Issue #5755, opened 2019-08-18
Music playback bugs/improvements


ToIGLAK:
I'm very happy this functionality was added to lazer (#5268)!
(and I'm referring to it saying keybindings)

Now, there are a few things I want to mention.
- osu intro (when you open osu) can be bugged if you switch song while the animation is playing (intro freezes)
- keybindings don't work in song select (using media keys)
- in song select, if you pause a track and change _**beatmap**_, music is not unpausing (it would be more intuitive if it did)

peppy:
Please file one issue per issue.

PR #5759, opened 2019-08-18
Prevent beatmap change during intro playback

Game4all:
teractions with Playback control.

Fixes first point of #5755


osu.Game/Screens/Menu/IntroScreen.cs
```diff
@@ -28,11 +28,18 @@ public abstract class IntroScreen : StartupScreen
...
+        private LeasedBindable<WorkingBeatmap> beatmap;
+
+        public new Bindable<WorkingBeatmap> Beatmap { get => beatmap; }
+
         protected override BackgroundScreen CreateBackground() => new BackgroundScreenBlack();
 
         [BackgroundDependencyLoader]
         private void load(OsuConfigManager config, BeatmapManager beatmaps, Framework.Game game)
         {
+            //we take a lease on the beatmap bindable to prevent music playback from changing / pausing music during intros, as it is causing weird interactions with certains intros
```
iiSaLMaN:
```suggestion
            // we take a lease on the beatmap bindable to prevent music playback from changing / pausing during intros, as it is causing weird interactions with certains intros.
```

osu.Game/Screens/Menu/IntroScreen.cs
```diff
@@ -108,6 +115,8 @@ protected void PrepareMenuLoad()
         protected void LoadMenu()
         {
             DidLoadMenu = true;
+            beatmap.Return(); //we return the lease to the beatmap bindable as we're pushing the main menu.
```
iiSaLMaN: unnecessary imo

frenzibyte (changes requested):

osu.Game/Screens/Menu/IntroScreen.cs
```diff
@@ -28,11 +28,18 @@ public abstract class IntroScreen : StartupScreen
 
         private Bindable<bool> menuVoice;
 
+        private LeasedBindable<WorkingBeatmap> beatmap;
+
+        public new Bindable<WorkingBeatmap> Beatmap { get => beatmap; }
```
peppy: i'd rather this didn't exist

osu.Game/Screens/Menu/IntroScreen.cs
peppy: you mean the comment, right?

peppy (changes requested):

osu.Game/Screens/Menu/IntroScreen.cs
iiSaLMaN: yeah

peppy:
I was actually looking at fixing this in a different way (fixing the triangles intro relying on the current playing track for timing), so this may not be necessary.

Game4all:
Closing this then

peppy:
This may still be something we want in extension, since it's weird UX to allow the track to be changed during the intro in the first place.

osu.Game/Screens/Menu/IntroScreen.cs
Game4all: I don't have a more elegant solution in mind other than shadowing Beatmap bindable as it can't be overriden.

peppy (approved):

PR #5762, opened 2019-08-19
Use a nearest value to stable's required spins

frenzibyte:
That's the nearest I got, but not accurate.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Spinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Spinner.cs
@@ -27,7 +27,7 @@ namespace osu.Game.Rulesets.Osu.Objects
             SpinsRequired = (int)(Duration / 1000 * BeatmapDifficulty.DifficultyRange(difficulty.OverallDifficulty, 3, 5, 7.5));
 
             // spinning doesn't match 1:1 with stable, so let's fudge them easier for the time being.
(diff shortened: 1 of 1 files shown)


73794449 (approved):
All Ok. A bit harder, but in normal ranges.

iiSaLMaN:
> All Ok. A bit harder, but in normal ranges.

Yeah I'm not quite sure about this, with my bonus spinning changes, i've realized the required spins are not accurate at all, and it seems a big issue for that change.

peppy:
What is this based on

iiSaLMaN:
> What is this based on

I did comparisons between stable and lazer on the bonus score gained from the spinner and I found that the required spins on lazer are a bit smaller than on stable, so I bumped up the value a bit

peppy:
The comment above the value you changed states this, though.

Did you test against multiple length spinners? If you did, you'll find that it is *impossible* to match stable as there is acceleration involved that makes the scoring non-linear.

iiSaLMaN:
> The comment above the value you changed states this, though.
> 
> Did you test against multiple length spinners? If you did, you'll find that it is _impossible_ to match stable as there is acceleration involved that makes the scoring non-linear.

In that case, would it be fine to adding bonus score even if it differs between stable and lazer?

peppy:
Yes, should be fine.

PR #5771, opened 2019-08-19
Add easing to osu! hit object fade in

### LINKED ISSUE #5770: Slider head and slider body uneven fade in opacity
**Describe the bug:** 
The slider head does not fade in with the same opacity as sliderbody
[https://streamable.com/jcl40](url)
**osu!lazer version:** 
2019.814.0

frenzibyte:
Closes #5770

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
@@ -36,7 +36,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 
         protected sealed override double InitialLifetimeOffset => HitObject.TimePreempt;
 
-        protected override void UpdateInitialTransforms() => this.FadeIn(HitObject.TimeFadeIn);
+        protected override void UpdateInitialTransforms() => this.FadeIn(HitObject.TimeFadeIn, Easing.Out);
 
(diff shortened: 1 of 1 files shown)


peppy:
What is this based on?

iiSaLMaN:
> What is this based on?

I'm not sure what you mean..?

peppy:
How are you finding that this is the correct value for easing? Have you done frame-by-frame comparisons? Are you referring to a wiki?

iiSaLMaN:
> How are you finding that this is the correct value for easing? Have you done frame-by-frame comparisons? Are you referring to a wiki?

I hope I can do frame-by-frame comparisons but it seems impossible with my current computer, 
I was gonna ask if BLuR can do comparisons to see if my PR makes it 1:1 to stable, should've linked the branch instead...

peppy:
You can't just guess at these values. I know for a fact this is 100% incorrect. There is no easting on hitobject fade in.

BLuRTuts:
If there is no easing the only thing I can think of is just a simple delay. If so then you'll probably have to look for the problem in the code. I could try upping the fps of the recording and finding a time value but that will be a janky solution.

peppy:
The original issue points out a potential discrepancy between sliders and hitcircles. You've changed the duration universally. I'm not sure how this works.

iiSaLMaN:
Will close this as an incorrect solution.

PR #5773, opened 2019-08-19
Avoid updating results of nested objects repeatedly

### LINKED ISSUE #1970: DrawableHitObject performs miss updates on nested hitobjects twice
`DrawableHitObject` performs `UpdateJudgement` in `UpdateAfterChildren`. This, in-turn, prior to checking for judgements on itself, performs `UpdateJudgement` on all `NestedHitObjects`.

So in the example:
```
HitObject: A
    -> HitObject: B
        -> HitObject: C
```

C will perform `UpdateJudgement` three times, B twice, and A once, all in the same frame. Beyond potential performance issues, this may be very problematic if judgements are added that are not marked as `Final` for the current time.

frenzibyte:
- Remove updating results of nested hit objects on `UpdateResult`, instead rely on `JudgementOccurred` since nested hit objects are updated before the main object, and it holds mostly the return value.

Closes #1970


peppy:
@smoogipoo requested you check this one, not 100% sure about it.

PR #5497, opened 2019-07-28
Make legacy skins animatable

jorolf:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/2738

This implements general handling for animatable components, however most components haven't been marked as animatable yet.

Closes #4627

PR #5776, opened 2019-08-20
Add slider ball animation support

peppy:
- adds support for the diffuse and specular layers used in `osu-stable`'s default skin.
- adds rotation support to the slider ball
- [x] Depends on #5497.

Not yet done (can be done in later PRs)

- Limit rotation application to only the main ball layer (not the lighting layers)
- Adjust frame rate based on slider velocity


osu.Game/Skinning/LegacySkin.cs
```diff
@@ -43,6 +43,7 @@ public class LegacySkin : Skin
         public LegacySkin(SkinInfo skin, IResourceStore<byte[]> storage, AudioManager audioManager)
             : this(skin, new LegacySkinResourceStore<SkinFileInfo>(skin, storage), audioManager, "skin.ini")
         {
+            if (!Configuration.CustomColours.ContainsKey("SliderBall")) Configuration.CustomColours["SliderBall"] = new Color4(2, 170, 255, 255);
```
smoogipoo: Maybe we ought to make a `LegacySkinConfiguration` instead? Because this is probably going to come up more in the future.

peppy: This isn't so simple to do as these defaults should only be applied via this particular constructor.

Unless you have any ideas, I can't see a low-overhead path forward to making this work with the current structure, so maybe best leaving for a future PR.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -344,6 +348,37 @@ private void load(ISkinSource skin)
...
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(ISkinSource skin, DrawableHitObject drawableObject)
+            {
+                animationContent.Colour = skin.GetValue<SkinConfiguration, Color4?>(s => s.CustomColours.ContainsKey("SliderBall") ? s.CustomColours["SliderBall"] : (Color4?)null) ?? Color4.White;
+
+                InternalChildren = new[]
+                {
+                    new Sprite
```
smoogipoo: This and the spec texture don't seem to be working with the stable default skin during gameplay.

smoogipoo (changes requested):

osu.Game/Skinning/LegacySkin.cs
peppy: they won't. they aren't included in the skin.

smoogipoo: Oh, ok then.

smoogipoo (approved):
This is fine for a first pass.

PR #5777, opened 2019-08-20
Apply osu! SpriteText text builder changes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/2781

### DIFF
--- a/osu.Game/Skinning/LegacySkin.cs
+++ b/osu.Game/Skinning/LegacySkin.cs
@@ -15,6 +15,7 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.Textures;
 using osu.Framework.IO.Stores;
+using osu.Framework.Text;
 using osu.Game.Database;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Sprites;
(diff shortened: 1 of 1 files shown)


osu.Game/Skinning/LegacySkin.cs
```diff
@@ -234,37 +235,50 @@ public void Dispose()
...
-                this.font = font;
-
                 Shadow = false;
                 UseFullGlyphHeight = false;
-            }
 
-            protected override Texture GetTextureForCharacter(char c)
-            {
-                string textureName = $"{font}-{c}";
+                Font = new FontUsage(font, 16);
```
peppy: where does 16 come from?

smoogipoo: Can be removed.

smoogipoo: Actually I take that back since the text builder isn't overridden with a custom size. It's just what it was previously (should be OsuFont.DEFAULT_FONT_SIZE).

peppy: so `Font = OsuFont.Default.With(font);` should be safest?

smoogipoo: That would also work fine, yeah.

PR #5779, opened 2019-08-20
Disallow adding bonus judgements to statistics

frenzibyte:
Matching stable behavior, bonus results should not count as they don't affect combo.


smoogipoo (approved):
This is fine for now. In the future we'll separate bonus and combo judgements because we _will_ want statistics for them.

PR #5780, opened 2019-08-20
Use correct type for value of text transforms

frenzibyte:
- Fixes text transforms throwing invalid operation exception.
- [ ] Depends on ppy/osu-framework#2783


iiSaLMaN:
Closing this PR as it's not the proper way.

PR #5728, opened 2019-08-15
Add other sample functions to SkinnableSound

frenzibyte:
- Adds `Stop()`, `AddAdjustment()`, `RemoveAdjustment()` functions to allow adjusting the playback of the `SkinnableSound`

- [x] Depends on ppy/osu-framework#2768

PR #5782, opened 2019-08-20
Implement gaining bonus scores from spinning a spinner

### LINKED ISSUE #7428: New HP Drain feature doesn't working well on spinner
Describe the bug:
-
As the title, HP doesn't regenerate by spinning spinner

-
[HP Drain.zip](https://github.com/ppy/osu/files/4018461/HP.Drain.zip)

**osu!lazer version:** 
-
Lazer 2019.1227.0

**Logs:** 
-
_No need I guess_

frenzibyte:
Resolves https://github.com/ppy/osu/issues/7428

Matching stable behavior (100 points for normal spins and 1100 for bonus spins)
https://streamable.com/7ec22

- [x] Depends on #5728
- [x] Depends on #5779 
- [x] Depends on #5795 
- [x] Depends on #6031 
- [x] Depends on #6040
- [x] Depends on #9635


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -190,6 +214,28 @@ protected override void UpdateAfterChildren()
...
+                if (i < 0 || i >= ticks.Count)
+                    break;
+
+                var tick = ticks[i];
+
+                tick.HasBonusPoints = Progress >= 1 && i > Spinner.SpinsRequired;
+
+                if (tick.HasBonusPoints)
+                    bonusCounter
+                        .TransformTextTo($"{(i - Spinner.SpinsRequired) * 1000}")
```
smoogipoo: As mentioned in the framework PR, we don't want these functions existing (and will probably be removed). You should just set `Text` instead.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
smoogipoo: You could try using `OnRevertResult`, but all this code needs to be more resilient to rewinding in general and you may not need to use it in the end.

iiSaLMaN: The way I was thinking this to work in terms of rewinding is that the rotation will decrement and the `currentSpin` would decrement with it, but the problem is that the rotation still increments because we're using `RotationAbsolute`, I proposed that spinners should stick to 1 direction, so you can spin in either direction, but you would stick to it and not change the direction (which no one does that), so that it would be implemented like this:
```csharp
if (direction == 0)
    int direction = Math.Sign(Disc.Rotation * 2); // current rotation is divided by 2 ( (see SpinnerDisc:L128)

int newSpins = (Disc.Rotation * direction / 360) - currentSpins;
```

but that doesn't seem like a good idea since not sure if we would want users to see their progress decrementing when spinning in another direction.

peppy: You can check whether elapsed time is decreasing or not as a potential avenue for handling this?

iiSaLMaN:
I've provided tests for this along with a bit of refactoring and catching up to latest master.

peppy:
The counter doesn't rewind, and the SPM shows NAN.

https://www.youtube.com/watch?v=e4JQC3BAXLc&feature=youtu.be

iiSaLMaN:
The SPM showing `NaN` is a separate issue where the Spinner SPM counter doesn't handle frozen clocks properly, will fix it.

peppy:
What about the counter rewinding issue? Is that going to be addressed in this PR?

iiSaLMaN:
Yeah but I don't have the time to look into it (final exams coming starting from tomorrow), will check once I can.

iiSaLMaN:
Just noting here: Slowly rewinding between spinner bonus ticks somehow keeps the bonus score in forever, even when all ticks are rewound back to 0, https://streamable.com/h5mnp.

Haven't identified the underlying bug behind this yet.

osu.Game.Rulesets.Osu/Objects/Spinner.cs
```diff
@@ -31,6 +33,15 @@ protected override void ApplyDefaultsToSelf(ControlPointInfo controlPointInfo, B
             SpinsRequired = (int)Math.Max(1, SpinsRequired * 0.6);
         }
 
+        protected override void CreateNestedHitObjects()
+        {
+            base.CreateNestedHitObjects();
+
+            var maximumSpins = OsuAutoGeneratorBase.SPIN_RADIUS * (Duration / 1000) / MathHelper.TwoPi;
```
smoogipoo: Is this basically a guess at the maximum number of ticks, or is this grounded in mathematical truth? I'm finding the 2pi value to be rather arbitrary here...

What if a user spins closer than the spin radius (which generally players that spin 477 do)? Is it possible for the user to out-spin the number of ticks?
At the moment the spinner isn't capped to a maximum speed (for better or worse) so it seems like this should be an issue.

I'm also wondering if we can add the ticks dynamically to avoid this issue, since they're counted as bonus as far as the score processor is concerned.

smoogipoo: fwiw you can test this with auto generation by changing the following two lines to be `/ 10` rather than `/ 20` in `OsuAutoGenerator.cs`:

image

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerBonusComponent.cs
```diff
@@ -0,0 +1,90 @@
...
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+                Font = OsuFont.Numeric.With(size: 24),
+                Alpha = 0,
+            };
+        }
+
+        private int currentSpins;
+
+        public void UpdateRotation(double rotation)
```
smoogipoo: Probably rename this to `SetRotation` to avoid adding one extra dialect to setting rotation, as previously mentioned.

peppy:
@iiSaLMaN planning on taking a look at applying reviews for this?

iiSaLMaN:
> @iiSaLMaN planning on taking a look at applying reviews for this?

Yeah, but after I finish up my current task (ppy/osu-framework#1618), will work on this in about a week. (I need to look for implementing dynamically-added nested hit objects)

peppy:
As we need this sooner rather than later I'm going to take control of this PR and figure the direction forward for it.

osu.Game.Rulesets.Osu/Objects/Spinner.cs
peppy: You know, I think this could be a pretty cool way of limiting sliders: removing the max rpm limitation but limiting the maximum bonus score. @smoogipoo thoughts?

smoogipoo: What do you achieve by doing that? In the first place I can't spin >=477 so I don't have the ability to determine what the maximum bonus points should be by myself.
The situation I'd worry about is underestimating that value and then causing someone who spins 430 to be valued the same as someone that spins 350. Or in lazer terms, someone spinning like 750 vs someone spinning 550.

peppy: Just another step in the direction of removing complex interactions that really don't need to be there.

My thinking is that it will make calculations much simpler (and also simplify implementation of spinners as it would be 1:1 to input rather than artificially limited at a per-frame basis).

We've seen how annoying spinners can be with stable; part of that comes from the per-frame limiting. Even with our existing replay system we would be potentially losing data – input/update at 1000hz but recording far lower (obviously we could attach as extra data in the replay frames, but that makes things more complex once again).

If we were to remove the limiter, it is 1:1 all the way, and the SPM meter would also match near-perfectly where that would otherwise not be possible.

The value can/would of course be tweaked to ensure what you say doesn't happen.

smoogipoo: Sure, let's see how it behaves for now.

enoslayd:
Question

1. Should the bonus point be positioned on bottom?
2. Does this include the health portion?

peppy:
Ready for initial review

frenzibyte (commented):
Looks nicer than my old code, can't tell if it's because I wasn't much good at contributing in that time and can do better now 😛; anyways here's a couple of things I found (one of which seems major):

osu.Game.Rulesets.Osu/Objects/Spinner.cs
```diff
@@ -26,6 +26,8 @@ public double EndTime
         /// </summary>
         public int SpinsRequired { get; protected set; } = 1;
 
+        public int MaximumBonusSpins => SpinsRequired;
```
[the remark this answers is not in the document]
peppy: Feel free to suggest a value that more closely matches expectations.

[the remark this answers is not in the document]
smoogipoo: Maybe assume an arbitrary rotations-per-second (`R=477/60` to keep in line with stable) and use the formula 
`BonusSpins = R * Duration / 1000 - SpinsRequired`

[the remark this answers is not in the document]
peppy: `SpinsRequired` is already based on arbitrary rotation, which `BonusSpins` is based on, so the only change to make would be the ratio applied. I'll roughly match stable at 477 for now I guess.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -156,6 +199,10 @@ protected override void CheckForResult(bool userTriggered, double timeOffset)
             if (userTriggered || Time.Current < Spinner.EndTime)
                 return;
 
+            // Trigger a miss result for remaining ticks to avoid infinite gameplay.
+            foreach (var tick in ticks.Where(t => !t.IsHit))
+                tick.TriggerResult(false);
```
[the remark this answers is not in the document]
peppy: no.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -185,15 +232,50 @@ protected override void UpdateAfterChildren()
...
         }
 
+        private int wholeSpins;
+
+        private void updateBonusScore()
+        {
+            if (ticks.Count == 0)
+                return;
+
+            int spins = (int)(Disc.CumulativeRotation / 360);
```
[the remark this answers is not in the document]
peppy: i think it's fine

osu.Game.Rulesets.Osu/Objects/Spinner.cs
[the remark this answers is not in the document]
peppy: Updated to roughly match stable. Roughly.

peppy:
@smoogipoo requesting re-review.

osu.Game.Rulesets.Osu/Objects/Spinner.cs
```diff
@@ -26,7 +26,10 @@ public double EndTime
         /// </summary>
         public int SpinsRequired { get; protected set; } = 1;
 
-        public int MaximumBonusSpins => SpinsRequired;
+        /// <summary>
+        /// Number of spins available to give bonus, beyond <see cref="SpinsRequired"/>.
+        /// </summary>
+        public int MaximumBonusSpins => (int)(SpinsRequired * 1.8f); // roughly matches stable
```
smoogipoo, line 32: This is only true at OD10. It's 2.5x less than maximum for stable at OD0.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Spinner.cs
smoogipoo: I think the true value is `7.95 / ({post-od value} * 0.6)`, which could be rounded up (just don't want to under-estimate stable).

peppy: Are we actually looking to match stable 100%? stable's spin calculations are absolutely impossible to understand as a mapper or player IMO.

smoogipoo: I'm only concerned about under-estimating. As it is, as long as you spin faster than 200RPM you'll run out of bonus ticks at OD0. At OD5 it's 324.

peppy: So you think lower OD should have more bonus spins? Rather than being an understandable ratio from base spins?

smoogipoo: If you wanted to go down that route you'd have to convince me that OD0 and OD10 should have the same amount of maximum number of spins. So long as you can't simplify that, I see no reason to simplify the number of bonus spins as if people are going to care about that and not the minimum RPM they have to spin to finish the spinner in the first place.

smoogipoo: I think it's actually quite understandable anyway.

```
// 2-part linear curve between OD0 and OD10
MinimumRPS = 0.6 * { 3, 5, 7.5 }

RequiredSpins = RPS * Duration

// Capped at 480RPM
MaximumRPS = 8
BonusSpins = MaximumRPS / RPS
```

peppy: So as OD drops, the total score attainable for a map will increase beyond 1mil?

smoogipoo: Sure. Is that what you're trying to simplify?

peppy: I'm just thinking we should reconsider spinning as we are with most other elements, because that seems really weird to me (easy mod score will be higher than no-mod due to bonus spin count). Maybe it makes sense? I'm not sure. We can go with it for now if you think that's best.

smoogipoo: That's impossible, because EZ halves your multiplier anyway. Meaning you'd need to spin ~450x more to get a higher score than a nomod player just through bonus points, which just isn't possible via the multipliers here.

Unless you're thinking mod multiplier shouldn't be applied (as we once discussed), at which point separate leaderboards or a completely different leaderboard system altogether needs to be envisioned. I don't really want to get into all that in this thread though.

How about from a player's perspective - most new players will flail their mouse around spinning, and I can't get any lower than 240-280RPM (lazer) by doing that. Considering most newbie maps are around OD2-OD4, you're looking at around max 305RPM to get all bonus points. So once a new player does anything different they're going to very quickly cap out and have nothing to work towards there.

peppy: Yeah, thinking no-multipliers. Not that we are going to ever compare 1:1, but still feels a bit weird. Let's go with matching for now and we can re-discuss at a later point with more consideration.

peppy: I've implemented and tested the code your specified, but it feels further off stable than what I had. Can you take a look? Do we need to increase the "max spin" rate?

smoogipoo:
I haven't looked into the reason why but this doesn't seem to handle rewinding all that well. It's very easy to make the counter go negative.

peppy:
Which counter? The bonus display?

smoogipoo:
Yeah just the bonus display. It doesn't look to affect the score.

peppy:
Just going to get this in for now. Can fix replay along with the many other replay issues still remaining.

PR #5789, opened 2019-08-21
Ensure triangle intro can complete even when no audio device is present

peppy:


peppy:
Don't merge this one yet – on testing this will not (alone) solve the issue.

peppy:
Closing for now.

PR #5793, opened 2019-08-21
Add android build automation via fastlane

peppy:
PRing for further testing. Currently creates a very small apk file.

Note that running fastlane is roughly equivalent to `MSBuild "/t:restore" "/t:PackageForAndroid" "/p:Configuration=Release" "osu.Android/osu.Android.csproj"` for testing purposes (both generate the incorrect apk).

- [x] Depends on https://github.com/ppy/osu/pull/6238.


miterosan:
The required msbuild Target is SignAndroidPackage

Not sure where you got PackageForAndroid from.

peppy:
`SignAndroidPackage` has the same result. I left it as `PackageForAndroid` as this creates an apk that can be extracted using `zip` for inspection.

smoogipoo (approved):

PR #5781, opened 2019-08-20
Expand hit sample lookup names

frenzibyte:
- Expand lookup names to allow sample names without a bank (e.g `spinnerbonus`)

PR #5795, opened 2019-08-21
Try retrieving samples without bank names

frenzibyte:
Allow non-bank samples to have banks (e.g `spinnerbonus`) as suggested by @smoogipoo.
Supersedes #5781 

- [x] Depends on ppy/osu-resources#71


peppy:
Is there a reason this couldn't have been added in `LookupNames` rather than this hacky method?

iiSaLMaN:
> Is there a reason this couldn't have been added in `LookupNames` rather than this hacky method?

I don't understand what you mean? we were planning on adding banks for these samples and currently redirect them to non-bank names, doing that in `LookupNames` will force those samples to stay non-bank.

peppy:
Check `HitSampleInfo` class. Not only is the way you have it non-intuitive, but you're actually changing the order of lookups for the worst. I believe the non-bank lookups should only occur at the end.

peppy:
This surfaces a larger concern: we are likely doing sample lookup incorrectly at the moment. We are checking each `LookupName` against the whole skin source hierarchy and matching the first match, whereas it should likely perform *all* fallback matches before checking the next skin source.

To correctly support this, `ISkin` should probably be changed to:

```csharp
SampleChannel GetSample(ISampleInfo sample);
```

An implementation in LegacySkin would do:
```csharp
        public override SampleChannel GetSample(ISampleInfo sample)
        {
            foreach (var n in sample.LookupNames)
            {
                var s = Samples.Get(n);
                if (s != null) return s;
            }

            if (sample is HitSampleInfo hsi)
            {
                // create name without bank without string manipulation.
                // return if available.
            }

            return null;
        }
```

iiSaLMaN:
What about the default skin, we would need the lookup names for it as well?

peppy:
why would it need to change at all from the custom implementation? i don't think that is in scope.

iiSaLMaN:
> why would it need to change at all from the custom implementation? i don't think that is in scope.

Aren't we moving the lookup names loop to `LegacySkin.GetSample(info)`? we would need to do lookup names loop at the default skin as well, for example: taiko hit samples have a namespace `Taiko` while other rulesets don't require any namespace, how can we know if we should load with a namespace or not

peppy:
because you look at `osu-resources` and see what is available?

one could argue we eventually move the resource retrieval to the `DefaultSkin` class if that results in better code quality.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -212,7 +213,22 @@ public override Texture GetTexture(string componentName)
...
+            {
+                var sample = Samples.Get(lookup);
+
+                if (sample != null)
+                    return sample;
+            }
+
+            if (sampleInfo is HitSampleInfo hsi)
+                // Try fallback to non-bank samples.
+                return Samples.Get($"{hsi.Name}");
```
peppy: probably doesn't need string literal

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -43,26 +42,19 @@ protected override void SkinChanged(ISkinSource skin, bool allowFallback)
...
             {
-                var ch = loadChannel(s, skin.GetSample);
+                var ch = skin.GetSample(s);
+
                 if (ch == null && allowFallback)
-                    ch = loadChannel(s, audio.Samples.Get);
-                return ch;
-            }).Where(c => c != null).ToArray();
-        }
+                    if (s is HitSampleInfo hsi)
```
peppy: can you explain why this is used here instead of `lookupnames`?

iiSaLMaN: i thought we're not gonna use the lookup names on default skin? are you fine with a duplicate loop from `LegacySkin`?

peppy: please do.

peppy (approved):

PR #5800, opened 2019-08-22
Fix IDE0008 (use explicit type instead of var) to modernize the codebase

ghost:
Every change was done automatically by the IDE. This is being done to ensure the codebase meets modern C# standards. No behavioral changes.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
+++ b/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
@@ -13,7 +13,7 @@ namespace osu.Game.Rulesets.Osu.Edit
     {
         public override void HandleDrag(SelectionBlueprint blueprint, DragEvent dragEvent)
         {
-            foreach (var h in SelectedHitObjects.OfType<OsuHitObject>())
+            foreach (OsuHitObject h in SelectedHitObjects.OfType<OsuHitObject>())
             {
                 if (h is Spinner)
                 {
--- a/osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
+++ b/osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
@@ -73,7 +73,7 @@ namespace osu.Game.Rulesets.Osu.UI
 
         private void addApproachCircleProxy(Drawable d)
         {
-            var proxy = d.CreateProxy();
+            Drawable proxy = d.CreateProxy();
             proxy.LifetimeStart = d.LifetimeStart;
             proxy.LifetimeEnd = d.LifetimeEnd;
             approachCircles.Add(proxy);
--- a/osu.Game/Database/MutableDatabaseBackedStore.cs
+++ b/osu.Game/Database/MutableDatabaseBackedStore.cs
@@ -50,7 +50,7 @@ namespace osu.Game.Database
         /// <param name="item">The item to update.</param>
         public void Update(T item)
         {
-            using (var usage = ContextFactory.GetForWrite())
+            using (DatabaseWriteUsage usage = ContextFactory.GetForWrite())
                 usage.Context.Update(item);
 
             ItemRemoved?.Invoke(item);
@@ -131,15 +131,15 @@ namespace osu.Game.Database
         {
             using (var usage = ContextFactory.GetForWrite())
             {
-                var context = usage.Context;
+                OsuDbContext context = usage.Context;
 
-                var lookup = context.Set<T>().Where(s => s.DeletePending);
+                IQueryable<T> lookup = context.Set<T>().Where(s => s.DeletePending);
 
                 if (query != null) lookup = lookup.Where(query);
 
                 lookup = AddIncludesForDeletion(lookup);
 
-                var purgeable = lookup.ToList();
+                List<T> purgeable = lookup.ToList();
 
                 if (!purgeable.Any()) return;
 
--- a/osu.Game/Graphics/Backgrounds/Triangles.cs
+++ b/osu.Game/Graphics/Backgrounds/Triangles.cs
@@ -228,7 +228,7 @@ namespace osu.Game.Graphics.Backgrounds
                 {
                     var offset = triangle_size * new Vector2(particle.Scale * 0.5f, particle.Scale * 0.866f);
 
-                    var triangle = new Triangle(
+                    Triangle triangle = new Triangle(
                         Vector2Extensions.Transform(particle.Position * size, DrawInfo.Matrix),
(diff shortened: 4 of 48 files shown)


osu.Game/Graphics/UserInterface/OsuSliderBar.cs
```diff
@@ -129,10 +129,10 @@ protected override void OnUserChange(T value)
 
         private void playSample(T value)
         {
-            if (Clock == null || Clock.CurrentTime - lastSampleTime <= 50)
```
peppy: were these changes also automated? could you tell us which plugin / refactoring tool you used to perform these?

peppy:
Can you explain where these var changes were made? There are still many existing (potentially ambiguous) usages like:

image

where one may argue that not using `var` is more readable.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -86,7 +86,7 @@ public override void PostProcess()
 
         private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
         {
-            if (!judgedObject.DisplayResult || !DisplayJudgements.Value)
+            if (!(judgedObject.DisplayResult && DisplayJudgements.Value))
```
iiSaLMaN: how is this related?

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -13,7 +13,7 @@ public class OsuSelectionHandler : SelectionHandler
     {
         public override void HandleDrag(SelectionBlueprint blueprint, DragEvent dragEvent)
         {
-            foreach (var h in SelectedHitObjects.OfType<OsuHitObject>())
+            foreach (OsuHitObject h in SelectedHitObjects.OfType<OsuHitObject>())
```
iiSaLMaN: is there a problem with using `var` here? it can be clearly seen that the type is `OsuHitObject` because of the filtering `Enumerable.OfType<FooType>()`

osu.Game/Database/MutableDatabaseBackedStore.cs
```diff
@@ -131,15 +131,15 @@ public void PurgeDeletable(Expression<Func<T, bool>> query = null)
...
 
-                var lookup = context.Set<T>().Where(s => s.DeletePending);
+                IQueryable<T> lookup = context.Set<T>().Where(s => s.DeletePending);
 
                 if (query != null) lookup = lookup.Where(query);
 
                 lookup = AddIncludesForDeletion(lookup);
 
-                var purgeable = lookup.ToList();
+                List<T> purgeable = lookup.ToList();
```
iiSaLMaN: also here, we're converting the `IQueryable<T>` to list, isn't it obvious that the type would be `List<T>`

osu.Game/Graphics/Backgrounds/Triangles.cs
```diff
@@ -228,7 +228,7 @@ public override void Draw(Action<TexturedVertex2D> vertexAction)
                 {
                     var offset = triangle_size * new Vector2(particle.Scale * 0.5f, particle.Scale * 0.866f);
 
-                    var triangle = new Triangle(
+                    Triangle triangle = new Triangle(
```
iiSaLMaN: we're creating an instance of `Triangle`, is there a reason not to use `var` here as well?

frenzibyte (changes requested):

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
pi1024e: Undid that.

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
pi1024e: To make it clear it is a hit object, because "var h" is not a clear enough

osu.Game/Database/MutableDatabaseBackedStore.cs
pi1024e: There are many subclasses of List<T>, so yes, it would not be as obvious as you think.

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
peppy: it's made clear by the source in this case.

osu.Game/Graphics/Backgrounds/Triangles.cs
pi1024e: The object being returned is a Triangle, so making it Triangle instead of var makes it stand out.

iiSaLMaN: this makes no sense, you're returning a `Triangle`, we would expect `var` to be `Triangle`?

osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapConverter.cs
```diff
@@ -22,11 +22,11 @@ public CatchBeatmapConverter(IBeatmap beatmap)
 
         protected override IEnumerable<CatchHitObject> ConvertHitObject(HitObject obj, IBeatmap beatmap)
         {
-            var curveData = obj as IHasCurve;
-            var positionData = obj as IHasXPosition;
-            var comboData = obj as IHasCombo;
-            var endTime = obj as IHasEndTime;
-            var legacyOffset = obj as IHasLegacyLastTickOffset;
+            IHasCurve curveData = obj as IHasCurve;
```
iiSaLMaN: it's casted to `IHasCurve`, why is it not var...?

iiSaLMaN:
Please test your changes locally before pushing to reduce failed builds, you can do that by running `build.ps1` for windows or `build.sh` for mac/linux

peppy:
Please do not force push after making a PR.

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
peppy: please don't resolve conversations

peppy:
This is mostly non-required changes. Our current policy on `var` usage is as follows:

- If the type can be discerned from the line (ie. there is `as Type` or `new Type`) we use `var`
- If the type is a primitive type, we don't use var.

Anywhere else we use var or not var (it's personal preference). Generally we use `var` for longer type names or to avoid type repetition.

I'm going to close this PR. Feel free to open another one following the above guidelines if you find any that you feel are out of line.

PR #5801, opened 2019-08-22
Fix video decoding loop running permanently in the background

peppy:


swoolcock:
Nice catch.  I'm interested to know what kind of a performance hit that was causing, as the decoding loop needs some work anyway.

peppy:
Yes, may be something we want to look into in more detail.

image

Note that the total time can be disregarded as it's likely incorrectly measured sleep time (I hope, anyway). The GC overhead should probably be investigated at very least.

peppy:
(and the queue access)

swoolcock (approved):

PR #5804, opened 2019-08-22
Fix IDE0008 (use explicit type instead of var) to modernize the codebase

ghost:
I did this again, but only with primitive types and few objects to comply with the current var policy on this project.


peppy:
If you are going to make commit messages like that, please don't contribute to this project.

pi1024e:
That PR message was supposed to be squashed and edited. I apologize about that.

PR #5813, opened 2019-08-23
Some clean.

73794449:


osu.Game/Graphics/Containers/OsuScrollContainer.cs
```diff
@@ -95,6 +95,7 @@ protected class OsuScrollbar : ScrollbarContainer
 
             private readonly Box box;
 
+            private int FadeDuration = 100;
```
turbedi: `private const int FadeDuration = 100;`

iiSaLMaN: private const double fade_duration = 100;

peppy:
Please run against `build.sh` before PRing for stuff liek this.

PR #5816, opened 2019-08-23
Save user's group and sort filter choices

### LINKED ISSUE #5389: Group & Sort filters don't save user's choice
**Describe the bug:** Group and Sort filters don't keep the user's choice when re-entering SongSelect.

**Screenshots or videos showing encountered issue:** https://streamable.com/1xrng

**osu!lazer version:** 2019.717.1

Game4all:
Closes #5389


frenzibyte (commented):
Would be better to set `SortMode` and `GroupMode` to `protected` and expose that in the tests since that's where it's used

osu.Game/Configuration/OsuConfigManager.cs
```diff
@@ -150,6 +154,8 @@ public enum OsuSetting
         SaveUsername,
         DisplayStarsMinimum,
         DisplayStarsMaximum,
+        SelectGroupingMode,
```
peppy: `SongSelect`

peppy (approved):

Issue #4929, opened 2019-06-05
Difficulties do not wrap correctly on the BeatmapSetOverlay or DirectOverlay


welsar55:
**Describe the bug:** 
On sets with the 21 or more difficulties (e.g. Sotark's Quaver) difficulties flow off the right-hand side of the screen.

**Screenshots or videos showing encountered issue:** 
![osu_2019-06-05_12-12-06](https://user-images.githubusercontent.com/6157330/58975711-420f1780-878b-11e9-9385-c21f54c83d0d.png)

**osu!lazer version:** 
2019.530.0

lunafreya:
I believe this might be fixed with the new beatmap page design to be implemented both on o!web and the new o!client, pending approval and testing. Right, @peppy and @arflyte?

smoogipoo:
For now it should wrap as on o!web: https://osu.ppy.sh/beatmapsets/873811

arphox:
Performing this change helps with the wrapping but causes additional problems: contents overlap. I'll investigate it further soon.
![kép](https://user-images.githubusercontent.com/17320094/59377100-d102cd80-8d51-11e9-8513-254bad0d81f2.png)

welsar55:
Just discovered the same issue is present in the direct overlay. 
![Capture](https://user-images.githubusercontent.com/6157330/59459726-14f1e300-8de3-11e9-842d-a016fb028a68.PNG)

iiSaLMaN:
Thanks, will include the fix in the PR, i'll handle it just like the web version
![image](https://user-images.githubusercontent.com/22781491/59459887-899a4100-8e26-11e9-892f-65c22c69f53b.png)

Joehuu:
Fixed in referencing PRs.

PR #5823, opened 2019-08-23
Implement grouped difficulty icons

frenzibyte:
- This fixes direct panels having out-of-bounds difficulties by replacing them with counters for each ruleset, Matching the web for now (same for carousel).
- [x] Depends on #5822

Fixes panel part of #4929

### DIFF
--- a/osu.Game/Beatmaps/Drawables/GroupedDifficultyIcon.cs
+++ b/osu.Game/Beatmaps/Drawables/GroupedDifficultyIcon.cs
@@ -11,6 +11,12 @@ using osuTK.Graphics;
 
 namespace osu.Game.Beatmaps.Drawables
 {
+    /// <summary>
+    /// A difficulty icon that contains a counter on the right-side of it.
+    /// </summary>
+    /// <remarks>
+    /// Used in cases when there are too many difficulty icons to show.
+    /// </remarks>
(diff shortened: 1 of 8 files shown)


osu.Game/Beatmaps/Drawables/GroupedDifficultyIcon.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Linq;
+using osu.Framework.Graphics;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
+using osu.Game.Rulesets;
+using osuTK.Graphics;
+
+namespace osu.Game.Beatmaps.Drawables
+{
+    public class GroupedDifficultyIcon : DifficultyIcon
```
peppy: could we please at very least xmldoc classes describing what they are when the name is not enough to do so?

iiSaLMaN: oh sorry, i forgot to add xmldocs

peppy:
I've basically rewritten most pieces of this into a usable state. Please make sure to go through my commits and check the changes one by one.

peppy (approved):

peppy:
Also as a side-note, this should have been split into two PRs - one for the direct overlay and one for song select.

PR #5833, opened 2019-08-26
Add ruleset-level legacy skin provider

### LINKED ISSUE #5786: Make LegacySkin implementations per-ruleset
Currently it is not able to DI a DrawableHitObject and cast to a ruleset specific type, as `LegacySkin` is in the `osu.Game` project.

The proposal would be to have a method of making per-ruleset legacy skin implementations to work around this.

peppy:
Closes #5786.

This allows a ruleset to provide a custom implementation for legacy skinning.

The hierarchy is a bit convoluted from the change (due to the source being the parent `SkinProvidingContainer`) but this is the best way I can find for this for now to ensure correct lookups.


osu.Game/Skinning/BeatmapSkinProvidingContainer.cs
```diff
@@ -0,0 +1,38 @@
...
+    /// <summary>
+    /// A container which overrides existing skin options with beatmap-local values.
+    /// </summary>
+    public class BeatmapSkinProvidingContainer : SkinProvidingContainer
+    {
+        private readonly Bindable<bool> beatmapSkins = new Bindable<bool>();
+        private readonly Bindable<bool> beatmapHitsounds = new Bindable<bool>();
+
+        protected override bool AllowConfigurationLookup => beatmapSkins.Value;
+        protected override bool AllowDrawableLookup(string componentName) => beatmapSkins.Value;
```
smoogipoo: Seems to break beatmap skins on first load (if you uncheck-check the checkbox it'll work). Probably due to load() order of the hierarchy.

smoogipoo (changes requested):
Looks ok otherwise

osu.Game/Skinning/BeatmapSkinProvidingContainer.cs
peppy: Fix via loading `DrawableRuleset` in a second pass.

peppy:
Hold off on merging this one briefly. Going to look into how easy it is to add a player-based test against beatmap/skin providers.

peppy:
Tests added.

osu.Game.Rulesets.Osu.Tests/TestSceneOsuPlayer.cs
```diff
@@ -1,17 +1,157 @@
...
+            {
+                this.skin = skin;
+            }
+
+            protected override ISkin GetSkin() => skin;
+        }
+
+        public class SkinProvidingPlayer : TestPlayer
+        {
+            private readonly TestSource userSkin;
```
iiSaLMaN:
```suggestion
            [Cached(typeof(ISkinSource))]
            private readonly TestSource userSkin;
```

smoogipoo (approved):
Lgtm

PR #5836, opened 2019-08-26
Implement relative timing point scaling for scrolling rulesets

### LINKED ISSUE #5802: change scroll rate values in osu!mania to be a multiplier if it is BPM scale
i suggest that the scroll rate in osu!mania to change from a ms value to a multiplier on BPM scale because a ms value is inaccurate because the BPM also changes the scroll speed

smoogipoo:
Resolves #5802 

Prereqs:
- [x] #5834
- [x] #5835

In osu!stable mania, timing points in non-bpm-scale mode are relatively-scaled between each other, based on the "most common" BPM by duration.

Only implemented as an option on `DrawableScrollingRuleset` for now - we could make it optional in the future, but I need to look into the invalidation pathway for that.


peppy:
Looks sane. We may need to tweak the default settings going forward.

peppy (approved):

PR #5846, opened 2019-08-26
Show storyboard icon on beatmap wedge

### LINKED ISSUE #5752: display on the beatmap block on the beatmap carasoul the storyboard/video symbol if it has one
It would display a symbol showing that it has a storyboard/video if it has one like the website

The symbol should be put next to the ranked status

EVAST9919:
Partially fixes https://github.com/ppy/osu/issues/5752

### DIFF
--- a/osu.Game/Screens/Select/BeatmapInfoWedge.cs
+++ b/osu.Game/Screens/Select/BeatmapInfoWedge.cs
@@ -26,6 +26,7 @@ using osu.Framework.Localisation;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.UI;
+using osu.Game.Overlays.Direct;
 
 namespace osu.Game.Screens.Select
 {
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -131,6 +132,7 @@ public class BufferedWedgeInfo : BufferedContainer
             public OsuSpriteText TitleLabel { get; private set; }
             public OsuSpriteText ArtistLabel { get; private set; }
             public BeatmapSetOnlineStatusPill StatusPill { get; private set; }
+            public IconPill StoryboardPill { get; private set; }
```
peppy: public but untested? the other attributes are public with reason.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -214,11 +216,16 @@ private void load(LocalisationManager localisation)
                         Name = "Topright-aligned metadata",
                         Anchor = Anchor.TopRight,
                         Origin = Anchor.TopRight,
-                        Direction = FillDirection.Vertical,
+                        Direction = FillDirection.Horizontal,
                         Margin = new MarginPadding { Top = 14, Left = 10, Right = 18, Bottom = 20 },
```
peppy: needs spacing, probably 3 pixels or so

peppy (changes requested):
missing any method of testing

peppy:
On top of mentioned issues, this is causing a full load of the storyboard each time a beatmap is selected at song select. This cannot happen.

PR #5847, opened 2019-08-27
Implement local sorting by name and country in SocialOverlay

### LINKED ISSUE #4995: "Friends" not ordering based on selected tab
**Describe the bug:** 
Lazer doesn't order "Friends" by the tab selected on the top(which I am very sure orders them by either rank, name, or location)

osu_2019-06-11_18-25-31

**osu!lazer version:** 
2019.611.0 (Latest)

**Logs:** (I don't think they are necessary but will send if asked to)

EVAST9919:
Partially fixes https://github.com/ppy/osu/issues/4995


osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -166,20 +170,68 @@ private void recreatePanels(PanelDisplayStyle displayStyle)
...
+            IEnumerable<User> sortedUsers = newUsers;
+
+            if (sortedUsers.Any())
+            {
+                switch (Filter.Tabs.Current.Value)
+                {
+                    case SocialSortCriteria.Location:
+                        switch (sortDirection)
+                        {
+                            case SortDirection.Ascending:
```
peppy: why not use `.Reverse()` for ascending/descending rather than duplicating this stuff? or if you want to avoid using `Reverse` (potentially requires more iteration than otherwise), store the predicate to a variable and at very least only have the `OrderBy` / `OrderByDescending` statements once.

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -166,20 +171,51 @@ private void recreatePanels(PanelDisplayStyle displayStyle)
...
+                loading.Hide();
                 ScrollFlow.Add(panels = newPanels);
             });
         }
 
         private void updateUsers(IEnumerable<User> newUsers)
         {
-            Users = newUsers;
-            loading.Hide();
+            loading.Show();
```
peppy: this looks wrong?

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -166,20 +171,51 @@ private void recreatePanels(PanelDisplayStyle displayStyle)
...
+
+            var sortDirection = Filter.DisplayStyleControl.Dropdown.Current.Value;
+
+            IEnumerable<User> sortedUsers = newUsers;
+
+            if (sortedUsers.Any())
```
peppy: can't this just early return?

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -166,20 +171,51 @@ private void recreatePanels(PanelDisplayStyle displayStyle)
...
         }
 
+        private IEnumerable<User> sortBy<T>(IEnumerable<User> users, Func<User, T> condition, SortDirection sortDirection) =>
+            sortDirection == SortDirection.Ascending ? users.OrderBy(condition) : users.OrderByDescending(condition);
+
+        private void onDropdownChanged()
+        {
+            if (Users == null)
+            {
+                queueUpdate();
```
peppy: i don't think this needs to exist. either there is an already pending update or there are no users, no?

peppy (changes requested):

osu.Game/Overlays/SocialOverlay.cs
EVAST9919: Nope, we need that because `Clear` function call happens in `recreatePanels` which will be called after sorting

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -71,38 +72,37 @@ public SocialOverlay()
...
-        private ScheduledDelegate queryChangedDebounce;
-
         private void queueUpdate() => Scheduler.AddOnce(updateSearch);
 
+        private CancellationTokenSource loadCancellation;
+
         private void updateSearch()
         {
-            queryChangedDebounce?.Cancel();
+            Scheduler.CancelDelayedTasks();
```
peppy: I'm hesitant about this. The scheduler could be used by `SearchableListOverlay` or `CompositeDrawable` and this would be potentially cancelling scheduled tasks that we are not responsible for.

peppy (changes requested):

peppy:
This seems to still be doing a server request even though it's locally filtering. Is that intended?

smoogipoo:
@peppy requesting review since I've changed a lot of the code in here to fix sorting issues when changing the ascending/descending mode.

smoogipoo (dismissed):

PR #5848, opened 2019-08-27
Use automatic state logic for taiko ruleset

frenzibyte:


smoogipoo (approved):

smoogipoo:
@iiSaLMaN In general, don't use `InitialLifetimeOffset` for timing purposes - it should only ever be used for optimisation.

Issue #5841, opened 2019-08-26
Use new automatic state logic for all rulesets


iiSaLMaN:
Currently, rulesets other than osu! use the legacy `UpdateState()` function.

- [x] Taiko ruleset (see #5848)
- [x] Catch ruleset (see #5842)
- [x] Mania ruleset (see #5844)

PR #5849, opened 2019-08-27
Remove DrawableHitObject legacy state update function

frenzibyte:
Removes legacy `UpdateState()` function and `UseTransformStateManagement` property in `DrawableHitObject`.
- [x] Depends on #5841


smoogipoo (approved):

smoogipoo:
On further discussion, we've decided to obsolete these members for the next 6 months.

iiSaLMaN:
> On further discussion, we've decided to obsolete these members for the next 6 months.

In this case, I think it would be better to reopen this and mark it as blocked until the 6 months pass.

peppy:
pointless imo

PR #5778, opened 2019-08-20
Add kudosu history to UserProfileOverlay

EVAST9919:
- [x] Depends on https://github.com/ppy/osu/pull/5858
Screenshot_1

PR #5858, opened 2019-08-27
Refactor PaginatedContainer to avoid code duplication

EVAST9919:
Split from https://github.com/ppy/osu/pull/5778

### DIFF
--- a/osu.Game/Overlays/Profile/Sections/PaginatedContainer.cs
+++ b/osu.Game/Overlays/Profile/Sections/PaginatedContainer.cs
@@ -116,7 +116,7 @@ namespace osu.Game.Overlays.Profile.Sections
                     return;
                 }
 
-                LoadComponentsAsync(items.Where(AllowCreate).Select(CreateDrawableItem), drawables =>
+                LoadComponentsAsync(items.Select(CreateDrawableItem).Where(d => d != null), drawables =>
                 {
                     missingText.Hide();
                     moreButton.FadeTo(items.Count == ItemsPerPage ? 1 : 0);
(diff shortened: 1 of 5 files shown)


osu.Game/Overlays/Profile/Sections/PaginatedContainer.cs
```diff
@@ -11,22 +11,27 @@
...
 using osu.Game.Rulesets;
 using osu.Game.Users;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
 
 namespace osu.Game.Overlays.Profile.Sections
 {
-    public abstract class PaginatedContainer : FillFlowContainer
+    public abstract class PaginatedContainer<T> : FillFlowContainer
```
iiSaLMaN: Ensure given argument is an `APIRequest`
```suggestion
    public abstract class PaginatedContainer<T> : FillFlowContainer
        where T : APIRequest
```

iiSaLMaN:
```suggestion
    public abstract class PaginatedContainer<TRequest> : FillFlowContainer
```

EVAST9919: Please read the code. T is not an api request. T is response of created request.

iiSaLMaN: oh sorry, nvm this then

EVAST9919: May be `TAPIResponse` would be better

smoogipoo (approved):

PR #5865, opened 2019-08-28
 Fix osu! approach circles fading in too late

peppy:
A fade was being applied to the object itself *in addition* to the approach circle. For sliders, there were two external fades being applied in addition (3 total).

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.TypeExtensions;
+using osu.Framework.Graphics;
 using osu.Framework.Graphics.Primitives;
 using osu.Game.Audio;
 using osu.Game.Rulesets.Judgements;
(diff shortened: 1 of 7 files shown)


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -186,13 +187,16 @@ private void updateState(ArmedState newState, bool force = false)
...
         /// Apply (generally fade-in) transforms leading into the <see cref="HitObject"/> start time.
         /// The local drawable hierarchy is recursively delayed to <see cref="LifetimeStart"/> for convenience.
+        ///
+        /// By default this will fade in the object from zero with no duration.
         /// </summary>
         /// <remarks>
         /// This is called once before every <see cref="UpdateStateTransforms"/>. This is to ensure a good state in the case
         /// the <see cref="JudgementResult.TimeOffset"/> was negative and potentially altered the pre-hit transforms.
         /// </remarks>
         protected virtual void UpdateInitialTransforms()
```
iiSaLMaN:
```suggestion
         protected virtual void UpdateInitialTransforms() => this.FadeInFromZero();
```

peppy: either or. if we wanted to enforce this we could do it via r# rules.

iiSaLMaN: there are functions that just pass an argument to a field like:
```csharp
[BackgroundDependencyLoader]
private void load(Dependency dependency)
{
    this.dependency = dependency;
}
```
and it looks better with a code block so let's not do that

PR #5878, opened 2019-08-29
Encapsulate editor hitobject additions/removals

smoogipoo:
Want to encapsulate the logic surrounding additions/removals so that there's a specific flow - `HitObjectComposer` handles adding to/removing from the beatmap and processing it, `BlueprintContainer` reads from the events and adds/removes blueprints, `DrawableEditRuleset` reads from the events and adds/removes drawable visualisations.

Eventually, the grid layers will also read from the beatmap to perform their hitobject lookups.

Next PR is going to be a refactoring of the editor, in much the similar way that `DrawableRuleset` was refactored (moving most things to generic versions).

### DIFF
--- a/osu.Game/Rulesets/Edit/HitObjectComposer.cs
+++ b/osu.Game/Rulesets/Edit/HitObjectComposer.cs
@@ -20,6 +20,7 @@ using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.UI;
 using osu.Game.Screens.Edit;
 using osu.Game.Screens.Edit.Components.RadioButtons;
+using osu.Game.Screens.Edit.Compose;
 using osu.Game.Screens.Edit.Compose.Components;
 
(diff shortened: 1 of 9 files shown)


osu.Game/Rulesets/Edit/HitObjectComposer.cs
```diff
@@ -177,17 +171,74 @@ protected override void UpdateAfterChildren()
...
+            dependencies.CacheAs<IEditorBeatmap>(editorBeatmap);
+            dependencies.CacheAs<IEditorBeatmap<TObject>>(editorBeatmap);
+
+            return base.CreateChildDependencies(dependencies);
+        }
+
+        private void addHitObject(HitObject hitObject)
+        {
+            // Process object
+            var processor = Ruleset.CreateBeatmapProcessor(playableBeatmap);
```
peppy: do we really want to construct this ever add / remove?

peppy (approved):

PR #5879, opened 2019-08-29
Add "osu!classic" as a bundled skin choice

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/73
- [x] Depends on osu-resources bump.
- [x] Depends on framework bump (for https://github.com/ppy/osu-framework/pull/2807).

I'm not 100% happy with this method of adding (bypassing the database) but I'd rather have it present for the time being to allow easier testing.


osu.Game/OsuGame.cs
```diff
@@ -182,7 +182,26 @@ private void load(FrameworkConfigManager frameworkConfig)
...
+            configSkin.ValueChanged += skinId =>
+            {
+                var skinInfo = SkinManager.Query(s => s.ID == skinId.NewValue);
+
+                if (skinInfo == null)
+                {
+                    switch (skinId.NewValue)
+                    {
+                        case -1:
+                            skinInfo = DefaultLegacySkin.Info;
```
iiSaLMaN: I don't like how skin info of legacy default is inside its class while lazer default is in `SkinInfo`, you should either move the default legacy info to `SkinInfo` or move default lazer info to its class (`DefaultSkin`) which seems better imo.

osu.Game/OsuGame.cs
```diff
@@ -182,7 +182,26 @@ private void load(FrameworkConfigManager frameworkConfig)
             // bind config int to database SkinInfo
             configSkin = LocalConfig.GetBindable<int>(OsuSetting.Skin);
             SkinManager.CurrentSkinInfo.ValueChanged += skin => configSkin.Value = skin.NewValue.ID;
-            configSkin.ValueChanged += skinId => SkinManager.CurrentSkinInfo.Value = SkinManager.Query(s => s.ID == skinId.NewValue) ?? SkinInfo.Default;
+            configSkin.ValueChanged += skinId =>
```
iiSaLMaN:
```csharp
            configSkin.BindValueChanged(skinId => ..., true);
```

osu.Game/Skinning/SkinManager.cs
```diff
@@ -56,6 +60,9 @@ public SkinManager(Storage storage, DatabaseContextFactory contextFactory, IIpcH
             };
         }
 
+        private Skin createIfNotExisting(SkinInfo skinInfo) =>
```
iiSaLMaN: where is this called?

osu.Game/Skinning/SkinManager.cs
```diff
@@ -102,9 +110,15 @@ protected override async Task Populate(SkinInfo model, ArchiveReader archive, Ca
         /// <returns>A <see cref="Skin"/> instance correlating to the provided <see cref="SkinInfo"/>.</returns>
         public Skin GetSkin(SkinInfo skinInfo)
         {
+            if (skinInfo == null)
```
iiSaLMaN: is `skinInfo` ever null? even so, there are no null checks in it's usage which would likely result in null reference exception, like here for example: https://github.com/ppy/osu/blob/40c1c6072ed587b0f9f3c8ec6d720117788fd62b/osu.Game/Skinning/SkinManager.cs#L92-L97

frenzibyte (changes requested):

osu.Game/OsuGame.cs
peppy: why?

osu.Game/OsuGame.cs
peppy: ?

iiSaLMaN: instead of
```csharp
configSkin.ValueChanged += ...;
configSkin.TriggerChange();
```

peppy: kind of out of scope of this PR, but sure.

osu.Game/OsuGame.cs
iiSaLMaN: actually nvm on this one, I was thinking differently about the default legacy skin

osu.Game/Skinning/SkinManager.cs
peppy: cross-talk from the [branch i based this off](https://github.com/peppy/osu/tree/add-user-skin) (i was toying with actually importing this skin to the database, but ended up cutting that out for the initial implementation). will remove.

smoogipoo (approved):

PR #5894, opened 2019-08-30
Move beat divisor colour retrieval to BindableBeatDivisor

smoogipoo:
Makes it more accessible to outside users.


osu.Game/Screens/Edit/BindableBeatDivisor.cs
```diff
@@ -35,5 +38,41 @@ public override int Value
...
         protected override int DefaultMaxValue => VALID_DIVISORS.Last();
         protected override int DefaultPrecision => 1;
+
+        /// <summary>
+        /// Retrieves the appropriate colour for a beat divisor.
+        /// </summary>
+        /// <param name="beatDivisor">The beat divisor.</param>
+        /// <param name="colours">The set of colours.</param>
+        /// <returns>The applicable colour from <paramref name="colours"/> for <paramref name="beatDivisor"/>.</returns>
+        public ColourInfo GetColourFor(int beatDivisor, OsuColour colours)
```
peppy: feels like a weird place to put this. can be static at least?

peppy (approved):

PR #5908, opened 2019-08-30
Add beatmap video support

EVAST9919:
Still has some issues. Pr'ing this for getting suggestions, because I'm 100% sure this is a "wrong direction"


osu.Game/Beatmaps/IWorkingBeatmap.cs
```diff
@@ -25,6 +26,11 @@ public interface IWorkingBeatmap
         /// </summary>
         Texture Background { get; }
 
+        /// <summary>
+        /// Retrieves the video file for this <see cref="WorkingBeatmap"/>.
```
iiSaLMaN:
```suggestion
        /// Retrieves the video background file for this <see cref="WorkingBeatmap"/>.
```

osu.Game/Beatmaps/WorkingBeatmap.cs
```diff
@@ -183,9 +185,16 @@ public IBeatmap Beatmap
         public bool BackgroundLoaded => background.IsResultAvailable;
         public Texture Background => background.Value;
         protected virtual bool BackgroundStillValid(Texture b) => b == null || b.Available;
+
```
iiSaLMaN: why?

osu.Game/Configuration/OsuConfigManager.cs
```diff
@@ -136,6 +137,7 @@ public enum OsuSetting
         DimLevel,
         BlurLevel,
         ShowStoryboard,
+        ShowVideo,
```
iiSaLMaN:
```suggestion
        ShowVideoBackground,
```

frenzibyte (changes requested):
This is likely not the direction we want, the proper way to support this as discussed is to move all types of backgrounds into one component that can be dimmed and blurred.

osu.Game/Migrations/20171019041408_InitialCreate.Designer.cs
```diff
@@ -123,6 +123,8 @@ protected override void BuildTargetModel(ModelBuilder modelBuilder)
 
                     b.Property<string>("BackgroundFile");
 
+                    b.Property<string>("VideoFile");
```
peppy: i don't think this needs to exist in the database

peppy:
The database migration likely aren't needed, but however you made them is also incorrect (you somehow rewrote every migration ever). There's instructions [here](https://github.com/ppy/osu/wiki/Database-Migrations) for future reference.

osu.Game/Migrations/20171019041408_InitialCreate.Designer.cs
iiSaLMaN: it should? how can we know the video file of the beatmap

peppy:
Your migration is still completely wrong. Did you not follow the instructions I gave?

EVAST9919:
> Your migration is still completely wrong. Did you not follow the instructions I gave?

My bad, wasn't paying enough attention

peppy (approved):

PR #5911, opened 2019-08-31
Fix ruleset skins incorrectly providing configuration defaults

peppy:


osu.Game/Skinning/SkinConfiguration.cs
```diff
@@ -7,28 +7,25 @@
...
-        public string HitCircleFont { get; set; } = "default";
+        public string HitCircleFont { get; set; }
 
         public int HitCircleOverlap { get; set; }
 
         public float? SliderBorderSize { get; set; }
 
         public float? SliderPathRadius { get; set; }
 
-        public bool? CursorExpand { get; set; } = true;
```
smoogipoo: Doesn't this now mean that the default for an `OsuLegacySkin` is now `false`? It doesn't seem to have any form of fallback if null...

smoogipoo (approved):

PR #5928, opened 2019-08-31
Fix inconsistent sound effects on mod buttons

### LINKED ISSUE #5916: Inconsistent sound effects when clicking mod icons
**Describe the bug:**

When left-clicking on a mod icon, you can hear both checkbox and button sounds.
When right-clicking, you can only hear the checkbox sound.

https://streamable.com/n66rj

**osu!lazer version:** 2019.830.0

**Logs:** irrelevant

bdach:
Fixes #5916.

Because `HoverClickSounds.OnClick()` does not fire upon right clicking on mod buttons, the sound effects that play on left and right click were inconsistent. Introduce `HoverMouseUpSounds` drawable that allows to play the click sound effect upon mouse up events for an arbitrary set of mouse buttons and use it on mod buttons.

FWIW, this works, but it feels a little hacky, so if this is not the way you'd like the issue fixed, feel free to close this.


iiSaLMaN:
Can't this be done by adding a field to `HoverClickSounds` that would allow right clicks? or adding a constructor parameter to `HoverClickSounds` and make it optional `new[] { MouseButton.Left }` which might be better in case we want click sounds for middle mouse button.

bdach:
I'm not honestly sure, but I don't think that's possible without any deeper changes. As I mentioned in the issue, `HoverClickSounds.OnClick(ClickEvent)` doesn't even fire when right click is used to toggle mods - there is no inherent filtering by button in there, so it doesn't appear the event is even propagated to that method.

I admit I don't know the input-handling code, but it feels that either the event is filtered/handled somewhere higher in the drawable tree, or the `OnClick()` handler _semantically_ means left click and left click only, in which case I don't think we even want it to get right-click events because _a lot_ of things could start acting funny.

iiSaLMaN:
Simply what I mean is to handle that on `OnMouseUp`,  because it's equivalent to `OnMouseClick` but for any other mouse button.

bdach:
Oh, that's what you meant - sure, that would also work if that's the preferred approach.

osu.Game/Graphics/UserInterface/HoverClickSounds.cs
```diff
@@ -16,16 +18,38 @@ namespace osu.Game.Graphics.UserInterface
...
+            : this(new[] { MouseButton.Left }, sampleSet)
+        {
+        }
+
+        /// <summary>
+        /// Creates an instance that adds sounds on hover and on click for any of the buttons specified.
+        /// </summary>
+        /// <param name="buttons">Array of button codes which should trigger the click sound.</param>
+        /// <param name="sampleSet">Set of click samples to play.</param>
+        public HoverClickSounds(MouseButton[] buttons, HoverSampleSet sampleSet = HoverSampleSet.Normal)
```
iiSaLMaN:
```suggestion
        public HoverClickSounds(MouseButton[] buttons = new[] { MouseButton.Left }, HoverSampleSet sampleSet = HoverSampleSet.Normal)
```

frenzibyte (changes requested):

osu.Game/Graphics/UserInterface/HoverClickSounds.cs
bdach: This won't actually compile, a default value parameter must be a compile-time constant. Best I could do here is default to `null` and then coalesce:
```csharp
public HoverClickSounds(MouseButton[] buttons = null, HoverSampleSet sampleSet = HoverSampleSet.Normal)
    : base(sampleSet)
{
    this.buttons = buttons ?? new[] { MouseButton.Left };
}
```

iiSaLMaN: Better than another constructor.

frenzibyte (dismissed):
LGTM 👍

osu.Game/Graphics/UserInterface/HoverClickSounds.cs
```diff
@@ -16,16 +18,32 @@ namespace osu.Game.Graphics.UserInterface
...
         {
+            this.buttons = buttons ?? new[] { MouseButton.Left };
         }
 
-        protected override bool OnClick(ClickEvent e)
+        protected override bool OnMouseUp(MouseUpEvent e)
         {
-            sampleClick?.Play();
-            return base.OnClick(e);
+            var index = Array.IndexOf(buttons, e.Button);
```
peppy: please use `Contains` instead.

peppy (approved):

PR #5934, opened 2019-09-01
Ensure playing track of beatmap selected only if a track change occurred

### LINKED ISSUE #5920: Changing the ruleset resumes a paused track in song select
**Describe the bug:**

Changing the ruleset resumes a paused track in song select.
This only seems to happen in song select. Did not happen when testing in main menu.

https://streamable.com/y412o

**osu!lazer version:** 2019.830.0

**Logs:** 

[runtime.log](https://github.com/ppy/osu/files/3562438/runtime.log)

frenzibyte:
- Fixes [performing a beatmap selection update](https://github.com/ppy/osu/blob/master/osu.Game/Screens/Select/SongSelect.cs#L378) on non-track change resumes it. 

Closes #5920


smoogipoo (approved):

peppy:
I guess I missed reviewing this PR, but it is wrong in multiple ways:
- Doesn't set the preview point in all cases
- Doesn't ensure music is playing in all cases (only the local change one)

PR #5935, opened 2019-09-01
Remove unnecessary osu! endpoint from changelog builds

frenzibyte:
URL is already fed the API endpoint on `OsuGame.OpenUrlExternally()`.


peppy:
Any specific reason for this? There's a lot of other occurrences of the domain in the code. It should probably all be handled together (after making a decision on how we want to handle it). Just relying on it being added in `OpenURLExternally` isn't correct because it may be used in another scenario.

iiSaLMaN:
Found out that the tooltip mess up as well, i'll close this for now.

PR #5945, opened 2019-09-02
Add precision to screen scaling / positioning sliders

### LINKED ISSUE #5940: Strange behavior with some settings sliders and the revert button
**Describe the bug:** 

I suppose this is due to floating point imprecision, but I don't understand why only some sliders are affected, hence why I'm creating this issue on the osu! side instead of framework.

1. The orange revert button does not appear even though UI scaling was changed to `1.01x`
image
image

2. Screen position/scale sliders always display the revert button even though the value appears to be the default.
Before changing:
image
After nudging slightly using a mouse:
image
Bringing the knob back to its original position:
image

**osu!lazer version:** 2019.831.0

**Logs:** 

[runtime.log](https://github.com/ppy/osu/files/3563309/runtime.log)

frenzibyte:
- Adds precision to screen scaling / positioning sliders (Horizontal position, Vertical position, Horizontal scale, Vertical scale).
- [x] Depends on ppy/osu-framework#2816

Closes #5940


peppy:
can you explain why they need to be changed to `double`? are you suggesting we can never use floats for some reason?

iiSaLMaN:
> can you explain why they need to be changed to `double`? are you suggesting we can never use floats for some reason?

I wasn't sure about this fix but it turns out `float` loses precision and makes `BindableFloat.IsDefault` inaccurate, for example, setting the slider to `1.01f` while the precision is `0.01f` doesn't show the revert button because `BindableFloat.IsDefault` returns true due to `(1.01f - 1f = 0.00999999046 < precision = 0.00999999978) == true`, hope it clears it up.

peppy:
So shouldn't bindable be fixed then? What you've done here is work around broken framework behaviour. That's generally not something we do.

peppy:
Can you confirm whether this is still required with the framework side change?

iiSaLMaN:
> Can you confirm whether this is still required with the framework side change?

this is still required for the screen-scaling precision, will rename the PR.

peppy:
So you're saying without this the default button still doesn't work?

peppy:
It doesn't work, but this turns out to be a regression / test shortcoming of the framework side change made recently: https://github.com/peppy/osu-framework/tree/broken-default

iiSaLMaN:
> It doesn't work, but this turns out to be a regression / test shortcoming of the framework side change made recently: https://github.com/peppy/osu-framework/tree/broken-default

see if ppy/osu-framework#2826 solves the problem

peppy:
Confirming fixed via framework change.

PR #5949, opened 2019-09-02
Fix taiko swell tick judgements having non-zero time offsets

### LINKED ISSUE #5912: Hit arrow estimation exceeds Hit error height
As explained from the title, the Hit Arrow Estimation exceeds the Hit Error height.

image

In order to reproduce this, hit atleast 5-10 objects late/early and you will see the arrow go above/below the Hit error height

smoogipoo:
Resolves part of #5912


peppy (approved):

peppy:
Note that this likely won't completely resolve the issue (as discussed).

PR #5955, opened 2019-09-02
Extend SearchBeatmapSetsRequest by adding page parameter

EVAST9919:


osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -11,23 +12,38 @@ namespace osu.Game.Online.API.Requests
...
         {
             this.query = System.Uri.EscapeDataString(query);
             this.ruleset = ruleset;
+            this.page = page;
             this.searchCategory = searchCategory;
             this.sortCriteria = sortCriteria;
             this.direction = direction;
         }
 
         // ReSharper disable once ImpureMethodCallOnReadonlyValueField
```
jorolf: this can probably be removed now

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -213,11 +216,12 @@ private void recreatePanels(PanelDisplayStyle displayStyle)
                 })
             };
 
+            loadCancellation = new CancellationTokenSource();
+
             LoadComponentAsync(newPanels, p =>
             {
-                if (panels != null) ScrollFlow.Remove(panels);
```
peppy: why was this removed?

peppy:
Is this PR supposed to add pagination? The branch name suggests it is but it sin't.

EVAST9919:
It was, but I thought it would be better to extend the request in it's own pr first

peppy:
Need some way to test, though.

peppy:
This seems weird now. You're adding a bindable which is going to be removed (or never used) in future changes.

If you can't test a change without adding cruft, you need to include more of what you changed. Or less. I would argue what you're doing here is pointless to PR separately from the actual pagination implementation.

EVAST9919:
How we want to handle the pagination? Like infinite scroll or by just adding a ShowMore button?

EVAST9919:
Will be reopened when after the actual scrolling implementation

PR #5959, opened 2019-09-02
Use float types for user visual levels

frenzibyte:
- Uses `float` type for user visual levels (`BlurLevel` and `DimLevel`) as it should've been.
- [x] Depends on ppy/osu-framework#2816

### DIFF
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
@@ -295,8 +295,8 @@ namespace osu.Game.Tests.Visual.Background
             }
 
             public readonly Bindable<bool> DimEnabled = new Bindable<bool>();
-            public readonly Bindable<double> DimLevel = new Bindable<double>();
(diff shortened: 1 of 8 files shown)


peppy:
Have you tested that this transfers values correctly in cases users already had a value set?

iiSaLMaN:
Bringing this topic here: `BlurLevel` sometimes have `0.39999..` instead of `0.4`, not sure whether to round that value or just close this PR?

bdach:
I'm not near a C# interactive shell right now, but I'm almost sure neither `double` or `float` can ever accurately represent a value of 0.4 due to the fact base 10 fractions don't work well with IEEE 754 floating point representations. You might get slightly *over* 0.4, instead of slightly under, but this should be handled at display time regardless of type, and for graphics calculations, if there are no extreme usages, single precision should be okay. Key thing to check at this point is that this migration doesn't mangle existing user preferences, as we're going from a more precise type to a less precise one - so you want to make sure no *horrific* truncations will take place (small ones are natural and expected in cases like this)

iiSaLMaN:
Closing this PR for now, will work on to later.

PR #5969, opened 2019-09-03
Refactor skin configuration lookups to be more flexible

peppy:
Rulesets can now define configuration enums as they wish. Note that namspacing has not yet been added, but this is unnecessary for current legacy implementations (and will be added in the near future).

Removes ugly delegate-based configuration lookups which were prone to error in null/fallback handling.


peppy:
- [ ] Probably depends on framework to mark test as headless only (https://github.com/ppy/osu-framework/pull/2818).

smoogipoo (dismissed):

smoogipoo:
@peppy review my latest changes, otherwise seems good 👍

PR #5970, opened 2019-09-03
Fix possible broken links in DrawableKudosuHistoryItem

EVAST9919:
Can happen because current MessageFormatter regex can't handle brackets in title
64130812-57690180-cdcd-11e9-8624-6d7f9eafb3cc


peppy:
Please fix the formatter instead (and make sure to add tests)

```markdown
[test [] ()](https://github.com)
```

->

[test [] ()](https://github.com)

EVAST9919:
Should be fixed formatter instead

PR #5979, opened 2019-09-03
Fix bracket and parenthesis handling in existing link formats

### LINKED ISSUE #5772: Message formatter can't create a correct link with opened square bracket in the link title
Example:
Screenshot_3
Screenshot_4

Usage: Kudosu History can return link to the post, in which there's only one opened square bracket in the name.
Screenshot_2

bdach:
Resolves #5772 (at least partially, see *Potential issues* for the asterisk on that).

# Description

For users to be able to add square brackets inside of links using the new format, the regular expression used for parsing those links contained a balancing group, which can be used for matching pairs of tokens (in this case, opening and closing brackets, in that order). However, this means that users could not post links with unmatched brackets inside of them (ie. ones that contain single brackets, or a closing bracket and then an opening one). Allow for escaping opening and closing brackets using the backslash character.

The change substitutes this old fragment of the regex in the display text group:

    [^\[\]]*        // any character other than closing/opening bracket, zero or more times

for this one: 

    (((?<=\\)[\[\]])|[^\[\]])*

The second pattern in the alternative remains the same; the first one performs the escaping, as follows:

    (
        (?<=\\)     // positive lookbehind expression:
                    // this match will succeed, if the next expression
                    // is preceded by a single backslash
        [\[\]]      // either an opening or closing brace
    )

Since the entire display group is matched and then passed down to actually format the link, unfortunately the lookbehind expression does not actually strip the backslashes, so they are manually stripped in `handleMatches`.

As demonstrated in the unit tests attached, this also allows balanced brackets to be mixed with escaped ones.

Analogous changes were made to Markdown and old-style link formats.

For the sake of readability, consistency and to make further changes easier, named groups `(?<text>)` and `(?<url>)` were also introduced to all link parsing regexes which have parts containing the desired link text and (optionally) URL.

The introduction of the named groups additionally simplifies `handleMatches()` and makes all calls to it consistent.

# Potential issues

* I'm not entirely sure about that escape replace part; entirely willing to swap that `Aggregate()` out for some `StringBuilding`.
* Open question: should I add the same handling to wiki links?
* Follow-up: to fully resolve #5772 in the original case of stray brackets in the link display text, it might also be required to escape all brackets in link texts in the kudosu list view, but that seems kind of out-of-scope for this PR. But it might be less painful for everyone involved if I just include it here.
* Documentation - I pulled this PR's description almost entirely from commit messages, but maybe some inline comments would also do?
* Readability - I could extract that magic `(((?<=\\)[\[\]])|[^\[\]])*` part to a variable and string together the regexes with that.
* Testing - I think I covered the obvious test cases, but suggestions for more are very welcome.

### DIFF
--- a/osu.Game.Tests/Chat/MessageFormatterTests.cs
+++ b/osu.Game.Tests/Chat/MessageFormatterTests.cs
@@ -120,6 +120,53 @@ namespace osu.Game.Tests.Chat
         }
 
         [Test]
+        public void TestOldFormatLinkWithBalancedBrackets()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "This is a (tricky (one))[https://osu.ppy.sh]!" });
+
+            Assert.AreEqual("This is a tricky (one)!", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
+            Assert.AreEqual("https://osu.ppy.sh", result.Links[0].Url);
+            Assert.AreEqual(10, result.Links[0].Index);
+            Assert.AreEqual(12, result.Links[0].Length);
+        }
+
+        [Test]
+        public void TestOldFormatLinkWithEscapedBrackets()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "This is (another loose bracket \\))[https://osu.ppy.sh]." });
+
+            Assert.AreEqual("This is another loose bracket ).", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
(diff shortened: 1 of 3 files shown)


smoogipoo:
Let's get this in as is for now, there's sufficient testing for me to be comfortable with it. 👍 

Probably not necessary adding this for wiki links until a case comes up, I'm not aware of such a case currently.

smoogipoo (approved):

peppy:
Thanks for the high quality submission!

PR #5998, opened 2019-09-05
Isolate "No beatmaps available" from DummyWorkingBeatmap

frenzibyte:
- Allows for checking if `no beatmaps available!` is the current beatmap and also for creating other types-of `DummyWorkingBeatmap` like `NoBeatmapsFound` for example.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -97,9 +97,9 @@ namespace osu.Game.Tests.Visual.SongSelect
         public void TestDummy()
         {
             createSongSelect();
-            AddAssert("dummy selected", () => songSelect.CurrentBeatmap == defaultBeatmap);
+            AddAssert("no beatmaps selected", () => songSelect.CurrentBeatmap == defaultBeatmap);
 
-            AddUntilStep("dummy shown on wedge", () => songSelect.CurrentBeatmapDetailsBeatmap == defaultBeatmap);
+            AddUntilStep("no beatmaps shown on wedge", () => songSelect.CurrentBeatmapDetailsBeatmap == defaultBeatmap);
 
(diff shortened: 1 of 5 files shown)


smoogipoo:
I'm not sure when there will ever be another type of beatmap other than DummyWorkingBeatmap.

iiSaLMaN:
> I'm not sure when there will ever be another type of beatmap other than DummyWorkingBeatmap.

First and mostly, it's for allowing to check that a `no beatmaps available!` is the current beatmap and not a `DummyWorkingBeatmap` (tests uses that to create a dummy beatmap), and also to allow dummy beatmaps with other metadata like `no beatmaps found!`

smoogipoo:
This is not the purpose of `DummyWorkingBeatmap`. It shouldn't used as a signal for what went wrong, but only as a signal that no beatmap is selected.

I don't see "no beatmaps found!" ever being a thing as part of `DummyWorkingBeatmap`.

iiSaLMaN:
> This is not the purpose of `DummyWorkingBeatmap`. It shouldn't used as a signal for what went wrong, but only as a signal that no beatmap is selected.

well, tests uses `DummyWorkingBeatmap` as a dummy beatmap to test song select components: `TestSceneBeatmapDetailsArea` for example.

smoogipoo:
Let me reiterate. `DummyWorkingBeatmap` is an object that provides sane defaults, not a marker for various game states.

If the wedge wants to display something different depending on the filter state/other scenarios, it should internally type-match on `Beatmap.Value is DummyWorkingBeatmap` and decide what to display based on other game states (maybe it could DI in the beatmapmanager, maybe it could receive the filter state from a bindable from SongSelect, etc). Same goes for the leaderboard.

As for testscenes using `DummyWorkingBeatmap` as this marker (e.g. `TestSceneBeatmapDetailArea`), they are doing so incorrectly.

smoogipoo:
As mentioned, let me know if you need more explanation.

PR #5999, opened 2019-09-05
Add "please select a beatmap!" leaderboard placeholder state

### LINKED ISSUE #5300: Leaderboard sections should be hidden if no beatmap are available.
When you don't have beatmaps, `DummyWorkingBeatmap` is used as the working beatmap to tell on the wedge that there are no beatmaps, the problem is you get wrong messages when viewing the leaderboard `Leaderboards are not available for this beatmap`

the leaderboard should probably be hidden if there are no beatmaps available.

frenzibyte:
- Adds a `please select a beatmap!` placeholder state on leaderboards once there is no beatmap selected.
- [x] Depends on #6050

Closes #5300


peppy:
Care to update this to just check against `DummyWorkingBeatmap` instead?

iiSaLMaN:
> Care to update this to just check against `DummyWorkingBeatmap` instead?

Yes, I'll do it in a bit.

smoogipoo (approved):

PR #6007, opened 2019-09-05
Add "Larger Circle" mod

ojh050118:


osu.Game/Rulesets/Mods/ModLargerCircle.cs
```diff
@@ -0,0 +1,26 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Beatmaps;
+
+namespace osu.Game.Rulesets.Mods
+{
+    public abstract class ModLargerCircle : Mod, IApplicableToDifficulty
```
iiSaLMaN: Isn't this supposed to be an osu!-ruleset mod only? this should probably be moved to `OsuModLargerCircle`

Anthony1x (dismissed):

osu.Game/Rulesets/Mods/ModLargerCircle.cs
ojh050118: I didn't understand what you mean. Can you tell me in detail?

iiSaLMaN: this is being created in a game-level when it supposed to be created in a ruleset-level, just move all the code from here into `OsuModLargerCircle`

peppy:
Please read some tutorials on how to use git before contributing to this repository.

Issue #6005, opened 2019-09-05
Hitobjects have no color and sound with default skin


Gabixel:
**Describe the bug:** All hitobjects are white with the default Lazer skin, and they have no hitsound.

**Screenshots or videos showing encountered issue:** https://streamable.com/x4455

**osu!lazer version:** 2019.905.0

bdach:
Yeah, I also have this.

Seems to have regressed in two separate commits. The lack of colours bisected down to bebc330, but hitsounds were working at f58ca82. I can investigate further and/or bisect the hitsounds too if necessary.

bdach:
Hitsounds disappearing bisected down to 6197c7f.

Gabixel:
Fixed by https://github.com/ppy/osu/commit/fbdef9749828326767e6c999e5b3300a38c2976b I guess

PR #6008, opened 2019-09-05
Add back missing sample fallback to default skin

peppy:
Open for discussion. Obviously not the best way to approach this, but can be used as a hotfix.

Fixes sample part of https://github.com/ppy/osu/issues/6005.


osu.Game/Rulesets/UI/FallbackSampleStore.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void AddAdjustment(AdjustableProperty type, BindableDouble adjustBindable)
+        {
+            primary.AddAdjustment(type, adjustBindable);
+            secondary.AddAdjustment(type, adjustBindable);
+        }
+
+        public void RemoveAdjustment(AdjustableProperty type, BindableDouble adjustBindable)
+        {
+            primary.RemoveAdjustment(type, adjustBindable);
+            primary.RemoveAdjustment(type, adjustBindable);
```
iiSaLMaN: `primary` x2

osu.Game/Rulesets/UI/FallbackSampleStore.cs
```diff
@@ -0,0 +1,64 @@
...
+        public FallbackSampleStore(ISampleStore primary, ISampleStore secondary)
+        {
+            this.primary = primary;
+            this.secondary = secondary;
+        }
+
+        public void Dispose()
+        {
+            primary.Dispose();
+            secondary.Dispose();
```
smoogipoo: This is 100% prone to error / dangerous. Probably should be the consumer's responsibility to dispose.

smoogipoo (changes requested):

osu.Game/Rulesets/UI/FallbackSampleStore.cs
peppy: oh yeah, i meant to remove this (i already moved handling to the consumption point after noticing this).

smoogipoo (approved):

PR #6011, opened 2019-09-05
Add link to templates in README.md

ticotaco72:
Probably could add something more... Idk


peppy:
Ruleeset development should be a full wiki documentation section with a very short mention in the readme. Also the template repository does not need to be linked. Its purpose is to provide the template but should not be visited by ruleset developers as it serves no further purpose.

peppy:
Closing for now, needs more thought.

PR #6019, opened 2019-09-06
Refactor HitWindows for legibility

peppy:
Removes the concept of "HalfHitWindows" and converts all existing values to +/-ms (seems to be the accepted industry standard).


osu.Game/Rulesets/Scoring/HitWindows.cs
```diff
@@ -4,51 +4,61 @@
...
         /// </summary>
-        public double Perfect { get; protected set; }
+        private double perfect;
 
         /// <summary>
         /// Hit window for a <see cref="HitResult.Great"/> result.
         /// </summary>
-        public double Great { get; protected set; }
+        /// <remarks>
+        /// Note that this value includes both the early and late region.
```
smoogipoo: This is now incorrect

smoogipoo (changes requested):

smoogipoo (approved):

PR #6021, opened 2019-09-06
Add "Larger Circle" mod

ojh050118:
Mod for people who are difficult to osu! (I'm sorry about what happened yesterday)


ojh050118:
> > Mod for people who are difficult to osu!
> 
> We already have "Easy" mod. Why do we need another one mod?

Increase circle size without affecting play.

iiSaLMaN:
> Increase circle size without affecting game play.

Should be incompatible with `Easy`

abraker95:
We need customizable CS and AR, not a million mods

peppy:
Please wait for mod level settings, which will allow for this degree of customisation. If you are looking to improve the mobile situation, the playfield *itself* should be scaled larger to take up more of the screen as a first step.

PR #6025, opened 2019-09-06
Add bottom bar hit error meter

### LINKED ISSUE #5932: Centered hit errors (like osu!stable)
I can't see the hit errors when it's set to the left or right side. It's further away from the sides than the bottom (Widescreen problem,it's not in the field of view) It would be nice if osu!lazer has the same hit error option as stable, as well as the scaling option.

One Option missed same on the stable:
ss (2019-08-30 at 08 16 00)

Like this (is a screenshot montage, edited by paint.net)
ss (2019-08-30 at 08 29 02)

Game4all:
Closes #5932 

Mostly adapted existing code of the vertical hit bar error meter for an horizontal use

It is for now bottom centered between the song graph and the progress counter
image


osu.Game/Screens/Play/HUD/HitErrorMeters/BottomBarHitErrorMeter.cs
```diff
@@ -0,0 +1,286 @@
...
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Objects;
+using osu.Game.Rulesets.Scoring;
+using osuTK;
+using osuTK.Graphics;
+using System.Linq;
+
+namespace osu.Game.Screens.Play.HUD.HitErrorMeters
+{
+    public class BottomBarHitErrorMeter : HitErrorMeter
```
peppy: Why is this a complete copy-paste? It should 100% be able to share at least 99% of logic with the side bsrs.

peppy (changes requested):

osu.Game/Screens/Play/HUD/HitErrorMeters/BottomBarHitErrorMeter.cs
Game4all: Decided to go with the copy-paste way for the sake of code readability (else code would be filled with inline conditionals) . Will refactor to at least share used components.

Game4all:
Did a bit of refactoring:

- Both bar hit error meters inherit of `BarHitErrorMeter` 
which wraps the judgement, and bar creation from hitwindows.

- Bar hit error meters code is now mostly layout code.

- Bar hit error meters have to implement `CreateJudgement(JudgementResult)` and `CreateColourBar(Color4, HitWindow, bool)` 

- JudgementLine is now public and takes a bool indicating whether it should be vertically displayed or not.

osu.Game/Screens/Play/HUD/HitErrorMeters/Bar/SidebarHitErrorMeter.cs
```diff
@@ -0,0 +1,196 @@
...
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Game.Screens.Play.HUD.HitErrorMeters.Bar
+{
+    public class SidebarHitErrorMeter : BarHitErrorMeter
```
iiSaLMaN: matching the capitalization with the inherited name
```suggestion
    public class SideBarHitErrorMeter : BarHitErrorMeter
```

frenzibyte (commented):
I would honestly prefer checks to pick the `Anchor` and `Origin` and whether to set `Width` or `Height`, but I guess this is all for readability

smoogipoo:
You've still duplicated 70% of the code between the vertical and horizontal ones. All the constants are the same, the way in which the gradient is generated is the same...

The only thing that changes is the layout, and that's a matter of correct anchors/origins and positionings.

Game4all:
Will reopen a PR using the method as proposed above (using checks to determine anchors & origin).

PR #6028, opened 2019-09-07
Reset DrawableHitObject lifetimes on state change

peppy:
I believe this was expected from the start. First step to fixing osu!catch rewind logic.


smoogipoo:
This looks so wrong. The lifetime is set once by the scrolling hitobject container on user scroll speed change/map load, so this looks like it could break a lot. This is not what I'd expect at all.

I'm not sure what you're trying to fix here, the catcher area creates new hitobjects when they're added to the plate, and as far as I'm aware the non-caught fruit already handle lifetimes properly? It's the caught ones that remain in a broken state.

peppy:
This PR has nothing to do with the catching part, just the resetting of lifetime on state change. Currently rulesets generally call `Expire()` from within their `UpdateStateTransforms`. Currently that operation is not reset, which means they can get into an incorrect state after rewinding.

You can see how this affects catch by running through gameplay (I've been using the juice streams autoplay test) after first disabling the catcher `Add` function to reduce visual noise.

An alternative may be to implicitly call the `Expire` part in all cases?

peppy:
To summarise what this is fixing:

- Currently rulesets generally call `Expire` in their `UpdateStateTransforms` implementation which sets a `LifetimeEnd`.
- In the case they do not call `Expire` or reset the end lifetime, it can potentially be in a state which does not allow correct input or judgement handling after a rewind.
- This is the case for osu!catch, which relies on judgement checking in `UpdateAfterChildren`. This may not be called in time (before a previously set `LifetimeEnd`) which causes judgements that are delayed until the `OnKilled` safety measure (https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L323).
- `LifetimeStart` being reset is (as you mentioned) incorrectly in this PR and may need further thought, if we want to support/guard against `Expire(true)` calls.

Reproduction:

- Disable fruit catching (reduce visual noise) by commenting out https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Catch/UI/CatcherArea.cs#L78
- Play through a few juice streams in `TestSceneAutoJuiceStream`.
- Rewind to start.
- Observe delayed judgements on subsequent plays.

smoogipoo:
Perhaps what we may want to do is reset _just_ the lifetime end. We'd need sufficient documentation to let people know they must `.Expire()` in Hit/Miss cases though.

peppy:
Are you against having `Expire` called automatically?

smoogipoo:
I'm not.

smoogipoo (approved):

smoogipoo:
@peppy quick check plz

peppy:
👍

PR #6031, opened 2019-09-07
Fix results screen pushed at the beginning of a gameplay

frenzibyte:
Due to autoplay simulation failing if the first judgement doesn't increase health.

### DIFF
--- a/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
@@ -55,7 +55,7 @@ namespace osu.Game.Rulesets.Scoring
         /// <summary>
         /// The current health.
         /// </summary>
-        public readonly BindableDouble Health = new BindableDouble { MinValue = 0, MaxValue = 1 };
+        public readonly BindableDouble Health = new BindableDouble(1) { MinValue = 0, MaxValue = 1 };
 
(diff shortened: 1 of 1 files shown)


peppy:
Is there a reason you chose this solution as opposed to calling `Reset()` before autoplay simulation? I'd say calling `Reset()` is probably something we want to do as a safety measure (in addition to the default, maybe).

iiSaLMaN:
> Is there a reason you chose this solution as opposed to calling `Reset()` before autoplay simulation? I'd say calling `Reset()` is probably something we want to do as a safety measure (in addition to the default, maybe).

I chose it because I don't really see any issue with not calling `Reset()`, keep in mind it's a simulation to find the maximum hit judgements, not something seen on screen, although I'll add it maybe to make the simulation accurate?

peppy:
@smoogipoo requesting your review with regards to `Reset()` placement in this change.

peppy (dismissed):

smoogipoo:
I would also prefer `Reset()` to be called before the simulation.

peppy:
It has been added to this PR, just not sure if the placement is correct (should it be in the `ctor` instead?)

smoogipoo:
That's what I meant, yes.

smoogipoo (approved):

PR #6032, opened 2019-09-07
Fix osu!catch rewind handling

peppy:
Keep in mind this solution is a best-effort. The whole management of fruit transfer needs to be re-thought to correctly play in reverse in a frame perfect manner (have the fruit re-appear on the plate and move backwards correctly).

Currently we only rewind at a very fast speed so this shortcoming is not visually noticeable.

- [x] Depends on https://github.com/ppy/osu/pull/6028.


smoogipoo:
Quick summary of the issues I've found:
1. Additive hyperdash layers aren't being added after rewinding.
2. The fruit removal code isn't quite working (likely `Clock.ElapsedFrameTime` is never going < 0 in the broken cases).

smoogipoo (approved):

PR #6033, opened 2019-09-07
Add android configuration defaults to props

miterosan:
Previously they may have been defined too late in the build process.

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
     <OutputPath>bin\$(Configuration)</OutputPath>
     <WarningLevel>4</WarningLevel>
     <SchemaVersion>2.0</SchemaVersion>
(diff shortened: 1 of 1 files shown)


osu.Android.props
```diff
@@ -1,5 +1,6 @@
 <Project>
   <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
```
peppy: This line does not seem necessary.

miterosan: the build will fail if you do not specify a configuration. This happens because osu.Android.props is imported before setting Debug as the default configuration.

You will get an empty APK when you do ``msbuild /t:SignAndroidPackage`` without this change.

peppy: so can all this be fixed by doing `/property:Configuration=Release`?

miterosan: No. The problem is because properties are checked before the set. (osu.android.props is imported before Configuration and Platform are set to their default values)

peppy: But specifying a default at command line should work, no?

peppy (approved):

PR #6035, opened 2019-09-07
Implement PageSelector component

EVAST9919:
Screenshot_1


peppy:
The way you've structured this means it's impossible to animate the buttons, because they are reconstructed on click.

EVAST9919:
Yep, I just copied logic from web-side. Looks like, that's not a good thing to do

peppy:
Generally avoid recreating drawables/buttons where they persist on screen.

EVAST9919:
Elements are refactored in latest commits for animations in mind (still not finished, but a good step forward)

EVAST9919:
I'm giving up on this one. Not sure I know the way (and direction) to implement this _properly_

smoogipoo:
Let me look at it again before you do. Last time I tried to rewrite just to add more animations and it became complex, but we’ve since decided to not require animations so long as they’re easy to add down the track.

peppy:
I've brought this up-to-date so we can get it merged for future use. It's still in a bit of a non-animated state but seems better than closing the PR.

I've changed `CurrentPage` to be zero-based, as that feels better to me. Open to opinions on the matter.

bdach:
* `TestResetCurrentPage` is failing. Likely something to do with the indexing changes.
* There is a white flash of the page items when the page is changed. Is that intentional? It was sort of throwing me off, I wouldn't expect that to happen.

  https://user-images.githubusercontent.com/20418176/148114396-3e87aef9-e106-4c58-9aec-f52d239deccb.mp4

bdach (changes requested):
as above

peppy:
Something must be super wrong with the tests, as they passed in visual test fine.

As for the animation: I left it as-is because the idea is that it really shouldn't be repopulating each time the selected state changes (see the history of this PR). I'd rather leave the animation broken as a hint that this needs to be resolved, rather than hiding it from now. We can reconsider when this is actually used somewhere. Main goal is to merge this PR, though.

bdach (approved):
seems inoffensive enough

PR #6037, opened 2019-09-07
Open login overlay if user isn't logged in after intro sequence

### LINKED ISSUE #6030: Login prompt should automatically appear if user isn’t logged in.
Upon entering the game, login prompt should appear once clicking the osu! logo if user isn’t signed in. Or, it appears when intro sequence ends, matching stable behaviour.

LeNitrous:
Matches osu-stable behavior. Resolves https://github.com/ppy/osu/issues/6030


bdach:
Those two merge commits at the start are really confusing, and in the github UI they point to the entirely wrong PRs from *this* repository. I'm not even sure how the branch history would look like if this was merged.

It seems you're updating your fork by PRing upstream back to it, this is not the right way to do it - see [*Syncing a fork*](https://help.github.com/en/articles/syncing-a-fork) for instructions.

peppy:
Please rebase and force push (only this once, we usually don't allow this) without the weird merge commits.

LeNitrous:
I'll redo this PR to avoid rebasing.

PR #6038, opened 2019-09-08
Open login overlay if user isn't signed in after intro sequence

### LINKED ISSUE #6030: Login prompt should automatically appear if user isn’t logged in.

LeNitrous:
Matches osu-stable behavior. Resolves #6030. Supersedes #6037.


osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -128,6 +135,9 @@ public override void OnEntering(IScreen last)
                     track.Seek(metadata.PreviewTime != -1 ? metadata.PreviewTime : 0.4f * track.Length);
                     track.Start();
                 }
+
+                if (api?.State == APIState.Offline)
```
peppy: which is this inside a track-related conditional?

LeNitrous: didn't realize it was checking the track as well, I'll make a separate clause where it only checks if the last screen was an `IntroScreen`

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -130,6 +137,9 @@ public override void OnEntering(IScreen last)
                 }
             }
 
+            if (last is IntroScreen && api?.State == APIState.Offline)
```
iiSaLMaN: never null
```suggestion
            if (last is IntroScreen && api.State == APIState.Offline)
```

peppy:
You probably want `api.IsLoggedIn` instead (should be safer). Also, I don't think you need the `last is IntroScreen` check – have you investigated why that is there?

LeNitrous:
Oh right. `MainMenu` just suspends into the background when transitioning to Song Select or Editor.

peppy:
This looks visually jarring. THe login overlay should not be shown without this toolbar present.

peppy:
I've fixed some remaining issues up. Seems okay to me now.

peppy (dismissed):

peppy (dismissed):

PR #6040, opened 2019-09-08
Decrease Spinner Disc Rotation on Rewind

frenzibyte:
- Decrease `RotationAbsolute` on rewinding.


osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
```diff
@@ -111,7 +113,8 @@ protected override void Update()
                     lastAngle -= 360;
 
                 currentRotation += thisAngle - lastAngle;
-                RotationAbsolute += Math.Abs(thisAngle - lastAngle);
+                BidirectionalRotation += Math.Abs(thisAngle - lastAngle) * Math.Sign(Time.Current - lastTime);
```
peppy: `Clock.ElapsedTime`?

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
```diff
@@ -82,10 +82,12 @@ protected override bool OnMouseMove(MouseMoveEvent e)
 
         private float lastAngle;
         private float currentRotation;
-        public float RotationAbsolute;
+        public float BidirectionalRotation;
```
peppy: i don't understand this variable name change. the old one makes more sense to me still?

peppy (dismissed):

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
hbnrmx: You wouldn't expect to find a variable named `RotationAbsolute` to be a negative value, which it may become when rewinding.

peppy: It should *never* be negative.

iiSaLMaN: Ah correct, I was thinking differently for this one.

smoogipoo:
This needs tests to ensure that:

1. Rewinding from a complete spinner to the start results in a `RotationAbsolute` of 0 (or close to).
2. Rewinding into the middle of a spinner and playing back results in the same `RotationAbsolute` as completing the spinner (or close to).

iiSaLMaN:
> This needs tests to ensure that:
> 
> 1. Rewinding from a complete spinner to the start results in a `RotationAbsolute` of 0 (or close to).
> 2. Rewinding into the middle of a spinner and playing back results in the same `RotationAbsolute` as completing the spinner (or close to).

Keep in mind seeking forward wouldn't increase the `RotationAbsolute` since the spinner never spins without a cursor spinning it. so `RotationAbsolute` must be manually set, is that fine?

peppy:
Why can't you use an autoplay test? Then you would have a cursor spinning, no?

iiSaLMaN:
@smoogipoo Tests added.

smoogipoo (dismissed):

peppy:
Going to need further review on the tests, based on intermittent failure.

iiSaLMaN:
@peppy What do you mean by intermittent failure? also you mean further review as in how accurate is the test?

peppy:
See the CI failure above. It's two commits ago.

https://ci.appveyor.com/project/peppy/osu/builds/27617340/tests

smoogipoo (approved):

PR #6041, opened 2019-09-08
Remove unnecessary AndroidNativeLibrary specification

miterosan:
Is depended on https://github.com/ppy/osu-framework/pull/2824.

This pr removes a workaround

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -49,7 +49,6 @@
     <None Include="$(MSBuildThisFileDirectory)\osu.licenseheader">
       <Link>osu.licenseheader</Link>
     </None>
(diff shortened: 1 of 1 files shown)


peppy:
> This pr remove a workaround

What do you mean by this?

miterosan:
The native libraries are copied into output directory and then collected as Android native libraries. (That is bad, because you need to build the project twice: first time to have the native libraries copied into the output directory. The libraries are only collected at the second build going forward. )

That is not necessary when you use EmbeddedAndroidlibrary in osu.Framework.Android. 
EmbeddedAndroidlibrary automatically adds the native libraries to the project, that references osu.framework.Android.

PR #6043, opened 2019-09-08
Add fallback option for badly-headered beatmap files

bdach:
Going to preempt this by saying that I don't like how this PR turned out at all, so posting as a draft and if you feel like this is too wrong I'm completely fine with this being closed.

Due to quite wide refactoring I recommend reading the diffs commit-by-commit instead of in bulk. If it's deemed necessary I can also split this out across multiple PRs.

# Description

This is an attempt to address #5522 while preserving the current public interface. Unfortunately, due to stream foibles the only way I could come up with something is by introducing state to the `Decoder`s. I dislike stateful objects, but I *really* tried to think of something else for like two days and I got nothing.

The reason for the refactor is, that when calling `Decoder<T>.GetDecoder()`, the method examines the first line of the stream being decoded for a magic string and uses it to return a correct decoder type. Because it operates on the stream directly, that first line is consumed - the actual decoder will read from the second line onwards.

It is however possible that this magic string will not be present in the first line of the file, and will instead actually be a part of the content to be decoded (most likely, a section header). In such a case
it is desired that this line is not discarded, and instead parsed the same way normal lines are. If it isn't, then the entire section will be discarded.

Rewinding the `StreamReader` is not correct due to encoding issues. If the file being parsed contains a byte sequence indicating encoding, like the UTF-8 byte-order mark, [it will be read with the first line after the rewind and break parsing](https://stackoverflow.com/a/31444226/6817321). There are workarounds floating around the Web that don't seek to offset 0, but instead use the methods of the `StreamReader` to get the preamble length. (I tested that, those methods don't actually *check* that the content has that preamble, so that hack is wrong when the BOM mark is *missing*.)

To bypass this roadblock, extract current configuration section to a protected field, and extract logic of parsing a single line to a protected method. This allows `GetDecoder()` to fall back to the normal line parsing path, when a file without a header is encountered.

Additionally, to be able to do something with that line, refactor base `Decoder` to contain an `Output` property with the decoded object instance, which is lazily constructed on-demand using the implementation of `CreateTemplateObject()`.

This allows to remove fields in `LegacyBeatmapDecoder` and `LegacyStoryboardDecoder` which functionally had the same usage.

Finally the fallback is implemented using a dictionary mapping types to fallback constructor delegates. I would have liked to not use a second collection, but `Dictionary` does not guarantee preserving order, and `OrderedDictionary` has a horrible untyped interface.

The main problem with this (besides introducing state), is that it makes instances of `Decoder` non-thread-safe. That's sort of easily fixable though, I guess, if unperformant.

# (Potential) To-do list

Of course this applies only if it even makes sense to continue in this manner.
* [ ] Right now even garbage files will get parsed; nothing valuable will be written to `Output`, but a still non-`null` object will be returned. It'd probably be better to fail out if nothing sensible ever gets read (will fix failing test: `osu.Game.Tests.Beatmaps.IO.ImportBeatmapTest.TestRollbackOnFailure()`)
* [ ] Make everything thread-safe (guard the `Lazy<TObject>` and `Decode()`)
* [ ] At current the API is sort of consumer-unsafe. That could be addressed by consolidating `GetDecoder()` and `Decode()` into one operation and hiding everything ugly with stricter access modifiers. Although there are tests now that rely on those two being separate (they call `GetDecoder()`, do some casting to set a property on the decoder, and only then call `Decode()`).

### DIFF
--- a/osu.Game.Tests/Beatmaps/Formats/LegacyBeatmapDecoderTest.cs
+++ b/osu.Game.Tests/Beatmaps/Formats/LegacyBeatmapDecoderTest.cs
@@ -494,5 +494,39 @@ namespace osu.Game.Tests.Beatmaps.Formats
                 Assert.DoesNotThrow(() => decoder.Decode(badStream));
             }
         }
+
+        [Test]
+        public void TestDecodeBeatmapWithCorruptedHeader()
+        {
+            Decoder<Beatmap> decoder = null;
+            Beatmap beatmap = null;
+
+            using (var resStream = TestResources.OpenResource("corrupted-header.osu"))
+            using (var stream = new StreamReader(resStream))
+            {
+                Assert.DoesNotThrow(() => decoder = Decoder.GetDecoder<Beatmap>(stream));
+                Assert.IsInstanceOf<LegacyBeatmapDecoder>(decoder);
+                Assert.DoesNotThrow(() => beatmap = decoder.Decode(stream));
+                Assert.AreEqual("Beatmap with corrupted header", beatmap.Metadata.Title);
+                Assert.AreEqual("Evil Hacker", beatmap.Metadata.AuthorString);
+            }
+        }
+
+        [Test]
+        public void TestDecodeBeatmapWithMissingHeader()
+        {
+            Decoder<Beatmap> decoder = null;
+            Beatmap beatmap = null;
+
+            using (var resStream = TestResources.OpenResource("missing-header.osu"))
+            using (var stream = new StreamReader(resStream))
+            {
+                Assert.DoesNotThrow(() => decoder = Decoder.GetDecoder<Beatmap>(stream));
+                Assert.IsInstanceOf<LegacyBeatmapDecoder>(decoder);
+                Assert.DoesNotThrow(() => beatmap = decoder.Decode(stream));
+                Assert.AreEqual("Beatmap with no header", beatmap.Metadata.Title);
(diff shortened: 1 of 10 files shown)


Aergwyn:
I couldn't give this a comprehensive look thus my comment may look very stupid.

What about just extending `StreamReader` to give something like `PeekLine`?
It reads the line, keeps it and when you call `ReadLine` it returns any "peeked" line first before actually continuing reading the stream?
If the decoder only peeks the first line it wouldn't be lost.

Wouldn't that potentially solve the entire issue?

bdach:
Sure, adding another line-oriented buffering layer inside a custom extension of `StreamReader` would solve the biggest problem. The changes required for this would then be less invasive, although I don't know of anything like that in the standard library, so it would have to be handwritten and require testing and maintenance. Whether that's preferable, I don't know, but it's a good suggestion.

bdach:
I'm trying out @Aergwyn's suggestion and so far I'm liking that approach a lot more. Due to the amount of usages of `StreamReader`s the total line change count might be a little bigger than even here, but the changes are a lot more mechanical (replace old `StreamReader` with the new reader) and most importantly preserve decoder purity - so if there are no protests, I'll supersede this with that approach.

Task no. 1 on the list above is however still relevant and I'll still need to address it.

peppy:
@bdach For clarity can you provide a reason for closing this? Are you doing to re-PR with the alternative solution?

bdach:
Yep, I have another branch propped up that will supersede this, I just need to give it another look and test it out in-game (unit tests already present). Sorry for not clarifying.

PR #6074, opened 2019-09-11
Fix key counter lit animation slowing down on fail

### LINKED ISSUE #4993: Key counter lit animation slows down with the fail one
**Describe the bug:** The key counter `lit` animation seems to slow down with the fail animation. (Not a big issue, but that's probably not the correct design)

**Screenshots or videos showing encountered issue:** https://streamable.com/sd7o7

**osu!lazer version:** 2019.608.0

frenzibyte:
- Uses `GameplayClock?.CurrentTime ?? Clock.CurrentTime` for key states logic instead of binding our clock to it.

Closes #4993

### DIFF
--- a/osu.Game/Screens/Play/KeyCounter.cs
+++ b/osu.Game/Screens/Play/KeyCounter.cs
@@ -25,6 +25,11 @@ namespace osu.Game.Screens.Play
         private readonly List<KeyCounterState> states = new List<KeyCounterState>();
         private KeyCounterState currentState;
 
+        [Resolved(canBeNull: true)]
+        private GameplayClock gameplayClock { get; set; }
+
+        public double CurrentTime => gameplayClock?.CurrentTime ?? Clock.CurrentTime;
+
         public bool IsCounting { get; set; } = true;
         private int countPresses;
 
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Play/KeyCounter.cs
```diff
@@ -25,6 +25,11 @@ public abstract class KeyCounter : Container
         private readonly List<KeyCounterState> states = new List<KeyCounterState>();
         private KeyCounterState currentState;
 
+        [Resolved(canBeNull: true)]
+        private GameplayClock gameplayClock { get; set; }
+
+        public double CurrentTime => gameplayClock?.CurrentTime ?? Clock.CurrentTime;
```
peppy: why public?

peppy:
Did you test this with the... test? It crashes.

iiSaLMaN:
Yeah and on top of that, it works with CI so not on this PR's end

osu.Game/Screens/Play/KeyCounter.cs
iiSaLMaN: Was used to `Clock.CurrentTime`..., anyways will remove.

PR #6078, opened 2019-09-11
Fix dialog overlay playing double samples on show/hide

Joehuu:


peppy:
I'm not sure we want to bypass the `FocusedOverlay` hierarchy. Could you give some more background on why you think this change is correct and what testing you've done?

I can't seem to find this regresses behaviour, but I'm still cautious as focus is no longer being requested by the dialog while it is displayed.

peppy (dismissed):

peppy (approved):

PR #6079, opened 2019-09-11
Add spacing to mod icons on leaderboards

Joehuu:


peppy:
Is there a reason this isn't applied to `DrawableProfileScore` and `ModDisplay`? Maybe it would work better as a padding or margin if it is to be applied everywhere mods are used.

peppy (approved):

PR #6087, opened 2019-09-12
Key counter improvements

### LINKED ISSUE #4993: Key counter lit animation slows down with the fail one

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/6086
- Supersedes and closes https://github.com/ppy/osu/pull/6074 (by removing the gameplay clock usage).
- Closes #4993.
- Plays back actions in rewind in real-time.
- Removes complex state storage logic and instead uses real-time rewind playback.
- Adjusts visuals to feel better during very fast key presses.
- Remotes some unnecessary properties that will never be customised.


osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -21,7 +21,8 @@ public ManiaReplayFrame()
         public ManiaReplayFrame(double time, params ManiaAction[] actions)
             : base(time)
         {
-            Actions.AddRange(actions);
+            if (actions.Length > 0)
```
smoogipoo: Is this required?

peppy: i don't actually think so, else it would have been failing until now.

smoogipoo (approved):

PR #6095, opened 2019-09-13
Fix incorrect DI usage of IAPIProvider in many tests

peppy:
`OsuTestScene` was caching its dummy API and many tests were trying to access it in `BackgrounDependencyLoader`. It would get the parent API.

- Fixes `TestSceneAccountCreation` incorrectly binding (and never unbinding) from a bindable.
- Fixes (hopefully) randomly getting logged out after running tests.

### DIFF
--- a/osu.Game/Tests/Visual/OsuTestScene.cs
+++ b/osu.Game/Tests/Visual/OsuTestScene.cs
@@ -52,7 +52,7 @@ namespace osu.Game.Tests.Visual
             get
             {
                 if (UseOnlineAPI)
-                    throw new Exception("Using the OsuTestScene dummy API is not supported when UseOnlineAPI is true");
+                    throw new Exception($"Using the {nameof(OsuTestScene)} dummy API is not supported when {nameof(UseOnlineAPI)} is true");
 
                 return dummyAPI;
             }
@@ -64,7 +64,7 @@ namespace osu.Game.Tests.Visual
 
         /// <summary>
         /// Whether this test scene requires real-world API access.
-        /// If true, this will bypass the local <see cref="dummyAPI"/> and use the <see cref="OsuGameBase"/> provided one.
+        /// If true, this will bypass the local <see cref="DummyAPIAccess"/> and use the <see cref="OsuGameBase"/> provided one.
         /// </summary>
         protected virtual bool UseOnlineAPI => false;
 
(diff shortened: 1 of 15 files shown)


smoogipoo (dismissed):

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -46,13 +46,27 @@ public abstract class OsuTestScene : TestScene
         protected Storage LocalStorage => localStorage.Value;
 
         private readonly Lazy<DatabaseContextFactory> contextFactory;
+
+        protected IAPIProvider API
+        {
+            get
```
iiSaLMaN: this can be shortened to `get => dummyAPI ?? throw new Exception(...)` since `dummyAPI` is created only if `UseOnlineAPI` is false

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -46,13 +46,27 @@ public abstract class OsuTestScene : TestScene
...
+
+        private DummyAPIAccess dummyAPI;
+
         protected DatabaseContextFactory ContextFactory => contextFactory.Value;
 
         /// <summary>
-        /// Whether this test scene requires API access
-        /// Setting this will cache an actual <see cref="APIAccess"/>.
+        /// Whether this test scene requires real-world API access.
+        /// If true, this will bypass the local <see cref="dummyAPI"/> and use the <see cref="OsuGameBase"/> provided one.
```
iiSaLMaN:
```suggestion
        /// If true, this will bypass the local <see cref="DummyAPIAccess"/> and use the <see cref="OsuGameBase"/> provided one.
```

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -46,13 +46,27 @@ public abstract class OsuTestScene : TestScene
         protected Storage LocalStorage => localStorage.Value;
 
         private readonly Lazy<DatabaseContextFactory> contextFactory;
+
+        protected IAPIProvider API
+        {
+            get
+            {
+                if (UseOnlineAPI)
+                    throw new Exception("Using the OsuTestScene dummy API is not supported when UseOnlineAPI is true");
```
iiSaLMaN: should be `InvalidOperationException`?

iiSaLMaN:
```suggestion
                    throw new Exception($"Using the {nameof(OsuTestScene)} dummy API is not supported when {nameof(UseOnlineAPI)} is true");
```

frenzibyte (changes requested):

osu.Game/Tests/Visual/OsuTestScene.cs
peppy: i think it reads better as-is.

PR #6106, opened 2019-09-14
Fix RankingsHeader dropdown can be clickable when not visible

EVAST9919:


peppy:
Is there a reason for this container needing to exist? Can you not just apply the properties to the dropdown itself and remove the container?

peppy (approved):

PR #6113, opened 2019-09-14
Abstractify gameplay input handling out of replay

frenzibyte:
- Allows for creating other input handlers for gameplay (upcoming `VirtualInputHandler` for example).


peppy:
@frenzibyte please check back on this and determine if it is still the direction you see this going. I'm guessing this PR was largely ignored because there is zero context as to the path forward here, and no follow-up PRs showing how this would be used.

frenzibyte:
Right, the path forward I had in mind for this was attaching an `InputHandler` to the ruleset input manager and generate buttons associated to each ruleset action in which pressing it would enqueue an `IInput` with the action to that handler.

I've rechecked #6114 and it doesn't look to be in a good state to update and review, I'll close this chain and investigate if this can be done in a better way without having virtual input logic all over `DrawableRuleset`, or at least do it in a tidied way if it has to stay there.

PR #6114, opened 2019-09-14
Add a virtual input handler for gameplay

frenzibyte:
- Adds a virtual input handler to allow pressing actions by adding them to the actions list `VirtualInputHandler.Actions`.
- A dependency for implementing a virtual touch overlay for mobile users.
- [ ] Depends on #6113


iiSaLMaN:
**NOTE**: Positional input is not supported as of yet due to no multi-touch support, but this can be useful for the taiko ruleset currently.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -196,13 +201,30 @@ private void load(OsuConfigManager config, CancellationToken? cancellationToken)
...
+
             if ((ResumeOverlay = CreateResumeOverlay()) != null)
             {
-                AddInternal(CreateInputManager()
+                AddInternal(resumeInputManager
                     .WithChild(CreatePlayfieldAdjustmentContainer()
                         .WithChild(ResumeOverlay)));
             }
 
+            if (RuntimeInfo.IsMobile)
```
peppy: why only mobile?

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -196,13 +201,30 @@ private void load(OsuConfigManager config, CancellationToken? cancellationToken)
...
+            {
+                if (!(KeyBindingInputManager is IHasVirtualHandler virtualInputManager))
+                    throw new InvalidOperationException($"A {nameof(KeyBindingInputManager)} which supports virtual input handling is not available");
+
+                virtualInputManager.VirtualInputHandler = CreateVirtualInputHandler();
+                resumeInputManager.UseParentInput = false;
+
+                if (!ProvidingUserCursor)
+                {
+                    // The cursor is hidden by default (see Playfield.load()), but should be shown when there's a replay
```
peppy: this feels very weird/hacky.

peppy (changes requested):

peppy:
I see no method of testing this?

osu.Game/Rulesets/UI/DrawableRuleset.cs
iiSaLMaN: oh right touch users.., not sure how I can check that?

peppy: There's nothing to check. All users should be able to choose to use a virtual input if they want to.

iiSaLMaN:
> I see no method of testing this?

I was gonna PR the virtual touch overlay component along with tests but there seem to be some issues I have to fix, I'll write a simple test for now.

osu.Game.Rulesets.Catch/Input/CatchVirtualInputHandler.cs
```diff
@@ -0,0 +1,21 @@
...
+
+namespace osu.Game.Rulesets.Catch.Input
+{
+    public class CatchVirtualInputHandler : VirtualInputHandler
+    {
+        public override List<IInput> GetPendingInputs() => new List<IInput>
+        {
+            new GameplayInputState<CatchAction>
+            {
+                PressedActions = Actions.Cast<CatchAction>().ToList(),
```
peppy: why is `VirtualInputHandler` not generic instead of cast?

osu.Game/Rulesets/UI/DrawableRuleset.cs
iiSaLMaN: I can't really think of a proposal that isn't hacky to fix this, have any?

iiSaLMaN:
> I see no method of testing this?

Resolved.

PR #6117, opened 2019-09-14
Add fallback decoder option for badly-headered files

### LINKED ISSUE #5522: IOException: Unknown file format ([General])
Sentry Issue: [OSU-LAZER-2JB](https://sentry.io/organizations/ppy/issues/1089044667/?referrer=github_integration)

```
IOException: Unknown file format ([General])
  Module "osu.Game.Beatmaps.Formats.Decoder", line 171, in GetDecoder
    osu.Game.Beatmaps.Formats.Decoder`1[T] GetDecoder[T](System.IO.StreamReader)
  Module "osu.Game.Beatmaps.BeatmapManager", line 107, in createBeatmapDifficulties
    System.Collections.Generic.List`1[osu.Game.Beatmaps.BeatmapInfo] createBeatmapDifficulties(osu.Game.IO.Archives.ArchiveReader)
  Module "osu.Game.Beatmaps.BeatmapManager", line 3, in Populate
    System.Threading.Tasks.Task Populate(osu.Game.Beatmaps.BeatmapSetInfo, osu.Game.IO.Archives.ArchiveReader, System.Threading.CancellationToken)
  Module "osu.Game.Database.ArchiveModelManager`2+<>c__DisplayClass28_0+<<Import>b__0>d", line 122, in MoveNext
    Void MoveNext()

Unknown file format ([General])
```

Should not be reported as an error. Probably shouldn't be an unknown section either.

bdach:
Attempt number two. Supersedes #6043, aims to resolve #5522.

# Description

This PR introduces a way to register `LegacyDecoder`s as fallbacks for beatmaps and storyboards. This prevents import failures in the case of a corrupted or missing header in the decoded file.

The fallback registration isn't really much to talk about (one new `Dictionary`, same as before), but as before, the issue of `StreamReader` consuming input has to be addressed. As per the excellent suggestion of @Aergwyn, a line-buffered reader decorator operating on `StreamReader` instances was added. The decorator has a very limited API consisting of three main operations - `PeekLine()`, which allows to see the next line in the stream without consuming it, `ReadLine()`, which consumes and returns the next line in the stream, and `ReadToEnd()` which reads all the remaining text in the stream (including the unconsumed peeked line). Peeking line-per-line uses an internal queue of lines that have been read ahead from the underlying stream.

The addition of the line-buffered reader is a workaround solution to a problem with decoding. At current selecting a decoder works by irreversibly reading the first line from the stream and looking for a magic string that indicates the type of decoder to use.

It might however be possible for a file to be valid in format, just missing a header. In such a case a lack of a line-buffered reader makes it impossible to reparse the content of that first line. Introducing it however allows to peek the first line for magic first.

 - If magic is found in the first line, `GetDecoder()` will peek it and use to return the correct `Decoder` instance. Note that in the case of `JsonBeatmapDecoder` the magic is the opening JSON object brace, and therefore must not be consumed. This is the most finicky part of this PR, so I left some comments that hopefully guide whoever else happens to touch those areas next.
 - If magic is not found, the fallback decoder will be able to use it.

Due to some tricky logic this PR additionally introduces unit tests both to the new `LineBufferedReader` and the pre-existing `LegacyBeatmapDecoder`.

# Garbage files

Due to the above change, there is one major caveat to address. Right now an empty or whitespace-only file will fail to import, as will a file that violates the parsing constraints (ie. a map with circle size that's too large and won't parse), but a non-empty file with complete garbage (like a sectionless text file, might as well be the complete works of Shakespeare) will not fail to import, and instead show up as the default no-data beatmap in the song select as shown below:

osu_2019-09-15_01-10-52

This is why I changed one of the pre-existing tests to use an empty `MemoryStream` instead of a byte stream of `0x1, 0x3, 0x3, 0x7`. Now I've already noted previously that this is a possibility, and I have an idea how to fix it. A draft is [over in my fork](https://github.com/bdach/osu/commit/2629e50186a8c3ba85214c0571cd321e3c9f506a) - basically, the idea is to make all of the line-parsing functions return `true` on valid input and error out in the end of parsing if not even one valid line was read.

The reason I'm not including that commit in here (yet?) is because I find it ugly in how it really muddies the control flow due to the multiple-indented `switch` statements. If I had to add that, I would first start by factoring out the more complex `switch` cases to methods to reduce nesting, and then recursively apply the `bool` change until the entire call tree returns `bool`s, but it seems to me that such a change would muddle this PR and really should be in a separate one. If you feel I'm wrong, I'm more than happy to include that here.


smoogipoo (approved):
I think this is fine as it is for now 👍

PR #6126, opened 2019-09-15
Show notification when volume is muted

Craftplacer:
This took around under a hour to implement, it has the same behavior and content from osu!stable.
A notification will show up when the user has either its master or track volume set to the minimum value, clicking it will set it to the default values.

- [x] Depends on https://github.com/ppy/osu/pull/6216


osu.Game/Screens/Play/PlayerLoader.cs
```diff
@@ -101,6 +106,10 @@ private void load()
             });
 
             loadNewPlayer();
+
+            this.notificationOverlay = notificationOverlay;
```
iiSaLMaN: Can be resolved with the `Resolved` attribute instead since it's usage is out of the BDL
```csharp
[Resolved]
private NotificationOverlay notificationOverlay { get; set; }
```

iiSaLMaN:
What if the user muted the audio with the mute button?

Craftplacer:
> What if the user muted the audio with the mute button?

Doesn't check that, thanks for the info, I'll look into it.

Craftplacer:
Video of the pull request, if interested: https://streamable.com/vj0bo

osu.Game/Overlays/VolumeOverlay.cs
```diff
@@ -32,6 +32,12 @@ public class VolumeOverlay : OverlayContainer
 
         private readonly BindableDouble muteAdjustment = new BindableDouble();
 
+        public bool IsMuted
```
iiSaLMaN: Why not expose the bindable?
```csharp
public BindableBool IsMuted => muteButton.Current;
```

Craftplacer: Yeah right.

osu.Game/OsuGame.cs
```diff
@@ -485,7 +485,8 @@ protected override void LoadComplete()
                 toolbarElements.Add(d);
             });
 
-            loadComponentSingleFile(volume = new VolumeOverlay(), leftFloatingOverlayContent.Add);
+            dependencies.Cache(loadComponentSingleFile(volume = new VolumeOverlay(), leftFloatingOverlayContent.Add));
```
peppy:
```suggestion
            loadComponentSingleFile(volume = new VolumeOverlay(), leftFloatingOverlayContent.Add, true);
```

check `loadComponentSingleFile`'s implementation you can cache via a `true` parameter.

peppy (dismissed):

peppy:
In stable, this notification is only shown once per osu! session. We may want to consider this to reduce frustration for a user that may be playing intentionally muted (can be done via a `private static` in `PlayerLoader`

osu.Game/Screens/Play/PlayerLoader.cs
```diff
@@ -66,6 +66,8 @@ public class PlayerLoader : ScreenWithBeatmapBackground
         [Resolved]
         private VolumeOverlay volumeOverlay { get; set; }
 
+        private bool muteWarningShownOnce = false;
```
peppy: needs to be static else won't work (did you test this?)

peppy:
As this is quite hidden functionality, it's probably a good idea to add a test scene for this to avoid regressing at a future stage. Let me know if you need any guidance on how to make this.

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -24,19 +30,35 @@ namespace osu.Game.Tests.Visual.Gameplay
...
+            loader = new TestPlayerLoader(() => player);
+            container = new TestPlayerLoaderContainer(loader);
+            InputManager.Child = container;
+        }
+
         [Test]
         public void TestBlockLoadViaMouseMovement()
         {
-            AddStep("load dummy beatmap", () => stack.Push(loader = new TestPlayerLoader(() => new TestPlayer(false, false))));
+            AddStep("load dummy beatmap", () => ResetPlayer(false));
```
iiSaLMaN: Already reset in the setup? keep in mind `SetUp` is called once per `[Test]`

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -49,13 +71,13 @@ public void TestLoadContinuation()
...
 
-            AddStep("load dummy beatmap", () => stack.Push(loader = new TestPlayerLoader(() => player = new TestPlayer(false, false))));
+            AddStep("load dummy beatmap", () => ResetPlayer(false));
             AddUntilStep("wait for current", () => loader.IsCurrentScreen());
             AddStep("mouse in centre", () => InputManager.MoveMouseTo(loader.ScreenSpaceDrawQuad.Centre));
             AddUntilStep("wait for player to be current", () => player.IsCurrentScreen());
             AddStep("load slow dummy beatmap", () =>
             {
-                stack.Push(loader = new TestPlayerLoader(() => slowPlayer = new SlowLoadPlayer(false, false)));
+                InputManager.Children = container = new TestPlayerLoaderContainer(loader = new TestPlayerLoader(() => slowPlayer = new SlowLoadPlayer(false, false)));
```
iiSaLMaN: maybe pass the player creation to `ResetPlayer` and rename it to `LoadPlayer` instead
```suggestion
                LoadPlayer(() => new SlowLoadPlayer(false, false));
```

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -24,19 +30,35 @@ namespace osu.Game.Tests.Visual.Gameplay
...
-            InputManager.Child = stack = new OsuScreenStack { RelativeSizeAxes = Axes.Both };
+            ResetPlayer(false);
             Beatmap.Value = CreateWorkingBeatmap(new OsuRuleset().RulesetInfo);
         });
 
+        /// <summary>
+        /// Sets the input manager child to a new test player loader container instance.
+        /// </summary>
+        /// <param name="interactive">If the test player should behave like the production one.</param>
+        public void ResetPlayer(bool interactive)
```
iiSaLMaN: why public?
```suggestion
        private void resetPlayer(bool interactive)
```

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -97,6 +118,61 @@ public void TestModReinstantiation()
             AddAssert("player mods applied", () => playerMod2.Applied);
         }
 
+        [Test]
+        public void TestMutedNotification()
+        {
+            AddStep("set master volume to 0%", () => audioManager.Volume.Value = 0);
+            AddStep("reset notification lock", () => PlayerLoader.ResetNotificationLock());
+            //AddStep("reset notification overlay", () => notificationOverlay);
```
iiSaLMaN: can be removed

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -97,6 +118,61 @@ public void TestModReinstantiation()
...
+            {
+                var lastPlayer = player;
+                player = null;
+                lastPlayer.Restart();
+            });
+        }
+
+        private class TestPlayerLoaderContainer : Container
+        {
+            private TestPlayerLoader loader;
```
iiSaLMaN: useless.

frenzibyte (changes requested):
Make sure to do a self-review at your changes to avoid pushing code with commented lines and unnecessary members.

Craftplacer:
> Make sure to do a self-review at your changes to avoid pushing code with commented lines and unnecessary members.

Those changes weren't supposed to be reviewed yet, as the latest commit was a (stupid) attempt at updating this branch.

peppy:
Let us know if you need help/direction. It shouldn't be a hard one to implement.

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
Craftplacer: Can't be removed, other tests run before TestMutedNotification() and might change state of the game like the notification lock by PlayerLoader, and the master volume might be 100% for AppVeyor and other new clients.

Craftplacer:
> Let us know if you need help/direction. It shouldn't be a hard one to implement.

It isn't hard but adding a test scene seems to require me rework it. For example, the current state of my changes, I implemented now the Container approach with a NotificationOverlay and VolumeOverlay but the other tests seem to fail or get stuck now. I did set local variables like TestPlayer but yet it fails. https://streamable.com/183ky

I have a feeling that something is wrong with the approach since it requires me to change other things that were already implemented, if you have any suggestions how to proceed I'd love to hear them.

Current changes:
https://gist.github.com/Craftplacer/1814970991dc4ec7543fe59106ca6e19

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
Craftplacer: This is one approach peppy told me I could do, and the only one which seems to work with resetting.

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
iiSaLMaN: Why is this comment here? (it's not even executable)

Craftplacer: https://github.com/ppy/osu/pull/6126#issuecomment-533142187
Like I said this was a kinda an unintended commit without PR quality control.

osu.Game/Screens/Play/PlayerLoader.cs
```diff
@@ -56,6 +59,17 @@ public class PlayerLoader : ScreenWithBeatmapBackground
...
+        [Resolved]
+        private NotificationOverlay notificationOverlay { get; set; }
+
+        [Resolved]
+        private VolumeOverlay volumeOverlay { get; set; }
+
+        [Resolved]
+        private AudioManager audioManager { get; set; }
+
+        private static bool muteWarningShownOnce;
```
iiSaLMaN: You can use #6216 for this instead.

Craftplacer: Kinda conflicts what peppy said to me :shrug:
> ppy 09/16/2019
really these kinds of things should be stored somewhere more central, but we don't have anything for that yet
something like ConfigManager which doesn't save to disk
`class SessionStatics : ConfigManager` (cached in `GameBase`) could work well, but it's out of scope of what you're implementing

peppy: How does it conflict?

Craftplacer: Then it does not, I recalled hearing something like SessionStatics somewhere and it not fitting what I want to do.

iiSaLMaN: just a dependency you can use to save values for the whole session of the game
```csharp
[BackgroundDependencyLoader]
private void load(SessionStatics sessionStatics)
{
    if (!sessionStatics.Get(Statics.UsefulValue))
        // do something once.

    sessionStatics.Set(Statics.UsefulValue, true);
}

**Comment by peppy** [comment]
I've fixed up your tests. Let's wait for the static config manager to be merged before getting this in, though.

**Review by frenzibyte — DISMISSED** [review]

**Comment by smoogipoo** [comment]
Please update this to make use of https://github.com/ppy/osu/pull/6216 .

File: osu.Game/Screens/Play/PlayerLoader.cs [hunk 11]
```diff
@@ -473,5 +506,43 @@ private void load()
...
+
+                    return true;
+                };
+            }
+        }
+
+        /// <summary>
+        /// Sets <see cref="muteWarningShownOnce"/> to false, reserved for testing.
+        /// </summary>
+        public static void ResetNotificationLock()
```
iiSaLMaN: This function should go away. instead, you should resolve the `SessionStatics` in the test and reset it.

frenzibyte (changes requested):

frenzibyte (dismissed):

osu.Game/Screens/Play/PlayerLoader.cs
```diff
@@ -473,5 +506,35 @@ private void load()
...
+        private class MutedNotification : SimpleNotification
+        {
+            public MutedNotification()
+            {
+                Text = "Your music volume is set to 0%! Click here to restore it.";
+            }
+
+            public override bool IsImportant => true;
+
+            public override bool RequestsFocus => true;
```
smoogipoo: What's this?

Craftplacer: Boolean overrides so that the notification is focused. The RequestsFocus can be removed as it's from Drawable. What's wrong with it?

iiSaLMaN: @Craftplacer Is there an issue with removing it? if no, then this is just redundant.

iiSaLMaN:
@Craftplacer Next time when you want to fix / add something, make sure to split it as possible as you can. It would help us review your PR faster.

smoogipoo (approved):
Seems ok to me.

PR #6128, opened 2019-09-15
Fix Perfect/SuddenDeath legacy mod conversion during score import

### LINKED ISSUE #6091: PF mods should shows as a PF mod not SD and PF
**Describe the bug:** 
The mods in perfect should shows as a perfect only not perfect and sudden death.
image

**osu!lazer version:** 
2019.091.0
**Logs:** 
No Logs

bdach:
Resolves #6091.

# Description

Upon investigating the user report in #6091 that indicated that viewing replays using the Perfect mod would also display the Sudden Death mod icon despite Perfect being the more restrictive of the two, it turned out that the logic of importing legacy scores was missing that corner case. A similar case of Double Time/Nightcore mutual exclusion was handled, but PF/SD was missed.

Add analogous handling of PF/SD legacy mods for all four rulesets, and additionally cover a tiny fraction of all cases with unit tests. The most problematic cases (NC+HD and PF+SD) are covered in all four basic rulesets.

(Is there such a thing as "too many tests" over here?)

# Remarks

Since this is an import fix and the mod list is persisted to the database as a JSON array, this won't retroactively fix already imported scores and replays:

image

Fixing this retroactively would require some sort of fixup data migration, but I'm not sure if that's a thing that's worth doing.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/CatchLegacyModConversionTest.cs
+++ b/osu.Game.Rulesets.Catch.Tests/CatchLegacyModConversionTest.cs
@@ -1,0 +1,29 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using NUnit.Framework;
+using osu.Game.Beatmaps.Legacy;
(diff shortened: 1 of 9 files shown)


peppy (approved):
solid

PR #6132, opened 2019-09-16
Implement HumanizedNumber component for rankings overlay

EVAST9919:


osu.Game/Overlays/Rankings/HumanizedNumber.cs
```diff
@@ -0,0 +1,43 @@
...
+
+            Text = numberToText(value);
+        }
+
+        private string numberToText(long value)
+        {
+            var suffixes = new[]
+            {
+                "",
+                "k",
```
peppy: where are you getting these from? why is thousands "k" but the reset not shortened? also i'm pretty sure there's already an implementation of this exact thing somewhere else in the code base.

EVAST9919: it's just copy-paste from web side

EVAST9919: The rest isn't shortened because it's using full names in web
Screenshot_2

EVAST9919: I could just use `Humanizer.ToMetric`, but it works only for 32bit int and you can't manually modify the text after the number

peppy: There should be a comment showing where you got the implementation from.

peppy:
This should not be a drawable class. You should implement the "humanize" part as a unit tested helper class (probably in https://github.com/ppy/osu/blob/master/osu.Game/Utils/HumanizerUtils.cs). Also call it something that doesn't have "humanize" in the name since that is really misleading.

EVAST9919:
Will be reopened with suggested changes

PR #6134, opened 2019-09-16
Replace DrawableRuleset class with an interface

frenzibyte:
- Replaces abstract class `DrawableRuleset` with a drawable interface `IDrawableRuleset` as told by the xmldoc remarks.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -161,7 +161,7 @@ namespace osu.Game.Screens.Play
                 .WithChild(beatmapSkinProvider
                     .WithChild(target = rulesetSkinProvider)));
 
-            target.AddRange(new Drawable[]
+            target.AddRange(new[]
             {
                 DrawableRuleset as Drawable,
                 new ComboEffects(ScoreProcessor)
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -71,7 +71,7 @@ namespace osu.Game.Screens.Play
         private BreakOverlay breakOverlay;
 
         protected ScoreProcessor ScoreProcessor { get; private set; }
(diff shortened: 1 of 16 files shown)


peppy:
What's the point / reasoning of this change?

osu.Game/Screens/Play/Player.cs
```diff
@@ -161,9 +161,9 @@ private void addGameplayComponents(Container target, WorkingBeatmap working)
                 .WithChild(beatmapSkinProvider
                     .WithChild(target = rulesetSkinProvider)));
 
-            target.AddRange(new Drawable[]
+            target.AddRange(new[]
             {
-                DrawableRuleset,
+                DrawableRuleset as Drawable,
```
peppy: this is the reason for the xmldoc that you removed. casting an interface to a drawable like this is super-dangerous.

iiSaLMaN: but it's an `IDrawable`, it can't be inherited from a normal class, no?

iiSaLMaN:
> What's the point / reasoning of this change?

Just for code quality, I saw it while reading on the drawable ruleset implementation and thought why not do that because `IDrawable` already exists

osu.Game/Screens/Play/Player.cs
peppy: there's nothing in place to enforce as such, which is why we weren't using an interface until now. if you can explain what this change is for (ie. what is dependent on switching to an interface) that may help figure a correct direction here.

peppy:
On a quick git history check, `IDrawable` has always existed (since 2017) but that comment was added in 2019. Until `IDrawable` can be used in place of `Drawable` (without casts) switching to using it for cases like this is not feasible.

iiSaLMaN:
Closing this PR until `IDrawable` is used instead of `Drawable`

PR #6141, opened 2019-09-17
Randomise beatmap playback order on startup

### LINKED ISSUE #6135: The startup menu song should be randomised
**Describe the bug:** when triangles main theme ends the song played should be random, not always the same

**osu!lazer version:** 

**Logs:**

peppy:
Closes #6135.

Note that due to limitations of `BindableList`, newly imported beatmaps are always added at the end. This doesn't seem like much of an issue, though.

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -8,6 +8,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Input.Bindings;
+using osu.Framework.MathUtils;
 using osu.Framework.Threading;
 using osu.Game.Beatmaps;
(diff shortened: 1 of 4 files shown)


osu.Game/Overlays/MusicController.cs
```diff
@@ -24,7 +25,7 @@ public class MusicController : Component, IKeyBindingHandler<GlobalAction>
         [Resolved]
         private BeatmapManager beatmaps { get; set; }
 
-        private List<BeatmapSetInfo> beatmapSets;
+        public readonly BindableList<BeatmapSetInfo> BeatmapSets = new BindableList<BeatmapSetInfo>();
```
smoogipoo: This should at the very least be an IBindableList if this is how this is going to be done.

smoogipoo (changes requested):

smoogipoo (approved):

PR #6151, opened 2019-09-18
Always update drawable hitobject state on skin change

peppy:
Allows mods to apply changes, and state to potentially be more correctly restored after skin-specific changes.


peppy:
This may have regressed rewind behaviour somehow, blocking for now.

smoogipoo (approved):

PR #3594, opened 2018-10-14
Allowing auto-restarts on fail for Perfect and Sudden Death mod

plankp:
Partially fixes #3380 since this does perform the auto-restart, but it is not optional.

PR #6159, opened 2019-09-18
Add auto-restart functionality to sudden death and perfect mods

### LINKED ISSUE #3380: Add an auto-restart option after failing for Perfect
In osu!stable, whenever you fail on Perfect, the game automatically retries. In osu!lazer, that doesn't happen. This could be an additional option that could be picked in the settings. It could work not only for Perfect but also SuddenDeath. 

~~I might be going a bit too far with those stable to lazer ideas~~

peppy:
Imitates stable.

- Supersedes https://github.com/ppy/osu/pull/3594, which lost its fork.
- Closes #3380.


osu.Game/Rulesets/Mods/IApplicableFailOverride.cs
```diff
@@ -12,5 +12,10 @@ public interface IApplicableFailOverride : IApplicableMod
         /// Whether we should allow failing at the current point in time.
         /// </summary>
         bool AllowFail { get; }
+
+        /// <summary>
+        /// Whether we want to restart on fail. Only used if AllowFail is true.
```
iiSaLMaN:
```suggestion
        /// Whether we want to restart on fail. Only used if <see cref="AllowFail"/> is true.
```

smoogipoo (dismissed):

smoogipoo (approved):

PR #6166, opened 2019-09-19
Add setting to adjust hold-to-confirm activation time

### LINKED ISSUE #3731: Add hold esc for menu in options.
I asked some of my friends about this new feature. Most of them said it will be a little bit hard for old players to get used to this behavior. And it might causes note missing(or level fail). If there's a switch in options menu that can enable/disable this feature,it will be better.
Concern from #3686

### LINKED ISSUE #4714: The pause menu wait
When im playing osu!, im usually around people. So in osu stable i can easily pause the game and take off my headphones to chat to them and then continue, but with osu!lazer i cant do this because i need to wait, and while waiting for the pause menu to show up i miss a bunch of circles and sometimes die.

Please remove the wait and maybe add some sort of penalty for pausing too much? like in osu!stable when theres like a 5 second wait before pausing again, i think that would be good.

peppy:
Also adds a confirmation dialog when exiting the game with hold-to-confirm disabled.

- Closes https://github.com/ppy/osu/issues/4714.
- Closes https://github.com/ppy/osu/issues/3731

### DIFF
--- a/osu.Game/Overlays/HoldToConfirmOverlay.cs
+++ b/osu.Game/Overlays/HoldToConfirmOverlay.cs
@@ -51,7 +51,7 @@ namespace osu.Game.Overlays
 
         protected override void Dispose(bool isDisposing)
         {
-            audio.Tracks.RemoveAdjustment(AdjustableProperty.Volume, audioVolume);
+            audio?.Tracks.RemoveAdjustment(AdjustableProperty.Volume, audioVolume);
             base.Dispose(isDisposing);
         }
     }
--- a/osu.Game/Screens/Menu/MainMenu.cs
+++ b/osu.Game/Screens/Menu/MainMenu.cs
@@ -85,7 +85,11 @@ namespace osu.Game.Screens.Menu
                             OnEdit = delegate { this.Push(new Editor()); },
                             OnSolo = onSolo,
                             OnMulti = delegate { this.Push(new Multiplayer()); },
-                            OnExit = this.Exit,
+                            OnExit = delegate
+                            {
+                                exitConfirmed = true;
+                                this.Exit();
+                            },
                         }
                     }
                 },
--- a/osu.Game/Screens/Menu/MainMenu.cs
+++ b/osu.Game/Screens/Menu/MainMenu.cs
@@ -55,7 +55,7 @@ namespace osu.Game.Screens.Menu
         [Resolved]
         private IAPIProvider api { get; set; }
 
-        [Resolved]
+        [Resolved(canBeNull: true)]
         private DialogOverlay dialogOverlay { get; set; }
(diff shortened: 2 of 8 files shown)


smoogipoo:
This puts the game into a broken state when cancelling the confirmation.

smoogipoo (changes requested):
as mentioned

osu.Game/Overlays/HoldToConfirmOverlay.cs
```diff
@@ -51,7 +51,7 @@ private void load()
 
         protected override void Dispose(bool isDisposing)
         {
-            audio.Tracks.RemoveAdjustment(AdjustableProperty.Volume, audioVolume);
+            audio?.Tracks.RemoveAdjustment(AdjustableProperty.Volume, audioVolume);
```
iiSaLMaN: `audio` is never null.
```suggestion
            audio.Tracks.RemoveAdjustment(AdjustableProperty.Volume, audioVolume);
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -221,9 +239,48 @@ public override void OnResuming(IScreen last)
 
         public override bool OnExiting(IScreen next)
         {
+            if (holdDelay.Value == 0 && !exitConfirmed && dialogOverlay != null && !(dialogOverlay.CurrentDialog is ConfirmExitDialog))
+            {
+                dialogOverlay.Push(new ConfirmExitDialog(() =>
```
iiSaLMaN: duplicate action of exiting as the exit button, maybe move it to one function
```csharp
private void onExit()
{
    exitConfirmed = true;
    this.Exit();
}

File: osu.Game/Screens/Menu/MainMenu.cs [hunk 3]
```diff
@@ -221,9 +239,48 @@ public override void OnResuming(IScreen last)
...
+                {
+                    exitConfirmed = true;
+                    this.Exit();
+                }, () =>
```
iiSaLMaN: can be shortened (more readable to me)
```csharp
                }, () => exitConfirmOverlay.Abort()));
```

frenzibyte (changes requested):

osu.Game/Overlays/HoldToConfirmOverlay.cs
peppy: This is intentional for cases where there is an exception (and it never gets to BDL). Commonly experienced in tests. Basically we don't presume things always hit BDL for the sake of testing.

iiSaLMaN: Ah, I see.

PR #6171, opened 2019-09-19
Add "plus" type changelog entry

hbnrmx:
* adds "+" icon to changelog entries
* changes vertical alignment slightly

before:
before
after:
after


iiSaLMaN:
Can be closed in favour of #6177

PR #6177, opened 2019-09-19
Add support for "additions" in changelog

EVAST9919:


iiSaLMaN:
Supersedes #6171

bdach:
This kind of duplication of work could have been avoided. The other PR could have as well added the enumeration type instead of operating on magic strings if there was a prompt to do so in the review.

EVAST9919:
> Supersedes #6171

Oh, I didn't even notice

osu.Game/Overlays/Changelog/ChangelogBuild.cs
```diff
@@ -76,11 +76,11 @@ private void load(OsuColour colours)
 
                     var entryColour = entry.Major ? colours.YellowLight : Color4.White;
 
-                    title.AddIcon(FontAwesome.Solid.Check, t =>
+                    title.AddIcon(entry.Type == ChangelogEntryType.Fix ? FontAwesome.Solid.Check : FontAwesome.Solid.Plus, t =>
                     {
                         t.Font = fontSmall;
                         t.Colour = entryColour;
-                        t.Padding = new MarginPadding { Left = -17, Right = 5 };
+                        t.Padding = new MarginPadding { Left = -17, Top = 5 };
```
iiSaLMaN: can't this be done by positioning the icon at the left-center instead of adjusting the padding?
```csharp
t.Anchor = Anchor.CentreLeft;
t.Origin = Anchor.CentreLeft;
```

frenzibyte (changes requested):

osu.Game/Overlays/Changelog/ChangelogBuild.cs
peppy: Yep, These should hopefully not be necessary any more.

hbnrmx:
@bdach, no worries, code is cheap.  :)

osu.Game/Overlays/Changelog/ChangelogBuild.cs
EVAST9919: I tried. Not working as expected:
Screenshot_2

iiSaLMaN: Seems correct to me, we're talking about the vertical padding, you should adjust the horizontal padding yourself (i think setting `Origin` to `Anchor.CentreRight` will push it out to the left).

EVAST9919: Nope, that's not happening

smoogipoo: As in try:

```
Anchor = CentreLeft,
Origin = CentreLeft,
Padding = { Right = 5 }
```

EVAST9919: Still no

bdach: This works for me:

```csharp
title.AddIcon(entry.Type == ChangelogEntryType.Fix ? FontAwesome.Solid.Check : FontAwesome.Solid.Plus, t =>
{
    // ...
    t.Padding = new MarginPadding { Left = -4, Right = 5 };
    t.Anchor = Anchor.CentreRight;
    t.Origin = Anchor.CentreRight;
});

title.AddText(entry.Title, t =>
{
    // ...
    t.Anchor = Anchor.CentreLeft;
    t.Origin = Anchor.CentreLeft;
});
```

Result (with crude visual explanation - circles are anchor points):
image

I would have added the padding to the entry title text but it seems to affect spacing for some reason. Will recheck framework-side and open an issue where appropriate.

bdach: Ok, doing the padding on the title text affects spacing because it's in a text flow, so it's *not* a bug; if we wanted to set it there then this would need some container refactoring.

EVAST9919: > Ok, doing the padding on the title text affects spacing because it's in a text flow, so it's _not_ a bug; if we wanted to set it there then this would need some container refactoring.

It deserves an individual pull request then

peppy (approved):

PR #3085, opened 2018-07-20
Add labelled switch button component for the Beatmap Setup Screen

Rekkonnect:
Split from #3078

**Known issues:**
- [x] When changing the value of the switch button for the first time, the colours change in a strange way

PR #6187, opened 2019-09-20
Add a base class for all future labelled components

smoogipoo:
Split out from https://github.com/ppy/osu/pull/3085

Has two modes:

Padded:

image

Non-padded:

image

Some of the code is definitely going to change in the future since the designs require specific alignment, but I think this is fine for now?


peppy:
I think this is okay for now. We may need to adjust in the future if labels get long.

image

peppy (approved):

PR #6188, opened 2019-09-20
Make LabelledTextBox use the new LabelledComponent class

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/6187
- [x] https://github.com/ppy/osu/pull/6237


osu.Game/Screens/Edit/Setup/Components/LabelledComponents/LabelledTextBox.cs
```diff
@@ -3,127 +3,50 @@
...
-        private const float default_label_text_size = 16;
-
         public event TextBox.OnCommitHandler OnCommit;
 
-        public bool ReadOnly
-        {
-            get => textBox.ReadOnly;
-            set => textBox.ReadOnly = value;
-        }
+        protected new OsuTextBox Component => (OsuTextBox)base.Component;
```
peppy: any reason for casts rather than generics? also why calling this component not textbox?

peppy (changes requested):

peppy (approved):

peppy (approved):

PR #6190, opened 2019-09-20
Fix scores importing with deleted beatmap sets

### LINKED ISSUE #6109: osu!lazer doesn't run after update.
**The issue:**
osu!lazer doesn't run at all after the new update and rebooting the computer.

I don't know if it has anything to do with it, but before rebooting the PC the game was working, and I deleted all the beatmaps and reinstalled them from osu!stable for testing the new background video feature. Now when I open osu!lazer, it appears a black screen for 3 - 4 seconds and then the game closes.

I also proved to unistall the game and reinstall it, but with no success.

**osu!lazer version:** 
2019.913.0

**Logs:**

[database.log](https://github.com/ppy/osu/files/3612726/database.log)
[network.log](https://github.com/ppy/osu/files/3612727/network.log)
[runtime.log](https://github.com/ppy/osu/files/3612728/runtime.log)

**Computer Specifications:** 
CPU: Intel Core i7-3770
RAM: 8,00 GB
GPU: AMD Radeon R9 280X
OS: Windows 10 x64

Edit: Mispelled words.
Edit2: Logs added.

smoogipoo:
Fixes #6109

I've added a test which _should_ fail on master, but doesn't. Haven't yet determined the cause of this, but it seems that in this particular unit test EFCore is not detecting changes to `DeletePending`, even though the line is being hit.

### DIFF
--- a/osu.Game.Tests/Scores/IO/ImportScoreTest.cs
+++ b/osu.Game.Tests/Scores/IO/ImportScoreTest.cs
@@ -144,8 +144,8 @@ namespace osu.Game.Tests.Scores.IO
                     beatmapManager.Delete(beatmapManager.QueryBeatmapSet(s => s.Beatmaps.Any(b => b.ID == imported.Beatmap.ID)));
                     Assert.That(scoreManager.Query(s => s.ID == imported.ID).DeletePending, Is.EqualTo(true));
 
-                    await scoreManager.Import(imported);
-                    Assert.That(scoreManager.Query(s => s.ID == imported.ID).DeletePending, Is.EqualTo(true));
+                    var secondImport = await loadIntoOsu(osu, imported);
+                    Assert.That(secondImport, Is.Null);
                 }
                 finally
                 {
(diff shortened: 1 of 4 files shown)


osu.Game.Tests/Scores/IO/ImportScoreTest.cs
```diff
@@ -117,6 +117,43 @@ public async Task TestImportStatistics()
...
+
+                    var imported = await loadIntoOsu(osu, toImport);
+
+                    var beatmapManager = osu.Dependencies.Get<BeatmapManager>();
+                    var scoreManager = osu.Dependencies.Get<ScoreManager>();
+
+                    beatmapManager.Delete(imported.Beatmap.BeatmapSet);
+                    Assert.That(scoreManager.Query(s => s.ID == imported.ID).DeletePending, Is.EqualTo(true));
+
+                    await scoreManager.Import(imported);
```
peppy: shouldn't we be checking that the import failed here? might be better to revert the original `loadIntoOsu` code but change it to `FirstOrDefault`, and assert equal to null.

peppy (changes requested):

smoogipoo:
Adjusted the test as requested. Still doesn't fail on master, as expected.

peppy (approved):

PR #6192, opened 2019-09-20
Specify model name on initialize and fail import notification messages

### LINKED ISSUE #6180: Skins and Beatmaps "Import" notifications are identical
**Describe the issue:** Starting an import from scratch, Lazer starts an import for beatmaps and another import for skins, but both of them start with "_Import is initialising..._", I almost thought it was a bug (like if it started two imports for beatmaps):

image

**Proposal designs of the feature:** I guess it's better for skins a message like "_Skins import is initialising..._", and the same thing for beatmaps: "_Beatmaps import is initialising..._", and/or with a different icon (still missing) based on the import.

frenzibyte:
Closes #6180

### DIFF
--- a/osu.Game/Database/ArchiveModelManager.cs
+++ b/osu.Game/Database/ArchiveModelManager.cs
@@ -111,7 +111,7 @@ namespace osu.Game.Database
         protected async Task Import(ProgressNotification notification, params string[] paths)
         {
             notification.Progress = 0;
-            notification.Text = $"{HumanisedModelName.Pluralize(paths.Length == 1)} import is initialising...".Humanize();
+            notification.Text = $"{HumanisedModelName.Humanize().Pluralize(paths.Length == 1)} import is initialising...";
 
             int current = 0;
 
@@ -147,7 +147,7 @@ namespace osu.Game.Database
 
             if (imported.Count == 0)
             {
-                notification.Text = $"{HumanisedModelName.Pluralize(paths.Length == 1)} import failed!".Humanize();
+                notification.Text = $"{HumanisedModelName.Humanize().Pluralize(paths.Length == 1)} import failed!";
                 notification.State = ProgressNotificationState.Cancelled;
             }
             else
--- a/osu.Game/Database/ArchiveModelManager.cs
+++ b/osu.Game/Database/ArchiveModelManager.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
+using Humanizer;
 using JetBrains.Annotations;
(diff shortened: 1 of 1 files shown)


bdach (commented):
If you're doing `Humanize()` everywhere, you missed one on [line 134](https://github.com/ppy/osu/blob/8300e86f20815f9add57bdc4132437186af7346d/osu.Game/Database/ArchiveModelManager.cs#L134) and another on [line 157](https://github.com/ppy/osu/blob/8300e86f20815f9add57bdc4132437186af7346d/osu.Game/Database/ArchiveModelManager.cs#L157). Other methods also use `HumanisedModelName` raw, might be actually preferable to use Humanizer [in the getter itself](https://github.com/ppy/osu/blob/8300e86f20815f9add57bdc4132437186af7346d/osu.Game/Database/ArchiveModelManager.cs#L599).

Not an expert on Humanizer, but we might want to do `Pluralize()` instead wherever plurals are involved?

smoogipoo:
+1 to `Pluralize`

iiSaLMaN:
> Not an expert on Humanizer, but we might want to do `Pluralize()` instead wherever plurals are involved?

I actually used humanize to make the text start with a capital letter, will humanize the entire text instead.

peppy:
I find the `Humanize` call on the whole string a bit confusing here. Would probably rather it be local to the model name call.

iiSaLMaN:
> I find the `Humanize` call on the whole string a bit confusing here. Would probably rather it be local to the model name call.

Resolved.

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -110,7 +111,7 @@ public Task Import(params string[] paths)
         protected async Task Import(ProgressNotification notification, params string[] paths)
         {
             notification.Progress = 0;
-            notification.Text = "Import is initialising...";
+            notification.Text = $"{HumanisedModelName.Humanize().Pluralize(paths.Length == 1)} import is initialising...";
```
peppy: This should not be plural. Also you probably want `.Humanize(LetterCasing.Title)`.

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -146,7 +147,7 @@ protected async Task Import(ProgressNotification notification, params string[] p
 
             if (imported.Count == 0)
             {
-                notification.Text = "Import failed!";
+                notification.Text = $"{HumanisedModelName.Humanize().Pluralize(paths.Length == 1)} import failed!";
```
peppy: ditto, pluralize will be incorrect in every case.

peppy (approved):

PR #6197, opened 2019-09-20
Fix typo

jorolf:
Not sure why nobody has noticed this yet...


bdach:
You got ninja'd by #6194.

PR #6206, opened 2019-09-21
Song select filter improvements

### LINKED ISSUE #6186: Add creator= and artist= search filters in song select
Recently added in stable https://github.com/ppy/osu-stable-issues/issues/50.

bdach:
* Adds `creator=` and `artist=` filters to resolve #6186.
* Adds support for half-open interval queries.
* Fixes culture variance when parsing filters with decimal points.
* Fixes handling of floating-point interval queries.
* Adds comprehensive filter unit test suite.

# Summary

This is quite a big one - I kind of just wanted to implement #6186, but as there were no tests, I started adding some and hitting some things that I wanted to improve, so this grew in scope quite a bit. And splitting this into multiple PRs felt like just creating noise for the sake of it.

I realise this might be hard to review. It might be a little easier going commit-by-commit, but this is quite large anyways. Feel free to merge this at your leisure.

## New filters

To match newest stable additions, add `creator=` and `artist=` filters to the beatmap carousel on song select screen. Contrary to stable, this implementation supports phrase queries with spaces within using double quotes.

The quote handling is not entirely correct (can't nest), but quotes should rarely happen within names, and it is an edge case of an edge case - leaving best-effort as is.

## Support for half-open intervals

A single `IsInclusive` field caused unexpected issues when trying to formulate a half-open interval query (last operator chosen won). Split out `IsInclusive` into two fields, `Is{Lower,Upper}Inclusive` and update usages accordingly.

## Culture variance fix

Culture was not taken into account when parsing filters, which meant that in cultures that use the comma (`,`) as a decimal delimiter, it would conflict with the comma used to delimit search criteria. To remove any ambiguity, introduce local helper functions that allow the decimal point to be utilised, using the invariant culture. This also matches stable behaviour.

The decision to not reuse `osu.Game.Beatmaps.Formats.Parsing` was deliberate due to differing semantics (it's not really sane to throw exceptions on receiving user-facing input, and catching them didn't feel right).

## Floating-point interval fix

Due to floating-point rounding and representation errors, filters could wrongly display results incongruous with the wedge display text (ie. a beatmap with the BPM of 139.99999 would be displayed as having 140 BPM and also pass the `bpm<140` filter).

Apply tolerance when parsing floating-point constraints. The tolerance chosen is half of what the UI displays for the particular values (so for example half of 0.1 for AR/DR/CS, 0.01 for stars, etc.)

Spotted this at the very end, so that kind of invalidated most of the `IsInclusive` changes (but not all of them).

## Testing

Lots of unit tests added, both for parsing and filter evaluation, with one end-to-end visual test modification. For testability I extracted the parsing logic out of its original place at the very start.

### DIFF
--- a/osu.Game.Tests/NonVisual/Filtering/FilterMatchingTest.cs
+++ b/osu.Game.Tests/NonVisual/Filtering/FilterMatchingTest.cs
@@ -12,7 +12,7 @@ namespace osu.Game.Tests.NonVisual.Filtering
     [TestFixture]
     public class FilterMatchingTest
     {
-        private readonly BeatmapInfo exampleBeatmapInfo = new BeatmapInfo
+        private BeatmapInfo getExampleBeatmap() => new BeatmapInfo
         {
             Ruleset = new RulesetInfo { ID = 5 },
             StarDifficulty = 4.0d,
@@ -25,10 +25,10 @@ namespace osu.Game.Tests.NonVisual.Filtering
             Metadata = new BeatmapMetadata
             {
                 Artist = "The Artist",
-                ArtistUnicode = "The Artist",
+                ArtistUnicode = "check unicode too",
                 Title = "Title goes here",
                 TitleUnicode = "Title goes here",
-                AuthorString = "Author",
+                AuthorString = "The Author",
                 Source = "unit tests",
                 Tags = "look for tags too",
             },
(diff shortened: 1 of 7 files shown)


iiSaLMaN:
This PR seems so big to be reviewed and exceeded the maximum lines of code that a member would review in general: (500 loc), would be better to separate this PR into couple of small PRs for making it easier to be merged.

bdach:
Closing due to size, sorry. Will split out into several by commits in the near future. CodeFactor is complaining anyways, so going to have to think what to do with that.

peppy:
This is fine as-is. Most of the size are tests anyway.

peppy:
Thanks for the quality contribution! I can't stress how much we appreciate this level of thought and testing in community PRs.

peppy (approved):

PR #6210, opened 2019-09-22
Allow top-level menu key pressed to progress the osu! logo

### LINKED ISSUE #6207: Menu progression keyboard shortcuts don't work to enter menu from osu! logo
Pressing valid keyboard shortcuts (p for play, d for osu!direct, e for editor etc) while in the title screen (the one with just the logo and nothing else) will not progress the menus. Once you press enter or spacebar to get to the main menu will they work normally. 

ESC does not clear the beatmap search bar, instead it goes back to the main menu.

Hold-to-confirm activation setting affects both going to the menu in-game and quick-reset. In stable, the menu was instant but quick-reset had a delay.

If these aren't intentional behavior changes, these should be brought over from stable.
 
**Notes**
Personally, I am very used to launching stable and pressing "p" 3 times to go to beatmap select.

Interestingly enough, ESC does clear the search bar in osu!direct. If the search bar is clear, it will go back to the main menu. On stable's osu!direct however, ESC goes straight back to the title screen regardless of if the search bar is empty or not. I'd say it behaves worse on stable.

Regarding the hold-to-confirm behavior, I'd argue they deserve to have different timings. 
The menu button serves as a pause button, so no delay means you pause exactly where you want to and a situation where you didn't pause because you didn't hold down the key long enough doesn't happen.
The quick-reset button delay however prevents accidental presses from resetting your current progress. Unlike with a pause button, the delay isn't a problem as you weren't going to resume anyways. Accidental presses can happen when quick-reset is bound to be near something else. For me quick-reset is on tilde, so I've definitely had a situation where both ESC and tilde get pressed.
I realize this can be solved by just changing keybinds to be further away, but it's worth proposing in my opinion.
 
**osu!lazer version**
2019.919.0

peppy:
Closes #6207.

### DIFF
--- a/osu.Game/Screens/Menu/ButtonSystem.cs
+++ b/osu.Game/Screens/Menu/ButtonSystem.cs
@@ -185,7 +185,7 @@ namespace osu.Game.Screens.Menu
         {
             if (State == ButtonSystemState.Initial)
             {
-                if (buttonsTopLevel.Any(b => e.PressedKeys.Contains(b.TriggerKey)))
+                if (buttonsTopLevel.Any(b => e.Key == b.TriggerKey))
                 {
                     logo?.Click();
                     return true;
--- a/osu.Game/Screens/Menu/ButtonSystem.cs
+++ b/osu.Game/Screens/Menu/ButtonSystem.cs
@@ -187,7 +187,7 @@ namespace osu.Game.Screens.Menu
             {
                 if (buttonsTopLevel.Any(b => e.PressedKeys.Contains(b.TriggerKey)))
                 {
-                    logo.Click();
+                    logo?.Click();
                     return true;
                 }
             }
--- a/osu.Game/Screens/Menu/ButtonSystem.cs
+++ b/osu.Game/Screens/Menu/ButtonSystem.cs
@@ -14,6 +14,7 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Input.Bindings;
+using osu.Framework.Input.Events;
 using osu.Framework.Logging;
 using osu.Framework.Platform;
 using osu.Framework.Threading;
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Menu/ButtonSystem.cs
```diff
@@ -180,6 +181,20 @@ private void updateIdleState(bool isIdle)
                 State = ButtonSystemState.Initial;
         }
 
+        protected override bool OnKeyDown(KeyDownEvent e)
+        {
+            if (State == ButtonSystemState.Initial)
+            {
+                if (buttonsTopLevel.Any(b => e.PressedKeys.Contains(b.TriggerKey)))
+                {
+                    logo.Click();
```
iiSaLMaN: missing a null check?
```suggestion
                    logo?.Click();
```

frenzibyte (changes requested):

osu.Game/Screens/Menu/ButtonSystem.cs
peppy: Theoretically this should never occur, since it can only go null when switching out to another screen (at which point these hotkeys cannot be triggered due to `State`) but I guess it's fine to add for safety.

frenzibyte (dismissed):

osu.Game/Screens/Menu/ButtonSystem.cs
```diff
@@ -180,6 +181,20 @@ private void updateIdleState(bool isIdle)
                 State = ButtonSystemState.Initial;
         }
 
+        protected override bool OnKeyDown(KeyDownEvent e)
+        {
+            if (State == ButtonSystemState.Initial)
+            {
+                if (buttonsTopLevel.Any(b => e.PressedKeys.Contains(b.TriggerKey)))
```
smoogipoo: Is there a reason this isn't just `e.Key == `?

peppy: no, that sounds better.

smoogipoo (approved):

PR #6211, opened 2019-09-22
Add an onboarding screen for tournament client

peppy:
Helps solve the most common issues we experience with tournament client setup.

### DIFF
--- a/osu.Game.Tournament/Screens/SetupScreen.cs
+++ b/osu.Game.Tournament/Screens/SetupScreen.cs
@@ -7,6 +7,8 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Game.Graphics.Sprites;
 using osu.Game.Graphics.UserInterface;
+using osu.Game.Online.API;
+using osu.Game.Overlays;
 using osu.Game.Screens.Edit.Setup.Components.LabelledComponents;
 using osu.Game.Tournament.IPC;
 using osuTK;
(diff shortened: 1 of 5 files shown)


osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -0,0 +1,142 @@
...
+            InternalChild = fillFlow = new FillFlowContainer
+            {
+                RelativeSizeAxes = Axes.X,
+                AutoSizeAxes = Axes.Y,
+                Direction = FillDirection.Vertical,
+                Padding = new MarginPadding(10),
+                Spacing = new Vector2(10),
+            };
+
+            api.LocalUser.BindValueChanged(_ => Schedule(reload));
```
iiSaLMaN: I think it's fine triggering the change immediately instead of calling again
```suggestion
            api.LocalUser.BindValueChanged(_ => Schedule(reload), true);
```

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -0,0 +1,142 @@
...
+                        loginOverlay.State.Value = Visibility.Visible;
+                    },
+                    Value = api?.LocalUser.Value.Username,
+                    Failing = api?.IsLoggedIn != true,
+                    Description = "In order to access the API and display metadata, a login is required."
+                }
+            };
+        }
+
+        private class ActionableInfo : LabelledComponent
```
iiSaLMaN: What do you think about renaming this to `LabelledFailableAction` or similar?

osu.Game.Tournament/Screens/SetupScreen.cs
peppy: it changes behaviour. the initial load would be scheduled.

osu.Game.Tournament/Screens/SetupScreen.cs
smoogipoo: It's very local, it's fine.

smoogipoo (approved):

PR #6216, opened 2019-09-23
Add session-wide storage as a safer alternative to statics

frenzibyte:
- Introduces SessionStatics component which is based on an in-memory config manager that can be used to save values for the entire session of the game.


peppy:
Can you update the one case you found for use here so there is a sample usage? (I believe it was somewhere in the menu?)

iiSaLMaN:
> Can you update the one case you found for use here so there is a sample usage? (I believe it was somewhere in the menu?)

You mean whether the muted notification is shown?

peppy:
No, you mentioned one other case already in master, I believe.

smoogipoo:
Why SessionStatics and not SessionConfigManager?

iiSaLMaN:
> Why SessionStatics and not SessionConfigManager?

It's not really a configuration so it would be kinda misleading adding `ConfigManager` to it in my opinion.

iiSaLMaN:
Fixed crash issue, happened due to the config manager not initialized with values for their lookups, I moved the in-memory logic to a base class in case we wanted in-memory store later on.

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -198,10 +200,10 @@ protected override void LogoArriving(OsuLogo logo, bool resuming)
 
             bool displayLogin()
             {
-                if (!loginDisplayed)
+                if (!statics.Get<bool>(Statics.LoginOverlayDisplayed))
```
peppy: i think it'd be cleaner to use the BDL load method to retrieve a bindable only, rather than querying `statics` multiple times.

peppy (changes requested):

peppy (approved):

PR #6219, opened 2019-09-23
Move common logic to IntroScreen

LeNitrous:
Cleans up code for all intros as well as playing custom welcome and see ya samples if supporter.


smoogipoo:
I'd like these changes not include the seeya/welcome changes, because those are going to need a little more thought re: how that's going to work with the API. As it is it's going to sometimes fail and sometimes not depending on whichever happens first: login or BDL load.

peppy:
@LeNitrous Please check my further commits which bring this up to standard. You missed many other components which should be addressed when making this kind of code quality change.

peppy (approved):

smoogipoo (approved):

PR #6236, opened 2019-09-24
Fix mania scrolling speed not including beatmap slider multiplier

### LINKED ISSUE #2870: osu!mania scroll speed is indicated in milliseconds but scales with BPM
in stable, bpm scaling is standard behaviour and can be turned off in the options. In lazer however, that option doesn't exist and scrollspeed is indicated in scrolltime which suggests the speed to be fixed, however it's not. Currently it seems to be ```indicated_scrolltime x 50 / BPM = final_scrolltime``` (based on observation) making the result of a given value unpredictable (just like stable where no player knows what the number actually means and everyone has to try out different values to find one that "feels right").

Indicating the speed in scrolltime is a good thing as it gives the number an understandable meaning, but if BPM scaling is active then the overlay that shows up when you change the scrollspeed should display both the input value and the resulting time after BPM scaling is applied.

This brings along the problem that lazer currently doesn't have an option for fixed scrollspeed, which is a must-have. I'd even go so far to question whether or not we want to keep BPM scaling (and if yes, which of the two should be default). It is important to note that even though the default in stable is BPM scaling, almost all players use fixed scrollspeed.

smoogipoo:
Fixes #2870

Only affects osu!mania / `RelativeScaleBeatLengths`. Also fixes hitobject lifetime / scrolling HOC invalidation issues which are also tested by the new test.


osu.Game/Rulesets/UI/Scrolling/DrawableScrollingRuleset.cs
```diff
@@ -131,7 +131,7 @@ private void load()
                     if (duration > maxDuration)
                     {
                         maxDuration = duration;
-                        baseBeatLength = timingPoints[i].BeatLength;
+                        baseBeatLength = timingPoints[i].BeatLength / Beatmap.BeatmapInfo.BaseDifficulty.SliderMultiplier;
```
peppy: isn't it saner to just remove this from line line 160 below, rather than doing this? tested and seems to work correctly and pass tests.

peppy (changes requested):

osu.Game/Rulesets/UI/Scrolling/DrawableScrollingRuleset.cs
smoogipoo: But we do want this no? As we discussed yesterday it's the base beatmap speed. Timing point velocity is multiplied into this inside `MultiplierControlPoint`.

smoogipoo: Will add a comment + test.

peppy (approved):
i *think* this is ok

PR #5520, opened 2019-07-30
Implement Main menu background mode setting

EVAST9919:
Fixes https://github.com/ppy/osu/issues/5516

PR #6239, opened 2019-09-24
Add option to show beatmap backgrounds at main menu

### LINKED ISSUE #5516: Beatmap backgrounds in the main menu
We might want to add an ability to allow players select beatmap backgrounds instead of default in the main menu.
Since we have an ability to skin that, we can make a dropdown in settings like

Main menu background
- default
- skin
- beatmap
- seasonal (in the future)

Some elements like "skin" and "beatmap" will be visible for supporters only (or not).

LeNitrous:
Fixes #5516 and supersedes https://github.com/ppy/osu/pull/5520

Allows the user to use these background modes:
- Default
    - uses the default background or skin if the user is supporter and the skin has one
- Beatmap
    - uses the currently playing beatmap's background if the user is supporter and falls back to the default backgrounds if the beatmap doesn't have a background

This extracts `BeatmapBackground` from `WorkingBeatmap` to be used by `BackgroundScreenDefault`.

Example for Beatmap: 
image


LeNitrous:
Currently `ScalingBackgroundScreen` doesn't update. Will look into it.

EVAST9919:
Login/out will trigger background update even if it's not suppose to (for example if user is not a supporter and background mode is default)

LeNitrous:
Stable also behaves this way unless unintended.

peppy:
We don't make placeholders in lazer, so please remove that for now.

osu.Game/Graphics/Containers/ScalingContainer.cs
```diff
@@ -154,6 +155,15 @@ private void updateSize()
 
         private class ScalingBackgroundScreen : BackgroundScreenDefault
         {
+            private Bindable<WorkingBeatmap> beatmap;
+
+            [BackgroundDependencyLoader]
+            private void load(IBindable<WorkingBeatmap> beatmap)
+            {
+                this.beatmap = (Bindable<WorkingBeatmap>)beatmap;
+                this.beatmap.ValueChanged += _ => Next();
```
UselessToucan: Since there is only one field usage, I suggest to replace it with `(Bindable<WorkingBeatmap>)beatmap.ValueChanged += _ => Next();`

peppy: you shouldn't be casting in the first place. create a bound copy and bind to its event. the method here is 100% wrong and causes a memory leak.

peppy (approved):

PR #6249, opened 2019-09-24
Fix mania bar line alignment

### LINKED ISSUE #6215: Mis-aligned beat bar in mania mode
**Describe the bug:** 

Falling beat hint lines in mania mode are not aligned with actual beat nor the falling notes (with correct timing of course). The beatmap used in the screenshots is [787919](https://osu.ppy.sh/beatmapsets/787919#osu/1653685), but the issue isn't limited to this one only.

* Missing beat hint bars and later misaligned ones at commit `af3bb5a2c`
osu_2019-09-23_00-02-45

* Correct display of beat hint bars at commit `fc7e8dcbf1`
osu_2019-09-23_00-00-18

**osu!lazer version:** 

The issue was discovered on git master, at the time of writing, `9fe7675be81`. I attempted git bisect and I think the issue is introduced with commit `af3bb5a2c` and merge request #6054.

**Logs:** 
[network.log](https://github.com/ppy/osu/files/3640679/network.log)
[performance.log](https://github.com/ppy/osu/files/3640680/performance.log)
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/3640681/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/3640683/performance-gamethread.draw.log)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/3640685/performance-gamethread.input.log)
[performance-gamethread.update.log](https://github.com/ppy/osu/files/3640686/performance-gamethread.update.log)
[runtime.log](https://github.com/ppy/osu/files/3640687/runtime.log)

bdach:
Resolves #6215.

For the effect of this PR, see this [quick-and-dirty comparison video](https://streamable.com/jujoc), showing both mania (for both scrolling directions) and taiko gameplay (shown to demonstrate lack of regressions - although the framerate foiled my plans just a *little* bit).

# Summary

As stated in the cause analysis in the aforementioned issue, the refactoring in #6054 inadvertently caused mania's anchoring to be dropped, because migrating `BarLine` out of the ruleset-specific hitobject inheritance hierachies also lead to the drawables moving from `DrawableManiaHitObject` (which contained the appropriate anchoring logic) to `DrawableHitObject`.

To address this issue, introduce an `IBarLine` interface, which together with generic constraints helps decouple `BarLineGenerator` from the actual hitobject types it creates. Thanks to this, all rulesets that want bar lines can provide an implementation of `IBarLine` that also derives from the base hitobject
class.

This allows `DrawableBarLines` in taiko and mania to be migrated back to `DrawableTaikoHitObject` and `DrawableManiaHitObject` base classes respectively. This in turn resolves #6215 without code duplication, since the missing anchoring application is now done again in mania's `DrawableBarLine` through deriving from `DrawableManiaHitObject`.

# Remarks

I struggled with the form of this for a while. I didn't want to either widen the generic constraints on `Drawable{Mania,Taiko}HitObject`s, or to duplicate the anchoring logic (since it was tied to a bindable). This is a solution I got.

Also included are some new regression test steps, albeit I have to admit they're not very strong. The mania case tests the anchoring itself (while probably good enough, that's not really what I *wanted* to test per se, and it's hard to visually recognize if that test should pass or not), and the taiko step is purely a visual one, without asserts. I'd be incredibly willing to listen to feedback if there are ideas how to make those test cases stronger (best way would be to somehow inspect if the relative position of the bar lines to normal hitobjects is correct).


smoogipoo (approved):
good stuff 👍

PR #6258, opened 2019-09-25
Implement HumanizerUtils.Humanize for long numbers

EVAST9919:


osu.Game/Utils/HumanizerUtils.cs
```diff
@@ -26,5 +26,32 @@ public static string Humanize(DateTimeOffset input)
...
+                "billion",
+                "trillion",
+                "quadrillion",
+                "quintillion",
+            };
+
+            if (input < k)
+                return input.ToString();
+
+            int i = (int)Math.Floor(Math.Log(input) / Math.Log(k));
```
bdach: Mind covering this with edge case unit tests (10<sup>n - 1</sup> and 10<sup>n</sup>) just for safety's sake? It's not immediately obvious to me this will behave correctly due to taking the natural logarithm (base *e*) instead of base 10. It's *mathematically* sound, but floats do float things.

Also, you might as well take `Math.Log(input, k)`, no?

EVAST9919: Yeah, you are right about 10(3n - 1)

bdach:
The visual tests are nice, but reworking them a bit to use steps and asserts might be even nicer, then any regressions in the printing will fail the build.

osu.Game/Utils/HumanizerUtils.cs
```diff
@@ -26,5 +26,32 @@ public static string Humanize(DateTimeOffset input)
...
+                "billion",
+                "trillion",
+                "quadrillion",
+                "quintillion",
+            };
+
+            if (input < k)
+                return input.ToString();
+
+            int i = (int)Math.Floor(Math.Round(Math.Log(input, k)));
```
bdach: This rounds the edge case (3n-1)-digit numbers up now:

image

When I suggested checking I actually wanted to see if it would round *down* correctly for large values. Looking at what `Math.Log` does, it unfortunately doesn't handle large bases and large numbers well in that case:

```csharp
> Math.Round(Math.Log(999_999_999_999_999, 1000))
5
> Math.Log(1_000_000_000_000_000, 1000)
5
> Math.Log(1_000_000_000_000_000, 1000) - Math.Log(999_999_999_999_999, 1000)
0
```

Not sure what kind of behaviour we want to get here from a UX standpoint.

osu.Game.Tests/Visual/UserInterface/TestSceneHumanizedNumber.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+                AutoSizeAxes = Axes.Both,
+                Direction = FillDirection.Vertical,
+                Children = new Drawable[]
+                {
+                    new DrawableNumber(0),
+                    new DrawableNumber(1001),
+                    new DrawableNumber(999999),
```
bdach: Just FYI, you could use the optional separator literals for readability here:
```csharp
new DrawableNumber(999_999),
// ...
```

frenzibyte (changes requested):
As told in your previous attempt of implementing this and in additon to what @bdach said: `Call it something that doesn't have "humanize" in the name since that is really misleading.` (by @peppy).

maybe `ToMetric` with a note saying that thousands has a suffix of `k` while other units have their full name in the suffix instead or show an example in the summary like this:
```csharp
/// <summary>
/// Turns the current or provided big number into a readable string.
///
/// HumanizerUtils.Humanize(1234) => "1.24 k"
/// HumanizerUtils.Humanize(1069485) => "1.07 million"
///
/// </summary>
/// <param name="input">The number to be humanized.</param>
/// <returns>Simplified number with a suffix.</returns>
public static string ToMetric(long input)
```

osu.Game/Utils/HumanizerUtils.cs
```diff
@@ -26,5 +26,32 @@ public static string Humanize(DateTimeOffset input)
...
+        /// <summary>
+        /// Turns the current or provided big number into a readable string
+        /// </summary>
+        /// <param name="input">The number to be humanized</param>
+        /// <returns>simplified number with a suffix</returns>
+        public static string Humanize(long input)
+        {
+            const int k = 1000;
+
+            var suffixes = new[]
```
iiSaLMaN:
```suggestion
            const string suffixes = new[]
```

osu.Game/Utils/HumanizerUtils.cs
iiSaLMaN: Since the built-in `Math` methods are broken here, a temporary fix can be used I suppose, mine would be checking if `Math.Pow(k, i)` is bigger than the giving input and fix that by looping till it's less:
```csharp
while (Math.Pow(k, i) > input) i--;
```
with a note why this line is there.

osu.Game/Utils/HumanizerUtils.cs
EVAST9919: It doesn't work like this

EVAST9919:
> As told in your previous attempt of implementing this and in additon to what @bdach said: `Call it something that doesn't have "humanize" in the name since that is really misleading.` (by @peppy).
> 
> maybe `ToMetric` with a note saying that thousands has a suffix of `k` while other units have their full name in the suffix instead or show an example in the summary like this:

Not sure about having an example in summary. Also you can read about having suffix in the output

osu.Game/Utils/HumanizerUtils.cs
peppy: at very least you want to move this outside of the method. the way you're doing it, this needs to be reallocated every call.

`static readonly suffixes = new[] ...`

osu.Game/Utils/HumanizerUtils.cs
```diff
@@ -26,5 +26,32 @@ public static string Humanize(DateTimeOffset input)
                 return input.Humanize(culture: new CultureInfo("en-US"));
             }
         }
+
+        /// <summary>
+        /// Turns the current or provided big number into a readable string.
+        /// </summary>
+        /// <param name="input">The number to be humanized.</param>
+        /// <returns>Simplified number with a suffix.</returns>
+        public static string ToMetric(long input)
```
peppy: `ToReadableString`?

peppy:
As I mentioned elsewhere, a reference needs to be here in comment form as to what you are matching (ie. why "thousand" is "k" and not matching the rest).

EVAST9919:
> As I mentioned elsewhere, a reference needs to be here in comment form as to what you are matching (ie. why "thousand" is "k" and not matching the rest).

Not sure how exactly should I write it..

smoogipoo:
I'm not really sure about this, because it makes it kinda hard to translate. I think we should rethink the relevant displays to avoid this kind of manual formatting.

EVAST9919:
> I'm not really sure about this, because it makes it kinda hard to translate. I think we should rethink the relevant displays to avoid this kind of manual formatting.

That's just how web handles it

smoogipoo:
Fwiw there is a related issue on osu-web where localisation is already broken for this: https://github.com/ppy/osu-web/issues/4192

huoyaoyuan:
Curious why not using `Humanizer.ToMetric`.

> because it makes it kinda hard to translate

As a Chinese user, I must point out that `0.85 million` is neither **more readable** nor **humanized**, in Chinese view.
Note that east-Asian languages scales in 10^4n. Although larger units are not used in morden Chinese, 10^4 and 10^8 are still used everyday. We even treat thousands-separator decreasing readability in some cases. For huger or smaller numbers, we accept SI prefixes like k/M/G/T/m/μ,  but not billion/trillion in daily uses.

I only agree to implement SI representation. InvariantCulture is not English everywhere.

smoogipoo:
`ToMetric` is what should be used, yes. My only qualm is the use of "G" to indicate billions - for us it's more natural to say "1.23 billion score" rather than "1.23 giga score".

EVAST9919:
Closing as not used/not necessary yet

PR #6260, opened 2019-09-25
Fix footer button overlap

### LINKED ISSUE #5923: Small overlap between buttons at the bottom of song select
**Describe the bug:**

There is a small (1 pixel?) overlap between the buttons at the bottom of song select.

image

**osu!lazer version:** 2019.831.0

LanTenDragon:
Fixes #5923

I have tried 1.0f and 2.0f for adjusting the spacing, it did not look good. 
So I settled on 1.5f instead.

Before 
1 0f

After
1 5f


peppy:
Thanks for the contribution!

This looks to scale incorrectly:

image

image

Which suggests there's something going on with resolution dependency here. You can reproduce the screenshotted cases by using min or max UI scale and resizing your window accordingly.

LanTenDragon:
I thought of this: Update the amount of spacing to suitable amounts when the UI changes.

Is this considered a hacky way to solve this problem?

Should I continue trying the above, or let framework fix this instead?

LanTenDragon:
Attempt 2 at fixing footer button overlap. 

Added a visual test with only dummy button to keep my sanity.

Minimum UI scale:
osu_2019-09-27_20-04-37

Maximum UI scale:
osu_2019-09-27_20-35-32

osu.Game.Tests/Visual/SongSelect/TestSceneFooterButtons.cs
```diff
@@ -0,0 +1,39 @@
...
+using osu.Game.Graphics;
+using osu.Game.Screens.Select;
+
+namespace osu.Game.Tests.Visual.SongSelect
+{
+    public class TestSceneFooterButtons : ScreenTestScene
+    {
+        public override IReadOnlyList<Type> RequiredTypes => new[]
+        {
+            //typeof(Screens.Select.SongSelect),
```
iiSaLMaN: Forgotten to-remove comment?

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -65,6 +66,9 @@ public FooterButton()
         {
             AutoSizeAxes = Axes.Both;
             Shear = SHEAR;
+            //Masks away one or two overlapping pixel
```
iiSaLMaN: We normally add a space after a comment prefix
```suggestion
            // Masks away one or two overlapping pixel
```

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -74,11 +78,23 @@ public FooterButton()
...
+                    RelativeSizeAxes = Axes.Both,
+                    //Make container for light slightly smaller than parent.
+                    Size = new Vector2(0.996f, 4),
+                    //Anchor is at top right because ModIcons are added to the right
+                    //and rounding errors (relative size od x) are hidden to the left.
+                    Anchor = Anchor.TopRight,
+                    Origin = Anchor.TopRight,
+                    Child = light = new Box
+                    {
+                        Size = new Vector2(0.998f, 4),
```
iiSaLMaN: Can inherit parent's sizings (width and height are 1 by default):
```suggestion
                        Width = 0.998f,
                        RelativeSizeAxes = Axes.Both,
```

frenzibyte (changes requested):

osu.Game/Screens/Select/FooterButton.cs
LanTenDragon: I think I'm missing something here. It does not inherit.

This is the result:
osu_2019-09-27_21-11-55

iiSaLMaN: Actually, to be more correct here, the container should set `AutoSizeAxes` to `Y` and `RelativeSizeAxes` to `X` and move the height to the light box. (so it would auto-size height of the container to the light box which would be `4`).

peppy:
image

At very high resolutions, this solution will still fail. I think you may need to re-think your approach on this. If we are do be fractionally decreasing the size, it likely needs to be based on the current display resolution.

LanTenDragon:
Closing for now, as I don't see myself having the time to work on this (will be back in one or two months)

Plus, to really solve it, I need to bind to the screen resolution and also UI scale, which will be another long code read for me.

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -56,7 +59,8 @@ public Color4 SelectedColour
             }
         }
 
-        protected readonly Container TextContainer;
+        protected readonly Container TextContainer, LightContainer, HeightAndAxisContainer;
+        protected readonly ScalingDrawSizePreservingFillContainer PreservingLightContainer;
```
peppy: unnecessary variable

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -56,7 +59,8 @@ public Color4 SelectedColour
             }
         }
 
-        protected readonly Container TextContainer;
+        protected readonly Container TextContainer, LightContainer, HeightAndAxisContainer;
```
peppy: `HeightAndAxisContainer` also unnecessary variable

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -142,5 +161,34 @@ protected override bool OnKeyDown(KeyDownEvent e)
...
         }
+
+        protected class ScalingDrawSizePreservingFillContainer : DrawSizePreservingFillContainer
+        {
+            private readonly bool applyUIScale;
+            private Bindable<float> uiScale;
+
+            public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => true;
+
+            public ScalingDrawSizePreservingFillContainer(bool applyUIScale)
```
peppy: unnecessary parameter. should always be applied

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -142,5 +161,34 @@ protected override bool OnKeyDown(KeyDownEvent e)
 
             return base.OnKeyDown(e);
         }
+
+        protected class ScalingDrawSizePreservingFillContainer : DrawSizePreservingFillContainer
```
peppy: name this something more local, and make it private. it should only be used here.

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -142,5 +161,34 @@ protected override bool OnKeyDown(KeyDownEvent e)
...
+                if (applyUIScale)
+                {
+                    uiScale = osuConfig.GetBindable<float>(OsuSetting.UIScale);
+                    uiScale.BindValueChanged(scaleChanged, true);
+                }
+            }
+
+            private void scaleChanged(ValueChangedEvent<float> args)
+            {
+                this.ScaleTo(new Vector2(args.NewValue), 500, Easing.Out);
```
peppy: probably needs a local comment explaining why this is done this way / required. since i'm not sure how this is working by reading your code.

peppy (changes requested):

osu.Game/Screens/Select/FooterButton.cs
```diff
@@ -56,7 +56,7 @@ public Color4 SelectedColour
             }
         }
 
-        protected readonly Container TextContainer;
+        protected readonly Container TextContainer, LightContainer;
```
peppy: this change can be reverted.

osu.Game/Screens/Select/Footer.cs
```diff
@@ -60,10 +65,24 @@ private void showOverlay(OverlayContainer overlay)
 
         private void updateModeLight() => modeLight.FadeColour(buttons.FirstOrDefault(b => b.IsHovered)?.SelectedColour ?? Color4.Transparent, TRANSITION_LENGTH, Easing.OutQuint);
 
+        // Extra  pacing between FooterButtons is added at all UI scales.
+        // The higher the UI scale, the less spacing added.
+        // Maximum extra spacing  is 0.625f at 1.6 UI.
+        // Minimum extra spacing  is 0.320f at 0.8 UI.
+        private void updateSpacing(ValueChangedEvent<float> args) => buttons.Spacing = new Vector2(-FooterButton.SHEAR_WIDTH + (1f / args.NewValue) * 0.5f, 0);
```
peppy: this still needs proper investigation. i'd hope to avoid ui spacing specific adjustments if possible.

osu.Game/Screens/Select/Footer.cs
```diff
@@ -60,10 +65,24 @@ private void showOverlay(OverlayContainer overlay)
 
         private void updateModeLight() => modeLight.FadeColour(buttons.FirstOrDefault(b => b.IsHovered)?.SelectedColour ?? Color4.Transparent, TRANSITION_LENGTH, Easing.OutQuint);
 
+        // Extra  pacing between FooterButtons is added at all UI scales.
```
peppy:
```suggestion
        // Extra spacing between FooterButtons is added at all UI scales.
```

peppy (changes requested):

osu.Game/Screens/Select/Footer.cs
LanTenDragon: The UI scaling based approach was meant to counteract this remaining overlap. 

Screenshot_20200222_214246

Without the UI scaling elements, the overlap is still visible but not as obvious as before. You probably would miss it if I didnt mention that the overlap is still there. Still, it is kind of a half-hearted approach. If this is acceptable, then I would push the remaning commits and fix the conflicts. Else, I would be re-closing this PR.

Theoretically, the problems should already be solved when ``Masking=true`` is set, as it should mask the pixels outside of respective containers already. I tried investigating, but still can't determine why the pixels are "leaking out".

bdach: Does the overlap depicted above grow & shrink with size with UI scaling changes? It seems to be a pixel-ish wide in size from the attached screenshot, which makes me think it is due to blending & antialiasing at the drawable edges causing a single pixel of overlap. If it does grow then it probably is something else.

LanTenDragon: It looks like it stays at the same size, but it gets more obvious? (brighter?) when UI scale is smaller.

All images below taken at 1080p

UI Scale at 1.6
image

UI Scale at 1.0 
image

UI Scale at 0.8
image

peppy:

The PR specifically is changing a pixel level error which potentially won't exist in an upcoming design update. On top of that, it's doing things in an unmaintainable way (should not be consuming UI scaled from settings).

PR #6270, opened 2019-09-25
Replace menu button text with "press for menu" on 0ms activation delay

### LINKED ISSUE #6262: "Hold for menu" is incorrect for people who set "Hold-to-confirm activation time" to 0 ms
**Describe the bug:** 
Pressing key and holding key is a different thing so the label with text "Hold for menu"  should be changed to "Press for menu" for people who set their "Hold-to-confirm activation time" to 0 ms.

osu_2019-09-25_18-58-20
osu_2019-09-25_18-58-53

**osu!lazer version:** 
2019.919.0

frenzibyte:
Closes #6262

### DIFF
--- a/osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
+++ b/osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
@@ -13,6 +13,7 @@ using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Events;
 using osu.Framework.MathUtils;
+using osu.Game.Configuration;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.Sprites;
@@ -45,7 +46,6 @@ namespace osu.Game.Screens.Play.HUD
             {
                 text = new OsuSpriteText
                 {
-                    Text = "hold for menu",
                     Font = OsuFont.GetFont(weight: FontWeight.Bold),
                     Anchor = Anchor.CentreLeft,
                     Origin = Anchor.CentreLeft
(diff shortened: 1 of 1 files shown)


peppy:
This won't update if the setting is changed post-load.

osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
```diff
@@ -60,6 +60,18 @@ public HoldForMenuButton()
             AutoSizeAxes = Axes.Both;
         }
 
+        [BackgroundDependencyLoader]
+        private void load(OsuConfigManager config)
+        {
+            var activationDelay = config.GetBindable<int>(OsuSetting.UIHoldActivationDelay).GetBoundCopy();
```
smoogipoo: This is going to be GC'd and references are going to be lost.

smoogipoo (changes requested):

osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
peppy: in two ways. spot them both (hint: check *any* other usage of config manager)

osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
```diff
@@ -60,6 +60,18 @@ public HoldForMenuButton()
...
+            activationDelay.BindValueChanged(v =>
```
smoogipoo: Generally, we should bind value changed events in `load()`. Move to `LoadComplete()

iiSaLMaN: Adding why would be much helpful to avoid that.

peppy: load is potentially in an async context. if the bound bindable changes during load, it may trigger a `ValueChanged` which isn't prepared for this to happen (bindables are generally considered update-thread-only).

You can fix this by adding a `Schedule` in the value changed event, but as a rule of thumb we avoid this practice to keep things simple.

smoogipoo (approved):

PR #6274, opened 2019-09-26
Implement rankings API models

EVAST9919:
Still without charts due to web reasons


bdach:
What kind of web reasons are we talking about?

EVAST9919:
> What kind of web reasons are we talking about?

https://github.com/ppy/osu-web/issues/4898

osu.Game.Tests/Visual/Online/TestSceneRankingsApi.cs
```diff
@@ -0,0 +1,138 @@
...
+using osu.Game.Rulesets.Taiko;
+using osu.Game.Rulesets.Catch;
+using osu.Game.Graphics.UserInterface;
+
+namespace osu.Game.Tests.Visual.Online
+{
+    public class TestSceneRankingsApi : OsuTestScene
+    {
+        [Resolved]
+        private IAPIProvider api { get; set; }
```
peppy: Should be using `API` instead, with `UseOnlineAPI => true`

peppy (changes requested):

osu.Game/Online/API/Requests/GetRankingsRequest.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            this.ruleset = ruleset;
+            this.page = page;
+        }
+
+        protected override WebRequest CreateWebRequest()
+        {
+            var req = base.CreateWebRequest();
+
+            req.AddParameter("page", page.ToString());
```
smoogipoo: I think this is fine for now as this isn't the final API.

EVAST9919:
Closing because it's already included in https://github.com/ppy/osu/pull/6293 and there's no a good way to test it without ui (which is already included in https://github.com/ppy/osu/pull/6293)

PR #6275, opened 2019-09-26
Add bindable versions of IHasComboInformation properties

smoogipoo:
Split out from https://github.com/smoogipoo/osu/tree/editor-grids

In the process this also makes the circle number pieces + accent colour use bindable indexes.


peppy:
Is the eventual plan to remove the non-bindable properties?

smoogipoo:
Maybe, but we decided on the current structure along with all other properties just for simplicity.

peppy (approved):

PR #6288, opened 2019-09-27
Fix cursor trail appearing regardless of the setting in replays

### LINKED ISSUE #6269: you can still see the cursor trail when its disabled and you watch autoplay/replays
https://streamable.com/10uip

**osu!lazer version:**
2019.925.0

frenzibyte:
Closes #6269

### DIFF
--- a/osu.Game.Rulesets.Osu/UI/Cursor/OsuCursorContainer.cs
+++ b/osu.Game.Rulesets.Osu/UI/Cursor/OsuCursorContainer.cs
@@ -40,6 +40,11 @@ namespace osu.Game.Rulesets.Osu.UI.Cursor
         private void load(OsuRulesetConfigManager config)
         {
             config?.BindWith(OsuRulesetSetting.ShowCursorTrail, showTrail);
+        }
+
+        protected override void LoadComplete()
+        {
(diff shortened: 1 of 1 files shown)


peppy:
Did you figure why this only affected autoplay/replays?

iiSaLMaN:
> Did you figure why this only affected autoplay/replays?

Not yet (I'm curious too), but I tried applying this change and it really fixed that issue.

peppy:
Please debug and figure out the "why", else it's very hard to merge this one (it may have moved the issue elsewhere). Even more so as there's no tests covering this.

iiSaLMaN:
> Please debug and figure out the "why"

After a deep investigation in the cursor trail, I've realized the alpha transform is not being applied at all, hanging in the `Transforms` list, and that's due to the current time of the application is somehow less than the start time of the transform, that's likely due to a clock change, so the issue is a matter of racing I suppose, though the weird part is why does this happen in replays and `Autoplay` only...

peppy:
It's important to understand how clocks work, and that they are not transferred or in their final state at the point of BDL. Transforms should never be applied in BDL.

peppy (approved):

PR #6289, opened 2019-09-27
Hide unnecessary profile sections for bots

### LINKED ISSUE #6147: Profiles like BanchoBot still need to become similar to the web version
**Describe the bug:** (You're probably aware of this, but I would like it to report to see if anyone already wants to implement it)
If Lazer "pages" have to be similar to web, there are a few problems with profiles like [**BanchoBot**](https://osu.ppy.sh/users/3).

* The _Saint Helena_ flag is missing on Lazer:
web
Lazer

* And on the web-side, the page only has the _me!_ tab (still missing on Lazer, but I guess it's a future thing), but on Lazer everything is visible (like _total plays_ on the left and all other tabs like _Recents_, _Ranks_, _Kudosu!_ etc.)
web
Lazer

**osu!lazer version:** 2019.913.0

EVAST9919:
Fixes https://github.com/ppy/osu/issues/6147

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneUserProfileOverlay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneUserProfileOverlay.cs
@@ -106,6 +106,14 @@ namespace osu.Game.Tests.Visual.Online
                 Country = new Country { FullName = @"Japan", FlagName = @"JP" },
                 CoverUrl = @"https://osu.ppy.sh/images/headers/profile-covers/c6.jpg"
             }, api.IsLoggedIn));
+            AddStep("Show bancho", () => profile.ShowUser(new User
+            {
+                Username = @"BanchoBot",
+                Id = 3,
+                IsBot = true,
+                Country = new Country { FullName = @"Saint Helena", FlagName = @"SH" },
+                CoverUrl = @"https://osu.ppy.sh/images/headers/profile-covers/c4.jpg"
+            }, api.IsLoggedIn));
 
             AddStep("Hide", profile.Hide);
             AddStep("Show without reload", profile.Show);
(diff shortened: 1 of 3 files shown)


peppy:
How are you testing this?

iiSaLMaN:
Also when adding test, please make it useful for CI, by having a way to ensure the sections are not added to the container when the loaded user is a bot.

osu.Game.Tests/Visual/Online/TestSceneUserProfileOverlay.cs
```diff
@@ -106,6 +106,14 @@ protected override void LoadComplete()
                 Country = new Country { FullName = @"Japan", FlagName = @"JP" },
                 CoverUrl = @"https://osu.ppy.sh/images/headers/profile-covers/c6.jpg"
             }, api.IsLoggedIn));
+            AddStep("Show bancho", () => profile.ShowUser(new User
```
bdach: CI's not complaining, but for consistency with the other steps and to separate them out visually I'd add a blank line before this.

bdach (commented):
Might have been nice to add an `Assert` step to check that the irrelevant profile sections aren't loaded after `ShowUser` with `IsBot = true`, but that would require to make them at least `protected`, so I'm unsure if it's worth the hassle of breaking encapsulation to test it.

iiSaLMaN:
> But that would require to make them at least `protected`, so I'm unsure if it's worth the hassle of breaking encapsulation to test it.

It's worth it and would make the tests useful for CI, we set the accessibility of the field to `protected` and expose it with a subclass of it `TestMyComponent` if it cannot be accessed from anywhere else, but for this one, the sections can be exposed like this `overlay.OfType<ProfileSectionsContainer>().First().OfType<ProfileSections>()`.

smoogipoo (approved):

PR #6291, opened 2019-09-27
Rework osu! selection blueprint updates

smoogipoo:
During grid implementation I've found that using bindables is very tricky since the hitobjects have `ApplyDefaults()` called on them when they're moved in the timeline. This is a simple case but the same will happen with slider control point adjustments/repetition adds/spinner length changes, etc.

So I've moved things to basically re-update inside `Update()`. This is a little bit of a proposal to see if the structure is adequate.

This doesn't update mania.


smoogipoo:
Blocking for a bit, pending tests.

PR #6293, opened 2019-09-27
Rankings tables implementation

EVAST9919:


EVAST9919:
I guess, I'll just remove https://github.com/ppy/osu/pull/6258 dependency for now since it has too many problems

smoogipoo:
image

Probably want some minimum with for the name column.

osu.Game/Overlays/Rankings/Tables/ScoresTable.cs
```diff
@@ -0,0 +1,102 @@
...
+    public class ScoresTable : RankingsTable<APIUserRankings>
+    {
+        public ScoresTable(int page = 1)
+            : base(page)
+        {
+        }
+
+        protected override TableColumn[] CreateHeaders() => new[]
+        {
+            new TableColumn("", Anchor.Centre, new Dimension(GridSizeMode.Absolute, 50)), // place
```
peppy: use `string.Empty`

osu.Game/Overlays/Rankings/Tables/PerformanceTable.cs
```diff
@@ -0,0 +1,95 @@
...
+    public class PerformanceTable : RankingsTable<APIUserRankings>
+    {
+        public PerformanceTable(int page = 1)
+            : base(page)
+        {
+        }
+
+        protected override TableColumn[] CreateHeaders() => new[]
+        {
+            new TableColumn("", Anchor.Centre, new Dimension(GridSizeMode.Absolute, 50)), // place
```
peppy: if you are repeating the same code in three files, you are doing it wrong. the shared portion should be shared.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -0,0 +1,108 @@
...
+                Content = null;
+                backgroundFlow.Clear();
+
+                if (value?.Any() != true)
+                    return;
+
+                value.ForEach(_ => backgroundFlow.Add(new TableRowBackground()));
+
+                Columns = CreateHeaders();
+                Content = value.Select((s, i) => CreateContent(page * items_per_page - (items_per_page - i), s)).ToArray().ToRectangular();
```
peppy:
```suggestion
                Content = value.Select((s, i) => CreateContent((page - 1) * items_per_page + i, s)).ToArray().ToRectangular();
```

i've never seen such convoluted number logic

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -0,0 +1,108 @@
...
+    {
+        protected const int TEXT_SIZE = 14;
+        private const float horizontal_inset = 20;
+        private const float row_height = 25;
+        private const int items_per_page = 50;
+
+        private readonly int page;
+        private readonly FillFlowContainer backgroundFlow;
+
+        public IReadOnlyList<TModel> Rankings
```
peppy: below constructor

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -0,0 +1,108 @@
...
+                    return;
+
+                value.ForEach(_ => backgroundFlow.Add(new TableRowBackground()));
+
+                Columns = CreateHeaders();
+                Content = value.Select((s, i) => CreateContent(page * items_per_page - (items_per_page - i), s)).ToArray().ToRectangular();
+            }
+        }
+
+        protected RankingsTable(int page)
```
peppy: do we ever start on a page that isn't the first page?

osu.Game.Tests/Visual/Online/TestSceneRankingsTables.cs
```diff
@@ -0,0 +1,148 @@
...
+            AddStep("Mania scores", () => createScoreTable(new ManiaRuleset().RulesetInfo));
+            AddStep("Taiko country scores", () => createCountryTable(new TaikoRuleset().RulesetInfo));
+            AddStep("Catch US performance page 10", () => createPerformanceTable(new CatchRuleset().RulesetInfo, "US", 10));
+        }
+
+        private void createCountryTable(RulesetInfo ruleset, int page = 1)
+        {
+            loading.Show();
+
+            request?.Cancel();
```
peppy: at my previous job we used to call this style of coding "copy-waste". you need to learn to self-review and not do this. if you see the same code written more than once it's an error.

osu.Game/Online/API/Requests/Responses/APICountryRankings.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Newtonsoft.Json;
+using osu.Game.Users;
+
+namespace osu.Game.Online.API.Requests.Responses
+{
+    public class APICountryRankings : CountryStatistics
+    {
+        [JsonProperty]
+        public Country Country;
```
peppy: just add this to `CountryStatistics`. the class name here is hard to understand.

osu.Game/Online/API/Requests/Responses/APIUserRankings.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using Newtonsoft.Json;
+using osu.Game.Users;
+
+namespace osu.Game.Online.API.Requests.Responses
+{
+    public class APIUserRankings : UserStatistics
+    {
+        [JsonProperty]
+        public User User;
```
peppy: just add this to `UserStatistics`. the class name here is hard to understand.

osu.Game/Overlays/Rankings/Tables/PerformanceTable.cs
```diff
@@ -0,0 +1,95 @@
...
+                    Font = OsuFont.GetFont(size: TEXT_SIZE, weight: FontWeight.Bold)
+                },
+            };
+
+            var username = new LinkFlowContainer(t => t.Font = OsuFont.GetFont(size: TEXT_SIZE)) { AutoSizeAxes = Axes.Both };
+            username.AddUserLink(item.User);
+
+            content.AddRange(new Drawable[]
+            {
+                new FillFlowContainer
```
peppy: repeated implementation *3

peppy (changes requested):

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
peppy: Can you elaborate on this one? Still unanswered.

EVAST9919: We need that variable only to calculate indexes on the left side of the table.

peppy: yes, but is it required in the ctor?

EVAST9919: Umm, for me yes. Do you have any suggestions about this may be?

peppy: So you will be reconstructing the table each page? Why is the `Rankings` property even a thing then? If you're only going to be populating it once ever, why not also put that in the ctor?

peppy:
Remember to [do drawable initialisation in the BDL method](https://github.com/ppy/osu/pull/6293/commits/e678fe7a773336e5ee2b8b7ed2bb0579101b9748). Moving stuff to the constructor means that `LoadComponentAsync` becomes pointless.

peppy (approved):

PR #6299, opened 2019-09-28
Fix track-adjusting mods not applying after resuming from another screen

### LINKED ISSUE #6298: Wind up and down effects not always playing in song select
**Describe the bug:** music sometimes isn't speeded up or down when wind up or wind down mods are turned on.
steps to reproduce:
1. chose wind up/wind down
2. enter beatmap
3. let it start and go for a few seconds
4. go out of gameplay
5. switch to other difficulty/ beatmap to hear the diffrence

https://streamable.com/ck79o

**osu!lazer version:** 2019.925.0

**Logs:** 
[database.log](https://github.com/ppy/osu/files/3664290/database.log)
[network.log](https://github.com/ppy/osu/files/3664291/network.log)
[performance.log](https://github.com/ppy/osu/files/3664292/performance.log)
[runtime.log](https://github.com/ppy/osu/files/3664293/runtime.log)

frenzibyte:
- Fixes track-adjusting mods not applying after resuming from another screen (`PlayerLoader` / `Player`) by reapplying adjustments `OnResuming()`

This issue happens because when starting a beatmap, the track is bind to the gameplay clock which means resetting speed adjustments of the gameplay clock applies the same for the track which results in the track being played with normal adjustments on resuming.

Closes #6298

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -57,7 +57,7 @@ namespace osu.Game.Overlays
         protected override void LoadComplete()
         {
             beatmap.BindValueChanged(beatmapChanged, true);
-            mods.BindValueChanged(_ => updateAudioAdjustments(), true);
+            mods.BindValueChanged(_ => ResetTrackAdjustments(), true);
             base.LoadComplete();
         }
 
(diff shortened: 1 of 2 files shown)


peppy:
Seems fair for now. We may want to re-think this in the future.

peppy (approved):

iiSaLMaN:
> Seems fair for now. We may want to re-think this in the future.

Same goes for the track loop I believe.

PR #6306, opened 2019-09-28
Fix player loader incorrectly checking for push

### LINKED ISSUE #5609: Player loader may not give enough time to change settings
**Describe the bug:** beatmap loads that fast that you don't even have a chance to edit settings

[2019-08-05 18-01-11.flv.zip](https://github.com/ppy/osu/files/3468449/2019-08-05.18-01-11.flv.zip)
delete .zip from filename

**osu!lazer version:**  2019.730.0

**Logs:** 
[database.log](https://github.com/ppy/osu/files/3468453/database.log)
[network.log](https://github.com/ppy/osu/files/3468455/network.log)
[performance.log](https://github.com/ppy/osu/files/3468456/performance.log)
[runtime.log](https://github.com/ppy/osu/files/3468457/runtime.log)

frenzibyte:
- Fixes player loader not blocking loading when a focused overlay is opened.

The underlying issue is that the player loader push conditions are a bit wrong, current master checks whether it is hovered (false when opening focused overlay) **or** the user is idle (doesn't track input for a focused overlay), this PR changes it to track whether it is hovered **and** the user is not idle. 

- [x] Depends on https://github.com/ppy/osu-framework/pull/2940

Closes #5609


peppy:
Did you consider just changing the usage parttern here to have one IdleTracker at a game level which is DI'd in? This may work better.

iiSaLMaN:
> Did you consider just changing the usage parttern here to have one IdleTracker at a game level which is DI'd in? This may work better.

But idle tracker is meant to track whether the user is currently idle or not (not only on the hierarchy) , but if you're insisting to, I can implement it like that.

peppy:
This is a memory leak, and I still disagree with this structure. Having a component that adds things at a game level is probably not something we want.

iiSaLMaN:
@peppy My other proposal for this is having a idle tracker manager component that you can register inside idle trackers. Do you have a better idea?

smoogipoo:
How about always caching idle trackers (add `[Cached]` to the class), and have the lower level idle trackers resolve in the higher level ones and compute the idle state via something like  
`IsIdle.Value = /* local idle state */ && /* parent idle state */`

peppy:
I was thinking of having the game-wide one cached and checking both in the case of `PlayerLoader`. But automating this via hierarchy as suggested by @smoogipoo may work better.

iiSaLMaN:
@smoogipoo I can't understand your proposal, can you explain a bit more?

smoogipoo:
```yaml
- Container
    - IdleTracker1
    - Container
        - IdleTracker2
```

`IdleTracker2` resolves in `IdleTracker1`.

`IdleTracker2` computes the idle state based on its own properties and the relevant properties of `IdleTracker1`.

`IdleTracker1` also itself resolves in a parenting tracker if it exists, and does the same thing, so it all transfers down.

Also that means attaching `[Cached]` to the class itself isn't going to work - it'll have to be attached to the field, but that's ok I think.

peppy:
Check something like `SkinProvidingContainer` as an example. 

- Cache self
- Resolve parent and consume parent (combine result with self)

in the case of idle checking, you probably want to get a parent bindable (if exists) and add it as an OR operation with self to check if idle at any level.

iiSaLMaN:
Hmm, if I'm checking `IsIdle` of the parenting tracker, that means the user will wait till the parent tracker "time-to-idle" length, in this case, it would have to wait 6 seconds before is idle? 

As for my proposal, it would be by creating an idle tracker manager component that you can register inside idle trackers, and the manager will simply add a receptor and calculate idle state from it for every idle tracker by doing: `tracker.IsIdle.Value = receptor.TimeSpentIdle > tracker.TimeToIdle && tracker.AllowIdle`

peppy:
Out of curiosity, have you tried checking the hover state? I think when overlays are present you will likely lose hover, which could be an alternative to all of this.

iiSaLMaN:
> Out of curiosity, have you tried checking the hover state? I think when overlays are present you will likely lose hover, which could be an alternative to all of this.

What if the user idles? unless the tracker wasn't for that case?

peppy:
I'm not sure player loading should continue as long as an overlay is present... would need to see how it feels.

peppy:
I didn't mean removing idle tracking completely. It's there for a reason – if you leave the cursor idle on top of a UI pane it will never load, leaving the user in confusion.

peppy:
I believe test failures are related to the changes.

peppy:
What is going on here? Are the tests passing for you locally?

peppy:
@iiSaLMaN can you please check this and close it if you're not planning on fixing?

iiSaLMaN:
Oh sorry, will try to fix this.

iiSaLMaN:
Turns out it was actually the notifications overlay being somehow open for a headless game host, and hover not being updated if the mouse hasn't moved. (a framework-side issue should likely be opened for that)

smoogipoo:
This needs some more testing. Gameplay should still start if you move your mouse around on the loader without hovering any overlay.

smoogipoo:
Furthermore, hovering over the loader overlays should not block continuation to gameplay indefinitely.

osu.Game.Tests/Visual/Menus/TestSceneScreenNavigation.cs
```diff
@@ -112,6 +112,12 @@ public void TestSongContinuesAfterExitPlayer(bool withUserPause)
 
             AddStep("press enter", () => pressAndRelease(Key.Enter));
 
+            // Player loading may be blocked by either the cursor staying at 0,0 or a focused overlay being opened taking away hover from player loader
```
peppy: the 0,0 cursor thing sounds like a bug or regression?

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -68,8 +69,22 @@ public void TestBlockLoadViaMouseMovement()
         {
             AddStep("load dummy beatmap", () => ResetPlayer(false));
             AddUntilStep("wait for current", () => loader.IsCurrentScreen());
-            AddRepeatStep("move mouse", () => InputManager.MoveMouseTo(loader.VisualSettings.ScreenSpaceDrawQuad.TopLeft + (loader.VisualSettings.ScreenSpaceDrawQuad.BottomRight - loader.VisualSettings.ScreenSpaceDrawQuad.TopLeft) * RNG.NextSingle()), 20);
+            AddRepeatStep("move mouse on settings", () => InputManager.MoveMouseTo(loader.VisualSettings.ScreenSpaceDrawQuad.TopLeft + (loader.VisualSettings.ScreenSpaceDrawQuad.BottomRight - loader.VisualSettings.ScreenSpaceDrawQuad.TopLeft) * RNG.NextSingle()), 20);
             AddAssert("loader still active", () => loader.IsCurrentScreen());
+            AddRepeatStep("move mouse on loader", () => InputManager.MoveMouseTo(loader.ScreenSpaceDrawQuad.TopLeft + (new Vector2(loader.VisualSettings.ScreenSpaceDrawQuad.TopLeft.X, loader.ScreenSpaceDrawQuad.BottomLeft.Y) - loader.ScreenSpaceDrawQuad.TopLeft) * RNG.NextSingle()), 20);
```
peppy: this is testing the opposite of what we want. this should still load.

peppy (changes requested):

osu.Game.Tests/Visual/Menus/TestSceneScreenNavigation.cs
iiSaLMaN: need to rename this comment, the `0,0` bug has been already fixed by the framework-side PR but I was referring to the toolbar as well.

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
iiSaLMaN: moving mouse on loader repeatedly is what we call not idle, it should be ready to push only when the user is idle?

peppy: the idea is we *only want to block* if the user is doing something meaningful. check on master – it blocks if they are adjusting settings only. your change means if someone is playing with their mouse it will never load the beatmap.

iiSaLMaN:
Closing in favour of #7836

PR #6308, opened 2019-09-28
Migrate beatmap carousel test to AddUntilStep

bdach:
Attempts to resolve `osu.Game.Tests.Visual.SongSelect.TestSceneBeatmapCarousel.TestFiltering` visual test non-deterministically failing builds.

# Summary

Due to non-deterministic test failures in `TestSceneBeatmapCarousel`, migrate the `checkSelected` helper step from `AddAssert` to `AddUntilStep`. This adds more leniency for performance-related issues while still checking the desired behaviour.

# Remarks

I am moderately certain that the root cause is the step progression racing with the filtering logic, which this sort of change would resolve, but I'm not sure how to convincingly prove that this fix succeeds, other than by saying the following:
* The test failure sometimes reproduced locally, usually at the first couple of test runs after opening Visual Studio.
* I never got that symptom again after applying the fix, I restarted VS like 5-6 times doing that.
* I ran five AppVeyor builds with the fix on my fork, here's a summary:

| Build | Test result | Build result |
|-|:-:|:-:|
| [fix-test-races-9](https://ci.appveyor.com/project/bdach/osu/builds/27737496/tests) | pass | pass |
| [fix-test-races-10](https://ci.appveyor.com/project/bdach/osu/builds/27737621/tests) | pass | pass |
| [fix-test-races-11](https://ci.appveyor.com/project/bdach/osu/builds/27737744/tests) | pass | pass |
| [fix-test-races-12](https://ci.appveyor.com/project/bdach/osu/builds/27737861/tests) | pass | fail |
| [fix-test-races-13](https://ci.appveyor.com/project/bdach/osu/builds/27737942/tests) | pass | pass |

The one failure comes from `osu.Game.Tests.Visual.Menus.TestSceneScreenNavigation`. I've also seen that fail sometimes, along with other tests, but after looking at them I did not have any idea of a fix that I'd be willing to PR.

Sorry for introducing this in the first place, I will bear the possibility of this sort of issue in mind in the future.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -83,7 +83,7 @@ namespace osu.Game.Tests.Visual.SongSelect
             AddAssert("ensure prev random fetch worked", () => selectedSets.Peek() == carousel.SelectedBeatmapSet);
 
         private void checkSelected(int set, int? diff = null) =>
-            AddAssert($"selected is set{set}{(diff.HasValue ? $" diff{diff.Value}" : "")}", () =>
+            AddUntilStep($"selected is set{set}{(diff.HasValue ? $" diff{diff.Value}" : "")}", () =>
             {
                 if (diff != null)
                     return carousel.SelectedBeatmap == carousel.BeatmapSets.Skip(set - 1).First().Beatmaps.Skip(diff.Value - 1).First();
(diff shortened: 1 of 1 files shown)


osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -83,7 +83,7 @@ private void loadBeatmaps(List<BeatmapSetInfo> beatmapSets = null)
             AddAssert("ensure prev random fetch worked", () => selectedSets.Peek() == carousel.SelectedBeatmapSet);
 
         private void checkSelected(int set, int? diff = null) =>
```
iiSaLMaN:
```suggestion
         private void waitForSelection(int set, int? diff = null) =>
```

frenzibyte (changes requested):

peppy (approved):

PR #6310, opened 2019-09-28
Add the ability to choose a random intro sequence each startup

### LINKED ISSUE #6284: Random intro sequence
**Describe the new feature:** Seeing as the Circles intro with its song is still in the game and can be chosen if desired, how about adding a random option in the dropdown so that both can be used at the same time and for a bit of variety? ~~Hell, maybe even add more sequences in the future, or support for custom ones?~~

image

V1ntagezTV:
Simple "Random" intro

- Closes #6284.

### DIFF
--- a/osu.Game/Screens/Loader.cs
+++ b/osu.Game/Screens/Loader.cs
@@ -60,10 +60,10 @@ namespace osu.Game.Screens
 
         private IntroScreen getIntroSequence()
         {
-            Random random = new Random();
             switch (introSequence)
             {
                 case IntroSequence.Random:
+                    var random = new Random();
                     if (random.Next(2) == 0)
                         return new IntroCircles();
                     else
--- a/osu.Game/Screens/Loader.cs
+++ b/osu.Game/Screens/Loader.cs
@@ -59,7 +59,7 @@ namespace osu.Game.Screens
         }
 
         private IntroScreen getIntroSequence()
-        {//вот именно что не показывает ни всплывающих подсказок нихера
+        {
             Random random = new Random();
             switch (introSequence)
             {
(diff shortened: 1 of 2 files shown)


elleyer:
Any reason for that? Also, you should delete not English comments.

bdach:
@elleyer The submitter aims to resolve #6284 with this.

osu.Game/Screens/Loader.cs
```diff
@@ -58,9 +59,16 @@ protected virtual OsuScreen CreateLoadableScreen()
         }
 
         private IntroScreen getIntroSequence()
-        {
+        {//вот именно что не показывает ни всплывающих подсказок нихера
```
bdach: Please keep all source, including comments, in English. Additionally your whitespace needs fixing (as reported by CI).

osu.Game/Screens/Loader.cs
```diff
@@ -58,9 +59,16 @@ protected virtual OsuScreen CreateLoadableScreen()
...
+            Random random = new Random();
```
bdach: * Use `var` here.
* This allocation can be moved into the `switch` case.

bdach (changes requested):

frenzibyte (changes requested):
This can be fairly simplified by just picking a random intro sequence enum value.

```csharp
switch (introSequence)
{
    case IntroSequence.Random:
        var introSequences = Enum.GetValues(typeof(IntroSequence)).Except(new[] { IntroSequence.Random }).ToList();
        return getIntroSequence(introSequences[RNG.Next(introSequences.Count)]);
}
```

Notes to be took from the above snippet:
 - `RNG` is used instead of `Random` which is a simplified way of seedless randomizing.

peppy (approved):

smileyhead:
Nice, thanks! :-)

PR #6321, opened 2019-09-29
Do not close osu! when the back button is pressed on android.

### LINKED ISSUE #6325: Pressing Back Button kill the game not back to previous menu (Android)
**Describe the bug:** 
**Expected Behavior:**
When pressing back button will back to the previous menu
**Reality:**
Pressing back button will kill the game.
**Solution:**
Please make pressing the back button willl go to the previous menu. Thanks For the attention

https://youtu.be/mTrdI9_D_Po

**osu!lazer version:** 
2019.925.0

**Logs:** 
No Logs

miterosan:
Overiding this method prevents that the back button closes the app.

Resolves https://github.com/ppy/osu/issues/6325

### DIFF
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -21,5 +21,12 @@ namespace osu.Android
             Window.AddFlags(WindowManagerFlags.Fullscreen);
             Window.AddFlags(WindowManagerFlags.KeepScreenOn);
         }
+
+        /// <summary>
+        /// Overriding this method avoids that osu! closes when the backbutton is pressed.
+        /// </summary>
+        public override void OnBackPressed()
+        {
+        }
     }
 }


peppy:
Are you sure we don't want to do this in `AndroidGameActivity`? I think we probably do.

iiSaLMaN:
@miterosan Just for making it easier, can you modify the description to actually close the issue, change to `Resolves`, `Fixes` or `Closes`.

ojh050118:
How do I exit? And I tested, but it doesn't go back to the previous screen.

swoolcock:
The hardware back button should really be mapped to `GlobalAction.Back`.

LukaTV939:
I constantly press back for no reason on my android phone, because I think it will go back to menu but no, it doesn't. That was also discovered back when 2019.704.0 was out on android.

swoolcock:
Please PR this in the framework and ensure that the hardware back button is generating a key event we can cleanly handle in osu!.  If it fires an escape key event that would magically handle the back action for us, too.

peppy:
@miterosan you still okay to move this one to a framework implementation?

miterosan:
Sure, just didn't had the time / focus to do that. Will see what I can do
today.

Dean Herbert <notifications@github.com> schrieb am Di., 8. Okt. 2019, 10:11:

bdach:
Closing in favour of ppy/osu-framework#2877.

PR #6331, opened 2019-09-30
Add SkinInfo property to ISkin interface

frenzibyte:
- Adds `SkinInfo` property to `ISkin` interface.
- Dependency for a DHO change that requires fetching information about current skin.


iiSaLMaN:
Closing as another direction has been chosen for this.

Issue #5537, opened 2019-07-30
Circle Expand Version issue (numbers expands)


Fatal3ty:
Circle expand from lazer version issue, that one use from like version 1:
![expand lazer](https://user-images.githubusercontent.com/53025167/62157957-a1625180-b30e-11e9-98eb-db787a073d0f.png)
Osu Stable with skin.ini Version 2.0+:
![expand stable](https://user-images.githubusercontent.com/53025167/62158004-c0f97a00-b30e-11e9-8895-879d4d54914f.png)

Same Skin and that use verison: 2 or latest
Skin: Clear Skin Ultra 3.0 (https://osu.ppy.sh/community/forums/topics/606260)
Same issue with another skins.

osu!lazer version: 2019.730.0
Logs: none

peppy:
Is this an issue? I was aware of this when implementing this, but it looked better with the number expanding to me. Would appreciate more detail on why you prefer it not expand.

Fatal3ty:
Mostly people like without numbers expand. I think that one can be config on skin.ini, like example this:

Example:
CircleExpand: 0 or 1 (sould be use circle expand or not?
NumbersExpand: 0 or 1 (only worked if CircleExpand is set to 1)

b4mbus88:
Please dont take us the possibility to change the fadeing. In my opinion as a skinner it would be great if we could costumize the fadeout time in skin.ini. I always played without the animation with only default0-9.png as hitcircle and i dont want to miss it. This also would solve the problem with numbers above 10 (stacking default pngs/numbers). So the ajustment in skin.ini would solve all problems and fit for every player i guess.

And it would make alot of skins out there right now playable for me :D

iiSaLMaN:
I'm actually a bit confused here, so the issue is that `NumbersExpand` is ignored?

PR #6332, opened 2019-09-30
Add Version property to SkinInfo

frenzibyte:
- Adds `Version` property to `SkinInfo` and retrieve it from the skin configuration.
- Dependency for a DHO change that requires fetching the version of the current skin.

### DIFF
--- a/osu.Game/Migrations/20190930135517_AddSkinInfoVersion.Designer.cs
+++ b/osu.Game/Migrations/20190930135517_AddSkinInfoVersion.Designer.cs
@@ -2,15 +2,17 @@
 using System;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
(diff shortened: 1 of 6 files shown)


peppy:
What is this for? It looks like you're implementing a legacy feature in the wrong place - legacy skin versions should be handled in an isolated way (via `LegacySkin`)

iiSaLMaN:
> What is this for? It looks like you're implementing a legacy feature in the wrong place - legacy skin versions should be handled in an isolated way (via `LegacySkin`)

This is a dependency for a PR that fixes #5537. would require knowing the version of the skin, If I add the property to `LegacySkin` I can't retrieve it since the decoder passes the skin configuration which holds the info only. The only solution may be to add it as a dictionary key and retrieve it from the hit circle directly with:
```csharp
skin.GetConfig<GlobalSkinConfiguration, string>(GlobalSkinConfiguration.Version);
```

iiSaLMaN:
Closing as another direction has been chosen for this.

PR #6344, opened 2019-10-01
Fix memory leaks due to audio track recycle order

smoogipoo:
In current master, the `Tracks` counter in `GameThread.Audio` always increases when selecting a new beatmap in song select. The reason for this is the order of recycling tracks vs other areas performing operations on them. For example:

https://github.com/ppy/osu/blob/4880fd9e02c6fe129fb3396f601fb3285b7e6cc7/osu.Game/OsuGame.cs#L327-L337

https://github.com/ppy/osu/blob/4880fd9e02c6fe129fb3396f601fb3285b7e6cc7/osu.Game/Screens/Select/SongSelect.cs#L418-L421

This results in huge memory use:

image

(All those `Byte[]`s are track data on the LOH, the streams themselves are on Gen2)

Simplest way I can think of to resolve this is to schedule after children. Also took this opportunity to refactor the relevant code - it had a weird condition + used old-style last value tracking.


peppy:
Should `BindableBetamap` just be nuked now? It was only there for that one purpose, I believe.

smoogipoo:
Oh yeah I was going to do that but forgot about it.

peppy:
I guess this is okay. We might want to consider finaliser disposal in WorkingBeatmap or Track in the future.

peppy (approved):

PR #6345, opened 2019-10-01
Expand number piece for legacy old skins

### LINKED ISSUE #5537: Circle Expand Version issue (numbers expands)

frenzibyte:
- Expand number piece for legacy skin that are older than v2.0
- This matches behaviour with stable: [Reference from wiki](https://osu.ppy.sh/help/wiki/Skinning/skin.ini#1.0).
- [x] Depends on #6450 

Closes #5537


osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -123,6 +123,10 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                                 return SkinUtils.As<TValue>(new BindableFloat(legacy_circle_radius));
 
                             break;
+
+                        case OsuSkinConfiguration.ExpandNumberPiece:
+                            string legacyVersion = source.GetConfig<string, string>("Version")?.Value ?? "1";
```
iiSaLMaN: Is this accepted? since this is considered legacy version of skins then this seems a good spot to retrieve the version from,  although it's possible to create `LegacySkinConfiguration` enum but unsure if it's worth it.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs
```diff
@@ -10,16 +10,21 @@
...
 
 namespace osu.Game.Rulesets.Osu.Objects.Drawables
 {
     public class DrawableHitCircle : DrawableOsuHitObject, IDrawableHitObjectWithProxiedApproach
     {
         public ApproachCircle ApproachCircle { get; }
 
+        public IBindable<bool> ExpandNumberPiece => expandNumberPiece;
+
+        private readonly BindableBool expandNumberPiece = new BindableBool();
```
smoogipoo: Why have this in DHC if it's not being used here? MainCirclePiece/LegacyMainCirclePiece already have access to the skin.

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -23,17 +26,24 @@ public LegacyMainCirclePiece()
             Size = new Vector2(OsuHitObject.OBJECT_RADIUS * 2);
         }
 
+        private Sprite hitCircleSprite;
+        private SkinnableSpriteText hitCircleText;
+
+        private List<Drawable> scalables;
```
smoogipoo: Unnecessary. Just transform each child.

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -58,13 +68,25 @@ private void load(DrawableHitObject drawableObject, ISkinSource skin)
             };
 
             state.BindTo(drawableObject.State);
-            state.BindValueChanged(updateState, true);
-
             accentColour.BindTo(drawableObject.AccentColour);
-            accentColour.BindValueChanged(colour => hitCircleSprite.Colour = colour.NewValue, true);
-
             indexInCurrentCombo.BindTo(osuObject.IndexInCurrentComboBindable);
+            expandNumberPiece.BindTo(drawableCircle.ExpandNumberPiece);
```
smoogipoo: As per above comment, use the skin directly.

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -58,13 +68,25 @@ private void load(DrawableHitObject drawableObject, ISkinSource skin)
...
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            state.BindValueChanged(updateState, true);
+            accentColour.BindValueChanged(colour => hitCircleSprite.Colour = colour.NewValue, true);
             indexInCurrentCombo.BindValueChanged(index => hitCircleText.Text = (index.NewValue + 1).ToString(), true);
+            expandNumberPiece.BindValueChanged(expand =>
```
smoogipoo: I'm fairly certain this doesn't need to exist.

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
smoogipoo: I'm not sure we want skins to be versioned like this. Would personally prefer a `LegacySkinConfiguration`.

Beyond that, this can't be done everywhere, and I'm not sure it's even correct because `latest` should be `2.7`, and not having a `Version` should default to `1`.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
iiSaLMaN: I accounted for this carefully, this will return `1` if `Version` doesn't exist in `skin.ini` (null is returned if `Version` key isn't in the `ConfigDictionary`, but if `skin.ini` doesn't exist, this will return `latest` (`DefaultSkinConfiguration` is passed to the legacy skin if `skin.ini` doesn't exist afaik)

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs
iiSaLMaN: Right 🤦‍♂️

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
smoogipoo: Wanna double check that? Both of these cases construct a `DefaultSkinConfiguration`:

https://github.com/ppy/osu/blob/cbdf42cd7b9cf782cb0aaa7e88b92ea95f806f67/osu.Game/Skinning/LegacySkin.cs#L37-L41

Fwiw I think we should be working with numeric version numbers all the time. That is having the mapping from latest -> 2.7 and nothing -> 1 explicitly defined rather than making these cases everywhere, because it's going to come up a lot more (spinners for example).

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -23,18 +23,24 @@ public LegacyMainCirclePiece()
...
         }
 
+        private Sprite hitCircleSprite, hitCircleOverlay;
+        private SkinnableSpriteText hitCircleText;
+
         private readonly IBindable<ArmedState> state = new Bindable<ArmedState>();
         private readonly Bindable<Color4> accentColour = new Bindable<Color4>();
         private readonly IBindable<int> indexInCurrentCombo = new Bindable<int>();
 
+        [Resolved]
```
peppy: why was this moved?

peppy (changes requested):

peppy:
The version code is all over the place. `DefaultSkinConfiguration` should probably not take a string, so maybe use that as a starting point.

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
iiSaLMaN: because of value changed bindings. I think it's safer to bind value changed events in `LoadComplete`?

peppy: This isn't used by them.

iiSaLMaN: I don't follow, what isn't used by them?

peppy: The thing this comment is attached to. I'm not sure how to be more clear.

iiSaLMaN: oh, you meant the `drawableObject` being resolved with the attribute, I thought you meant why I moved the bindings, sorry, will return back.

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -123,6 +124,10 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                                 return SkinUtils.As<TValue>(new BindableFloat(legacy_circle_radius));
 
                             break;
+
+                        case OsuSkinConfiguration.ExpandNumberPiece:
+                            decimal legacyVersion = source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value ?? 1.0m;
```
peppy: why 1.0m? where does this even come from? why not just 

`bool expand = source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value >= 2.0;`

peppy (changes requested):

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
iiSaLMaN: `value` could be null (if the provided skin itself provides skin.ini without version)? 

I thought you meant comparing null with non-null?

peppy: you have changed the null handling logic. see my previous statement. the statement i gave you did not change this. confirm this was intentional

peppy: previous code:
```csharp
decimal legacyVersion = source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value ?? 1.0m;
                            return SkinUtils.As<TValue>(new BindableBool(legacyVersion < 2.0m));
```
if Version is null or less than 2.0, expand

proposed code:
```csharp
return !(source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value >= 2.0);
```
if Version is null or less than 2.0, expand

your code:
```csharp
                            bool expand = source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value < 2.0m;
                            return SkinUtils.As<TValue>(new BindableBool(expand));
```
if Version is less than 2.0, expand

iiSaLMaN: I wrote it carelessly, after reading the docs regarding nullable comparsions, I understand now why you proposed the code to check if not greater or equal to 2.0, will fix it, sorry.

Joehuu:
Just want to bump this, as this visual difference makes lazer gameplay feel different from stable, probably what this player and maybe others are noticing.
osu!_obnFx7lGdt

As for my review, the fade out of the number should also be different between the two versions. The linked issue's (https://github.com/ppy/osu/issues/5537) description gives a good visual of that. Just not sure what that duration is.

peppy:
Will bump this into the next milestone

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -136,6 +137,10 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                                 return SkinUtils.As<TValue>(new BindableFloat(LEGACY_CIRCLE_RADIUS));
 
                             break;
+
+                        case OsuSkinConfiguration.ExpandNumberPiece:
+                            bool expand = !(source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value >= 2.0m);
```
peppy, line 142: are you sure this is correct? won't it stop the propagation of skin lookup at the first skin?

peppy: in addition, this *definitely* needs an in-line comment explaining this. ie. the fact that skin version 2.0 changed behaviour.

else it actually reads like it's a bug (and there's a missing configuration lookup).

peppy (dismissed):

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
frenzibyte: > are you sure this is correct? won't it stop the propagation of skin lookup at the first skin?

I don't think there's an issue with that, the `Version` is the lookup that would be propagated until top-parent, same is done with:
https://github.com/ppy/osu/blob/ccc20a93412ead98c9e80f2cdcd4b911ff703935/osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs#L93-L104

bdach: While we're on this I find reading `!(source.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value >= 2.0m);` super backwards. Couldn't this just be a `<` so that there's no need to mentally flip this check when reading this code?

frenzibyte: @bdach because C# has a very weird thing with nullables, check out the full convo. of https://github.com/ppy/osu/pull/6345#discussion_r358596459, I'm gonna try my best to split this.

bdach: I don't recall reading that, fair enough. I guess it can stay then to not belabour this further. Maybe leave an inline comment or something explaining why this is written this weirdly.

frenzibyte:
> As for my review, the fade out of the number should also be different between the two versions. The linked issue's (#5537) description gives a good visual of that. Just not sure what that duration is.

@Joehuu Right, couldn't do per-frame comparisons at the time I opened PR (should've mentioned that in the description...), here's how it currently looks (slowed down):

([video](https://streamable.com/il2tqn))

After applying a linear 60.0ms duration fade out, it looks much similar and would give same gameplay feels to osu!stable:

([video](https://streamable.com/6svm0b))

But now that `ExpandNumberPiece` doesn't only stop scaling, but applies immediate fading to the number piece, I'll discard it all and put the version check inside the legacy piece itself (as a condition to decide between whether to scale, or immediately fade)

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -45,12 +48,27 @@ private void load(DrawableHitObject drawableObject)
 
             InternalChildren = new Drawable[]
             {
-                hitCircleSprite = new Sprite
+                circlePieces = new Container<Sprite>
```
bdach, line 51: I'd just call this `hitCircleContainer` or something probably? Named a bit weird right now.

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -98,20 +101,36 @@ Texture getTextureWithFallback(string name)
...
-
-                    if (skin.GetConfig<OsuSkinConfiguration, bool>(OsuSkinConfiguration.ExpandNumberPiece)?.Value ?? true)
+                    circlePieces.FadeOut(legacy_fade_duration, Easing.Out);
+                    circlePieces.ScaleTo(1.4f, legacy_fade_duration, Easing.Out);
+
+                    var legacyVersion = skin.GetConfig<LegacySetting, decimal>(LegacySetting.Version)?.Value;
+
+                    if (legacyVersion >= 2.0m)
+                        // legacy skins of version 2.0 and newer apply immediate fade out to the number piece and only that.
+                        hitCircleText.FadeOut(legacy_fade_duration / 4, Easing.Out);
```
bdach, line 127: I'm really confused by this comment. It states that "newer skins apply immediate fade out to the number piece and only that", which is immediately followed by a fade of the text with a non-zero (albeit shorter) duration. I'm guessing you meant "scale"?

frenzibyte: not sure what you mean by "scale", immediate fade out as in the duration is extremely short (mostly 2 frames of transforming) if that's what you're confused about?

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
frenzibyte: wasn't sure about this name, maybe `circleSprites` would do?

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
bdach: Both the terminology (not sure what a "number piece" is supposed to mean in this context) and the non-immediate immediate fade out (for me immediate is duration 0 always; else this reads like a bug - what you have there is more like a very short fade consistent with stable).

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
bdach: Well seeing as both contained drawables are prefixed with `hitCircle` that's what I proposed. I mostly suggested a change as this screwed me up reading the other comment (here a piece is the *background*, and there the piece is the *text*...?)

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
frenzibyte: > Both the terminology (not sure what a "number piece" is supposed to mean in this context)

I'm not sure if there's a problem with saying "number piece" in a `LegacyMainCirclePiece` component referring to the hitcircle number sprite text, it's quite clear to me.

> and the non-immediate immediate fade out (for me immediate is duration 0 always; else this reads like a bug - what you have there is more like a very short fade consistent with stable).

Ah, if there's one thing I'm missing in my contributions, it's proper English; will go with `very short fade out`.

smoogipoo (approved):
This looks fine to me. One thing to note is that stable does all the fade transforms here with `Easings.None`.

PR #6351, opened 2019-10-01
Simplify exit logic of screens with textboxes using back button receptor

Joehuu:
Initially I just wanted to do the second commit, but it wasn't working with every screen as `Exit()` was happening on textboxes.

Since back button receptor is now the last `GlobalAction` when pressing escape, it won't interfere with the `GlobalAction`s of textboxes, etc.

`OnPressed` of `OsuTextBox` is already done [framework level](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/UserInterface/TextBox.cs#L708) using `OnKeyDown`. Using breakpoints, it was never executed.


smoogipoo (approved):
👍

PR #6353, opened 2019-10-01
Add key binding to restart current track

### LINKED ISSUE #6335: Play from beginning hotkey is missing
**Describe the new feature:** A hotkey that will seek to the start of the current song and starts playing it (It's F2 or X key in stable)

GSculerlor:
Fix #6335 

**Proof of concept:** https://streamable.com/u208i

**Description:** Matching up the behaviour with stable. When track is running it'll restart the current track, when track is paused it'll resume the current track. For now it showing "Restart Track" and "Play Track" (stable only showing "Play")

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -46,8 +46,6 @@ namespace osu.Game.Overlays
         [Resolved(canBeNull: true)]
         private OnScreenDisplay onScreenDisplay { get; set; }
 
-        private bool isRestart;
-
         [BackgroundDependencyLoader]
         private void load()
         {
@@ -183,12 +181,14 @@ namespace osu.Game.Overlays
             return false;
         }
 
-        private bool restartTrack()
+        private bool restartTrack(out bool isRestart)
         {
             var track = current?.Track;
 
             if (track == null)
             {
+                isRestart = false;
+
                 if (beatmap.Disabled)
                     return false;
 
@@ -297,7 +297,7 @@ namespace osu.Game.Overlays
                     return true;
 
                 case GlobalAction.MusicRestart:
-                    if (restartTrack())
+                    if (restartTrack(out bool isRestart))
                         onScreenDisplay?.Display(new MusicControllerToast(isRestart ? "Restart track" : "Play track"));
 
                     return true;
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -46,6 +46,8 @@ namespace osu.Game.Overlays
         [Resolved(canBeNull: true)]
         private OnScreenDisplay onScreenDisplay { get; set; }
 
+        private bool isRestart;
+
         [BackgroundDependencyLoader]
         private void load()
         {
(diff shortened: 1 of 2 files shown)


frenzibyte (changes requested):
Missing tests. (with asserts ensuring correct behaviour)

osu.Game/Overlays/MusicController.cs
```diff
@@ -181,6 +183,35 @@ private bool next(bool instant = false)
             return false;
         }
 
+        private bool restartTrack()
```
iiSaLMaN: You can use an out parameter instead of a private field that is seemingly misleading
```suggestion
        private bool restartTrack(out bool restarted)
```

frenzibyte (changes requested):

GSculerlor:
> Missing tests. (with asserts ensuring correct behaviour)

Not sure how to test it but will give a try

osu.Game/Overlays/MusicController.cs
peppy: what.

book return and bool our with no documentation this is shocking.

osu.Game/Overlays/MusicController.cs
```diff
@@ -181,6 +181,37 @@ private bool next(bool instant = false)
             return false;
         }
 
+        private bool restartTrack(out bool isRestart)
```
iiSaLMaN: The name of this function is kinda misleading, I'm not good with namings but maybe rename to `startOrRestartTrack`?

osu.Game/Overlays/MusicController.cs
iiSaLMaN: Forgot to mention that

osu.Game/Overlays/MusicController.cs
bdach: If there's an `Or` in a method name, it's a surefire indicator that the method is doing too much.

bdach (commented):
Personally I don't like that `isRestart` exists. I'd advocate for moving the on-screen display calls somewhere inside to avoid all this execution path branching.

osu.Game/Overlays/MusicController.cs
iiSaLMaN: Yeah again, I'm not good with namings.

GSculerlor:
> Personally I don't like that `isRestart` exists. I'd advocate for moving the on-screen display calls somewhere inside to avoid all this execution path branching.

Yeah I'm concern about execution branching also. But not sure should I call it inside the function

iiSaLMaN:
What about checking instead (like `GlobalAction.MusicPlay`): `IsPlaying ? "Restart track" : "Play track"

bdach:
> What about checking instead (like `GlobalAction.MusicPlay`): `IsPlaying ? "Restart track" : "Play track"`

Doesn't help here. `MusicPlay` switches on the track state *after* performing the action (it can do that - it's a toggle), in this case the switch happens on the track state *before* - pressing this new binding will make *some* track play if `true` is returned, the question is if it was in progress (in which case the text should say `Restart`) or if it was paused (in which case it should say `Play`)

iiSaLMaN:
just check if it *was* playing?
```csharp
var wasPlaying = IsPlaying;
if (restartTrack())
    ...(wasPlaying ? "Restart track" : "Play track")

smoogipoo (approved):
Seems ok.

peppy:
Blocking pending UX discussion in issue.

GSculerlor:
Closing it for now, will create another PR with intended UX behavior

PR #6359, opened 2019-10-02
Change HoldForMenu slider precision to 50

### LINKED ISSUE #6341: Hold-to-confirm should have a sane minimal value
Setting hold-to-confirm to 1ms is functionally identical to 0ms but removes all the confirmations 0ms has. A sane minimal value like 50ms (1/3 the average reaction time of a human) in addition to 0ms should be used.

LanTenDragon:
Fixes #6341

### DIFF
--- a/osu.Game/Configuration/OsuConfigManager.cs
+++ b/osu.Game/Configuration/OsuConfigManager.cs
@@ -114,7 +114,7 @@ namespace osu.Game.Configuration
 
             Set(OsuSetting.UIScale, 1f, 0.8f, 1.6f, 0.01f);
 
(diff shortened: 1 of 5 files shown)


smoogipoo (approved):

iiSaLMaN:
This should've been a `double` instead of `float`, will change in another PR.

smoogipoo:
what? why?

iiSaLMaN:
Because it's used in time calculations, which has a `double` just a code quality thing, nothing more.

bdach:
I'm not sure that is worth the time. The values of that slider are whole numbers that can be accurately represented without rounding errors using floating point variables, no matter if single or double precision. Conversions will be lossless.

PR #6360, opened 2019-10-02
Add bottom bar hit error meter

### LINKED ISSUE #5932: Centered hit errors (like osu!stable)

Game4all:
Closes #5932
Second attempt  using the approach proposed in #6025 

image


frenzibyte (changes requested):
Can't this be calculated by a private function that spits the correct vector / anchor / axes? so you would give it the default positioned value of the type and the alignment and it would invert it or something, what do you think?

`getAligned(alignment, new Vector2(x, y))`:
```csharp
private Vector2 getAligned(Anchor alignment, Vector2 value) => alignment == Anchor.x1 ? value.Yx : value;
```

osu.Game/Screens/Play/HUD/HitErrorMeters/BarHitErrorMeter.cs
```diff
@@ -233,29 +282,54 @@ public override void OnNewJudgement(JudgementResult judgement)
...
+                    RelativeSizeAxes = Axes.Y;
+                    RelativePositionAxes = Axes.X;
+                    Width = 3;
+                }
+                else
+                {
+                    RelativeSizeAxes = Axes.X;
+                    RelativePositionAxes = Axes.Y;
+                    Height = 3;
+                }
```
iiSaLMaN: These can be fairly simplified with:
```csharp
RelativeSizeAxes = isVertical ? Axes.Y : Axes.X;
RelativePositionAxes = isVertical ? Axes.X : Axes.Y;
Size = isVertical ? new Vector2(3, Size.Y) : new Vector2(Size.X, 3);
```

(making it more easier with what I proposal):
```csharp
RelativeSizeAxes = getAligned(alignment, Axes.X);
RelativePositionAxes = getAligned(alignment, Axes.Y);
Size = getAligned(alignment, new Vector2(Size.X, 3));
```

Game4all: I'm unsure about the utility of the code snippet you commented earlier as it is simply wrapping an inline conditionnal in a function. As for the properties , looks like I forgot to inline them

iiSaLMaN: > I'm unsure about the utility of the code snippet you commented earlier as it is simply wrapping an inline conditionnal in a function.

They make the code more readable tbh. I would implement this only if it's possible to and useful for the source code. (forgot to mention that)

osu.Game/Screens/Play/HUD/HitErrorMeters/BarHitErrorMeter.cs
```diff
@@ -55,89 +56,112 @@ private void load(OsuColour colours)
...
+            else
+            {
+                InternalChild.Height = bar_height;
+                judgementsContainer.Width = judgement_line_width;
+                colourBars.Width = bar_width;
+                colourBarsEarly.Height = 0.5f;
+                colourBarsLate.Height = 0.5f;
+                arrowContainer.Width = chevron_size;
+                arrow.Y = 0.5f;
+            }
```
iiSaLMaN: There are also other checks like here for example. just to make it clear, `Width` and `Height` are components of the vector `Size`, `X` and `Y` are components of `Position`, so you can apply the same thing as I said above.

frenzibyte (changes requested):

osu.Game/Screens/Play/HUD/HitErrorMeters/BarHitErrorMeter.cs
Game4all: Didn't know these were shorthand properties for accessing Size & Position properties

Game4all:
Closing this as I don't intend to work further on this one.

PR #6363, opened 2019-10-02
Only lock the database for the duration of a deletion

### LINKED ISSUE #6334: Freeze when finishing a play while deleting all beatmaps
**Describe the crash:** When deleting all beatmaps and then going to play, when you finish the play the results screen (SoloResults) never appears and the game freezes 

**Screenshots or videos showing encountered issue:** Black screen with elements that remain after finishing the play, when the song finishes. I wasn't able to screenshot this, for some reason it shows main menu instead or is just black screen: https://i.imgur.com/yIcD2Jc.png

**osu!lazer version:** 2019.930.0

**Logs:** Nothing meaningful seemed to be logged but here they are: [logs.zip](https://github.com/ppy/osu/files/3672447/logs.zip)

Only thing interesting might be this database line: ``2019-09-30 05:19:07: [?????] Beginning import...``

**Computer Specifications:** Win10, Pentium 4405U, no graphics card

smoogipoo:
Fixes #6334

A context is retrieved in each `Delete()`


peppy:
As a note, this was previously there as an optimisation (to speed up deletions). They will now take longer as there is one transaction per delete. Probably an OK trade-off, but that's why the outer context was there before (was not a mistake).

peppy:
Saying that, it's absolutely fast even after this change. May have been improved in EF since.

peppy (dismissed):

smoogipoo:
I see. Makes sense but it's much faster with this change anyway.

peppy (approved):

PR #6369, opened 2019-10-02
Fix Alt-F4 being blocked during gameplay

### LINKED ISSUE #2887: Alt F4 in-map brings up pause menu, then quit the game on second trigger
Like the title said. Since I'm familiar with Alt F4 directly quitting to map selection, this throws me off my habits. Except, Alt F4 the secons time at pause exits.

Edit: Also Alt F4 quits the game at autoplay

Joehuu:
Fixes #2887

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestScenePause.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestScenePause.cs
@@ -137,9 +137,35 @@ namespace osu.Game.Tests.Visual.Gameplay
         }
 
         [Test]
+        public void TestQuickRetryFromFailedGameplay()
+        {
+            AddUntilStep("wait for fail", () => Player.HasFailed);
+            AddStep("quick retry", () => Player.GameplayClockContainer.OfType<HotkeyRetryOverlay>().First().Action?.Invoke());
+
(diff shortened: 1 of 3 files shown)


peppy:
The functionality was implemented to match stable. The confirm exit dialog was intended (think of it as a "save changes?" confirmation, since it will eventually block when a task is in progress). Unsure how much of this change we want.

Open to feedback.

Joehuu:
https://github.com/ppy/osu/issues/2887#issuecomment-400641181

Did I misunderstand?

peppy:
You didn't, but that was specifically about pause behaviour. I need a bit of time to figure what feels correct here, but as above I believe the quit confirmation should still fire under cases it is saving the user from potential distress.

Joehuu:
Leaving the exit dialog and just letting alt-f4 bypass pause.

peppy (dismissed):

peppy:
@smoogipoo want to double-check this one? i had to add additional handling for fail, probably worth a second set of eyes.

peppy (dismissed):

smoogipoo (approved):
Looks ok

PR #6374, opened 2019-10-03
Fix key counters appearing negative on intense beatmaps

### LINKED ISSUE #6198: Key counters going negative on some beatmaps
**Describe the bug:**  after skipping and then it freezing i pressed the keys to warm up and got a negative input...

Screen Shot 2019-09-20 at 2 35 49 PM 1

**osu!lazer version:**  2019.919.0

**Logs:** 
[database.log](https://github.com/ppy/osu/files/3637466/database.log)
[network.log](https://github.com/ppy/osu/files/3637467/network.log)
[performance.log](https://github.com/ppy/osu/files/3637468/performance.log)
[runtime.log](https://github.com/ppy/osu/files/3637469/runtime.log)

peppy:
When `FrameStabilityContainer` decides it needs multiple updates on the same frame, it ends up with an elapsed time of zero. This was interacting badly with the condition used in `RulesetInputManager` to govern playback direction.

I have changed this to use `Rate` as exposed by the frame stable clock.

- Closes #6198.

### DIFF
--- a/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
+++ b/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
@@ -47,6 +47,11 @@ namespace osu.Game.Rulesets.UI
 
         private IFrameBasedClock parentGameplayClock;
 
+        /// <summary>
+        /// The current direction of playback to be exposed to frame stable children.
+        /// </summary>
+        private int direction;
+
         [BackgroundDependencyLoader(true)]
         private void load(GameplayClock clock)
         {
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -169,6 +164,13 @@ private void updateClock()
             }
             finally
             {
+                if (newProposedTime != manualClock.CurrentTime)
+                    direction = newProposedTime > manualClock.CurrentTime ? 1 : -1;
+
+                manualClock.CurrentTime = newProposedTime;
```
smoogipoo: This doesn't look correct, due to line 160.

smoogipoo (changes requested):
Causes auto to miss on centipede

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
peppy: ah, that was on my mental check list but i forgot to fix it. good catch.

smoogipoo (approved):

PR #6375, opened 2019-10-03
Fix random failures on BeatmapCarousel filter test

peppy:
The "un-filter" step causes a `SelectNextRandom` invocation. If this happens to select a difficulty in set 3 other than the previously buffered difficulty #2, the subsequent test would fail.

I've split this test out to remove the random element, but added a new assert to ensure buffered (previously visited?) difficulty is re-selected on return to the same set.


bdach:
Thank you for spotting this, I really should have dug into the actual test more instead of focusing on its execution.

smoogipoo (approved):

PR #6379, opened 2019-10-03
Fix possible nullref in MusicController

smoogipoo:
As seen in https://ci.appveyor.com/project/peppy/osu/builds/27846159/tests

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -75,7 +75,7 @@ namespace osu.Game.Overlays
         /// <summary>
         /// Returns whether the current beatmap track is playing.
         /// </summary>
-        public bool IsPlaying => beatmap.Value?.Track.IsRunning ?? false;
+        public bool IsPlaying => beatmap?.Value?.Track.IsRunning ?? false;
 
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/MusicController.cs
```diff
@@ -75,7 +75,7 @@ public void ChangeBeatmapSetPosition(BeatmapSetInfo beatmapSetInfo, int index)
         /// <summary>
         /// Returns whether the current beatmap track is playing.
         /// </summary>
-        public bool IsPlaying => beatmap.Value?.Track.IsRunning ?? false;
+        public bool IsPlaying => beatmap?.Value?.Track.IsRunning ?? false;
```
iiSaLMaN: Why not use `current` instead?
```suggestion
        public bool IsPlaying => current?.Track.IsRunning ?? false;
```

frenzibyte (changes requested):

osu.Game/Overlays/MusicController.cs
smoogipoo: Sure

frenzibyte (approved):

peppy (approved):

PR #6393, opened 2019-10-04
Implement beatmap ruleset selector component

frenzibyte:
- Implements beatmap ruleset selector component with tests ensuring correct behaviour.

- [x] Depends on #6522


osu.Game/Overlays/BeatmapSet/BeatmapRulesetSelector.cs
```diff
@@ -0,0 +1,50 @@
...
+            get => beatmapSet;
+            set
+            {
+                if (value == beatmapSet)
+                    return;
+
+                beatmapSet = value;
+
+                foreach (var tab in TabContainer.TabItems.OfType<BeatmapRulesetTabItem>())
+                    tab.SetBeatmaps(beatmapSet?.Beatmaps.FindAll(b => b.Ruleset.Equals(tab.Value)));
```
bdach: I find this naming a bit too general, since as far as I can see the tab does not hold the actual beatmaps, just a per-ruleset count - how about something like `SetCountForRuleset`/`CountBeatmapsForRuleset` (if my other suggestion is applied)?

osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
```diff
@@ -0,0 +1,148 @@
...
+
+            count.Colour = colour.Gray9;
+            bar.Colour = colour.Blue;
+
+            updateState();
+        }
+
+        public void SetBeatmaps(List<BeatmapInfo> beatmaps)
+        {
+            Trace.Assert(beatmaps?.TrueForAll(b => b.Ruleset.Equals(Value)) ?? true, "A beatmap has a ruleset not of this tab value");
```
bdach: This looks like a quite expensive assert, which could be avoided if the control flow was inverted, ie. the full list (or even `IEnumerable`, no need for a concrete list here) was passed down and the ruleset tabs counted the appropriate beatmaps themselves - then there's no chance of a mix-up.

bdach (commented):
I checked this out - visually looks ok. Can't comment on the visual component/bindable usage too much since I'm not very well-versed in those yet, but I have one suggestion to the structure.

osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
iiSaLMaN: The data required for this ruleset selector is the list of beatmaps of this tab's value, the design may changes and further data may be wanted for the tab, but i'll change it if it **should** receive the count only.

osu.Game/Overlays/BeatmapSet/BeatmapRulesetSelector.cs
iiSaLMaN: as i said down there, the data required for this tab is the beatmaps of it's value (ruleset), i'm concerned whether to change to what we actually need now (`SetBeatmapsCount`) or keep it as is..

iiSaLMaN:
> Can't comment on the visual component/bindable usage too much since I'm not very well-versed in those yet, but I have one suggestion to the structure.

I specifically picked caching the ruleset bindable as `IBindable<RulesetInfo>` is because it's the best way to know the current ruleset and know if it changed without passing it from the `Header`, I may also step forward and cache it inside the beatmap set overlay.

osu.Game/Overlays/BeatmapSet/BeatmapRulesetSelector.cs
peppy: wouldn't it be more in-line with the rest of our code to pass the beatmapset down (whether bindable or property) and have the count or whatever computed within the tab?

smoogipoo (changes requested):

peppy (approved):

PR #5126, opened 2019-06-23
Implement ruleset selection in beatmap set overlay

frenzibyte:
Closes #5111

PR #6394, opened 2019-10-04
Add ruleset selection to beatmap set overlay

### LINKED ISSUE #5111: Tabs for different Osu modes.
**Describe the new feature:** 
In the gray area above the beatmap page on osu!lazer, I would recommend adding tabs for beatmaps that are specific for different osu modes. This is already fixed on the website, but it doesn't exist in lazer.
osu_2019-06-21_18-34-28

Add navigation tabs for `osu!`, `osu!ctb`, `osu!catch`, and `osu!mania` on the gray area above the beatmap page overlay on Lazer(designed like on the website).

frenzibyte:
- Adds beatmap ruleset selector to the beatmap set overlay with tests ensuring correct behaviour with the beatmap picker.
- [x] Depends on #6393 

Supersedes #5126 
Closes #5111


osu.Game/Overlays/BeatmapSet/BeatmapPicker.cs
```diff
@@ -50,11 +52,11 @@ public BeatmapSetInfo BeatmapSet
 
         private void updateDisplay()
         {
-            difficulties.Clear();
+            Difficulties.Clear();
 
             if (BeatmapSet != null)
             {
-                difficulties.ChildrenEnumerable = BeatmapSet.Beatmaps.OrderBy(beatmap => beatmap.StarDifficulty).Select(b => new DifficultySelectorButton(b)
+                Difficulties.ChildrenEnumerable = BeatmapSet.Beatmaps.FindAll(b => b.Ruleset.Equals(ruleset.Value)).OrderBy(b => b.StarDifficulty).Select(b => new DifficultySelectorButton(b)
```
bdach: Any reason this is not a `.Where()`? Could save a list allocation.

iiSaLMaN: Right.. I actually thought list functions are faster than linq, guess I should check the perfomance..

peppy (approved):

PR #6396, opened 2019-10-04
Allow fallback decoder overwrite

bdach:
Follow-up to #6117. Resolves crashes on start-up in the difficulty calculator.

# Summary

`LegacyDifficultyCalculatorBeatmapDecoder` was registered as a fallback decoder in commit ffde389 for future use in the server-side difficulty calculation components. Due to the pre-existing fallback registrations this caused a runtime crash when the diffcalc components were started.

Add a test reproducing this scenario to prevent the issue from resurfacing in the future and remove the check for pre-existing fallback registration along with the exception. The xmldoc for the registration function has been extended to make users aware of possible consequences of calling it.

Sorry for not catching this in time in the original PR, saw that change slightly too late.

### DIFF
--- a/osu.Game.Tests/Beatmaps/Formats/LegacyBeatmapDecoderTest.cs
+++ b/osu.Game.Tests/Beatmaps/Formats/LegacyBeatmapDecoderTest.cs
@@ -591,5 +591,27 @@ namespace osu.Game.Tests.Beatmaps.Formats
                 Assert.Throws<IOException>(() => Decoder.GetDecoder<Beatmap>(stream));
             }
         }
+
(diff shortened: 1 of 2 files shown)


peppy (approved):

peppy:
The global `fallback_decoders` registration system is still something I don't hugely agree with (one of the only `static` components anywhere in the game) but this seems same for now.

I'd hope we can eventually refactor this out.

bdach:
Yeah, I agree that if we can get rid of the `static` we should. I don't have a suggestion as to how right now, but I'll have a think about it.

PR #6400, opened 2019-10-04
Fix keybinding order of beatmap options

### LINKED ISSUE #6399: Beatmap options order does not correspond with keybindings
Pressing numbers from 1-4 when in the beatmap options does not select the respective action as shown by the menu.

**Current** 
1 - remove from unplayed
2 - clear local scores
3 - edit beatmap
4 - delete all difficulties
**Corrected**
1 - delete all difficulties
2 - remove from unplayer
3 - clear local scores
4 - edit beatmap

image

Logs are irrellevant - Version 2019.930.0

Joehuu:
Fixes https://github.com/ppy/osu/issues/6399


bdach:
We might want to fix the button order instead [as outlined here](https://github.com/ppy/osu/issues/6399#issuecomment-538585390).

Joehuu:
Shouldn't `Edit beatmap` button be last as it is exclusive in `PlaySongSelect`? On `MatchSongSelect`, this button is gone and keybind 3 is skipped.

Proposed order:
- Remove from unplayed (1)
- Clear local scores (2)
- Delete all difficulties (3)
- Edit beatmap (the exclusive one) (4)

We can remove the depth specification all together and just move `Delete all difficulties` to last line *if this happens*.

bdach:
Not necessarily opposed to that to be fair, just wanted to make sure everything was clear before agreeing on a fix.

peppy (approved):

peppy:
I think this is fine and more sane for now 👍 .

PR #4856, opened 2019-05-26
Apply safe area padding and edge snapping

swoolcock:
WIP for design and implementation discussion.

Depends on https://github.com/ppy/osu-framework/pull/2420

Developed on desktop by hardcoding `Window.SafeAreaPadding.Value = new MarginPadding(100)`.
Tested on iPhone X and iPad Pro 12.9".

* Overlays and the screen stack are padded with a `SafeAreaContainer`.
* Backgrounds, `ParallaxContainer`s, etc. are wrapped in individual `EdgeSnappingContainer`s.
* `BeatmapInfoWedge` is wrapped in a masking container to hide the sheared edge effect.

WIP examples:

Screen Shot 2019-06-28 at 9 52 22 pm

Screen Shot 2019-06-28 at 9 52 54 pm

PR #6403, opened 2019-10-05
Apply safe area insets for overlays

swoolcock:
Applies the minimum required to implement safe area insets for overlays.

* The top level overlay containers are wrapped in a `SafeAreaContainer`.
* Some individual overlays have been tweaked to ensure that backgrounds run off the edge of the screen, etc.
* The screen stack takes into account the safe area when pushing it below the toolbar.

This is currently "opt-in", as individual screens will need to wrap their contents in a `SafeAreaContainer` on a per-case basis.  I thought this was a better solution than applying a blanket padding to the screen stack and risking regressions.

Supersedes #4856

### DIFF
--- a/osu.Game/Overlays/VolumeOverlay.cs
+++ b/osu.Game/Overlays/VolumeOverlay.cs
@@ -43,11 +43,16 @@ namespace osu.Game.Overlays
 
             AddRange(new Drawable[]
             {
-                new Box
+                new SafeAreaContainer
                 {
-                    RelativeSizeAxes = Axes.Y,
-                    Width = 300,
-                    Colour = ColourInfo.GradientHorizontal(Color4.Black.Opacity(0.75f), Color4.Black.Opacity(0))
+                    RelativeSizeAxes = Axes.Both,
+                    SafeAreaOverrideEdges = Edges.All,
+                    Child = new Box
+                    {
+                        RelativeSizeAxes = Axes.Y,
+                        Width = 300,
+                        Colour = ColourInfo.GradientHorizontal(Color4.Black.Opacity(0.75f), Color4.Black.Opacity(0))
+                    },
                 },
                 new FillFlowContainer
                 {
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -457,10 +457,17 @@ namespace osu.Game
                         logoContainer = new Container { RelativeSizeAxes = Axes.Both },
                     }
                 },
-                overlayContent = new Container { RelativeSizeAxes = Axes.Both },
-                rightFloatingOverlayContent = new Container { RelativeSizeAxes = Axes.Both },
-                leftFloatingOverlayContent = new Container { RelativeSizeAxes = Axes.Both },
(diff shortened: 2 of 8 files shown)


osu.Game/Overlays/VolumeOverlay.cs
```diff
@@ -43,11 +43,16 @@ private void load(AudioManager audio, OsuColour colours)
...
             AddRange(new Drawable[]
             {
-                new Box
+                new SafeAreaContainer
                 {
-                    RelativeSizeAxes = Axes.Y,
-                    Width = 300,
-                    Colour = ColourInfo.GradientHorizontal(Color4.Black.Opacity(0.75f), Color4.Black.Opacity(0))
+                    RelativeSizeAxes = Axes.Both,
+                    SafeAreaOverrideEdges = Edges.All,
```
bdach: Why is `Edges.Right` included here? From what I can tell it should work fine without.

bdach (commented):
Looks to be doing what is expected, from what I can tell.

I'm not sure if it's within scope here, but it seems `VolumeOverlay`'s mute button stays outside of the safe area instead of being pushed inside of it:

image

Just bringing it up since that file was modified here, but if it's a matter to be fixed in another PR later down the line, feel free to disregard.

osu.Game/Overlays/VolumeOverlay.cs
swoolcock: The background Box needs to snap to all edges of the screen.

swoolcock:
How did you test it to take that screenshot? Just hard coded the safe area margins?

bdach:
I breakpointed at `OsuGameDesktop.SetHost` and manually set `SafeAreaPadding` via debugger (100 on each side). I don't *think* that would be an improper way to set it, since I'm not seeing any `ValueChanged` subscribers at that point, but maybe it is and I should hard-code it somewhere in the framework itself?

It does make sense to me that this would happen judging from the source though - the `FillFlowContainer` with the volume controls is side-by-side with the new `SafeAreaContainer`, not within it.

swoolcock:
The `SafeAreaContainer` in `VolumeOverlay` is solely for overriding the background edges.  It's somewhat of a misnomer, because it's really a `NotSafeAreaContainer` since it's undoing the margins applied by the one that wraps the overlays.

swoolcock:
As for testing the safe areas, I just put this as the first line in the `OsuGame` BDL:

```
Host.Window.SafeAreaPadding.Value = new MarginPadding(100);
```

peppy:
Might be worth adding a comment about the `VolumeOverlay` usage for clarity.

swoolcock:
It's the same for all uses of `SafeAreaContainer` that have `SafeAreaOverrideEdges` set, not just the `VolumeOverlay` background.

frenzibyte:
Closing this as well in favour of #16772.

peppy:
Similarly to #4856, I'm going to close this as being superseded by #16772. We will likely want to revisit these cased on a per-implementation basis in the future, but that future is not quite here yet.

PR #6406, opened 2019-10-05
Resume music to same position when exiting gameplay

### LINKED ISSUE #4684: When exiting a play, track position seeks to preview point instead of staying
**Describe the new feature:** 
If the player quit the level during the game, the music should not reset to preview point. It feels more natural to keep playing.

Joehuu:
Fixes https://github.com/ppy/osu/issues/4684.

Didn't need to use music controller as pause condition check was removed in https://github.com/ppy/osu/pull/5764.

The restart bool parameter usages were removed sometime ago, so I'm using it again.


bdach (approved):
Works as advertised, I don't see any problems with the code either.

PR #6407, opened 2019-10-05
Autoscroll playlist on song change

### LINKED ISSUE #6368: Music player's list never jumps to now playing song
**Describe the bug:** It probably should when you open it and maybe when song changes

**osu!lazer version:** 2019.930

**Logs:**

UselessToucan:
resolves #6368


bdach:
The actual change works fine, but I'm unconvinced it needs its own separate setting in its own separate section.

UselessToucan:
> I'm unconvinced it needs its own separate setting in its own separate section

I may want not to have the autoscrolling. I.e. it should be configurable.

iiSaLMaN:
@UselessToucan I believe what @bdach said is that it shouldn't have it's own sub-section. but rather, I think this shouldn't be in the audio section, but maybe in the UI sub-section.

bdach:
> @UselessToucan I believe what @bdach said is that it shouldn't have it's own sub-section. but rather, I think this shouldn't be in the audio section, but maybe in the UI sub-section.

No, I actually don't think it needs to be a setting at all. I understand there might be a UX concern that autoscroll can interfere with the user manually scrolling to find a song, but that would be broken regardless of the setting. If we want to handle that cleanly then I guess a manual scroll could prevent an autoscroll until some event happens (i.e. user manually selects a song or closes the playlist).

peppy:
Should be default behaviour and not a setting.

peppy:
This fails if the playlist isn't visible at the point the scroll happens, but that is a separate issue which likely requires a framework fix, so I think it's fine to move ahead with this as-is.

PR #6411, opened 2019-10-06
Fix tab controls overflowing right content at max ui scale

### LINKED ISSUE #4415: hamburger menu not visible in chat if too many chats are open
osu_2019-03-05_21-10-45

Joehuu:
Also fixes https://github.com/ppy/osu/issues/4415.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapDetailAreaTabControl.cs
+++ b/osu.Game/Screens/Select/BeatmapDetailAreaTabControl.cs
@@ -20,7 +20,6 @@ namespace osu.Game.Screens.Select
         public static readonly float HEIGHT = 24;
         private readonly OsuTabControlCheckbox modsCheckbox;
         private readonly OsuTabControl<BeatmapDetailTab> tabs;
-        private readonly Container tabsContainer;
 
         public Action<BeatmapDetailTab, bool> OnFilter; //passed the selected tab and if mods is checked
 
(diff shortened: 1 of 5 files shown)


bdach:
For visual reference:

| Before | After |
| :-: | :-: |
| osu_2019-10-06_19-57-30 | osu_2019-10-06_19-58-41 |
| osu_2019-10-06_20-02-33 | osu_2019-10-06_20-03-16 |

osu.Game/Screens/Select/BeatmapDetailAreaTabControl.cs
```diff
@@ -69,6 +74,13 @@ private void load(OsuColour colour, OsuConfigManager config)
             tabs.Current.TriggerChange();
         }
 
+        protected override void Update()
+        {
+            base.Update();
+
+            tabsContainer.Padding = new MarginPadding { Right = modsCheckbox.Width };
```
bdach: Just to clarify, is this done in `Update` because the checkbox changes width dynamically?

bdach (commented):
Visually looks good which is probably enough seeing that this is a layout reorder, but I just want to clarify one thing.

Aergwyn:
Was the second part necessary or an accident from the changes? It looked fine to me.

bdach:
> Was the second part necessary or an accident from the changes? It looked fine to me.

I took that screenshot above where the overflow menu is behind the details container from current `master`, unsure if it was always broken or a regression.

On another note though I did notice that the on extreme scalings the `Selected Mods` checkbox item on leaderboard tabs breaks badly, I don't know if there is currently an easy way of fixing that, and probably out of scope anyways?

osu.Game/Screens/Select/BeatmapDetailAreaTabControl.cs
Joehuu: Checking again it does change width when the the text gets bold when checked, so this is probably right.

Initially, I just didn't want to specify some random fixed number as the padding.

peppy:
In the above screenshots, the "..." button seems to be moving. It shouldn't move (should remain right-aligned in all cases).

Joehuu:
Went an old approach (fixed numbers), since I think using `Update` is bad. Fixed the concerns.

peppy (approved):

PR #6416, opened 2019-10-07
Add API functionality to beatmap favourite button

EVAST9919:


osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
```diff
@@ -67,16 +53,22 @@ private void load()
...
+
+            Action = () =>
+            {
+                if (loading.State.Value == Visibility.Visible)
+                    return;
+
+                loading.Show();
+
+                request?.Cancel();
+                request = new PostBeatmapFavouriteRequest(BeatmapSet.Value?.OnlineBeatmapSetID ?? 0, favourited.Value ? BeatmapFavouriteAction.UnFavourite : BeatmapFavouriteAction.Favourite);
```
bdach: This probably shouldn't send a request at all if `BeatmapSet.Value` is null.

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
```diff
@@ -67,16 +53,22 @@ private void load()
...
+                request.Success += () => favourited.Value = !favourited.Value;
```
bdach: I'm a bit worried about this potentially desyncing with the actual web state. Does that endpoint not actually return whether or not the favourite was added or removed? I'd be less concerned about potential discrepancies then.

bdach (commented):
Looks to be working ok and I *think* it's race-free - I have some nitpicks though.

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
EVAST9919: afaik it returns total favourite beatmaps amount

bdach: Agh, that doesn't help here much. I guess this is passable then.

iiSaLMaN: I agree with @bdach, you should open an issue web-side for this, this may be dangerous.

peppy: Yep, should ask for API change.

bdach:
* Tagging as blocked pending resolution of ppy/osu-web#5108.
* By the way, another thing to check here is how the favourite button handles hitting the maximum allowed number of favourited beatmaps.

bdach:
Unblocking after close of web issue, I guess the toggle should be *mostly* safe as-is.

EVAST9919:
How about just send another request after the first one to redownload beatmap metadata and see, whether this beatmap is favourited?

smoogipoo:
You would need to constantly send requests if you wanted to keep it somewhat consistent. Seems not worthwhile.

EVAST9919:
Okay, after hitting a limit of favourited beatmaps there's no actual API response. It worth it to create a web issue.

nanaya:
That's not what that endpoint is for. Use `beatmapsets/:id` to get its latest data.

peppy:
To clarify, the endpoint *will* return an error if you are at the limit of allowed favourites.

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
```diff
@@ -68,6 +69,18 @@ private void load(IAPIProvider api)
                 request?.Cancel();
                 request = new PostBeatmapFavouriteRequest(BeatmapSet.Value?.OnlineBeatmapSetID ?? 0, favourited.Value ? BeatmapFavouriteAction.UnFavourite : BeatmapFavouriteAction.Favourite);
                 request.Success += () => favourited.Value = !favourited.Value;
+                request.Failure += exception =>
+                {
+                    if (exception.Message == "UnprocessableEntity")
```
bdach: This feels wrong due to hinging on a magic string (which *is* the name of an HTTP error code, but still), and [I don't think 422 isn't guaranteed to be returned on this error only either](https://github.com/ppy/osu-web/blob/666a8135dc576afffb185dd5315d95fb1c4ca302/app/helpers.php#L560-L563). If anything is to be displayed to the user here, it's probably whatever web returns in the `error` message.

peppy: osu-web should probably be returning an error in the body.

peppy: surprise, it already was

peppy:
@smoogipoo requesting a quick check on this, as i made many changes.

can split out `APIRequest `error handling and `DirectOverlay` fixes if you'd prefer (they are relatively minor, though).

peppy (dismissed):

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
```diff
@@ -55,28 +42,49 @@ private void load()
...
+                    favourited.Toggle();
+                    loading.Hide();
+                };
+
+                request.Failure += e =>
                 {
-                    pink.FadeOut(200);
-                    icon.Icon = FontAwesome.Regular.Heart;
-                }
+                    notifications.Post(new SimpleNotification
```
bdach: This leads to `TestSceneBeatmapSetOverlay.TestAvailability()` crashing the test browser - clicking the button in that test raises a 404, which then causes an attempt to access `notifications` which is null inside the test. This was a problem before too. Unsure if we just want to null-check this, refactor somehow or leave as-is.

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
```diff
@@ -55,28 +42,49 @@ private void load()
...
+            }, true);
+
+            Action = () =>
+            {
+                if (loading.State.Value == Visibility.Visible)
                     return;
 
-                favourited.Value = setInfo.NewValue.OnlineInfo.HasFavourited;
-            });
+                // guaranteed by disabled state abvove.
```
bdach: I feel horrible for even suggesting this, but
```suggestion
                // guaranteed by disabled state above.
```

bdach (commented):
Not @smoogipoo, but took a look. The new changes are good with me. I got some things to point out but they're not supremely relevant.

osu.Game/Overlays/BeatmapSet/Buttons/FavouriteButton.cs
iiSaLMaN: should null check as null notifications overlay is allowed here.

smoogipoo:
The button needs to disable when API is logged out.

smoogipoo:
It also needs to have a test. I highly suggest moving the fix out into its own separate PR.

smoogipoo (approved):

PR #6421, opened 2019-10-07
Update game code inline with TestSuite introduction

UselessToucan:
Update game code to make things work after [this](https://github.com/ppy/osu-framework/pull/2872) framework change.
This PR was made in order to demonstrate impact on framework consumers.
Note that game test code still violates SRP and needs refactoring. But it can be done later.

### DIFF
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
@@ -35,7 +35,6 @@ using osuTK.Graphics;
 
 namespace osu.Game.Tests.Visual.Background
 {
-    [TestFixture]
     public class TestSceneUserDimContainer : ManualInputManagerTestScene
     {
         public override IReadOnlyList<Type> RequiredTypes => new[]
--- a/osu.Game.Tests/Visual/Components/TestSceneIdleTracker.cs
+++ b/osu.Game.Tests/Visual/Components/TestSceneIdleTracker.cs
@@ -11,7 +11,6 @@ using osuTK.Graphics;
 
 namespace osu.Game.Tests.Visual.Components
 {
-    [TestFixture]
     public class TestSceneIdleTracker : ManualInputManagerTestScene
     {
         private IdleTrackingBox box1;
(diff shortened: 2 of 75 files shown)


bdach:
I don't want to be rude, but 3 PRs and with a total diffstat of `+3,934 -3,619` I don't see any tangible benefits from this series of changes. What's the end goal here? And for the record:

- I don't really like using the single responsibility principle as a primary reasoning, since it's been muddled to hell and back and you can argue all day on whether something does or doesn't violate it. 
- Decoupling the test scene from the test logic for me is *worse*, since now I have to navigate *two* files to make/read a test. It would be worth the cost if the scene was reused between tests, but I don't know how often that might be necessary and to me the gain does not justify the cost.
- Forcing the tests to use nunit attributes in my opinion might sacrifice readability.

Personally I'd want to see a killer example of a test where you think your approach will *definitively* beat the existing one. Until then, I think it's going to be very difficult to convince anyone to review and merge all of this with no real benefit in sight.

UselessToucan:
> What's the end goal here?

The end goal is to decompose a huge class with multiple responsibilities to a set of smaller classes. I believe that small classes are easier to maintain.

> I'd want to see a killer example of a test where you think your approach will _definitively_ beat the existing one.

The first thing that comes to mind is a test scene with nested classes. For example, look at the [TestSceneNestedMenus](https://github.com/ppy/osu-framework/blob/937ea02f265df873ff51b57d78d039439e71098a/osu.Framework.Tests/Visual/UserInterface/TestSceneNestedMenus.cs) class. There are 493 LOC of testing logic mixed with UI. Large files are hard to maintain(especially when there are multiple responsibilities mixed there) and I would prefer to avoid it. Also, it is uncomfortable to read large files without IDE.

bdach:
As a part of ongoing efforts to reduce the number of stale open pull requests, I am going to close this pull request for the following reasons:

* Lack of general support for the direction of changes suggested here.
* Lack of updates for over a year.
* Large amount of merge conflicts remaining.

If you still wish to push for this change to be made, we would need to be convinced that there are tangible benefits of this direction, which have not been demonstrated yet with a concrete example of a test scene that would read way better than what we have now.

I would however recommend looking at other pain points of our current testing setup that have been listed [in this comment](https://github.com/ppy/osu-framework/pull/3887#discussion_r489389316), if you're so inclined.

PR #6423, opened 2019-10-08
Fix song select not correctly playing tracks in some cases

### LINKED ISSUE #6413: Audio preview on beatmap starts from beginning when clicking osu!direct download notification
I think title does not describe it enough, so the beatmap’s audio is automatically paused. The audio does not start on preview as well if downloaded on direct and pressed its notification.
[Here’s the video](https://youtu.be/L2Njni1u1xI)

peppy:
Closes #6413. Basically reverts #5934.


smoogipoo:
This basically also unresolves https://github.com/ppy/osu/issues/5920, which I don't think we want.

peppy:
Did you test / reproduce thi?

smoogipoo:
Yes, I did.

peppy:
As discussed IRL, only happens when manually pausing the track. Will look into it.

smoogipoo (approved):

PR #6427, opened 2019-10-08
Better signify under construction screens in editor

peppy:
- Unimplemented composers will now show a message to the user
- Added missing placeholder screens for editor tabs


bdach (approved):
The screens do their job, and look really nice in the game 👍

PR #6428, opened 2019-10-08
Fix ignored song select test

peppy:
Was not working due to the convert setting. Reversing the test fixes this easily/quickly.


bdach (approved):
This fix makes sense to me 👍

smoogipoo (approved):

PR #6431, opened 2019-10-08
Move editor selections relative to hitobject start positions

smoogipoo:
Using the drag event as is does not work for snapping, where the screen-space coordinate mouse position needs to be snapped and a delta needs to be constructed from the resulting position.

With the way this is set up, all that should be required is snapping `movePosition`.


osu.Game.Rulesets.Mania/Edit/ManiaSelectionHandler.cs
```diff
@@ -31,13 +29,16 @@ private void load(IAdjustableClock clock)
...
         }
 
-        public override void HandleDrag(SelectionBlueprint blueprint, DragEvent dragEvent)
+        public override void HandleMovement(MoveSelectionEvent moveEvent)
         {
-            adjustOrigins((ManiaSelectionBlueprint)blueprint);
-            performDragMovement(dragEvent);
-            performColumnMovement(dragEvent);
+            var maniaBlueprint = (ManiaSelectionBlueprint)moveEvent.Blueprint;
+            int lastColumn = maniaBlueprint.HitObject.HitObject.Column;
```
bdach: The nested `HitObject`s had me in a double take, but I guess there's not really a better way to name them...

bdach (approved):
Checked this out - can't say I'm in context on this, but I think I understand what this aims to do from the xmldocs, and I didn't manage to break it in visual tests, so looks good to go for me 👍

osu.Game.Rulesets.Mania/Edit/ManiaSelectionHandler.cs
iiSaLMaN: what about renaming the DHO `DrawbleObject`?

smoogipoo: It's something we can consider and discuss, but not necessary for now. It would need to be done everywhere (e.g. `Playfield.AllHitObjects`, `HitObjectContainer.Objects`, `HitObjectContainer.AliveObjects`.

Out of scope for this PR.

peppy (approved):

Issue #4391, opened 2019-03-02
Rates Mod


LeNitrous:
**Describe the feature:** A more flexible rate-changer mod.

**Proposal designs of the feature:**
For the context of this proposal, "rate" will be defined as the speed of the map being played. Currently in osu! there are two rate changing mods namely Double Time and Half Time which changes the rate of a map to 1.5x or 0.75x respectively.

The issue is that the options seem too limiting and not to mention not friendly. There are cases where other people prefer to speed up or slow down maps for the purpose of training. An example is from DJ Shrapnel's loved set [TAKECORE OF YOURSELF](https://osu.ppy.sh/beatmapsets/480479) where each difficulty is a separate audio with matching difficulty file for that specific rate.
![image](https://user-images.githubusercontent.com/20495991/53682147-c6d23b80-3d2d-11e9-852e-28b7cd6633c1.png)

The community already made third party tools such as smoogipooo's [Map Speed Changer](https://osu.ppy.sh/community/forums/topics/114765) although it only makes a difficulty that is tied to what rate you set not to mention you still have to edit the audio file in any audio editing software.

I would also like to participate in implementing this although I reckon that I'd post this as an issue first to get the general gist of how the implementation should be if ever this gets accepted.

peppy:
At some point we will be adding per-mod settings, at which point DT/HT could have such a setting.

LeNitrous:
Closing as this can now be achieved under the rate adjustment mods.

Issue #4175, opened 2019-01-28
Play only part or section of the map


GSculerlor:
**Describe the feature:** Sometimes players want to try to play only part of the map (where they fail or where the hard pattern appears) and not full of the map. At this point at stable, a lot of players can only create a new diff on the map that only contains a part that they want to play to do that.

**Proposal designs of the feature:** Add feature maybe like training mod or something like that, specify the start and end point/time on the map that player wants to try so players can play that specific part without creating new difficulty

ticotaco72:
It something like Autoplay but you can in any moment change it to normal gameplay

c0linw:
I think it would be nice to have it implemented as a separate part of the game instead of as a mod. An ability to "scroll" through the beatmap in real time a la McOsu would be nice in a "Practice" mode alongside "Play" and "Edit" out of the main menu.

Aergwyn:
Isn't Target Practice an unimplemented mod for exactly that or am I remembering wrong?

LeNitrous:
Target Practice is where random targets appear in triangles. Not the one they're looking for.

Issue #3201, opened 2018-08-10
Map settings


WebFreak001:
I think a new UI for configuring maps before playing would be a cool feature to add.

Being able to change settings like:
* Playback rate (more finely controllable than DT/HT)
* AR (instead of EZ/HR which have other effects)
* HP (instead of EZ/HR which have other effects)
* OD (instead of EZ/HR which have other effects)

Changing the settings would make it unranked.

It would practically be a replacement for some mods for practicing or doing cool plays. Also open for more settings suggestions.

It could also be a place where Mods could put in settings (like FL radius, floatiness the new arrange mod which is being discussed in the PRs, hidden fade in time, etc.)

Maybe it would be placed in the beatmap settings thing at the top right corner when you start a map (where you can enable/disable storyboard, change dim, etc.)

I am considering adding it, how would such a feature be received?

ZapTech43:
Same Like These? : 
- https://osu.ppy.sh/community/forums/topics/141204
- https://osu.ppy.sh/community/forums/topics/774895?start=6730968

WebFreak001:
yeah pretty much like those. Before you would always create copies of the difficulties in the editor, so I think this would be useful to simply skip that step.

jpsouzasilva:
The McOsu client has the whole beatmap configuration available even during gameplay (you press F1 and it brings up this overlay below).

![screenshot2](https://user-images.githubusercontent.com/17429390/46641182-91db0780-cb46-11e8-81c0-3eae1230f604.png)

What about creating a "Training" mod that enables this kind of panel when it's on? It could be brought up before the beatmap starts, like https://github.com/ppy/osu/blob/master/osu.Game/Screens/Play/PlayerSettings/VisualSettings.cs

PR #6432, opened 2019-10-08
Add per-mod settings

### LINKED ISSUE #6272: Allow mods to have their own settings
**Describe the new feature:** Allow mods to be customized with their own settings. For instance: https://github.com/ppy/osu/issues/4391, https://github.com/ppy/osu/issues/4175 and https://github.com/ppy/osu/issues/3201

Creating this issue for visibility as well as to discuss the implementation details.

**Proposal designs of the feature:** Create a new interface `IModHasSettings` which is different from `IReadFromConfig` as that persists through sessions and also an added benefit to avoid the need of creating new options and ending up cluttering the settings sidebar.

Mod settings will be found a sidebar alongside Input Settings and Visual Settings before starting the map. It's contents will be created from `IModHasSettings` method that should return a new drawable (or better yet implement a new class like `PlayerSettingItem` for streamlining) and the values of those controls will be read by the mod.

Mod settings will by dynamically created when we reach `PlayerLoader` depending on the mods selected

Current Implementation (Rate Mod as example)
image

LeNitrous:
Resolves #6272 and in turn allows #4391, #4175, and #3201 to be realized.

Creates a new interface `IModHasSettings` which allows a mod to create controls such as check boxes and sliders to allow the user control over it's properties.

A new button is found on `ModSelectOverlay` called "Customization" which opens a simple interface where all controls for customizing mods are found.

The Modify mod will be made in a separate PR and is only shown here as an example.

Example:
image

- [x] Depends on https://github.com/ppy/osu/pull/7081.


iiSaLMaN:
Can the customization area not hit the bottom area of the mod select overlay (where the mod buttons are)

osu.Game/Rulesets/Mods/IModHasSettings.cs
```diff
@@ -0,0 +1,15 @@
...
+using osu.Framework.Graphics;
+
+namespace osu.Game.Rulesets.Mods
+{
+    /// <summary>
+    /// An interface for mods that allows user control over it's properties.
+    /// </summary>
+    public interface IModHasSettings
+    {
+        Drawable[] CreateControls();
```
iiSaLMaN: This shouldn't pass a list of drawables, but rather, a list of settings controls maybe?

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -69,7 +69,7 @@ public abstract class Mod : IMod, IJsonSerializable
         /// <summary>
         /// Creates a copy of this <see cref="Mod"/> initialised to a default state.
         /// </summary>
-        public virtual Mod CreateCopy() => (Mod)Activator.CreateInstance(GetType());
+        public virtual Mod CreateCopy() => (Mod)MemberwiseClone();
```
iiSaLMaN: This should likely be in a different PR, kind-of out of scope for what this is implementing.

frenzibyte (changes requested):

osu.Game/Rulesets/Mods/IModHasSettings.cs
LeNitrous: I'd like to automatically generate it's settings based on the class' properties (maybe marked with an attribute?) but I'm not sure of that approach just yet.

osu.Game/Rulesets/Mods/Mod.cs
bdach: I think it's fine here, without this any mods with internal state would be thoroughly broken.

iiSaLMaN: This code is about fixing an issue when doing something to the mods as discussed, this should've been in another PR, just saying for the author to take note of that later on.

peppy:
This is missing any form of testing. How are you testing? Please make a test scene with a test mod which implements settings.

peppy:
Just a heads up that whatever you did to this PR to bring it up-to-date is not the correct way (`git merge upstream/master` is what you want in the future).

peppy:
Updated behaviour somewhat to use new logic in #7081.

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    public static class SettingSourceExtensions
+    {
+        public static IEnumerable<Drawable> CreateSettingsControls(this object obj)
+        {
+            var configProperties = obj.GetType().GetProperties(BindingFlags.GetProperty | BindingFlags.Public | BindingFlags.Instance).Where(p => p.GetCustomAttribute<SettingSourceAttribute>(true) != null);
+
+            foreach (var property in configProperties)
+            {
+                var attr = property.GetCustomAttribute<SettingSourceAttribute>(true);
```
huoyaoyuan: Logic neat:
```csharp
foreach (var property in obj.GetType.GetProperties(...))
{
    var attr = property.GetCustomAttribute<SettingSourceAttribute>(true);
    if (attr == null) continue;


}
```

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -0,0 +1,111 @@
...
+                        yield return new SettingsTextBox
+                        {
+                            LabelText = attr.Label,
+                            Bindable = bString
+                        };
+
+                        break;
+
+                    case IBindable bindable:
+
```
huoyaoyuan: Too many blank lines here.

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -0,0 +1,106 @@
...
+                    if (section.ModType == ModType.Conversion)
+                    {
+                        section.Mods = new Mod[]
+                        {
+                            new TestModCustomisable1(),
+                            new TestModCustomisable2()
+                        };
+                    }
+                    else
+                        section.Mods = new Mod[] { };
```
huoyaoyuan: `Array.Empty<>`. Can be addressed later.

osu.Game/Configuration/SettingSourceAttribute.cs
peppy: Applied in https://github.com/ppy/osu/pull/7081

osu.Game/Configuration/SettingSourceAttribute.cs
peppy: Applied in #7081

osu.Game/Rulesets/Mods/Mod.cs
smoogipoo: I'm still not quite sure about this. I believe in its current state it will break the easy mod.

What we really want here is to bind bindables between mods, so perhaps a custom implementation that _does_ use `CreateInstance` and also reflection to bind (or copy? unsure) bindables.

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -284,10 +331,32 @@ private void load(OsuColour colours, IBindable<RulesetInfo> ruleset, AudioManage
             Ruleset.BindTo(ruleset);
             if (mods != null) SelectedMods.BindTo(mods);
 
+            SelectedMods.ValueChanged += updateModSettings;
+            Ruleset.ValueChanged += _ => ModSettingsContent.Clear();
```
smoogipoo: Is this required? It should be handled by the `SelectedMods` bindable?

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -226,6 +232,17 @@ public ModSelectOverlay()
                                                     Right = 20
                                                 }
                                             },
+                                            CustomiseButton = new TriangleButton
+                                            {
+                                                Width = 180,
+                                                Text = "Customization",
```
smoogipoo:
```suggestion
                                                Text = "Customisation",
```

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -284,10 +331,32 @@ private void load(OsuColour colours, IBindable<RulesetInfo> ruleset, AudioManage
             Ruleset.BindTo(ruleset);
             if (mods != null) SelectedMods.BindTo(mods);
 
+            SelectedMods.ValueChanged += updateModSettings;
```
smoogipoo: This should probably be a `BindValueChanged(, true)`.

Also this shouldn't be inside `load()`.

Also this should probably not even exist, and `updateModSettings()` should be invoked by `selectedModsChanged` (see the existing `LoadComplete()`).

smoogipoo (changes requested):

osu.Game/Rulesets/Mods/Mod.cs
smoogipoo: Actually it probably won't break the easy mod since it's cloned with a zero-state, still feels kind of dodgy though.

smoogipoo (approved):

PR #6435, opened 2019-10-08
Fix cursor size trail scaling

### LINKED ISSUE #6103: Cursor trail doesn't resize when cursor scale is changed
When i try a on a cursorsize option, cursortrail don't resized smaller or bigger than 1x. (all skins with custom cursortrail)
And cursor trail repeating little a less.

Osu!Stable:
screenshotd10814
Osu!Lazer:
cursor bug

Skin: http://puu.sh/EgR6M/5d2addd9fc.osk

**osu!lazer version:** 
2019.913.0

**Logs:** 
Is a visual bug.

holly-hacker:
Fixes cursor trail not scaling with cursor in the osu! ruleset.

Old behaviour:
old trail x0.42

New behaviour:
new trail x0.42

Closes #6103


osu.Game.Rulesets.Osu/UI/Cursor/OsuCursorContainer.cs
```diff
@@ -37,9 +43,33 @@ public OsuCursorContainer()
...
-            config?.BindWith(OsuRulesetSetting.ShowCursorTrail, showTrail);
+            float scale = cursorScale.Value;
+
+            if (autoCursorScale.Value && beatmap.Value != null)
+            {
+                // if we have a beatmap available, let's get its circle size to figure out an automatic cursor scale modifier.
+                scale *= 1f - 0.7f * (1f + beatmap.Value.BeatmapInfo.BaseDifficulty.CircleSize - BeatmapDifficulty.DEFAULT_DIFFICULTY) / BeatmapDifficulty.DEFAULT_DIFFICULTY;
+            }
+
+            ActiveCursor.Scale = cursorTrail.Scale = new Vector2(scale);
```
iiSaLMaN: What about using this:
```suggestion
            Scale = new Vector2(scale);
            Size = 1 / Scale;
```

smoogipoo: Drawables generally shouldn't be changing their own properties. If a consumer were to set scale/size on this container, the changes you proposed would reset them.

I prefer the original change.

smoogipoo:
Some weird things are going on with scaling here: https://streamable.com/mt94h - size is not set when starting gameplay and is reverted when unpausing.

Also the pause cursor should scale too.

HoLLy-HaCKeR:
Should work now, I can't find anything odd outside of #6485 which seems out of scope.

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
```diff
@@ -38,7 +38,19 @@ public override void Show()
...
+                Add(localCursorContainer = newContainer);
+
+                clickToResumeCursor.CursorScale = newContainer.CalculatedCursorScale.Value;
+                clickToResumeCursor.Scale = new Vector2(newContainer.CalculatedCursorScale.Value);
+
+                newContainer.CalculatedCursorScale.ValueChanged += e =>
+                {
+                    clickToResumeCursor.CursorScale = e.NewValue;
+                    clickToResumeCursor.Scale = new Vector2(e.NewValue);
+                };
```
bdach: Possibly worth factoring the two-liners out into a method and calling that instead?

bdach (dismissed):
Concerns look addressed to me, so a tentative approve here - I've got just one note to the code.

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
HoLLy-HaCKeR: I don't think this should be factored out, it's only 2 lines that are only duplicated once. Even making it a local method would be more complicated than needed.

peppy: You can use `BindValueChanged(delegate, true)` to solve this. it's used in hundreds of places across the codebase.

peppy:
I'm so lost by this whole change.

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
```diff
@@ -57,6 +69,8 @@ public class OsuClickToResumeCursor : OsuCursor, IKeyBindingHandler<OsuAction>
 
             public Action ResumeRequested;
 
+            public float CursorScale;
```
peppy: this should not be required.

osu.Game.Rulesets.Osu/UI/Cursor/OsuCursorContainer.cs
```diff
@@ -37,9 +45,36 @@ public OsuCursorContainer()
...
+            this.beatmap.BindTo(beatmap);
+            this.beatmap.ValueChanged += _ => calculateScale();
+
+            cursorScale = config.GetBindable<float>(OsuSetting.GameplayCursorSize);
+            cursorScale.ValueChanged += _ => calculateScale();
+
+            autoCursorScale = config.GetBindable<bool>(OsuSetting.AutoCursorSize);
+            autoCursorScale.ValueChanged += _ => calculateScale();
+
+            calculatedCursorScale = new Bindable<float>();
```
peppy: why is this called "calculated"? just call it cursorscale and change the other one to `userCursorScale` or something.

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
```diff
@@ -82,7 +96,7 @@ public bool OnPressed(OsuAction action)
                     case OsuAction.RightButton:
                         if (!IsHovered) return false;
 
-                        this.ScaleTo(new Vector2(2), TRANSITION_TIME, Easing.OutQuint);
+                        this.ScaleTo(2 * CursorScale, TRANSITION_TIME, Easing.OutQuint);
```
peppy: this should not be a thing. make a container to apply the local scale to and set the other component externally.

HoLLy-HaCKeR: Changed it, please check if this is how you wanted it.

smoogipoo:
The scale seems to work on first pause, but it's slightly off in the 2nd pause: https://streamable.com/yquui

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
```diff
@@ -29,16 +29,25 @@ public class OsuResumeOverlay : ResumeOverlay
...
             base.Show();
-            clickToResumeCursor.ShowAt(GameplayCursor.ActiveCursor.Position);
+            cursorScaleContainer.MoveTo(GameplayCursor.ActiveCursor.Position);
+            clickToResumeCursor.Appear();
 
             if (localCursorContainer == null)
+            {
                 Add(localCursorContainer = new OsuCursorContainer());
+
+                localCursorContainer.CursorScale.BindValueChanged(scale => cursorScaleContainer.Scale = new Vector2(scale.NewValue), true);
```
peppy: why is this required? shouldn't the cursor container be handling this itself?

peppy: okay i just misread because this is really not the right way to use bindables. you can't `BindValueChanged` to a bindable owned by another class. make a local one.

HoLLy-HaCKeR:
The cursor being off during the second pause is because the gameplay cursor is expanded, so the pause-to-resume cursor doesn't cover it completely. Is the better solution to take the expanded cursor into account (which means exposing the exposed state + scaling) or to just hide the gameplay cursor?

osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs
HoLLy-HaCKeR: I'm not sure what you mean by that. Do you want me to make a local bindable that's bound to the external one and subscribe to that instead?

bdach: I think local copy bound to external is correct here - as far as I understand subscribing to the external directly is a memory leak once this particular subscriber goes out of scope.

bdach:
I'd be in favour of hiding the old expanded gameplay cursor, unless it was intended by design.

smoogipoo (approved):
seems ok for now

PR #6441, opened 2019-10-08
Separate skin colour configuration decoding into own decoder

### LINKED ISSUE #6433: Beatmap skin configuration decoding potentially conflicts with beatmap configuration
Beatmap skin decoding potentially conflicts with beatmap configuration decoding, this applies for `Version` where on beatmap configuration, it gives the difficulty name of the beatmap, on the other hand for skin configuration, it gives the legacy version for this skin.

Will take a look at it.

frenzibyte:
- Separates skin colour configuration decoding (`ComboColours` and `CustomColours`) into own decoder to avoid conflicting between the two file types (`.osu` files and `skin.ini`)

Closes #6433


peppy:
I'm so confused by this change. You've basically removed the ability to decode anything but colours if I'm not mistaken?

peppy:
Why not just early return after reading the `Version` attribute in the beatmap decoder and call it a day?

iiSaLMaN:
> I'm so confused by this change. You've basically removed the ability to decode anything but colours if I'm not mistaken?

What this change does is decode colours separately because they're only passed to the `SkinConfiguration` (that implements the two interfaces and their only usage `IHasComboColours` and `IHasCustomColours`)

> Why not just early return after reading the `Version` attribute in the beatmap decoder and call it a day?

you meant after reading it in the skin configuration decoder, `LegacyBeatmapSkin` inherits `LegacySkin` and pass the filename of the beatmap as the "skin configuration file", and what the legacy skin decoder does is simply decode every single line of the beatmap (which affects the legacy version of the skin that i'm gonna implement because the beatmap key `Version` (diff name) conflicts with the skin key `Version` (skin version).

peppy:
image

iiSaLMaN:
I can simplify this to just not do anything on any non-colours section (that was my first approach), but I realized it just looks a bit awkward to me (rejected from merge), and the only usage for beatmap in skins are the colours section of it. so I can just split the colours section thing into own decoder and pass it to the legacy skin for decoding it.

peppy:
It's not just colours.

peppy:
If the goal here is to close the attached issue, why could version not just be moved out of the base decoder? I'm still not sure I understand the need for all this refactoring.

- You are now parsing the whole .osu file twice instead of once.
- What's the reason for the `Colours` subclass when it's basically the only thing in `SkinConfiguration`? This is increasing complexity.

I can't see how the related issue even occurs in the first place. `Version` is being parsed in the `LegacyBeatmapDecoder` only, so how does this conflict?

iiSaLMaN:
> I can't see how the related issue even occurs in the first place. Version is being parsed in the LegacyBeatmapDecoder only, so how does this conflict?

As I mentioned in the attached issue, `Version` has a definition in `.osu` files and `skin.ini` files. Confliction occurs when trying to decode beatmap skin by decoding with `LegacySkinDecoder` on `.osu` files. (After `Version` is merged)

Anyways will go with your approach if it's that complex.

peppy:
I'm not asking for an explanation. I'm saying it's convoluted and hard to understand.

iiSaLMaN:
Closing as I'm going with the other approach (filter decoding)

PR #6444, opened 2019-10-09
Add abstract implementation for editor screens which display a timeline

peppy:
To be used by compose/design/timing screens. Transition isn't perfect, but will do for now. Note that going forward each of these screens are going to require slightly different implementations of the timeline display, which is why I opted to do it this way (rather than moving the timeline to a higher level).


bdach (approved):
Net negative lines is always nice to have.

smoogipoo (approved):

PR #6445, opened 2019-10-09
Use empty HitWindows to fix editor slider display

### LINKED ISSUE #5995: Editor : Clicking in timeline object loading issues
**Describe the bug:** 
Moving to somewhere in the song by clicking the timeline (to be more reliable, preferably a moment you've not been yet, seems to be connected to loading of the map) shows all objects that are in between the two locations at once, also causing a massive lag spike from having too many objects loaded at once. Having song on pause makes it clear, a 4 minute map will drop the fps to below 1.

https://gyazo.com/87ab3baedd1943989b8a70cca6e8d571

**osu!lazer version:** 
2019.903.0

peppy:
With a previous change, `HitWindows` was allowed to be null. This is part of a chain effect which caused slider balls to appear in the editor post-seek, even though the end time of the slider was far in the past.

This replaced null `HitWindows` with an empty implementation, containing only `Miss` or `Perfect` (both with zero time values). It also enforces that `HitWindows` have at least two values, and always include miss.

Finally, the clamping of result offsets is now moved to a higher level to provide correct clamping of misses on the `HitResult`, not just the transforms.

Personally I feel this is a code quality improvement, removing nullable `HitWindows` from the equation.

Closes #5995 (i think).

### DIFF
--- a/osu.Game/Rulesets/Objects/HitObject.cs
+++ b/osu.Game/Rulesets/Objects/HitObject.cs
@@ -66,7 +66,6 @@ namespace osu.Game.Rulesets.Objects
         /// <summary>
         /// The hit windows for this <see cref="HitObject"/>.
         /// </summary>
-        [CanBeNull]
         public HitWindows HitWindows { get; set; }
 
         private readonly List<HitObject> nestedHitObjects = new List<HitObject>();
@@ -113,10 +112,7 @@ namespace osu.Game.Rulesets.Objects
             nestedHitObjects.Sort((h1, h2) => h1.StartTime.CompareTo(h2.StartTime));
 
             foreach (var h in nestedHitObjects)
-            {
-                h.HitWindows = HitWindows;
                 h.ApplyDefaults(controlPointInfo, difficulty);
-            }
         }
 
         protected virtual void ApplyDefaultsToSelf(ControlPointInfo controlPointInfo, BeatmapDifficulty difficulty)
@@ -147,7 +143,7 @@ namespace osu.Game.Rulesets.Objects
         /// This will only be invoked if <see cref="HitWindows"/> hasn't been set externally (e.g. from a <see cref="BeatmapConverter{T}"/>.
         /// </para>
         /// </summary>
-        [CanBeNull]
+        [NotNull]
         protected virtual HitWindows CreateHitWindows() => new HitWindows();
     }
 }
--- a/osu.Game/Rulesets/Scoring/HitWindows.cs
+++ b/osu.Game/Rulesets/Scoring/HitWindows.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using osu.Game.Beatmaps;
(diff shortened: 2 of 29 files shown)


osu.Game/Rulesets/Scoring/HitWindows.cs
```diff
@@ -168,6 +182,29 @@ public double WindowFor(HitResult result)
         /// Defaults are provided but can be overridden to customise for a ruleset.
         /// </summary>
         protected virtual DifficultyRange[] GetRanges() => base_ranges;
+
+        public class EmptyHitWindows : HitWindows
+        {
+            private static readonly DifficultyRange[] ranges =
+            {
+                new DifficultyRange(HitResult.Perfect, 0, 0, 0),
```
iiSaLMaN: I think you meant `Great` here?
```suggestion
                new DifficultyRange(HitResult.Great, 0, 0, 0),
```

peppy: I didn't, as I typed Perfect.

Can you explain why Great should be preferred?

osu.Game/Rulesets/Scoring/HitWindows.cs
```diff
@@ -168,6 +182,29 @@ public double WindowFor(HitResult result)
...
+                new DifficultyRange(HitResult.Miss, 0, 0, 0),
+            };
+
+            public override bool IsHitResultAllowed(HitResult result)
+            {
+                switch (result)
+                {
+                    case HitResult.Great:
```
iiSaLMaN: Sorry, I meant this value should be `Perfect`
```suggestion
                    case HitResult.Perfect:
```

frenzibyte (approved):

osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -113,10 +112,7 @@ public void ApplyDefaults(ControlPointInfo controlPointInfo, BeatmapDifficulty d
             nestedHitObjects.Sort((h1, h2) => h1.StartTime.CompareTo(h2.StartTime));
 
             foreach (var h in nestedHitObjects)
-            {
-                h.HitWindows = HitWindows;
```
bdach: Just out of curiosity, why'd the propagation of hit windows to the lower nested objects go away?

bdach (approved):
Does resolve the issue for me too, and always the less `null`s the better in my book 👍

osu.Game/Rulesets/Objects/HitObject.cs
peppy: ah, i forgot to mention that. the behaviour wasn’t well tested or documented, and would result in nested hitobjects that create their own hitwindows having them silently overwritten.

smoogi mentioned this was a memory optimisation for difficulty calculation. we can revisit it if it’s an issue later.

smoogipoo:
I 100% guarantee we will need another factor other than `WindowFor(Miss) != 0`, but I suppose this is fine for now.

smoogipoo (approved):

PR #6449, opened 2019-10-09
Fallback to current skin combo colours if none provided on beatmap skin

frenzibyte:
- Fixes using beatmap skin with no combo colours provides defualt combo colours instead of current skin's.

- To reproduce this issue on current master:

1. Select a skin with custom combo colours.
2. Play a beatmap that doesn't provide combo colours with `beatmap skin` enabled.
3. Result: Combo colours shown are of the default skin's combo colours not the current skin's custom combo colours.


osu.Game.Tests/Resources/skin-empty.ini
```diff
@@ -1,2 +0,0 @@
-[General]
-Name: test skin
```
peppy: what's the reason for removing this?

iiSaLMaN: Doesn't seem to be used, although i've realized it has a usage with something i'm working on, will return.

osu.Game.Tests/Skins/LegacySkinDecoderTest.cs
```diff
@@ -13,28 +13,23 @@ namespace osu.Game.Tests.Skins
     [TestFixture]
     public class LegacySkinDecoderTest
     {
-        [TestCase(true)]
-        [TestCase(false)]
-        public void TestDecodeSkinColours(bool hasColours)
```
peppy: why did you remove the test for no colours?

osu.Game/Skinning/LegacyBeatmapSkin.cs
```diff
@@ -9,6 +9,9 @@ namespace osu.Game.Skinning
 {
     public class LegacyBeatmapSkin : LegacySkin
     {
+        // Null should be returned in the case of no colours provided to fallback into current skin's colours.
```
peppy: this comment is no sense. null what? no colours provided to fallback?

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -63,7 +68,12 @@ protected override void Dispose(bool isDisposing)
                     switch (global)
                     {
                         case GlobalSkinConfiguration.ComboColours:
-                            return SkinUtils.As<TValue>(new Bindable<List<Color4>>(Configuration.ComboColours));
+                            if (Configuration.ComboColours.Any())
```
peppy: you've bypassed `Configuration`. configuration is public. now you can get two values for combo colours depending on how you access.

osu.Game/Skinning/LegacyBeatmapSkin.cs
iiSaLMaN: Was afraid of that (my explanation is shit). null should be returned if the beatmap skin doesn't provide any combo colours when trying to retrieve with `GetConfig(...)`, this will make `SkinProvidingContainer` retrieve from the current skin's (`SkinProvidingContainer.fallbackSource`) combo colours.

osu.Game.Tests/Skins/LegacySkinDecoderTest.cs
iiSaLMaN: As told in the description of that commit, checked differently with `TestSceneSkinConfigurationLookup.TestEmptyComboColours(false)`

peppy: no reason to not leave the test here?

iiSaLMaN: it'll fail as the combo colours are gonna be empty? unless it's fine for it to be there

osu.Game/Skinning/LegacySkin.cs
iiSaLMaN: I fail to understand what you mean, care to explain further?

osu.Game.Tests/Skins/LegacySkinDecoderTest.cs
peppy: i really don't want to have to spell it out, but it seems i have to:

update the test asserts to match the *expected* output.

osu.Game/Skinning/LegacySkin.cs
peppy: `skin.Configuration.ComboColours` is going to return different results to `skin.GetConfig(ComboColours)`

iiSaLMaN: I'm not sure if this is causing issues though? if the combo colours retrieved is null (in DHO), the white colour will be used. what I'm trying to say is that the correct way to use this is by using a `SkinProvidingContainer`.

bdach: Having two ways of accessing the same value (or at least what *looks like* it should be the same value, as this does) return two different results is bug-prone by default and a violation of encapsulation. We don't want to have to remember in the future which way of accessing it is *right*. There should be only one way of accessing a particular value.

iiSaLMaN: are you talking about `skin.Configuration.ComboColours` returning empty list and `GetConfig(ComboColours` returning null?

osu.Game/Beatmaps/Formats/IHasComboColours.cs
```diff
@@ -8,6 +8,19 @@ namespace osu.Game.Beatmaps.Formats
...
+
+        /// <summary>
+        /// Adds combo colours to the list.
+        /// </summary>
+        void AddComboColours(params Color4[] colours);
+
+        /// <summary>
+        /// Clear current combo colours from the list.
+        /// </summary>
+        void ClearComboColours();
```
peppy: when would we every want to clear combo colours?

osu.Game/Beatmaps/Formats/IHasComboColours.cs
```diff
@@ -8,6 +8,19 @@ namespace osu.Game.Beatmaps.Formats
...
-        List<Color4> ComboColours { get; set; }
+        /// <summary>
+        /// Retrieves the list of combo colours for presentation only.
+        /// </summary>
+        IReadOnlyList<Color4> ComboColours { get; set; }
+
+        /// <summary>
+        /// Adds combo colours to the list.
+        /// </summary>
+        void AddComboColours(params Color4[] colours);
```
peppy: why was this added? i'm looking for reason for the added complexity in this PR.

osu.Game/Skinning/SkinConfiguration.cs
```diff
@@ -14,7 +16,39 @@ public class SkinConfiguration : IHasComboColours, IHasCustomColours
...
+
+                if (AllowDefaultComboColoursFallback)
+                    return DefaultComboColours;
+
+                return null;
+            }
+            set => comboColours = value.ToList();
+        }
+
+        public void AddComboColours(params Color4[] colours) => colours.ForEach(c => comboColours.Add(c));
```
peppy: if this is necessary, `AddRange` like before?

peppy (changes requested):

osu.Game/Beatmaps/Formats/IHasComboColours.cs
iiSaLMaN: To allow passing default combo colours when trying to read via `ComboColours`, and avoid having 2 different values like this PR used to be: (`GetConfig(ComboColours)`, `SkinConfiguration.ComboColours`).

osu.Game/Beatmaps/Formats/IHasComboColours.cs
iiSaLMaN: Wasn't sure about leaving an add function without remove/clear function.., will remove.

osu.Game/Skinning/SkinConfiguration.cs
iiSaLMaN: Right.. 🤦‍♂️

osu.Game.Tests/Skins/LegacySkinDecoderTest.cs
```diff
@@ -15,14 +15,24 @@ public class LegacySkinDecoderTest
     {
         [TestCase(true)]
         [TestCase(false)]
-        public void TestDecodeSkinColours(bool hasColours)
+        [TestCase(false, false)]
+        public void TestDecodeSkinColours(bool hasColours, bool canFallback = true)
```
smoogipoo: Please split out into multiple tests (at least the `canFallback` part). Each method should test at most one case.

smoogipoo (changes requested):
This doesn't seem like a bad change.

osu.Game.Tests/Skins/LegacySkinDecoderTest.cs
iiSaLMaN: Does this mean to not always use `TestCase(val, val)` in future changes?

smoogipoo: In general, yes, they should be avoided. The typical use case is having expected values and such, e.g.:

`TestMethod(bool arg1, bool expectedResult)`

or similar. Having them as args that affect the test case increases the human parsing complexity a lot.

peppy:
This is missing a test for the case you're trying to resolve (ie. there is no beatmap test).

iiSaLMaN:
Wrote a beatmap test for `LegacyBeatmapSkin` that retrieves combo colours of the `Player`'s `BeatmapSkinProvidingContainer` to assert with, a bit complicated but hopefully good to go.

peppy (approved):

peppy:
@iiSaLMaN still required a considerable amount of tidy-up. please check my commits, especially https://github.com/ppy/osu/pull/6449/commits/783159f7a06de9438552f0de2b9654662b78b6ac and https://github.com/ppy/osu/pull/6449/commits/145ac8e0b547c3f8472f4a0344dec7b6e3a4730a.

PR #6450, opened 2019-10-09
Add access to numerical version of legacy skin

frenzibyte:
- Adds access to the numerical legacy version of legacy skins, allows for applying behaviour on specific legacy skins, for example expanding number piece of the hit circle if the skin version is older than v2.0
- One thing to note here is that the legacy configuration lookup enum name ends with an `s` intendedly, as it cannot be singular because it conflicts with the component name.


osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace osu.Game.Skinning
+{
+    public class LegacySkinConfiguration : DefaultSkinConfiguration
+    {
+        public const double LATEST_VERSION = 2.5;
+
+        /// <summary>
+        /// Legacy version of this skin.
+        /// </summary>
+        public double? LegacyVersion { get; internal set; }
```
bdach: `double` seems like an unfortunate data type choice for me here, worried about floating-point shenanigans. I guess either `decimal`, `string` or a proper `struct` with major-minor fields would be better choices if I were to suggest something else.

iiSaLMaN: I think it's fine to use `decimal` for now as this is a legacy version.

peppy:
Note that a user skin must [always force latest version too](https://osu.ppy.sh/help/wiki/Skinning/skin.ini).

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Skinning
+{
+    public class LegacySkinConfiguration : DefaultSkinConfiguration
+    {
+        public const decimal LATEST_VERSION = 2.5m;
```
peppy: should be 2.7

image

osu.Game/Skinning/LegacySkinDecoder.cs
```diff
@@ -32,6 +32,14 @@ protected override void ParseLine(DefaultSkinConfiguration skin, Section section
                             case @"Author":
                                 skin.SkinInfo.Creator = pair.Value;
                                 return;
+
+                            case @"Version":
+                                if (pair.Value == "latest" || pair.Value == "User")
```
peppy: Please re-read the wiki. "User" is *not* a valid version. It would be based off the skin's name.

osu.Game/Skinning/LegacySkinDecoder.cs
```diff
@@ -32,6 +32,14 @@ protected override void ParseLine(DefaultSkinConfiguration skin, Section section
...
+                                    skin.LegacyVersion = LegacySkinConfiguration.LATEST_VERSION;
+                                else if (decimal.TryParse(pair.Value, out var version))
```
peppy: missing culture specification

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+
+        /// <summary>
+        /// Legacy version of this skin.
+        /// </summary>
+        public decimal? LegacyVersion { get; internal set; }
+    }
+
+    public enum LegacySkinConfigurations
```
peppy: I'd sooner rename the `LegacySkinConfiguration` class than do this weirdness in naming.

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Skinning
+{
+    public class LegacySkinConfiguration : DefaultSkinConfiguration
+    {
+        public const decimal LATEST_VERSION = 2.5m;
+
+        /// <summary>
+        /// Legacy version of this skin.
```
peppy: what does null mean?

peppy (changes requested):

osu.Game/Skinning/LegacySkinConfiguration.cs
iiSaLMaN: `StableSkinConfiguration`..?

peppy: no, synonyms to avoid naming clashes are invalid

osu.Game/Skinning/LegacySkinConfiguration.cs
iiSaLMaN: that no version was set in the decoder, that leads into returning a `null` in `GetConfig(Version)`, as the user is forced into putting a default value in case the version retrieved is null.

```csharp
skin.GetConfig<LegacySkinConfigurations, decimal>(LegacySkinConfigurations.Version)?.Value ?? 1;
```

osu.Game/Skinning/LegacySkinConfiguration.cs
iiSaLMaN: Oops. I've read on the wiki that the latest version is `2.5`, should probably be updated.

osu.Game/Skinning/LegacySkinDecoder.cs
iiSaLMaN: I'm gonna support that later actually, seems a bit complicated to implement...

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -71,6 +73,18 @@ protected override void Dispose(bool isDisposing)
                 case GlobalSkinColour colour:
                     return SkinUtils.As<TValue>(getCustomColour(colour.ToString()));
 
+                case LegacySkinConfigurations legacy:
+                    switch (legacy)
+                    {
+                        case LegacySkinConfigurations.Version:
+                            if (Configuration.LegacyVersion.HasValue)
+                                return SkinUtils.As<TValue>(new Bindable<decimal>(Configuration.LegacyVersion.Value));
```
huoyaoyuan: Note: I'm preparing a pull request to ban `HasValue` and `Value` in framework side. Consider to use pattern matching because it has slightly better performance, and good readability.

osu.Game/Skinning/LegacySkinConfiguration.cs
iiSaLMaN: how about renaming the whole configuration enums to `Property`; 
 - `LegacySkinConfigurations` -> `LegacySkinProperty`
 - `GlobalSkinConfiguration` -> `GlobalSkinProperty` | `SkinProperty`

osu.Game/Skinning/LegacySkin.cs
huoyaoyuan: Well I meant 'is decimal version'... If you are not familiar with this I'll cover this in later PR by finding all usages.

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Skinning
+{
+    public class LegacySkinConfiguration : DefaultSkinConfiguration
+    {
+        public const decimal LATEST_VERSION = 2.7m;
+
+        /// <summary>
+        /// Legacy version of this skin. Null if no version was set to allow fallback to a parent skin version.
```
peppy: what parent?

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -74,6 +80,18 @@ protected override void Dispose(bool isDisposing)
                 case GlobalSkinColour colour:
                     return SkinUtils.As<TValue>(getCustomColour(colour.ToString()));
 
+                case LegacySkinConfiguration.LegacySetting legacy:
+                    switch (legacy)
+                    {
+                        case LegacySkinConfiguration.LegacySetting.Version:
```
peppy: not used or tested anywhere

peppy (dismissed):
as suggested

osu.Game/Skinning/LegacySkinConfiguration.cs
iiSaLMaN: What about this?
```csharp
/// <summary>
/// Legacy version of this skin. Null if no version was set to allow <see cref="SkinProvidingContainer"/> fallback system.
/// </summary>

peppy: At that point I don't even think it is necessary to mention this.

smoogipoo (approved):
Seems ok

PR #6451, opened 2019-10-10
Fix audio playback position being reset after resuming to song select

### LINKED ISSUE #4684: When exiting a play, track position seeks to preview point instead of staying

peppy:
- Closes #4684.
- Adds test (this is not going to regress again).
- [x] Depends on osu-resources bump


smoogipoo:
After changing the beatmap set, the track should be unpaused.

Joehuu:
That was an issue before this PR was created. Another regression from https://github.com/ppy/osu/pull/6423.

peppy:
fix on the way

smoogipoo (approved):

PR #6457, opened 2019-10-10
Implement CommentsHeader component

EVAST9919:


osu.Game/Overlays/Comments/SortSelector.cs
```diff
@@ -0,0 +1,106 @@
...
+
+                protected override void OnHoverLost(HoverLostEvent e)
+                {
+                    if (!Active.Value) base.OnHoverLost(e);
+                }
+            }
+        }
+    }
+
+    public enum SortCommentsBy
```
iiSaLMaN: this name seems inconsistent to me, what about `CommentSortCriteria` or just `SortCriteria`

frenzibyte (changes requested):

osu.Game/osu.Game.csproj
```diff
@@ -32,4 +32,7 @@
     <PackageReference Include="SharpRaven" Version="2.4.0" />
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.6.0" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Overlays\Comments\" />
+  </ItemGroup>
```
bdach: There doesn't seem to be a need for this to land in the `.csproj`.

bdach (commented):
Visual reference:

image

Looks to be in line with web design and working properly - I just have one comment on the `.csproj` change which seems to be unnecessary.

osu.Game/osu.Game.csproj
EVAST9919: It was an auto addition, will remove

bdach (dismissed):
Looks good to go as far as I can tell 👍

peppy:
Just as a note, there's basically zero point in making constants for values which only appear once. The exception is if it's not easy to know what the constant *is*, but in all the cases in your new class, they are used next to a keyword which defines them:

image

peppy:
@EVAST9919 i've made several refactors to improve code quality, please make sure to review them.

peppy (approved):

PR #6458, opened 2019-10-10
Update to .NET core 3.0

peppy:


osu.Game.Tests/Beatmaps/IO/LineBufferedReaderTest.cs
```diff
@@ -113,19 +113,14 @@ public void TestReadToEndNoPeeks()
...
 
             using (var stream = new MemoryStream(Encoding.UTF8.GetBytes(contents)))
             using (var bufferedReader = new LineBufferedReader(stream))
             {
                 Assert.AreEqual("this line is gone", bufferedReader.ReadLine());
                 Assert.AreEqual("this one shouldn't be", bufferedReader.PeekLine());
-                const string ending = @"this one shouldn't be
-these ones
-definitely not";
+                const string ending = "this one shouldn't be\nthese ones\ndefinitely not";
```
smoogipoo: Needs investigation as to why `ReadToEnd()` is providing a different output to the input above.

peppy: @bdach if you beat me to it, wanna take a look into this? not sure if it's a regression in .NET core 3.0 or an expected change (seems that `\r\n` are being replaced with `\n` on macOS, at least).

bdach: Sure, I'll have a look at it as the first thing today.

peppy: Looks like this is also reproducing on master, so may have just never been run/tested on a non-windows host.

smoogipoo (approved):

peppy:
Going to merge this one even with the random test failure (will fix tomorrow). I'd like to test the deploy process and see how we fare with iOS/android crashes.

PR #6461, opened 2019-10-10
Fix platform dependency in buffered reader test

bdach:
Fixes `BufferedLineReader` test fails on non-Windows platforms.

# Summary

Tests for the line-buffered reader added in 7b1ff38 were subtly dependent on the execution environment due to differing end-of-line markers on Windows and Unix-based systems.

Because `StreamReader` discards all newlines when reading line-by-line, `LineBufferedReader` used a `StringBuilder` to patch the peeked lines back together with the remaining contents of the file being read. As `StringBuilder.AppendLine` uses the environment-specific newline delimiter, the delimiters after the peeked-but-unconsumed lines can therefore be substituted by the platform-specific variants, causing the test failures due to the overly-simplified way they were written.

Reformulate the test to avoid such issues from resurfacing again by splitting lines by `\r` or `\n` and then testing each line individually. Additionally remove all raw literals in favour of explicitly mixing various line delimiter character sequences for additional coverage.

# Remarks

To avoid this sort of issue from happening again in the future I will be adjusting my workflow to include Linux builds as well as Windows ones wherever necessary. To that end I ran an appveyor build on the `Ubuntu` image as well as a normal Windows one:

* [Windows build test run](https://ci.appveyor.com/project/bdach/osu/builds/28017980/tests)
* [Linux build test run](https://ci.appveyor.com/project/bdach/osu/builds/28019215/tests)

Unfortunately the whole Linux build isn't green due to visual tests failing which I don't want to start fixing right now (likely the image doesn't have a graphics device enabled), but all tests from `LineBufferedReaderTest` are green on both platforms.


peppy (approved):

smoogipoo:
I don't like that `LineBufferedReader` is mangling the line endings itself. Have you considered seeking the base stream back after peeking a line?

smoogipoo:
I guess it's fine if there's no way to preserve line endings...

PR #6464, opened 2019-10-10
Make previous track button restart current track on first click

### LINKED ISSUE #6335: Play from beginning hotkey is missing

GSculerlor:
Fix #6335 supersedes #6353 

Following up the [voting result](https://twitter.com/ppy/status/1181773682898661377), this PR will change the previous button behavior. So when current track's time is below 5 seconds it'll change the track to previous track. And if current track's time is above 5 second it'll seek the track to 0 (restart the track). 

5 seconds is chosen based on my observation on some music player (like Groove Music, Spotify, etc.). Using `TestSceneNowPlayingOverlay` for testing (actually I placed it there to make easier to check using `NowPlayingOverlay`)

**Proof of concept:** [test](https://streamable.com/e2na3) **|** [in-game](https://streamable.com/p98dc)


osu.Game/Overlays/MusicController.cs
```diff
@@ -136,11 +136,20 @@ public bool TogglePause()
...
         public bool PrevTrack()
         {
+            var currentTrackPosition = current?.Track.CurrentTime;
+
+            if (currentTrackPosition >= 5000)
+            {
+                SeekTo(0);
+
+                return true;
+            }
```
iiSaLMaN: I think this logic should go into the `case GlobalAction.MusicPrev` as it seems out of scope for what this function should do. (play previous track)

osu.Game/Overlays/MusicController.cs
```diff
@@ -136,11 +136,20 @@ public bool TogglePause()
...
         /// <summary>
-        /// Play the previous track.
+        /// Play the previous track or restart the current track if it's current time below 5000ms
         /// </summary>
         /// <returns>Whether the operation was successful.</returns>
         public bool PrevTrack()
         {
+            var currentTrackPosition = current?.Track.CurrentTime;
+
+            if (currentTrackPosition >= 5000)
```
iiSaLMaN: the number of milliseconds to allow playing the previous track should probably be a `double` constant.

frenzibyte (changes requested):
I think the tests will fail if there is one beatmap available.

osu.Game/Overlays/MusicController.cs
bdach: It's not that simple, this logic is shared with `NowPlayingOverlay`, whose previous button should probably behave the same way.

osu.Game/Overlays/MusicController.cs
bdach: Yeah, I'd extract something like a `RestartCutoffPoint` constant here. Additionally this check should not be duplicated verbatim twice.

I'd extract a private method at the least (`bool ShouldReset(Track track)`), or, more controversially, introduce a tri-value inner enum:
```csharp
enum PrevActionResult
{
    None,
    Restart,
    Previous
}
```
and switch on that in `OnPressed` to get the appropriate message.

osu.Game/Overlays/MusicController.cs
```diff
@@ -136,11 +136,20 @@ public bool TogglePause()
         }
 
         /// <summary>
-        /// Play the previous track.
+        /// Play the previous track or restart the current track if it's current time below 5000ms
```
bdach: The xmldoc should not refer to a concrete value like this. Once you refactor out the constant you can `<see cref="RestartCutoffPoint" />` it.

bdach (changes requested):

osu.Game.Tests/Visual/UserInterface/TestSceneNowPlayingOverlay.cs
```diff
@@ -49,9 +49,10 @@ public void TestPrevTrackBehavior()
             AddStep(@"Call PrevTrack", () => musicController.PrevTrack());
             AddAssert(@"Check if it restarted", () => currentTrack == Beatmap.Value);
 
-            AddStep(@"Seek track to 1 second", () => musicController.SeekTo(1000));
+            AddStep(@"Seek track to 2 second", () => musicController.SeekTo(2000));
             AddStep(@"Call PrevTrack", () => musicController.PrevTrack());
-            AddAssert(@"Check if it changed to prev track'", () => currentTrack != Beatmap.Value);
+            // If the track isn't changing, check the current track's time instead
+            AddAssert(@"Check if it changed to prev track'", () => currentTrack != Beatmap.Value || currentTrack.Track.CurrentTime < 2000);
```
smoogipoo: Why would the track not change?

GSculerlor: If the available track is only one, it's basically just play the same track and restart it (both go to previous or go to next track)

smoogipoo: That needs to be split into multiple tests. As it is this test is potentially bypassing expected behaviour.

bdach: The presence of more than one beatmap/track might not be guaranteed in a test environment, possibly we might want to ensure it by manually importing before the test run?

osu.Game/Overlays/MusicController.cs
```diff
@@ -135,13 +137,26 @@ public bool TogglePause()
             return true;
         }
 
+        private PreviousButtonAction? prevAction;
```
bdach: This should not be a field. The point of introducing this was to return an enum instead of a bool in `PrevTrack`.

osu.Game/Overlays/MusicController.cs
```diff
@@ -261,7 +276,7 @@ public bool OnPressed(GlobalAction action)
 
                 case GlobalAction.MusicPrev:
                     if (PrevTrack())
-                        onScreenDisplay?.Display(new MusicControllerToast("Previous track"));
+                        onScreenDisplay?.Display(new MusicControllerToast(prevAction == PreviousButtonAction.Restart ? "Restart track" : "Previous track"));
```
bdach: I'd honestly rather have this as a `switch` after you change the return value.

bdach (dismissed):

osu.Game.Tests/Visual/UserInterface/TestSceneNowPlayingOverlay.cs
GSculerlor: Is there any testcase example that I can follow to encapsulate the beatmap list per test?

bdach: I suppose [`ImportBeatmapTest`](https://github.com/ppy/osu/blob/master/osu.Game.Tests/Beatmaps/IO/ImportBeatmapTest.cs) has the closest thing to an example of an import in tests, although I'm not entirely certain whether we want to go down that route or not.

osu.Game/Overlays/MusicController.cs
```diff
@@ -260,10 +270,19 @@ public bool OnPressed(GlobalAction action)
...
+                        case PreviousButtonAction.Restart:
+                            onScreenDisplay?.Display(new MusicControllerToast("Restart track"));
+                            return true;
+
+                        case PreviousButtonAction.Previous:
+                            onScreenDisplay?.Display(new MusicControllerToast("Previous track"));
+                            return true;
+
+                        default:
+                            return false;
```
bdach: Should return `true` instead - the return value indicates whether the action is handled or not, and it is, even if it does nothing (see the neighboring cases - `return true` is done unconditionally, disregarding whether the hotkeys did anything or not). Therefore I'd probably move the `return`s outside, and not even handle `default` explicitly.

osu.Game/Overlays/MusicController.cs
```diff
@@ -136,11 +138,19 @@ public bool TogglePause()
         }
 
         /// <summary>
-        /// Play the previous track.
+        /// Play the previous track or restart the current track if it's current time below <see cref="restart_cutoff_point"/>
         /// </summary>
         /// <returns>Whether the operation was successful.</returns>
```
peppy: returns something different.

peppy (changes requested):

peppy:
Adjusted tests, xmldoc and naming, seems good otherwise.

peppy (dismissed):

peppy:
As an update, this needs more investigation into why tests are failing. Cannot repro locally.

bdach:
I managed to reproduce the failure locally inside Rider using the NUnit runner. It's not consistent (fails about 60% of the time). Doing this seems to help:
```diff
@@ -55,6 +56,7 @@ public void TestPrevTrackBehavior()
             AddStep(@"Seek track to 6 second", () => musicController.SeekTo(6000));
             AddUntilStep(@"Wait for current time to update", () => currentBeatmap.Track.CurrentTime > 5000);
             AddAssert(@"Check action is restart track", () => musicController.PreviousTrack() == PreviousTrackResult.Restart);
+            AddUntilStep(@"Wait for current time to update", () => Precision.AlmostEquals(currentBeatmap.Track.CurrentTime, 0));
             AddAssert(@"Check track didn't change", () => currentBeatmap == Beatmap.Value);
             AddAssert(@"Check action is not restart", () => musicController.PreviousTrack() != PreviousTrackResult.Restart);
         }
```
`Precision.AlmostEquals` used to silence IDE errors about floating-point equality comparison. I can push that in here, if you think that's a correct fix, or any of you can apply it, not too fussed.

peppy:
@bdach go ahead and push that, it seems correct

bdach:
Done, should be all ready now - I suppose I can't approve now, though.

bdach:
Necroing this one since it seemed to be pretty ready to merge and just fell through the cracks.

smoogipoo (approved):
lgtm

PR #6465, opened 2019-10-10
Filter out OS-generated files from ZIP archives

bdach:
Follow-up of sorts to #6170.

# Summary

Add a filename ignore list to `ZipArchiveReader` to filter out superfluous OS-generated files from archives during the import process. In addition to decreasing the size of files imported this allows imports of some incorrectly-constructed archives. An example is the case of having a `__MACOSX` directory next to a single directory with the actual files - filtering out the former at `ZipArchiveReader` allows the fallback added in #6170 to work.

Comes with a test demonstrating the reproduction case, and some slight refactorings of the neighbouring test code, because I like `nameof`s.

# Remarks

This edge case was spotted "in the wild" while I was helping out @X4X5 with a skin import on ChromeOS through the Linux shell. Aside from that actual case I added some other files that don't really need to be imported ever (`.DS_Store` files for Macs and their ugly Windows analogues, the infamous `Thumbs.db`). Suggestions for more welcome, if you can think of any.

I was thinking a little about adding this to `LegacyFileArchiveReader` too or even moving the filtering higher up to `ArchiveReader` itself, but I'm not entirely convinced it's a good idea.


osu.Game/IO/Archives/ZipArchiveReader.cs
```diff
@@ -43,7 +58,9 @@ public override void Dispose()
             archiveStream.Dispose();
         }
 
-        public override IEnumerable<string> Filenames => archive.Entries.Select(e => e.Key).ToArray();
+        private static bool canBeIgnored(IEntry entry) => filename_ignore_list.Any(ignoredName => entry.Key.IndexOf(ignoredName, StringComparison.InvariantCultureIgnoreCase) >= 0);
```
peppy: Just something to be aware of, but this may match `test/MyMapBackground.DS_Store.jpg`. It's probably never going to be an issue, but could be fixed by matching the end of the string or otherwise.

smoogipoo:
```suggestion
        private static bool canBeIgnored(IEntry entry) => filename_ignore_list.Any(ignoredName => entry.Key.IndexOf(ignoredName, StringComparison.OrdinalIgnoreCase) >= 0);
```

Culture is unnecessary, has theoretical improvements in performance.

smoogipoo (dismissed):

osu.Game/IO/Archives/ZipArchiveReader.cs
bdach: Yeah, I felt like I was skimping on doing the checks properly, as I didn't want to split into two lists (one for full path and one for filename only). I think I'm going to go back and make it safer so that it can't ignore files wrongly.

peppy: I don't think the path needs to be considered, can just do an `EndsWith` or similar (and optionally a check for the path separator). We be ignoring these files no matter what path they are in.

bdach: Sure, but the `__MACOSX` folder might require a different way of handling it entirely as it can show in the middle of the path (and I think ZIP folder entries are separate from the entries of files contained within, if not entirely nonexistent), that's why the split into two ignore lists would be necessary. I'd have to actually check to be certain, though.

peppy: Aaah, I forgot this can be a folder.

To be honest, I think it's fine to just merge as-is. Not worth spending extra effort on.

smoogipoo (approved):

PR #6473, opened 2019-10-11
Implement abstract class and structure for editor grids

smoogipoo:
Abstract class + tests implementation.


bdach (approved):
Looks cool in visual tests, the test scenarios seem to be behaving correctly, code seems good too 👍

PR #6474, opened 2019-10-11
Implement osu!'s beat snapping grid

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/6473

Same as the PR - implementation + tests. Not integrated yet - integration is not far off but there's a few further prerequisites/cleanups to do.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneOsuBeatSnapGrid.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneOsuBeatSnapGrid.cs
@@ -5,9 +5,11 @@ using System;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.UserInterface;
 using osu.Framework.Input.Events;
+using osu.Framework.MathUtils;
 using osu.Game.Beatmaps;
 using osu.Game.Beatmaps.ControlPoints;
 using osu.Game.Rulesets.Objects;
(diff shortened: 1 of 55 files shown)


osu.Game.Rulesets.Osu.Tests/TestSceneOsuBeatSnapGrid.cs
```diff
@@ -0,0 +1,210 @@
...
+        [TestCase(1.5f, 150)]
+        public void TestSliderMultiplier(float multiplier, float expectedSpacing)
+        {
+            AddStep($"set speed multiplier = {multiplier}", () => editorBeatmap.BeatmapInfo.BaseDifficulty.SliderMultiplier = multiplier);
+            createGrid();
+            AddAssert($"spacing = {expectedSpacing}", () => Precision.AlmostEquals(expectedSpacing, grid.DistanceSpacing));
+        }
+
+        [Test]
+        public void TestCursorInCentre()
```
bdach: This test is giving me a consistent crash locally in the visual test runner on the last step (same stack trace as the CI build is reporting for it), which is weird, as all of the others run fine to the end.

bdach (commented):
I like how this is panning out, works really nice. I have some suggestions/notes in comments, and some general notes that are probably outside of the scope of this - you might be aware of them or not, but I thought I'd want to mention them:

- I feel that for the finer spacings (like on `TestBeatDivisor(16)`) the usefulness of the grid is greatly diminished by the density of the circular rulers. I'd consider some sort of hover behaviour for it to be easier to tell visually which circle is being snapped to.
- The rulers seem to be aliased badly, which is quite jarring - see magnified image below:
  image

bdach:
A random thought I had on the problem of denser grids not really helping much for snapping was applying a notion of [hysteresis](https://en.wikipedia.org/wiki/Hysteresis#User_interface_design) to the grid lines. I'm not sure if I'll be able to explain clearly how this applies, but here goes.

The rulers could be designed such that the snapping range of the currently active ruler is not exactly half of the distance between rulers, but a little more than that (so like 75%). To switch to snap to another ruler you'd have to get closer than halfway (following the number above, within 25% of the distance to it), at which point it becomes the active ruler and has the wider margin from that point on. Could even combine that with a hover effect that that makes the active ruler glow and show its snapping range for clear UX.

This is a suggestion that I haven't tried at all and most likely out of scope of this, I guess we'd have to try it and see if it feels ok.

smoogipoo:
Hysteresis is a little bit out of scope for now, just want to get things working and usable for now.

smoogipoo:
> The rulers seem to be aliased badly, which is quite jarring

Can't really address this one easily, but it would be nice to fix. It's a framework-side fix to a component that doesn't really support AA in the form we usually do.

bdach (dismissed):
Looks ready to go as far as I can tell :shipit:

peppy (approved):

PR #6504, opened 2019-10-14
Implement CommentsContainer component

EVAST9919:
Still no sub-comments loading because it's already a quite big pull request


osu.Game/Graphics/UserInterface/ShowMoreButton.cs
```diff
@@ -126,19 +131,21 @@ private class ChevronIcon : SpriteIcon
         {
             private const int icon_size = 8;
 
+            private Color4 accentColour;
+
+            public Color4 AccentColour
+            {
+                get => accentColour;
+                set => accentColour = Colour = value;
+            }
```
bdach: Why does this exist? It seems you wanted an alias for `Colour`, but if so, this is wrong due to the fact you can do the following:
```csharp
spriteIcon.AccentColour = Color4.Black;
spriteIcon.Colour = Color4.White;
// now the two properties return different values
```

osu.Game/Online/API/Requests/GetCommentsRequest.cs
```diff
@@ -0,0 +1,47 @@
...
+            return req;
+        }
+
+        protected override string Target => "comments";
+    }
+
+    public enum CommentableType
+    {
+        Build,
+        Beatmapset,
```
bdach: `BeatmapSet` for consistent casing.

osu.Game.Tests/Visual/Online/TestSceneShowMoreButton.cs
```diff
@@ -17,11 +18,11 @@ public class TestSceneShowMoreButton : OsuTestScene
 
         public TestSceneShowMoreButton()
```
bdach: `ShowMoreButton` was renamed, but the test scene wasn't.

osu.Game/Online/API/Requests/Responses/APICommentsController.cs
```diff
@@ -0,0 +1,80 @@
...
+                        comments.ForEach(parent =>
+                        {
+                            if (parent.Id == child.ParentId)
+                            {
+                                parent.ChildComments.Add(child);
+                                child.ParentComment = parent;
+                            }
+                        });
+                    }
+                });
```
bdach: I think you could take this down from being quadratic to at least quasilinear (`n log n`) with some dictionaries. And this could matter for long comment threads, too.

osu.Game/Online/API/Requests/Responses/APICommentsController.cs
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    Comments.ForEach(c =>
+                    {
+                        if (c.UserId == u.Id)
+                            c.User = u;
+
+                        if (c.EditedById == u.Id)
+                            c.EditedUser = u;
+                    });
+                });
```
bdach: Same comment about performance/time complexity as above applies here. This is *sort of* better in that it will regress to quadratic only in threads where there are as many users as comments, but still could be faster.

osu.Game/Graphics/UserInterface/ShowMoreButton.cs
```diff
@@ -1,30 +1,36 @@
-// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
```
bdach: What's happened here? This looks like a newline change, I'd double-check your `git` settings.

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -0,0 +1,115 @@
...
+        public DateTimeOffset? EditedAt { get; set; }
+
+        [JsonProperty(@"edited_by_id")]
+        public long? EditedById { get; set; }
+
+        public User EditedUser { get; set; }
+
+        public bool IsTopLevel { get; set; }
+
+        public bool IsDeleted { get; set; }
```
bdach: This shouldn't be settable, just delegate to the actual property:
```csharp
public bool IsDeleted => DeletedAt.HasValue;
```
Avoids potential update bugs.

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -0,0 +1,115 @@
...
+
+        public string GetMessage()
+        {
+            if (IsDeleted)
+                return @"deleted";
+
+            return WebUtility.HtmlDecode(Regex.Replace(MessageHtml, @"<(.|\n)*?>", string.Empty));
+        }
+
+        public int GetDeletedChildsCount()
```
bdach: `GetDeletedChildrenCount()`

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -0,0 +1,115 @@
...
+            int count = 0;
+
+            if (ChildComments.Any())
+                ChildComments.ForEach(child =>
+                {
+                    if (child.IsDeleted)
+                        count++;
+                });
+
+            return count;
```
bdach: I'm not sure this needs to be a manual count loop instead of a linq expression.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,203 @@
...
+        private void getComments(bool initial = true)
+        {
+            if (initial)
+            {
+                currentPage = 1;
+                loadedTopLevelComments = 0;
+                deletedChildsPlaceholder.DeletedCount.Value = 0;
+                moreButton.IsLoading = true;
+                content.Clear();
+            }
```
bdach: Does this need to be in here behind a flag, instead of somewhere where setup/loading is normally done? (be it ctor, or `Load()`)?

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,203 @@
...
+        [Resolved]
+        private IAPIProvider api { get; set; }
+
+        [Resolved]
+        private OsuColour colours { get; set; }
+
+        private GetCommentsRequest request;
+        private CancellationTokenSource loadCancellation;
+        private int currentPage;
+        private int loadedTopLevelComments;
```
bdach: I'm not sure the performance benefit of having this in a field outweighs the cost of potential bugs when the state of this gets polluted.

osu.Game/Overlays/Comments/DeletedChildsPlaceholder.cs
```diff
@@ -0,0 +1,74 @@
...
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics;
+using osu.Game.Graphics;
+using osu.Framework.Graphics.Sprites;
+using osuTK;
+using osu.Framework.Bindables;
+
+namespace osu.Game.Overlays.Comments
+{
+    public class DeletedChildsPlaceholder : FillFlowContainer
```
bdach: `DeletedChildrenPlaceholder`

osu.Game/Overlays/Comments/DeletedChildsPlaceholder.cs
```diff
@@ -0,0 +1,74 @@
...
+            if (!canBeShown)
+            {
+                Hide();
+                return;
+            }
+
+            string str = $@"{count.NewValue} deleted comment";
+
+            if (!(count.NewValue.ToString().EndsWith("1") && !count.NewValue.ToString().EndsWith("11")))
+                str += "s";
```
bdach: This shouldn't be done manually, [use Humanizer's `.ToQuantity` instead](https://github.com/Humanizr/Humanizer#toquantity).

bdach (dismissed):
Honestly this would greatly benefit from splitting, and there are multiple things that I believe need to be fixed - and I had just a rudimentary look without too much digging.

osu.Game.Tests/Visual/Online/TestSceneShowMoreButton.cs
EVAST9919: It wasn't renamed. It's still there, but removed to another namespace

bdach: OK, that's my bad, but in that case why does the test use `ProfileShowMoreButton` exclusively? Kind of confusing naming here.

osu.Game/Online/API/Requests/GetCommentsRequest.cs
EVAST9919: Can't do this due to using `Underscore` function (used to convert `NewsPost` to `news_post`)

osu.Game.Tests/Visual/Online/TestSceneShowMoreButton.cs
EVAST9919: Will use TestButton or something

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: Are you talking about `loadedTopLevelComments` line?

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: May be it's worth it to move that stuff to it's own function

osu.Game/Overlays/Comments/CommentsContainer.cs
bdach: Yeah, I'd honestly ideally want to get rid of that field, it seems like it'll cause problems down the line. Would rather iterate over the comments every time and be sure the count is correct than risk potential bugs arising from forgetting to update it.

osu.Game/Overlays/Comments/CommentsContainer.cs
bdach: Sure, it'd be a start at least.

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        public string GetMessage()
+        {
+            if (IsDeleted)
+                return @"deleted";
+
+            return WebUtility.HtmlDecode(Regex.Replace(MessageHtml, @"<(.|\n)*?>", string.Empty));
+        }
+
+        public int GetDeletedChildrenCount => ChildComments.Select(c => c.IsDeleted).Count(c => c);
```
smoogipoo: Rename to `DeletedChildrenCount`.

```suggestion
        public int GetDeletedChildrenCount => ChildComments.Count(c => c.IsDeleted);
```

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,198 @@
...
+                    {
+                        ShowDeleted = { BindTarget = ShowDeleted }
+                    });
+            }
+
+            LoadComponentAsync(page, loaded =>
+            {
+                content.Add(loaded);
+
+                int deletedComments = response.Comments.Select(c => c.IsDeleted && c.IsTopLevel).Count(c => c);
```
smoogipoo:
```suggestion
                int deletedComments = response.Comments.Count(c => c.IsDeleted && c.IsTopLevel);
```

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,198 @@
...
+
+                deletedChildrenPlaceholder.DeletedCount.Value = initial ? deletedComments : deletedChildrenPlaceholder.DeletedCount.Value + deletedComments;
```
smoogipoo:
```suggestion
                deletedChildrenPlaceholder.DeletedCount.Value += deletedComments;
```

This `initial` bool can be removed

osu.Game/Overlays/Comments/CommentsContainer.cs
smoogipoo: I'd also like to see it removed.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,198 @@
...
+                    moreButton.Current.Value = response.TopLevelCount - loadedTopLevelComments;
+                    moreButton.IsLoading = false;
+                }
+
+                moreButton.FadeTo(response.HasMore ? 1 : 0);
+            }, loadCancellation.Token);
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
```
smoogipoo: `load()` should be just below the constructor.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        protected override void LoadComplete()
+        {
+            Sort.BindValueChanged(onSortChanged, true);
+            base.LoadComplete();
+        }
+
+        private void onSortChanged(ValueChangedEvent<CommentsSortCriteria> sort) => getComments();
+
+        private void getComments(bool initial = true)
```
smoogipoo: Not really a fan of this `initial` parameter. Would prefer a `clearComments()` method instead.

osu.Game/Overlays/Comments/DeletedChildrenPlaceholder.cs
```diff
@@ -0,0 +1,70 @@
...
+        protected override void LoadComplete()
+        {
+            DeletedCount.BindValueChanged(onCountChanged, true);
+            ShowDeleted.BindValueChanged(onShowDeletedChanged, true);
+            base.LoadComplete();
+        }
+
+        private void onShowDeletedChanged(ValueChangedEvent<bool> showDeleted)
+        {
+            if (canBeShown)
```
smoogipoo: This is very flimsy. So if showdeleted is set to true it hides it, but if the count is then shown it will go to line 67 which shows it. These cases can all be grouped in an `updateDisplay()` method or similar, and the `canBeShown` bool defined locally.

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -0,0 +1,356 @@
...
+                    Origin = Anchor.CentreLeft,
+                    Font = OsuFont.GetFont(size: 12),
+                    Text = $@"edited {HumanizerUtils.Humanize(comment.EditedAt.Value)} by {comment.EditedUser.Username}"
+                });
+            }
+
+            if (!comment.IsDeleted)
+            {
+                var formattedSource = MessageFormatter.FormatText(comment.GetMessage());
+                message.AddLinks(formattedSource.Text, formattedSource.Links);
```
smoogipoo: As mentioned before, should be able to use markdown?

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -0,0 +1,356 @@
...
+            }
+            else
+            {
+                content.FadeColour(OsuColour.Gray(0.5f));
+                votePill.Hide();
+            }
+
+            if (comment.IsTopLevel)
```
smoogipoo: What about for comments nested multiple levels deep? Example: https://osu.ppy.sh/comments/915348

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -0,0 +1,356 @@
...
+                info.Add(new SpriteText
+                {
+                    Anchor = Anchor.CentreLeft,
+                    Origin = Anchor.CentreLeft,
+                    Font = OsuFont.GetFont(size: 12),
+                    Text = $@"edited {HumanizerUtils.Humanize(comment.EditedAt.Value)} by {comment.EditedUser.Username}"
+                });
+            }
+
+            if (!comment.IsDeleted)
```
smoogipoo: This needs to support dynamic updating (based on `ShowDeleted`?)

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -0,0 +1,356 @@
...
+            {
+                var formattedSource = MessageFormatter.FormatText(comment.GetMessage());
+                message.AddLinks(formattedSource.Text, formattedSource.Links);
+            }
+            else
+            {
+                content.FadeColour(OsuColour.Gray(0.5f));
```
smoogipoo: `content` is effectively empty yeah? It needs to contain the message for admins when `ShowDeleted = true`.

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -0,0 +1,356 @@
...
+
+            comment.ChildComments.ForEach(c => childCommentsContainer.Add(new DrawableComment(c)
+            {
+                ShowDeleted = { BindTarget = ShowDeleted }
+            }));
+        }
+
+        protected override void LoadComplete()
+        {
+            ShowDeleted.BindValueChanged(onShowDeletedChanged, true);
```
smoogipoo: Do it inline? `BindValueChanged(show => this.FadeTo(show.NewValue ? 1 : 0), true)`

smoogipoo (changes requested):

smoogipoo:
Testing could be improved. There should be tests that don't require the API that test most of everything that you can see (e.g. nesting, show more correctly appearing in nested levels, multiple nested levels, showing + hiding replies, etc).

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -0,0 +1,83 @@
...
+        public bool IsTopLevel => !ParentId.HasValue;
+
+        public bool IsDeleted => DeletedAt.HasValue;
+
+        public string GetMessage()
+        {
+            if (IsDeleted)
+                return @"deleted";
+
+            return WebUtility.HtmlDecode(Regex.Replace(MessageHtml, @"<(.|\n)*?>", string.Empty));
```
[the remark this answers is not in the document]
EVAST9919: Message is null for most cases

osu.Game/Online/API/Requests/Responses/APICommentsController.cs
```diff
@@ -0,0 +1,80 @@
...
+        public bool HasMore { get; set; }
+
+        [JsonProperty(@"has_more_id")]
+        public long? HasMoreId { get; set; }
+
+        [JsonProperty(@"user_follow")]
+        public bool UserFollow { get; set; }
+
+        [JsonProperty(@"included_comments")]
+        public List<Comment> IncludedComments { get; set; }
```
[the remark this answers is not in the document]
EVAST9919: Included comments are used for comment sub-request (which is not included in this pr, so I don't think it's necessary for now.

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: It works, as you can see in tests
Screenshot_2

osu.Game/Online/API/Requests/Responses/Comment.cs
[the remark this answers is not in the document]
smoogipoo: Right. Deserves a comment similar to the one in ChangelogBuild then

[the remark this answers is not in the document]
EVAST9919: Can you please explain what `Message` should include? Because I have no idea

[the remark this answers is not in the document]
smoogipoo: `Message` doesn't seem to be implemented yet according to the web guys. It should be the markdown version of `MessageHtml`. It's fine to use `MessageHtml` for now.

osu.Game/Overlays/Comments/DrawableComment.cs
smoogipoo: image

EVAST9919: So API sends the `MessageHtml` of deleted comments only for several types of users, right?

smoogipoo: For admins, yup.

EVAST9919: Should be ok now

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: The goal of this pr is just a read-only access without the actual interaction with comments, but yeah, sure

smoogipoo: I misread this one as `if (!ShowDeleted)`... Don't worry about it.

smoogipoo (changes requested):

smoogipoo:
Remaining issues (found on `Big Black` and sort by old):
1. https://streamable.com/qrj9d - Replies may not be sent and need to have their own request. This can be seen on the comment by KinomiCandy.
2. Related to (1), osu-web displays a "show more" link for long reply threads image
This can be seen on the comment by rrtyui.

EVAST9919:
I literally wrote that in the pr description: pr is already too big, gonna include that in another one.

smoogipoo:
I see. This is fine for now in that case.

smoogipoo (approved):

PR #6508, opened 2019-10-14
Lazy ruleset loading

miterosan:
Loads the rulessets lasily. This allows other code to run before the rulesets are determined.

PR #6510, opened 2019-10-14
Fix incorrect current directory on Android

### LINKED ISSUE #6507: osu!lazer not loading rulesets  (Android)
**Describe the bug:** I don't really know how serious this is, mainly because Lazer ran fine with few lag spikes, but whenever I open Lazer, I keep getting a notification saying, "Could not load rulesets from directory."

Screenshot_20191013-204820

**Logs:** 
[runtime.log](https://github.com/ppy/osu/files/3725517/runtime.log)

miterosan:
* depends on #6508
* fixes #6507


osu.Android/OsuGameActivity.cs
```diff
@@ -16,6 +16,11 @@ public class OsuGameActivity : AndroidGameActivity
 
         protected override void OnCreate(Bundle savedInstanceState)
         {
+            // The default current directory on android is '/'.
+            // On some devices '/' maps to the app data directory. On others it maps to the root of the internal storage.
+            // In order to have a consitend current directory on all devices the full path of the app data directory is set as the current directory.
```
bdach:
```suggestion
            // In order to have a consistent current directory on all devices the full path of the app data directory is set as the current directory.
```

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -16,17 +16,11 @@ namespace osu.Game.Rulesets
     /// </summary>
     public class RulesetStore : DatabaseBackedStore
     {
-        private static readonly Dictionary<Assembly, Type> loaded_assemblies = new Dictionary<Assembly, Type>();
+        private static readonly Lazy<Dictionary<Assembly, Type>> loaded_assemblies = new Lazy<Dictionary<Assembly, Type>>(() => loadRulesets());
```
bdach: Can you guarantee that accesses to this are thread-safe? (as in, this can't be called multiple times and cause races) If not, might be worth sticking a `isThreadSafe: false` on this.

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -113,64 +107,73 @@ private void addMissingRulesets()
...
             try
             {
                 string[] files = Directory.GetFiles(Environment.CurrentDirectory, $"{ruleset_library_prefix}.*.dll");
 
                 foreach (string file in files.Where(f => !Path.GetFileName(f).Contains("Tests")))
-                    loadRulesetFromFile(file);
+                {
+                    try
+                    {
+                        rulesetAssemblies.Add(Assembly.LoadFrom(file));
```
bdach: So I assume the `Lazy` was added because `OsuGameActivity` doesn't get the chance to overwrite `Environment.CurrentDirectory` before `RulesetStore` starts looking. Leaving aside the hackiness of that (is that execution order guaranteed?), can xamarin even dynamically load in assemblies from DLLs?

bdach (commented):
I'm not sure if this was worth splitting into 2 PRs for the sake of 5 lines of difference, so I'll leave all my comments here.

osu.Game/Rulesets/RulesetStore.cs
miterosan: >can xamarin even dynamically load in assemblies from DLLs?
yes

osu.Game/Rulesets/RulesetStore.cs
miterosan: The initialization logic is thread safe. The only thing I could imagine creating problems is Assembly.Load. But that method is also thread safe according to [here](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly?view=netframework-4.8#thread-safety)

So this should be fine.

osu.Game/Rulesets/RulesetStore.cs
miterosan: > So I assume the Lazy was added because OsuGameActivity doesn't get the chance to overwrite Environment.CurrentDirectory before RulesetStore starts looking.

That is correct. Enforcing the direct path ensures that the working directory will point to correct directory.

> is that execution order guaranteed?

The line https://github.com/ppy/osu/pull/6510/files#diff-1afa33156b358a1e34958fc4fbdadf87R23 gets called before the framework loads. (The framework gets loaded in the base.OnCreate)
Therefore yes.

swoolcock:
Have you tried this on iOS?  Xamarin.iOS can't dynamically load assemblies, as they have to be AOT-compiled.

miterosan:
@swoolcock this does not change the logic that happenes for iOS.

If it did work before for iOS, it will after this pr.

Of course it cannot hurt somebody testing this pr with iOS.

smoogipoo:
This change can be simplified to moving those function calls to within the ctor, and removing all statics, right? Would highly prefer that.

bdach (dismissed):
As #6515 got merged, the lazy-loading is no longer necessary and I guess 6da1b4d is the only real change we want now.

bdach:
The typo in the comment is back.

smoogipoo (approved):
Will follow the same process as the back button fix - easier to wait for issues than test.

PR #6514, opened 2019-10-15
Make OverlayContainers with no blocking input VisibilityContainers

Joehuu:
- `OverlayContainer`s with only `BlockPositionalInput` overridden to false are changed to `VisibilityContainer`s.
- `VersionManager` shouldn't block input in the first place, and thus changed.
- `PlaylistOverlay` is a child to `NowPlayingOverlay`, and thus changed.


smoogipoo (approved):
seems to work well

PR #6515, opened 2019-10-15
Make RulesetStore initialise at construction time

smoogipoo:
Replaces the `RulesetStore` changes in https://github.com/ppy/osu/pull/6510 with not using statics altogether.

### DIFF
--- a/osu.Game/Rulesets/RulesetStore.cs
+++ b/osu.Game/Rulesets/RulesetStore.cs
@@ -14,25 +14,22 @@ namespace osu.Game.Rulesets
     /// <summary>
     /// Todo: All of this needs to be moved to a RulesetStore.
     /// </summary>
-    public class RulesetStore : DatabaseBackedStore
+    public class RulesetStore : DatabaseBackedStore, IDisposable
     {
-        private static readonly Dictionary<Assembly, Type> loaded_assemblies = new Dictionary<Assembly, Type>();
+        private const string ruleset_library_prefix = "osu.Game.Rulesets";
 
-        static RulesetStore()
-        {
-            AppDomain.CurrentDomain.AssemblyResolve += currentDomain_AssemblyResolve;
-
-            // On android in release configuration assemblies are loaded from the apk directly into memory.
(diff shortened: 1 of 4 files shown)


osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -158,19 +150,35 @@ private static void loadRulesetFromFile(string file)
...
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            AppDomain.CurrentDomain.AssemblyResolve -= resolveRulesetAssembly;
+        }
```
bdach: This `Dispose()` implementation seems to go against what I thought I knew about the dispose pattern, namely that:

* finalizers should be used pretty much only when there are managed resources to dispose, and
* disposing managed resources inside `Dispose(bool disposing)` should be guarded by the `disposing` parameter.

As far as I understand this could attempt to unsubscribe from `AssemblyResolve` twice (I do however know [finalizers are not guaranteed to run in general](https://ericlippert.com/2015/05/18/when-everything-you-know-is-wrong-part-one/)). I *could* very well be full of it, though (I'd even say it's likely).

bdach: Oh, and [part two of that post](https://ericlippert.com/2015/05/21/when-everything-you-know-is-wrong-part-two/) says that apparently it might be wrong to even assume the `AssemblyResolve` event was even subscribed to in the first place, since the ctor didn't necessarily fully execute...

smoogipoo: Unsubscribing from events that haven't been subscribed to has no ill effects.

It's sometimes fine to use the finalizer in a managed context (e.g. `ArrayPool<>`). So long as you understand the implications of doing so (thread safety, etc). In this case the usage is fine, but it won't actually do anything since a managed reference to the object is held from an external source if `Dispose()` is not invoked, so I'll remove it.

peppy (approved):

PR #6522, opened 2019-10-15
Fix incorrect beatmap set info equality check for local beatmapsets

frenzibyte:


peppy:
Can you explain the case this is fixing? Is there a reproduction somewhere we can test against?

osu.Game/Beatmaps/BeatmapSetInfo.cs
```diff
@@ -63,6 +63,12 @@ public class BeatmapSetInfo : IHasPrimaryKey, IHasFiles<BeatmapSetFileInfo>, ISo
 
         public bool Protected { get; set; }
 
-        public bool Equals(BeatmapSetInfo other) => OnlineBeatmapSetID == other?.OnlineBeatmapSetID;
+        public bool Equals(BeatmapSetInfo other)
+        {
+            if (!OnlineBeatmapSetID.HasValue || !(other?.OnlineBeatmapSetID.HasValue ?? false))
+                return ReferenceEquals(this, other);
```
smoogipoo: Should check by `ID` first, if it's non-zero.

smoogipoo (changes requested):

iiSaLMaN:
> Can you explain the case this is fixing? Is there a reproduction somewhere we can test against?

Found this issue after creating a bindable of `BeatmapSetInfo`, realized it checks whether it's equal with another value by `BeatmapSetInfo.Equals`, which made issues to the bindable not calling value changed events, will provide tests.

bdach:
> Found this issue after creating a bindable of `BeatmapSetInfo`, realized it checks whether it's equal with another value by `BeatmapSetInfo.Equals`, which made issues to the bindable not calling value changed events, will provide tests.

Mind expounding on what the actual *values* were? Was it the case of a pair of maps that were different but didn't have an online ID, or some other less obvious scenario?

iiSaLMaN:
> > Found this issue after creating a bindable of `BeatmapSetInfo`, realized it checks whether it's equal with another value by `BeatmapSetInfo.Equals`, which made issues to the bindable not calling value changed events, will provide tests.
> 
> Mind expounding on what the actual _values_ were? Was it the case of a pair of maps that were different but didn't have an online ID, or some other less obvious scenario?

2 beatmaps with same online ID (null)

bdach:
In that case I'm not completely sold that `ReferenceEquals` is correct here, could be too strict in cases there are duplicate references that actually point to the same beatmap set - `BeatmapSetInfo.Hash` could be more appropriate here? Would ensure that the actual beatmap set content is the same and account for multiple instances representing a single underlying set.

smoogipoo (approved):

PR #6523, opened 2019-10-15
Make most textbox carets movable

### LINKED ISSUE #6484: Some TextBoxes should be able to move the caret between letters
**Describe the bug:** TextBoxes like the settings' one, the _filter_ one from the music controller and the social/osu!direct's one can't move the caret between letters. I think they should be able to do that, like in the chat overlay (except for the carousel's one, because with <kbd>left</kbd> and <kbd>right</kbd> you change beatmap): it could be useful for typos (especially when searching for payers/maps in the social/osu!direct's one, instead of deleting the entire text with <kbd>Escape</kbd>).

Joehuu:
Resolves https://github.com/ppy/osu/issues/6484.

### DIFF
--- a/osu.Game/Graphics/UserInterface/FixedSearchTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/FixedSearchTextBox.cs
@@ -1,0 +1,10 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Graphics.UserInterface
+{
+    public class FixedSearchTextBox : SearchTextBox
+    {
+        public override bool HandleLeftRightArrows => false;
+    }
+}
--- a/osu.Game/Graphics/UserInterface/SearchTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/SearchTextBox.cs
@@ -14,8 +14,6 @@ namespace osu.Game.Graphics.UserInterface
     {
         protected virtual bool AllowCommit => false;
 
-        public override bool HandleLeftRightArrows => false;
-
         public SearchTextBox()
         {
             Height = 35;
--- a/osu.Game/Graphics/UserInterface/SeekLimitedSearchTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/SeekLimitedSearchTextBox.cs
@@ -3,7 +3,7 @@
 
 namespace osu.Game.Graphics.UserInterface
 {
-    public class FixedSearchTextBox : SearchTextBox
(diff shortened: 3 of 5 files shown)


frenzibyte (dismissed):

bdach:
Just for completeness' sake, here's what I think is an exhaustive list of the locations whose textboxes will now have movable carets (compiled purely by looking through references):

* `osu.Game.Overlays.SettingsPanel`
* `osu.Game.Overlays.Chat.Selection.ChannelSelectionOverlay`
* All users of `osu.Game.Overlays.Music.FilterControl`:
	- `osu.Game.Overlays.Music.PlaylistOverlay`
* All users of `osu.Game.Overlays.SearchableList.SearchableListFilterControl` and derivatives:
	- `osu.Game.Overlays.DirectOverlay`
	- `osu.Game.Overlays.SocialOverlay`
	- `osu.Game.Screens.Multi.Lounge.LoungeSubScreen`

smoogipoo:
Should be disabled in `SettingsPanel` due to `SliderBars` being adjustable via left-right.

bdach:
Also, on a general note, how about making `HandleLeftRightArrows` explicitly settable (be it framework- or lazer-side)? Having to subclass to change behaviour for this one property alone seems suboptimal to me.

peppy:
Almost makes me wonder if we could somehow have a non-focused input path for such cases so we wouldn't have to specify it in the first place (not sure how feasible this is though).

- textbox focused OnKeyDown handling
- rest of heirarchy until textbox
- textbox unfocused OnKeyDown handling

osu.Game/Graphics/UserInterface/FixedSearchTextBox.cs
```diff
@@ -0,0 +1,10 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Graphics.UserInterface
+{
+    public class FixedSearchTextBox : SearchTextBox
```
peppy: What does "fixed" mean here?

peppy (changes requested):

osu.Game/Graphics/UserInterface/FixedSearchTextBox.cs
Joehuu: The caret is "fixed". Sounds misleading now, but wasn't too sure of what to name it in the first place.

peppy: `SeekLimitedSearchTextBox` maybe... hard to come up with a good name

peppy (approved):

PR #6527, opened 2019-10-17
Implement nested drawable hitobject reconstruction

smoogipoo:
This is made primarily for the editor but I've kept hitobject pooling in the back of my mind. The reason this is required is two scenarios:
1. In the editor, `ApplyDefaults` is invoked, which creates new nested `HitObject`s. Nested `DrawableHitObject`s are not recreated so they become out-of-date.
2. In pooling, we'll need to give `DrawableHitObject`s a model from which they'll update.

To do this I've introduced 3 virtual methods:
1. `CreateNestedHitObject` - Creates the drawable representation for a nested `HitObject`. Eventually with pooling this will likely be moved to a higher level (something like `pool.GetDrawableRepresentation(HitObject)`).
2. `AddNestedHitObject` - Adds the nested `DrawableHitObject` to the hierarchy.
3. `ClearNestedHitObjects` - Removes all previously added `DrawableHitObject`s from the hierarchy.

The general invocation structure is:
```
ClearNestedHitObjects()
for each hitObject:
    drawable = CreateNestedHitObject(hitObject)
    AddNestedHitObject(drawable)
```

The existing method, `AddNested()`, has been obsoleted for removal on 20200417.

This all happens in the `Apply(HitObject)` method, which is currently protected but will be made public with hitobject pooling. Eventually it'll also have another method such as `ApplyToSelf()` which will contain the ability to copy over properties (position/etc).

### DIFF
--- a/osu.Game.Rulesets.Mania/Edit/Blueprints/HoldNoteSelectionBlueprint.cs
+++ b/osu.Game.Rulesets.Mania/Edit/Blueprints/HoldNoteSelectionBlueprint.cs
@@ -20,28 +20,34 @@ namespace osu.Game.Rulesets.Mania.Edit.Blueprints
 
         private readonly IBindable<ScrollingDirection> direction = new Bindable<ScrollingDirection>();
 
-        private readonly BodyPiece body;
+        [Resolved]
+        private OsuColour colours { get; set; }
 
         public HoldNoteSelectionBlueprint(DrawableHoldNote hold)
             : base(hold)
         {
-            InternalChildren = new Drawable[]
-            {
-                new HoldNoteNoteSelectionBlueprint(hold.Head),
-                new HoldNoteNoteSelectionBlueprint(hold.Tail),
-                body = new BodyPiece
-                {
-                    AccentColour = Color4.Transparent
-                },
-            };
         }
 
         [BackgroundDependencyLoader]
-        private void load(OsuColour colours, IScrollingInfo scrollingInfo)
+        private void load(IScrollingInfo scrollingInfo)
         {
-            body.BorderColour = colours.Yellow;
-
             direction.BindTo(scrollingInfo.Direction);
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            InternalChildren = new Drawable[]
+            {
+                new HoldNoteNoteSelectionBlueprint(HitObject.Head),
+                new HoldNoteNoteSelectionBlueprint(HitObject.Tail),
+                new BodyPiece
+                {
+                    AccentColour = Color4.Transparent,
+                    BorderColour = colours.Yellow
+                },
+            };
         }
 
         protected override void Update()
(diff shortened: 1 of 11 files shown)


osu.Game.Rulesets.Mania/Edit/Blueprints/HoldNoteSelectionBlueprint.cs
```diff
@@ -20,30 +20,36 @@ public class HoldNoteSelectionBlueprint : ManiaSelectionBlueprint
...
-            body.BorderColour = colours.Yellow;
-
             direction.BindTo(scrollingInfo.Direction);
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            InternalChildren = new Drawable[]
```
peppy: why was this moved?

smoogipoo: Because previously mania expected the nested hitobjects to be available in the ctor. osu! handles this by not passing down the nested hitobjects anymore, instead using the `UpdateFrom()` method.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -134,6 +137,68 @@ protected override void LoadComplete()
...
+
+            if (nestedHitObjects.IsValueCreated)
+            {
+                nestedHitObjects.Value.Clear();
+                ClearNestedHitObjects();
+            }
+
+            foreach (var h in hitObject.NestedHitObjects)
+            {
+                var drawableNested = CreateNestedHitObject(h) ?? throw new InvalidOperationException($"{nameof(CreateNestedHitObject)} returned null for {h.GetType().ReadableName()}.");
```
peppy: are you sure we don't want to allow null here? potentially some hitobjects don't need drawable pairs. for example slider end circles (which currently have a drawable but don't necessarily need it; it shows nothing).

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -134,6 +137,68 @@ protected override void LoadComplete()
...
+
+                drawableNested.OnNewResult += (d, r) => OnNewResult?.Invoke(d, r);
```
peppy: should probably move this out into a separate method (duplicated code with `AddNested` below).

osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
```diff
@@ -123,12 +123,20 @@ private void addBlueprintFor(DrawableHitObject hitObject)
             if (blueprint == null)
                 return;
 
-            blueprint.Selected += onBlueprintSelected;
-            blueprint.Deselected += onBlueprintDeselected;
-            blueprint.SelectionRequested += onSelectionRequested;
-            blueprint.DragRequested += onDragRequested;
+            if (hitObject.IsLoaded)
```
peppy: comment plz

peppy (changes requested):

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
peppy: discussed IRL, makes sense (DHOs are required to handle judgement results).

osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
smoogipoo: Turns out this isn't actually required anymore due to https://github.com/ppy/osu/pull/6527/commits/510ce9345f79b3deee4f6ed1fc72e46ebed0ed0e

peppy (approved):

PR #6529, opened 2019-10-17
Implement working VotePill for CommentsContainer

EVAST9919:


osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -0,0 +1,179 @@
...
+        {
+            request = new CommentVoteRequest(comment.Id, isVoted.Value ? CommentVoteAction.UnVote : CommentVoteAction.Vote);
+            request.Success += onSuccess;
+            api.Queue(request);
+        }
+
+        private void onSuccess(CommentBundle response)
+        {
+            isVoted.Value = !isVoted.Value;
+            setCount(response.Comments.First().VotesCount);
```
bdach: How about doing `.Single()` instead? Is there any threat this will ever return more than one comment in the response?

osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -0,0 +1,179 @@
...
+                borderContainer.BorderThickness = 3;
+
+                if (!isVoted.Value)
+                    sideNumber.Show();
+                else
+                    hoverLayer.Show();
+            }
+        }
+
+        private void setCount(int count) => votesCounter.Text = $"+{count}";
```
bdach: This feels not quite right. I'd ideally like to see a properly-typed `int` property which then auto-updates when the count changes - maybe a case for a `Bindable`?

bdach (commented):
I have some comments code-wise, but not very major and I don't know if I'm right on these, so take with a <strike>grain</strike> pillar of salt?

bdach:
I do however have a major *design* complaint to make here, though. I can see that the design is *mostly* in line with web, but I feel like it has a major accessibility flaw for voted comments in that you can't see the vote count properly due to the low contrast of the green and white. This is exacerbated in lazer especially, since the drop shadow used isn't as diffused as it is web-side:

| web | lazer |
| :-: | :-: |
| image | image |

I was going to try suggest something alternative, but I'm trying to grasp how the drawable structure works, so bear with me while I try to figure that out.

osu.Game/Graphics/UserInterface/LoadingButton.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly LoadingAnimation loading;
+        private readonly Drawable content;
+
+        protected LoadingButton()
+        {
+            Container background;
+
+            Child = background = CreateBackground();
+
+            background.AddRange(new[]
```
peppy: why are you adding elements inside a "background"? this seems counter-intuitive.

EVAST9919: Still bad at naming. Any suggestions?

peppy: It's content if you're adding content to it.

osu.Game/Graphics/UserInterface/LoadingButton.cs
```diff
@@ -0,0 +1,97 @@
...
+                return base.OnClick(e);
+            }
+            finally
+            {
+                // run afterwards as this will disable this button.
+                IsLoading = true;
+            }
+        }
+
+        protected virtual void OnLoadingStart()
```
smoogipoo: `OnLoadStarted` or `OnLoading`

osu.Game/Graphics/UserInterface/LoadingButton.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+        }
+
+        protected virtual void OnLoadingFinished()
```
smoogipoo: To compliment the above:

`OnLoadFinished` or `OnLoaded`

osu.Game/Graphics/UserInterface/LoadingButton.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+        }
+
+        protected abstract Container CreateContent();
```
smoogipoo: I don't feel good with this returning a container which has things added to it. It's not what I'd expect from "Create**Content**".

It'd be nice to have `CreateContent()` be a final thing, returning a non-descript `Drawable`. This could be accomplished by removing `CreateText()` and letting consumers decide what to hide inside `OnLoading` / `OnLoaded`.

smoogipoo (changes requested):

smoogipoo (approved):

PR #6532, opened 2019-10-17
changing repetition of code to foreach

yurastico:


bdach (changes requested):
Please trim unnecessary whitespace as per CI suggestions. You can [run code analysis locally](https://github.com/ppy/osu#code-analysis) to ensure style correctness before pushing again.

osu.Game/OsuGameBase.cs
```diff
@@ -134,29 +155,33 @@ private void load()
...
+
+
+
+
+
+
+
+
+
+
```
TPGPL: you should probably delete all this whitespace

TPGPL (changes requested):

frenzibyte (changes requested):
If this is to be corrected for, It should probably be looking up for any font inside the `Fonts` folder rather than having a hardcoded list containing the fonts.

osu.Game/OsuGameBase.cs
```diff
@@ -120,6 +120,27 @@ public OsuGameBase()
 
         protected override UserInputManager CreateUserInputManager() => new OsuUserInputManager();
 
+        private readonly IList<string> fontPaths = new List<string>()
```
swoolcock: Use `IEnumerable<string>` and an array.

peppy:
Closing as low quality / abandoned. Please re-open if you wish to fix this.

PR #6535, opened 2019-10-18
Update WaveContainer to support framework changes

peppy:
Should be merged to fix issues with upstream framework changes in https://github.com/ppy/osu-framework/pull/2902.

- [x] Depends on framework build with https://github.com/ppy/osu-framework/pull/2902


smoogipoo:
Seems to break initial overlay transform: https://streamable.com/3551y

smoogipoo:
Turns out I wasn't checked out locally.

smoogipoo (approved):

PR #6542, opened 2019-10-19
Implement comment replies loading ability

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu/pull/6722
- [x] https://github.com/ppy/osu/pull/6529

Implements:
- "show more" button
- "load replies" button


osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -44,8 +44,40 @@ public List<Comment> Comments
         [JsonProperty(@"user_follow")]
         public bool UserFollow { get; set; }
 
+        private List<Comment> includedComments;
+
         [JsonProperty(@"included_comments")]
-        public List<Comment> IncludedComments { get; set; }
+        public List<Comment> IncludedComments
```
bdach: I'm confused - what is `IncludedComments` and how does it differ from `Comments`? Why does `IncludedComments`'s getter modify `Comments` too?

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -44,8 +44,40 @@ public List<Comment> Comments
...
+                {
+                    if (child.ParentId != null)
+                    {
+                        value.ForEach(parent =>
+                        {
+                            if (parent.Id == child.ParentId)
+                            {
+                                parent.ChildComments.Add(child);
+                                child.ParentComment = parent;
+                            }
```
bdach: This snippet shows up thrice in this file, possibly factorable into a method.

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -59,6 +91,12 @@ private List<long> userVotes
                         if (v == c.Id)
                             c.IsVoted = true;
                     });
+
+                    IncludedComments.ForEach(c =>
+                    {
+                        if (v == c.Id)
+                            c.IsVoted = true;
+                    });
```
bdach: Delegate duplicated twice.

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -83,6 +121,15 @@ public List<User> Users
...
                     });
+
+                    IncludedComments.ForEach(c =>
+                    {
+                        if (c.UserId == u.Id)
+                            c.User = u;
+
+                        if (c.EditedById == u.Id)
+                            c.EditedUser = u;
+                    });
```
bdach: Delegate duplicated twice.

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -24,23 +25,27 @@ public class DrawableComment : CompositeDrawable
         private const int margin = 10;
 
         public readonly BindableBool ShowDeleted = new BindableBool();
+        public readonly Bindable<CommentsSortCriteria> Sort = new Bindable<CommentsSortCriteria>();
 
         private readonly BindableBool childrenExpanded = new BindableBool(true);
+        private readonly Bindable<List<Comment>> childComments = new Bindable<List<Comment>>();
+        private readonly List<Comment> loadedChildren = new List<Comment>();
```
bdach: How does `childComments` differ from `loadedChildren`?

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -257,39 +273,60 @@ protected override void LoadComplete()
                     this.FadeTo(show.NewValue ? 1 : 0);
             }, true);
             childrenExpanded.BindValueChanged(expanded => childCommentsVisibilityContainer.FadeTo(expanded.NewValue ? 1 : 0), true);
+            childComments.BindValueChanged(children =>
+            {
+                children.NewValue.ForEach(c =>
+                {
+                    if (!loadedChildren.Contains(c))
```
bdach: Possibly nitpicky again, but this `Contains` on a list makes this quadratic. If possible I'd want to see `loadedChildren` being a `Dictionary` or `HashSet`, unless the order of elements matters.

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -298,7 +335,12 @@ public RepliesButton(int count)
 
             protected override void OnExpandedChanged(ValueChangedEvent<bool> expanded)
             {
-                text.Text = $@"{(expanded.NewValue ? "[+]" : "[-]")} replies ({count})";
+                text.Text = $@"{(expanded.NewValue ? "[+]" : "[-]")} replies ({repliesCount})";
+            }
+
+            protected override void OnChildrenChanged(ValueChangedEvent<List<Comment>> children)
+            {
+                Alpha = (children.NewValue.Count == 0 && repliesCount > 0) || repliesCount == 0 ? 0 : 1;
```
bdach: First off, this check can be simplified to
```csharp
repliesCount == 0 || children.NewValue.Count == 0
```
but mostly I'm confused with what this does. Does `children` contain only the comments that are going to be *added* below (therefore meaning that there's nothing more to load, so the button can be hidden)? I don't see any other reason for hiding the button if it's empty; if so, mind changing the parameter naming to reflect that?

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
```diff
@@ -0,0 +1,94 @@
...
+            request.Success += onSuccess;
+            api.Queue(request);
+        }
+
+        private void onSuccess(CommentBundle response)
+        {
+            var children = ChildComments.Value.ToList();
+            response.Comments.ForEach(c =>
+            {
+                if (c.ParentId == Comment.Id && children.Count(u => u.Id == c.Id) == 0)
```
bdach: * Why is the parent id checked here? I assume web should take care of that, since it's passed in the request.
* `Count` is slightly inefficient here, since it won't stop at the first matching result. Use negated `.Any()` or `.All()` instead.

osu.Game/Overlays/Comments/ShowMoreRepliesButton.cs
```diff
@@ -0,0 +1,36 @@
...
+        [BackgroundDependencyLoader]
+        private void load(OsuColour colours)
+        {
+            IdleColour = colours.Blue;
+            HoverColour = colours.BlueLighter;
+        }
+
+        protected override void OnChildrenChanged(ValueChangedEvent<List<Comment>> children)
+        {
+            Alpha = (!children.NewValue.Any() && Comment.RepliesCount > 0) || children.NewValue.Count == Comment.RepliesCount ? 0 : 1;
```
bdach: I don't understand the first condition here. I assume the second is to cover the case where all the replies are loaded at once, but what's with the first?

Also in general I think this is too complicated - as far as I understand you *know* how many replies are expected in total, so if you can access the *total* number of children at the point of calling this, this should reduce to checking if the counts are equal.

bdach (changes requested):
Notes from a read-through of the source alone. Haven't ran the code. I assume the tests from the previous PR will also cover this, since they're fetching online data.

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: `ChildComments` is being modified on every `GetCommentReplies` request. However, the response of this request can include already loaded comments from the main request. so `losdedChildren` is used just to check, whether the comment already exists

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: Order of the elements matters because of the `Sort` usage. We are receiving already ordered comments

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
EVAST9919: I could just use `Contains` but this is not working because some data can be differ for one comment but different requests (for example, in one request comment don't have children, but in another it has)

osu.Game/Overlays/Comments/ShowMoreRepliesButton.cs
EVAST9919: First condition is for the case, when we know, what there are replies exist, but they are no loaded and atm we don't have any children (in this case `LoadMoreCommentsButton` will be shown)

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
EVAST9919: We can receive the same comment twice (first time in the main request, second - in the `GetRepliesRequest`)

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
EVAST9919: `IncludedComments` contains all non-top-level comments (in case of `GetRepliesRequest`)

osu.Game/Overlays/Comments/DrawableComment.cs
bdach: I'd either advocate for a naming change or an xmldoc here for clarity (ie. `requestedComments` vs `loadedComments`?)

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
```diff
@@ -0,0 +1,94 @@
...
+            request.Success += onSuccess;
+            api.Queue(request);
+        }
+
+        private void onSuccess(CommentBundle response)
+        {
+            var children = ChildComments.Value.ToList();
+            response.Comments.ForEach(c =>
+            {
+                if (!children.Any(child => child.Id == c.Id))
```
bdach:
```suggestion
                if (children.All(child => child.Id != c.Id))
```
Clears up a resharper warning visible in the CI build log.

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: Removed by using `children.OldValue` in `childComments.BindValueChanged()`

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
```diff
@@ -0,0 +1,92 @@
...
+            request.Success += onSuccess;
+            api.Queue(request);
+        }
+
+        private void onSuccess(CommentBundle response)
+        {
+            var children = ChildComments.Value.ToList();
+            response.Comments.ForEach(c =>
+            {
+                if (children.All(child => child.Id != c.Id))
```
smoogipoo: Is this not guaranteed by the request? If it is I'm thinking this can be refactored further by using `BindableList<T>`.

In here:
```
// perform request
foreach (var comment in response)
    list.Add(comment)
```

In DrawableComment:
```
list.OnAdded += c =>  /* add new drawable comment */
```

which would remove the whole bindvaluechanged + foreach stuff.

osu.Game/Overlays/Comments/GetCommentRepliesButton.cs
```diff
@@ -0,0 +1,92 @@
...
+using osu.Game.Online.API;
+using osu.Framework.Bindables;
+using osu.Game.Online.API.Requests;
+using System.Linq;
+
+namespace osu.Game.Overlays.Comments
+{
+    public abstract class GetCommentRepliesButton : LoadingButton
+    {
+        public readonly Bindable<List<Comment>> ChildComments = new Bindable<List<Comment>>();
```
smoogipoo: I'm not really a fan of this, modifying `ChildComments` from a button. I'd much rather have this button expose an event like `Action<List<Comment>> OnCommentsReceived`

osu.Game/Overlays/Comments/LoadRepliesButton.cs
```diff
@@ -0,0 +1,29 @@
...
+        public LoadRepliesButton(Comment comment)
+            : base(comment)
+        {
+            IdleColour = OsuColour.Gray(0.7f);
+            HoverColour = Color4.White;
+        }
+
+        protected override void OnChildrenChanged(ValueChangedEvent<List<Comment>> children)
+        {
+            Alpha = children.NewValue.Any() || Comment.RepliesCount == 0 ? 0 : 1;
```
smoogipoo: Relating to my previous comment re: exposing an event instead... This isn't the correct way to do this anyway - shouldn't be changing your own visual state.

osu.Game/Overlays/Comments/ShowChildrenButton.cs
```diff
@@ -3,32 +3,36 @@
...
+            Action = () => Expanded.Value = !Expanded.Value;
+
+            IdleColour = OsuColour.Gray(0.7f);
+            HoverColour = Color4.White;
         }
 
         protected override void LoadComplete()
         {
             Expanded.BindValueChanged(OnExpandedChanged, true);
+            ChildComments.BindValueChanged(OnChildrenChanged, true);
```
smoogipoo: Same here, this is also not the proper way to do this (mutating your own visual state). It's the responsibility of the user of this button to do so.

smoogipoo (changes requested):

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: Resolved by latest refactoring

bdach (dismissed):
Source definitely looking better.

Ran the code, and unfortunately some Markdown links don't seem to be working correctly:
image
The format seems a little bogus, might be potentially a web-side issue?

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -292,7 +378,7 @@ public RepliesButton(int count)
 
             protected override void OnExpandedChanged(ValueChangedEvent<bool> expanded)
             {
-                text.Text = $@"{(expanded.NewValue ? "[+]" : "[-]")} replies ({count})";
+                text.Text = $@"{(expanded.NewValue ? "[+]" : "[-]")} replies ({repliesCount})";
```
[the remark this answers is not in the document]
EVAST9919: oops

EVAST9919:
> Ran the code, and unfortunately some Markdown links don't seem to be working correctly:
> image
> The format seems a little bogus, might be potentially a web-side issue?

This is how message looks like
`[http://puu.sh/a9Clg/143ed6c...](http://puu.sh/a9Clg/143ed6c903.png "http://puu.sh/a9Clg/143ed6c903.png") ..... D:`

bdach:
I checked one of the other steps (the one with the changelog) and the link worked properly. I assume the above example is some vestige of a partially-broken conversion.

A bit tentative on approving since the others might have objections, but looks alright to me - dismissing review.

bdach:
Just as a side note, I have a draft of a fix for the markdown formatting, but it needs some work still and I don't want to rush it. Expect it within a day or two, though.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -13,17 +13,25 @@
...
     {
-        private readonly CommentableType type;
-        private readonly long id;
+        private CommentableType type;
+        private long id;
 
         public readonly Bindable<CommentsSortCriteria> Sort = new Bindable<CommentsSortCriteria>();
         public readonly BindableBool ShowDeleted = new BindableBool();
 
+        public void ShowComments(CommentableType type, long id)
```
smoogipoo: Methods shouldn't be above ctor

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -35,15 +43,14 @@ public class CommentsContainer : CompositeDrawable
         private int currentPage;
 
         private readonly Box background;
+        private readonly Box placeholderBackground;
         private readonly FillFlowContainer content;
-        private readonly DeletedChildrenPlaceholder deletedChildrenPlaceholder;
+        private readonly DeletedCommentsPlaceholder deletedCommentsPlaceholder;
         private readonly CommentsShowMoreButton moreButton;
+        private readonly Container placeholder;
```
smoogipoo: Maybe rename to `noCommentsPlaceholder`

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -13,17 +13,25 @@
...
+        private long id;
 
         public readonly Bindable<CommentsSortCriteria> Sort = new Bindable<CommentsSortCriteria>();
         public readonly BindableBool ShowDeleted = new BindableBool();
 
+        public void ShowComments(CommentableType type, long id)
+        {
+            this.type = type;
+            this.id = id;
+            Sort.TriggerChange();
```
smoogipoo: How about this being the method that clears and retrieves comments, and having the sort change event trigger `ShowComments(type, id)`?

smoogipoo (changes requested):

EVAST9919:
Will be reopened after some refactoring and "offline-mode" implementation

bdach:
I've got to admit I'm a little bit miffed this has grown in scope again. Your changes *need* to be more incremental, or else it's hard to review things properly.

I'll take a look at this later, but please split things more for the good of everyone in the future (and that includes you - smaller PRs = faster merging).

EVAST9919:

PR #6556, opened 2019-10-21
Fix sliders not moving with stacking changes

smoogipoo:
It was completely broken.


peppy (dismissed):

peppy:
Actual test failure here

smoogipoo:
For future me: this is an issue with stable's conversion generation...

peppy (approved):

PR #6557, opened 2019-10-21
Fix hitcircle selection area being too large

smoogipoo:


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs
```diff
@@ -172,7 +170,7 @@ protected override void UpdateStateTransforms(ArmedState state)
 
         public Drawable ProxiedLayer => ApproachCircle;
 
-        private class HitArea : Drawable, IKeyBindingHandler<OsuAction>
+        public class HitReceptor : Drawable, IKeyBindingHandler<OsuAction>
```
peppy: why did you rename the class but not the variable?

smoogipoo: The rename prevents this:
```
public OsuAction? HitAction => HitArea.HitAction; // Compiler error (what is `HitArea`?)
public HitArea HitArea;
```

Open to suggestions.

peppy: right... i guess it's ok for now then.

PR #6569, opened 2019-10-21
Implement day divider for chat

### LINKED ISSUE #6521: Day divider for chat
**Describe the new feature:** A simple divider that separates the chat in days.

**Proposal designs of the feature:** I don't know if you already have a design for this, but here are some examples I've done (with a little inspiration from Discord in the first one, but with a brighter color of the Lazer chat):
<details>
<summary>Design 1</summary>

Design 1
</details>
<details>
<summary>Design 2</summary>

Design 2
</details>
<details>
<summary>Design 3</summary>

Design 3
</details>
<details>
<summary>Design 4</summary>

Design 4
</details>

EVAST9919:
Resolves https://github.com/ppy/osu/issues/6521

Screenshot_1
Screenshot_2

### DIFF
--- a/osu.Game/Overlays/Chat/DrawableChannel.cs
+++ b/osu.Game/Overlays/Chat/DrawableChannel.cs
@@ -95,14 +95,14 @@ namespace osu.Game.Overlays.Chat
 
             Message lastMessage = chatLines.LastOrDefault()?.Message;
 
-            displayMessages.ForEach(m =>
+            foreach (var message in displayMessages)
             {
-                if (lastMessage == null || lastMessage.Timestamp.ToLocalTime().Date != m.Timestamp.ToLocalTime().Date)
-                    ChatLineFlow.Add(CreateDaySeparator(m.Timestamp));
+                if (lastMessage == null || lastMessage.Timestamp.ToLocalTime().Date != message.Timestamp.ToLocalTime().Date)
+                    ChatLineFlow.Add(CreateDaySeparator(message.Timestamp));
 
-                ChatLineFlow.Add(CreateChatLine(m));
-                lastMessage = m;
-            });
+                ChatLineFlow.Add(CreateChatLine(message));
+                lastMessage = message;
+            }
 
             if (scroll.IsScrolledToEnd(10) || !chatLines.Any() || newMessages.Any(m => m is LocalMessage))
                 scrollToEnd();
--- a/osu.Game/Overlays/Chat/DrawableChannel.cs
+++ b/osu.Game/Overlays/Chat/DrawableChannel.cs
@@ -12,6 +12,8 @@ using osu.Framework.Graphics.Containers;
 using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.Cursor;
 using osu.Game.Online.Chat;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Extensions.IEnumerableExtensions;
 
 namespace osu.Game.Overlays.Chat
 {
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -74,17 +82,34 @@ protected override void Dispose(bool isDisposing)
...
+
         private void newMessagesArrived(IEnumerable<Message> newMessages)
         {
             // Add up to last Channel.MAX_HISTORY messages
             var displayMessages = newMessages.Skip(Math.Max(0, newMessages.Count() - Channel.MaxHistory));
 
-            ChatLineFlow.AddRange(displayMessages.Select(CreateChatLine));
+            var existingChatLines = getChatLines();
+
+            Message lastMessage = existingChatLines.Any() ? existingChatLines.Last().Message : null;
```
bdach: Could simplify to just `.LastOrDefault()?.Message`.

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -112,19 +137,90 @@ private void pendingMessageResolved(Message existing, Message updated)
 
         private void messageRemoved(Message removed)
         {
-            ChatLineFlow.Children.FirstOrDefault(c => c.Message == removed)?.FadeColour(Color4.Red, 400).FadeOut(600).Expire();
+            getChatLines().FirstOrDefault(c => c.Message == removed)?.FadeColour(Color4.Red, 400).FadeOut(600).Expire();
         }
 
+        private IEnumerable<ChatLine> getChatLines() => ChatLineFlow.Children.OfType<ChatLine>();
```
bdach: Seems to me this would be better suited as a property (private, but still).

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -112,19 +137,90 @@ private void pendingMessageResolved(Message existing, Message updated)
...
+
         private void scrollToEnd() => ScheduleAfterChildren(() => scroll.ScrollToEnd());
 
-        protected class ChatLineContainer : FillFlowContainer<ChatLine>
+        protected class ChatLineContainer : FillFlowContainer
         {
             protected override int Compare(Drawable x, Drawable y)
```
bdach: This comparator worries me a little bit with the fallback to `base`. I'm not sure what this does, or if it's even still needed?

bdach (commented):
Looks mostly ok to me from the source alone, besides some stylistic comments.

osu.Game/Overlays/Chat/DrawableChannel.cs
EVAST9919: Couldn't find any issues by removing this

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -112,19 +135,74 @@ private void pendingMessageResolved(Message existing, Message updated)
 
         private void messageRemoved(Message removed)
         {
-            ChatLineFlow.Children.FirstOrDefault(c => c.Message == removed)?.FadeColour(Color4.Red, 400).FadeOut(600).Expire();
+            getChatLines.FirstOrDefault(c => c.Message == removed)?.FadeColour(Color4.Red, 400).FadeOut(600).Expire();
         }
 
+        private IEnumerable<ChatLine> getChatLines => ChatLineFlow.Children.OfType<ChatLine>();
```
iiSaLMaN: Seems like a wrong name for a private property, although what do you think about returning the chat line container and apply the filtering inside:
```csharp
public IEnumerable<ChatLine> ChatLines => this.OfType<ChatLine>();
public IEnumerable<DaySeperator> Seperators => this.OfType<DaySeperator>();
```

frenzibyte (changes requested):

osu.Game/Overlays/Chat/DrawableChannel.cs
iiSaLMaN: Another thing to note here this should be protected as the parent is.

bdach:
I had a look at blame in regards to the comparator, and it seems to have been added in 877c69d. I'm not entirely certain what the context was, but from what I can tell this commit added support for echoing back sent messages, so I guess that's the thing that should be checked after the removal.

peppy:
Would've been nice to see some asserts in the new (and existing) tests.

peppy (dismissed):

peppy (approved):

osu.Game/Overlays/Chat/DrawableChannel.cs
peppy: I've also double-checked the cases I remember being unreasonable and cannot seem to reproduce any more.

PR #6574, opened 2019-10-22
Fix registration textboxes always focusing after pressing escape

### LINKED ISSUE #3903: User registration form keeps focus when pressing escape
When at the email/username/password form, pressing escape will not close the form.

Joehuu:
Fixes https://github.com/ppy/osu/issues/3903.

### DIFF
--- a/osu.Game/Overlays/AccountCreation/ScreenEntry.cs
+++ b/osu.Game/Overlays/AccountCreation/ScreenEntry.cs
@@ -143,7 +143,8 @@ namespace osu.Game.Overlays.AccountCreation
             base.OnEntering(last);
             processingOverlay.Hide();
 
-            focusNextTextbox();
+            if (host?.OnScreenKeyboardOverlapsGameWindow != true)
+                focusNextTextbox();
         }
 
         private void performRegistration()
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/AccountCreation/ScreenEntry.cs
```diff
@@ -138,18 +138,12 @@ private void load(OsuColour colours, IAPIProvider api, GameHost host)
             passwordTextBox.Current.ValueChanged += password => { characterCheckText.ForEach(s => s.Colour = password.NewValue.Length == 0 ? Color4.White : Interpolation.ValueAt(password.NewValue.Length, Color4.OrangeRed, Color4.YellowGreen, 0, 8, Easing.In)); };
         }
 
-        protected override void Update()
-        {
-            base.Update();
-
-            if (host?.OnScreenKeyboardOverlapsGameWindow != true && !textboxes.Any(t => t.HasFocus))
```
bdach: The `host?.OnScreenKeyboardOverlapsGameWindow != true` check seems to have been a mobile usability improvement (see #4179 and 23d0399) - did you check that this doesn't regress behaviour?

peppy:
I think this should be fine, but the mobile conditional should be added back.

peppy (approved):

PR #6580, opened 2019-10-23
Don't snap slider control point placement

smoogipoo:
Not sure if this is the best way to handle this. May be okay since it's kinda a custom thing and generally not desired? Only 1 or 2 rulesets will do something like this.


peppy:
Needs an inline comment explaining what/why it's doing this. Very confusing without this.

peppy (approved):

PR #6581, opened 2019-10-23
Fix distance snapping grid not updating on scroll

smoogipoo:
When scrolling, the elapsed frame time is 0, so the grid doesn't get refreshed.

### DIFF
--- a/osu.Game/Rulesets/Edit/HitObjectComposer.cs
+++ b/osu.Game/Rulesets/Edit/HitObjectComposer.cs
@@ -169,10 +169,7 @@ namespace osu.Game.Rulesets.Edit
             base.Update();
 
             if (EditorClock.CurrentTime != lastGridUpdateTime && blueprintContainer.CurrentTool != null)
-            {
                 showGridFor(Enumerable.Empty<HitObject>());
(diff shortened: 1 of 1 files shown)


osu.Game/Rulesets/Edit/HitObjectComposer.cs
```diff
@@ -212,6 +217,8 @@ private void showGridFor(IEnumerable<HitObject> selectedHitObjects)
             {
                 distanceSnapGridContainer.Child = distanceSnapGrid;
                 distanceSnapGridContainer.Show();
+
+                lastGridUpdateTime = EditorClock.CurrentTime;
```
peppy: can't this just be moved outside of the `if` to avoid duplicating it in two places?

smoogipoo: The other one is in a completely separate block, I wanted to avoid potentially calling a secondary `showGridFor()` by adding it here too. With that being said, it should be moved out of the `if` block here.

smoogipoo: Misunderstood, this method is being called.

peppy (approved):

PR #6590, opened 2019-10-24
Implement slider distance snapping

smoogipoo:
I don't know why `SliderPath` returns a `double` distance. Will probably fix later.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
@@ -139,7 +139,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders
             var unsnappedPath = new SliderPath(newControlPoints.Length > 2 ? PathType.Bezier : PathType.Linear, newControlPoints);
             var snappedDistance = composer?.GetSnappedDistance((float)unsnappedPath.Distance) ?? (float)unsnappedPath.Distance;
 
-            HitObject.Path = new SliderPath(newControlPoints.Length > 2 ? PathType.Bezier : PathType.Linear, newControlPoints, snappedDistance);
+            HitObject.Path = new SliderPath(unsnappedPath.Type, newControlPoints, snappedDistance);
 
             bodyPiece.UpdateFrom(HitObject);
             headCirclePiece.UpdateFrom(HitObject.HeadCircle);
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
@@ -33,7 +33,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders
 
         private PlacementState state;
 
-        [Resolved]
+        [Resolved(CanBeNull = true)]
         private HitObjectComposer composer { get; set; }
 
         public SliderPlacementBlueprint()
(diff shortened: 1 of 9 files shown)


peppy:
I'm confused by the title of this PR. Is this supposed to be "implement slider beat snapping"?

Distance snap is snapping hitobjects a certain distance from each other. This PR seems to be quantizing sliders to be correct length in time with the beat, which I wouldn't call distance snap.

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
```diff
@@ -131,8 +134,12 @@ protected override void Update()
 
         private void updateSlider()
         {
-            var newControlPoints = segments.SelectMany(s => s.ControlPoints).Concat(cursor.Yield()).ToArray();
-            HitObject.Path = new SliderPath(newControlPoints.Length > 2 ? PathType.Bezier : PathType.Linear, newControlPoints);
+            Vector2[] newControlPoints = segments.SelectMany(s => s.ControlPoints).Concat(cursor.Yield()).ToArray();
+
+            var unsnappedPath = new SliderPath(newControlPoints.Length > 2 ? PathType.Bezier : PathType.Linear, newControlPoints);
+            var snappedDistance = composer?.GetSnappedDistance((float)unsnappedPath.Distance) ?? (float)unsnappedPath.Distance;
```
peppy: i would call this length, not distance

osu.Game/Rulesets/Edit/HitObjectComposer.cs
```diff
@@ -257,9 +257,11 @@ public void EndPlacement(HitObject hitObject)
 
         public void Delete(HitObject hitObject) => EditorBeatmap.Remove(hitObject);
 
-        public override Vector2 GetSnappedPosition(Vector2 position) => distanceSnapGrid?.GetSnapPosition(position) ?? position;
+        public override Vector2 GetSnappedPosition(Vector2 position) => distanceSnapGrid?.GetSnappedPosition(position) ?? position;
 
-        public override double GetSnappedTime(double startTime, Vector2 position) => distanceSnapGrid?.GetSnapTime(position) ?? startTime;
+        public override double GetSnappedTime(double startTime, Vector2 position) => distanceSnapGrid?.GetSnappedTime(position) ?? startTime;
+
+        public override float GetSnappedDistance(float distance) => distanceSnapGrid?.GetSnappedDistance(distance) ?? distance;
```
peppy: I'm a tad confused by this flow. This doesn't seem like the correct class to be using for this.

A visual grid is providing slider lengths makes me want to take a step back. I think I mentioned before, but I believe the beat snapping / velocity code should probably be in another place (maybe the `Beatmap` class or similar?), as it will be used in multiple cases (like the one in this PR).

peppy (changes requested):

smoogipoo:
By slider distance snapping I meant it like “snapping the distance of sliders”.

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
smoogipoo: Care to explain why? I specifically asked this IRL.

osu.Game/Rulesets/Edit/HitObjectComposer.cs
smoogipoo: I think it makes perfect sense for a distance snapping grid to take a distance and snap that intervals on the grid. This is effectively doing the same thing, except that the physical display of the distance isn’t necessarily on the grid itself.

Can you explain where else distance snap will be used, if it were to be moved in the beatmap?

Also consider that calls to it likely won’t be able to go through the beatmap but the composer as it currently is, because a keypress or setting should globally disable the distance snap.

smoogipoo:
One issue I see with this is snapping the first hitobject. Will think about it a bit more.

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderPlacementBlueprint.cs
peppy: correct me if i'm wrong, but it's snapping the length of the path, not the distance of the path from the previous hitobject.

smoogipoo:
Not going to change this PR. See: https://github.com/ppy/osu/pull/6599

peppy:
https://github.com/ppy/osu/pull/6599

PR #6591, opened 2019-10-24
Add support for Markdown inline links

bdach:
# Summary

While reviewing #6542 it became apparent that there was another Markdown link format variant, used in comments that came from the web API, called the "inline link" style. It allows to specify the tooltip title within the actual URL portion, as such:
```
[link text](https://osu.ppy.sh "tooltip text")
```
Extend the Markdown parsing regex to allow parsing these inline links. Within the parenthesis () part of the Markdown URL syntax, introduce a new capturing group:
```
    (
      \s+              // whitespace between actual URL and inline title
      (?<title>        // start of "title" named group
        ""             // opening double quote (doubled inside @ string)
        (
          [^""]        // any character but a double quote
          |            // or
          (?<=\\)      // the next character should be preceded by a \
          ""           // a double quote
        )*             // zero or more times
        ""             // closing double quote
      )
    )?                 // the whole group is optional
```
This allows for parsing the inline links as-provided by web, with support for escaping double quotes inside of the title part. Correctness is displayed by the passing tests.

Testing with #6542 additionally surfaced a latent bug leading to a crash, caused by formatted links that had URLs in the display text, for example
```
[mean example - https://osu.ppy.sh](https://osu.ppy.sh)
```
In that case the outer Markdown link would get picked up once, and then reduced to the link text when looking for other links, leading to it being picked up again the second time when the raw link is found.

To resolve that case, add a check in the raw link parsing path that ensures that the found URL is not a part of a bigger, pre-existing link.

# Remarks

I'm not currently using the `title` named group to change mouseover text as web seems to do, since as discussed in Discord it could lead to users posting links with in chat or in comments with misleading tooltip text.

### DIFF
--- a/osu.Game.Tests/Chat/MessageFormatterTests.cs
+++ b/osu.Game.Tests/Chat/MessageFormatterTests.cs
@@ -274,6 +274,54 @@ namespace osu.Game.Tests.Chat
         }
 
         [Test]
+        public void TestMarkdownFormatLinkWithInlineTitle()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "I haven't seen [this link format](https://osu.ppy.sh \"osu!\") before..." });
+
+            Assert.AreEqual("I haven't seen this link format before...", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
+            Assert.AreEqual("https://osu.ppy.sh", result.Links[0].Url);
+            Assert.AreEqual(15, result.Links[0].Index);
+            Assert.AreEqual(16, result.Links[0].Length);
+        }
+
+        [Test]
+        public void TestMarkdownFormatLinkWithInlineTitleAndEscapedQuotes()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "I haven't seen [this link format](https://osu.ppy.sh \"inner quote \\\" just to confuse \") before..." });
+
+            Assert.AreEqual("I haven't seen this link format before...", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
+            Assert.AreEqual("https://osu.ppy.sh", result.Links[0].Url);
+            Assert.AreEqual(15, result.Links[0].Index);
+            Assert.AreEqual(16, result.Links[0].Length);
+        }
+
+        [Test]
+        public void TestMarkdownFormatLinkWithUrlInTextAndInlineTitle()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "I haven't seen [https://osu.ppy.sh](https://osu.ppy.sh \"https://osu.ppy.sh\") before..." });
+
+            Assert.AreEqual("I haven't seen https://osu.ppy.sh before...", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
+            Assert.AreEqual("https://osu.ppy.sh", result.Links[0].Url);
+            Assert.AreEqual(15, result.Links[0].Index);
+            Assert.AreEqual(18, result.Links[0].Length);
+        }
+
+        [Test]
+        public void TestMarkdownFormatLinkWithMisleadingUrlInText()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "I haven't seen [https://google.com](https://osu.ppy.sh) before..." });
+
+            Assert.AreEqual("I haven't seen https://google.com before...", result.DisplayContent);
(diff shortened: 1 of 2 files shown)


osu.Game.Tests/Chat/MessageFormatterTests.cs
```diff
@@ -273,6 +273,66 @@ public void TestMarkdownFormatLinkWithEscapedAndBalancedBrackets()
             Assert.AreEqual(21, result.Links[0].Length);
         }
 
+        [Test]
+        public void TestMarkdownFormatLinkWithInlineTitle()
+        {
+            Message result = MessageFormatter.FormatMessage(new Message { Content = "I haven't seen [this link format](https://osu.ppy.sh \"osu!\") before..." });
```
Zeroks77: //Arragne 
var blub = "someStuff";
var expected = "SomeStuff";

//Act 
var result = doSomeStuff(blub)

//Assert 
Assert.Something(expected, result);
 ?


have a detailed look [here](https://docs.microsoft.com/en-us/visualstudio/test/unit-test-basics?view=vs-2019)

it would make the test more readable and mantainable

osu.Game.Tests/Chat/MessageFormatterTests.cs
```diff
@@ -273,6 +273,66 @@ public void TestMarkdownFormatLinkWithEscapedAndBalancedBrackets()
...
+
+            Assert.AreEqual("I haven't seen this link format before...", result.DisplayContent);
+            Assert.AreEqual(1, result.Links.Count);
+            Assert.AreEqual("https://osu.ppy.sh", result.Links[0].Url);
+            Assert.AreEqual(15, result.Links[0].Index);
+            Assert.AreEqual(16, result.Links[0].Length);
+        }
+
+        [Test]
+        public void TestMarkdownFormatLinkWithUrlInTextAndInlineTitle()
```
Zeroks77: Also what does the test do? If it fails, if it works, please rename the test so you can see what it should do.

Example: ifAThanReturnSomething 

The test name must make clear what is tested and how

osu.Game/Online/Chat/MessageFormatter.cs
```diff
@@ -20,7 +20,7 @@ public static class MessageFormatter
         private static readonly Regex new_link_regex = new Regex(@"\[(?<url>[a-z]+://[^ ]+) (?<text>(((?<=\\)[\[\]])|[^\[\]])*(((?<open>\[)(((?<=\\)[\[\]])|[^\[\]])*)+((?<close-open>\])(((?<=\\)[\[\]])|[^\[\]])*)+)*(?(open)(?!)))\]");
 
         // [test](https://osu.ppy.sh/b/1234) -> test (https://osu.ppy.sh/b/1234) aka correct markdown format
```
Zeroks77: Please provide a more detailed description of what the regex should do.

not everyone is able to understand regex right away.

Zeroks77 (changes requested):

osu.Game.Tests/Chat/MessageFormatterTests.cs
Zeroks77: same goes to every other Testname

osu.Game.Tests/Chat/MessageFormatterTests.cs
bdach: I am familiar with the arrange-act-assert pattern (also known as the given-when-then pattern). I use it in my own projects. If you notice these tests also are arranged similarly, but don't have the comment lines - and I don't believe adding them is so important as to break existing conventions in this repository.,

osu.Game.Tests/Chat/MessageFormatterTests.cs
bdach: Again, I am familiar with that convention and sometimes use it, but choose not to here due to pre-existing conventions. The value of such naming is disputable to me, and anyone lacking context can check out the commit log and PR description.

osu.Game/Online/Chat/MessageFormatter.cs
bdach: I'm not sure what I could explain further here without restating the regex verbatim, part-by-part, which seems counterproductive - this does not seem like the correct place to explain regular expression syntax. (And if you check the commit message, I explained the added fragment part-by-part - the explanation is in the PR description, too).

bdach:
I realised that the previous containment check was not strong enough and needs to be bumped up to an overlap test - changed and added appropriate test cases.

smoogipoo (approved):

PR #6599, opened 2019-10-25
Refactor distance snapping to be handled by HitObjectComposer

smoogipoo:
Prereqs:
- [ ] https://github.com/ppy/osu/pull/6590


peppy (approved):
seems okay for a first pass

PR #6601, opened 2019-10-25
Replace local Equatable implementations with abstract EquivalentTo

peppy:
`EquivalentTo` will be used once again in a follow-up PR. Meanwhile, this makes implementation less prone to error by forcing implementation via abstract.

### DIFF
--- a/osu.Game/Beatmaps/Formats/LegacyDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyDecoder.cs
@@ -204,8 +204,8 @@ namespace osu.Game.Beatmaps.Formats
             }
 
             public override bool EquivalentTo(ControlPoint other) =>
-                base.EquivalentTo(other)
-                && CustomSampleBank == ((LegacySampleControlPoint)other).CustomSampleBank;
+                base.EquivalentTo(other) && other is LegacySampleControlPoint otherTyped &&
+                CustomSampleBank == otherTyped.CustomSampleBank;
         }
     }
 }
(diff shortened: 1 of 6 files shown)


osu.Game/Beatmaps/Formats/LegacyDecoder.cs
```diff
@@ -203,9 +203,9 @@ public override HitSampleInfo ApplyTo(HitSampleInfo hitSampleInfo)
                 return baseInfo;
             }
 
-            public bool Equals(LegacySampleControlPoint other)
-                => base.Equals(other)
-                   && CustomSampleBank == other?.CustomSampleBank;
+            public override bool EquivalentTo(ControlPoint other) =>
+                base.EquivalentTo(other)
+                && CustomSampleBank == ((LegacySampleControlPoint)other).CustomSampleBank;
```
bdach: Not going to pretend this isn't a nitpick, but is there any reason this is a straight cast, instead of a type-safe `is` check?

peppy: Good catch, I thought the inheritance here was the other way around and that this was safe. It *is* currently safe due to `EquivalentTo` calls being very limited (and always type-checking first) but best not to do this.

huoyaoyuan: Why not making it protected? It has exact same signature with `Equals`.

huoyaoyuan: Worth `/// <inheritdoc />` on overrides, then the user won't get confused when calling on derived classes.

peppy: equal checks the time as well

smoogipoo (approved):

PR #6609, opened 2019-10-25
Hide navigation bar on android

### LINKED ISSUE #6266: Doesn't hide android navigation buttons
the game doesn't go fully fullscreen (keeps the navigation buttons on screen)

Screenshot_20190925-203858

build: 2019.925.0 (official apk)
Android 7.0 on snapdragon 430

Game4all:
- [ ] Depends on framework pr ppy/osu-framework#2927

Closes #6266 

This makes the game run in fullscreen mode, hiding the navigation bar using some `SystemUiFlags`

Video demo avalaible at: https://streamable.com/vpo78

### DIFF
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -20,6 +20,8 @@ namespace osu.Android
 
             Window.AddFlags(WindowManagerFlags.Fullscreen);
             Window.AddFlags(WindowManagerFlags.KeepScreenOn);
+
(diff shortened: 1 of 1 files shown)


peppy:
Would it not be better to set these defaults at a framework side? Can you think of a scenario we wouldn't want them?

Game4all:
Sounds good as sane default as most (to not say all) mobile games run in immersive mode nowadays (can't remember of a notable title who does not).

Game4all:
Flags have been set as sane defaults on framework side. This is no longer required.

PR #6611, opened 2019-10-25
Hide resume overlay when re-pausing gameplay

### LINKED ISSUE #6442: Orange cursor to resume can be clicked on pause overlay
**Screenshots or videos showing encountered issue:** https://puu.sh/EqquA/9a3f82537f.mp4

**osu!lazer version:** latest master

nwabear:
Fixes https://github.com/ppy/osu/issues/6442.

Added a way to cancel the resume overlay if the user is pausing while the overlay is active.


osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -239,6 +239,11 @@ public override void RequestResume(Action continueResume)
                 continueResume();
         }
 
+        public override void CancelResume()
+        {
+            ResumeOverlay.Hide();
```
bdach: I think this needs a null check, `ResumeOverlay` doesn't look to be guaranteed to be non-null (see [line 199 where it is constructed](https://github.com/ppy/osu/blob/b0e21c2749d24d58de0b4c92569244bdeed77460/osu.Game/Rulesets/UI/DrawableRuleset.cs#L199), and then also [line 232 when a resume is requested](https://github.com/ppy/osu/blob/b0e21c2749d24d58de0b4c92569244bdeed77460/osu.Game/Rulesets/UI/DrawableRuleset.cs#L232)).

bdach (commented):
One note to the code, but I don't think I see any problems with this on a conceptual level? Seemed to work okay when I tested it.

osu.Game/Rulesets/UI/DrawableRuleset.cs
nwabear: Thanks! I didn't realize that, I added a null check just now!

bdach:
I suppose it could be nice to have some visual tests here too.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -453,6 +458,8 @@ public HitWindows FirstAvailableHitWindows
         /// <param name="continueResume">The action to run when resuming is to be completed.</param>
         public abstract void RequestResume(Action continueResume);
 
+        public abstract void CancelResume();
```
bdach: This addition is going to be a part of the public API surface, since custom rulesets might want to override this, so an xmldoc here is probably necessary.

osu.Game/Screens/Play/Player.cs
```diff
@@ -443,6 +443,11 @@ public void Pause()
         {
             if (!canPause) return;
 
+            if (IsResuming)
+            {
+                DrawableRuleset.CancelResume();
+            }
```
iiSaLMaN: Passing those lines of code without a comment makes it a bit confusing as to why it's there (or for me at least), please provide a comment as to why the resume overlay should be canceled here.

frenzibyte (changes requested):

osu.Game/Screens/Play/Player.cs
iiSaLMaN: Also, resuming check doesn't seem to be required, it's fine to hide a drawable that is already hidden. (nothing will change)

nwabear: I did add the commenting, but I tried to remove the check and it threw an error with normal pausing, so I had to keep it in.

bdach: I had a look at why that fails - without the check `OsuResumeOverlay.Hide()` eats a `NullReferenceException` here at first pause:

https://github.com/ppy/osu/blob/50b674249ba921bbcf2a0145062639d62249c8c6/osu.Game.Rulesets.Osu/UI/OsuResumeOverlay.cs#L62

due to `GameplayCursor` being null. Unsure whether we want to guard for null there and remove this check, or just leave the check as-is here.

nwabear: I think it would be better to do a null check there, for no other reason than it saves a bit of space, and makes it a little more organized in the Pause method.

iiSaLMaN:
@nwabear Please refrain from constantly merging the upstream master branch into your branch unless you need that.

nwabear:
oh sorry, I won't.

osu.Game/Screens/Play/Player.cs
iiSaLMaN: I advocate for a null check as there's simply no reason for the resume overlay to be only hidden once.

peppy (dismissed):

peppy (approved):

PR #6614, opened 2019-10-25
Add temporary mobile report moratorium issue template

bdach:
Due to an overwhelming amount of mobile reports that are not actively being worked on (neither by the core team, due to more pressing priorities, nor by external contributors) and take up considerable time to manage, add an issue template that aims to enforce a temporary moratorium on accepting mobile issues.

In case that isn't enough, I myself will be setting up a saved reply with pretty much the same content to clean up anything that gets past the template.


peppy:
Might be worth mentioning that opening an issue is fine if the user plans to submit a fix themselves, or maybe just mentioning that PRs are still welcome for mobile issues.

PR #6622, opened 2019-10-26
Update OsuGameDesktop.cs

Niyizibyose:
Nightmod


bdach:
First of all, I don't understand what this change is supposed to even *be* (no description, no useful commit message). Secondly, the code you've added doesn't seem to *do* anything at all or even *be even valid C#*. Thirdly, judging by the *Verified* tag on your only commit in this PR, I assume you've made this change by committing in the github interface, which makes me think you haven't even *attempted to compile* this code.

Closing due to zero added value.

PR #6631, opened 2019-10-28
Implement ability to post new comments

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu/pull/6542

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
@@ -21,39 +21,34 @@ namespace osu.Game.Tests.Visual.Online
             typeof(DrawableComment),
             typeof(HeaderButton),
             typeof(SortTabControl),
-            typeof(ShowChildrenButton),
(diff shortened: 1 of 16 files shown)


peppy:
Split this out. ResponseContainer should be its own stand-alone PR with no prereqs, at *very least*.

bdach:
I checked this out today - only ran the code this time, as this is very hard to review in its current state (with an unmerged prerequisite and due to being large on its own anyway - please split this; I will most likely have comments to the code, judging from a very lackadaisical skim once I saw this submitted first). It seems to be working fine as far as the posting goes, but I checked only the obvious happy paths.

In honesty I really dislike how this can only be tested by hitting the actual real live API, though. I'm not convinced this should be the case. Sure, an assert-less visual test with full integration is nice to quickly decouple the component from everything else, but as for now I'm forced to actually post actual real comments - and that's mostly why I'm not too comfortable *actually* trying to break it.

EVAST9919:
Closing until https://github.com/ppy/osu/pull/6542 is being merged

PR #6634, opened 2019-10-28
Add editor timing screen

peppy:
This is a first stage implementation with support for adding / removing control points (and associated groups). It does not yet support writing to parameters of control points, only displaying them.

- [x] Depends on https://github.com/ppy/osu/pull/6603


osu.Game/Screens/Edit/Timing/SampleSection.cs
```diff
@@ -0,0 +1,47 @@
...
+                SampleBank = reference.SampleBank,
+                SampleVolume = reference.SampleVolume,
+            };
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            ControlPoint.BindValueChanged(point =>
```
smoogipoo: Not a super big deal for now but I believe this can be refactored better by way of exposing an `OnControlPointChanged()` method from the base class. Reasoning being is because I'm pretty sure this will already be handled incorrectly due to not having `runOnceImmediately = true` here, and every usage would have to know about that.

smoogipoo (dismissed):

smoogipoo (approved):

PR #6635, opened 2019-10-28
Bindable control point properties

peppy:
Preparation for use in editor


smoogipoo:
Would be nice if there was a beatmap parsing test case for that last commit. Only found it by processing 200K beatmaps...

smoogipoo (approved):

PR #6636, opened 2019-10-28
Make OsuButton correctly block hover events

peppy:


bdach (approved):
Looks straightforward, correct and I understand the reason for this after checking out the dependent PR, so I feel pretty comfortable giving this an approve 👍

smoogipoo (approved):

PR #6638, opened 2019-10-28
Add ability to edit individual ControlPoint attributes from timing screen

peppy:
- [x] Depends on #6637.
- [x] Depends on #6636.
- [x] Depends on #6635.
- [x] Depends on #6634.


smoogipoo:
Requires conflict resolution.

smoogipoo (approved):

PR #6639, opened 2019-10-28
Fix download tracking components getting stuck on import failures

peppy:


peppy:
Can be tested by attempting to download an old graveyarded beatmap (try searching osu!direct for `53412` with filter set to "any).

smoogipoo (approved):

Issue #6643, opened 2019-10-28
Move IntroSequence out of OsuLogo


peppy:
Should now be in `IntroCircles`.

iiSaLMaN:
I think this issue should not be considered a "good first issue" due to the difficulty of matching intro timing correctly.

peppy:
As an aside, a 1:1 refactor moving the intro should not change any timing as it is just moving code.

PR #6644, opened 2019-10-28
Fix game-wide performance drop when triangles intro is used

peppy:
This is a hotfix. More correct fix will come via https://github.com/ppy/osu/issues/6643.


frenzibyte (approved):
Seems good for now.

Joehuu (approved):

smoogipoo (approved):

PR #6653, opened 2019-10-29
Fix day separators never being removed from chat history

peppy:
Separators would line up in a row and look silly.

- [x] Depends on https://github.com/ppy/osu/pull/6649 due to restructuring of nearby code.

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneStandAloneChatDisplay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneStandAloneChatDisplay.cs
@@ -9,6 +9,7 @@ using osuTK;
 using System;
 using System.Linq;
 using osu.Framework.Graphics.Containers;
+using osu.Game.Graphics.Containers;
 using osu.Game.Overlays.Chat;
 
 namespace osu.Game.Tests.Visual.Online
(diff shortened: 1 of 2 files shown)


osu.Game.Tests/Visual/Online/TestSceneStandAloneChatDisplay.cs
```diff
@@ -137,6 +138,17 @@ protected override void LoadComplete()
                         });
                 }, Channel.MAX_HISTORY / messages_per_call + 5);
 
+            AddAssert("Ensure no adjacent day separators", () =>
+            {
+                var indices = chatDisplay.FillFlow.OfType<DrawableChannel.DaySeparator>().Select(ds => chatDisplay.FillFlow.IndexOf(ds));
```
bdach: Insanely nitpicky, since it's test code and so it doesn't matter, but I guess you could do it like this:
```suggestion
                var indices = chatDisplay.FillFlow.OfType<DrawableChannel.DaySeparator>().Select((_, index) => index);
```

bdach (approved):
Does what it says on the label, I don't see any faults with it either 👍

osu.Game.Tests/Visual/Online/TestSceneStandAloneChatDisplay.cs
bdach: The above is incorrect. Brain lapse, sorry.

smoogipoo (approved):

PR #6658, opened 2019-10-29
Fix 1x1 white pixel appearing in the centre of hitcircles on default skin

peppy:


bdach (approved):
Seems gone for me 👍 

image

smoogipoo (approved):

Issue #4086, opened 2019-01-17
User login/oauth failures provide no feedback


GSculerlor:
**Describe the feature:** Basically right now, no feedback or any indicator why user can't login in-game (like wrong password, can't connect to server, etc)

**Proposal designs of the feature:** Adding a notification or maybe pop-up, or the simplest one is an message on login box that indicate why user can't login. Using popup dialog like when user went to registration is quite too much tho

yusdacra:
Something like [this](https://streamable.com/nhkmz) maybe? I could try adding this, seems easy enough. But i have yet to understand most of the codebase so it will take time for me to do it.

GSculerlor:
Maybe like that one, but not sure pushing notification panel when login window is active is correct behavior

ticotaco72:
@arfylte should say what is correct.

PR #6662, opened 2019-10-29
Add feedback to user when sign in form isn't filled

GSculerlor:
Basically just unifying the behavior with register form. When the username or password text box isn't filled it will trigger the shake container.
Partially working on #4086 to give user more feedback on sign in flow, will do the error from API response on separate PR

https://streamable.com/ky8nz


bdach (commented):
I think this is okay? Works for sure. Not too certain why two nested containers are necessary here, but I might just not be well-versed in the layouting, since I've never written one before.

smoogipoo:
Nested containers are required because this is a fill flow, which doesn't support drawables having custom positions (since it sets their positions).

smoogipoo (approved):

PR #6663, opened 2019-10-29
Disallow voting for own comments

EVAST9919:
Depends on
- [x] https://github.com/ppy/osu/pull/6749


osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -69,6 +70,8 @@ protected override void LoadComplete()
             votesCount.Value = comment.VotesCount;
             isVoted.BindValueChanged(voted => background.Colour = voted.NewValue ? AccentColour : OsuColour.Gray(0.05f), true);
             votesCount.BindValueChanged(count => votesCounter.Text = $"+{count.NewValue}", true);
+
+            api.LocalUser.BindValueChanged(user => disabled = user.NewValue?.Id == comment.UserId ? true : false, true);
```
huoyaoyuan: I can't imagine how `? true : false` is meanful.

huoyaoyuan (changes requested):

osu.Game/Overlays/Comments/VotePill.cs
EVAST9919: oopsie

osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -145,16 +148,30 @@ protected override void OnLoadFinished()
 
         protected override bool OnHover(HoverEvent e)
         {
+            if (disabled)
+                return false;
```
bdach: I'm not sure if I wouldn't prefer `true` to be returned here to prevent events from propagating further. It doesn't seem to interfere with drag in my testing.

osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -43,6 +43,7 @@ public class VotePill : LoadingButton, IHasAccentColour
 
         private readonly BindableBool isVoted = new BindableBool();
         private readonly BindableInt votesCount = new BindableInt();
+        private bool disabled;
```
bdach: I don't like that this exists. I understand why, but this at least needs to be negated (because now you have `Enabled` in the base and `disabled` in the derived), renamed (maybe something like `canVote`), or even removed (you should be able to just set `Enabled` to false in the ctor - if it's going to start disabled, it won't ever get to the loading part, so it shouldn't interfere - this might warrant leaving a comment in source, though).

bdach:
Thoughts on the following refactoring proposal?

```diff
diff --git a/osu.Game/Overlays/Comments/VotePill.cs b/osu.Game/Overlays/Comments/VotePill.cs
index a895ad2..4f8c179 100644
--- a/osu.Game/Overlays/Comments/VotePill.cs
+++ b/osu.Game/Overlays/Comments/VotePill.cs
@@ -62,7 +62,7 @@ private void load(OsuColour colours)
             AccentColour = borderContainer.BorderColour = sideNumber.Colour = colours.GreenLight;
             hoverLayer.Colour = Color4.Black.Opacity(0.5f);
 
-            disabled = api.LocalUser.Value.Id == comment.UserId;
+            Enabled.Value = api.LocalUser.Value.Id != comment.UserId;
         }
 
         protected override void LoadComplete()
@@ -142,36 +142,21 @@ protected override void OnLoadFinished()
         {
             votesCounter.FadeIn(duration, Easing.OutQuint);
 
-            if (IsHovered)
-                onHoverAction();
+            onHoverAction();
         }
 
         protected override bool OnHover(HoverEvent e)
         {
-            if (disabled)
-                return false;
-
             onHoverAction();
             return base.OnHover(e);
         }
 
         protected override void OnHoverLost(HoverLostEvent e)
         {
-            if (disabled)
-                return;
-
             updateDisplay();
             base.OnHoverLost(e);
         }
 
-        protected override bool OnClick(ClickEvent e)
-        {
-            if (disabled)
-                return false;
-
-            return base.OnClick(e);
-        }
-
         private void updateDisplay()
         {
             if (isVoted.Value)
@@ -187,7 +172,7 @@ private void updateDisplay()
 
         private void onHoverAction()
         {
-            if (!IsLoading)
+            if (Enabled.Value && !IsLoading)
                 updateDisplay();
         }
```
Could even move the `Enabled` check further down to `updateDisplay()`.

smoogipoo:
Needs a test.

osu.Game.Tests/Visual/Online/TestSceneVotePill.cs
```diff
@@ -0,0 +1,65 @@
...
+
+            AddStep("User comment", () => addVotePill(userComment));
+            AddStep("Click", () => votePill.Click());
+            AddAssert("Not loading", () => !votePill.IsLoading);
+
+            AddStep("Random comment", () => addVotePill(randomComment));
+            AddStep("Click", () => votePill.Click());
+            AddAssert("Loading", () => votePill.IsLoading);
+
+            AddStep("Log out", API.Logout);
```
bdach: Due to this logout not having a matching login at the start, this test is not idempotent (if you run the test scenario twice, it'll succeed the first time but then fail the second time due to staying logged in).

osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -60,6 +58,9 @@ private void load(OsuColour colours)
         {
             AccentColour = borderContainer.BorderColour = sideNumber.Colour = colours.GreenLight;
             hoverLayer.Colour = Color4.Black.Opacity(0.5f);
+
+            if (api.LocalUser.Value.Id != comment.UserId && api.LocalUser.Value.Id != 1)
```
bdach: What's the `!= 1` check for? I don't remember seeing that before, is local user guaranteed to always have ID 1?

osu.Game/Overlays/Comments/VotePill.cs
```diff
@@ -157,15 +158,18 @@ protected override void OnHoverLost(HoverLostEvent e)
 
         private void updateDisplay()
         {
-            if (isVoted.Value)
+            if (Action != null)
```
bdach: A nitpick, but I'd invert this (check for null and immediately return if null, like a guard). Reduces nesting and makes it a bit easier to read for me.

bdach (commented):
The tests are definitely nice, but this could do with some more improvements in my opinion.

osu.Game/Overlays/Comments/VotePill.cs
EVAST9919: Local (logged-out) user has ID = 1

bdach: Okay, I suppose this is fine then (there seems to be no online user with id 1 anyways).

bdach (dismissed):
I think this should be ok now 👍

osu.Game/Overlays/Comments/VotePill.cs
peppy: `api.IsLoggedIn`?

osu.Game.Tests/Visual/Online/TestSceneVotePill.cs
```diff
@@ -0,0 +1,71 @@
...
+        private void load()
+        {
+            var randomComment = new Comment
+            {
+                IsVoted = false,
+                UserId = 4444,
+                VotesCount = 2,
+            };
+
+            AddStep("Log in", () => API.LocalUser.Value = new User
```
peppy: can you split these three tests up until `[Test]` methods with names explaining what they are testing?

peppy (changes requested):

osu.Game.Tests/Visual/Online/TestSceneVotePill.cs
```diff
@@ -21,6 +22,11 @@ public class TestSceneVotePill : OsuTestScene
             typeof(VotePill)
         };
 
+        protected override bool UseOnlineAPI => true;
```
peppy: why are we using online api now? this is going to be triggering login attempts on the actual site, no? which could get IPs blocked and trigger security countermeasures.

how about fixing `IsLoggedIn` in `DummyAPIAccess` instead rather than working around the true issue.

peppy (approved):

PR #6666, opened 2019-10-30
Fix crash when quickly exiting multiplayer after joining a room

### LINKED ISSUE #6664: Leaving a multiplayer lobby while loading causes a crash
**Describe the crash:** 
Join a lobby and quickly press the back button. 

https://streamable.com/jb4iw

**osu!lazer version:** 
commit e1d0d78

**Logs:** 
[runtime.log](https://github.com/ppy/osu/files/3785573/runtime.log)

**Computer Specifications:** 
i3 6100 8GB RAM NVIDIA GTX1050Ti Windows 10 x64 1909

smoogipoo:
Fixes #6664

This stops the callbacks from occurring after multiplayer has been exited. This doesn't ensure the room is actually parted, but that's not too big of an issue right now and will be fixed with a refactor of how `PartRoom()` works - this code will remain the same in any case.

### DIFF
--- a/osu.Game/Screens/Multi/Multiplayer.cs
+++ b/osu.Game/Screens/Multi/Multiplayer.cs
@@ -212,6 +212,8 @@ namespace osu.Game.Screens.Multi
 
         public override bool OnExiting(IScreen next)
         {
+            roomManager.PartRoom();
+
             if (screenStack.CurrentScreen != null && !(screenStack.CurrentScreen is LoungeSubScreen))
             {
                 screenStack.Exit();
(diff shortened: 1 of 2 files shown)


bdach (commented):
Checked this out; the original bug was quite finicky, but I got it to crash a couple of times on `master`. Haven't been able to get any crashes with this fix. I added a temporary debug statement to see if requests ever get canceled when I was trying, and I saw some cancels, so I suppose this is ok 👍

peppy (approved):

PR #6668, opened 2019-10-30
Share framework file-exclusion function

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/2933.
- [x] Depends on framework version bump.


bdach (commented):
Guess I was too slow to submit this 😄 Locally tests pass with local framework with the prereq, so this looks good to go along with the next framework bump.

smoogipoo (approved):

PR #6671, opened 2019-10-30
Port merged tooling updates from framework

huoyaoyuan:
Simple port of ppy/osu-framework#2896, ppy/osu-framework#2906, ppy/osu-framework#2935 and part of ppy/osu-framework#2921.


huoyaoyuan:
CodeFactor claims for `build.sh`, but it doesn't triggers for framework.

bdach (commented):
A bit tiring to review manually, but on a cursory glance looks correct.

smoogipoo:
The blocker on this for us right now is having rider configs - maybe it should be copied from the osu.Desktop configs.

huoyaoyuan:
> The blocker on this for us right now is having rider configs - maybe they should be copied from the existing configs.

Which setting is blocking? I merged the three setting files, and the merge tool changed some tabs. Or may I do a manual merge only for custom dictionary?

smoogipoo:
It’s missing the run configs for the filtered solutions. May need me or peppy to go through it if you don’t have rider.

huoyaoyuan:
I do have rider, but only use it for feature parity checking. Please help to add them because I'm not familiar.

bdach:
I think if you copy the entire `.idea/.idea.osu/.idea/runConfigurations` folder over to `.idea/.idea.osu.Desktop/.idea/runConfigurations` they'll show up when opening `osu.Desktop.slnf`. It's not ideal, I'd make that a symlink, but Windows. Also possibly debatable if having run configs for the full solution is necessary any longer.

peppy (approved):

smoogipoo (approved):

PR #6679, opened 2019-10-31
Implement slider control point deletion

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/6678

https://streamable.com/llscc


osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
```diff
@@ -70,5 +77,49 @@ private void selectPiece(int index)
                     piece.IsSelected.Value = piece.Index == index;
             }
         }
+
+        protected override bool OnKeyDown(KeyDownEvent e)
+        {
+            switch (e.Key)
```
peppy: wonder if we should be using `PlatformAction` here?

swoolcock: I agree, since macOS apps tend to respond to backspace (also labelled delete on a Mac keyboard) in a similar destructive fashion to Windows.

Also note that the iPad Smart Keyboard does not have a key that will trigger this enum case, only the `Backspace’ case.

peppy (approved):

PR #6682, opened 2019-10-31
Improve issue templates

jai-x:
Reorders the issue templates such that the warning about mobile issues is first to make it more visible.

Adds an issue template config that links to the [osu-stable-issues](https://github.com/ppy/osu-stable-issues) repo osu!stable issues and disables blank issues from being opened.

Removes "Missing for live" issue template as it's no longer needed.

---

*Before*: 
image

---

*After*: 
image


peppy:
"missing for live" can be removed too

smoogipoo (approved):

PR #5574, opened 2019-08-02
Improve iOS importing and file system

dtsdao:
- Fixes #5545.
- [x] Depends on https://github.com/ppy/osu/pull/6685.

### Custom URL Scheme

- Added `HandleUrl` function in `OsuGame`

- Added custom URL types

- Might be easier to complete this part on other platforms?

### Fix importing issue

- Fixes #5500. Fixes #5485.

- Enable "import from stable"

- Be able to open osu! files

Be tested on iOS 12.4 Simulator and real device.

## Function preview:

https://streamable.com/bf0lw

PR #6685, opened 2019-11-01
Move link handling code to OsuGame

peppy:
Split out of https://github.com/ppy/osu/pull/5574.

Added protection for potentially being called before the game / gamehost / target overlay are in a correct state.

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -215,16 +215,20 @@ namespace osu.Game
 
         private ExternalLinkOpener externalLinkOpener;
 
-        public void HandleLink(string url) => HandleLink(MessageFormatter.GetLinkDetails(url));
+        /// <summary>
+        /// Handle an arbitrary URL. Displays via in-game overlays where possible.
+        /// This can be called from a non-thread-safe non-game-loaded state.
+        /// </summary>
+        /// <param name="url">The URL to load.</param>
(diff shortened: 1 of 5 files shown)


bdach (dismissed):
I think this makes the code look better overall. Shame it makes `OsuGame` grow in size a little, but not much that can be done about that I guess.

peppy:
If we get to the point we want to move this out, a `static class LinkRouter { void Route(string link, OsuGame game) { ... }` could work, since the target functions are all `public`

osu.Game/OsuGame.cs
```diff
@@ -215,31 +215,102 @@ private void load(FrameworkConfigManager frameworkConfig)
 
         private ExternalLinkOpener externalLinkOpener;
 
-        public void OpenUrlExternally(string url)
+        /// <summary>
+        /// Handle an arbitrary URL. Displays via in-game overlays where possible.
+        /// This can be called from a non-thread-safe non-game-loaded state.
+        /// </summary>
+        /// <param name="url">The URL to load.</param>
+        public void HandleLink(string url) => Schedule(() => HandleLink(MessageFormatter.GetLinkDetails(url)));
```
smoogipoo: The schedule should exist on the other method. This also means that the other method isn't actually thread-safe.

smoogipoo (changes requested):

osu.Game/OsuGame.cs
peppy: Fair call

smoogipoo (approved):

PR #6690, opened 2019-11-01
Moving intro sequence from osu logo to intro circles

ghost:
Moved IntroSequence to IntroCircles from OsuLogo in an attempt to solve issue #6643 .


bdach:
You've made all commits using the github interface - did you test this or try to run this at all? Leaving aside the broken whitespace reported by CI.

(You shouldn't be using those default commit messages at all, too, they're noise - `git diff` is for checking which files a commit touched, not the message.)

peppy:
This doesn't even do what it says it does.

PR #6691, opened 2019-11-01
Add retry button and hotkey support to results screen

### LINKED ISSUE #6439: Result screen should have a restart button
**Describe the new feature:** 
Just like on stable you should have the ability to restart a map from the result screen

osu_2019-10-08_23-21-25

peppy:
- Closes https://github.com/ppy/osu/issues/6439.
- [x] Depends on https://github.com/ppy/osu/pull/6689.

### DIFF
--- a/osu.Game/Screens/Ranking/Results.cs
+++ b/osu.Game/Screens/Ranking/Results.cs
@@ -19,6 +19,7 @@ using osu.Game.Graphics;
 using osu.Framework.Graphics.Shapes;
 using osu.Game.Graphics.Sprites;
 using osu.Game.Scoring;
+using osu.Game.Screens.Play;
 
 namespace osu.Game.Screens.Ranking
 {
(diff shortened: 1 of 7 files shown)


osu.Game/Screens/Ranking/Results.cs
```diff
@@ -100,10 +104,7 @@ public override void OnEntering(IScreen last)
 
         public override bool OnExiting(IScreen next)
         {
-            allCircles.ForEach(c =>
-            {
-                c.ScaleTo(0, transition_time, Easing.OutSine);
-            });
+            allCircles.ForEach(c => { c.ScaleTo(0, transition_time, Easing.OutSine); });
```
iiSaLMaN:
```suggestion
            allCircles.ForEach(c => c.ScaleTo(0, transition_time, Easing.OutSine));
```

frenzibyte (changes requested):

bdach:
Both happy paths seem to work nicely, and I got to play some while testing this, so definitely a fun one to review 😄 

I've got one small visual issue though - it's very likely out of scope of this, but I'm not too keen on the fact that the fade to black when using the hotkey doesn't cover the top bar ([see video](https://streamable.com/dku5b)). To me it's especially jarring on beatmaps with light backgrounds.

peppy:
That is indeed weird. I'd probably open that in a separate issue as it's a bit out of scope of this change.

bdach (dismissed):

smoogipoo (dismissed):

PR #6694, opened 2019-11-01
Move circles intro sequence out of OsuLogo

### LINKED ISSUE #6643: Move IntroSequence out of OsuLogo

frenzibyte:
Closes #6643


smoogipoo:
The animation seems to have changed a little bit as the OsuLogo starts appearing.

iiSaLMaN:
@smoogipoo Should be more correct now. (I assume you're talking about the logo appearing before the 4 circles animation are complete)

osu.Game/Screens/Menu/IntroCircles.cs
```diff
@@ -15,8 +22,8 @@ public class IntroCircles : IntroScreen
 
         protected override string BeatmapFile => "circles.osz";
 
-        private const double delay_step_one = 2300;
-        private const double delay_step_two = 600;
+        private const double load_menu_delay = 2400;
```
peppy: can you explain the 100ms change here?

iiSaLMaN: Seems that the current load menu delay is a bit earlier with my new changes (start intro at sequence load completion), so I optimized the delay a bit, should be no harm?

peppy: When variables change, I'd want to understand why, not just "seems to be different".

bdach:
Here's [a comparison video I took](https://streamable.com/2zxxb) (1920x1080@60fps, played at 25% speed). There are some timing differences, unsure if actual ones or it's just a discrepancy caused by frame samples not being synced. Don't see any visual problems.

osu.Game/Screens/Menu/IntroCircles.cs
peppy: @iiSaLMaN i hope you understand i *cannot* go further without this being resolved. are you planning on looking into this?

iiSaLMaN: yeah sorry for this delay. I didn't have time to look into this, will check it after finishing from other PR reviews atm.

iiSaLMaN:
After doing a very deep search in the circles intro sequence changes I made (by trial-and-error), I've found out that the timing differences (especially the `delay_step_one`/`load_menu_delay` change) occurred because of loading the circles intro sequence in the intro screen other than in `OsuLogo`, but got nowhere after that with no knowledge towards the timing change, I've even printed out the current time of the clock and transform start time and they were almost the same (with 5 ~ 10 ms change).

Closing as I don't have the time to work on this, will return to this in the near future if not fixed yet.

PR #6699, opened 2019-11-03
Fix slider creation regressing with path selection changes

peppy:
Regressed in https://github.com/ppy/osu/pull/6678.


peppy:
This breaks clicking to drag when over the selection area of another slider. Which says to me that all of these things need better tests in place, since it's very easy to regress intertwining behaviours.

smoogipoo (approved):

PR #6700, opened 2019-11-03
Fix mod section overflowing mod select overlay at higher ui scale

Joehuu:
Before:
osu_2019-11-03_07-27-06

After:
osu_2019-11-03_07-26-39

Not sure how to fix this pop in/out transition at certain resolutions (spacing changes in the fill flow right now).


frenzibyte (approved):

bdach:
I took a look at this, and it looks like this might be a framework bug; the draw visualiser seems to show that the child `FillFlowContainer` has its width intact when its `X` position is being transformed, but the children within adhere to the `FillFlowContainer`'s _parent's_ width ([see video](https://streamable.com/wujfb) - right edge of the containers is the interesting part there).

Unsure whether this was intended/the visualiser is confused or this is an actual bug, however.

Joehuu:
> spacing changes in the fill flow right now

https://github.com/ppy/osu/blob/7d5b6226a9ee97bcfdbd2b1d672e68cbeab1bea3/osu.Game/Overlays/Mods/ModSelectOverlay.cs#L329

Let me clarify on this, there is also a spacing transform on pop in/out and that affects the fill flow container’s real width (i.e. the width the children are taking, not the relative size that makes it flow). Removing the spacing transforms will fix the problem as the real width will now be the same.

bdach:
I was aware of the transition, but the flow break looked like it was respecting the parent width. I suppose I might have mistakenly extrapolated.

I experimented with applying transformable negative padding too, which seemed to fix it in some cases, but the big problem is how much of it to actually apply - that would depend on the number of items in a row, which seems wrong to rely on/try to even access since it's a flow container...

peppy:
I think the spacing transform issue is best tracked separately. This PR fixes a larger usability issue which we should get in ASAP.

peppy (approved):

PR #6716, opened 2019-11-05
CommentsContainer improvements

EVAST9919:
Includes:
- ability test local comment bundles
- placeholder for "no comments yet" case
- reload comments on user change
- works for logged-out users
- better loading animation
- added `TotalCommentsCounter`
- fix async load wasn't really async


osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -40,11 +38,16 @@ public class Comment
         [JsonProperty(@"votes_count")]
         public int VotesCount { get; set; }
 
+        public CommentableType CommentableType;
+
         [JsonProperty(@"commenatble_type")]
-        public string CommentableType { get; set; }
+        private string commentableType
+        {
+            set => Enum.TryParse(value.Pascalize().Replace("_", string.Empty), true, out CommentableType);
```
bdach: Pretty sure [`.Pascalize()` will already remove underscores](https://humanizr.net/#pascalize), so the `.Replace()` is gratuitous.

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -70,12 +73,10 @@ public class Comment
 
         public bool IsDeleted => DeletedAt.HasValue;
 
-        public bool HasMessage => !string.IsNullOrEmpty(MessageHtml);
-
-        public bool IsVoted { get; set; }
+        public bool IsReply => ParentId.HasValue;
 
-        public string GetMessage => HasMessage ? WebUtility.HtmlDecode(Regex.Replace(MessageHtml, @"<(.|\n)*?>", string.Empty)) : string.Empty;
+        public bool HasMessage => !string.IsNullOrEmpty(Message);
```
bdach: Seems sort of out-of-scope, but ok.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -117,39 +136,69 @@ public CommentsContainer(CommentableType type, long id)
...
+        {
+            if (parameters == null)
+                return;
+
+            request = new GetCommentsRequest(parameters, Sort.Value, currentPage++);
+            request.Success += onSuccess;
+            api.Queue(request);
         }
 
         private void onSuccess(CommentBundle response)
```
bdach: Maybe not necessarily relevant right here, but this handler is getting really big and its body is hard to parse. Would be useful to split it into smaller, more well-defined methods, to be able to tell quicker what is going on.

osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
```diff
@@ -30,30 +34,117 @@ public class TestSceneCommentsContainer : OsuTestScene
...
 
-            AddStep("Airman comments", () =>
-            {
-                scrollFlow.Clear();
-                scrollFlow.Add(new CommentsContainer(CommentableType.Beatmapset, 24313));
-            });
+        [BackgroundDependencyLoader]
+        private void load(IAPIProvider api)
+        {
+            AddStep("trigger logout", api.Logout);
```
bdach: You're using the actual online API, and adding a step to logout to check comment availability when logged out. However, due to the fact you're using the actual online API, you actually *can't log back in again in this same test*. Moreover, *this silently logs you out from the actual lazer client*. This level of state pollution is unacceptable in tests as far as I'm concerned.

What this basically means is this is too tightly coupled to online components now to be testable, which is not good. [There were warnings about that too, by the way.](https://github.com/ppy/osu/pull/6504#issuecomment-542063407)

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -40,11 +38,16 @@ public class Comment
         [JsonProperty(@"votes_count")]
         public int VotesCount { get; set; }
 
+        public CommentableType CommentableType;
+
         [JsonProperty(@"commenatble_type")]
```
bdach: Must have missed this before - how is/was this working?
```suggestion
         [JsonProperty(@"commentable_type")]
```

bdach (changes requested):
Again not sure if this is ready to go now in your opinion or not, but since you requested a review...

osu.Game/Online/API/Requests/Responses/Comment.cs
EVAST9919: I never used it so yeah, good catch

osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
EVAST9919: I just splitted the test scene into two: online and offline ones. In online you can just check whether all the stuff is working and how it looks and for offline you can test your local bundles

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -117,41 +138,93 @@ public CommentsContainer(CommentableType type, long id)
...
         {
             background.Colour = colours.Gray2;
+            placeholderBackground.Colour = colours.Gray3;
         }
 
         protected override void LoadComplete()
         {
-            Sort.BindValueChanged(onSortChanged, true);
+            Sort.BindValueChanged(_ => updateComments());
+            api.LocalUser.BindValueChanged(_ => updateComments());
```
bdach: Just noticed - isn't this a memory leak? You should probably make a local bindable and bind to `api.LocalUser`.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -117,41 +138,93 @@ public CommentsContainer(CommentableType type, long id)
...
+        {
+            clearComments();
+
+            if (parameters == null)
+            {
+                if (commentBundle != null)
+                    onSuccess(commentBundle);
+
+                return;
+            }
```
bdach: I don't like that this silently returns if both `parameters` and `commentBundle` are null. I'm pretty sure one of them is guaranteed to be non-null, right? It would be better to add an assert for that.

EVAST9919: Not really. They both can be null if you are in idle state and press `sort` buttons. It will trigger `updateComments` but since there's nothing to update it will just returns

EVAST9919: Also using that we can use `ShowComments(null)` and that will just trigger idle state

bdach: To tell you the truth I'm not too hot on this myself... Passing `null` in ctor shouldn't be a thing.

EVAST9919: So throwing an exception if `commentBundle` is null is a good way to handle this, right?

EVAST9919:
You can see advantages of splitting `CommentsContainer` to Online and Offline ones in https://github.com/ppy/osu/pull/6968

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
```diff
@@ -0,0 +1,60 @@
...
+    public class OnlineCommentsContainer : CommentsContainer
+    {
+        private GetCommentsRequest request;
+        private int currentPage;
+        private CommentBundleParameters parameters;
+
+        public void ShowComments(CommentableType type, long id)
+        {
+            parameters = new CommentBundleParameters(type, id);
+            Sort.TriggerChange();
```
smoogipoo: The way this class is implemented is pretty weird. You're doing calls on members of the base class that seem like they should go in the base class themselves. Why is this done here?

Furthermore, why are the implementations of `OnSortChanged()`, `OnUserChanged()` and `OnLoadStarted()` not in the base class?

This all may be simplified via a better implementation of the retrieval in the first place. Consider looking at `Leaderboard` as a reference of the implementation of comment retrieval. In general - expose a `FetchComments()` method, and do the success/cancelling stuff in the base class.

You'll also find that you may not need to expose the methods listed above through this refactoring.

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            request = new GetCommentsRequest(parameters, Sort.Value, ++currentPage);
+            request.Success += response =>
+            {
+                if (currentPage == 1)
+                    ResetComments(response);
+                else
+                    AddComments(response, false);
+            };
+            Task.Run(() => request.Perform(API));
```
smoogipoo: `API.Queue(request)`

smoogipoo (changes requested):

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
EVAST9919: No. `Api.Queue()` doesn't work for logged-out users

osu.Game/Online/API/Requests/Responses/Comment.cs
```diff
@@ -40,11 +38,16 @@ public class Comment
...
         public int VotesCount { get; set; }
 
-        [JsonProperty(@"commenatble_type")]
-        public string CommentableType { get; set; }
+        public CommentableType CommentableType;
+
+        [JsonProperty(@"commentable_type")]
+        private string commentableType
+        {
+            set => Enum.TryParse(value.Pascalize(), true, out CommentableType);
```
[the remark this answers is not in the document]
EVAST9919: Used this as a reference
https://github.com/ppy/osu/blob/db34ea88bc26626b8ffc0e0e64f00e1507f2c455/osu.Game/Online/API/Requests/Responses/APIKudosuHistory.cs#L56

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
EVAST9919: We can't do much with sort for "offline mode" (when we only have a local bundle)
The overall idea was just move all the online logic outside of the class itself (in tjis case to it's sub-class)

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
smoogipoo: Ah. Probably a good time to make a method on `IAPIProvider` for this sort of functionality, like `PerformAsGuest()`.

osu.Game/Online/API/Requests/Responses/Comment.cs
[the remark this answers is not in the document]
EVAST9919: Using your suggestion `CommentableType` for news post is trying to be `news_post`, when it must be `NewsPost`

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
EVAST9919: > Furthermore, why are the implementations of OnSortChanged(), OnUserChanged() and OnLoadStarted() not in the base class?
`OnSortChanged` do nothing in the base class when it fetches new Bundle in Online one
`OnUserChanged` reloads the same comment bundle in the base class and refetches new Bundle in Online one
`OnLoadStarted` in the `OnlineCommentsContainer` is just an expansion (as you can see by using `base.OnLoadStarted`)
> do the success/cancelling stuff in the base class.
The overall idea was just move all the online logic outside of the class itself (in this case to it's sub-class)

osu.Game/Online/API/Requests/Responses/Comment.cs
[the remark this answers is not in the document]
smoogipoo: I'd see a use for having a converter similar to `StringEnumConverter` that does this process for us, since this isn't going to be the only place where this is done.

osu.Game/Overlays/Comments/OnlineCommentsContainer.cs
smoogipoo: Well all comments are online comments. The only reason the base class needs to be provided without online functionality is for testing purposes (although in the future even that will be arguable).
So I'd like to see the logic unified - always refetch when sort is changed, always refetch when the user is changed, always cancel request / reset state in `OnLoadStarted()`.

EVAST9919: Okay, I'll remove separated `OnlineCommentsContainer` and will make a single one again

bdach: I don't think merging the containers back again is necessary for what @smoogipoo wants to be done here, if I understood correctly.

smoogipoo: Correct, I just want the logic to be standardised. The online one can still exist for performing requests, but that's _all_ it should be doing, and exposing the requests to the base class via APIRequest or similar (again, look at `Leaderboard` for an example).

EVAST9919: Should be better now. I tried to mimic `Leaderboard` logic

osu.Game/Overlays/Comments/CommentBundleParameters.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Game.Online.API.Requests;
+
+namespace osu.Game.Overlays.Comments
+{
+    public class CommentBundleParameters
```
peppy: missing xmldoc. i have no idea was "Id" is.

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -26,7 +26,7 @@ public List<Comment> Comments
                         {
```
peppy: this is very weird logic. there's no need to `ForEach` since there should be either zero or one match. Try using `FirstOrDefault` instead?

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -47,17 +47,27 @@ public List<Comment> Comments
...
+            get => userVotes;
+            set
             {
-                Comments.ForEach(c =>
+                userVotes = value;
+                userVotes.ForEach(v =>
                 {
-                    if (v == c.Id)
-                        c.IsVoted = true;
+                    foreach (var c in Comments)
```
peppy: not sure if this can be easily resolved, but this could nullref depending on the order in which `Comments` and `UserVotes` are initialised.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -13,37 +12,82 @@
...
+using osu.Game.Users;
+using osu.Game.Graphics.UserInterface;
+using osu.Game.Graphics.Sprites;
+using osu.Framework.Threading;
+using System;
 
 namespace osu.Game.Overlays.Comments
 {
-    public class CommentsContainer : CompositeDrawable
+    public abstract class CommentsContainer : CompositeDrawable
```
peppy: making this abstract *just* for testing feels very very backwards.

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: So, there's no reason to make a separate `OnlineCommentsContainer` at all or I should make the APIRequest virtual with null in a base one?

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
EVAST9919: So far there were no such cases

osu.Game/Overlays/Comments/CommentsContainer.cs
peppy: the default and only usage in osu! is "online". so the implementation should be as such. class inheritance shouldn't be structured around making tests work – just do this via virtual `Create` or settable properties.

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
peppy: Obviously. Because you were aware of it. But if the order happened to get changed for whatever reason (ie. the API returned it in a different order) it will crash.

EVAST9919: How about making something like `CommentBundleController` which will set all the relations between objects?

peppy: How about binding to `Success` in the `APIRequest` itself?

EVAST9919:
Will be splitted to smaller pr's, but leaving this open: will be used as a reference.
I guess this one should be marked with a `blocked` label then.

peppy:
Is this still a valid PR? How's progress towards closing this?

EVAST9919:
Well, I can use it for reference even if it will be closed, so no reason to leave it open, I guess.

PR #6722, opened 2019-11-05
Implement total comments counter for CommentsContainer

EVAST9919:
Depends on 
- [ ] https://github.com/ppy/osu/pull/6716


bdach:
I'm going to be honest here, I quite dislike how this is structured. I don't like that the counter is not a separate component, and I don't like the flag shenanigans with the counter hiding. It makes it very hard to understand when and *why* will the counter hide and when not.

I'd say this needs rethinking and restructuring. Right now the cognitive load involved with reviewing changes around this area is getting a bit too big for me, and I'm not convinced this should be that hard.

EVAST9919:
So you think it should be always visible and updated only when new `CommentBundle` arrived?

EVAST9919:

PR #6723, opened 2019-11-05
Comments container recolour

EVAST9919:
- [ ] Depends on https://github.com/ppy/osu/pull/6542

Before:
Screenshot_2
After:
Screenshot_1


bdach:
Comparison:

| lazer | web |
| :-: | :-: |
| image | image |

I suppose if we're going for parity with web then the *Show more* button's colour is off.

EVAST9919:

PR #6728, opened 2019-11-06
Fix adjustments not being removed correctly on retry from pause

### LINKED ISSUE #6704: Glitching audio when restarting beatmap through pause overlay
**Describe the bug:** When playing a beatmap, if you pause the game or lose, after clicking on retry or clicking on quit and then playing the same beatmap again, the game starts to glitch a lot. 
PS: The game works as expected with continue button at pause and quick reload (unless on the pause overlay).

I've tried to solve this issue, but no luck in past hours. I believe the problem is within the classes FailOverlay, PauseOverlay, or it's base classes (such as GameplayMenuOverlay). When there is no (FailOverlay/PauseOverlay).Show(), the bug isn't present, which is weird, since it would indicate that the problem is with Hide() messing up something Show() has done, but the continue button works just fine with Show and Hide of overlays, so I really don't know where to look.
Actually, I've started looking at osu!lazer code for the first time today, so maybe I'm totally wrong.

Demonstration Video: https://youtu.be/otgRI_P8bb4

**osu!lazer version:**  2019.1029.0

**Logs:** https://pastebin.com/c7BnafTB (Equivalent to demonstration video)

I'm using majaro linux, with dotnet 3.0.

peppy:
- Closes https://github.com/ppy/osu/issues/6704.

### DIFF
--- a/osu.Game/Screens/Play/GameplayClockContainer.cs
+++ b/osu.Game/Screens/Play/GameplayClockContainer.cs
@@ -162,14 +162,10 @@ namespace osu.Game.Screens.Play
             if (sourceClock != beatmap.Track)
                 return;
 
+            removeSourceClockAdjustments();
+
             sourceClock = new TrackVirtual(beatmap.Track.Length);
             adjustableClock.ChangeSource(sourceClock);
-        }
-
-        public void ResetLocalAdjustments()
-        {
-            // In the case of replays, we may have changed the playback rate.
-            UserPlaybackRate.Value = 1;
         }
 
         protected override void Update()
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -198,6 +194,14 @@ private void updateRate()
         protected override void Dispose(bool isDisposing)
         {
             base.Dispose(isDisposing);
+
+            removeSourceClockAdjustments();
```
smoogipoo: I'm a little weary that this isn't going to happen in time due to async disposal...

peppy: It's still called in `StopUsingBeatmapClock`, which is called in `OnExiting`. I just combined them because it makes more sense.

smoogipoo: Ah I missed that.

smoogipoo (approved):

PR #6743, opened 2019-11-06
Less sickening parallax

peppy:
Parallax should match the mouse movement more closely to reduce sickening feeling. Also reduces parallax in editor to same level as gameplay.

### DIFF
--- a/osu.Game/Graphics/Containers/ParallaxContainer.cs
+++ b/osu.Game/Graphics/Containers/ParallaxContainer.cs
@@ -69,9 +69,11 @@ namespace osu.Game.Graphics.Containers
             {
                 Vector2 offset = (input.CurrentState.Mouse == null ? Vector2.Zero : ToLocalSpace(input.CurrentState.Mouse.Position) - DrawSize / 2) * ParallaxAmount;
 
-                double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, 1000);
+                const float parallax_duration = 100;
 
-                content.Position = Interpolation.ValueAt(elapsed, content.Position, offset, 0, 100, Easing.OutQuint);
+                double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, parallax_duration);
+
+                content.Position = Interpolation.ValueAt(elapsed, content.Position, offset, 0, parallax_duration, Easing.OutQuint);
                 content.Scale = Interpolation.ValueAt(elapsed, content.Scale, new Vector2(1 + System.Math.Abs(ParallaxAmount)), 0, 1000, Easing.OutQuint);
             }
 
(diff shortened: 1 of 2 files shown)


osu.Game/Graphics/Containers/ParallaxContainer.cs
```diff
@@ -71,7 +71,7 @@ protected override void Update()
 
                 double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, 1000);
```
bdach: Tests seem to be failing due to this being unchanged, which sends the interpolated position out of whack:
```suggestion
                 double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, 100);
```
Probably best not to apply this outright, but to extract the value to a constant instead.

bdach (changes requested):
I agree with the change itself, less movement is definitely good for me and I know delays like the old ones can really make people a bit nauseous. There's one thing that needs to be changed though to fix relevant test failures.

bdach (dismissed):

Joehuu (approved):

osu.Game/Graphics/Containers/ParallaxContainer.cs
```diff
@@ -69,9 +69,11 @@ protected override void Update()
...
                 Vector2 offset = (input.CurrentState.Mouse == null ? Vector2.Zero : ToLocalSpace(input.CurrentState.Mouse.Position) - DrawSize / 2) * ParallaxAmount;
 
-                double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, 1000);
+                const float parallax_duration = 100;
 
-                content.Position = Interpolation.ValueAt(elapsed, content.Position, offset, 0, 1000, Easing.OutQuint);
+                double elapsed = MathHelper.Clamp(Clock.ElapsedFrameTime, 0, parallax_duration);
+
+                content.Position = Interpolation.ValueAt(elapsed, content.Position, offset, 0, parallax_duration, Easing.OutQuint);
                 content.Scale = Interpolation.ValueAt(elapsed, content.Scale, new Vector2(1 + System.Math.Abs(ParallaxAmount)), 0, 1000, Easing.OutQuint);
```
bdach: Missed this, but looking at the surrounding code again this has the old value too (therefore will stop a tenth of the way in interpolating - which due to the quint easing function isn't *too* bad, but still slightly off), and so does `load()`:
https://github.com/ppy/osu/blob/d400e4a5f6a576b83a355bd34ef405d774382811/osu.Game/Graphics/Containers/ParallaxContainer.cs#L50
Sorry, should have pointed that out before...

peppy: AH, I actually intentionally left this one, since it only happens on transitions and feels better as a gradual change.

smoogipoo (approved):

PR #6749, opened 2019-11-06
Fix DummyAPIAccess.IsLoggedIn returning incorrect values

frenzibyte:


bdach (approved):
I suppose this is straightforward enough?

PR #6750, opened 2019-11-06
Disallow parsing key "Version" for legacy beatmap skins

### LINKED ISSUE #6433: Beatmap skin configuration decoding potentially conflicts with beatmap configuration

frenzibyte:
Closes #6433

### DIFF
--- a/osu.Game/Skinning/LegacyBeatmapSkin.cs
+++ b/osu.Game/Skinning/LegacyBeatmapSkin.cs
@@ -7,7 +7,7 @@ using osu.Game.Beatmaps;
 
 namespace osu.Game.Skinning
 {
-    public class LegacyBeatmapSkin : LegacySkin
+    public class LegacyBeatmapSkin : LegacySkin<LegacyBeatmapSkinDecoder>
     {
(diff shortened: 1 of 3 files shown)


bdach:
Could you provide an example of a beatmap that hits this problem? Are you sure this ambiguity cannot be resolved by checking the file section?

iiSaLMaN:
> Could you provide an example of a beatmap that hits this problem?

Basically, any beatmap, check the referenced issue for more information.

> Are you sure this ambiguity cannot be resolved by checking the file section?

You mean early return on the `Metadata` section? that's another way to do it. (seems nicer than filtering the key `Version`)

bdach:
The issue does not have the kind of info I'd want either. I'd want to see an actual example file.

I looked at the wiki, and if the information there is correct, skin version should be in `[General]`, and beatmap version should be in `[Metadata]` - does that sound correct? If it does, you need to inspect the section parameter and switch on it appropriately to handle both without having to ignore one.

iiSaLMaN:
https://github.com/ppy/osu/issues/6433#issuecomment-555733478

PR #6752, opened 2019-11-07
Fix possible crash when loading mania compose screen

### LINKED ISSUE #6746: Hard crash when switching from the timing panel to the compose panel while editing an osu!mania beatmap.
Switching from the timing to compose tab **when a timing section is selected** when editing an osu!mania map hard crashes the game. More info in logs.
occurs in version 2019.1106.0
[Logs](https://www1.zippyshare.com/v/Y6CcEhKt/file.html)

i5-7500
GTX1060
8GB DDR4

smoogipoo:
Fixes #6746

Doesn't fix the fact that mania is completely broken right now, it's not a focus.


peppy (approved):
will do for now

PR #6758, opened 2019-11-07
Fix home button not closing login and now playing overlays

Joehuu:


smoogipoo (dismissed):

osu.Game/OsuGame.cs
```diff
@@ -671,6 +671,8 @@ protected override void LoadComplete()
                 };
             }
 
+            overlays.AddRange(new OverlayContainer[] { login, nowPlaying });
```
peppy: this may be changing the behaviour of `CloseAllOverlays` (see the `bool` in the method). needs further investigation.

peppy (changes requested):

osu.Game/OsuGame.cs
Joehuu: We can basically revert https://github.com/ppy/osu/pull/4765 as closing the now playing overlay should be done even when the bool is false.

frenzibyte (dismissed):

peppy:
I've simplified the process to not require manually adding overlays to the list (this will be handled by the load method).

peppy (approved):

PR #6760, opened 2019-11-08
Implement stateful menu items

smoogipoo:
This implements 3 types of stateful menu items:

1. The base type, which can be used for custom states like the following:

https://streamable.com/qzjpk

2. A toggle menu item, where items can be in a disabled or enabled state.

https://streamable.com/zj8xe

2. A "three state" menu item like the toggle menu item but with a third "indeterminate" state. This follows the naming convention of .NET's `CheckBox.ThreeState`.

https://streamable.com/crmda

Furthermore this:
1. Fix the design of context sub-menus - the wrong class was being ctor'd in `CreateSubMenu()`.
2. Moves `DrawableOsuMenuItem` out of `OsuMenu`.
3. Refactors ctors of `OsuMenuItem`.

In the future I'd like to see the state icon be right-aligned, however right now custom positioning and sizing inside a menu item is quite a mess. It's likely that the whole `ContentDrawWidth` stuff of `DrawableMenuItem` needs to be reworked.


osu.Game/Graphics/UserInterface/DrawableStatefulMenuItem.cs
```diff
@@ -0,0 +1,72 @@
...
+                base.LoadComplete();
+                state.BindValueChanged(updateState, true);
+            }
+
+            protected override void Update()
+            {
+                base.Update();
+
+                // Todo: This is bad. This can maybe be done better with a refactor of DrawableOsuMenuItem.
+                stateIcon.X = BoldText.DrawWidth + 10;
```
bdach: I suppose to make this nicer, something like an abstract `CreateIconContainer()` could be done in `DrawableOsuMenuItem` (defaults to `null`, only used/shown when overridden)? I could PR that.

bdach (dismissed):
All looks good and working to me 👍

osu.Game/Graphics/UserInterface/DrawableStatefulMenuItem.cs
smoogipoo: I think the bigger issue to tackle is sizing/positioning. `Menu` invokes `ContentDrawWidth` on the menu items which refers a child of the `DrawableMenuItem`s to retrieve the size. I'd like to see that addressed somehow first before refactoring this further, since the icons should also be right-aligned in the end.
I can imagine something like using a `GridContainer` for the items and allowing parts of the items to be added to different columns.

peppy (dismissed):

smoogipoo:
Looks ok

PR #6762, opened 2019-11-08
Implement hitobject sound adjustment

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/6760

https://streamable.com/4uub0

(ignore shaking objects)

Currently only supported via right-click

Does not support nodular samples (e.g. parts of sliders), which will be added via a better interface. In the case of nodular samples, this doesn't change the existing nodular samples at all, so sliders may have a different head/repeat sound than what was selected, but will have the correct tail sound.

I'm not sure if this is the final form of any of this yet, in terms of where it's located in code, etc:
1. I'm not sure yet how nodular sound editing will work because it may just be a completely separate path so I'm hesitant to move this logic into the selection blueprints themselves.
2. I've extracted logic for addition/removal into individual methods with the expectation that they're going to be called via buttons/hotkeys.
3. I'm not sure on the whole `BindableList` stuff yet. It may well be easier to change samples and call `HitObjectComposer.UpdateHitObject()`, and do a full `ApplyDefaults()`. Doing this is kinda weird with the current structure of sample loading (async + immediate `onApplyDefaults()` invocation in LoadComplete...).  
I can explore this option further if desired.


smoogipoo:
Also ignore menu weirdness, various framework issues. https://github.com/ppy/osu-framework/issues/2965

bdach:
I looked through this, can't find too much to gripe about. The bindable lists don't seem that horrible from the diff alone but if it's possible to centralise application then that might be better and safer.

As a UX thing it could be nice to have some indication that a menu has a submenu under it, but that's out of scope here.

peppy:
One thing I noticed which is not covered by the framework issue:

- Right clicks that aren't on a blueprint shouldn't show a blueprint/selection level menu (and optionally should deselect the current selection). This conforms to Windows/macOS selection behaviours:

screencast 2019-11-12 11-20-44

screencast 2019-11-12 11-19-43

smoogipoo:
I've made the context menu only show when any of the blueprints are hovered. I'm not changing right-click to deselect in this PR since every OS behaves differently.

peppy:
I can't hear hitsounds in the editor with this change.

peppy (approved):
Seems fine for a first pass

Issue #6724, opened 2019-11-05
Audio preview in osu!direct doesn't stop if its "stop" sprite icon is hidden


Gabixel:
**Steps for the bug:**
1. Open osu!direct while there's a beatmap (in the music player) playing
2. Click the preview button of a beatmap in osu!direct
3. Hide its "stop" sprite icon (or the entire panel of that beatmap)
4. **Result:** the sprite icon doesn't update and its preview progress bar doesn't disappear. The beatmap that was playing in the music player remains muted until you stop hiding the sprite icon (or you can just close osu!direct)

**Screenshots or videos showing encountered issue:** https://streamable.com/uc6m3

**osu!lazer version:** 2019.1105.0 (but this seems to also happen on previous versions)

PR #6768, opened 2019-11-08
Fix preview track manager allowing playback of multiple tracks

### LINKED ISSUE #6757: Can preview two songs at the same time in osu!direct
**Describe the bug:** I can get the game to play two songs at the same time in osu!direct.

Steps to reproduce:

1. Preview song 1
2. Preview song 2. This will stop song 1 as expected.
3. Preview song 1 when song 2 is still playing. This doesn't stop song 2.

Screenshot

**osu!lazer version:** 2019.1106.0

frenzibyte:
- `current?.Stop()` inside the **scheduled** track start event **schedules** a null assign to `current` which is to be invoked after the track start event **schedule**. due to that, preview track manager loses it's current track and multiple preview tracks can be played.

Closes #6757


bdach:
The diagnosis of the issue seems to be correct, and the code seems to be working fine, but I'm a little bit worried about the naked stop call there. Unless my brain isn't working today I think there's a slim chance it might end up hitting a race on the null check.

I'd be much more comfortable with this if there was a way to call a `ScheduledDelegate` and continue it with something else (or otherwise ensure order of scheduling, so that the `Stopped` schedule gets to run before the `Started` one does).

It'd be also nice to have a regression test for #6724 so that there is conclusive evidence this does not end up reverting #6738, but I think you told me that isn't that easy.

iiSaLMaN:
> The diagnosis of the issue seems to be correct, and the code seems to be working fine, but I'm a little bit worried about the naked stop call there. Unless my brain isn't working today I think there's a slim chance it might end up hitting a race on the null check.
> 
> I'd be much more comfortable with this if there was a way to call a `ScheduledDelegate` and continue it with something else (or otherwise ensure order of scheduling, so that the `Stopped` schedule gets to run before the `Started` one does).
> 
> It'd be also nice to have a regression test for #6724 so that there is conclusive evidence this does not end up reverting #6738, but I think you told me that isn't that easy.

Thinking about it now, I think I can make this by having a non-present preview track (0 alpha), will give it a try.

iiSaLMaN:
@bdach Resolved.

osu.Game/Audio/PreviewTrackManager.cs
```diff
@@ -48,13 +48,18 @@ public PreviewTrack Get(BeatmapSetInfo beatmapSetInfo)
 
             track.Started += () => Schedule(() =>
             {
-                current?.Stop();
+                var last = current;
                 current = track;
+                last?.Stop();
```
bdach: Needs an inline comment explaining what the local is for.

bdach (commented):
Looking at the code itself (haven't checked out and run it yet) I think it's better now, but I might be wrong. I don't see this causing leaks since the local should go out of scope at the end of the delegate but it might be worth checking empirically.

peppy:
This is a very convoluted way to fix the issue. How about just not nulling `current` if `current` isn't the `track` that was stopped?

iiSaLMaN:
> This is a very convoluted way to fix the issue. How about just not nulling `current` if `current` isn't the `track` that was stopped?

I believe I did exactly like that..? save the last track and stop it after assigning `current` to the new track?

peppy:
The local shouldn't be required then.

peppy (dismissed):

PR #6769, opened 2019-11-08
Fix incorrect DI usage in some tournament screen tests

peppy:


bdach:
Just out of curiosity, why was the old version incorrect? I can see a couple of possibilities (the same bindable instance ended up in places it shouldn't? it was some lifetime woes?), but I'm not exactly sure what the actual problem was.

peppy:
the cached values are not resolved anywhere, as far as i could tell.

bdach:
Ah... right, I even looked at all `Bindable<TournamentMatch>` occurrences and somehow missed that not a single one had the attribute on it, my bad. Not my day I suppose.

bdach (approved):

smoogipoo (approved):

PR #6770, opened 2019-11-08
Add delete option to the right-click menu

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/6762 (context menu)


bdach (approved):
Looks straightforward on its own I suppose 👍

peppy (approved):

PR #6807, opened 2019-11-11
Lookup tournament player statistics for the correct ruleset

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/2976


smoogipoo (dismissed):

smoogipoo:
Requires conflict resolution

osu.Game/Graphics/UserInterfaceV2/LabelledDropdown.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using osu.Framework.Graphics;
+using osu.Game.Graphics.UserInterface;
+
+namespace osu.Game.Graphics.UserInterfaceV2
+{
+    public class LabelledDropdown<T> : LabelledComponent<OsuDropdown<T>, T>
```
smoogipoo: This should be split out into its own PR because it feels pretty bad atm: https://streamable.com/0s7js

peppy: Will move local for now, the fix for this isn't so straightforward.

smoogipoo (dismissed):

smoogipoo (approved):

PR #5634, opened 2019-08-07
Create a control for filtering mods on a leaderboard

EVAST9919:
Split from https://github.com/ppy/osu/pull/6812

PR #5451, opened 2019-07-25
Add mod selector to beatmap overlay

EVAST9919:
Depends on:
- [x] https://github.com/ppy/osu/pull/5627
- [x] https://github.com/ppy/osu/pull/5629
- [ ] https://github.com/ppy/osu/pull/5634

Screenshot_3

PR #6812, opened 2019-11-11
Add mod selector to beatmap overlay

EVAST9919:
Supersedes https://github.com/ppy/osu/pull/5451

Depends on:
- [x] https://github.com/ppy/osu/pull/5634

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -7,6 +7,12 @@
       "commands": [
         "dotnet-cake"
       ]
+    },
+    "dotnet-format": {
+      "version": "3.1.37601",
+      "commands": [
+        "dotnet-format"
+      ]
     }
   }
 }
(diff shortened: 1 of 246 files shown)


bdach:
I think even purely visually this is quite a ways away from proper integration. Problems I noticed on a rudimentary look:

* Touch Device mod has default fallback icon and colours.
* No placeholder when filtering out scores when the filter matches no scores - compare with web:
  image
* When filtering scores the whole score container disappears and scrolls the entire overlay up, which leads to some pretty bad UX. Scores should probably fade partially out and add spinner on top in line with web.

[See video](https://streamable.com/is4hf) for visual demonstration of all three problems.

EVAST9919:
@bdach mostly fixed, but I don't really wanna touch the touchscreen mod since:
- it has no such icon in the framework
- it was moved to `System` mods and mostly invisible in-game on purpose, I suppose. So not sure I can decide here

peppy:
This implementation is missing placeholders when not supporter, as the web has.

image

Going to merge without this, but do consider adding this as a follow-up addition.

peppy (approved):

PR #6823, opened 2019-11-12
Apply new language features

huoyaoyuan:
ppy/osu-framework#2897 and ppy/osu-framework#2948 .


osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -288,17 +288,15 @@ protected override void OnHoverLost(HoverLostEvent e)
...
 
             public override bool Contains(Vector2 screenSpacePos) => content.Contains(screenSpacePos);
 
-            public string TooltipText => name;
+            public string TooltipText { get; }
 
             public ScoreComponentLabel(LeaderboardScoreStatistic statistic)
             {
-                name = statistic.Name;
+                TooltipText = statistic.Name;
```
huoyaoyuan: These classes are very small; the field isn't used elsewhere, and doesn't has a special meaning like `isDisposed`. So I think it's fine to lose the name of the field.

osu.Game/OsuGameBase.cs
```diff
@@ -85,7 +83,7 @@ public class OsuGameBase : Framework.Game, ICanAcceptFiles
         [Cached(Type = typeof(IBindable<IReadOnlyList<Mod>>))]
         protected readonly Bindable<IReadOnlyList<Mod>> Mods = new Bindable<IReadOnlyList<Mod>>(Array.Empty<Mod>());
 
-        protected Bindable<WorkingBeatmap> Beatmap => beatmap;
+        protected Bindable<WorkingBeatmap> Beatmap { get; private set; } // cached via load() method
```
huoyaoyuan: Not sure about this. Since there's no attribute, I just copied the comment.

osu.Game/Online/Leaderboards/LeaderboardScore.cs
bdach: I sort of disagree with these sorts of renames, would rather rename properties back to `Name` and add delegating explicit `IHasTooltip` implementations.

osu.Game/Overlays/KeyBinding/VariantBindingsSubsection.cs
```diff
@@ -7,8 +7,7 @@ namespace osu.Game.Overlays.KeyBinding
 {
     public class VariantBindingsSubsection : KeyBindingsSubsection
     {
-        protected override string Header => variantName;
-        private readonly string variantName;
+        protected override string Header { get; }
```
bdach: Same thing here, would rather have `VariantName` back and `Header` delegating to it myself.

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
```diff
@@ -379,26 +379,25 @@ protected virtual DrawableSlider CreateDrawableSlider(Slider slider) => new Draw
 
         private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
         {
-            var osuObject = judgedObject as DrawableOsuHitObject;
-            if (osuObject == null)
-                return;
-
-            OsuSpriteText text;
-            Add(text = new OsuSpriteText
+            if (judgedObject is DrawableOsuHitObject osuObject)
```
bdach: Personally I'd rather negate, use as guard and then keep using `osuObject` exclusively.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/HitObjectPatternGenerator.cs
```diff
@@ -303,8 +303,7 @@ private Pattern generateRandomPatternWithMirrored(double centreProbability, doub
 
             var pattern = new Pattern();
 
-            bool addToCentre;
-            int noteCount = getRandomNoteCountMirrored(centreProbability, p2, p3, out addToCentre);
+            int noteCount = getRandomNoteCountMirrored(centreProbability, p2, p3, out var addToCentre);
```
bdach: Maybe `out bool` would be better here, since this is the only instance wherein the type is not immediately obvious from the function call/variable name?

bdach (commented):
Some comments. Mostly stylistic nitpicks, probably need backup of everyone else whether they agree with me or not.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/HitObjectPatternGenerator.cs
huoyaoyuan: This could be handled by "don't use var for predefined type". I just applied the auto fix here.

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
huoyaoyuan: For consistence, fast-return seems not aggressively used in the code base.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/HitObjectPatternGenerator.cs
bdach: I think the guideline for `var` was that it should be used only if the type can be inferred from the context? Which is unfortunately kind of hard to enforce with an analyzer, true. I'd rather not enable "don't use var for predefined type", since the same commit here shows examples where that doesn't make a lot of sense (those `int.Parse` calls have no reason to then have `out int`)

huoyaoyuan: The code base doesn't have wide consistency about `var` now. I'd prefer it discussed separately.

osu.Game/Overlays/KeyBinding/VariantBindingsSubsection.cs
smoogipoo: I prefer this personally. Bit of a much of a muchness.

osu.Game/OsuGameBase.cs
smoogipoo: I don't really mind it.

smoogipoo (commented):
I'm fine with this in general.

huoyaoyuan:
Index expression crashes compiler in VS 16.3, and successes in 16.4.
Well let's separate it.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/HitObjectPatternGenerator.cs
```diff
@@ -168,43 +168,43 @@ private Pattern generate()
...
 
-                    return pattern = generateRandomPattern(1, 0.62, 0, 0);
+                    return generateRandomPattern(1, 0.62, 0, 0);
                 }
 
                 if (ConversionDifficulty > 4)
                 {
                     if (convertType.HasFlag(PatternType.LowProbability))
-                        return pattern = generateRandomPattern(0.35, 0.08, 0, 0);
+                        return generateRandomPattern(0.35, 0.08, 0, 0);
```
smoogipoo: All of this changes the code that will be run in `finally`

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
smoogipoo: I have pushed for early-return multiple times in the project. I highly prefer that pattern for large code blocks from a reading/maintenance perspective - preconditions can generally be discarded when reading the following code while otherwise you have to keep them in mind.

With that in mind, I don't particularly disagree with it in the cases in this PR.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
peppy: I'm fine with early return in most cases, but still find this one a particularly hard read when negated, since the line is creating a variable for future but not indenting code (like `using` would do). I find it easy to miss that the variable is being created in the first place.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/HitObjectPatternGenerator.cs
huoyaoyuan: OK this should be another false positive about `finally`.

huoyaoyuan: I'm doing a minor refactor on this. Using `try` as control flow is strongly discouraged and somehow misleading. Since the method has only one caller, I've made it a local function and move post-process in `finally` to the caller.

smoogipoo (approved):

smoogipoo:
Test regressed from a previously merged PR, disregard.

PR #6824, opened 2019-11-12
Handle SharpRaven deprecation

huoyaoyuan:
SharpRaven is [deprecated](https://www.nuget.org/packages/SharpRaven/2.4.0) by its owner. Switching to supported new SDK.

Note: the new SDK is using much more static members. The instance members actually used are internal. Since we are subscribing to static event, it feels OK to me.

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -71,7 +71,7 @@ namespace osu.Game
         [Cached]
         private readonly ScreenshotManager screenshotManager = new ScreenshotManager();
 
-        protected SentryLogger RavenLogger;
+        protected SentryLogger SentryLogger;
 
         public virtual Storage GetStorageForStableInstall() => null;
 
@@ -110,7 +110,7 @@ namespace osu.Game
 
             forwardLoggedErrorsToNotifications();
 
-            RavenLogger = new SentryLogger(this);
+            SentryLogger = new SentryLogger(this);
         }
 
         private void updateBlockingOverlayFade() =>
@@ -166,7 +166,7 @@ namespace osu.Game
 
             dependencies.CacheAs(this);
 
-            dependencies.Cache(RavenLogger);
+            dependencies.Cache(SentryLogger);
 
             dependencies.Cache(osuLogo = new OsuLogo { Alpha = 0 });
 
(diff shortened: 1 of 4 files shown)


osu.Game/OsuGame.cs
```diff
@@ -71,7 +71,7 @@ public class OsuGame : OsuGameBase, IKeyBindingHandler<GlobalAction>
         [Cached]
         private readonly ScreenshotManager screenshotManager = new ScreenshotManager();
 
-        protected RavenLogger RavenLogger;
+        protected SentryLogger RavenLogger;
```
bdach: Member can be renamed too.

osu.Game/Utils/SentryLogger.cs
```diff
@@ -81,19 +81,9 @@ private bool shouldSubmitException(Exception exception)
...
-            {
-                lock (tasks)
-                    tasks.Remove(task);
-            });
-        }
-
         #region Disposal
 
-        ~RavenLogger()
+        ~SentryLogger()
```
bdach: Does this finalizer even need to exist?

osu.Game/Utils/SentryLogger.cs
```diff
@@ -112,7 +102,8 @@ protected virtual void Dispose(bool isDisposing)
                 return;
 
             isDisposed = true;
-            lock (tasks) Task.WaitAll(tasks.ToArray(), 5000);
+            sentry.Dispose();
```
bdach: I'd probably guard this for null despite `isDisposed` anyways, if this is going to be called in the finalizer - no guarantees of anything there. Better safe than sorry with these things.

osu.Game/Utils/SentryLogger.cs
```diff
@@ -7,26 +7,26 @@
...
 namespace osu.Game.Utils
 {
     /// <summary>
     /// Report errors to sentry.
     /// </summary>
-    public class RavenLogger : IDisposable
+    public class SentryLogger : IDisposable
     {
-        private readonly RavenClient raven = new RavenClient("https://5e342cd55f294edebdc9ad604d28bbd3@sentry.io/1255255");
+        private IDisposable sentry;
```
bdach: Weird API choice *to have `IDisposable`* as the returned type, but alas...

huoyaoyuan: There is a good sense `SentryClient`, but missing `Breadcrumb` support.
An alternative is use a permanent scope.

peppy (approved):
Looks sane

PR #6825, opened 2019-11-12
Fix cancel button animating when entering a dialog

### LINKED ISSUE #6822: Dialog cancel button animates while others do not
Cancel button looks like it was hovered for a second
nice

GSculerlor:
Fix #6822 

https://streamable.com/vfx2w


bdach:
[See comment](https://github.com/ppy/osu/pull/6826#issuecomment-552920729) for a special case concerning the confirm dialog overlay we might also want to handle separately.

peppy (approved):
works as expected

PR #6826, opened 2019-11-12
Remove default-to-cancel assumption on exiting popup dialogs

### LINKED ISSUE #6822: Dialog cancel button animates while others do not

mcendu:
Fix #6822 Aside from some minor refactoring in the same PR, I think that #6825 might be a better solution if the default-to-cancel assumption is correct.

### DIFF
--- a/osu.Game/Graphics/UserInterface/DialogButton.cs
+++ b/osu.Game/Graphics/UserInterface/DialogButton.cs
@@ -20,6 +20,7 @@ namespace osu.Game.Graphics.UserInterface
 {
     public class DialogButton : OsuClickableContainer
     {
+        private const float idle_width = 0.8f;
         private const float hover_width = 0.9f;
         private const float hover_duration = 500;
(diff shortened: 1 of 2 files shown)


bdach:
I think we'd want to keep the effect of pressing the cancel button on hide for UX purposes.

mcendu:
> I think we'd want to keep the effect of pressing the cancel button on hide when no action was triggered for UX purposes.

This assumption is not correct in some cases. I'll take the exit confirm dialog as an example here: When the player clicks the close window button the second time, the player is doing the equivalent of pressing "Good bye", instead of the cancel button.

bdach:
I see - in that particular case could we possibly add a special path that would trigger the confirm button then? I'd say that'd be the best possible fix of all available.

mcendu:
Please close this issue if #6825 is merged. The minor code refactor here is no longer necessary as the same change is featured in #6839

PR #6833, opened 2019-11-13
Fix carousel filtering twice on startup due to unpopulated ruleset

### LINKED ISSUE #3058: Going into song select lets maps for another mode disappear
Not sure if that is intended but that beeing said it causes stutters when going into songselect

https://streamable.com/53ics

peppy:
Fixes #3058.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -46,7 +46,7 @@ namespace osu.Game.Screens.Select
         protected const float BACKGROUND_BLUR = 20;
         private const float left_area_padding = 20;
 
-        public FilterControl FilterControl;
+        public FilterControl FilterControl { get; private set; }
 
         protected virtual bool ShowFooter => true;
 
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -260,7 +260,11 @@ namespace osu.Game.Screens.Select
             }
         }
 
-        protected virtual void ApplyFilterToCarousel(FilterCriteria criteria) => Carousel.Filter(criteria);
+        protected virtual void ApplyFilterToCarousel(FilterCriteria criteria)
+        {
+            if (this.IsCurrentScreen())
+                Carousel.Filter(criteria);
+        }
 
         private DependencyContainer dependencies;
 
@@ -432,6 +436,8 @@ namespace osu.Game.Screens.Select
         public override void OnEntering(IScreen last)
         {
             base.OnEntering(last);
+
+            Carousel.Filter(FilterControl.CreateCriteria(), false);
 
             this.FadeInFromZero(250);
             FilterControl.Activate();
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -88,8 +88,6 @@ namespace osu.Game.Screens.Select
         [Resolved(canBeNull: true)]
         private MusicController music { get; set; }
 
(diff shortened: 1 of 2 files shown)


bdach (dismissed):
I suppose I'll trust the test on this one 👍

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -90,8 +90,14 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
         [Cached(Type = typeof(IBindable<IReadOnlyList<Mod>>))]
         private readonly Bindable<IReadOnlyList<Mod>> mods = new Bindable<IReadOnlyList<Mod>>(Array.Empty<Mod>()); // Bound to the game's mods, but is not reset on exiting
 
-        protected SongSelect()
+        protected virtual void ApplyFilterToCarousel(FilterCriteria criteria) => Carousel.Filter(criteria);
+
+        [BackgroundDependencyLoader(true)]
+        private void load(BeatmapManager beatmaps, AudioManager audio, DialogOverlay dialog, OsuColour colours, SkinManager skins, ScoreManager scores)
         {
+            // initial value transfer is requried for FilterControl (it uses our re-cached bindables in its asynd loac for the initial filter).
```
bdach: Sorry for doing this, but
```suggestion
            // initial value transfer is required for FilterControl (it uses our re-cached bindables in its async load for the initial filter).
```

bdach (dismissed):

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -46,40 +46,40 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
...
-        public readonly FilterControl FilterControl;
+        public FilterControl FilterControl;
 
         protected virtual bool ShowFooter => true;
 
         /// <summary>
         /// Can be null if <see cref="ShowFooter"/> is false.
         /// </summary>
-        protected readonly BeatmapOptionsOverlay BeatmapOptions;
+        protected BeatmapOptionsOverlay BeatmapOptions;
```
smoogipoo: These need private setters.

smoogipoo (changes requested):

peppy:
Should be good 👍

smoogipoo:
Seems to cause some weird filtering: https://streamable.com/ed1pw

Steps to repro:
1. Enter song select with taiko ruleset.
2. Exit song select.
3. Change to osu! ruleset.
4. Enter song select again.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -46,48 +46,56 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
...
-        protected readonly BeatmapDetailArea BeatmapDetails;
+        protected BeatmapDetailArea BeatmapDetails { get; private set; }
 
         private readonly Bindable<RulesetInfo> decoupledRuleset = new Bindable<RulesetInfo>();
 
         [Resolved(canBeNull: true)]
         private MusicController music { get; set; }
 
-        protected SongSelect()
+        protected virtual void ApplyFilterToCarousel(FilterCriteria criteria) => Carousel.Filter(criteria);
```
smoogipoo: Move before `load()`

smoogipoo (changes requested):

peppy:
Hmm, I'm not sure how we should handle that case you mentioned, if at all. The changes I've made do not re-filter in the background - they just guarantee a second filter won't happen if nothing changed externally.

In the case you've video'd there *is* an external change. It kind of makes sense it's doing a second filter on entry in this case, although optimally it may be better if we can delay the screen push until it is in a good state, not sure.

smoogipoo (approved):
Let’s go with this for now

PR #6834, opened 2019-11-13
Clean up carousel tests

peppy:
I thought I was going to use this but I didn't end up doing so. Still a worthwhile change, I believe (causes the carousel to be recreated per test).


bdach (dismissed):

smoogipoo (dismissed):

smoogipoo:
The tests still fail in the visual browser, indicating some inter-dependence.

smoogipoo (approved):

PR #6836, opened 2019-11-13
Fix background loading twice when entering song select

peppy:


bdach (approved):
Verified that this works when re-entering song select by breakpointing at `BackgroundScreenBeatmap.load()`

PR #6837, opened 2019-11-13
Move to .NET Standard 2.1

huoyaoyuan:
Doing at game side first to unblock framework side.
Reference: https://dotnet.microsoft.com/platform/dotnet-standard , for supported frameworks

#### Advantages
- Allow default interface methods
  Minor, but I do find some places suitable for this.
- Allow consuming EF Core 3
  But they choose to move back to .NET Standard 2.0 in EF Core 3.1, which is coming in early December.
- Allow optimizing streams using span
- Allow index and range expressions

#### Works to do
- Android compatibility
  - Require installing Android 10 SDK, which is currently only available from Google source. Seems installed in AppVeyor.
  - Minimal supported version in manifest kept. Not sure for it's compatibility, because UWP's adoption of netstandard required minversion change.
- iOS compatibility
  - I can't verify anything at all because I don't have Apple device.
  - No clear `TargetFrameworkVersion` property, seems bound to installed VS/VSMac version. Since no property is actually changed, I assume compatibility is kept same.
- Notify external consumers
  - Seems all listed by GitHub `Used by`?
  - Should also include external consumers of framework.
- A step leaving further from .NET Framework
  - .NET Framework has got feature freeze. This will require external consumers to not targeting it.

### DIFF
--- a/osu.iOS.props
+++ b/osu.iOS.props
@@ -75,4 +75,17 @@
     <PackageReference Include="ppy.osu.Game.Resources" Version="2019.1010.0" />
     <PackageReference Include="ppy.osu.Framework.iOS" Version="2019.1112.0" />
   </ItemGroup>
+  <!-- Xamarin.iOS does not automatically handle transitive dependencies from NuGet packages. -->
+  <ItemGroup Label="Transitive Dependencies">
+    <PackageReference Include="Humanizer" Version="2.7.9" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.2.6" />
+    <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="2.2.6" />
+    <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
+    <PackageReference Include="ppy.osu.Framework" Version="2019.1112.0" />
+    <PackageReference Include="SharpCompress" Version="0.24.0" />
+    <PackageReference Include="NUnit" Version="3.12.0" />
+    <PackageReference Include="SharpRaven" Version="2.4.0" />
+    <PackageReference Include="System.ComponentModel.Annotations" Version="4.6.0" />
+    <PackageReference Include="ppy.osu.Framework.NativeLibs" Version="2019.1112.0" ExcludeAssets="all" />
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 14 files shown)


smoogipoo:
I think the most we'll need to do is add an item to the `Breaking Changes ` section in the wiki, listing the move to 2.1 and the requirement of netcore 3.0 alongside it.

huoyaoyuan:
Marking as ready.
After a glance I haven't seen any external dependencies on .NET Framework.
The only work remaining should be mobile compatibility check. @swoolcock should be the one interesting on them?

swoolcock:
Have you tried building Xamarin iOS and Android?

huoyaoyuan:
iOS builds as before, but It's not providing usable result without a Mac.
Android claims missing SDK(jar), instead of cannot target. This matches the behavior when I've not installed v8.0. I cannot install from google source easily, microsoft source hasn't added 10.0.

swoolcock:
Xamarin.iOS (specifically `mtouch`) does not correctly handle transitive NuGet package dependencies, so unfortunately the "redundant package references" must remain.  The framework-side projects (Tests/SampleGame) do not have the same problem as they use project references.

Also important is the `ExcludeAssets="all"` on NativeLibs, otherwise mtouch will attempt to AOT the ffmpeg dll files, assuming they are .NET assemblies.

I can confirm that when replacing the package references, osu! compiles and runs on iOS.

huoyaoyuan:
Confirmed osu.Android successfully runs on my device at Android 9.0.

smoogipoo (approved):
Depends on @peppy updating his toolsets to make sure it compiles (Xamarin.iOS/Android specifically).

osu.iOS.props
```diff
@@ -75,4 +75,17 @@
     <PackageReference Include="ppy.osu.Game.Resources" Version="2019.1010.0" />
     <PackageReference Include="ppy.osu.Framework.iOS" Version="2019.1112.0" />
   </ItemGroup>
+  <!-- Xamarin.iOS does not automatically handle transitive dependencies from NuGet packages. -->
+  <ItemGroup Label="Transitive Dependencies">
```
peppy: how has it been working until now?

swoolcock: @huoyaoyuan removed these explicit dependencies from osu.iOS.props assuming they were unrequired.  My commit readded them.

peppy: they aren't in master – does that mean master is broken?

huoyaoyuan: Yes. It's not effectively verified in CI.

PR #6838, opened 2019-11-13
Fix osu!direct previews muting game audio permanently

### LINKED ISSUE #6835: Music permanent muted if song previewed in direct and imported while sitting in song select
**Describe the bug:** 
Go into song select and let the music play. Open Direct and preview a song. while its previewing, the current song gets muted. Download the song while its previewing and let it import. Result is that the music is perma muted until the game is restarted.

https://streamable.com/cd5bc

**osu!lazer version:** 
2019.1113.0

**Logs:** 
[runtime.log](https://github.com/ppy/osu/files/3840571/runtime.log)

frenzibyte:
- [x] Depends on ppy/osu-framework#3000

Closes #6835

### DIFF
--- a/osu.Game.Tests/Visual/Components/TestScenePreviewTrackManager.cs
+++ b/osu.Game.Tests/Visual/Components/TestScenePreviewTrackManager.cs
@@ -13,7 +13,7 @@ namespace osu.Game.Tests.Visual.Components
 {
     public class TestScenePreviewTrackManager : OsuTestScene, IPreviewTrackOwner
     {
-        private readonly PreviewTrackManager trackManager = new TestPreviewTrackManager();
+        private readonly TestPreviewTrackManager trackManager = new TestPreviewTrackManager();
 
         protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
         {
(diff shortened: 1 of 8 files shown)


bdach:
A bug-reproducing test case would be nice to have.

iiSaLMaN:
Seems that this alone doesn't fix the issue correctly, will keep investigating further.

peppy:
Explanation as to why? Test?

iiSaLMaN:
> Explanation as to why? Test?

Nulling the current right after stopping track will make the stopped event check (`current != track`) becomes true and early-return without unmuting the game.

peppy:
Needs an inline comment saying as such (right where the `= null` line used to be) so we don't accidentally regress this in the future.

bdach:
Why isn't this testable? Is this a race condition or something equally non-deterministic?

peppy:
It is testable, and there are already good tests setup for this component.

osu.Game.Tests/Visual/Components/TestScenePreviewTrackManager.cs
```diff
@@ -108,6 +108,26 @@ public void TestNonPresentTrack()
...
+
+            AddStep("get track", () => Add(owner = new TestTrackOwner(track = getTrack())));
+            AddUntilStep("wait loaded", () => track.IsLoaded);
+            AddStep("start track", () => track.Start());
+            AddAssert("current is track", () => trackManager.CurrentTrack == track);
+            AddStep("pause manager updates", () => trackManager.AllowUpdate = false);
+            AddStep("stop any playing", () => trackManager.StopAnyPlaying(owner));
+            AddAssert("current not changed", () => trackManager.CurrentTrack == track);
+            AddStep("resume manager updates", () => trackManager.AllowUpdate = true);
+            AddAssert("current is null", () => trackManager.CurrentTrack == null);
```
peppy: how is this testing that mute was undone? and why is the "allowupdate" required?

iiSaLMaN: well, this test is ensuring that the current track changes correctly as we're assuming `Stop` will remove the audio bindable, `AllowUpdate` is required to check if the `CurrentTrack` didn't change before the scheduled stopped event is invoked.

peppy: why not test for it? that's generally what we do.

bdach: In general `current == null` is the less important half of the desired behaviour we want here. With regression tests you generally want to test the actually important user-facing part of the regression, which is the volume (from what I should see it should be easily exposed). `current` being `null` could be well considered an implementation detail.

swoolcock:
@iiSaLMaN how's this going?  https://github.com/ppy/osu-framework/pull/3000 was a breaking change for osu!, so this PR needs to go in.

iiSaLMaN:
> how's this going? [ppy/osu-framework#3000](https://github.com/ppy/osu-framework/pull/3000) was a breaking change for osu!, so this PR needs to go in.

Everything seems to be alright, I thought this PR is waiting for a framework bump, no?

swoolcock:
Looks like we can't do a framework bump until `DrawableRuleset.FallbackSampleStore` is updated to implement `IAggregateAudioAdjustment`.  Could you please do that?

peppy (dismissed):

peppy (approved):

PR #6839, opened 2019-11-13
Correct DialogButton click animation

mcendu:
This PR is made to fix a possible bug in the click animation of `DialogButton`. In `PopupDialog` instances, after either of _OK_ or _Cancel_ is clicked, the dialog would (sometimes) also close. The close process seems to cause a `HoverLost` event to be fired alongside the expected `Click` event, causing the possibly incorrect behavior.
[Here is a demonstration of the original behavior.](https://github.com/ppy/osu/files/3840830/glitch.mkv.gz) Note how the cancel button behave when it is clicked.


peppy:
Still reproduces in tests with your changes.

mcendu:
> Still reproduces in tests with your changes.

How is the bug reproduced?

peppy:
Run `TestSceneDialogButton`.

peppy:
Please check and fix CI inspections.

peppy:
Are you actually testing this? It seems to be getting more broken.

https://www.youtube.com/watch?v=s8Sx9GK6ROY

EVAST9919:
Fixed via https://github.com/ppy/osu/pull/6825

peppy:
Seems that way.

mcendu:
This PR is intended to fix another bug, instead of #6822.

peppy:
Feel free to reopen but please provide a hood explanation of what it’s fixing

PR #6841, opened 2019-11-13
Always trust official site in external link opener

### LINKED ISSUE #5926: Remove unnecessary safeguard against official osu! domain
**Describe the new feature:** 

I don't think it's necessary to check whether the user really wants to visit the official osu! website.

osu_2019-08-31_23-57-27

huoyaoyuan:
Fixes #5926 .
It's OK to hardcode official site because it's also hardcoded in API access.

### DIFF
--- a/osu.Game/Online/Chat/ExternalLinkOpener.cs
+++ b/osu.Game/Online/Chat/ExternalLinkOpener.cs
@@ -27,10 +27,16 @@ namespace osu.Game.Online.Chat
 
         public void OpenUrlExternally(string url)
         {
-            if (externalLinkWarning.Value)
+            if (externalLinkWarning.Value && !isDomainTrusted(url))
                 dialogOverlay.Push(new ExternalLinkDialog(url, () => host.OpenUrlExternally(url)));
             else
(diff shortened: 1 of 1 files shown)


iiSaLMaN:
I believe this was agreed to be in the form of a domain whitelist, not only osu! domain.

bdach:
I also don't think this check is lenient enough. It's case-sensitive, so for example in chat `https://osu.ppy.sh/` won't require confirmation, while `https://osu.ppy.SH/` will. It should probably be case-insensitive.

Perhaps a little debatable if the `https` protocol specification should be used, but I'm in favour of its current inclusion, since naked HTTP shouldn't really be a thing anymore.

peppy:
I think this needs to be coupled with an option to always ask. Half of the reason of the confirmation popup is to let people know they are going to exit the game experience, not that the site is "untrusted".

This is especially important when running exclusive fullscreen, or on a mobile device (especially android which will quit the game right now).

peppy:
Going to close this for now due to reason state above. Can revisit in the future.

PR #6843, opened 2019-11-13
Remove usages of untyped equality

huoyaoyuan:
Part of ppy/osu-framework#2901 .


bdach (dismissed):
Build failure is relevant due to unprotected access to possibly-null beatmap in `SongSelect`.

peppy:
The location of the failure also changes behaviour, as it no longer runs the containing code on `null == null`, which was intended.

peppy (approved):

smoogipoo (approved):

PR #4613, opened 2019-04-09
Remove mods from WorkingBeatmap

smoogipoo:
Resolves https://github.com/ppy/osu/issues/4610

1. `OsuScreen` now has a `SelectedMods` bindable, similar to beatmap and ruleset.
2. Mods now have to be provided to `WorkingBeatmap.GetPlayableBeatmap()` and `DrawableRuleset()`.
3. Track adjustments are applied at music controlled right now, because that's the component which handles music playing right now. I think eventually we may want to move it into `SongSelect`, but I kept the changes simple for now.
4. Due to (3), `MatchSubScreen` now plays the modded track, which may be expected functionality?

PR #6846, opened 2019-11-14
Fix selected mods not being retained on exiting song select

### LINKED ISSUE #3186: Mod selection resets when exiting song select
It regressed. Doesn't apply selected mods when entering back anymore.

peppy:
Regressed due to incorrect implementation transfer in https://github.com/ppy/osu/pull/4613 (the previous logic which was removing the mod from the `WorkingBeatmap` was incorrectly transferred to remove it from the global mod bindable).

- Closes #3186 for the third time, now with regression testing.


smoogipoo:
This should still reset the mods on exiting, DT shouldn't remain applied in the menu, for example.

smoogipoo (changes requested):
As mentioned

smoogipoo (approved):
Seems ok

PR #6851, opened 2019-11-14
Use audio backend to create preview tracks

swoolcock:
Related to https://github.com/ppy/osu-framework/issues/2448
- [ ] Depends on https://github.com/ppy/osu-framework/pull/2999
- [ ] Depends on framework bump


swoolcock:
Closing, see https://github.com/ppy/osu-framework/pull/2999

PR #6852, opened 2019-11-14
Fix database unique constraint issues on some imports

### LINKED ISSUE #6572: Unique constraint errors when importing from stable
**Describe the bug:** 
Got a lot of DB errors while importing maps from stable.
"osu.game.database.osudbcontext"
"dbcommand"

**osu!lazer version:** 
osu 2019.1021.0.0

**Logs:** 
[database.log](https://github.com/ppy/osu/files/3755778/database.log)

peppy:
- Fixes #6572.

Two things were incorrect here:

- The beatmap online retrieval response running incorrectly on the API scheduler, causing incorrect import behaviour (in many cases the online IDs would not be imported correctly as they weren't populated in time).

- Online retrieved IDs could be double if two .osu files with the same content (and hash) existed. I'm making as second PR to dedupe these before we hit online retrieval, but I think the check is still valuable to have (see inline comment).

Requires login / api to test, can easily test using: https://puu.sh/EEpm3/bf5f955d21.osz

Note that as this is a thread race condition the failure may take multiple attempts, but usually happens on first.

### DIFF
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -129,9 +129,12 @@ namespace osu.Game.Beatmaps
         {
             var beatmapIds = beatmapSet.Beatmaps.Where(b => b.OnlineBeatmapID.HasValue).Select(b => b.OnlineBeatmapID).ToList();
 
+            LogForModel(beatmapSet, "Validating online IDs...");
+
             // ensure all IDs are unique
             if (beatmapIds.GroupBy(b => b).Any(g => g.Count() > 1))
             {
+                LogForModel(beatmapSet, "Found non-unique IDs, resetting...");
                 resetIds();
                 return;
             }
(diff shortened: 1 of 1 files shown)


smoogipoo:
Test?

bdach:
Might be difficult to add a reproducibly failing test for the first part, since it's a race? The hash thing has a test in the other PR, could maybe add coverage for the check included here though.

smoogipoo:
Yeah I'm interested in the test for duplicate beatmap IDs

peppy:
See description, I mention that a test is unfeasible due to online requirement.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -380,22 +387,25 @@ private void update(BeatmapSetInfo set, BeatmapInfo beatmap)
...
                     beatmap.Status = res.Status;
                     beatmap.BeatmapSet.Status = res.BeatmapSet.Status;
                     beatmap.BeatmapSet.OnlineBeatmapSetID = res.OnlineBeatmapSetID;
-                    beatmap.OnlineBeatmapID = res.OnlineBeatmapID;
-                };
 
-                req.Failure += e => { LogForModel(set, $"Online retrieval failed for {beatmap} ({e.Message})"); };
+                    // note that this check only needs to be here if two identical hashed beatmaps exist int he same import.
+                    // probably fine to leave it for safety.
+                    if (set.Beatmaps.All(b => b.OnlineBeatmapID != res.OnlineBeatmapID))
```
smoogipoo: Just documenting although we've discussed this IRL.

This is going to fail if a bad beatmap in the set already contains the online beatmap ID of another beatmap in the set.

smoogipoo (approved):

Issue #6854, opened 2019-11-14
Searching for multiplayer rooms does not work


recapitalverb:
**Describe the bug:** Searching for multiplayer rooms using the `type to search` bar does not work.
How to reproduce:
1. Navigate to multi lounge: (osu!) -> play -> multi
2. Type anything into search bar and hit <kbd>Enter</kbd>
3. Observe.
The results shown below should update, but instead nothing happens and still shows the default list of rooms that show up shortly after entering multi lounge.

**Screenshots or videos showing encountered issue:**
Before: ![Screenshot 1](https://i.imgur.com/i80nMue.png)
After: ![Screenshot 2](https://i.imgur.com/sergrlc.png)

**osu!lazer version:** 2019.1113.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/3846880/database.log)
[network.log](https://github.com/ppy/osu/files/3846887/network.log)
[performance.log](https://github.com/ppy/osu/files/3846882/performance.log)
[runtime.log](https://github.com/ppy/osu/files/3846883/runtime.log)

PR #6857, opened 2019-11-14
Hook up multiplayer search filter

smoogipoo:
Resolves https://github.com/ppy/osu/issues/6854


peppy:
Doesn't correctly apply if you change display tabs (open / recently ended / owned). Also might be best to add a short debounce (same as song select should work fine).

peppy:
I'm not sure I completely agree with the debounce application here. I'd prefer it was done at `FilterControl` only for the text value changing. Right now it's adding an extra delay when clicking the tabs (which i think we want to avoid).

And also a big ugly having it in two places (leading on from discussions today, I think reducing debounce stuff is in our best interest).

recapitalverb:
What are we currently waiting for before merging this?

peppy (approved):

PR #6860, opened 2019-11-15
Implement slider path segments

smoogipoo:
image

Currently breaks the editor, which will now only show the first segment's control points, but that will be added eventually.

I'm running this on ~400k beatmaps just to make sure there aren't any regressions (I've added one test relating to perfect curves). Out of the first 4000 there aren't any differences compared to master.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
@@ -272,7 +272,7 @@ namespace osu.Game.Rulesets.Osu.Tests
                 Position = new Vector2(-(distance / 2), 0),
                 Path = new SliderPath(new[]
                 {
-                    new PathSegment(PathType.PerfectCurve, new[]
+                    new PathSegment(PathType.Linear, new[]
                     {
                         Vector2.Zero,
                         new Vector2(distance, 0),
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
@@ -187,6 +187,45 @@ namespace osu.Game.Rulesets.Osu.Tests
             bool assertSamples(HitObject hitObject) => hitObject.Samples.All(s => s.Name != HitSampleInfo.HIT_CLAP && s.Name != HitSampleInfo.HIT_WHISTLE);
         }
 
+        [Test]
+        public void TestSliderWithMultipleSegments()
+        {
+            AddStep("create slider", () =>
+            {
+                var bezierSegment = new List<Vector2> { Vector2.Zero };
+
+                for (int i = 1; i < 50; i++)
+                {
+                    bezierSegment.Add(new Vector2(i * 10, -300));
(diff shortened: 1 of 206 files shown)


osu.Game/Rulesets/Objects/SliderPath.cs
```diff
@@ -35,35 +27,24 @@ public struct SliderPath : IEquatable<SliderPath>
...
-        /// <param name="controlPoints">The control points of the path.</param>
-        /// <param name="expectedDistance">A user-set distance of the path that may be shorter or longer than the true distance between all
-        /// <paramref name="controlPoints"/>. The path will be shortened/lengthened to match this length.
-        /// If null, the path will use the true distance between all <paramref name="controlPoints"/>.</param>
+        /// <param name="segments">The path segments.</param>
+        /// <param name="expectedDistance">A user-set distance of the path that may be shorter or longer than the true distance. The path will be shortened/lengthened to match this length.
+        /// If null, the path will use the true distance.</param>
         [JsonConstructor]
-        public SliderPath(PathType type, Vector2[] controlPoints, double? expectedDistance = null)
+        public SliderPath(PathSegment[] segments, double? expectedDistance = null)
```
peppy: can we add a fallback constructor for single segment paths rather than complicate all the existing usages with single item array logic?

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
```diff
@@ -187,6 +187,45 @@ public void TestChangeSamplesWithNodeSamples()
...
+                    bezierSegment.Add(new Vector2(i * 10, 300));
+                }
+
+                var slider = create(new Vector2(-200),
+                    new PathSegment(PathType.Bezier, new[]
+                    {
+                        Vector2.Zero,
+                        new Vector2(80, -125),
+                        new Vector2(150, 105),
+                        new Vector2(260, 0),
```
peppy: shouldn't there be tests and exception logic if the previous segment's last vector doesn't equal the first of the next segment?

peppy (changes requested):

osu.Game.Rulesets.Osu.Tests/TestSceneSlider.cs
smoogipoo: Will add a test but an exception isn't necessary since it works just fine (just interpolates linearly between the segments).

Original:
Screen Shot 2019-11-19 at 12 56 49 pm

Separated:
Screen Shot 2019-11-19 at 12 55 32 pm

peppy: hmm i guess that's okay, seems a bit weird though. i guess it's hard to avoid duplication of the end / start vectors of each segment to ensure this can't happen in the first place?

just thinking forward mappers may use these kinds of edge cases to create broken game mechanics.

smoogipoo: Given that sliders are computed as linear interpolations between pieces, it's more expected functionality rather than an edge case. Wouldn't be hard to add a check to avoid duplication but I see it as unnecessary.

smoogipoo:
Closing as this is being reimplemented.

PR #6861, opened 2019-11-15
Add black layer for storyboard

EVAST9919:
For some storyboards there's a moment when we can see beatmap background (which in stable cannot be visible)

[Good example](https://osu.ppy.sh/beatmapsets/914120#osu/1922594)
**Stable**
Screenshot_1
**Lazer (master)**
Screenshot_2
**Lazer (with pr)**
Screenshot_3


EVAST9919:
Oops, now that layer can be visible on beatmaps with no storyboard

PR #6867, opened 2019-11-16
Implement TotalCommentsCounter component for CommentsContainer

EVAST9919:


osu.Game/Overlays/Comments/TotalCommentsCounter.cs
```diff
@@ -0,0 +1,80 @@
...
+        {
+            RelativeSizeAxes = Axes.X;
+            Height = 50;
+            AddInternal(new FillFlowContainer
+            {
+                Anchor = Anchor.CentreLeft,
+                Origin = Anchor.CentreLeft,
+                AutoSizeAxes = Axes.Both,
+                Direction = FillDirection.Horizontal,
+                Margin = new MarginPadding { Left = 50 },
```
peppy: This shouldn't be defined in here. It should be handled wherever this display is used. There's no way it's guaranteed to be the same for each use case.

peppy (approved):

osu.Game/Overlays/Comments/TotalCommentsCounter.cs
peppy: On second thoughts let's just leave it until the full picture is realised. May be better solved by a rename of this class.

PR #6872, opened 2019-11-17
Fix bad references in xml doc

huoyaoyuan:
Same with ppy/osu-framework#3006 .

### DIFF
--- a/osu.Game/Beatmaps/BeatmapDifficulty.cs
+++ b/osu.Game/Beatmaps/BeatmapDifficulty.cs
@@ -56,10 +56,22 @@ namespace osu.Game.Beatmaps
         /// Maps a difficulty value [0, 10] to a two-piece linear range of values.
         /// </summary>
         /// <param name="difficulty">The difficulty value to be mapped.</param>
-        /// <param name="range">The values that define the two linear ranges.</param>
-        /// <param name="range.od0">Minimum of the resulting range which will be achieved by a difficulty value of 0.</param>
-        /// <param name="range.od5">Midpoint of the resulting range which will be achieved by a difficulty value of 5.</param>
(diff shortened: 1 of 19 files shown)


osu.Game/Beatmaps/BeatmapDifficulty.cs
```diff
@@ -56,10 +56,22 @@ public static double DifficultyRange(double difficulty, double min, double mid,
...
+        ///     <description>Midpoint of the resulting range which will be achieved by a difficulty value of 5.</description>
+        ///   </item>
+        ///   <item>
+        ///     <term>od10</term>
+        ///     <description>Maximum of the resulting range which will be achieved by a difficulty value of 10.</description>
+        ///   </item>
+        /// </list>
+        /// </param>
         /// <returns>Value to which the difficulty value maps in the specified range.</returns>
         public static double DifficultyRange(double difficulty, (double od0, double od5, double od10) range)
```
huoyaoyuan: This overload is better to be removed. The only call [here](https://github.com/ppy/osu/tree/master/osu.Game/Rulesets/Scoring/HitWindows.cs#L90) passes `min, mid, max`, not `od0, od5, od10` to it.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -351,7 +351,7 @@ private void select(CarouselItem item)
         /// <summary>
         /// Half the height of the visible content.
         /// <remarks>
-        /// This is different from the height of <see cref="ScrollContainer{T}.displayableContent"/>, since
+        /// This is different from the height of <see cref="ScrollContainer{T}"/>.displayableContent, since
```
huoyaoyuan: Referring to private member will surprise user more when sources of framework and game are separated.

smoogipoo (approved):

PR #6878, opened 2019-11-19
Remove usages of FileSafety class

huoyaoyuan:
As https://github.com/ppy/osu-framework/pull/3004#discussion_r347746661 , the entire class is going to be removed. Since game side is using totally different methods with framework side, just copying implementations of the 2 used methods.

- [x] Depends on ppy/osu-framework#3004 after change.

### DIFF
--- a/osu.Game/Beatmaps/WorkingBeatmap.cs
+++ b/osu.Game/Beatmaps/WorkingBeatmap.cs
@@ -82,7 +82,6 @@ namespace osu.Game.Beatmaps
         /// <returns>The absolute path of the output file.</returns>
         public string Save()
         {
-            // copied from osu.Framework.IO.File.FileSafety.GetTempPath
             string directory = Path.Combine(Path.GetTempPath(), @"osu!");
             Directory.CreateDirectory(directory);
 
--- a/osu.Game/Beatmaps/WorkingBeatmap.cs
+++ b/osu.Game/Beatmaps/WorkingBeatmap.cs
@@ -7,7 +7,6 @@ using osu.Game.Rulesets.Mods;
 using System;
 using System.Collections.Generic;
 using osu.Game.Storyboards;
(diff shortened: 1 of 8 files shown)


osu.Game/IO/FileStore.cs
```diff
@@ -130,5 +130,12 @@ public override void Cleanup()
                 }
             }
         }
+
+        /// <summary>
+        /// Moved from osu.Framework.IO.FileSafety.
+        /// </summary>
+        /// <param name="path">The path to standardise.</param>
+        /// <returns>The standardised path using '/' for directory separator.</returns>
+        internal static string PathStandardise(string path) => path.Replace('\\', '/');
```
huoyaoyuan: A temporary position not unsuitable for the method to live. May be factored out further.

peppy: Disagree with this placement. How about just making it an extension method and putting in `ExtensionMethods`?

huoyaoyuan: Should be sure, but then this PR must be merged together with updating framework.

osu.Game/Beatmaps/WorkingBeatmap.cs
```diff
@@ -83,7 +82,11 @@ protected virtual Track GetVirtualTrack()
         /// <returns>The absolute path of the output file.</returns>
         public string Save()
         {
-            var path = FileSafety.GetTempPath(Guid.NewGuid().ToString().Replace("-", string.Empty) + ".json");
+            // copied from osu.Framework.IO.File.FileSafety.GetTempPath
```
smoogipoo: unnecessary

smoogipoo (changes requested):

peppy (dismissed):

smoogipoo (approved):

PR #6886, opened 2019-11-20
Fix crash when loading results after gameplay

### LINKED ISSUE #6884: Results screen crashing when loading scores
Sentry Issue: [OSU-LAZER-38N](https://sentry.io/organizations/ppy/issues/1326709716/?referrer=github_integration)

```
InvalidOperationException: Sequence contains no matching element
  Module "osu.Game.Scoring.LegacyDatabasedScore", line 44, in .ctor
    Void .ctor(osu.Game.Scoring.ScoreInfo, osu.Game.Rulesets.RulesetStore, osu.Game.Beatmaps.BeatmapManager, osu.Framework.IO.Stores.IResourceStore`1[System.Byte[]])
  Module "osu.Game.Scoring.ScoreManager", line 34, in GetScore
    osu.Game.Scoring.Score GetScore(osu.Game.Scoring.ScoreInfo)
  Module "osu.Game.OsuGame", line 319, in PresentScore
    Void PresentScore(osu.Game.Scoring.ScoreInfo)
  Module "osu.Game.Screens.Ranking.Pages.ReplayDownloadButton+<>c__DisplayClass5_0", line 40, in load { <lambda> }
    Void <load>b__0()
  Module "osu.Framework.Graphics.Containers.ClickableContainer", line 30, in OnClick
    Boolean OnClick(osu.Framework.Input.Events.ClickEvent)
...
(17 additional frame(s) were not displayed)

An unhandled error has occurred.
```

smoogipoo:
Fixes https://github.com/ppy/osu/issues/6884

Online IDs can incorrectly match in the case of offline plays, so Beatmap ID is used.  
Beatmap IDs can incorrectly match in the case of offline plays on the same Beatmap, so hash is used.
Scores imported from stable will have a non-null hash, scores imported by lazer have a null hash as they don't currently have files. This'll change in the future. The only case to consider is `Hash (null) == Hash (null)` for scores imported from lazer, in which case there is a secondary check in `CheckLocalAvailability` for having files.

Online id and beatmap id can match in the case of offline plays. Hash is the fallback which essentially ensures that a s

### DIFF
--- a/osu.Game/Scoring/ScoreInfo.cs
+++ b/osu.Game/Scoring/ScoreInfo.cs
@@ -183,6 +183,9 @@ namespace osu.Game.Scoring
 
         public override string ToString() => $"{User} playing {Beatmap}";
 
-        public bool Equals(ScoreInfo other) => other?.OnlineScoreID == OnlineScoreID;
+        public bool Equals(ScoreInfo other) =>
+            other?.OnlineScoreID == OnlineScoreID
+            && other?.BeatmapInfoID == BeatmapInfoID
+            && other.Hash == Hash;
     }
 }
--- a/osu.Game/Scoring/ScoreInfo.cs
+++ b/osu.Game/Scoring/ScoreInfo.cs
@@ -184,8 +184,9 @@ namespace osu.Game.Scoring
         public override string ToString() => $"{User} playing {Beatmap}";
 
         public bool Equals(ScoreInfo other) =>
-            other?.OnlineScoreID == OnlineScoreID
-            && other?.BeatmapInfoID == BeatmapInfoID
+            other != null
+            && other.OnlineScoreID == OnlineScoreID
+            && other.BeatmapInfoID == BeatmapInfoID
             && other.Hash == Hash;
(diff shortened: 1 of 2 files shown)


osu.Game/Scoring/ScoreInfo.cs
```diff
@@ -183,6 +183,9 @@ protected class DeserializedMod : IMod
 
         public override string ToString() => $"{User} playing {Beatmap}";
 
-        public bool Equals(ScoreInfo other) => other?.OnlineScoreID == OnlineScoreID;
+        public bool Equals(ScoreInfo other) =>
+            other?.OnlineScoreID == OnlineScoreID
```
peppy: just for readability and safety, i'd prefer doing a null check as the first condition here. right now the first condition (`other?.OnlineScoreID == OnlineScoreID`) can pass even if other is null (as `OnlineScoreID` is nullable)

peppy (changes requested):

peppy (approved):

PR #6890, opened 2019-11-20
Fix song select not always scrolling to the correct location

peppy:
Fixed issue mentioned [here](https://github.com/ppy/osu/pull/6833#issuecomment-554322387).


smoogipoo (approved):
Seems ok

PR #6892, opened 2019-11-20
Change SO pp multiplier to be based on amount of spinners in the map

stanriders:
Currently SO cuts too much pp for what's essentialy a tiny part of map difficulty. This should fix it (and make deranking using SO almost impossible)


smoogipoo:
I'm fine with this, any thoughts @peppy @Tom94 ?

peppy:
It's a *huge* reduction in the negative multiplier. Is there any discussion around this? I'd like to see how this conclusion was come to, and get more player feedback.

stanriders:
Disscussion didn't lead anywhere besides "nobody except new players and derankers uses SO lol". 

The idea is to make SO punish as much as it impacts difficulty, which in result makes easy maps (assuming easy maps are all low on object count) with SO lose a noticable amount of pp forcing new players to learn to spin, and hard maps (again, assuming hard maps have more objects) almost not losing anything to make SO pretty much impossible to derank with. 
This obviously won't solve deranking (people will just play offline, as they mostly do already), but it'll make SO pp reduction more reasonable
This also fixes pp reduction when there're no spinners in a map

smoogipoo:
Please open a related PR in https://github.com/ppy/osu-performance . The code changes required should be very similar, but that will lead to more discussions (and hopefully others testing).

peppy:
Thanks for opening on the remote repo. We'll close this one for now pending discussion/resolution, at which point it can be reconsidered.

PR #6893, opened 2019-11-20
Add some extra search keywords to settings

### LINKED ISSUE #6888: Descriptive settings search
**Describe the new feature:** 

Searching in settings is inconvenient because it only filters using the description text. Having a set of keywords associated with an option would be good.

For example, when searching for UI scaling, "scal" or "scaling" is accepted but "scale" is not.

image
image

Other similar cases:
- Frame limiter settings with "fps"
- Audio device selector with "speaker"
- Volume sliders with "sound"
- Mouse sensitivity slider with "speed"

EVAST9919:
Fixes https://github.com/ppy/osu/issues/6888

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
@@ -22,6 +22,8 @@ namespace osu.Game.Overlays.Settings.Sections.Audio
                 new SettingsSlider<double> { LabelText = "Effect", Bindable = audio.VolumeSample, KeyboardStep = 0.01f },
                 new SettingsSlider<double> { LabelText = "Music", Bindable = audio.VolumeTrack, KeyboardStep = 0.01f },
             };
+
+            Keywords = new[] { "Sound" };
         }
     }
 }
--- a/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Collections.Generic;
-using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Audio;
 using osu.Framework.Graphics;
@@ -13,8 +11,6 @@ namespace osu.Game.Overlays.Settings.Sections.Audio
     public class VolumeSettings : SettingsSubsection
     {
         protected override string Header => "Volume";
-
-        public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "Sound" });
 
         [BackgroundDependencyLoader]
         private void load(AudioManager audio, OsuConfigManager config)
--- a/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Audio;
 using osu.Framework.Graphics;
@@ -12,6 +14,8 @@ namespace osu.Game.Overlays.Settings.Sections.Audio
     {
         protected override string Header => "Volume";
 
+        public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "Sound" });
+
         [BackgroundDependencyLoader]
         private void load(AudioManager audio, OsuConfigManager config)
         {
@@ -22,8 +26,6 @@ namespace osu.Game.Overlays.Settings.Sections.Audio
                 new SettingsSlider<double> { LabelText = "Effect", Bindable = audio.VolumeSample, KeyboardStep = 0.01f },
                 new SettingsSlider<double> { LabelText = "Music", Bindable = audio.VolumeTrack, KeyboardStep = 0.01f },
             };
-
-            Keywords = new[] { "Sound" };
         }
     }
 }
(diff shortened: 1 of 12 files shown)


osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -38,6 +38,7 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Show health display even when you can't fail",
                     Bindable = config.GetBindable<bool>(OsuSetting.ShowHealthDisplayWhenCantFail),
+                    Keywords = new[] { "bar" }
```
phosphene47: I think this should be capitalized as "Bar". It doesn't matter, but just for consistency.

EVAST9919:
I think `Keywords` should be moved to `IHasFilterTerms` interface, but not sure

osu.Game/Overlays/Settings/SettingsSubsection.cs
```diff
@@ -21,7 +21,10 @@ public abstract class SettingsSubsection : FillFlowContainer, IHasFilterableChil
         protected abstract string Header { get; }
 
         public IEnumerable<IFilterable> FilterableChildren => Children.OfType<IFilterable>();
-        public IEnumerable<string> FilterTerms => new[] { Header };
+
+        public virtual IEnumerable<string> FilterTerms => Keywords == null ? new[] { Header } : new List<string>(Keywords) { Header }.ToArray();
+
+        public IEnumerable<string> Keywords { get; set; }
```
jai-x: Why add a new `Keywords` property and not just add the relevant search terms to the `FilterTerms` property directly? This is already being done similarly in `SettingsDropdown`.

EVAST9919: Sounds fair, but with this implementation we need to create a new class for every object for which we want to add keywords

EVAST9919: However it can be simplified for subsections since they are already splitted classes

osu.Game/Overlays/Settings/Sections/Graphics/RendererSettings.cs
```diff
@@ -12,6 +14,8 @@ public class RendererSettings : SettingsSubsection
     {
         protected override string Header => "Renderer";
 
+        public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "FPS" });
```
peppy: not sure i understand

osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
```diff
@@ -27,12 +27,14 @@ private void load(OsuConfigManager osuConfig, FrameworkConfigManager config)
                 new SettingsCheckbox
                 {
                     LabelText = "Raw input",
-                    Bindable = rawInputToggle
+                    Bindable = rawInputToggle,
+                    Keywords = new[] { "Speed" }
```
peppy: this also seems wrong

osu.Game/Overlays/Settings/Sections/Gameplay/ModsSettings.cs
```diff
@@ -18,7 +18,8 @@ private void load(OsuConfigManager config)
                 new SettingsCheckbox
                 {
                     LabelText = "Increase visibility of first object when visual impairment mods are enabled",
-                    Bindable = config.GetBindable<bool>(OsuSetting.IncreaseFirstObjectVisibility)
+                    Bindable = config.GetBindable<bool>(OsuSetting.IncreaseFirstObjectVisibility),
+                    Keywords = new[] { "Hidden", "Traceable" }
```
peppy: no. "mod" maybe but not this.

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -38,6 +38,7 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Show health display even when you can't fail",
                     Bindable = config.GetBindable<bool>(OsuSetting.ShowHealthDisplayWhenCantFail),
+                    Keywords = new[] { "Bar" }
```
peppy: add "hp" too?

osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
```diff
@@ -12,6 +14,8 @@ public class VolumeSettings : SettingsSubsection
     {
         protected override string Header => "Volume";
 
+        public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "Sound" });
```
peppy: shouldn't this be at the audio *section* level?

osu.Game/Overlays/Settings/Sections/Graphics/RendererSettings.cs
EVAST9919: This one has been taken from issue suggestion. I'll remove this

osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
EVAST9919: as above

osu.Game/Overlays/Settings/Sections/Graphics/RendererSettings.cs
peppy: if it's placed on the FPS display (and wasn't already a keyword) i'd be okay with it.

peppy (dismissed):
guess this is ok for now.

osu.Game/Overlays/Settings/Sections/Gameplay/SongSelectSettings.cs
```diff
@@ -26,18 +26,21 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Show converted beatmaps",
                     Bindable = config.GetBindable<bool>(OsuSetting.ShowConvertedBeatmaps),
+                    Keywords = new[] { "convert" }
```
EVAST9919: Not necessary, since it's already in the label text. (That's why I used "converts", not "convert")

peppy: Adding plurals feels completely backwards. I'd sooner advocate implementing a similar mechanism to stable where the search will stop filtering if it can see zero results would be returned.

PR #6911, opened 2019-11-22
Fix usage of SpriteText instead of OsuSpriteText

peppy:
WIP. PRing with analysis only to test appveyor's response.


smoogipoo:
blocking until the erroneous compiler warnings are fixed

peppy:
Should be good.

smoogipoo (approved):

Issue #3433, opened 2018-09-15
Unclear Mods Overlay can be scrolled?


plankp:
It took me a while to realize the newly introduced mods are only accessible after you scroll up the mods overlay.

Is there anyway to make it more obvious that the mods overlay can be scrolled up-down? (like having a scrollbar on the side or something to show there is still content below)

jorolf:
The scrollbar has been hidden intentionally.

Gabixel:
Maybe just adding "(scroll down for more mods)" in the Mods Description can be a good solution?

peppy:
The scrollability needs to be exposed either via UI elements or animation on appearing.

Joehuu:
Closing in favor of https://github.com/ppy/osu/issues/16917 (already resolved in new design).

PR #6916, opened 2019-11-22
Increase padding on mod section title

aspriddell:
Minor change, increase the margin/padding of the mod section's title to look better

Before:
image

After:
image


Kashouryo:
I think this change can also tell the player that the mod list is scrollable.

bdach:
If you end up trying to add a scroll indicator see #3433 for the discussion of what could be done to explicitly show scrollability.

aspriddell:
I think if the padding was increased again slightly (maybe from 20 to 25) it would make it more obvious it is a scrollable element. The fact it cuts off the autoplay text should be enough of a hint that the container has more content underneath

peppy (approved):

PR #6917, opened 2019-11-22
Add placeholder for non-supporters in beatmap overlay

EVAST9919:
Screenshot_1

### DIFF
--- a/osu.Game/Overlays/BeatmapSet/Scores/NotSupporterPlaceholder.cs
+++ b/osu.Game/Overlays/BeatmapSet/Scores/NotSupporterPlaceholder.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Graphics.Sprites;
+using osuTK;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Containers;
+
+namespace osu.Game.Overlays.BeatmapSet.Scores
+{
+    public class NotSupporterPlaceholder : Container
+    {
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/BeatmapSet/Scores/NotSupporterPlaceholder.cs
```diff
@@ -0,0 +1,49 @@
...
+                        Anchor = Anchor.TopCentre,
+                        Origin = Anchor.TopCentre,
+                        Direction = FillDirection.Horizontal,
+                        AutoSizeAxes = Axes.Both,
+                    }
+                }
+            };
+
+            text.AddText("Click ");
+            text.AddLink("here", "/home/support");
```
bdach: Why's this not a full URL like here?
https://github.com/ppy/osu/blob/fd3e2375bb166624783f8f3676e70493868ab565/osu.Game/Screens/Menu/Disclaimer.cs#L123

osu.Game/Overlays/BeatmapSet/Scores/ScoresContainer.cs
```diff
@@ -223,6 +235,18 @@ private void getScores()
...
+            {
+                Scores = null;
+                notSupporterPlaceholder.Show();
+                loading.Hide();
+                return;
+            }
+            else
+            {
+                notSupporterPlaceholder.Hide();
+            }
```
bdach: If you're going to `return` in the first branch then there's no need to have an `else` at all.

osu.Game/Overlays/BeatmapSet/Scores/NotSupporterPlaceholder.cs
EVAST9919: Inspired by
https://github.com/ppy/osu/blob/8da65fc9094282a458662d32c4aeb55deccfa721/osu.Game/Overlays/AccountCreation/ScreenWarning.cs#L27

smoogipoo (approved):

PR #6922, opened 2019-11-23
Display useful statistics on song bar depending on current ruleset

peppy:


osu.Game.Tournament/Components/SongBar.cs
```diff
@@ -174,12 +180,43 @@ private void update()
 
             if ((mods & LegacyMods.DoubleTime) > 0)
             {
-                //ar *= 1.5f;
+                ar *= 1.5f;
```
smoogipoo: This isn't accurate. AR9 + DT is equivalent of AR10.33. It should be calculated based on hit window timing (see https://github.com/ppy/osu/blob/4063135a3a894e262f880cca2b91904c51fb33c2/osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs#L56 )

smoogipoo (changes requested):

osu.Game.Tournament/Components/SongBar.cs
peppy: yeah, that was understood when this was originally added. if you can make the fix to make it accurate that would definitely be appreciated.

smoogipoo (approved):

PR #6933, opened 2019-11-23
Fix incorrect NativeLibs version reference

peppy:
I thought we already fixed this one, but it's still wrong.


huoyaoyuan:
@swoolcock added it for mtouch not correctly handing transitive reference, but assumed that nativelibs is shipped in the same pace of managed.

smoogipoo (approved):

PR #6940, opened 2019-11-23
Add "Cinema" mod

### LINKED ISSUE #6029: Cinema mod is unaccessable
**Describe the bug:** You can't access Cinema mod by double clicking Auto mod.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=_CZ8KV77CHc

**osu!lazer version:** 2019.906.0

**Logs:** -

aspriddell:
Resolves #6029.

Add cinema mod support for all modes. In short it hides the `HUDOverlay` and `drawableRuleset.Playfield` but sets `AlwaysPresent` to `true` on each (otherwise you cannot quit the player and there will be no hitsounds)

All ruleset-specific classes have also been added to `ConvertLegacyMods` in the same way that Sudden Death and Perfect are listed (ie check for `Perfect` flag first then `else if` against the `SuddenDeath` flag)

- [x] Depends on https://github.com/ppy/osu/pull/7168


GSculerlor:
Probably we can set the background dim to 0 since it kinda useless if we set dim to 100 and using Cinema mod for example (I believe stable also set the dim to 0 on Cinema mod if I'm not mistaken)

bdach:
I'd say that's a prerequisite, given that the entire purpose of the mod is to watch the storyboard.

aspriddell:
I'll take a look sometime tomorrow. Need to do a bit more digging...
So far I can use the configmanager to set the dim but it doesn't reset after the map has finished

bdach:
I'm pretty sure `ConfigManager` should *not* be used here, since that might irrevocably lose user settings. Something more local would be a lot better - perhaps we'd want to add a way for mods to disable/override display settings.

aspriddell:
I agree @bdach it seems like there is no way to revert the changes with what is currently available. 

I will look into adding an interface to deal with this (potentially called `ITempConfigOverride` or something respective)

osu.Game/Rulesets/Mods/ModCinema.cs
```diff
@@ -3,15 +3,35 @@
...
 
 namespace osu.Game.Rulesets.Mods
 {
-    public class ModCinema : ModAutoplay
+    public abstract class ModCinema<T> : ModCinema, IApplicableToDrawableRuleset<T>
+        where T : HitObject
+    {
+        public virtual void ApplyToDrawableRuleset(DrawableRuleset<T> drawableRuleset)
+        {
+            drawableRuleset.SetReplayScore(CreateReplayScore(drawableRuleset.Beatmap));
```
iiSaLMaN: What about using `IApplicableFailOverride.AllowFail = false` instead?

aspriddell: The inheritance chain should mean it already takes the settings from the Autoplay mod. `ModCinema` should inherit `IApplicableFailOverride` settings [as seen here](https://github.com/ppy/osu/blob/8da65fc9094282a458662d32c4aeb55deccfa721/osu.Game/Rulesets/Mods/ModAutoplay.cs#L21)

osu.Game/Screens/Play/Player.cs
```diff
@@ -498,8 +498,24 @@ public override void OnEntering(IScreen last)
                 .Delay(250)
                 .FadeIn(250);
 
-            Background.EnableUserDim.Value = true;
-            Background.BlurAmount.Value = 0;
+            var screenOverride = Mods.Value.OfType<IApplicableToScreen>();
```
aspriddell: Was thinking that if more than one mod in the list had the `IApplicableToScreen` modifier it threw an exception. In this case line 503 would become a switch with `0`, `1` and `default`, with default throwing the exception.

Any ideas on an appropriate exception to throw, or should I make my own, like `IncompatableModSelection`?

osu.Game/Rulesets/Mods/IApplicableToScreen.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Rulesets.Mods
+{
+    /// <summary>
+    /// An interface for a mod which can temporarily override screen settings.
+    /// </summary>
+    public interface IApplicableToScreen : IApplicableMod
```
peppy: this interface is too specific. you should be providing access to the required instances, not exposing the three things specifically required by cinema mod alone.

osu.Game/Screens/Play/Player.cs
peppy: there's already a system in place for incompatible mods (see `IncompatibleMods`). this kind of logic should not exist.

peppy (changes requested):

osu.Game/Rulesets/Mods/IApplicableToPlayer.cs
```diff
@@ -0,0 +1,15 @@
...
+using osu.Game.Screens.Play;
+
+namespace osu.Game.Rulesets.Mods
+{
+    /// <summary>
+    /// An interface for a mod which can temporarily override the <see cref="Player"/> settings.
+    /// </summary>
+    public interface IApplicableToPlayer : IApplicableMod
+    {
+        void ApplyToPlayer(Player player);
```
aspriddell: Much less specific now, but can still do everything from before

peppy:
If we are using the `IgnoreUserSettings` bindable, it needs test coverage in `TestSceneUserDimContainer`.

osu.Game/Graphics/Containers/UserDimContainer.cs
```diff
@@ -90,7 +91,7 @@ protected virtual void UpdateVisuals()
             ContentDisplayed = ShowDimContent;
 
             dimContent.FadeTo(ContentDisplayed ? 1 : 0, BACKGROUND_FADE_DURATION, Easing.OutQuint);
-            dimContent.FadeColour(OsuColour.Gray(1 - (float)DimLevel), BACKGROUND_FADE_DURATION, Easing.OutQuint);
+            dimContent.FadeColour(IgnoreUserSettings.Value ? OsuColour.Gray(1 - (float)DimLevel) : Color4.White, BACKGROUND_FADE_DURATION, Easing.OutQuint);
```
peppy: Is this line required? Shouldn't it be handled by the `DimLevel` one above?

peppy (changes requested):

osu.Game/Graphics/Containers/UserDimContainer.cs
aspriddell: resolved in caa9286

osu.Game/Rulesets/Mods/ModCinema.cs
```diff
@@ -3,15 +3,43 @@
...
+
+            drawableRuleset.Playfield.AlwaysPresent = true;
```
peppy: Can you explain the reason behind this being required? I'm going through and trying to remove this requirement at the `Player` level.

peppy: Required for hitsounds.

peppy:
I've made a few changes:

- Also hides break overlay
- Removes AlwaysPresent requirement on HUD overlay via https://github.com/ppy/osu/pull/7168
- Adds a comment about playfield `AlwaysPresent` requirement

aspriddell:
👍

peppy:
This is the absolute last change to be made to `UserDimContainer`. That class needs a rewrite to remove the 10+ settings it has. I'm OK with this happening so we at least have documentation of all the different parameters we need to account for when refactoring the class.

peppy (approved):

PR #6941, opened 2019-11-24
IApplicableToDrawableHitObjects should be able to get all the hitobject in nasted Playfield

andy840119:
`IApplicableToDrawableHitObjects` should be able to get all the `HitObjects` in main `Playfield` and it's `NastedPlayfield`


iiSaLMaN:
What is this fixing?

andy840119:
Some of the `HitObject` like `ManiaHitObject` is placed in `NastedPlayfield`.
And it cannot be get.

smoogipoo (approved):

PR #6958, opened 2019-11-25
Fix appveyor integration by not using cake for build

### LINKED ISSUE #6913: Fix AppVeyor integration
Appveyor only collects warnings automatically when either
- specify `project: xxx.sln` in yml
- load their logger, which is only available for `MSBuild.exe` that runs at .NET Framework (the one shipped with VS)

Here're ways I've found to fix it:
#### Use some cake addin to report issues to AppVeyor.
I do find certain addin to do that. But, cake support for dotnet core is not so good. I haven't found necessary apis to setup the logger bridge in minutes, but nearly. If you prefer this I can still investigate it.

#### Don't use cake for build. Only use it for InspectCode.
Personally, I'm a non-fan of custom build scripts.

Unlike .NET Framework Era, in .NET Core we have `dotnet xxx` to drive everything. There's no need for custom scripts to simulate IDE commands in command line.
Let's have a peek about what's contained in the build processes:
- Build desktop components. It's perfectly covered by every IDE and `dotnet build`.
- Run unit tests. I tried the simple command `dotnet test Desktop.proj` and it works very well. We also carry settings for every IDE.
- Build/Test Xamarin components. Current cake script is even skipping them because of invoking `dotnet build` only! I've tested they are happy with VS MSBuild.
- Run tests with HttpBin. It's usually only running in CI, and can be handled perfectly by before_test/after_test.
- Publish to nuget. AppVeyor has a straight built-in support for it.
- Deploy Xamarin apps. Seems not executed by CI at all? Or at some other mac pool not integrated. 
- Run InspectCode. This is the only task that cake **doesn't make regressions**.

#### Don't use cake totally.
In fact InspectCode can be installed and invoked using nuget without cake. Due to the Windows-only fact, a ps1 should be enough.
The only difference with the old CI-only inspection is the way to install it. Not everyone has chocolatey, but everyone is definitely to have nuget.

huoyaoyuan:
Fixes #6913 .
The key improvements: **compiler and nuget warnings are populated, Xamarin projects are covered**

I'm keeping in mind that making build configuration straightforward.
AppVeyor directly integrates to MSBuild. The configuration is quite simple.
Note that **much more warnings are populated**, including ones during restore and packing. It's very boring and hard to collection all them in cake.

InspectCode integrates very badly with the .NET Core world. It's strongly depending "desktop" MSBuild which is only accessible from Visual Studio Developer CLI, and supporting only Windows. **Integrating it into the experience of VS will probably break command line on .NET Core.** So I'm keeping the current progress with cake unchanged.

I also removed the "1 script to do what CI does at local", because:
- Many developers doesn't have Xamarin installed. Then they will require manually skipping them in the script.
- It's only possible to do exactly everything on Windows now.
- It doesn't fit the workflow, especially with IDE.
- User wants to fix violations instead of just prompting them, for example the dry run of `dotnet format`.

I've tried a lot at my local fork to make sure everything is working. Deployment is proved by the correct artifacts collected by AppVeyor. (https://ci.appveyor.com/project/huoyaoyuan/osu-deploy/builds/29081533/artifacts)


huoyaoyuan:
Validated at my fork. The validation of deployment is a bit out of date. It requires to deploy to a fake environment.

huoyaoyuan:
Deployment proved:
Build at https://ci.appveyor.com/project/huoyaoyuan/osu-deploy/builds/29099897
Package pushed to https://www.myget.org/feed/huoyaoyuan/package/nuget/ppy.osu.Game/1.2.3

huoyaoyuan:
Changing the tests directory organization should be suitable for another PR.

huoyaoyuan:
No: currently android test will fail on loading beatmap resource. Since no one is actually using it, let's delay it.

osu.Game/osu.Game.csproj
```diff
@@ -7,7 +7,9 @@
   <PropertyGroup Label="Nuget">
     <Title>osu!</Title>
     <PackageId>ppy.osu.Game</PackageId>
+    <Version>0.0.0</Version>
     <PackageIcon>icon.png</PackageIcon>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
peppy: Not sure we want this, as it will add overhead for normal development.

README.md
```diff
@@ -73,13 +74,19 @@ If you are not interested in debugging osu!, you can add `-c Release` to gain pe
...
+
 ### Testing with resource/framework modifications
 
 Sometimes it may be necessary to cross-test changes in [osu-resources](https://github.com/ppy/osu-resources) or [osu-framework](https://github.com/ppy/osu-framework). This can be achieved by running some commands as documented on the [osu-resources](https://github.com/ppy/osu-resources/wiki/Testing-local-resources-checkout-with-other-projects) and [osu-framework](https://github.com/ppy/osu-framework/wiki/Testing-local-framework-checkout-with-other-projects) wiki pages.
 
 ### Code analysis
 
-Code analysis can be run with `powershell ./build.ps1` or `build.sh`. This is currently only supported under Windows due to [ReSharper CLI shortcomings](https://youtrack.jetbrains.com/issue/RSRP-410004). Alternatively, you can install ReSharper or use Rider to get inline support in your IDE of choice.
+Before committing your code, please run code formatter. It can be achieved with command `dotnet format`, or `Format code` command in your IDE. (Defaults to `Ctrl+K,Ctrl+D` in Visual Studio, `Shift+Alt+F` in Visual Studio Code, or `Ctrl+Alt+L` in Rider)
```
peppy:
```suggestion
Before committing your code, please run a code formatter. It can be achieved with command `dotnet format`, or `Format code` command in your IDE.```

I don't think we need to specify hotkeys (and the default changes in some IDEs depending on which profile you set the IDE up with).

File: README.md [hunk 3]
```diff
@@ -73,13 +74,19 @@ If you are not interested in debugging osu!, you can add `-c Release` to gain pe
 
 If the build fails, try to restore NuGet packages with `dotnet restore`.
 
+_Due to historical feature gap between .NET Core and Xamarin, running `dotnet` CLI from the root directory will probably not work. You can try to run the CLI towards single .NET Core project, or helper project at `build/Desktop.proj`. Or you can get help from your IDE of choice to run the correct commands. We are keeping configuration files in the repository for the popular IDEs._
```
peppy:
```suggestion
_Due to historical feature gap between .NET Core and Xamarin, running `dotnet` CLI from the root directory will not work for most commands. This can be resolved by specifying a target `.csproj` or the helper project at `build/Desktop.proj`. Configurations have been provided to work around this issue for all supported IDEs mentioned above._
```

peppy (changes requested):

osu.Game/osu.Game.csproj
huoyaoyuan: This can help finding packing warnings at normal development, for example, depending on preview package from stable.
Packing is relatively faster than building. `dotnet pack --no-build` takes ~0.5s on my machine.

huoyaoyuan:
AppVeyor does't allow mixture of propertyset and command line. Enabling WarningAsError without incluencing normal build requires explicit specifying everything.
Alternatively, I'm relatively OK with WarningAsError during normal development.
Or let AppVeyor "just work without regression" (not prompted at all -> reported to discord)?

peppy:
I'm also not strongly against `WarningAsError` being always on.

@smoogipoo thoughts?

huoyaoyuan:
Or a condition in csproj to detect weather it's built by appveyor, similar to what we did for referencing native libs from framework.

smoogipoo:
I think it’s fine as always on. We’ve already had a warning in osu.Game that we’ve ignored for months.

huoyaoyuan:
Windows build for iOS provides warnings, and is skipped now via solution configuration. Once ppy/osu-framework#3044 get resolved, it can have managed side covered by Windows pool.

peppy (approved):
let's give this a shot

PR #5639, opened 2019-08-07
Lighten background on breaks

frenzibyte:
Supersedes #2629 

- [x] Depends on #5467 
- [ ] Depends on #5656 

Closes #2502

PR #6961, opened 2019-11-25
Don't pause during breaks

peppy:
- [x] Depends on #6960.
- [ ] Depends on #5639.


osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
```diff
@@ -313,7 +332,7 @@ private void load(OsuConfigManager config)
                 config.BindWith(OsuSetting.BlurLevel, BlurLevel);
             }
 
-            public bool IsBackgroundDimmed() => ((FadeAccessibleBackground)Background).CurrentColour == OsuColour.Gray(1 - (float)DimLevel.Value);
+            public bool IsBackgroundDimmed() => ((FadeAccessibleBackground)Background).CurrentColour == OsuColour.Gray(Math.Min(1 - (float)((FadeAccessibleBackground)Background).DimLevel, 1));
```
bdach: I'm a bit worried about the argument to `OsuColour.Gray` potentially going below 0 - might be worth throwing in an assert or a clamp?

osu.Game/Graphics/Containers/UserDimContainer.cs
```diff
@@ -37,7 +46,7 @@ public abstract class UserDimContainer : Container
 
         protected Bindable<bool> ShowVideo { get; private set; }
 
-        protected double DimLevel => EnableUserDim.Value ? UserDimLevel.Value : 0;
+        protected double DimLevel => DimOffset.Value + (EnableUserDim.Value ? UserDimLevel.Value : 0);
```
bdach: This seems potentially unclamped too, from both sides this time. Probably best to just apply the clamp here and leave the other expressions as they are (or even remove the `Math.Min(1 - DimLevel, 1)` calls altogether?)

peppy:
Re-opening without dependency.

PR #6968, opened 2019-11-25
Implement ability to delete own comments

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu/pull/6716


EVAST9919:
Closing because of unexpected refactoring in https://github.com/ppy/osu/pull/6716. It'll be better to make it a new pr rather than remake this one

PR #6970, opened 2019-11-25
Correct banned symbols configuration

huoyaoyuan:
I missed the previous pull request.
Syntax reference: https://github.com/dotnet/roslyn-analyzers/blob/master/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md
`M` represents method and `T` represents type here.

Do I need to add comment at somewhere? The BannedSymbols.txt itself doesn't support comment, but can be simulated.


huoyaoyuan:
If it's correctly configured, warnings will populate in command line and IDE builds immediately.

smoogipoo (approved):

PR #6979, opened 2019-11-27
Reduce the framebuffer scale of background blurs

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3033

Differences are unnoticeable.

| test | master | this branch |
| ---- | ------- | ------- |
| 0 blur (no scale) | osu_2019-11-27_15-22-47 | osu_2019-11-27_15-25-28 |
| 50% blur | osu_2019-11-27_15-21-22 | osu_2019-11-27_15-20-57 |
| 100% blur | osu_2019-11-27_15-22-26 | osu_2019-11-27_15-26-39 |
| 0 blur with 0.5 scale | - | osu_2019-11-27_15-28-09 |

With this I can maintain 60fps at retina-4k when dragging the slider, and no more framedrops when switching beatmap sets in song select.


Tom94:
I am confused about the 0 blur with 0.5 scale result. The image appears to be at the non-downscaled resolution, which is strange. In the code there is a special path that handles 0-blur amounts---you're probably just hitting that.

I think it's important that you only kick in with the 0.5 scale for sufficiently large sigmas... perhaps even make the scale 0.25 (and successive powers of 1/2) proportional to how big the blur sigma is?

smoogipoo:
I specifically disabled that, just to show that something _was_ happening. Which it is, it's very minor though.

That's actually a reasoning against increasing complexity and kicking it in for large sigmas - the reduction is almost unnoticeable anyway.

Tom94:
I see; sounds good for small blur sigmas! :)

There's a ton of performance improvement on the table for larger blur amounts, though, isn't it? Why stop at 0.5 when the blur might span gigantic amounts of pixels?

Tom94:
Also, this seems fundamental enough that we might actually want to have it at the framework-level within `BufferedContainer`. I think this can be flat-out applied to _any_ place that does blurring.

smoogipoo:
Not sure if I'm not implementing it correctly but I tried to see how it feels and it's choppy using POT (https://streamable.com/835jb):

```
if (bufferedContainer != null)
{
    Vector2 downScaleFactor = new Vector2(
        Math.Max(1, (int)MathF.Log(newBlurSigma.X + 1, 2) + 1),
        Math.Max(1, (int)MathF.Log(newBlurSigma.Y + 1, 2) + 1)
    );

    Vector2 blurScale = Vector2.Divide(Vector2.One, downScaleFactor);

    bufferedContainer.FrameBufferScale = blurScale;
    bufferedContainer.BlurTo(newBlurSigma * blurScale, duration, easing);
}
```

Linear factor introduces shimmering, and log_4 does much of the same. Don't really want to spend too much time perfecting this.

Tom94:
Sounds to me like this is a matter of tweaking the factors to get good results. I.e. downscale + blur being more aggressive and therefore needing a non-1:1 factor. Should only take a couple minutes to zero-in on the right factor using binary search (e.g. start with a way too large correction factor and then bisecting until there is no jumpyness left).

smoogipoo:
I don't really want to spend too much time tweaking this to get it perfect, in favour of completing the game in other ways. The primary reason for making this PR is because every beatmap change in song select causes a pretty large stutter for me and likely for others at retina+ resolutions. Same with the blur transitions into gameplay (drop to <30 fps).

As it is, it's a minimal implementation that lays the groundwork without the potential for causing regressions via something similar to the aforementioned issues. There are more considerations to take into account that I haven't in my code above, such as how much you want to scale in the first place to counter huge losses of quality.

If anyone else wants to improve on this further, feel free to.

peppy (dismissed):

PR #6991, opened 2019-11-28
Add support for --sdl command line arg

swoolcock:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3013 and framework update


huoyaoyuan:
Launch profile for VS and VSC required. And potentially in README.

peppy:
Probably not required as this is a temporary/experimental setting which will be removed once things are ready.

bdach:
I agree, let's not expose this too loudly to avoid getting unnecessary reports about things that still need ironing out.

peppy (approved):

osu.Desktop/Program.cs
```diff
@@ -22,8 +22,9 @@ public static int Main(string[] args)
         {
             // Back up the cwd before DesktopGameHost changes it
             var cwd = Environment.CurrentDirectory;
+            bool useSdl = args.Contains("--sdl");
```
UselessToucan: There is no need for the variable since there is only one usage

swoolcock: It looks cleaner to me.  If you disagree, make a pull request.

PR #7016, opened 2019-11-29
Add NewsArticleCover class and add missing content async loading logic

Game4all:
# Summary

- This PR implements the content async loading logic (`LoadChildContent`) which should have been implemented in the News Overlay skeleton PR.

- This PR also implements the `NewsArticleCover` class, used to load asynchronously and display the cover of news article cards. 

Article cards and News front page are coming in a next PR.

image

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
@@ -31,10 +31,6 @@ namespace osu.Game.Tests.Visual.Online
 
         private class TestNewsOverlay : NewsOverlay
         {
-            public TestNewsOverlay()
-            {
-            }
-
             public void ShowContent(NewsContent content) => LoadChildContent(content);
         }
 
--- a/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics.Shapes;
 using osu.Game.Graphics;
 using osu.Game.Overlays;
 using osu.Game.Overlays.News;
+using osuTK;
 
 namespace osu.Game.Tests.Visual.Online
 {
@@ -41,7 +42,7 @@ namespace osu.Game.Tests.Visual.Online
         {
             public TestNewsContent()
             {
-                Spacing = new osuTK.Vector2(0, 10);
+                Spacing = new Vector2(0, 10);
                 Add(new TestCard("https://osu.ppy.sh/help/wiki/shared/news/banners/CWC_2019_banner.jpg"));
                 Add(new TestCard("https://osu.ppy.sh/help/wiki/shared/news/banners/CWC_2019_banner.jpg"));
                 Add(new TestCard("https://osu.ppy.sh/help/wiki/shared/news/banners/CWC_2019_banner.jpg"));
--- a/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
@@ -1,6 +1,11 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Colour;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osu.Game.Graphics;
 using osu.Game.Overlays;
 using osu.Game.Overlays.News;
 
@@ -19,6 +24,8 @@ namespace osu.Game.Tests.Visual.Online
 
             AddStep(@"Show front page", () => news.ShowFrontPage());
             AddStep(@"Custom article", () => news.Current.Value = "Test Article 101");
+
+            AddStep(@"News article covers", () => news.ShowContent(new TestNewsContent()));
         }
 
         private class TestNewsOverlay : NewsOverlay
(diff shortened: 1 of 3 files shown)


osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
```diff
@@ -1,23 +1,74 @@
...
+            }
+
+            public void ShowContent(NewsContent content) => LoadChildContent(content);
+        }
+
+        private class TestNewsContent : NewsContent
+        {
+            public TestNewsContent()
+            {
+                Spacing = new osuTK.Vector2(0, 10);
```
bdach: This is going to be quite a nit, but why not just import the namespace? I don't see any potential pollutions.

osu.Game/Overlays/News/NewsArticleCover.cs
```diff
@@ -0,0 +1,44 @@
...
+                Anchor = Anchor.Centre;
+                Origin = Anchor.Centre;
+                RelativeSizeAxes = Axes.Both;
+                FillMode = FillMode.Fill;
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(LargeTextureStore store)
+            {
+                Texture = store.Get(url ?? "Headers/changelog");
```
bdach: Seems like a strange choice for a default for the *news* cover, there's even a
```suggestion
                Texture = store.Get(url ?? "Headers/news");
```
[in `osu-resources`](https://github.com/ppy/osu-resources/blob/master/osu.Game.Resources/Textures/Headers/news.png)?

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -56,13 +57,31 @@ private void load(OsuColour colours)
...
+            }, (loadChildCancellation = new CancellationTokenSource()).Token);
+        }
+
+        public void ShowArticle(string article)
+        {
+            Current.Value = article;
             Show();
         }
+
+        public void ShowFrontPage() => ShowArticle(null);
```
bdach: This looks a bit like a "magic" API. What's this `article` string? An URL? Article slug? Why's `null` the front page? I'd like this to be easier to understand and read.

osu.Game/Overlays/News/NewsArticleCover.cs
Game4all: Good catch, I completely missed that 😅

osu.Game/Overlays/NewsOverlay.cs
Game4all: Actually in the final implementation, the Current bindable should contain a struct about a specific news article but since the web API struct for news isn't defined yet, I use string as a placeholder for the moment. Asking to show article null should instead should show the news front page (whose implementation is to come later)

osu.Game/Overlays/News/NewsArticleCover.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.Textures;
+
+namespace osu.Game.Overlays.News
+{
+    public class NewsArticleCover : CompositeDrawable
```
peppy: shouldn't this class handle showing the title/author/date?

image

osu.Game/Overlays/News/NewsArticleCover.cs
```diff
@@ -0,0 +1,44 @@
...
+        public NewsArticleCover(string url)
+        {
+            NewsArticleCoverBackground background;
+            InternalChild = new DelayedLoadWrapper(background = new NewsArticleCoverBackground(url));
+
+            background.OnLoadComplete += bg => bg.FadeInFromZero(400, Easing.Out);
+        }
+
+        [LongRunningLoad]
+        private class NewsArticleCoverBackground : Sprite
```
peppy: Note that this has a gradient when displayed in the news post
`linear-gradient(180deg,transparent,hsla(var(--base-hue),10%,10%,.5))`

osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
```diff
@@ -1,23 +1,71 @@
...
+            }
+
+            private class TestCard : Container
+            {
+                public TestCard(string url)
+                {
+                    RelativeSizeAxes = Axes.X;
+                    Height = 250;
+                    Masking = true;
+                    CornerRadius = 4;
```
peppy: shouldn't this be part of the `NewsCover`?

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -56,13 +57,31 @@ private void load(OsuColour colours)
...
+            loadChildCancellation?.Cancel();
+
+            LoadComponentAsync(newContent, c =>
+            {
+                content.Child = c;
+                content.FadeIn(300, Easing.OutQuint);
+            }, (loadChildCancellation = new CancellationTokenSource()).Token);
+        }
+
+        public void ShowArticle(string article)
```
peppy: this isn't being used anywhere, right? probably best save it for the PR where it's being used.

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -56,13 +57,31 @@ private void load(OsuColour colours)
...
 
             header.Current.BindTo(Current);
+
             Current.TriggerChange();
         }
 
-        public void ShowFrontPage()
+        private CancellationTokenSource loadChildCancellation;
+
+        protected void LoadChildContent(NewsContent newContent)
```
peppy: `LoadAndShowContent` probably more clear.

peppy (changes requested):

osu.Game/Overlays/News/NewsArticleCover.cs
Game4all: This class was intended to only provide async background image loading for a news article card (image shown above), whose implementation is to come soon (as there is no API struct avalaible for news articles to base the component on, unless it is fine to use a dummy struct to develop the component?).

Game4all:
On a second thought, implementing this as its own separated component isn't needed as it fairly easy to implement async background image loading directly in the component. Closing.

PR #7017, opened 2019-11-30
Implement basic layout for rankings overlay

EVAST9919:


osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -0,0 +1,198 @@
...
+
+        public void ShowCountry(Country requested)
+        {
+            if (requested == null)
+                return;
+
+            Show();
+
+            if (country.Value?.FlagName == requested.FlagName)
+                return;
```
bdach: This guard feels wrong and kind of defeats the point of `country` being a bindable. I'd probably rather have `Country` implement `IEquatable<Country>`, in which case this goes away magically and you get the same semantics.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -0,0 +1,198 @@
...
+            request?.Cancel();
+
+            cancellationToken = new CancellationTokenSource();
+
+            switch (scope.Value)
+            {
+                default:
+                    contentPlaceholder.Clear();
+                    loading.Hide();
+                    return;
```
bdach: I don't quite understand this case. I understand `RankingsScope.Spotlights` you can't do for now (due to ppy/osu-web#5356?) but other than that, you shouldn't just display nothing.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        private void loadTable(Drawable table)
+        {
+            LoadComponentAsync(table, t =>
+            {
+                contentPlaceholder.Clear();
+                contentPlaceholder.Add(t);
+                loading.Hide();
+            }, cancellationToken.Token);
```
bdach: I'm not 100% certain this cancellation logic is bug-free due to the shared field. I'll have to look in detail in an IDE, but if `this` is captured as a closure instead of just `cancellationToken` this might result in races.

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: It exists only in the test scene(until everything is ready), so I see nothing wrong with having this as a temporary solution

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: I have almost the same implementation in `RankingsTablesTestScene`, but if it can be improved - I'm happy to do so

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -184,12 +185,12 @@ private void createScoreTable()
 
         private void loadTable(Drawable table)
         {
-            LoadComponentAsync(table, t =>
+            showTableDelegate = Schedule(() => LoadComponentAsync(table, t =>
```
peppy: could you explain how the schedule helps here?

EVAST9919: I thought this is what this comment is about
>I'm not 100% certain this cancellation logic is bug-free due to the shared field. I'll have to look in detail in an IDE, but if this is captured as a closure instead of just cancellationToken this might result in races.

osu.Game/Overlays/RankingsOverlay.cs
peppy: moving the `cancellationToken` construction was enough to make it clear there was no issue. adding a schedule/cancel call has no meaning and is not correct.

peppy:
Is there is a list of things remaining before this can be added to the toolbar? Is the charts section blocking it?

EVAST9919:
> Is there is a list of things remaining before this can be added to the toolbar? Is the charts section blocking it?
Yes and also this one : https://github.com/ppy/osu/pull/6035

peppy:
I've done some clean-up on this. Please check my commits and make sure you agree with them. https://github.com/ppy/osu/pull/7017/commits/5f9b9631ef38bae53c320a9aa1512250399b1f47 is of importance – "title" or "header" components shouldn't be enforcing constraints. It took me minutes to find how this whole system was even working, which shouldn't be the case.

peppy (dismissed):

peppy:
Please fix up the failing tests if you're happy with the change in logic (you may want to add similar tests to the overlay test scene now).

EVAST9919:
Tests are failing because you changed the logic a bit (switching scope doesn't affect the country and vice versa). For example: having selecting a country must trigger the scope to became `Performance`, and selecting another scope must reset the country

peppy:
Please re-read what I said. I am aware of why they are failing – this is intentional (read https://github.com/ppy/osu/pull/7017#issuecomment-561976426) and tests should be added to the overlay test scene ensuring the correct behaviour there.

smoogipoo (approved):

PR #7034, opened 2019-12-01
Fix overlays closing when dragging from in/out or out/in

### LINKED ISSUE #6950: Left/Right clicking/holding & dragging combinations have an unexpected behaviour.
**Describe the bug:** 
I encountered this on the confirmation dialog when opening external links. Basically when you get prompted and click outside the confirmation dialog the expected behaviour, the dialog getting closed, is the outcome. 
While holding the left mouse button outside the dialog and dragging it into the dialog though, should not disgard the dialog, which is not the case. 
Holding the left mouse button and additionally holding the right mouse button and dragging the mouse outside the dialog closes the dialog, which should not happen. 

You can compare the general behaviour of how it should look like with the close button of any windows window. Dragging out the mouse, while holding the left mouse button over the close button does not close the window. Either does the combination with the right mouse button do.
https://www.youtube.com/watch?v=ZnK1xEE-i5o
**osu!lazer version:** 
2019.1122.0
**Computer Specifications:** 
i7 6800k, GTX 1080, 16GB ram, 500gb ssd, Windows 10 64-bit Pro

Joehuu:
Fixes https://github.com/ppy/osu/issues/6950.

I've simplified it to `OnMouseDown` and `OnMouseUp`.

`OnMouseDown` is equivalent to `OnDragStart`, but `OnDragStart` needs a drag. It needs to handle the bools, ~since~ so that if we left click outside of the overlay, then right click inside and drag outside, the `closeOnDragEnd` will be false and the overlay will not close.

`OnMouseUp` is equivalent to `OnClick` and `OnDragEnd`, but `OnClick` is only left click and `OnDragEnd` needs a starting drag. It'll handle `Hide()`.

Issues fixed in issue:
- [X] While holding the left mouse button outside the dialog and dragging it into the dialog though, should not disgard the dialog, which is not the case.
- [X] Holding the left mouse button and additionally holding the right mouse button and dragging the mouse outside the dialog closes the dialog, which should not happen.
	- Note: this can be done with just one of the buttons.

Other changes:
- Right clicking outside in place will close the overlay (from what I said on top).

### DIFF
--- a/osu.Game/Graphics/Containers/OsuFocusedOverlayContainer.cs
+++ b/osu.Game/Graphics/Containers/OsuFocusedOverlayContainer.cs
@@ -67,33 +67,21 @@ namespace osu.Game.Graphics.Containers
         // receive input outside our bounds so we can trigger a close event on ourselves.
         public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => BlockScreenWideMouse || base.ReceivePositionalInputAt(screenSpacePos);
 
-        protected override bool OnClick(ClickEvent e)
-        {
-            if (!base.ReceivePositionalInputAt(e.ScreenSpaceMousePosition))
-                Hide();
+        private bool closeOnMouseUp;
 
-            return base.OnClick(e);
+        protected override bool OnMouseDown(MouseDownEvent e)
+        {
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

smoogipoo:
Seems to work fine.

PR #7045, opened 2019-12-02
Display user rank in menu

### LINKED ISSUE #2510: Player rank not displayed in menu
image

There is no 'easy' way to view your rank in the menu, it requires going into the social browser.
Rank progress is also not displayed anywhere like rank.

Leostayner:
Resolves #2510
This PR is not ready yet, I am thinking of setting up in the tab user a container with local and global raking equal to the web page (I tried to get this layout ready but apparently is not possible).

image   or   image

But being new to C # I have some difficulties, I managed to get the rakings values ​​and put them in a simple text, now i am trying to assemble a proper container for this information, i am trying to create an `updatableRank.cs` file and a `drawableRank.cs` file, does this seem correct?

for now it's like this, but it was a test to get the values.
image

### DIFF
--- a/osu.Game/Users/Drawables/DrawableFlag.cs
+++ b/osu.Game/Users/Drawables/DrawableFlag.cs
@@ -19,7 +19,7 @@ namespace osu.Game.Users.Drawables
         {
             this.country = country;
         }
-
+        
         [BackgroundDependencyLoader]
         private void load(TextureStore ts)
         {
--- a/osu.Game/Users/Drawables/DrawableFlag.cs
+++ b/osu.Game/Users/Drawables/DrawableFlag.cs
@@ -19,7 +19,7 @@ namespace osu.Game.Users.Drawables
         {
             this.country = country;
         }
-        
+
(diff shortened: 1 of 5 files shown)


bdach:
* Please use descriptive names for PR titles. It's impossible to tell what a change does by looking at an issue number. Do however refer to the issue in the PR description.
* If you'd like feedback on a PR, using the *Draft* option seems more appropriate.
* Please don't commit commented code. You either need it, in which case it should be uncommented, or no longer do, in which case there's git history.
* Appearance-wise your changes don't look great and are confusing to new users (to a newcomer, what do these numbers mean?) That's why I mentioned that this needs an actual design from the team in the issue before proceeding with implementation.

peppy:
@Leostayner are you planning to continue work on this?

EVAST9919:
Also what about switching rulesets? It should refetch the data somehow

peppy:
Closing this for now. Feel free to reopen after addressing the concerns above.

PR #7047, opened 2019-12-03
Decouple LegacyScoreInfo/APILegacyScoreInfo from ScoreInfo

peppy:
As part of an effort to remove inheritance of model classes.

### DIFF
--- a/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs
+++ b/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs
@@ -34,12 +34,10 @@ namespace osu.Game.Rulesets.Catch.Difficulty
         {
             mods = Score.Mods;
 
-            var legacyScore = Score as LegacyScoreInfo;
-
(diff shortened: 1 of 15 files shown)


smoogipoo (approved):

osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs
```diff
@@ -34,12 +34,10 @@ public override double Calculate(Dictionary<string, double> categoryDifficulty =
         {
             mods = Score.Mods;
 
-            var legacyScore = Score as LegacyScoreInfo;
-
-            fruitsHit = legacyScore?.Count300 ?? Score.Statistics[HitResult.Perfect];
-            ticksHit = legacyScore?.Count100 ?? 0;
-            tinyTicksHit = legacyScore?.Count50 ?? 0;
-            tinyTicksMissed = legacyScore?.CountKatu ?? 0;
+            fruitsHit = Score?.GetCount300() ?? Score.Statistics[HitResult.Perfect];
```
huoyaoyuan: `?.` not needed here.
Can be caught later when reviewing nullable reference type.

PR #7050, opened 2019-12-03
Remove some remaining untyped boxing calls

huoyaoyuan:
The remaining scenarios are all `Nullable<T>` related.
Since CLR doesn't support implementing interfaces conditionally, `Nullable<T>` can't implement any one.

Please always prefer `==` over `Equals` when writing code.

Note: `==` and `Equals` have different behavior on `NaN`. I think `NaN` should be unexpected in these scenarios.

### DIFF
--- a/osu.Game/Rulesets/Objects/SliderPath.cs
+++ b/osu.Game/Rulesets/Objects/SliderPath.cs
@@ -273,14 +273,6 @@ namespace osu.Game.Rulesets.Objects
             return p0 + (p1 - p0) * (float)w;
         }
 
-        public bool Equals(SliderPath other)
-        {
-            if (ControlPoints == null && other.ControlPoints != null)
-                return false;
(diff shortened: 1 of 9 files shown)


osu.Game/Rulesets/Objects/SliderPath.cs
```diff
@@ -273,14 +273,6 @@ private Vector2 interpolateVertices(int i, double d)
             return p0 + (p1 - p0) * (float)w;
         }
 
-        public bool Equals(SliderPath other)
-        {
-            if (ControlPoints == null && other.ControlPoints != null)
-                return false;
-            if (other.ControlPoints == null && ControlPoints != null)
-                return false;
```
huoyaoyuan: Unrelated, but here's the only location I caught together.
`null` converts to empty span.

smoogipoo (approved):

Issue #7039, opened 2019-12-02
BeatmapSetOverlay doesn't work as expected on first ShowBeatmapSet call


EVAST9919:
I'm talking about `ShowBeatmapSet(BeatmapSetInfo set)` (already populated beatmap set). For `FetchAndShowBeatmapSet` it works fine.

How to reproduce: just simply reorder tests in `TestSceneBeatmapSetOverlay` in which local beatmap will be first test.

How it's looks like: 
![Screenshot_1](https://user-images.githubusercontent.com/22874522/69940933-82f6a780-14f4-11ea-8c5e-2861fd825260.png)

How it's suppose to looks like:
![Screenshot_2](https://user-images.githubusercontent.com/22874522/69940950-8b4ee280-14f4-11ea-85ec-11cb99ae7bcf.png)

iiSaLMaN:
I have a good fix for this (one-liner), but adding a test to avoid regressions from it doesn't seem to be straightforward at all..

peppy:
Can you explain the difficulties you are having with the test? Maybe we can provide a suggestion.

iiSaLMaN:
So the underlying issue here is that the logic of selecting the first enabled tab doesn't work properly when invoked before the tabs are "alive" (added to `AliveInternalChildren`), the fix is simply to use `InternalChildren` instead:
- `TabContainer.TabItems.FirstOrDefault(...)` -> `TabContainer.FirstOrDefault(...)`


But the difficulty with testing this is I have to set a beatmap before the ruleset selector is loaded and the children become alive to replicate this issue, and I have no idea how to do this properly.

bdach:
Does ppy/osu-framework#2813 seem relevant? Seems like this also could be a tab lifetime problem which has shown up in other places before, so maybe what we're looking for is a framework-side resolution.

iiSaLMaN:
Will take a look at that.

PR #7054, opened 2019-12-03
Fix unnecessary beatmap fetch for beatmap panel in multiplayer room

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu/issues/7039

In multiplayer room we already have a populated beatmapset, so there's no neeed to fetch one when clicking on beatmap panel

### DIFF
--- a/osu.Game/Overlays/Direct/DirectPanel.cs
+++ b/osu.Game/Overlays/Direct/DirectPanel.cs
@@ -44,10 +44,13 @@ namespace osu.Game.Overlays.Direct
 
         protected override Container<Drawable> Content => content;
 
-        protected DirectPanel(BeatmapSetInfo setInfo)
+        private readonly bool fetchOnline;
+
+        protected DirectPanel(BeatmapSetInfo setInfo, bool fetchOnline = false)
         {
             Debug.Assert(setInfo.OnlineBeatmapSetID != null);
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/Direct/DirectPanel.cs
```diff
@@ -44,10 +44,13 @@ public abstract class DirectPanel : Container
 
         protected override Container<Drawable> Content => content;
 
-        protected DirectPanel(BeatmapSetInfo setInfo)
+        private readonly bool fetchOnline;
+
+        protected DirectPanel(BeatmapSetInfo setInfo, bool fetchOnline = false)
```
smoogipoo: Any reason this is `false` by default whereas `DirectGridPanel` has `true` by default?

smoogipoo (changes requested):

peppy:
I don't really like this flow, and think it may be best to always fetch for safety.

Passing bools down the heirarchy is prone to error – if we're going to bypass fetches it should be done in an automated and well-tested way (ie. BeatmapSetOverlay should decide whether it needs to fetch or not, based on whether the incoming request has everything it needs for display).

EVAST9919:
As mentioned above, this needs different approach

PR #7067, opened 2019-12-05
Make expert plus colour on tooltip more readable

### LINKED ISSUE #7058: Colour of X difficulty is hard to read in tooltip
Comparing:
image
image

Needs to change the colour of either background or the text.

GSculerlor:
Resolve #7058 

Using same colour with website (# 999)

Example: 
image
image

### DIFF
--- a/osu.Game/Beatmaps/Drawables/DifficultyIcon.cs
+++ b/osu.Game/Beatmaps/Drawables/DifficultyIcon.cs
@@ -168,7 +168,7 @@ namespace osu.Game.Beatmaps.Drawables
 
                 difficultyName.Text = beatmap.Version;
                 starRating.Text = $"{beatmap.StarDifficulty:0.##}";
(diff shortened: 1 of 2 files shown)


peppy:
Is there a reason you didn't make the change in the function?

GSculerlor:
`ForDifficultyRating` is also used on some places like [here](https://github.com/ppy/osu/blob/e213b338cea584e5d4b66c43b53a4d7fdef465dc/osu.Game/Beatmaps/Drawables/DifficultyIcon.cs#L78) for example to provide colour to the difficulty icon itself

peppy:
Have you confirmed the colour on osu-web is *only* different in tooltips? That sounds like a weird decision..

GSculerlor:
Ah, missed a spot there, the `Expert` is also different (web is #aa88ff and lazer using #8866ee). Working on it

peppy (approved):

PR #7070, opened 2019-12-05
Fix DialogButton snapping back on click

mcendu:
Copy of #6839. Fixes `DialogButton` ["snapping back"](https://github.com/ppy/osu/files/3840830/glitch.mkv.gz) on click.

Bug demonstrated near 0:00:05 when button was clicked.


peppy:
I can't reproduce this issue. Could you please provide reproduction steps?

mcendu:
In `TestSceneDialogOverlay`:
* Click any of the buttons.
* Open another dialog, then press Esc. This animation is different from when clicking a button.

peppy:
Thanks. Your PR isn't rebased correctly and I wanted to fix the aniamtion a bit more, so I took what you did and committed under https://github.com/ppy/osu/pull/7082 for cleanliness.

PR #7080, opened 2019-12-06
Fix SettingsItem bindable logic

peppy:
More correctly use the new `BindableWithCurrent` backing.

Previous logic would mean the `restoreDefaultButton` could potentially be binding to a bindable that went out of context.

- [x] Depends on https://github.com/ppy/osu-framework/issues/3088.

### DIFF
--- a/osu.Game/Overlays/Settings/SettingsItem.cs
+++ b/osu.Game/Overlays/Settings/SettingsItem.cs
@@ -53,27 +53,10 @@ namespace osu.Game.Overlays.Settings
             }
         }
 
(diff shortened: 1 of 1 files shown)


smoogipoo:
This seems to have broken all slider controls in the settings (e.g. UI Scale).

peppy:
Tracking at https://github.com/ppy/osu-framework/issues/3088.

smoogipoo (approved):

smoogipoo:
Approved but not going to merge without a framework release.

PR #7081, opened 2019-12-06
Add the ability to create settings UI components from a class with exposed drawables

peppy:
- [x] Depends on #7080.


smoogipoo (approved):
g2g but depends on the underlying PR

PR #7082, opened 2019-12-06
Improve animation of popup dialog buttons

peppy:


smoogipoo:
Seems like the force push broke the changes in this PR (supposed to have some animation improvements).

peppy:
Oops, fixed.

smoogipoo (approved):
seems ok

Issue #5845, opened 2019-08-26
Creating 3 point sliders still uses Bezier curve shape


malachimdube:
**Describe the missing feature:** 
In Osu!Stable, creating a slider with 3 points in the editor creates a curve based on the circle containing all three points rather than using the standard bezier curve shape. Currently Osu!Lazer is capable of handling such sliders in existing maps correctly (displays correctly and behaves correctly when moving points around), but will always create bezier sliders when making new hit objects in the editor. 

A large portion of mappers would also likely agree that any segment of a slider between red anchors or ends with only one grey tick should use this circular curve shape as it is much easier to work with and looks a lot better, but this would be a new feature.

**Proposal designs of the feature:** 
If a slider is created with exactly three points, it should use the circular curve shape rather than the bezier curve shape to match Osu!Stable. It should not be impossible to create 3 point bezier sliders, as creating a slider with 4 points and then deleting one results in a 3 point Bezier curve in Osu!Stable.

Making all slider sections containing three points use this shape could be achieved via approximation, such as described here: http://spencermortensen.com/articles/bezier-circle/.

peppy:
Both of these additions are planned.

FullTablet:
While mappers from stable are already used to 3 points making a circular slider, and 4 points + 1 anchor deletion making a quadratic Bezier curve. Don't you think that behavior is counter-intuitive for new mappers?

Also, the current slider format doesn't allow Bézier sliders that have consecutive anchor points in the same position (since those are interpreted as Red anchor points by the game client). Nor does it allow concatenating Bézier curves with circular curves (like Bézier curves are concatenated by using red anchor points).

What about making a new anchor color (for example, green) for making circular sliders?

1. When clicking 3 points to make a slider, the mid point is made Green. This makes a circular slider.

2. Anchor points colors can be modified by right clicking and Ctrl-click, just like in stable.
Delete <-> White Anchor <-> Red Anchor <-> Green Anchor.
Two Green anchors can't be consecutive (and even if someone made one by editing the .osu file, the color of the anchor point that follows the first green anchor point is ignored when generating the slider).

3. When there is only White and Red anchors, the slider behaves as a concatenation of Bezier curves (like it does in stable for Bezier type sliders).

4. Each Green anchor is the middle anchor point of a circular path. Regardless of its color, the point that follows it is both the end point of the circular curve, and the start of the next curve if there are more anchor points. The path of the slider can continue either with a Bezier curve or with another circular curve.

[...]

PR #7083, opened 2019-12-06
Allow PerfectCircle sliders to be placed and edited

mcendu:
Implements #5845


smoogipoo:
Sorry for not assigning the original issue to myself. I've actually completely reworked slider paths to handle this stuff a little better, including supporting multiple path types. As a result, these files have changed significantly so I'm going to close this PR in favour of my upcoming one.

PR #7084, opened 2019-12-06
Add rate adjustment setting to HT/DT/DC/NC

peppy:
- [ ] Depends on https://github.com/ppy/osu/pull/6432.


peppy:
Will re-open as new PR. Removing hackiness.

PR #7099, opened 2019-12-08
Fix mod buttons being selected when drag scrolling overlay

### LINKED ISSUE #3174: Gameplay mod buttons should not be selected while drag scrolling
The overlay is now a scroll container and buttons should not be selected when dragging up and down while scrolling.

Joehuu:
Fixes https://github.com/ppy/osu/issues/3174.

`OnMouseUp` is still there to handle right clicks. It can be completely removed once https://github.com/ppy/osu-framework/issues/1594 is resolved.

### DIFF
--- a/osu.Game/Overlays/Mods/ModButton.cs
+++ b/osu.Game/Overlays/Mods/ModButton.cs
@@ -178,11 +178,9 @@ namespace osu.Game.Overlays.Mods
 
         protected override bool OnClick(ClickEvent e)
         {
-            scaleContainer.ScaleTo(1, 500, Easing.OutElastic);
-
             SelectNext(1);
 
-            return base.OnClick(e);
+            return true;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/Mods/ModButton.cs
```diff
@@ -180,6 +176,15 @@ protected override bool OnMouseUp(MouseUpEvent e)
             return true;
         }
 
+        protected override bool OnClick(ClickEvent e)
+        {
+            scaleContainer.ScaleTo(1, 500, Easing.OutElastic);
```
peppy: should not be required.

osu.Game/Overlays/Mods/ModButton.cs
```diff
@@ -180,6 +176,15 @@ protected override bool OnMouseUp(MouseUpEvent e)
...
+
+            SelectNext(1);
+
+            return base.OnClick(e);
```
peppy:
```suggestion
            return true;
```

peppy (changes requested):

peppy (approved):

PR #7101, opened 2019-12-08
Add missing async content loading logic to NewsOverlay

Game4all:
Split from #7016

Adds the missing bits for loading asynchronously and showing content on the NewsOverlay

### DIFF
--- a/osu.Game/Overlays/NewsOverlay.cs
+++ b/osu.Game/Overlays/NewsOverlay.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Threading;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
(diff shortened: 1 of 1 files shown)


osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
```diff
@@ -2,22 +2,28 @@
...
             AddStep(@"Show", news.Show);
             AddStep(@"Hide", news.Hide);
 
             AddStep(@"Show front page", () => news.ShowFrontPage());
             AddStep(@"Custom article", () => news.Current.Value = "Test Article 101");
         }
+
+        private class TestNewsOverlay : NewsOverlay
+        {
+            public new void LoadAndShowChild(NewsContent content) => base.LoadAndShowChild(content);
```
peppy: no usage of this?

Game4all: This is being used in upcoming PR for the news overlay for testing purposes. Guess it is better to keep that for when it is being used?

peppy: generally we avoid adding any new functions without having test coverage.

peppy:
I'm still confused. You forced pushed this but the new method is still not used. Either make a test for it or combine it with the follow-up PR.

Game4all:
Closing as the commit is being cherry-picked in #7192

PR #7114, opened 2019-12-09
Allow mods to apply to track, not clock

peppy:


osu.Game/Rulesets/Mods/ModTimeRamp.cs
```diff
@@ -59,23 +58,8 @@ private void applyAdjustment(double amount)
...
-                    tempo.TempoAdjust *= adjust;
-                    break;
-
-                default:
-                    clock.Rate /= lastAdjust;
-                    clock.Rate *= adjust;
-                    break;
-            }
+            track.Tempo.Value /= lastAdjust;
+            track.Tempo.Value *= lastAdjust;
```
bdach: Am I missing something, or should this be
```suggestion
            track.Tempo.Value *= adjust;
```
I haven't run this to be fair, but looks suspicious on my eye test of the diff.

peppy: good catch. this ugly code won't exist in the follow-up PR, thankfully.

smoogipoo (approved):

PR #7116, opened 2019-12-09
Implement segmented slider paths

### LINKED ISSUE #3949: Rework of slider anchors behavior
Its quite hard to deal with sliders when you want to make a high precised shape, basically for now you cant make a perfect circle shape without using either 3-white anchor thing, or some external tool like [Slider Studio](https://slider.little.moe/) by [Little](https://osu.ppy.sh/users/Little)

So yeah, I think there should be an option of using these 3-white-anchored circle sliders inside the already created slider 
Lets say, I want to make that circle in between of the 2 red anchors, for now its impossible:

What should appear is this:

So even using that, the shape isn't 100% precise. 

So another thing, is that when you use lots of white anchors, if you move one white anchor, the entire slider shape moves for some reason and it becomes quite frustrating to perfect the shape.
I would want to add a video about that issue, but I guess you can just try making a blanket around a circle with a slider (using many white anchors) to understand what I am talking about.

Its close to impossible to do that thing using 4+ white anchors.

My guess is that you can add the 3rd type of anchors that would make a perfect circle.

### LINKED ISSUE #5845: Creating 3 point sliders still uses Bezier curve shape

smoogipoo:
Demonstrated in https://twitter.com/smoogipoo/status/1202897871088939008

Resolves #5845
Resolves #3949

It's a pretty tough change to go through because `SliderPath` is so integral to everything. The only sort of unnecessary change is getting segmented placements to work - all of which is contained within `SliderPlacementBlueprint` and a pretty small change itself.

The rest is updating various code paths to support the new control point models.

There are still further refactorings to be done in future PRs, including: binding to the list to handle additions/removals in `PathControlPointVisualiser` and giving `PathControlPointPiece` a reference rather than an index and binding to position/type changes internally.


osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -259,6 +253,45 @@ private void readCustomSampleBanks(string str, SampleBankInfo bankInfo)
             bankInfo.Filename = split.Length > 4 ? split[4] : null;
         }
 
+        private PathControlPoint[] convertControlPoints(Vector2[] vertices, PathType type)
+        {
+            if (type == PathType.PerfectCurve)
+            {
+                if (vertices.Length == 3)
```
peppy: might read better as
```csharp
	            if (vertices.Length != 3)
                    type = PathType.Bezier;
                else if (isLinear(vertices))
                    // osu-stable special-cased colinear perfect curves to a linear path
                    type = PathType.Linear;
```

osu.Game/Rulesets/Objects/PathControlPoint.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        /// <summary>
+        /// Invoked when any property of this <see cref="PathControlPoint"/> is changed.
+        /// </summary>
+        internal event Action Changed;
+
+        /// <summary>
+        /// Creates a new <see cref="PathControlPoint"/>.
+        /// </summary>
+        public PathControlPoint()
```
peppy: i'd usually do this the other way around (have the `ctor` with parameters call the one without) since you're initialising with already-initialised values.

peppy (changes requested):

osu.Game/Rulesets/Objects/PathControlPoint.cs
smoogipoo: I'm confused, this is doing exactly what you want?

peppy:
```
        /// <summary>
        /// Creates a new <see cref="PathControlPoint"/> with a provided position and type.
        /// </summary>
        /// <param name="position">The initial position.</param>
        /// <param name="type">The initial type.</param>
        public PathControlPoint(Vector2 position, PathType? type = null) : this()
        {
            Position.Value = position;
            Type.Value = type;
        }


/// <summary>
        /// Creates a new <see cref="PathControlPoint"/>.
        /// </summary>
        public PathControlPoint()
{
            Position.ValueChanged += _ => Changed?.Invoke();
            Type.ValueChanged += _ => Changed?.Invoke();
}
```

smoogipoo: Oh I misread, thought this comment was inside SliderPath... >_<

peppy (approved):

PR #7121, opened 2019-12-09
Add rate adjustment settings for all time mods (DT/HT/NC/DC/WU/WD)

### LINKED ISSUE #7162: Mods that change clock speed are not resetting
**Bug**
When a song is selected and a mod like `DT`, `NC`, `HT` or `DC` is picked, the music clock is changed but is being multiplied and not reset. For example, if nightcore is picked, it applies the clock speed multiplier for DT then adds NC multipliers on top, making the map a lot faster (or slower in daycore's case) than expected. If one of these mods is selected, enabling/disabling any other mod causes them to multiply up as well.

**Video** 
https://drive.google.com/open?id=1iSnLUYN7ZTmkvnwbdTrrh9XREPRoT81d
https://drive.google.com/open?id=1jNNJnNQ7jDOK20SDlLuKKB6jRs7JEbqo

**Version**
osu!lazer version 052d0f4 release build

peppy:
Resolves #7162.

- [x] Depends on https://github.com/ppy/osu/pull/6432.
- [x] Depends on https://github.com/ppy/osu-framework/pull/3095.
- [x] Depends on https://github.com/ppy/osu-framework/pull/3087.
- [x] Depends on https://github.com/ppy/osu/pull/7172.


huoyaoyuan:
How would these mods be ranked/scored then? It makes sense to be ranked only at default rate, but that requires some more mechanism.

peppy:
No idea. Ranking is not a consideration yet.

peppy:
Test failure fixed in https://github.com/ppy/osu-framework/pull/3095

smoogipoo (dismissed):
👍

smoogipoo:
One thing I noticed is that if you enable DT then exit song select and go back in, the customise button will be greyed out.

peppy:
Once https://github.com/ppy/osu/pull/7172 is merged, I will rebase this PR, removing the #7170 commits and using https://github.com/ppy/osu/pull/7172 logic instead.

l3lackShark:
I encountered 2 bugs with this branch. 
1. Approach circle of the first note makes "jumps" in speed when the speed multiplier is applied.
2. Several beatmaps crash on transition to gameplay, including this one: https://osu.ppy.sh/beatmapsets/659007#osu/1768802
<details>
  <summary>Spoiler</summary>
  
  
2019-12-13 00:13:16: Unhandled exception has been allowed with 0 more allowable exceptions .
2019-12-13 00:13:16: An unhandled error has occurred.
2019-12-13 00:13:16: System.ArgumentNullException: Value cannot be null.
2019-12-13 00:13:16: at System.Threading.Monitor.ReliableEnter(Object obj, Boolean& lockTaken)
2019-12-13 00:13:16: at osu.Framework.Graphics.Drawable.dispose(Boolean isDisposing)
2019-12-13 00:13:16: at osu.Framework.Graphics.Drawable.Finalize()
  
</details>

peppy:
@l3lackShark the second (crash) bug has been fixed via dependent PRs. The jump in speed may still exist, but can be addressed elsewhere. after this is merged into master, please open an issue if it still occurs!

smoogipoo (approved):

PR #7129, opened 2019-12-09
Fix osu!catch catcher not scaling down correctly

### LINKED ISSUE #7100: Catcher sprite size too large
**Describe the bug:** Catcher sprite size in catch in the default lazer skin has regressed since allowing display always at native sizes (#7078). Since I'm unsure how this should be dealt with (whether by reverting or resizing the asset at `osu-resources`), opening an issue.

| Before | After |
| :-: | :-: |
| image | image |

**osu!lazer version:** commit 21ceb7f and later

**Logs:** not required (visual problem)

peppy:
Closes https://github.com/ppy/osu/issues/7100.


bdach (approved):
Did run this one, looks all good now 👍

PR #7130, opened 2019-12-09
Fix song select filters not reapplied if in a child screen

### LINKED ISSUE #6980: "Show converted beatmaps" doesn't take effect when resuming song select
https://streamable.com/bh4fm

Seems to only take effect when changing the ruleset or _entering_ song select, but not when it's resumed from gameplay.

peppy:
Closes https://github.com/ppy/osu/issues/6980.


smoogipoo (approved):
Seems ok.

PR #7140, opened 2019-12-10
Add VS launcher profile for tournament client

huoyaoyuan:
Tournament should be the only one that doesn't have entry itself?


peppy:
Am i missing something here? https://github.com/ppy/osu/blob/master/.vscode/launch.json#L77

peppy:
Oh, this is for visual studio, not code?

huoyaoyuan:
Right. It doesn't live in `.vs` folder.

PR #7142, opened 2019-12-10
Lighten background during breaks

### LINKED ISSUE #2502: Background stays dimmed during break times
This feature was just changed (on 100% dim) on stable to match lazer's current behaviour, which is to keep the background dim the same during gameplay and break time. I would argue that one of the few times you actually see the background is during break times and thus there should be an option to show the background during them.

I don't know what the general consensus around this issue is as I have not heard any player mention that the background should stay dimmed (even if it's set to 100% or something)

frenzibyte:
- Adds "Lighten playfield background during break times" feature as a setting in Gameplay→General

Supersedes #5639 – due to wrong pushed merge commits in old branch.
Closes #2502


peppy:
no test

iiSaLMaN:
> no test

Tests added :P

peppy:
Can you explain the rationale behind renaming the previous test and putting this in a new test? Generally we name test scenes after the class they are testing.

osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
```diff
@@ -1,423 +1,82 @@
...
+            AddStep("set break", () => isBreakTime.Value = true);
+            AddWaitStep("wait for dim", 5);
+            AddAssert($"is current dim {userDim}", () => container.DimEqual(userDim));
 
-            AddUntilStep("Wait for Player Loader to load", () => playerLoader.IsLoaded);
-            AddStep("Move mouse to center of screen", () => InputManager.MoveMouseTo(playerLoader.ScreenPos));
-            AddUntilStep("Wait for player to load", () => player.IsLoaded);
-        }
+            AddStep("set lighten during break", () => lightenDuringBreaks.Value = true);
+            AddWaitStep("wait for dim", 5);
```
peppy: use until, not wait.

osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
```diff
@@ -1,423 +1,82 @@
...
-
-            /// <summary>
-            /// Make sure every time a screen gets pushed, the background doesn't get replaced
-            /// </summary>
-            /// <returns>Whether or not the original background (The one created in DummySongSelect) is still the current background</returns>
-            public bool IsBackgroundCurrent() => ((FadeAccessibleBackground)Background).IsCurrentScreen();
+            AddStep("clear break", () => isBreakTime.Value = false);
+            AddStep("set lighten during break", () => lightenDuringBreaks.Value = true);
+            AddWaitStep("wait for dim", 5);
+            AddAssert($"is current dim {userDim}", () => container.DimEqual(userDim));
```
peppy: your tests don't really cover all scenarios. i'd split it out into four tests or so:

- turn on lighten mode, toggle break on, check, toggle break off, check
- toggle break on, check, turn on lighten mode, check
- turn on lighten mode, toggle break on, check, turn off lighten mode, check, toggle break off, check

or so

iiSaLMaN:
> Can you explain the rationale behind renaming the previous test and putting this in a new test? Generally we name test scenes after the class they are testing.

Previous Test scene seem to be testing components using `UserDimContainer` while the New Test plays with `UserDimContainer` directly. Though maybe we can rename the new test to `TestSceneBreakLightening` and rename the previous test back to `TestSceneUserDimContainer`.

peppy:
Hmmmmmm, fair enough. The existing tests are written in a pretty bad way (hard coupled with player).

Let's stay with how you have it now (although I wouldn't say no to more tests being added/migrated across to your new class while you're here).

peppy:
I made some changes to your tests for readability and reliability – please make sure to skim over them for future contributions.

peppy (approved):

PR #7143, opened 2019-12-10
Move UserDimContainer to an appropriate namespace

frenzibyte:
- Moves `UserDimContainer` to the `osu.Game.Screens.Play` namespace due to the component being out-of-scope to general container as an upcoming PR is adding break time logic inside this container.


peppy:
This is used by `BackgroundScreenBeatmap`, which is more general than `Play`. Not 100% on this one.

peppy:
Leaving this for now.

PR #7157, opened 2019-12-11
Add difficulty adjustment mod

### LINKED ISSUE #3201: Map settings

LeNitrous:
Adds a new mod "Difficulty Adjust" where the user can override a beatmap's difficulty settings.
Resolves #3201.

Requires https://github.com/ppy/osu/pull/7175 for `IApplicableToBeatmapDifficulty` to apply during song selection.

- [x] Default not updating correctly on sliders (depends on https://github.com/ppy/osu-framework/pull/3123 and subsequent logic changes osu-side). Can probably go ahead with this PR without that being fixed as long as it isn't forgotten.
- [x] Depends on https://github.com/ppy/osu/pull/7303 to show up-to-date statistics in song select details.
- [x] Split implementation into per-ruleset.


peppy:
can this feasibly just be settings on HR?

LeNitrous:
Sounds fine.

peppy:
I'd try and make the settings on easy / HR (limited in opposite directions), as multipliers.

LeNitrous:
I just realized that HR in the osu! ruleset alters the positions of the hitobjects and what I'm aiming for is to purely change the beatmap settings alone.

peppy:
fair enough

peppy:
Would it not make more sense to draw defaults from the selected beatmap? I'm not 100% sure how most people would use, but it seems like you'd be adjusting from the mapper's settings rather than always using a set combination.

LeNitrous:
It would certainly make more sense with that.

peppy:
Are you going to give that a shot? Should be a pretty simple change.

LeNitrous:
I will. I'm still skimming over the code to know how to set the slider's default values based on the beatmap since `IApplicableToDifficulty` only applies when we move over to the `Player` screen.

peppy:
I see. Changes may need to happen to have this trigger from song select, on beatmap change.

peppy:
- Default not updating correctly on sliders (depends on https://github.com/ppy/osu-framework/pull/3123 and subsequent logic changes osu-side). Can probably go ahead with this PR without that being fixed as long as it isn't forgotten.
- Display values not updating after mod settings change. This needs to be fixed (as a separate PR) before this is merged.

peppy:
@LeNitrous this needs to be split out into per-ruleset implementations. Mania shouldn't have "circle size" adjustments, for instance.

LeNitrous:
I've taken considerations with Taiko and Mania where CS and AR are unused according to the wiki

For Circle Size:
> This option is visible in osu!taiko, but it does not do anything that affects gameplay.

> Key Count, also known as Keys, replaces the Circle Size metric for osu!mania-specific beatmaps.

For Approach Rate:
> This option is visible in osu!taiko and osu!mania, but it does not do anything that affects gameplay. This is mainly because both use a scroll speed, which is based on BPM (beats per minute).

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public override ModType Type => ModType.Conversion;
+
+        public override IconUsage Icon => FontAwesome.Solid.Hammer;
+
+        public override double ScoreMultiplier => 1.0;
+
+        public override Type[] IncompatibleMods => new[] { typeof(ModEasy), typeof(ModHardRock) };
+
+        public virtual BindableNumber<float> DrainRate { get; }
```
peppy: `DrainRate` is present in all rulesets, so can be in the base class.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public virtual BindableNumber<float> CircleSize { get; }
```
peppy: Move unused members *out* of the base class.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -0,0 +1,64 @@
...
+        public virtual void ApplyToDifficulty(BeatmapDifficulty difficulty)
+        {
+            if (this.difficulty == null || this.difficulty.ID != difficulty.ID)
+            {
+                this.difficulty = difficulty;
+
+                if (DrainRate != null)
+                    DrainRate.Value = DrainRate.Default = difficulty.DrainRate;
+
+                if (CircleSize != null)
```
peppy: do this in the implementation, not base class.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public override Type[] IncompatibleMods => new[] { typeof(ModEasy), typeof(ModHardRock) };
+
+        public virtual BindableNumber<float> DrainRate { get; }
+
+        public virtual BindableNumber<float> CircleSize { get; }
+
+        public virtual BindableNumber<float> ApproachRate { get; }
+
+        public virtual BindableNumber<float> OverallDifficulty { get; }
```
peppy: same for OD. it's in all four rulesets so didn't need to be moved.

peppy (changes requested):

PR #7170, opened 2019-12-12
Fix mod overlay not showing settings on returning from main menu

peppy:
Resolves issue pointed out in https://github.com/ppy/osu/pull/7121#issuecomment-564874764.

Improves tests (previous tests were incorrectly overwriting mods in the mod overlay).


peppy:
Closing in favour of https://github.com/ppy/osu/pull/7172.

PR #3272, opened 2018-08-22
Visualization of change in star difficulty from applied mods

zachllogan:
- Closes #3270

---

Adds the ability for StarCounter to visualize the change in difficulty do to applied mods.
moddifficultystarvisualization

PR #7175, opened 2019-12-12
Display current difficulty statistics with mods applied

LeNitrous:
Allows `BeatmapDetailArea` to display the most current difficulty settings from mods that alter difficulty. This also allows `IApplicableToBeatmapDifficulty` to be called when a new beatmap is selected in `SongSelect`.

No Mod
image
Easy Mod
image
Hard Rock Mod
image
Colors are based off of stable.


peppy:
May be worth looking at https://github.com/ppy/osu/pull/3272 too.

osu.Game/Screens/Select/Details/AdvancedStats.cs
```diff
@@ -65,12 +60,60 @@ public AdvancedStats()
...
+            mods.ValueChanged += _ => updateStatistics();
+        }
+
+        private void updateStatistics()
+        {
+            BeatmapInfo processed = Beatmap?.Clone();
+
+            if (processed != null && mods.Value.Any(m => m is IApplicableToDifficulty))
+            {
+                processed.BaseDifficulty = processed.BaseDifficulty.Clone();
```
peppy: is it necessary to clone a second time?

peppy (changes requested):

peppy:
It would be nice if you could see the actual change in difficulty vs the original. Maybe show the adjusted difficulty in the bottom-half of each line (using the colours you have)?

image

Or use the full height but keep the white and put the colour behind (for harder) or in front (for easier) so you can see the difference

image

Should be considered a temporary solution either way (fine to implement) until we can get a design for this.

osu.Game/Screens/Select/Details/AdvancedStats.cs
LeNitrous: For some reason we need to get a shallow clone of `BeatmapDifficulty`. Certain issues such as the bars not getting colored and the difficulty not showing correct values when re-entering song select will show up.

It is also done in `WorkingBeatmap`:https://github.com/ppy/osu/blob/0627c814d66c34925057fa38a5dc50bd33144d5e/osu.Game/Beatmaps/WorkingBeatmap.cs#L124-L125

LeNitrous:
I've considered using the last suggestion where there is one bar in front with transparency since this is just a temporary solution. I've also updated the value text to change color as well as a quick addition.

The screenshots have been updated to reflect this.

peppy:
I made a couple of fixes to this which I recommend you check out:

- Removed the beatmap cloning – only the difficulty needed to be cloned (for now).
- Used tuple for settings. The way you had it, it would have displayed incorrectly if the Base/Mod value were set back-to-front. By combining into a tuple it not only cleans up the code but ensure a stable state when arriving in the display context.

peppy (dismissed):

smoogipoo (approved):

PR #7183, opened 2019-12-13
Add initial implementation of a legacy beatmap encoder

smoogipoo:
Status:

#### `[General]`
osu!lazer properties only, but mostly complete

#### `[Editor]`
complete

#### `[Metadata]`
complete

#### `[Difficulty]`
complete

#### `[Events]`
beatmap events only, no storyboards

#### `[TimingPoints]`
complete

#### `[HitObjects]`
`Rulesets.Osu` only


osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Beatmaps.Legacy;
+using osu.Game.Rulesets.Objects;
+using osu.Game.Rulesets.Objects.Types;
+
+namespace osu.Game.Beatmaps.Formats
+{
+    public class LegacyBeatmapEncoder
+    {
+        public const int LATEST_VERSION = 128;
```
bdach: 14 was the latest as per the wiki - is this a deliberate choice?

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+            writer.WriteLine(FormattableString.Invariant($"Title: {beatmap.Metadata.Title}"));
+            writer.WriteLine(FormattableString.Invariant($"TitleUnicode: {beatmap.Metadata.TitleUnicode}"));
+            writer.WriteLine(FormattableString.Invariant($"Artist: {beatmap.Metadata.Artist}"));
+            writer.WriteLine(FormattableString.Invariant($"ArtistUnicode: {beatmap.Metadata.ArtistUnicode}"));
+            writer.WriteLine(FormattableString.Invariant($"Creator: {beatmap.Metadata.AuthorString}"));
+            writer.WriteLine(FormattableString.Invariant($"Version: {beatmap.BeatmapInfo.Version}"));
+            writer.WriteLine(FormattableString.Invariant($"Source: {beatmap.Metadata.Source}"));
+            writer.WriteLine(FormattableString.Invariant($"Tags: {beatmap.Metadata.Tags}"));
+            writer.WriteLine(FormattableString.Invariant($"BeatmapID: {beatmap.BeatmapInfo.OnlineBeatmapID ?? 0}"));
+            writer.WriteLine(FormattableString.Invariant($"BeatmapSetID: {beatmap.BeatmapInfo.BeatmapSet.OnlineBeatmapSetID ?? 0}"));
```
bdach: Not sure about stable interop, but on the eye test this doesn't seem to preserve nulls (encoding and decoding would as far I can tell silently coalesce these `null`s to 0).

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+
+                    case 2:
+                        handleCatchHitObject(writer, h);
+                        break;
+
+                    case 3:
+                        handleManiaHitObject(writer, h);
+                        break;
+                }
+            }
```
bdach: This reads weird to me with the switch inside of the loop instead of the opposite. It'll most likely have no effect on anything, I'm guessing not even performance (due to branch prediction), but it pops out as a strange construct to me.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+
+            if (hitObject is IHasCurve curveData)
+            {
+                PathType? lastType = null;
+
+                for (int i = 0; i < curveData.Path.ControlPoints.Count; i++)
+                {
+                    PathControlPoint point = curveData.Path.ControlPoints[i];
+
+                    if (point.Type.Value != null)
```
bdach: I'm not sure how you feel about `continue`s, but this seems like a good place to put one and invert the check to get the path out of the way.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+        {
+        }
+
+        private void handleCatchHitObject(TextWriter writer, HitObject hitObject)
+        {
+        }
+
+        private void handleManiaHitObject(TextWriter writer, HitObject hitObject)
+        {
+        }
```
bdach: Could at least throw `NotImplemented` for now maybe? Unless you're aiming for fast integration with standard editor in which case I guess not throwing is slightly more preferable.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+
+        private string getSampleBank(IList<HitSampleInfo> samples, bool banksOnly = false, bool zeroBanks = false)
+        {
+            LegacySampleBank normalBank = toLegacySampleBank(samples.FirstOrDefault(s => s.Name == HitSampleInfo.HIT_NORMAL)?.Bank);
+            LegacySampleBank addBank = toLegacySampleBank(samples.FirstOrDefault(s => !string.IsNullOrEmpty(s.Name) && s.Name != HitSampleInfo.HIT_NORMAL)?.Bank);
+
+            string customSampleBank = toLegacyCustomSampleBank(samples.FirstOrDefault()?.Suffix);
+            string sampleFilename = samples.FirstOrDefault(s => string.IsNullOrEmpty(s.Name))?.LookupNames.First() ?? string.Empty;
+
+            int volume = samples.First().Volume;
```
bdach: Maybe these should be `Single(OrDefault)`s?

bdach (commented):
Some things from a read-through, most of which aren't probably all that valid anyways.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
mcendu: #7116 introduced sliders with multiple segments. As each segment can have a different curve type, maps with such kind of pathes might be incompatible with 14.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+
+        private void handleGeneral(TextWriter writer)
+        {
+            writer.WriteLine("[General]");
+
+            writer.WriteLine(FormattableString.Invariant($"AudioFilename: {Path.GetFileName(beatmap.Metadata.AudioFile)}"));
+            writer.WriteLine(FormattableString.Invariant($"AudioLeadIn: {beatmap.BeatmapInfo.AudioLeadIn}"));
+            writer.WriteLine(FormattableString.Invariant($"PreviewTime: {beatmap.Metadata.PreviewTime}"));
+            // Todo: Not all countdown types are supported by lazer yet
+            writer.WriteLine(FormattableString.Invariant($"Countdown: {(beatmap.BeatmapInfo.Countdown ? "1" : "0")}"));
```
huoyaoyuan: `'1' ? '0'` for slightly better performance

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+                        break;
+                }
+            }
+
+            return type;
+        }
+
+        private LegacySampleBank toLegacySampleBank(string sampleBank)
+        {
+            switch (sampleBank?.ToLower())
```
huoyaoyuan:
```suggestion
            switch (sampleBank?.ToLowerInvariant())
```

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -0,0 +1,399 @@
...
+
+                writer.Write(FormattableString.Invariant($"{group.Time},"));
+                writer.Write(FormattableString.Invariant($"{beatLength},"));
+                writer.Write(FormattableString.Invariant($"{(int)beatmap.ControlPointInfo.TimingPointAt(group.Time).TimeSignature},"));
+                writer.Write(FormattableString.Invariant($"{(int)toLegacySampleBank(tempHitSample.Bank)},"));
+                writer.Write(FormattableString.Invariant($"{toLegacyCustomSampleBank(tempHitSample.Suffix)},"));
+                writer.Write(FormattableString.Invariant($"{tempHitSample.Volume},"));
+                writer.Write(FormattableString.Invariant($"{(timingPoint != null ? "1" : "0")},"));
+                writer.Write(FormattableString.Invariant($"{(int)effectFlags}"));
+                writer.Write("\n");
```
huoyaoyuan:
```suggestion
                writer.WriteLine();
```

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
peppy: intentionally leaving some headroom in case we need to add more stable version increments (without conflicting with lazer).

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
smoogipoo: Coalescing is fine, although OnlineBeatmapSetID should coalesce to -1 instead.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
smoogipoo: I did it this way because of the `if (i != 0)` check below. Unless I'm missing something, I'll either have to dupe that block into the `continue`ing condition (which seems weird due to the code duping), or add a try-finally (which isn't saving much).

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
smoogipoo: I've made the change where possible, although this is not the case for every `First*()` in this method.

peppy (dismissed):

PR #7192, opened 2019-12-13
Implement NewsArticleCover class

Game4all:
This PR implements the `NewsArticleCover` class.
Note that I'm using a dummy struct (see `ArticleInfo` class in`NewsArticleCover` ) as there is no API struct for news article info for now

Visual reference | This PR
----------------|--------
70392687-f6814300-1a25-11ea-9453-74008d2f0fc9 | image |

### DIFF
--- a/osu.Game/Overlays/News/NewsArticleCover.cs
+++ b/osu.Game/Overlays/News/NewsArticleCover.cs
@@ -90,7 +90,6 @@ namespace osu.Game.Overlays.News
             bg.OnLoadComplete += d => d.FadeIn(250, Easing.In);
         }
 
-        //news article cover background
         [LongRunningLoad]
         private class NewsBackground : Sprite
         {
@@ -108,7 +107,6 @@ namespace osu.Game.Overlays.News
             }
         }
 
-        //date container
         private class DateContainer : Container, IHasTooltip
         {
             private readonly DateTime date;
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/News/NewsArticleCover.cs
```diff
@@ -0,0 +1,159 @@
...
+                    },
+                    Font = OsuFont.GetFont(Typeface.Exo, 16, FontWeight.Bold),
+                    Text = "by " + info.Author
+                }
+            };
+
+            bg.OnLoadComplete += d => d.FadeIn(250, Easing.In);
+        }
+
+        //news article cover background
```
bdach: Kind of a redundant comment (as is the next one)

osu.Game/Overlays/News/NewsArticleCover.cs
```diff
@@ -0,0 +1,159 @@
...
+                    {
+                        RelativeSizeAxes = Axes.Both,
+                        Colour = Color4.Black.Opacity(0.5f),
+                    },
+                    new OsuSpriteText
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Font = OsuFont.GetFont(Typeface.Exo, 12, FontWeight.Black, false, false),
+                        Text = date.ToString("dd MMM yyy"),
```
bdach: Don't know if we're after parity with web here, but if we are then this should probably be
```suggestion
                        Text = date.ToString("d MMM yyy").ToUpper(),
```

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -59,6 +59,21 @@ private void load(OsuColour colours)
             Current.TriggerChange();
         }
 
+        private CancellationTokenSource loadChildCancellation;
+
+        protected void LoadAndShowChild(NewsContent newContent)
```
peppy: content, not child

peppy (approved):
seems fine for now

PR #7196, opened 2019-12-13
KeyBindingRow: only consider buttons that were pressed after getting focus

galister:
This aims to fix an issue with key binding while a joystick is connected:
When trying to bind a key, Osu will record any non-centered axes and non-momentary switches into the key combination.

With this PR, such will be ignored unless activated while the KeyBindingRow is being focused.

### DIFF
--- a/osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
+++ b/osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
@@ -12,6 +13,7 @@ using osu.Framework.Graphics.Effects;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input.Bindings;
 using osu.Framework.Input.Events;
+using osu.Framework.Input.States;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Sprites;
 using osu.Game.Graphics.UserInterface;
@@ -150,6 +152,9 @@ namespace osu.Game.Overlays.KeyBinding
 
         private KeyButton bindTarget;
 
+        private ICollection<InputKey> previouslyHeldKeys;
+        private bool canHandleRelease;
+
         public bool AllowMainMouseButtons;
 
         public IEnumerable<KeyCombination> Defaults;
@@ -173,7 +178,8 @@ namespace osu.Game.Overlays.KeyBinding
                 }
             }
 
-            bindTarget.UpdateKeyCombination(KeyCombination.FromInputState(e.CurrentState));
+            bindTarget.UpdateKeyCombination(keyCombinationSinceFocused(e.CurrentState));
+            canHandleRelease = true;
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
```diff
@@ -173,7 +178,8 @@ protected override bool OnMouseDown(MouseDownEvent e)
                 }
             }
 
-            bindTarget.UpdateKeyCombination(KeyCombination.FromInputState(e.CurrentState));
+            bindTarget.UpdateKeyCombination(keyCombinationSinceFocused(e.CurrentState));
```
bdach: Even though it's one line, I'd probably make a method out of this usage pattern myself, as it's not immediately obvious `keyCombinationSinceFocused` should be used in all places.

osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
```diff
@@ -150,6 +152,9 @@ protected override void OnHoverLost(HoverLostEvent e)
 
         private KeyButton bindTarget;
 
+        private ICollection<InputKey> previouslyHeldKeys;
+        private bool canHandleRelease;
```
bdach: If I'm not mistaken, this flag is not necessary provided we save the state of buttons at the moment the binding component becomes focused.

* If a button went up and it was in `previouslyHeldKeys` before, then exclude it from `previouslyHeldKeys` and do not finalize binding input.
* If a button went up and it wasn't in `previouslyHeldKeys` before, then finalise.

I might be wrong though because this logic is quite hard to keep in my head.

bdach (commented):
Tests for this would be nice, and I don't think they'd be too bad to write with a `ManualInputManager`.

osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
galister: the current implementation takes the KeyCombination.FromInputState() at the moment a non-modified button is pressed and saves that as the binding. the issue with that is that it'll record things such as axes being off-center and switches (like actual 2-position switches), some of which don't even have an off-position say, down position activates button 30 while up position activates button 31. high-end devices like to do this and even though I'm not planning to use these to play Osu, this would make me have to unplug my devices to make it work

bdach: I understand all that, I thought I was proposing a simpler implementation of the same behaviour, and I'm still not entirely sure the two differ (as in I don't see a concrete case where the two would differ). I'll have a think about it with a fresher mind tomorrow - but honestly I'd like to see tests for the scenarios you describe regardless.

osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
galister: already have a method, might as well have this cover this line as well.

osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
galister: I started out with the same thing you've said, however I had to work around the edge case of having a button that's being held before the container is focused, releasing and pushing it again would register an empty KeyCombination.

I'll  get some tests going.

peppy:
Is this still an issue? I've tried testing with key / key repeat but it doesn't seem to immediately save. I would hope that the axis-sourced button presses only fire once on press / release (seems to be the case from checking the framework flow), did this change at some point in the last year?

galister:
This happens when I press the Up button while a joystick (that's not completely centered) is plugged in. Note the joystick is not being moved at all, and the binding is accepted when I press the Up button on the keyboard.

image

peppy:
I guess it's due to jitter in and out of the dead zone? Although in that case I'm not sure how this will work to fix the issue.

As I said above, joysticks off-centre should only fire events once as far as I can tell from the code and from testing.

peppy:

In particular, this needs test coverage to find out why the last comment I left in this PR is not what's happening. In theory this seems like a change we *will* want at some point, but the fix may not be in the correct place. Understanding the base behaviour is required before attempting to fix it.

PR #7208, opened 2019-12-14
Update sdk and InspectCode version

huoyaoyuan:
This will unblock switch expressions, default interface implementations, and some other new features.

Notes for sdk version in `global.json`:
Previously, `global.json` doesn't support rolling forward, thus can only be used to lock the sdk version. From .NET Core 3, it can be used to specify _minimal required SDK version_ now. Reference: https://github.com/dotnet/designs/blob/master/accepted/global-json-updates.md
My opinions for setting rolling level:
- Prefer lower to higher. This helps finding usages requiring SDK version update
- Only rolling feature brands seems too strict. Will major rolling bring unpredictable changes? Minor should be a medium level.

Including index and range expressions in this PR as they are small changes.

NOT updating the target framework to 3.1. Should we do this now? Updating target framework will be much easier in .NET Core.

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneBeatmapSetOverlayDetails.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneBeatmapSetOverlayDetails.cs
@@ -44,7 +44,7 @@ namespace osu.Game.Tests.Visual.Online
             AddStep("set second set", () => details.BeatmapSet = secondSet);
             AddAssert("ratings set", () => details.Ratings.Metrics == secondSet.Metrics);
 
-            BeatmapSetInfo createSet() => new BeatmapSetInfo
+            static BeatmapSetInfo createSet() => new BeatmapSetInfo
             {
                 Metrics = new BeatmapSetMetrics { Ratings = Enumerable.Range(0, 11).Select(_ => RNG.Next(10)).ToArray() },
                 Beatmaps = new List<BeatmapInfo>
--- a/osu.Game/Graphics/Containers/LinkFlowContainer.cs
+++ b/osu.Game/Graphics/Containers/LinkFlowContainer.cs
@@ -37,7 +37,7 @@ namespace osu.Game.Graphics.Containers
 
             foreach (var link in links)
             {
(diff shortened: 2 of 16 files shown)


osu.Game.Tests/Visual/Online/TestSceneBeatmapSetOverlayDetails.cs
```diff
@@ -44,7 +44,7 @@ public void TestMetrics()
             AddStep("set second set", () => details.BeatmapSet = secondSet);
             AddAssert("ratings set", () => details.Ratings.Metrics == secondSet.Metrics);
 
-            BeatmapSetInfo createSet() => new BeatmapSetInfo
+            static BeatmapSetInfo createSet() => new BeatmapSetInfo
```
huoyaoyuan: This is a fatal bug in VS16.3/roslyn3.3 and fixed now. Compiler is now bound to SDK version, forcing SDK 3.1.100 is required for this fix.

huoyaoyuan:
It's a bit strange that my local InspectCode run catches some of index usages, but not all. Maybe due to cache issues.

smoogipoo:
Welcome to JetBrains _still_ only having a primitive support of C#8.

Generally I’m ok with these changes. Just gonna ping @peppy to make sure we’re on the same page.

smoogipoo (approved):

osu.Game/Graphics/Containers/LinkFlowContainer.cs
```diff
@@ -37,7 +37,7 @@ public void AddLinks(string text, List<Link> links)
 
             foreach (var link in links)
             {
-                AddText(text.Substring(previousLinkEnd, link.Index - previousLinkEnd));
+                AddText(text[previousLinkEnd..link.Index]);
```
peppy: am i alone in thinking this reads really badly?

smoogipoo: It's much safer this way, though. Very easy to get into off-by-one traps with index+length combos.

peppy: ok, let's go with it for now. looks alright with font ligatures.

PR #7222, opened 2019-12-15
Add a reverse setting for Winding mods

### LINKED ISSUE #4863: reverse of the winding mods
**Describe the new feature:** 
it is the winding mods but it starts at the fastest/slowest speed and it goes to normal speed over the length of the beatmap
Untitled

LeNitrous:
Closes https://github.com/ppy/osu/issues/4863.

- [x] Depends on #7645
- [x] Depends on ppy/osu-web#5557
- [x] Depends on web deploy

Simply adds a new setting for `ModTimeRamp`.


peppy:
If we're going to do this "WindUp" and "WindDown" need to be re-thought. Since the names are no longer correct.

LeNitrous:
Making it a separate mod as the issue originally thought off may kinda be too much. Renaming the mod would be better.

peppy:
In discussion, we don't think this is a sane direction forward (the mod names lose their meaning). `WindDown` and `WindUp` are good presets so we don't really want to combine them into one mod.

Rather than this, the proposal is to:

- Expose a "start" value for ramping mods.
- For WindUp, the start value should be default 1 but range 0.5-1.
- For WindDown, the start value should be default 1 but range 1-1.5

Wanna give that a shot instead and see how it feels?

LeNitrous:
So in a nutshell, making the whole ramp customizable. For starters, it sounds off from the original proposal wherein the original goal was to flip the start and end times of the ramp.

Regardless, it still sounds an interesting setting to include for the time ramp mods. I'll see what I can do with it as well.

peppy:
Customisable, but still ramping in the direction of the mod name, never reverse.

It allows what you have added here (ie. WindDown with start value 1.5, end value 0.99 would be the same as WindUp + Reverse).

LeNitrous:
For some reason the tooltip in `ModWindUp`'s `InitialRate` slider is in percent. I'll look into it.

osu.Game/Rulesets/Mods/ModTimeRamp.cs
```diff
@@ -69,6 +72,7 @@ public virtual void Update(Playfield playfield)
         /// </summary>
         /// <param name="amount">The amount of adjustment to apply (from 0..1).</param>
         private void applyAdjustment(double amount) =>
-            SpeedChange.Value = 1 + (FinalRate.Value - 1) * Math.Clamp(amount, 0, 1);
+            //  SpeedChange.Value = 1 + (FinalRate.Value - 1) * Math.Clamp(amount, 0, 1);
```
bdach: Commented code shouldn't stay in if no longer needed.

bdach (dismissed):
The percents are due to you hitting this check:

https://github.com/ppy/osu/blob/2ef8714d6713730bcb33f665842aa3aaeca94fff/osu.Game/Graphics/UserInterface/OsuSliderBar.cs#L175-L176

which is kind of a pain. Percent display should be explicitly requested IMO, but that requires going over all usages and introducing some sort of flag manually which is probably worthy of a seperate PR.

osu.Game/Rulesets/Mods/ModTimeRamp.cs
LeNitrous: whoops my bad

smoogipoo (approved):

smoogipoo:
For now we'll go with displaying as plain-text decimals. We'll want some way to format the text in the future (e.g. these should be `1.53x` rather than `1.53` or `153%`).
Otherwise lgtm.

smoogipoo:
Blocking on osu-web changes.

PR #7223, opened 2019-12-15
Deassociate LegacyLastTick from slider tail

mcendu:
Cases that mean to use the real tail instead of `LegacyLastTick` remain.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Slider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Slider.cs
@@ -107,7 +107,7 @@ namespace osu.Game.Rulesets.Osu.Objects
         public double TickDistanceMultiplier = 1;
 
         public HitCircle HeadCircle;
-        public SliderTailCircle TailCircle;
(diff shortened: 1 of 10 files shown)


osu.Game.Rulesets.Osu/Objects/Slider.cs
```diff
@@ -162,9 +162,16 @@ protected override void CreateNestedHitObjects()
...
-                        // it is *okay* to use this because the TailCircle is not used for any meaningful purpose in gameplay.
-                        // if this is to change, we should revisit this.
-                        AddNested(TailCircle = new SliderTailCircle(this)
+                        AddNested(LastTick = new SliderLegacyLastTick(this)
+                        {
+                            StartTime = e.Time,
+                            Position = EndPosition,
+                            StackHeight = StackHeight
+                        });
+                        break;
```
mcendu: Seems like this addition breaks tests.

mcendu:
Unnecessary

PR #7227, opened 2019-12-16
Add notification support for private messages and username mentions

Craftplacer:
With this pull-request I added the following features:
- Notifications upon user mention
- Notifications upon incoming private message

- [x] Depends on #7425


osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -220,5 +327,53 @@ public DaySeparator(DateTimeOffset time)
                 };
             }
         }
+
+        private class MentionNotification : SimpleNotification
```
peppy: should be two separate classes. a "chanel" isn't a "username" and you're conditional-ing every line of the class.

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -134,6 +167,62 @@ void expireAndAdjustScroll(Drawable d)
                 scrollToEnd();
         }
 
+        private void checkForMentions(IEnumerable<Message> messages)
+        {
+            // only send notifications when chat overlay is **closed**
```
peppy: comment doesn't match the code.

osu.Game/Overlays/Settings/Sections/Online/AlertsAndPrivacySettings.cs
```diff
@@ -0,0 +1,32 @@
...
+            Children = new Drawable[]
+            {
+                new SettingsCheckbox
+                {
+                    LabelText = "Show a notification popup when someone says your name",
+                    Bindable = config.GetBindable<bool>(OsuSetting.ChatHighlightName)
+                },
+                new SettingsCheckbox
+                {
+                    LabelText = "Show chat message notifications",
```
peppy: isn't it more like "private message notifications"?

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -148,6 +237,13 @@ private void pendingMessageResolved(Message existing, Message updated)
             }
         }
 
+        public void HighlightMessage(Message message)
```
peppy: `ScrollToAndHighlightMessage`

osu.Game/Overlays/Chat/DrawableChannel.cs
```diff
@@ -148,6 +237,13 @@ private void pendingMessageResolved(Message existing, Message updated)
...
+        {
+            var chatLine = findChatLine(message);
```
peppy: handle null case. it may no longer be in buffer.

peppy (changes requested):

osu.Game/Overlays/Settings/Sections/Online/AlertsAndPrivacySettings.cs
Craftplacer: Taken from stable. I'll be happy to change that.

peppy:
First of all, a lot of this functionality is going to be moved to the API-side, where notifications are already created / served. So at best, this would be a temporary implementation. As such it should be easy to remove and contained in its own class.

You've also added a lot of complexity and responsibilities to `DrawableChannel` here. I'd consider moving all the highlight/notification related code into its own `Component`.

It could either live inside ChannelManager or ChatOverlay (probably the first to allow for notifications to be decoupled from the drawable channel / overlay stack) and DI in `ChatOverlay` as a nullable component for the places you need access.

If you want this functionality to live on, it would need to be written against osu-web, and the notification API used to source notifications into the game.

Craftplacer:
I will look into separating and moving later. But how is the web side of things going to work?

Craftplacer:
I still have to add MessageNotifier to channel events.
Also, should I force create channel drawable creation/joining of channels, so the notification feature actually works?

peppy:
Why do you have to force drawable creation for it to work? The idea is that `ChannelManager` can work without drawables.

Craftplacer:
> Why do you have to force drawable creation for it to work? The idea is that `ChannelManager` can work without drawables.

It doesn't have to be that but I was assuming in the comment what I'd have to do, since the pre-review code relied on a created ChannelDrawable.

osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -221,8 +221,14 @@ private void load(OsuConfigManager config, OsuColour colours, ChannelManager cha
...
                 channelManager.JoinedChannels.ItemsRemoved += onChannelRemovedFromJoinedChannels;
+
+                bool channelSelected = channelManager.CurrentChannel.Value != null;
+
                 foreach (Channel channel in channelManager.JoinedChannels)
-                    ChannelTabControl.AddChannel(channel);
+                    ChannelTabControl.AddChannel(channel, !channelSelected);
+
+                if (channelSelected)
+                    ChannelTabControl.Current.Value = channelManager.CurrentChannel.Value;
```
bdach: This logic is quite convoluted. You're handling the "no channel selected" case in `ChannelTabControl` and the "channel selected" case here, which is going to be confusing to whoever is going to be looking for the place where the channel can change down the line. Should handle both here and make `AddChannel` not change tabs if anything, IMO.

bdach (changes requested):
* This should have tests in general. Right now you need elaborate setup to even be able to tell if this code will work.
* From a UX standpoint, the "Chat highlight words" setting looks unusable in its current state. If you didn't know there was supposed to be a textbox there, you wouldn't be able to use it - looks like an empty space.

  image

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,257 @@
...
+                }
+
+                if (!string.IsNullOrWhiteSpace(highlightWords.Value))
+                {
+                    var matchedWord = hasCaseInsensitive(words, getWords(highlightWords.Value));
+
+                    if (matchedWord != null)
+                    {
+                        var notification = new HighlightNotification(message.Sender.Username, matchedWord, onClick);
+                        notificationOverlay?.Post(notification);
```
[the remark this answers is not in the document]
Craftplacer: Is the name "HighlightedWordNotification" more appropiate?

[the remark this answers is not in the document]
bdach: The entire name of the setting is misleading to me with this implementation. Highlighting implies a visual effect. If we're going to stick with firing notifications on keywords, that's what the setting should be called (so something like "Send notifications when the following words are used in chat")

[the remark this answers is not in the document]
Craftplacer: I intended at first to implement highlighting as well, but peppy suggested me to handle this in a separate pull request. You could tell me if I should implement this later or rename the names now.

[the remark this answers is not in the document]
bdach: In that case, IMO the highlighted word setting and all code involved should be entirely extracted out of this PR and moved into a separate one, leaving just the mention and PM notifications here.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,257 @@
...
+            [BackgroundDependencyLoader]
+            private void load(OsuColour colours, MessageNotifier notifier)
+            {
+                IconBackgound.Colour = colours.PurpleDark;
+                Activated = delegate
+                {
+                    onClick?.Invoke();
+
+                    if (notifier.privateMessageNotifications.Contains(this))
+                        notifier.privateMessageNotifications.Remove(this);
```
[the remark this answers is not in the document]
Craftplacer: I tried to make NotificationOverlay expose its notifications  so I can manipulate them later, that didn't work and a list containing all instances of the notification was working. To be sure that we don't change notifications that aren't even inside the NotificationOverlay anymore I made it to remove itself. I would be glad if you could guide on how to change this.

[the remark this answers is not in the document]
iiSaLMaN: Shouldn't you just include these 2 lines inside the `onClick` action.

osu.Game/Overlays/ChatOverlay.cs
Craftplacer: Are these changes fine?
```diff
@ -222,13 +222,10 @@ private void load(OsuConfigManager config, OsuColour colours, ChannelManager cha
                channelManager.JoinedChannels.ItemsAdded += onChannelAddedToJoinedChannels;
                channelManager.JoinedChannels.ItemsRemoved += onChannelRemovedFromJoinedChannels;

-               bool channelSelected = channelManager.CurrentChannel.Value != null;

                foreach (Channel channel in channelManager.JoinedChannels)
-                   ChannelTabControl.AddChannel(channel, !channelSelected);
+                   ChannelTabControl.AddChannel(channel);

-               if (channelSelected)
-                   ChannelTabControl.Current.Value = channelManager.CurrentChannel.Value;
+               ChannelTabControl.Current.Value = channelManager.CurrentChannel.Value == null ? channelManager.JoinedChannels.First() : channelManager.CurrentChannel.Value;

                channelManager.AvailableChannels.ItemsAdded += availableChannelsChanged;
                channelManager.AvailableChannels.ItemsRemoved += availableChannelsChanged;
```

```diff
@ -59,17 +59,12 @@ protected override TabItem<Channel> CreateTabItem(Channel value)

        /// <summary>
        /// Adds a channel to the ChannelTabControl.
-       /// The first channel added will automaticly selected if <paramref name="setChannel"/> is true.
        /// </summary>
        /// <param name="channel">The channel that is going to be added.</param>
-       /// <param name="setChannel">If the current channel should be changed if none was selected before</param>
-       public void AddChannel(Channel channel, bool setChannel = true)
+       public void AddChannel(Channel channel)
        {
            if (!Items.Contains(channel))
                AddItem(channel);
-
-           if (Current.Value == null && setChannel)
-               Current.Value = channel;
        }

        /// <summary>
```

osu.Game/Online/Chat/MessageNotifier.cs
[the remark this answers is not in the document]
Craftplacer: > Shouldn't you just include these 2 lines inside the `onClick` action.

Already did that with the suggestion of un-god-ing the HandleMessages method.

osu.Game/Overlays/ChatOverlay.cs
bdach: That is what I had in mind, so it looks good to me, yes.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,257 @@
...
+                MessageCount = 1;
+                this.onClick = onClick;
+            }
+
+            private int messageCount;
+
+            public int MessageCount
+            {
+                get => messageCount;
+                set => Text = (messageCount = value) > 1 ? $"You received {messageCount} private messages from '{Username}'. Click to read it!" : $"You received a private message from '{Username}'. Click to read it!";
```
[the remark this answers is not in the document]
Craftplacer: I wouldn't benefit from Humanizer because it still requires me to put a condition in place (e.g. `messageCount > 1 ? "message".Pluralize() : "message"`).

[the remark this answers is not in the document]
Craftplacer: Nevermind. https://github.com/Humanizr/Humanizer#toquantity

osu.Game/Online/Chat/MessageNotifier.cs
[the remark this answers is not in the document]
Craftplacer: @bdach any thoughts?

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            HandleMessages(messages.First().ChannelId, messages);
+        }
+
+        /// <remarks>
+        /// Resolves the channel id
+        /// </remarks>
+        public void HandleMessages(long channelId, IEnumerable<Message> messages)
+        {
+            var channel = channelManager.JoinedChannels.FirstOrDefault(c => c.Id == channelId);
```
bdach: Can probably be
```suggestion
            var channel = channelManager.JoinedChannels.SingleOrDefault(c => c.Id == channelId);
```

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            if (channel == null)
+            {
+                Logger.Log($"Couldn't resolve channel id {channelId}", LoggingTarget.Information);
+                return;
+            }
+
+            HandleMessages(channel, messages);
```
bdach: Not sure if _this_ split makes much sense now.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+        }
+
+        public void HandleMessages(Channel channel, IEnumerable<Message> messages)
+        {
+            // don't show if the ChatOverlay and the target channel is visible.
+            if (IsActive && channelManager.CurrentChannel.Value == channel)
+                return;
```
bdach: Not sure if I'm sleep deprived, but this comment threw me in for a loop. I had to think long and hard about "don't show _what_?".

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            foreach (var message in messages)
+            {
+                // ignore messages that already have been read
+                if (message.Id < channel.LastReadId)
+                    return;
```
bdach: Supposed to be a `continue`, or am I wrong? Assuming we're checking messages chronologically, oldest to newest, hitting one old read message will make it so that no notifications are ever fired?

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                // ignore messages from yourself
+                var localUsername = localUser.Value.Username;
+
+                if (message.Sender.Username == localUsername)
+                    continue;
```
bdach: Another confusing comment. Should either remove the blank line or move it down if you insist or having it. I'd just remove it as it restates the code. 

Checking by username is also probably a worse idea than checking the ID, which you should have access to here.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+                if (checkForPMs(channel, message))
+                    continue;
+
+                // change output to bool again if another "message processor" is added.
+                checkForMentions(channel, message, localUsername);
```
bdach: This seems plain wrong. Suppose you're in a normal channel (non-PM). `checkForPMs` will always return `false`, therefore making mentions not work at all?

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+            var words = getWords(message.Content);
+
+            if (!notifyOnMention.Value || !anyCaseInsensitive(words, username))
+                return;
+
+            var notification = new MentionNotification(message.Sender.Username, channel);
+            notificationOverlay?.Post(notification);
+        }
+
+        private static IEnumerable<string> getWords(string input) => Regex.Matches(input, @"\w+").Select(c => c.Value);
```
bdach: Looks like a regex you might want to compile and reuse.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+            HandleMessages(channel, messages);
+        }
+
+        public void HandleMessages(Channel channel, IEnumerable<Message> messages)
+        {
+            // don't show if the ChatOverlay and the target channel is visible.
+            if (IsActive && channelManager.CurrentChannel.Value == channel)
+                return;
+
+            foreach (var message in messages)
```
bdach: After reading this loop it appears clear to me that this structure isn't working for this usecase, there's too much branching. This would benefit in inversion of control via a message processor abstraction. Consider the following:

* Create N message processors (one for mentions, one for PMs, ...)
* Processors are applicable to certain channel types (mentions - global, PMs - private channels only)
* Processors take in messages and do whatever they need to with them

This yields:

* Less branching
* Better separation of concerns
* They're testable!
* You could even hide the lists of notifications sent lower down to the processors making them a little less painful

I haven't tried to implement, but that's the way I'd try to go with this code. At current it is really difficult to keep track of all of this.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,234 @@
...
+
+            return true;
+        }
+
+        private void checkForMentions(Channel channel, Message message, string username)
+        {
+            var words = getWords(message.Content);
+
+            if (!notifyOnMention.Value || !anyCaseInsensitive(words, username))
+                return;
```
bdach: There's some performance to be gained here; why even get words if you don't even have the notifications enabled?

osu.Game/Online/Chat/MessageNotifier.cs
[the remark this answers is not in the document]
bdach: I suppose if it's in one central place instead of present in one and absent in two others I can temporarily live with it, but it'd be better to have it gone.

bdach (changes requested):
After the god method split I have serious doubts about the current structure of this code. It's really hard to follow and I think it has at least one severe bug.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: This is intended, it doesn't make sense to check (and send notifications of) messages of a channel the user already has read.

bdach: I understand not showing notifications for the already-read messages, but either (a) my brain is no longer working or (b) since this returns from a loop over all messages, in the scenario of
```
read message 123
read message 124
unread message 125 with a mention
```
this code will see message 123, and since it's read it will no longer check the rest, therefore not firing a notification for message 125 that it should. Unless the messages are always read *newest-first*, in which case I suppose this is OK and optimal, but warrants a comment.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: What does "compile" mean in the context of a regex?

osu.Game/Online/Chat/MessageNotifier.cs
iiSaLMaN: I think what would be better here is to have a method for retrieving the channel for an ID.

osu.Game/Online/Chat/MessageNotifier.cs
bdach: See the [`RegexOptions.Compiled`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regexoptions?view=netframework-4.8#System_Text_RegularExpressions_RegexOptions_Compiled) flag, and [some more info here](https://docs.microsoft.com/en-us/dotnet/standard/base-types/regular-expression-options?view=netframework-4.8#compiled-regular-expressions). Should probably be worth it over multiple executions. Can always benchmark to be sure.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: The problem is, even if I would take this effort to make this all objects with properties and what not. This might or will get deprecated once web notifications get integrated into the game (like peppy warned me before hand), **unless** we have a place/use case for such system.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: The wording might be off, I could phrase it "// Only show if the ChatOverlay and the target channel aren't visible"

bdach: Sure, I was mostly missing the word "notifications".

osu.Game/Online/Chat/MessageNotifier.cs
bdach: While you're correct, I don't think this is in a mergeable state at current, even if this PR is temporary. However I'm not going to insist - there is already improvement on the original implementation as far as I'm concerned - I'll defer to the judgement of others.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: > but mentions still won't work in PM channels, which they probably should do.

It's like mentioning your friend in an one to one DM in Discord. It doesn't make much sense (in my opinion), sending two notifications about the same message with the same destination (your friends PMs).

bdach: Don't necessarily disagree with your conclusion either. The most painful thing about this to me is that the order of checking now silently matters and will generate regressions / unexpected behaviour if another conditional is improperly inserted. The processor proposal would help isolate concerns here.

Craftplacer: Is it fine if I include a warning comment, for the lifetime of this pull request living inside lazer?

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: I've inversed/sorted the incoming messages by their time/ID, thanks for making me take notice of this.

osu.Game/Online/Chat/MessageNotifier.cs
bdach: A comment wouldn't hurt if this is to stay as is. You could also describe the rationale behind this particular ordering (why PMs first, etc.)

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,238 @@
...
+        private ChannelManager channelManager { get; set; }
+
+        private Bindable<bool> notifyOnMention;
+        private Bindable<bool> notifyOnChat;
+        private Bindable<User> localUser;
+
+        /// <summary>
+        /// Determines if the user is able to see incoming messages.
+        /// </summary>
+        public bool IsActive => chatOverlay?.IsPresent == true;
```
peppy: bad name, also shouldn't be public. also is only used in one place so it might be fine to just inline it.

also the condition you want is probably `chatOverlay?.State == Visible`

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,238 @@
...
+                return false;
+
+            var existingNotification = privateMessageNotifications.FirstOrDefault(n => n.Username == message.Sender.Username);
+
+            if (existingNotification == null)
+            {
+                var notification = new PrivateMessageNotification(message.Sender.Username, channel, n => privateMessageNotifications.Remove(n));
+
+                notificationOverlay?.Post(notification);
+                privateMessageNotifications.Add(notification);
```
peppy: this is never emptied and will grow infinitely. as discussed in discord, i would not add this debounce functionality in this PR. there's already too much; debouncing notifications is something which requires further thought.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,238 @@
...
+        private void checkForMentions(Channel channel, Message message, string username)
+        {
+            if (!notifyOnMention.Value || !anyCaseInsensitive(getWords(message.Content), username))
+                return;
+
+            var notification = new MentionNotification(message.Sender.Username, channel);
+            notificationOverlay?.Post(notification);
+        }
+
+        private static IEnumerable<string> getWords(string input) => Regex.Matches(input, @"\w+", RegexOptions.Compiled).Select(c => c.Value);
```
peppy: I'm pretty sure you do not want to use the `Compiled` option with a `static` invocation.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            private int messageCount;
+
+            public int MessageCount
+            {
+                get => messageCount;
+                set
+                {
+                    messageCount = value;
+                    Text = $"You received {"private message".ToQuantity(messageCount)} from '{Username}'. Click to read it!";
```
peppy: number is not required.

peppy (changes requested):

peppy:
No test scene? How are you testing this?

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: Why is that so?

peppy: Due to the other comment above (don't do debounce - just shoot a notification every message for now).

Craftplacer:
> No test scene? How are you testing this?

As it's online related, I simply test it against the actual API. I would simply message a bot (e.g. Tillerino) on stable, and let lazer react to the chat update.
Integrating a test scene would probably mean to split up methods and conditions more.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,202 @@
...
+        private void checkForMentions(Channel channel, Message message, string username)
+        {
+            if (!notifyOnMention.Value || !anyCaseInsensitive(getWords(message.Content), username))
+                return;
+
+            var notification = new MentionNotification(message.Sender.Username, channel);
+            notificationOverlay?.Post(notification);
+        }
+
+        private IEnumerable<string> getWords(string input) => Regex.Matches(input, @"\w+", RegexOptions.Compiled).Select(c => c.Value);
```
smoogipoo: Why a regex and not `string.Split(' ')`? This will fail in the case of names like `[John]` or `Doe-`, or any other non-A/N characters we support.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        /// <summary>
+        /// Finds the first matching string/word in both <paramref name="x"/> and <paramref name="y"/> (case-insensitive)
+        /// </summary>
+        private static string hasCaseInsensitive(IEnumerable<string> x, IEnumerable<string> y) => x.FirstOrDefault(x2 => anyCaseInsensitive(y, x2));
```
smoogipoo: Seems unused and is weirdly named.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
```diff
@@ -59,16 +59,12 @@ protected override TabItem<Channel> CreateTabItem(Channel value)
...
         /// Adds a channel to the ChannelTabControl.
-        /// The first channel added will automaticly selected.
         /// </summary>
         /// <param name="channel">The channel that is going to be added.</param>
         public void AddChannel(Channel channel)
         {
             if (!Items.Contains(channel))
                 AddItem(channel);
-
-            if (Current.Value == null)
```
smoogipoo: Why was this removed? This will break in the case where you add the first channel (not from the initial load of channels).

smoogipoo (changes requested):

smoogipoo:
And @peppy 's comment means to imply that a test _should_ be written for this. You should already have the testing structure available for this, look into `TestSceneNotificationOverlay` and `TestSceneStandAloneChatDisplay` and `TestSceneChatOverlay` for direction. Essentially, create the hierarchy, get a channel, post to the channel, make sure that the notification overlay + notification is shown.

osu.Game/Online/Chat/MessageNotifier.cs
bdach: I initially proposed to change from splitting over whitespace because of punctuation. You're right that it will fail for search strings with non-word characters; should go back to splitting over whitespace, but check search term containment without case sensitivity.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
bdach: [I proposed moving this up higher earlier](https://github.com/ppy/osu/pull/7227#discussion_r368304286) to keep this type of logic in one place, but unsure if doing that does break anything.

osu.Game/Online/Chat/MessageNotifier.cs
smoogipoo: The other way to do this is not splitting, but by searching the string, which would cover the punctuation case. Stable already does this and I think it's worked well?

Also: need to replace any space in the name with `_`.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
Craftplacer: I don't have control which channel is selected without this change. The logic of ChatOverlay is, once loaded, the selected tab gets switched over to the first one, causing an event to fire that switches the ChannelManager's channel. I couldn't await the loading of ChatOverlay (because I'm not well inversed with scheduling inside the framework), which would actually help me to not use this change.

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: I will remove the word checking, but are you sure with the underscore (or is it stable behavior)?

smoogipoo: It's stable behaviour. Upon further discussion, do both (match on both no-underscores and with-underscores).

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
smoogipoo: The bigger issue here is that `ChannelManager` should already be doing this, but its current channel is listed as `"+"` which should never be the current channel.

Craftplacer: Have you noticed the changes inside ChatOverlay though?

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,205 @@
...
+
+            var notification = new MentionNotification(message.Sender.Username, channel);
+            notificationOverlay?.Post(notification);
+        }
+
+        /// <summary>
+        /// Checks if <paramref name="message"/> contains <paramref name="username"/>, if not, retries making spaces into underscores.
+        /// </summary>
+        /// <returns>If the <paramref name="message"/> mentions the <paramref name="username"/></returns>
+        public bool IsMentioning(string message, string username)
```
peppy: this shouldn't be public.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,205 @@
...
+        {
+            // sanitize input to handle casing
+            message = message.ToLower();
```
peppy: use `IndexOf` with `StringComparisonOptions.CaseInsensitive`

osu.Game.Tests/Chat/MessageNotifierTests.cs
```diff
@@ -0,0 +1,30 @@
...
+    [TestFixture]
+    public class MessageNotifierTests
+    {
+        private readonly MessageNotifier messageNotifier = new MessageNotifier();
+
+        [Test]
+        public void TestMentions()
+        {
+            // Message (with mention, different casing)
+            Assert.IsTrue(messageNotifier.IsMentioning("Hey, Somebody Playing OSU!", "Somebody playing osu!"));
```
peppy: these tests are really strange. you should be testing by passing in `Message` objects, not strings.

peppy (changes requested):

osu.Game/Online/Chat/MessageNotifier.cs
Craftplacer: How am I supposed to test it then?

peppy:
Please run [code inspection](https://github.com/ppy/osu#code-analysis) locally if you can, rather than relying on server-side CI and fixing issues one commit at a time.

Craftplacer:
I'll see what I can do, like clean cloning the repository, but at the moment the inspectcode script doesn't work for me.

On Wed, Jan 29, 2020, 02:53 Dean Herbert <notifications@github.com<mailto:notifications@github.com>> wrote:

Please run code inspection<https://github.com/ppy/osu#code-analysis> locally if you can, rather than relying on server-side CI and fixing issues one commit at a time.

peppy:
with what error

Craftplacer:
> with what error

```PS A:\Code\dotnet\craftplacer-osu> A:\Code\dotnet\craftplacer-osu\InspectCode.ps1
Tool 'cake.tool' (version '0.35.0') was restored. Available commands: dotnet-cake
Tool 'dotnet-format' (version '3.1.37601') was restored. Available commands: dotnet-format

Restore was successful.
Could not find any assemblies compatible with .NETCoreApp,Version=v3.1 in NuGet package CodeFileSanity. Falling back to using root folder of NuGet package.

========================================
CodeFileSanity
========================================

========================================
InspectCode
========================================
JetBrains Inspect Code 2019.1.1
Running in 64-bit mode, .NET runtime 4.0.30319.42000 under Microsoft Windows NT 6.2.9200.0
JetBrains.ProjectModel.ProjectsHost.Impl.DummySolutionMark+DummySolutionConfigurationAndPlatform
Warning: Data at the root level is invalid. Line 1, position 1.
The project file could not be loaded. Data at the root level is invalid. Line 1, position 1. A:\Code\dotnet\craftplacer-osu\osu.Desktop.slnf at (1:1)
No files to inspect were found.
An error occurred when executing task 'InspectCode'.
Error: One or more errors occurred. (InspectCode: Process returned an error (exit code 1).)
        InspectCode: Process returned an error (exit code 1).
PS A:\Code\dotnet\craftplacer-osu>
```

smoogipoo:
Try delete the `./build/tools` folder. For some reason it's trying to use `JetBrains Inspect Code 2019.1.1` whereas it should be `2019.3` as we have it in the cake file.

Craftplacer:
> Try delete the `./build/tools` folder. For some reason it's trying to use `JetBrains Inspect Code 2019.1.1` whereas it should be `2019.3` as we have it in the cake file.

Yep that fixed it, although I get some assembly warnings.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -0,0 +1,92 @@
...
+using osu.Game.Overlays;
+using osu.Game.Users;
+
+namespace osu.Game.Tests.Visual.Online
+{
+    public class TestSceneMessageNotifier : OsuTestScene
+    {
+        private User friend;
+        private Channel publicChannel;
+        private Channel privateMesssageChannel;
```
bdach:
```suggestion
        private Channel privateMessageChannel;
```

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -0,0 +1,92 @@
...
+            privateMesssageChannel = new Channel(friend) { Id = 2, Name = friend.Username, Type = ChannelType.PM };
+
+            Schedule(() =>
+            {
+                Child = testContainer = new TestContainer(new[] { publicChannel, privateMesssageChannel })
+                {
+                    RelativeSizeAxes = Axes.Both,
+                };
+
+                testContainer.ChatOverlay.Show();
```
bdach: Should also show notification overlay to be able to visually confirm notification presence.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -0,0 +1,92 @@
...
+            [BackgroundDependencyLoader]
+            private void load()
+            {
+                AddRange(new Drawable[] { ChannelManager, NotificationOverlay });
+
+                ((BindableList<Channel>)ChannelManager.AvailableChannels).AddRange(channels);
+
+                AddRange(new Drawable[] { ChatOverlay, MessageNotifier });
+
+                ((BindableList<Channel>)ChannelManager.JoinedChannels).AddRange(channels);
```
bdach: I guess you could also call `JoinChannel`. Test code though, so this time it's nothing too big.

osu.Game/Online/Chat/Message.cs
```diff
@@ -63,5 +63,7 @@ public int CompareTo(Message other)
 
         // ReSharper disable once ImpureMethodCallOnReadonlyValueField
         public override int GetHashCode() => Id.GetHashCode();
+
+        public override string ToString() => $"[{ChannelId}] ({Id}) {Sender}: {Content}";
```
bdach: Perhaps [`[DebuggerDisplay]`](https://docs.microsoft.com/pl-pl/visualstudio/debugger/using-the-debuggerdisplay-attribute?view=vs-2019) would be preferable for debug printing.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -0,0 +1,92 @@
...
+            public ChatOverlay ChatOverlay { get; } = new ChatOverlay();
+
+            [BackgroundDependencyLoader]
+            private void load()
+            {
+                AddRange(new Drawable[] { ChannelManager, NotificationOverlay });
+
+                ((BindableList<Channel>)ChannelManager.AvailableChannels).AddRange(channels);
+
+                AddRange(new Drawable[] { ChatOverlay, MessageNotifier });
```
bdach: I'm not sure why the `AddRange`s are split - is that because they influence order of DI? If so then I'd say that's not good.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -0,0 +1,92 @@
...
+        [Test]
+        public void TestPublicChannelMention()
+        {
+            AddStep("Switch to PMs", () => testContainer.ChannelManager.CurrentChannel.Value = privateMesssageChannel);
+
+            AddStep("Send regular message", () => publicChannel.AddNewMessages(new Message(messageIdCounter++) { Content = "Hello everyone!", Sender = friend, ChannelId = publicChannel.Id }));
+            AddAssert("Expect no notifications", () => testContainer.NotificationOverlay.UnreadCount.Value == 0);
+
+            AddStep("Send message containing mention", () => publicChannel.AddNewMessages(new Message(messageIdCounter++) { Content = $"Hello {API.LocalUser.Value.Username.ToLowerInvariant()}!", Sender = friend, ChannelId = publicChannel.Id }));
+            AddAssert("Expect 1 notification", () => testContainer.NotificationOverlay.UnreadCount.Value == 1);
```
bdach: Should try to switch to the tab visually after notification is received to confirm that the message has been sent.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,198 @@
...
+                Username = username;
+                Channel = channel;
+                Text = $"You received a private message from '{Username}'. Click to read it!";
+            }
+
+            public string Username { get; }
+
+            public Channel Channel { get; }
+
+            public Action<PrivateMessageNotification> OnRemove { get; set; }
```
bdach: Seems unused.

bdach (dismissed):
Getting closer, but tests are hard to visually understand. I'd also like to see more coverage for other test cases (ie. make sure no notifications are fired if the tab is active, notifications aren't re-fired if the messages have been read or if they've been sent by the user themselves).

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
Craftplacer: I had to fiddle with the code to get this working.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
Craftplacer: I should've figured the CI would fail, please rename references too thx.

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
bdach: To be blunt, that answer doesn't inspire confidence in this code.

Craftplacer: True true, I will check this part again and I'll tell you what exactly fails if I change the order.

osu.Game/Online/Chat/Message.cs
Craftplacer: I would have to regress my previous PR then, because it seems to fire off error messages with this class name without the current change.

bdach: I'm not sure I understand.

Craftplacer: Right now:
read state error: "Failed to mark channel up to 'osu.Game.Online.Chat.Message' as read"
debugger: osu.Game.Online.Chat.Message

Your suggestion:
read state error: "Failed to mark channel up to 'osu.Game.Online.Chat.Message' as read"
debugger: "[0] (1) Craftplacer: This doesn't seem right"

If the current code was being pushed:
read state error: "Failed to mark channel up to '[0] (1) Craftplacer: This doesn't seem right' as read"
debugger: "[0] (1) Craftplacer: This doesn't seem right"

bdach: Well, this doesn't mean that this *regresses* anything, but more that the previous logging code was *broken* in the first place. But fine, I suppose this needs to stay then.

Craftplacer: Well, regression is still a fancy word I didn't use yet, only heard from the dev team.

bdach:
Added some more tests here.

I don't think I have any major issues left with this anymore. It's not great, but as far as I'm aware also temporary. Not too sure whether this code really is fine or if I've read it too many times already to be able to tell, so another pair of eyes would be helpful here.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
smoogipoo: I have, and it is causing the issue mentioned here. Still needs to be fixed.

smoogipoo (changes requested):

smoogipoo:
The naming of `channel_NewMessagesArrived` still needs to be fixed.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
bdach: @smoogipoo it might be sleep deprivation talking, but I don't see what the problem is - would you mind elaborating on the problem so I can fix it?

Craftplacer: I mean, if we would go not modifying ChannelTabControl, we have to rework how ChatOverlay reacts to tab changes, as it's definitely unintentionally causing to change the current channel set from other code (in other words, this one right here).

smoogipoo: It seems to also be broken on master, but the intention is that the first channel joined is the first channel selected. This happens post-load, e.g. the user can close all channels and then open up a new channel.

Craftplacer: What I would propose, is that ChatOverlay, if it detects a tab change (from + to *channel) and the current channel (of ChannelManager) is not null, it would set the tab back to the ChannelManager's channel.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,193 @@
...
+            {
+                // ignore messages that already have been read
+                if (message.Id <= channel.LastReadId)
+                    return;
+
+                if (message.Sender.Id == localUser.Value.Id)
+                    continue;
+
+                // check for private messages first, if true, skip checking mentions to prevent duplicate notifications about the same message.
+                if (checkForPMs(channel, message))
```
iiSaLMaN: maybe this would be better.
```csharp
if (checkForPMs(channel, message))
    // check if the PM notification has been posted
    // to avoid potentially sending PM and mention notifications if mentioned on a PM.
    return;

**Review by frenzibyte — DISMISSED** [review]

File: osu.Game/Online/Chat/MessageNotifier.cs (hunk 36 above)
Comment (bdach) [reply osu.Game/Online/Chat/MessageNotifier.cs]: The inline comment is incorrect, as it would stop checking for all notifications altogether (note that this check is inside a `for` loop over a batch of messages). However that might actually not be entirely undesirable, if a little bit brittle.

**Comment by bdach** [comment]
I locally have fixes for everything else aside from the `AddChannel` removal. Readability of the chat overlay isn't great, as `channelManager.CurrentChannel` and `ChannelTabControl.Current` are somewhat cyclically linked (they assign values to each other at times). Will revisit over the coming days.

**Comment by Craftplacer** [comment]
@bdach What's the status of this? This PR is getting quite old...

**Comment by bdach** [comment]
As you can probably tell by the number of open PRs keeping up has been difficult, so I have made no further changes here. Over the coming weeks I want to focus on decreasing that number by applying changes myself whenever I can but it will take time.

However seeing that this implementation will be temporary at most, it's not exactly high priority, either.

**Comment by Craftplacer** [comment]
Ah got it, it's fine.

File: osu.Game/Online/Chat/MessageNotifier.cs [hunk 37]
```diff
@@ -0,0 +1,194 @@
...
+        private void load(OsuConfigManager config, IAPIProvider api)
+        {
+            notifyOnMention = config.GetBindable<bool>(OsuSetting.ChatHighlightName);
+            notifyOnPM = config.GetBindable<bool>(OsuSetting.ChatMessageNotification);
+            localUser = api.LocalUser;
+
+            channelManager.JoinedChannels.BindTo(joinedChannels);
+
+            // Listen for new messages
+            joinedChannels.ItemsAdded += joinedChannels =>
```
smoogipoo, line 50: This being bound after the `BindTo()` above is a little unsafe. It's potentially missing channels if they're added to `JoinedChannels` prior to this being loaded.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,194 @@
...
+        private Bindable<bool> notifyOnPM;
+        private Bindable<User> localUser;
+        private readonly BindableList<Channel> joinedChannels = new BindableList<Channel>();
+
+        [BackgroundDependencyLoader]
+        private void load(OsuConfigManager config, IAPIProvider api)
+        {
+            notifyOnMention = config.GetBindable<bool>(OsuSetting.ChatHighlightName);
+            notifyOnPM = config.GetBindable<bool>(OsuSetting.ChatMessageNotification);
+            localUser = api.LocalUser;
```
smoogipoo, line 45: Either `GetBoundCopy()` or `BindTo()`. Don't copy the bindable by-reference.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,194 @@
...
+                Text = $"Your name was mentioned in chat by '{username}'. Click to find out why!";
+                this.channel = channel;
+            }
+
+            private readonly Channel channel;
+
+            public override bool IsImportant => false;
+
+            [BackgroundDependencyLoader]
+            private void load(OsuColour colours, ChatOverlay chatOverlay, NotificationOverlay notificationOverlay, ChannelManager channelManager)
```
smoogipoo, line 179: Probably not a big deal, but code duplication can be reduced by adding a common base type here (like `OpenChannelNotification`).

smoogipoo (dismissed):

peppy:
@Craftplacer did you give up on this one? There are outstanding reviews which might be the only thing holding this PR back.

Craftplacer:
> @Craftplacer did you give up on this one? There are outstanding reviews which might be the only thing holding this PR back.

I mean I still can continue work on this, but I thought you were going to implement web notifications which are going to supersede this PR (from how I remember the discussion went).

peppy:
The fact that notifications still haven't been implemented probably means that we want to get this merged for the time being.

The mention support isn't even part of the notification system yet, so that will likely stick around for a longer length of time.

Craftplacer:
I see.

I'll check up the state of this pull request later today then and do necessary changes to get this merged.

peppy:
@Craftplacer I've done the initial work to bring this up-to-date (figured it would be quicker for me to do so as it's a pretty hefty merge conflict. Can you check through the code and make sure the behaviour is still as you expect it to be?

Craftplacer:
The current batch of commits fixes the last batch of unresolved reviews, also fixes a minor issue with tests.

At this point I am currently stuck though because I can't get the tests to run locally. Reason for this is that, when testing, `API.LocalUser` becomes unexpectedly null and fails the test. I can't figure out why is that the case.

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,190 @@
...
+
+            var notification = new MentionNotification(message.Sender.Username, channel);
+            notificationOverlay?.Post(notification);
+        }
+
+        /// <summary>
+        /// Checks if <paramref name="message"/> contains <paramref name="username"/>, if not, retries making spaces into underscores.
+        /// </summary>
+        /// <returns>If the <paramref name="message"/> mentions the <paramref name="username"/></returns>
+        private static bool isMentioning(string message, string username) => message.IndexOf(username, StringComparison.OrdinalIgnoreCase) != -1 || message.IndexOf(username.Replace(' ', '_'), StringComparison.OrdinalIgnoreCase) != -1;
```
turbedi, line 143:
```suggestion
        private static bool isMentioning(string message, string username) => message.Contains(username, StringComparison.OrdinalIgnoreCase) || message.Contains(username.Replace(' ', '_'), StringComparison.OrdinalIgnoreCase);
```

Craftplacer: Not bad find

osu.Game.Tests/Visual/Online/TestSceneMessageNotifier.cs
```diff
@@ -28,6 +29,14 @@ public class TestSceneMessageNotifier : OsuManualInputManagerTestScene
         [SetUp]
         public void Setup()
         {
+            // We blindly mark every request as success so that ChannelManager doesn't remove our channel again.
+            if (API is DummyAPIAccess daa)
+            {
+                daa.HandleRequest = (request) => {
+                    return true;
+                };
```
frenzibyte, line 37: It's usually best to only handle the request that the test scene expects to be fired, not handle every single API request. See other usages of `HandleRequest`.

Craftplacer:
There's one "accidential" deletion on ChannelTabControl, but I am holding off to commit the revert to not unnecessarily trigger the CI (push it with other changes if there are any improvements left to be done).

Other than that, there's nothing wrong with the pull request anymore in my eyes.

peppy:
I've brought this in line with our code standards. Should be good to go now.

Also as a side-note, this never worked correctly due to the load order being incorrect (`ChatOverlay` would always be `null` in actual game usage as it was loaded single-file after the notifier component itself).

peppy (dismissed):

osu.Game/Online/Chat/MessageNotifier.cs
```diff
@@ -0,0 +1,177 @@
...
+        [BackgroundDependencyLoader]
+        private void load(OsuConfigManager config, IAPIProvider api)
+        {
+            notifyOnUsername = config.GetBindable<bool>(OsuSetting.NotifyOnUsernameMentioned);
+            notifyOnPrivateMessage = config.GetBindable<bool>(OsuSetting.NotifyOnPrivateMessage);
+
+            localUser.BindTo(api.LocalUser);
+
+            joinedChannels.BindCollectionChanged(channelsChanged);
+            joinedChannels.BindTo(channelManager.JoinedChannels);
```
bdach, line 50: This does not look 100% right (callback registered before bind, and in BDL to boot), not sure if there's a valid reason for that. Can fix up, just marking down.

bdach (commented):
Doesn't seem like this should explode.

osu.Game/Online/Chat/MessageNotifier.cs
peppy: The callback before bind is a cheap way of handling the initial population from the `BindTo` call, but as we now have `BindCollectionChanged` this ordering is no longer required. But probably best to move out of BDL, sure.

bdach (approved):

PR #7242, opened 2019-12-17
Implement counter pills for UserOverlay

EVAST9919:
Screenshot_1
Screenshot_2
Screenshot_3

### DIFF
--- a/osu.Game/Overlays/Profile/Sections/Beatmaps/PaginatedBeatmapContainer.cs
+++ b/osu.Game/Overlays/Profile/Sections/Beatmaps/PaginatedBeatmapContainer.cs
@@ -38,5 +38,27 @@ namespace osu.Game.Overlays.Profile.Sections.Beatmaps
                 Anchor = Anchor.TopCentre,
                 Origin = Anchor.TopCentre,
             };
+
+        protected override int GetCount(User user)
+        {
+            switch (type)
+            {
+                default:
+                case BeatmapSetType.Favourite:
+                    return user.FavouriteBeatmapsetCount;
+
+                case BeatmapSetType.Graveyard:
+                    return user.GraveyardBeatmapsetCount;
+
+                case BeatmapSetType.Loved:
+                    return user.LovedBeatmapsetCount;
+
+                case BeatmapSetType.RankedAndApproved:
+                    return user.RankedAndApprovedBeatmapsetCount;
+
+                case BeatmapSetType.Unranked:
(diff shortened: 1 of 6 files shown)


huoyaoyuan:
The position in the image looks a bit higher than the text.

osu.Game/Overlays/Profile/Sections/CounterPill.cs
```diff
@@ -0,0 +1,61 @@
...
+            };
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            Current.BindValueChanged(onCurrentChanged, true);
+        }
+
+        private void onCurrentChanged(ValueChangedEvent<int> value)
```
huoyaoyuan: It looks a bit clearer to use lambda and `else` instead of `return` here.

osu.Game/Overlays/Profile/Sections/Beatmaps/PaginatedBeatmapContainer.cs
```diff
@@ -38,5 +38,27 @@ public PaginatedBeatmapContainer(BeatmapSetType type, Bindable<User> user, strin
                 Anchor = Anchor.TopCentre,
                 Origin = Anchor.TopCentre,
             };
+
+        protected override int GetCount(User user)
+        {
+            switch (type)
```
huoyaoyuan: Switch expressions can be correctly handled now. Here is the definitive use case. I turned off the suggestion, and will turn it on soon.

osu.Game/Overlays/Profile/Sections/Ranks/PaginatedScoreContainer.cs
```diff
@@ -32,6 +32,18 @@ public PaginatedScoreContainer(ScoreType type, Bindable<User> user, string heade
         protected override APIRequest<List<APILegacyScoreInfo>> CreateRequest() =>
             new GetUserScoresRequest(User.Value.Id, type, VisiblePages++, ItemsPerPage);
 
+        protected override int GetCount(User user)
+        {
+            switch (type)
```
huoyaoyuan: Also switch expression here.

osu.Game/Overlays/Profile/Sections/CounterPill.cs
smoogipoo: I think it's fine.

smoogipoo (approved):

PR #7244, opened 2019-12-17
Move CreateScoreProcessor() to Ruleset

smoogipoo:
Still with the goal of eventually being able to construct these things to reprocess scores.


peppy (approved):
looks sane

PR #7250, opened 2019-12-17
Fix editor crashing after re-ordering objects

### LINKED ISSUE #6733: Editor sometimes throws exceptions when deleting hitobjects
Seems to be a bug related to hitobject order inside the `LifetimeManagementContainer` - getting a "key was not present in dictionary" exception.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/6733

Not sure if this is the best/cleanest fix - I'm open for suggestions.

The way this works is by preserving the start time used for the comparer while the start time is changed. During the change process, the preserved start time allows us to remove and re-add the hitobjects, effectively re-sorting them.

This may be a bit over-engineered but I think it's best to be a bit safe (re: bound bindable copy).


peppy (dismissed):

peppy:
Probably nice to have a regression test for this one.

peppy (approved):

PR #7252, opened 2019-12-17
Add ability to delete individual scores

### LINKED ISSUE #7146: Ability to delete scores
Title is descriptive enough.

**Proposal designs of the feature:** Right-click for a context menu with item "Delete". Clicking it gives a dialog asking for confirmation.

wltu:
Resolves #7146

The changes added the ability to show a context menu with the option to delete on local `LeaderboardScore` only.  Clicking the delete option with pop up a `BeatmapClearScoresDialog` with the confirmation along with the information about the score.  Confirming it will delete the score and refresh the scoreboard.

The scope of this implementation as far as I know includes the `BeatmapDetailArea` and the `LocalLeaderboardPage`.

One issue that I am having with this implementation is that to refresh the scoreboard. To do that I created an Action variable in `LearderboardScore` and passed in `BeatmapLeaderboard.RefreshScores()` as an Action to that. I am not sure if that is the best way to do it.


osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
```diff
@@ -39,6 +39,28 @@ public BeatmapClearScoresDialog(BeatmapInfo beatmap, Action onCompletion)
             };
         }
 
+        public BeatmapClearScoresDialog(ScoreInfo score, Action onCompletion)
+        {
+            string accuracy = string.Format(score?.Accuracy % 1 == 0 ? @"{0:P0}" : @"{0:P2}", score?.Accuracy);
+
+            BodyText = $@"{score?.Beatmap?.Metadata?.Artist} - {score?.Beatmap?.Metadata?.Title} {Environment.NewLine} {score?.User} - Rank: {score?.Rank} - Max Combo: {score?.MaxCombo} - {accuracy} - {score?.Date.Date.ToShortDateString()}";
```
bdach: I'm pretty sure these null propagation operators are a bit too lenient - `score == null` means catastrophe and it'll most likely just fail somewhere down below like `ScoreManager`. I'd guard (either with an assert, or just plain exception) and drop the `?`s next to `score?`s.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
```diff
@@ -39,6 +39,28 @@ public BeatmapClearScoresDialog(BeatmapInfo beatmap, Action onCompletion)
             };
         }
 
+        public BeatmapClearScoresDialog(ScoreInfo score, Action onCompletion)
```
bdach: I'm a bit opposed against this landing in here. This new dialog is arguably *not* a `BeatmapClearScoresDialog` since it doesn't really clear scores from a beatmap. I'd argue this class needs at least a rename, if not some proper genericizing and refactoring.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
```diff
@@ -39,6 +39,28 @@ public BeatmapClearScoresDialog(BeatmapInfo beatmap, Action onCompletion)
...
+            BodyText = $@"{score?.Beatmap?.Metadata?.Artist} - {score?.Beatmap?.Metadata?.Title} {Environment.NewLine} {score?.User} - Rank: {score?.Rank} - Max Combo: {score?.MaxCombo} - {accuracy} - {score?.Date.Date.ToShortDateString()}";
+
+            Icon = FontAwesome.Solid.Eraser;
+            HeaderText = @"Clearing this local score. Are you sure?";
+            Buttons = new PopupDialogButton[]
+            {
+                new PopupDialogOkButton
+                {
+                    Text = @"Yes. Please.",
+                    Action = (() => scoreManager.Delete(score)) + onCompletion
```
bdach: While this isn't going to be as expensive as bulk removal of scores for one beatmap, there could still be some DB work involved, which means I/O. We might want this to be asynchronous mirroring how the other one is done.

bdach (commented):
Tests would be nice, but I imagine they might be a little complicated to write for this scenario.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: Ok. I'll fix this. 

Just to make sure we are on the same page. So we want to make sure score is not null and if it is null we want it to show an error in the dialog or just exist it.

> Tests would be nice, but I imagine they might be a little complicated to write for this scenario.

And for the testing, I had some issue setting up test environment for this.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: OK. Will fix it. I was also debating whether I want to create a new class for it or not.

How does `ClearLocalScoreDialog` sound for this?

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: I did have tried to do it asynchronously, but the score board does not refresh once the delete is completed.

Will look more into it and fix it.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
peppy: `BeatmapInfo` already has a `ToString`. Check other usages in the code and use what they do.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
peppy: no need for async for now. keep it simple.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
```diff
@@ -182,9 +182,15 @@ protected override APIRequest FetchScores(Action<IEnumerable<ScoreInfo>> scoresC
...
+        protected override LeaderboardScore CreateDrawableScore(ScoreInfo model, int index)
         {
-            Action = () => ScoreSelected?.Invoke(model)
-        };
+            model.Beatmap = beatmap;
+
+            return new LeaderboardScore(model, index, IsOnlineScope)
+            {
+                Action = () => ScoreSelected?.Invoke(model),
+                RefreshAction = () => this.RefreshScores()
```
peppy: this should not be a thing. subscribe to the `ScoreManager` for changes.

peppy (changes requested):

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
bdach: Sure, I'm fine with that name.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: I have updated to 
`BodyText = $@"{score} {Environment.NewLine} Rank: {score.Rank} - Max Combo: {score.MaxCombo} - {accuracy}";`

This will use the ToString for `ScoreInfo` which also uses the  ToString of `BeatmapInfo`. The rank and max combo is not included in the ToString, so I have to do that if we still want the information to be displayed.


This will display:
`{User} playing {Artist} - {Title} ({Author}) [{Version}]`
`{score.Rank} - Max Combo: {score.MaxCombo} - {accuracy}`

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: ok sounds good for now.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
wltu: I am a bit confused about what you meant by subscribing to `ScoreManager`. I did this in attempt to have the ability to refresh the score board once the local score is deleted. This is based on `BeatmapDetailArea.Leaderboard.RefreshScores()`. This is dependent on the learderboard and not the `ScoreManager`.

Do you have any suggestion on how to refresh the score board inside `LeaderboardScore`? Any advise would be appreciated. Unless we don't think that is a necessary feature.

bdach: `ScoreManager`, as all `ArchiveModelManager`s, exposes an event that fires whenever an item is deleted:
https://github.com/ppy/osu/blob/d63ab362dcb94f4e56f84853a2b36c0bbb0c8c93/osu.Game/Database/ArchiveModelManager.cs#L61-L65
I'm reasonably confident that's what's being proposed here. Mind the xmldoc though (the note about the callback not being called on the update thread) - you'll have to schedule the refresh. Have a look around the repository for other usages of `Item{Added,Removed}` for an example how to do so.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
peppy: Clear generally refers to clearing more than one object. Delete is better IMO.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
peppy: Precisely. Should be subscribing to both in the local leaderboard and updating display.

osu.Game/Screens/Select/BeatmapClearScoresDialog.cs
wltu: Yeah, I realized that a bit after. I have changed it to `LocalScoreDeleteDialog`.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
wltu: Thanks for the information. This made a lot more sense than whatever I was doing.

I have updated it and subscribed to the `scoreManager` from `BeatmapLeaderboard`  which would affect both the local leaderboard and the beatmap detail area (I am assuming that is what you are referring to for updating display).

After making the changes, the score board for both beatmap detail are and local leaderboard were able to refresh after deleting a score. However, the only issue is that the refresh screen will also get triggered when doing clearing all local score for every local score.

This might be an issue if there are a lot of local scores.

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
```diff
@@ -0,0 +1,62 @@
...
+                        Text = @"Yes. Please.",
+                        Action = () => scoreManager.Delete(score)
+                    },
+                    new PopupDialogCancelButton
+                    {
+                        Text = @"No, I'm still attached.",
+                    },
+                };
+            }
+            catch (Exception e)
```
smoogipoo: This implementation is unnecessary. It's a situation which should never happen and is indicative of a developer error. This should be rewritten in the form:

```csharp
public LocalScoreDeleteDialog(ScoreInfo score)
{
    Debug.Assert(score != null);

    string accuracy = ...;
    ...
}
```

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
```diff
@@ -0,0 +1,62 @@
...
+    {
+        private ScoreManager scoreManager;
+
+        public LocalScoreDeleteDialog(ScoreInfo score)
+        {
+            try
+            {
+                string accuracy = string.Format(score?.Accuracy % 1 == 0 ? @"{0:P0}" : @"{0:P2}", score?.Accuracy);
+
+                BodyText = $@"{score} {Environment.NewLine} Rank: {score.Rank} - Max Combo: {score.MaxCombo} - {accuracy}";
```
smoogipoo: I'd make this match beatmap deletion, rather than listing rank/combo/accuracy/etc. Here's a comparison:

image

image

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
```diff
@@ -182,9 +194,22 @@ protected override APIRequest FetchScores(Action<IEnumerable<ScoreInfo>> scoresC
             return req;
         }
 
-        protected override LeaderboardScore CreateDrawableScore(ScoreInfo model, int index) => new LeaderboardScore(model, index, IsOnlineScope)
+        protected override LeaderboardScore CreateDrawableScore(ScoreInfo model, int index)
         {
-            Action = () => ScoreSelected?.Invoke(model)
-        };
+            model.Beatmap = beatmap;
```
smoogipoo: What's this about?

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -359,5 +367,23 @@ public LeaderboardScoreStatistic(IconUsage icon, string name, string value)
...
+            if (score == null || score.ID <= 0) return;
+
+            dialogOverlay?.Push(new LocalScoreDeleteDialog(score));
+        }
+
+        public MenuItem[] ContextMenuItems
+        {
+            get
+            {
+                return (this.allowHighlight) ? null : new MenuItem[]
```
smoogipoo: This seems like the wrong variable to use here. Maybe you can just use the databased `score.ID` instead?

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -359,5 +367,23 @@ public LeaderboardScoreStatistic(IconUsage icon, string name, string value)
                 Value = value;
             }
         }
+
+        private void deleteLocalScore(ScoreInfo score)
+        {
+            if (score == null || score.ID <= 0) return;
```
smoogipoo: This should be impossible.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
```diff
@@ -182,9 +194,22 @@ protected override APIRequest FetchScores(Action<IEnumerable<ScoreInfo>> scoresC
...
+            return new LeaderboardScore(model, index, IsOnlineScope)
+            {
+                Action = () => ScoreSelected?.Invoke(model)
+            };
+        }
+
+        private void deleteLocalScore(ScoreInfo score)
+        {
+            if (score == null)
+                return;
```
smoogipoo: This is impossible.

smoogipoo (changes requested):

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
wltu: Sounds good to me.

I wasn't sure whether doing an assertion or exception is better. 

I have updated with an assertion.

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
wltu: I would argue that the score information are necessary since this feature is deleting individual local score. Please let me know if there are any specific reasons that you would recommend your suggestion.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
wltu: I have found that the `ScoreInfo` actually does not contain the `BeatmapInfo` for some reason. So this is to make sure that the score have that information.

osu.Game/Online/Leaderboards/LeaderboardScore.cs
wltu: My intention here is to have only the `ContextMenu` show up for local scores, so any online scores will not be affected. 

Please let me know if that is still the wrong choice of variable to use in this context.

osu.Game/Online/Leaderboards/LeaderboardScore.cs
wltu: Noted and have been removed.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
wltu: Noted and have been removed.

wltu (commented):
I have review the suggestion and made the changes, except for the information regarding to the delete dialog which required further discussion.

osu.Game/Online/Leaderboards/LeaderboardScore.cs
smoogipoo: My point is that allowHighlight doesn't indicate that a score is local or not. Maybe we'll want to allow highlighting local scores in the future, for example, in which case `allowHighlight` will be true for local scores too.

`score.ID` being non-zero should tells you the score was definitely sourced from the database (thus, definitely local).

wltu: I see what you are saying.

I got the impression that `allowHighlight` is directly related to local from `BeatmapLeaderboard` when it called the `LeaderboardScore`.
```
protected override LeaderboardScore CreateDrawableScore(ScoreInfo model, int index)
{
    model.Beatmap = beatmap;

    return new LeaderboardScore(model, index, IsOnlineScope)
    {
        Action = () => ScoreSelected?.Invoke(model)
    };
}
```

I was unaware that `score.ID` being non-zero basically mean it is local. That being said. I will made the changes accordingly.

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
Apo11oxd: at the very least, i would consider having the score information display in a more digestable format

"smoogipoo's 68.25% D Rank on TRF - Survival dAnce ~no no cry more~ [Insane]" might be an immediate improvement to the current format

wltu: Yeah, I agree. I will make the changes accordingly.

wltu:
I believe I made all the suggested changes except for the information on the `PopupDialog` for confirming the delete.


Currently I have made the changes to:
`{User}'s {accuracy} {Rank} Rank on {Beatmap}`

where `{Beatmap}` is converted to `{Artist} - {Title} ({Author}) [{Version}]` in `BeatmapInfo`'s ToString()

wltu:
@smoogipoo  and @peppy can you review the changes again?

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -62,9 +68,10 @@ public LeaderboardScore(ScoreInfo score, int rank, bool allowHighlight = true)
         }
 
         [BackgroundDependencyLoader]
-        private void load(IAPIProvider api, OsuColour colour)
+        private void load(IAPIProvider api, OsuColour colour, DialogOverlay overlay)
```
iiSaLMaN: Use `[Resolved]` attribute instead:
```csharp
    [Resolved]
    private DialogOverlay dialogs { get; set; }
```

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
```diff
@@ -0,0 +1,45 @@
...
+                },
+                new PopupDialogCancelButton
+                {
+                    Text = @"No, I'm still attached.",
+                },
+            };
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(ScoreManager scoreManager)
```
iiSaLMaN: same thing here.

osu.Game/Screens/Select/Leaderboards/BeatmapLeaderboard.cs
iiSaLMaN: Feels like this should be fixed somewhere else, probably where it's created.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapLeaderboard.cs
```diff
@@ -281,5 +290,13 @@ public void SetRetrievalState(PlaceholderState state)
                 PlaceholderState = state;
             }
         }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Dependencies.Cache(dialogOverlay);
+
+            Add(leaderboard);
```
iiSaLMaN: why is this added on load?

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapLeaderboard.cs
```diff
@@ -281,5 +290,13 @@ public void SetRetrievalState(PlaceholderState state)
                 PlaceholderState = state;
             }
         }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Dependencies.Cache(dialogOverlay);
```
iiSaLMaN: use `[Cached]` attribute:
```csharp
    [Cached]
    private readonly DialogOverlay dialogs;
```

frenzibyte (changes requested):

osu.Game/Online/Leaderboards/LeaderboardScore.cs
wltu: Thanks for the suggestion. I have made the changes.

osu.Game/Screens/Select/LocalScoreDeleteDialog.cs
wltu: I have also made the changes here. This was based on `BeatmapClearScoresDialog.cs`, so I have also made the changes in this file.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapLeaderboard.cs
wltu: This is based on `TestSceneSettingsPanel.cs`. I was looking for ways to get the dialogOverlay to be included in the test. I have updated it after resolving the issue regarding the `Cached` attribute.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapLeaderboard.cs
wltu: Thanks for the suggestion. I have made the changes accordingly.

EVAST9919:
> @smoogipoo and @peppy can you review the changes again?

Please, be patient: there are 82 pr's waiting for review. Noone will forgot about yours.

smoogipoo:
I've refactored a lot (most) of this PR to flow a little better. In particular:
1. A component such as the beatmap leaderboard should not modify models. There are two ways to solve this - either the model needs to already be returned to it with the correct data from the higher level, or (the path that I chose) the data should be retrieved in the correct format from the database.
2. We prefer to have automated tests for logic, rather than just visual "DIY" tests, especially in places where important logic is involved.

Other than that the general ideas and logic you had were fine. Please look through my changes to get an idea of how the above was achieved.

@peppy should be good to go. only things i'm unsure about are the button quips, but it's up to you how much variety we want there. requesting review.

smoogipoo (dismissed):

peppy:
New test looks to be failing

```
  X TestDeleteViaRightClick [92ms]
  Error Message:
   TearDown : System.InvalidOperationException : Sequence contains no elements
  Stack Trace:
  --TearDown
   at System.Linq.ThrowHelper.ThrowNoElementsException()
   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
   at osu.Game.Tests.Visual.UserInterface.TestSceneDeleteLocalScore.<TestDeleteViaRightClick>b__13_1() in C:\Users\Dean\Projects\osu\osu.Game.Tests\Visual\UserInterface\TestSceneDeleteLocalScore.cs:line 138
   at osu.Framework.Testing.Drawables.Steps.SingleStepButton.<.ctor>b__1_0()
   at osu.Framework.Testing.Drawables.Steps.StepButton.PerformStep(Boolean userTriggered)
   at osu.Framework.Testing.TestScene.runNextStep(Action onCompletion, Action`1 onError, Func`2 stopCondition)
--- End of stack trace from previous location where exception was thrown ---
   at osu.Game.Tests.Visual.OsuTestScene.OsuTestSceneTestRunner.RunTestBlocking(TestScene test) in C:\Users\Dean\Projects\osu\osu.Game\Tests\Visual\OsuTestScene.cs:line 320
   at osu.Framework.Testing.TestScene.RunTests()
```

wltu:
> I've refactored a lot (most) of this PR to flow a little better. In particular:
> 
> 1. A component such as the beatmap leaderboard should not modify models. There are two ways to solve this - either the model needs to already be returned to it with the correct data from the higher level, or (the path that I chose) the data should be retrieved in the correct format from the database.
> 2. We prefer to have automated tests for logic, rather than just visual "DIY" tests, especially in places where important logic is involved. Admittedly this was a chunky change, but I hope my changes can give you an idea of how to do it in the future.
> 
> Other than that the general ideas and logic you had were fine. Please look through my changes to get an idea of how the above was achieved.
> 
> @peppy should be good to go. only things i'm unsure about are the button quips, but it's up to you how much variety we want there. requesting review.

Thanks for the changes.

I have looked over the changes you made regarding the things you pointed out. 

I have a good understanding on how you were able to get the beatmap information in the DialogOverlay from the database instead of getting it from the leaderboard. My only concern is that wouldn't it be better if we were to go with the first approach you mentioned. Having the correct data at a higher level could allow other places to have the information without having extra access to the database. Given that the second approach is the easier fix, maybe the first approach could be the next step?

As for part regarding automated tests, I have a better understanding of how to set it up now. I have tried to do the automated tests before but could not get to the correct Drawable with `InputManager`. I was not aware that `ChildrenOfType<T>` exists.

Thanks for helping me understand the codebase flow better.

smoogipoo:
It depends on how much data we want to retrieve for a particular score from the database. In most cases having just the beatmap info is enough since we're using it to retrieve a `WorkingBeatmap`. This case is an exception to that.
This will probably change after we move to Realm, since it'll lazily/dynamically evaluate these properties.

peppy (dismissed):

peppy:
Not sure of how much value the clicking visual test is here (in terms of how much effort is required to implement it) but I guess it's nice to have.

peppy (dismissed):

PR #7269, opened 2019-12-18
Use switch expression in the code base

huoyaoyuan:
It's finally here.
I scanned literally every file by my eye and found cases not handled by the auto converter. This sounds insane, but I think it's worthwhile for such an important feature.

Kept ReSharper suggestion at HINT instead of SUGGESTION, because it's a little aggressive in some scenarios.

I'm also doing little refactors for scenarios that looks fit to me.

Not included:
- Not preferring switch expressions for event handling.
- Switch expression doesn't support LValues and ref returns now.

### DIFF
--- a/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
@@ -148,7 +148,7 @@ namespace osu.Game.Beatmaps.Formats
                         1 => new Rulesets.Objects.Legacy.Taiko.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
                         2 => new Rulesets.Objects.Legacy.Catch.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
                         3 => new Rulesets.Objects.Legacy.Mania.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
-                        _ => parser,
+                        _ => throw new FormatException($"Unknown ruleset ID {beatmap.BeatmapInfo.RulesetID}."),
                     };
                     break;
 
--- a/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
@@ -142,26 +142,14 @@ namespace osu.Game.Beatmaps.Formats
 
                 case @"Mode":
                     beatmap.BeatmapInfo.RulesetID = Parsing.ParseInt(pair.Value);
-
-                    switch (beatmap.BeatmapInfo.RulesetID)
+                    parser = beatmap.BeatmapInfo.RulesetID switch
                     {
-                        case 0:
-                            parser = new Rulesets.Objects.Legacy.Osu.ConvertHitObjectParser(getOffsetTime(), FormatVersion);
-                            break;
-
-                        case 1:
-                            parser = new Rulesets.Objects.Legacy.Taiko.ConvertHitObjectParser(getOffsetTime(), FormatVersion);
-                            break;
-
-                        case 2:
-                            parser = new Rulesets.Objects.Legacy.Catch.ConvertHitObjectParser(getOffsetTime(), FormatVersion);
-                            break;
-
-                        case 3:
-                            parser = new Rulesets.Objects.Legacy.Mania.ConvertHitObjectParser(getOffsetTime(), FormatVersion);
-                            break;
-                    }
-
+                        0 => new Rulesets.Objects.Legacy.Osu.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        1 => new Rulesets.Objects.Legacy.Taiko.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        2 => new Rulesets.Objects.Legacy.Catch.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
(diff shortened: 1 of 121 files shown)


peppy:
not sure about this one. i still prefer existing switch notation and find it more legible.

huoyaoyuan:
I'm also not sure too. If each case is in 1 or 2 lines, expression will be undoubtfully more legible.
Let's discuss case by case.

osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
```diff
@@ -142,26 +142,14 @@ private void handleGeneral(string line)
...
-                        case 3:
-                            parser = new Rulesets.Objects.Legacy.Mania.ConvertHitObjectParser(getOffsetTime(), FormatVersion);
-                            break;
-                    }
-
+                        0 => new Rulesets.Objects.Legacy.Osu.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        1 => new Rulesets.Objects.Legacy.Taiko.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        2 => new Rulesets.Objects.Legacy.Catch.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        3 => new Rulesets.Objects.Legacy.Mania.ConvertHitObjectParser(getOffsetTime(), FormatVersion),
+                        _ => throw new FormatException($"Unknown ruleset ID {beatmap.BeatmapInfo.RulesetID}."),
```
bdach: A bit of a sneaky semantics change here.

osu.Game/Online/Chat/MessageFormatter.cs
```diff
@@ -116,70 +116,35 @@ public static LinkDetails GetLinkDetails(string url)
 
             switch (args[0])
             {
-                case "http":
-                case "https":
-                    // length > 3 since all these links need another argument to work
-                    if (args.Length > 3 && (args[1] == "osu.ppy.sh" || args[1] == "new.ppy.sh"))
+                case "http" when args.Length == 4 && (args[1] == "osu.ppy.sh" || args[1] == "new.ppy.sh"):
+                case "https" when args.Length == 4 && (args[1] == "osu.ppy.sh" || args[1] == "new.ppy.sh"):
```
bdach: Opposed to this condition duplication.

bdach:
I scrolled through this. It's a mixed bag for me. My two cents:

* Neutral/not opposed to the most common cases of mapping short switches (4-5 cases max) to switch expressions.
* In favour of mapping long switches to switch expressions (examples: [`LegacyStoryboardDecoder`](https://github.com/ppy/osu/pull/7269/files#diff-d849ecd63b70bccb909caaf97618e476), [`SerializationReader`](https://github.com/ppy/osu/pull/7269/files#diff-26948673b950c819f519403030bd9b28), [second one in `DrawableRecentActivity`](https://github.com/ppy/osu/pull/7269/files#diff-0e28f973a978ef403674b80cf9f4b6d7)).
* Both cases with tuple shenanigans I saw ([`DrawableHistoryKudosuItem`](https://github.com/ppy/osu/pull/7269/files#diff-84c480f61c8e04886ff686d5ca6a678b) and [`BeatmapCarousel`](https://github.com/ppy/osu/pull/7269/files#diff-17cb1f4d2c61ec9f4581df7ddae8f3ad)) I like due to getting the point across briefly.
* Opposed to replacing switches with fallthroughs to switch expressions due to code duplications.
* Strongly opposed to switch expressions with property conditions due to lack of readability.

I'm also a bit miffed by you slipping through some other cool changes here, which might require splitting off into their own PRs anyway (and would benefit from doing so due to their obvious correctness), namely:
* [pattern matching in `SerializationWriter`](https://github.com/ppy/osu/pull/7269/files#diff-b907bd228c7d188d913a6d15ce6d29d3), and
* [introducing slice expressions in `OsuColour`](https://github.com/ppy/osu/pull/7269/files#diff-a23e7492e934985f94b8cb1f703df90a).

huoyaoyuan:
My personal judgement here:

> Neutral/not opposed to the most common cases of mapping short switches (4-5 cases max) to switch expressions

4 cases are giving a positive favour to me. Most of small cases (2-3) are auto converted.
In addition, it gives a strong positive favour if it can be inlined into the consuming expression.

> Both cases with tuple shenanigans I saw (DrawableHistoryKudosuItem and BeatmapCarousel) I like due to getting the point across briefly.

I find opportunity with tuple if their are 2 coupled values.

> Opposed to replacing switches with label fallthroughs to switch expressions due to code duplications.

The shorter the body, the more tolerance here. I tolerate it when duplications is lesser than 3 and the bodies are very short (enum or `new Color4`).

> Strongly opposed to switch expressions with property conditions due to lack of readability.

Interesting. I think it increases readability a lot.

> I'm also a bit miffed by you slipping through some other cool changes here

I realized this. I wrote the code many days ago with "how can I optimize it" in mind.

osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
huoyaoyuan: Yes, and I think they are necessary. Forgot to write in PR description.

osu.Game/Online/Chat/MessageFormatter.cs
huoyaoyuan: The duplication here eliminates the duplication of `External`. Duplicating here gives a weak positive favour to me.

huoyaoyuan:
Sure. I may provide some comparisons on stylish decisions.

PR #7270, opened 2019-12-18
Refactor storyboard timeline to reduce GC / fix crashes

### LINKED ISSUE #7268: Storyboard error
Some storyboards can't be visible after recent storyboard-related changes.
Beatmap example: https://osu.ppy.sh/beatmapsets/236697#osu/591445
Logs: [runtime.log](https://github.com/ppy/osu/files/3978754/runtime.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/7268

I pulled this out from my WIP storyboard performance branch - not only fixing the issue (Min/Max throw exceptions when there's no elements in the sequence), but also refactoring that stuff.

I've kept it as simple as possible for now - the `OrderBy` enumerable can/has been removed with a bit more work, but it's not done in this PR.

### DIFF
--- a/osu.Game/Storyboards/CommandTimeline.cs
+++ b/osu.Game/Storyboards/CommandTimeline.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osu.Framework.Caching;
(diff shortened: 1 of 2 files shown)


smoogipoo:
Also the actual GC reduction as a result of this is not all that much - most of the GC overhead resides in the areas around adding transforms to a `Drawable`.. It's maybe 1-2 seconds faster on `world.execute(me)` on a rough benchmark.

peppy (approved):

Issue #6125, opened 2019-09-15
Add cancel download/import button at osu!direct overlay


LanTenDragon:
**Describe the new feature:** In the osu!direct overlay, the download button should change to a cancel button after the player has started a download. 

This would be a quality of life improvement, as sometimes when I change my mind on downloading a map, I have to reach the notification overlay, and by that time, the import is already completed.   

**Proposal designs of the feature:** The download button changes to red and and shows an "X". Text on hover is "Cancel"

peppy:
Maybe a dialog on clicking the in-progress button to allow cancelling? Not sure.

elleyer:
I'm not sure it's an important thing because typical beatmap uses only few MB's of the memory, and you can delete it at any moment. Also it downloads in a few seconds, so you will not have a time for cancel it.

Gabixel:
That's true, but this feature can come in handy for some rare moments: the easiest one is the misclicking (but that's just an example). My idea is to double-click on the download button (or just pressing a cancel button next to the download one).

HerendraTJ:
In Stable version there isn't an option to cancel download. Unless in Lazer there is a download manager that control the download. Since before download , people being played the replay and right now you can see the beatmap listings without opening the browser. So you can consider which map and who mapped the beatmap.

peppy:
In stable there *is* an option. In lazer you *can* already cancel the download via the notification area.

iiSaLMaN:
Are we agreeing on a design here?

peppy:
A dialog popup allowing cancelling would be a good start, as I mentioned above.

PR #7297, opened 2019-12-20
Allow cancelling existing download request of a model

frenzibyte:
- Adds a method to cancel existing download request of a model.
- Dependency for resolving #6125

### DIFF
--- a/osu.Game/Database/DownloadNotification.cs
+++ b/osu.Game/Database/DownloadNotification.cs
@@ -1,0 +1,23 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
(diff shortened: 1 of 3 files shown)


peppy:
I'm not sure why the notification change is required (and the addition of the "package" class). Can't you just use the existing code and cancel the request?

iiSaLMaN:
> I'm not sure why the notification change is required (and the addition of the "package" class). Can't you just use the existing code and cancel the request?

If it's fine for the notifications to remain there as `Cancelled`, sure.

peppy:
The discord link doesn't work. Please post here what you need to ask.

iiSaLMaN:
I was asking you if cancelling a download request from out of the notifications overlay should close the downlaod notification immediately or keep it, I went with closing it immediately.

Here's how it looks with keeping the notification: https://streamable.com/mw39s

peppy:
That looks fine for now.

PR #7303, opened 2019-12-20
Update song select statistics when mod changes settings

peppy:
Using a temporary method of tracking these as alternatives were attempted but deemed too complex (reflection on reflection).

Will add a final test in https://github.com/ppy/osu/pull/7157.

- [x] Add tests
- [x] Depends on https://github.com/ppy/osu-framework/pull/3131


huoyaoyuan:
Idea:
- Add a non-generic `BindValueChanged` method to `IBindable`, and implement it in `IBindable<T>` via default interface method.
- Extract `CreateSettingsControls` into 2 steps: discovering settings into `IEnumerable<IBindable>` and converting them into controls.

peppy:
Good ideas, we have the same thoughts:

We looked at #1 but held off because it is hard to enforce correct event ordering in the bindable flow (between generic and non-generic versions).

We looked at #2 but it was reflection overload. It looked like it would have worked but we decided the complexity was best left for another time. If you want to have a go at this you're more than welcome to, but I'd hope we could do it in 3-4 reflection calls at most. We were at 8 or something before deciding against it.

smoogipoo:
In regards to (1), it isn't hard to enforce correct ordering, but it's hard to delegate the removal of event handlers. Or maybe I don't know if there's a way to do it without using a separate mapping dictionary.

```
event IBindable.ValueChanged
{
    add => ValueChanged += _ => value();
    remove => /* todo: how? */
}
```

huoyaoyuan:
Event handler removal has been one of the biggest pain I've experienced in UI binding.
```csharp
public interface IBindable
{
    event Action ValueChanged;
}

public interface IBindable<T>
{
    new event Action<ValueChangedEvent<T>> ValueChanged;

    protected sealed void TriggerValueChanged(ValueChangedEvent<T> e, Action<ValueChangedEvent<T>> generic, Action nonGeneric)
    {
        // or place into implementation
        generic?.Invoke(e);
        nonGeneric?.Invoke(e);
    }
}
```

Requires the implementation side to be aware about the non generic events, but totally achievable.

smoogipoo:
Doing it that way breaks event ordering. I'm not sure it's a bad thing since you shouldn't bind to bindables outside of your own self in the first place, but I'd need time to convince myself on that before settling on it, since we don't want to change that in the future and break every consumer of `Bindable` everywhere.

For example:

```
class MyParentObject:
    ctor():
        Add(new MyChildObject());
        MyChildObject.Bindable.ValueChanged += _ => { /* generic */ };

class MyChildObject:
    ctor():
        Bindable.ValueChanged += () =>  { /* non-generic */ };
```
Is it fine if `MyParentObject` receives the event before `MyChildObject`, even if `MyChildObject` bound to it first?

huoyaoyuan:
I see.
`GetBoundNonGenericCopy()` may also mess up the implementation of bound list.
Making a new class delegating non-generic events could get clean code and limited overhead.

smoogipoo:
A neat way to do this is to create a custom bindable type which exposes a non-generic action that is invoked from its own ValueChanged. Then you'd bind that bindable (via reflection) to the original type.

But yeah, the number of reflection calls just gets a little chunky so we kept it simple for now.

smoogipoo (approved):

PR #3210, opened 2018-08-12
Hit error judgements

zachllogan:
- Closes #3209 

---

Adds an option to enable Hit Error Judgements, which colors judgments based off of their hit error (perfect timing is white, and as hit error becomes earlier the judgement becomes bluer, as it becomes later the judgement becomes redder).

Issue #3209, opened 2018-08-12
Proposal for an additional Hit Error display method


zachllogan:
I find the current hit error display bar virtually unusable. During gameplay that is even remotely high level, one simply can't use it. Really, its only practical use for me is during replays, but even then I still find it a little difficult to understand.

My proposal is to integrate Hit Error feedback into hit judgments, as these are always near where the eye is looking.There are many ways this could be implemented. A couple of ideas:

1. Change the color, text, or skin element depending on whether the hit error is positive or negative. This could be having 2 skinable elements for each judgment, one for early and one for late, or for the current system (of text sprites), it could be appending or prepending a "+" or "-" to each judgement. Also a different color could be used for the sprites depending on whether they were early or late (so a total of 8 colors).

2. Set the color of the judgment based on the hit error using a gradient (one color for each end of the 50 hitwindow and one for the center)

I have taken the liberty of implementing my second idea, because it was simple and I believe it would be effective. I also created a couple of tests that show how it would look with overstreaming or understreaming. This is a video of these tests: https://youtu.be/C5-l92U_td4

I am willing to develop this myself and make whatever changes you believe should be made (though I can't promise I'll be quick).

peppy:
I think something like this could work very well! The judgement design is currently very placeholder, and will probably get a refresh in the near future (at which point adding this behaviour will probably become more visible to the user) but feel free to open a PR implementing this and we can take it from there.

iiSaLMaN:
> I think something like this could work very well! The judgement design is currently very placeholder, and will probably get a refresh in the near future (at which point adding this behaviour will probably become more visible to the user) but feel free to open a PR implementing this and we can take it from there.

Just for clarity, which system would work very well? Idea 1 or 2?

bdach:
Judging by the reason for closing the original PR submitted (#3210) I suppose that number 1, which was proposed in the duplicate as well, would be more feasible. That solution is nice in the way that it's backwards compatible - it won't break existing skins, if the extra sprites are missing then simply the normal ones could be used.

For extra style points (to save number 2's aesthetic) we might even want to consider alpha blending the sprites in the shader? For example it could be two sprites per judgement, early/late, which would be blended together based on the timed position within the hit window for that judgement.

peppy:
New skin able elements will not be added in `LegacySkin`. If this is implemented, it should be non-skinnable. And will require design.

AntonPetrochenko:
[...]

PR #7310, opened 2019-12-20
Display fast/slow alongside judges

Tina-otoge:
This PR adds a "Fast" or "Slow" text above judges when not hitting perfectly.

It integrates well in osu and mania rulesets, looks a bit weird in taiko, not completely sure if I could take a different approach for taiko.

Also I should probably need to add test cases for this?

Maybe this should also be made into a toggle in user options near hit error bar settings?


osu.Game/Rulesets/Judgements/JudgementResult.cs
```diff
@@ -30,6 +30,8 @@ public class JudgementResult
         [NotNull]
         public readonly Judgement Judgement;
 
+        public HitDetail Detail => TimeOffset < 0 ? HitDetail.Early : HitDetail.Late;
```
iiSaLMaN: What happens when hitting at perfect timing `TimeOffset = 0`?

osu.Game/Rulesets/Scoring/HitResult.cs
```diff
@@ -44,4 +44,13 @@ public enum HitResult
...
         Perfect,
     }
+
+    public enum HitDetail
+    {
+        [Description(@"Fast")]
+        Early,
+
+        [Description(@"Slow")]
+        Late,
```
iiSaLMaN: Pretty much unnecessary to add descriptions here, just name the enums `Fast` and `Late` then use `HitDetail.ToString()`.

frenzibyte (changes requested):

osu.Game/Rulesets/Scoring/HitResult.cs
```diff
@@ -44,4 +44,13 @@ public enum HitResult
         [Description(@"Perfect")]
         Perfect,
     }
+
+    public enum HitDetail
```
iiSaLMaN: Should have it's own `.cs` file.

osu.Game/Rulesets/Judgements/JudgementResult.cs
iiSaLMaN: xmldoc.

iiSaLMaN: Just realized that no detail text appears on `Great` judgement, although I advocate for adding a `Perfect` detail rather than having a false value.

skielred: It's hidden when `TimeOffset` is within the window of a Perfect or Great judgement, because the judgement text already gives the information that we hit perfectly.

skielred: Alright, actually done that at first but could not find a use-case where it would be used. If you think it's more fitting to have a third member in the Enum I'll be adding one

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -134,5 +146,32 @@ private Color4 judgementColour(HitResult judgement)
...
+        {
+            switch (result.Type)
+            {
+                case HitResult.Perfect:
+                case HitResult.Great:
+                    return "";
+                default:
+                    return result.Detail.GetDescription().ToUpperInvariant();
+            }
+        }
```
iiSaLMaN: What about `HitResult.Miss`?

iiSaLMaN:
Seems OK visually, the colours of `Fast` and `Slow` seems to be not fitting with the judgment result. How about Green for `Fast` and Yellow or Light Red for `Slow`?

skielred:
Blue for `Fast` and red for `Slow` are kinda standard across all rhythm games, sometime text is different, but the colors are blue and red in every rhythm game I can think of that has this feature. Also the contrast helps hinting your timing without actually having to read the text.
Haven't really played with other colors, I can try Light Red for `Slow`, but I believe we should stick to blue for `Fast`, maybe other hues of blue blends in better?

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -65,13 +66,24 @@ private void load(OsuColour colours)
...
+                        Text = Result.Type.GetDescription().ToUpperInvariant(),
+                        Font = OsuFont.Numeric.With(size: 20),
+                        Colour = judgementColour(Result.Type),
+                        Scale = new Vector2(0.85f, 1),
+                    }, confineMode: ConfineMode.NoScaling),
+                    new SkinnableDrawable(new GameplaySkinComponent<HitDetail>(Result.Detail), _ => JudgementDetailText = new OsuSpriteText
+                    {
+                        Text = detailText(Result),
+                        Font = OsuFont.Numeric.With(size: 15),
+                        Y = -20,
```
skielred: This -20 is kinda arbitrary, I may be missing a way to display this component above the other one nicely

iiSaLMaN:
> Haven't really played with other colors, I can try Light Red for `Slow`, but I believe we should stick to blue for `Fast`, maybe other hues of blue blends in better?

how about Cyan for `Fast`, fits well with light red I believe.

skielred:
Cyan and RedLight looks really nice! Thank you for the heads up

osu.Game/Rulesets/Scoring/HitDetail.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Rulesets.Scoring
+{
+    public enum HitDetail
```
swoolcock: I don’t think this really warrants its own enum, it could just be done directly from the `TimeOffset`.

Either way, `HitResultAccuracy` is a better name, “Detail” is very arbitrary.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
iiSaLMaN: you're probably looking for a `FillFlowContainer`, check its usages for understanding on how to use it.

swoolcock (commented):
Since this is a feature that affects gameplay, this PR needs:

* An accompanying proposal request issue
* Designs from the designer

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
VPeruS: FlowContainer with Direction = FillDirection.Vertical

GSculerlor:
The indicator overlapping the skinnable hit result https://streamable.com/sgyz6

skielred:
@GSculerlor Yeah that was the purpose of my question in https://github.com/ppy/osu/pull/7310#discussion_r360428851
I'll try using `FlowContainer` this should fix the overlapping issue

Weigsu:
### Problem
I feel this creates too much visual noise. Specially as the default skin is quite overwhelming already.
Also, if I understood correctly of how the new skins work I will be able to skin the HitResult (perfect, good, meh) and HitDetail (fast, slow) separately. This however doesn't offer enough flexibility for skinning in my opinion.

### My solution
My preferred way to skin this would be to leave only "Good", "Meh" and "Miss" texts. (and nothing on slider ends) and then color the text yellow if I'm going too fast and blue if too slow.
[Example of that (left is unchanged, right is as described above)](https://streamable.com/dxoxg)

### Possible implementation?
This can't be done with current implementation where I have to skin HitResult and HitDetail separately. And seems there is no standard way to have GameplaySkinComponent based on two values. I'm not quite sure how to do it.
Maybe combine HitResult and HitDetail in struct and implement ToString like that:
``` C#
struct MyHitStruct
{
    private HitResult HitResult;
    private HitDetail HitDetail;

    public MyHitStruct(HitResult hitResult, HitDetail hitDetail)
    {
        HitResult = hitResult;
        HitDetail = hitDetail;
    }

    public override string ToString()
    {
        return string.Join("/", HitResult.ToString(), HitDetail.ToString());
    }
}
// and then when used
new SkinnableDrawable(new GameplaySkinComponent<MyHitStruct>(new MyHitStruct(Result.Type, Detail)), ....)
```
or just add all possible combinations to HitResult enum, but this seems even worse.

### Motivation
Having the extra text was too much to process for me while playing and I just ignored it.  But when I color coated the HitResults it actually helped me to get into correct rhythm sometimes (not as much as I hoped though)

But the best effect this had was that I didn't feel the constant need to look at the Hit Error bars to see if I clicked too late or early (and then missing or losing acc because of that). This made the game experience feel better and "smoother/less abrasive?" for me.

Love this feature and I hope it will make it to lazer!

bdach:
@Weigzu Your proposal (or something extremely similar to it) has already been submitted as a PR once and rejected; see #3210 for rationale.

Weigsu:
> @Weigsu Your proposal (or something extremely similar to it) has already been submitted as a PR once and rejected; see #3210 for rationale.

This is very sad. I don't quite understand the reason for rejection though, peppy said in proposal that he liked the idea (so you're saying there is a chance? :) but in PR just that the colors shouldn't be applied directly to HitResult? Not a big issue

Just to be clear, the color thing I made example of was just one use case (of how I would use it) but the default skin would remain unchanged.
There could be skinnable object for every HitResult HitDetail combo like so:
 * `HitResult` - if the skin doesn't support fast/slow displaying or when it is disabled from options
 * `HitResult/HitDetail.Fast` - for every type of HitResult a skinnable object for when click was too early
 * `HitResult/HitDetail.Slow` - ... was too late
 * `HitResult/HitDetail.Flawles` - ...was on time and some HitResult Judgements are always flawless

Right now (in this PR) there are skinnable objects just for every HitResult and for every HitDetail but not for combinations of them.

I don't think this feature would be good addition to default skin, but it would be awesome to be able to apply custom skin elements to them  (if the skin author desires to).

bdach:
New legacy skin elements are also unlikely to happen ([see comment](https://github.com/ppy/osu/issues/3209#issuecomment-533942709)).

skielred:
Coloring the judge or adding a prefix or suffix (like + or -) are both options that were also discussed when I first talked about implementing this alongside my friends, and I think it's a good compromise for people who think it adds too much elements on screen and makes reading too complex.
It is already noticeable that people are bothered when there are too much elements on screen as most people hide Perfect/Greats from gameplay, most people want the less visual clutter possible.

However, adding an additional elements also allow compatibility with legacy skins by not modifying the behavior of the current judges element, while making it possible to make new skins that hide the judges and instead only display FAST or SLOW, both, or a "-" and "+" on the side of judges instead of "FAST" and "SLOW" above, etc.

As discussed before, I (or someone else) will have to submit a proposal request issue and discuss what works best for osu! in it.

My personal opinion is that it should be left as an option to the player or the skinner to decide how the "detail" / additional accuracy info should be displayed.

Either way, I think this is an important feature as reading the Hit Error Display bar is not intuitive at all while being aware of if we are hitting too early or too late is very much needed to adjust our typing while playing.

I'm completely willing to work again on this PR when we'll have discussed what would be the most fitting implementation for osu!.

Weigsu:
> New legacy skin elements are also unlikely to happen ([see comment](https://github.com/ppy/osu/issues/3209#issuecomment-533942709)).

I don't see any reason to add new legacy skin elements.

> Either way, I think this is an important feature as reading the Hit Error Display bar is not intuitive at all...

Yes, exactly. And it's hard to look at sides of the screen and keep playing :)

#### Clarification
I'll try to explain my issue with current implementation again.
If I wanted to hide "Great" judgments but show custom slow/fast elements on other judgments ("Meh", "Good") it would still display the slow/fast over "Great" even though "Great" itself is invisible. Also it would display it over misses etc.
Demonstrated in video in row "PR issue"

I just feel that if adding extra elements to Judgment  is the only way to skin, it won't be used very much. I made a video with two examples that I believe would be usable for gameplay. But what I think is often very different to what everyone else thinks. If it is so I can just play in dev mode no problem.

[Video with examples.](https://streamable.com/bnofn)

The two examples:
 * Color - Judgment text is colored based on fast/slow (yellow fast, blue slow in this case)
 * Text - Color is kept default but the text is changed to fast/slow (or +/- etc) Also the "Great" judgment would be hidden but it's there to show default colors.

#### In conclusion:
 * No new legacy skin elements added
 * Not supported in old osu, only in lazer
 * Optional for skinner to add
 * If skin is missing HitDetail (fast/slow) elements then this option will be disabled/hidden and HitResult skin is used
 * Not part of default lazer skin (unless someone figures out a way this could fit in default skin)
 
Also, here's a [osu community feature request](https://osu.ppy.sh/community/forums/topics/167932). Might be useful.

smoogipoo:
Personally quite like the "text" version in that example video. It could be adapted very well to more minimalistic skins that use dots - I imagine skins having up/down arrows coloured by the judgement instead of the fast/slow text.
Not a fan of the "flawless" though - empty text may work better, but would have to see how it feels during gameplay.

In general I think having two text lines is a bit cluttered.

osu.Game/Rulesets/Scoring/HitDetail.cs
LevFlavien: How about `HitTiming` or `HitOffset`?

skielred:
I updated my branch with current master, some stuffs that changed on master actually allowed rewriting some parts a bit cleaner.
This way if anyone want to still test or change things on this, at least it's up to date.

But feature-wise it's exactly the same as before and still suffer from the same problem as before.
The added fast/slow is still arbitrary put 20px above the judge, making it not very good visually using some skins as reported by https://github.com/ppy/osu/pull/7310#issuecomment-568040411 which is the top issue currently imo.

mcendu:
Is this still being worked on?

bdach:
given that the last commit was 11 months ago, i think the answer is "no". there is no point in bumping.

the most important part of this feature is going to be the visual design of the fast/slow display anyhow. the implementation is otherwise trivial. as i understand it this feature may come with future updates to the lazer default skins.

peppy:
A visual design for this has already been decided, although I'm not sure if it's exposed yet on figma. @arflyte can you advise?

What I can say, is that this PR won't be satisfactory for implementing the new design, as it requires more than three states. Also there's enough conflicts in here to make a rewrite easier, so let's close for now.

PR #7317, opened 2019-12-21
Implement ColourHitErrorMeter

### LINKED ISSUE #7292: Combined error and colour score meter
**Describe the new feature:** While one of the score meters from stable has been added, the other one hasn't. I suggest we change that. Here's a suggestion for a more modern-looking behaviour, though: instead of having the newest element scroll through the entire bar and go back to the beginning and start again, how about always having the topmost indicator be the most recent one, and with each new hit, the others would scroll down a spot (with a fluid animation) with the bottom-most one disappearing?
I would also like to take this chance to suggest an option to have both score meters displayed at the same time.

image

EVAST9919:
Closes https://github.com/ppy/osu/issues/7292

Video: https://streamable.com/subpe


osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
```diff
@@ -0,0 +1,108 @@
...
+using osuTK.Graphics;
+
+namespace osu.Game.Screens.Play.HUD.HitErrorMeters
+{
+    public class ColourHitErrorMeter : HitErrorMeter
+    {
+        private const int max_available_judgements = 20;
+
+        private readonly JudgementFlow judgementsFlow;
+        private readonly BindableList<(Color4 colour, JudgementResult result)> judgements = new BindableList<(Color4 colour, JudgementResult result)>();
```
peppy: curious on the reason for making a bindable list and a subcomponent. seems like unnecessary complexity.

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public override void OnNewJudgement(JudgementResult judgement)
+        {
+            judgements.Add((GetColourForHitResult(HitWindows.ResultFor(judgement.TimeOffset)), judgement));
+
+            if (judgements.Count > max_available_judgements)
+                judgements.RemoveAt(0);
+        }
+
+        private class JudgementFlow : Container<DrawableResult>
```
peppy: is there a reason you didn't use `FillFlow` for this?

peppy (changes requested):

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
EVAST9919: It's easier to maintain judgements capacity like this rather than counting drawables

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
EVAST9919: Just the first solution I came up with

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
peppy: i'm not sure i understand. you're still counting in the end.

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
peppy: you are reimplementing fillflow logic. try using a fillflow instead.

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
EVAST9919: I tried to make it as you suggested first, but this was not working correctly for high 7K mania difficulties. I tried to remove drawables using `Alpha == 1` for the last one, but it failed since there could be many of them fading out at the same time.
(This is what I was typing and then realized: for that reason I added `FinishTransforms` for each drawables on new judgement arrived)

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
EVAST9919: I guess, I should look into NotificationsOverlay for similar behavior

osu.Game/Screens/Play/HUD/HitErrorMeters/ColourHitErrorMeter.cs
peppy:
```csharp
DrawableResult
{
    public bool IsRemoved;

    void Remove()
    {
		IsRemoved = true;
        FadeOut(...);
        Expire();
    }
}

Children.Where(c => !c.IsRemoved)
```

is one simple way

GSculerlor:
This have different behavior compared with stable. In here the latest result will push the previous result while on stable it will stay on same position and then fading out

hiterrormeter

EVAST9919:
Ikr. Just implementing the behavior, explained in the issue

osu.Game/Configuration/ScoreMeterType.cs
```diff
@@ -18,5 +18,14 @@ public enum ScoreMeterType
...
         HitErrorBoth,
+
+        [Description("Colour (left)")]
+        ColourLeft,
+
+        [Description("Colour (right)")]
+        ColourRight,
+
+        [Description("Colour (both)")]
+        ColourBoth,
```
mcendu: Maybe split into "left score meter type" and "right score meter type"?

This would make it easy for one to have hit error on left and color on right.

peppy: The plan is to allow both on each side too. Not worth changing this enum here yet.

EVAST9919: Suggesting having two dropdowns: one for type and another one is for position. 
But still out of context of this pr

osu.Game.Tests/Visual/Gameplay/TestSceneHitErrorMeter.cs
```diff
@@ -91,17 +95,31 @@ private void recreateDisplay(HitWindows hitWindows, float overallDifficulty)
...
+                Origin = Anchor.CentreRight,
+                Margin = new MarginPadding { Right = 50 }
+            });
+
+            Add(colourMeter2 = new ColourHitErrorMeter(hitWindows)
+            {
+                Anchor = Anchor.CentreLeft,
+                Origin = Anchor.CentreLeft,
+                Margin = new MarginPadding { Left = 50 }
+            });
```
mcendu: Replace with `FillFlowContainer`?

mcendu (commented):
I suggest that we use two `FillFlowContainer`s, one for each side, to contain hit error meters. This can facilitate implementing the "Both" options.

peppy (approved):

PR #7318, opened 2019-12-21
Truncate rich presence status strings to 128 bytes

### LINKED ISSUE #7311: Discord rich presence throws exception on some beatmaps
https://osu.ppy.sh/beatmapsets/874178#osu/1827120
https://osu.ppy.sh/beatmapsets/925343#osu/1932924

Throws the following exception upon entering gameplay:

```
2019-12-20 15:46:57: DiscordRPC.Exceptions.StringOutOfRangeException: Length of string is out of range. Expected a value with a maximum length of 128
2019-12-20 15:46:57: at DiscordRPC.RichPresence.set_Details(String value)
2019-12-20 15:46:57: at osu.Desktop.DiscordRichPresence.updateStatus()
2019-12-20 15:46:57: at osu.Desktop.DiscordRichPresence.<load>b__10_4(ValueChangedEvent`1 _)
2019-12-20 15:46:57: at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
2019-12-20 15:46:57: at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
2019-12-20 15:46:57: at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
2019-12-20 15:46:57: at osu.Framework.Bindables.Bindable`1.set_Value(T value)
2019-12-20 15:46:57: at osu.Game.Screens.OsuScreen.updateActivity()
2019-12-20 15:46:57: at osu.Game.Screens.OsuScreen.set_Activity(UserActivity value)
2019-12-20 15:46:57: at osu.Game.Screens.OsuScreen.OnEntering(IScreen last)
2019-12-20 15:46:57: at osu.Game.Screens.Play.Player.OnEntering(IScreen last)
2019-12-20 15:46:57: at osu.Framework.Screens.ScreenStack.<>c__DisplayClass15_0.<Push>b__0(Drawable _)
2019-12-20 15:46:57: at osu.Framework.Graphics.Drawable.loadComplete()
2019-12-20 15:46:57: at osu.Framework.Graphics.Drawable.UpdateSubTree()
```

The game doesn't crash, `Player` finishes loading but does not continue playing the beatmap.

Game4all:
Closes #7311
`State` and `Detail` strings are now truncated to 128 bytes (as stated in the DiscordRPC docs), which should stop DiscordRichPresence from throwing exceptions

### DIFF
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -100,17 +100,23 @@ namespace osu.Desktop
 
         private static readonly int ellipsis_length = Encoding.UTF8.GetByteCount(new[] { '…' });
 
-        private string truncate(ReadOnlySpan<char> str)
+        private string truncate(string str)
         {
             if (Encoding.UTF8.GetByteCount(str) <= 128)
-                return new string(str);
+                return str;
+
+            ReadOnlyMemory<char> strMem = str.AsMemory();
 
             do
             {
-                str = str[..^1];
-            } while (Encoding.UTF8.GetByteCount(str) + ellipsis_length > 128);
+                strMem = strMem[..^1];
+            } while (Encoding.UTF8.GetByteCount(strMem.Span) + ellipsis_length > 128);
 
-            return new string(str) + '…';
+            return string.Create(strMem.Length + 1, strMem, (span, mem) =>
+            {
+                mem.Span.CopyTo(span);
+                span[^1] = '…';
+            });
         }
 
         private string getDetails(UserActivity activity)
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -98,17 +98,17 @@ namespace osu.Desktop
             client.SetPresence(presence);
         }
 
+        private static readonly int ellipsis_length = Encoding.UTF8.GetByteCount(new[] { '…' });
+
         private string truncate(ReadOnlySpan<char> str)
         {
             if (Encoding.UTF8.GetByteCount(str) <= 128)
                 return new string(str);
 
-            int ellipsisLength = Encoding.UTF8.GetByteCount(new[] { '…' });
-
             do
             {
                 str = str[..^1];
-            } while (Encoding.UTF8.GetByteCount(str) + ellipsisLength > 128);
+            } while (Encoding.UTF8.GetByteCount(str) + ellipsis_length > 128);
 
             return new string(str) + '…';
         }
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -1,7 +1,10 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Linq;
+using System.Text;
 using DiscordRPC;
+using DiscordRPC.Helper;
 using DiscordRPC.Message;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
@@ -78,8 +81,8 @@ namespace osu.Desktop
 
             if (status.Value is UserStatusOnline && activity.Value != null)
             {
-                presence.State = activity.Value.Status;
-                presence.Details = getDetails(activity.Value);
+                presence.State = truncate(activity.Value.Status);
+                presence.Details = truncate(getDetails(activity.Value));
             }
             else
             {
@@ -96,6 +99,8 @@ namespace osu.Desktop
 
             client.SetPresence(presence);
         }
+
+        private string truncate(string str) => str.WithinLength(128) ? str : new string (str.TakeWhile((c, i) => Encoding.UTF8.GetByteCount(str.Substring(0, i + 1)) <= 125).ToArray()) + '…'; //the ellipsis char is 3 bytes long in UTF8
 
         private string getDetails(UserActivity activity)
         {
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -1,10 +1,9 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Linq;
+using System;
 using System.Text;
 using DiscordRPC;
-using DiscordRPC.Helper;
 using DiscordRPC.Message;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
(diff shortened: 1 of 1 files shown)


smoogipoo:
128 bytes != 128 chars

E.g.:
`Console.WriteLine(System.Text.Encoding.UTF8.GetBytes(new string(Enumerable.Repeat('あ', 128).ToArray())).Length);`
Returns a length of 384 bytes.

HoLLy-HaCKeR:
How would this handle multi-byte characters? I think a more fool-proof way of doing this would be to convert individual characters until you get to 127/128 bytes, since you can also end on a character that's 2 or more bytes long.

Game4all:
> 
> 
> 128 bytes != 128 chars
> 
> E.g.:
> `Console.WriteLine(System.Text.Encoding.UTF8.GetBytes(new string(Enumerable.Repeat('あ', 128).ToArray())).Length);`
> Returns a length of 384 bytes.

I assumed that the input string was composed exclusively of ASCII chars (1 char = 1byte) because the stringified `BeatmapInfo` uses the romanised song title / artist name
(Guess I'll have to read up a bit about encodings)



> 
> 
> How would this handle multi-byte characters? I think a more fool-proof way of doing this would be to convert individual characters until you go over 127/128 bytes, since you can also end on a character that's 2 or more bytes long.
> 
> **EDIT**: The ellipsis character itself is also [3 bytes long in UTF8 encoding](https://apps.timwhitlock.info/unicode/inspect?s=%E2%80%A6).

So about 125 bytes of a UTF8 string + 3 bytes for the ellipsis character

peppy:
https://github.com/Lachee/discord-rpc-csharp/blob/13f18606f12fc667ff6222c6c46c8a2d370652fb/DiscordRPC/Helper/StringTools.cs#L40

is publicly exposed

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -97,6 +99,8 @@ private void updateStatus()
             client.SetPresence(presence);
         }
 
+        private string truncate(string str) => str.WithinLength(128) ? str : Encoding.UTF8.GetString(Encoding.UTF8.GetBytes(str), 0, 125) + '…'; //the ellipsis char is 3 bytes long in UTF8
```
mcendu:
```suggestion
        private string truncate(string str) => str.WithinLength(128) ? str : Encoding.UTF8.GetString(Encoding.UTF8.GetBytes(str), 0, 125) + '…'; // the ellipsis char is 3 bytes long in UTF8
```

peppy: i'm pretty sure this isn't what you want. you are potentially splitting a UTF8 character into pieces (not accepting all the bytes for it).

i'd highly suggest searching and using an existing solution rather than trying to reinvent the wheel. [this](https://stackoverflow.com/questions/1225052/best-way-to-shorten-utf8-string-based-on-byte-length) looks like a good starting point.

Game4all: went with a slighty modified version of the top answer of the post where the byte length is hardcoded to 125 (125 bytes + the 3 bytes of the ellipsis char)

Game4all: went with a slighty modified version of the top answer of the post where the byte length is hardcoded to 125 (125 bytes + the 3 bytes of the ellipsis char)

holly-hacker (commented):
Maybe you should just use an expression body, instead of forcing everything on 1 line using LINQ.

smoogipoo:
Thanks. I cleaned up the implementation for readability. Generally I find aggregates hard to read and it wasn't worth it in this scenario (also O(n^2) complexity).

smoogipoo (dismissed):

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -96,6 +98,21 @@ private void updateStatus()
             client.SetPresence(presence);
         }
 
+        private string truncate(ReadOnlySpan<char> str)
+        {
+            if (Encoding.UTF8.GetByteCount(str) <= 128)
+                return new string(str);
```
huoyaoyuan: The input should be a string to avoid extra allocation when not truncating. Use a `ReadOnlyMemory<char>` to truncate.

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -96,6 +98,21 @@ private void updateStatus()
...
+
+            int ellipsisLength = Encoding.UTF8.GetByteCount(new[] { '…' });
```
huoyaoyuan: Should be constant or static.

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -96,6 +98,21 @@ private void updateStatus()
...
+
+            do
+            {
+                str = str[..^1];
+            } while (Encoding.UTF8.GetByteCount(str) + ellipsisLength > 128);
+
+            return new string(str) + '…';
```
huoyaoyuan: Allocation here can be reduced:
assuming `ReadOnlyMemory<char> str`:
```csharp
return string.Create(str.Length + 1, str, (span, s) =>
{
    s.Span.Copyto(span);
    span[^1] = '…';
});
```

huoyaoyuan (commented):
How the allocations here _can_ be reduced.
Not strongly suggesting because of readability.

osu.Desktop/DiscordRichPresence.cs
huoyaoyuan: Additional note: I'm not sure about acceptance for the ellipsis character. It's 3 bytes in UTF8 like a common wide character, same with three dots `...`.
It's easy to input when using Chinese IME, and often used in Chinese text. However, in English text, I'm seeing three dots used much more.

peppy: it's more correct to use this character, but not easily accessible to normal users (no key binding). for automated cases i believe it should be preferred.

osu.Desktop/DiscordRichPresence.cs
smoogipoo: ReadOnlyMemory gets ugly. You lose any performance benefits as soon as you call `.Span` on it,`GetByteCount()` doesn't support `ReadOnlyMemory<>`, and you have to decorate callsites with `.AsMemory()`.

osu.Desktop/DiscordRichPresence.cs
smoogipoo: It's good to know, but I'm not interested in micro-optimisations like this in a low-overhead component in the first place. See other comment also - this also gets kinda ugly due to `ReadOnlyMemory<>` shortcomings.

osu.Desktop/DiscordRichPresence.cs
huoyaoyuan: `ReadOnlyMemory<T>.Span` has 128 bytes in asm. Really surprises me.
Since `GetByteCount` supports string, it should be better assigning the span after the early return route.

smoogipoo: Fyi the `.Span` inside the loop will cause the same runaway condition. Here's a benchmark I've done:

|         Method | Length |        Mean |      Error |     StdDev | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|--------------- |------- |------------:|-----------:|-----------:|------------:|------------:|------------:|--------------------:|
|   TruncateSpan |      1 |    23.66 ns |  0.2237 ns |  0.2093 ns |      0.0076 |           - |           - |                24 B |
| TruncateMemory |      1 |    17.61 ns |  0.2475 ns |  0.2316 ns |           - |           - |           - |                   - |
|   TruncateSpan |    128 |    51.66 ns |  1.0912 ns |  1.7929 ns |      0.0892 |           - |           - |               280 B |
| TruncateMemory |    128 |    25.07 ns |  0.2214 ns |  0.2071 ns |           - |           - |           - |                   - |
|   TruncateSpan |    200 | 1,858.51 ns | 16.5747 ns | 15.5040 ns |      0.1755 |           - |           - |               552 B |
| TruncateMemory |    200 | 1,981.22 ns | 22.4968 ns | 19.9428 ns |      0.0877 |           - |           - |               280 B |

peppy (approved):

PR #7328, opened 2019-12-23
Simplify/improve beatmap conversion precheck

smoogipoo:
1. It was too strict - mania beatmaps should "convert" to mania beatmaps, even though they don't implement `IHasXPosition`.
2. It was pretty confusing because `ValidConversionTypes` to me implies "i can convert any of these types" not "i can only convert hitobjects implementing all these types".


huoyaoyuan:
Why not making `CanConvert` virtual on hit objects? Any reason not to centralize the `All` call?

```csharp
bool CanConvert => Beatmap.HitObjects.All(h => CanConvert(h));

protected virtual bool CanConvertObject(HitObject hitObject) => true;
```

InspectCode 2019.3 can handle default interface methods.

smoogipoo:
Because doing so requires having already-converted hitobjects (e.g. mania's `Note`), which the converter does not have.

peppy (approved):

PR #7329, opened 2019-12-23
Fix mania hold notes not handling input correctly

### LINKED ISSUE #7309: Maina long hold note wired behaviour.
Hold note count as miss at the end regardless you release early or late, it only happen to the long hold notes, but suddenly release and press back the hold note (red cross appears but no combo break)fix it.

redcross no break:
ezgif-4-deba9ee83509

release at the end but_break:
ezgif-4-c026985ee60f

smoogipoo:
Fixes https://github.com/ppy/osu/issues/7309

Prereqs:
- [x] https://github.com/ppy/osu/pull/7328 (for testing purposes)

This is pretty much a complete rewrite of how mania hold-note input works. There were issues with how the tail note was being judged that go beyond the issue above, see the following test failures:

image

Input/judgements now matches scorev2.


peppy (approved):
looks sane

PR #7331, opened 2019-12-23
Make the Easy mod's extra life count customizable

### LINKED ISSUE #7327: Easy mod should have adjustable life count
Should be able to pick between a sane range, e.g. 0-10.

Game4all:
Closes #7327

This makes the number of allowed retries with the easy mod customizable.

### DIFF
--- a/osu.Game/Rulesets/Mods/ModEasy.cs
+++ b/osu.Game/Rulesets/Mods/ModEasy.cs
@@ -5,6 +5,7 @@ using System;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Beatmaps;
+using osu.Game.Configuration;
 using osu.Game.Graphics;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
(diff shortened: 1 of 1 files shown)


osu.Game/Rulesets/Mods/ModEasy.cs
```diff
@@ -21,7 +22,14 @@ public abstract class ModEasy : Mod, IApplicableToDifficulty, IApplicableFailOve
         public override bool Ranked => true;
         public override Type[] IncompatibleMods => new[] { typeof(ModHardRock) };
 
-        private int retries = 2;
+        [SettingSource("Extra Life count", "Number of extra lives")]
```
Game4all: Thoughts on the option name? Went with "extra life count" there as it sounded better in my head than "Number of allowed retries / attempts"

smoogipoo: Maybe just "Extra Lives"?

osu.Game/Rulesets/Mods/ModEasy.cs
```diff
@@ -21,7 +22,14 @@ public abstract class ModEasy : Mod, IApplicableToDifficulty, IApplicableFailOve
...
+        public Bindable<int> Retries { get; } = new BindableInt(2)
+        {
+            MinValue = 0,
+            MaxValue = 10
+        };
+
+        private int retries;
```
smoogipoo: Is this required? / any reason for not using the bindable directly?

Game4all: Decrementing the bindable directly would cause it to set the customized allowed retries count to 0  on the next game you start, thus the existence of this variable which stores the allowed number of retries in the current game

smoogipoo: Ugh, i'm still not happy with the bindable references being copied across... Alright.

peppy (approved):

hbnrmx:
should this not affect the score multiplier?

smoogipoo:
no

PR #7332, opened 2019-12-23
Disable automatic retry of failed discord RPC connections

peppy:
It seems like the overhead from retries was affecting some low-spec'd users on osu-stable. Disabling reconnect logic for now as a safety measure, pending further investigation.

### DIFF
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -43,6 +43,10 @@ namespace osu.Desktop
             };
 
             client.OnReady += onReady;
+
+            // safety measure for now, until we performance test / improve backoff for failed connections.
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

Ushimi:
Hi!
Im an osu user!
I have been playing this game for many months!
This morning, I encountered a big problem.

When i went into my osu lazer, I saw that i was not logged in.
I changed my problem but the problem still occured.
I went into my osu direct and it was blank.
I went into the chat and it was blank. 
And the multiplayer rooms were blank too! 
Please help! Ive been practicing all week to get a high score :(.
I think it might be an internet problem.
I'll try send the files later.

peppy:
you are in the wrong place.

Ushimi:
sorry!

PR #7334, opened 2019-12-23
Implement Mirror mod for Osu and Catch

### LINKED ISSUE #6954: Add Mirror Mod for osu! standard
**Describe the new feature:** Title says it all. Flipping the map horizontally can not just help players remove mindblocks on some maps, it's also neat for left-handed players (like me :]) since there's a number of maps that are mapped with right-handed players in mind. It would remove "unwanted" difficulty for lefties.
Potential problem is that maps that are supposed to be hard for the right hand (top left corner) could easily be bypassed by using the mod.

**Proposal designs of the feature:** I'm not designer at all, but using the mania mirror mod icon (as seen on the website) and replacing the lines with circles should look pretty good. Multiplier should be 1x since it literally does nothing but flipping the map. 
DKWYPGp

_Note that this is my first proposal so please tell me if I did something wrong_

huoyaoyuan:
Resolves #6954 .
The implementation is modified from `OsuModHardRock`. I'll work on a separated PR to make a shared implementation.


huoyaoyuan:
`XOffset` is affecting `CatchHitObject`.
Should I change
```csharp
        public float X
        {
            get => x + XOffset;
            set => x = value;
        }
```
to
```csharp
        public float X
        {
            get => x + XOffset;
            set => x = value - XOffset;
        }
```
?

smoogipoo:
That sounds fair for now.

huoyaoyuan:
Things getting a bit complicated about catch: the position from the beatmap is not identical to the position effective during gameplay. This may prevent further code sharing. `IHasXPosition` is not enough for catch.

Current implementation should make the applied order of `CatchModHardRock` and `CatchModMirror` arbitary.

smoogipoo:
`XOffset` shouldn't be included into `X`. That's only supposed to be a temporary thing to reduce implementation complexity during the push to make catch diffcalc identical to stable.

It should be provided as a separate property, like `GameplayX` or something.

huoyaoyuan:
> The mod would then invert both X and XOffset separately.

This requires Mirror to be applied after HardRock. To be order agnostic, the Mirror mod can only set a flag.

huoyaoyuan:
After finding all references, the places to distinguish `X` and `GameplayX` aren't so much.
Should I separate it into another PR? Does it need tests for offset working correctly?

huoyaoyuan:
It builds and passes unit tests now. Not sure if the design is expected.

osu.Game.Rulesets.Osu/Mods/OsuModMirror.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            osuObject.Position = new Vector2(OsuPlayfield.BASE_SIZE.X - osuObject.Position.X, osuObject.Y);
+
+            if (!(hitObject is Slider slider))
+                return;
+
+            slider.NestedHitObjects.OfType<SliderTick>().ForEach(h => h.Position = new Vector2(OsuPlayfield.BASE_SIZE.X - h.Position.X, h.Position.Y));
+            slider.NestedHitObjects.OfType<SliderRepeat>().ForEach(h => h.Position = new Vector2(OsuPlayfield.BASE_SIZE.X - h.Position.X, h.Position.Y));
+
+            foreach (var point in slider.Path.ControlPoints)
```
smoogipoo, line 28: These changes should be batched (there's been changes to `OsuModHardRock` lately, take a look for reference).

osu.Game.Rulesets.Catch/Mods/CatchModMirror.cs
```diff
@@ -0,0 +1,34 @@
...
+
+            foreach (var nested in catchObject.NestedHitObjects.Cast<CatchHitObject>())
+            {
+                nested.X = 1 - nested.X;
+                nested.XOffsetReversed = true;
+            }
+
+            if (catchObject is JuiceStream juiceStream)
+            {
+                foreach (var point in juiceStream.Path.ControlPoints)
```
smoogipoo, line 29: Same here, batch these changes.

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
```diff
@@ -15,19 +15,20 @@ public abstract class CatchHitObject : HitObject, IHasXPosition, IHasComboInform
...
 
         /// <summary>
         /// A random offset applied to <see cref="X"/>, set by the <see cref="CatchBeatmapProcessor"/>.
         /// </summary>
         internal float XOffset { get; set; }
 
+        /// <summary>
+        /// The effective X position during gameplay, applying with HardRock offset.
+        /// </summary>
+        public float GameplayX => XOffsetReversed ? X - XOffset : X + XOffset;
```
smoogipoo, line 30: Given the potential impact this can have and that osu!lazer is used for live diffcalc, I'd like this to have an HR diffcalc test.
See https://github.com/ppy/osu/wiki/Generating-osu!stable-conversion-mappings for steps how to do this (use the `diffcalc-test` beatmap already in the tests). Select the HR mod from stable and it'll generate the relevant mappings.

smoogipoo (changes requested):

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
huoyaoyuan: Created the test a bit tricky:
Create test case with HR in master to get correct SR in master, then paste it in this branch.

Surprisingly, HR gets the correct SR, but HRMR doesn't get the same,
```
Expected: 4.4882724679198711d
But was:  4.4848455546932069d
```

Investigating.

huoyaoyuan: For mirror only:
```
Expected: 4.050601681491468d
But was:  4.0506016929676321d
```

Seems like difficulty calculator itself having precision issue on this.

smoogipoo: MR diffcalc doesn't really matter, only interested in HR so it doesn't break existing stuff.

huoyaoyuan: Negate the X position in difficulty calculator:
None fails, MR passes, meaning it's a precision issue
HR fails, but HRMR cannot pass, meaning something is wrong in the interaction between HR and MR.

huoyaoyuan:
The interaction is correct now, but looks so ugly and fragile...

peppy:
@huoyaoyuan going to close this one for now. We've generally come to the agreement that mod PRs shouldn't be modifying anything outside of the mod class itself. This one is touching way too many locations. In addition it's conflicting in a lot of places to the point reimplementing from scratch seems like the simpler option.

`CatchHitObject` now has an `OriginalXBindable` which may be of use to you in the implementation (I haven't read into the details). It might be a good opportunity to take a fresh look at this if you're still interested in making it happen.

huoyaoyuan:
It's the right way to make the hit object more abstract instead of special case everywhere. If someone has interest in this, they can reuse some part of this PR.

PR #7336, opened 2019-12-24
Fix databased RulesetInfo not passed into an instance

### LINKED ISSUE #7335: RulesetInfo ID parameter gets passed incorrectly
regressed from https://github.com/ppy/osu/pull/7256/

Even though from the database you can clearly see that it has a set id, it however does not get passed when creating the Ruleset. therefore `CreateConfig()` does not get called

database
createconfig()

Though when debugging from RulesetStore it shows that it _does_ have an id.
rulesetstore

**osu!lazer version:** 2019.1219.0

frenzibyte:
Closes #7335

### DIFF
--- a/osu.Game/Rulesets/Ruleset.cs
+++ b/osu.Game/Rulesets/Ruleset.cs
@@ -26,7 +26,7 @@ namespace osu.Game.Rulesets
 {
     public abstract class Ruleset
     {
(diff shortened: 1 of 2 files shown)


peppy:
Disagree with this solution as it feels super round-about and hard to conceptually understand.

iiSaLMaN:
Closing in favour of #7338

PR #7343, opened 2019-12-25
Improve performance

greg6775:
- Adds .ConfigureAwait(false) to improve performance
- Remove GC.Collect() to improve performance


peppy:
- The GC collect is part of a test and should not be removed. The intention is to collect.
- Please cite the performance improvement of using `ConfigureAwait(false)` (and please read into what it does and where it serves a purpose, because I believe you will find it is not relevant for our usage).

peppy:
Closing this one for now for the stated reasons. Please feel free to follow up if you're unsure about the rationale, or disagree.

PR #7347, opened 2019-12-25
Fix sample lookup not working correctly for custom rulesets

peppy:
Needs to be tested; came across this potential flaw when updating template projects.


peppy:
Has been tested.

smoogipoo (approved):

PR #7351, opened 2019-12-25
Fix download manager potentially not handling cancel requests properly

frenzibyte:
- Fixes `DownloadableArchiveModelManager` not handling case where a request is cancelled outside of the notification `CancelRequested` event. (i.e `DownloadManager.GetExistingDownload(model).Cancel()`)

Alternative to / Closes #7297.


bdach:
No tests?

iiSaLMaN:
Seems very complex to write one, will check once I can...

peppy (approved):

PR #7354, opened 2019-12-26
Fix nub position getting offset by size changes

smoogipoo:
Since `RangePadding` is an absolute value, relative positioning still encounters the same issue.

This may be an intermediate fix - we may have to consider framework-side changes if we want this to be fixed in the general case (making `RangePadding` a relative value).


smoogipoo:
Closing in favour of a better solution.

PR #7358, opened 2019-12-26
Separate health and score processing components

smoogipoo:
There shouldn't be any changes in this PR. Parts common to both score and health processing have been abstracted into `JudgementProcessor`.

`ScoreProcessor` handles all score-related processing (including completions).
`HealthProcessor` handles all health-related processing (including fails).

For ruleset creators, generally all they'll have to do is move most code from their custom `ScoreProcessor` into the new `HealthProcessor` class. It shares the same interface as `ScoreProcessor` (e.g. beatmap provided via ctor), which will change after this.


osu.Game.Rulesets.Mania/Scoring/ManiaHealthProcessor.cs
```diff
@@ -0,0 +1,69 @@
...
+{
+    public class ManiaHealthProcessor : HealthProcessor
+    {
+        /// <summary>
+        /// The hit HP multiplier at OD = 0.
+        /// </summary>
+        private const double hp_multiplier_min = 0.75;
+
+        /// <summary>
+        /// The hit HP multiplier at OD = 0.
```
huoyaoyuan: fix the docs together with copying?

smoogipoo: What part of it am I fixing, the spelling? This is all going to be removed in a future PR anyway, so...

huoyaoyuan: The OD value were wrong in the original ScoreProcessor(should be 5 here). Sure to keep it if it'll be removed.

smoogipoo: Will keep it for now. See: https://github.com/ppy/osu/pull/7361 where this entire class has been removed.

peppy (approved):

PR #7361, opened 2019-12-26
Implement HP drain

### LINKED ISSUE #5946: Passive HP drain implementation
Health drain is not yet implemented but should be (to match stable).

smoogipoo:
Resolves https://github.com/ppy/osu/issues/5946

Prereqs:
- [x] https://github.com/ppy/osu/pull/7358

Taiko still uses its own "accumulating" health processor, but maybe that'll change in the future too.

This requires the processors to be added to the draw hierarchy. Currently, this is achieved by moving the overlays container of `DrawableRuleset` into the frame stability container and adding the processors to that container. As discussed, a future refactoring will move the processors into `DrawableRuleset` itself (although still be constructed via `Ruleset`).

In terms of rulesets, the breaking change is that their judgement processors don't accept a beatmap via ctor anymore, and is rather given in `ApplyBeatmap()`.


bdach (commented):
I can't find much fault with this after scrolling through, albeit I mostly took the actual drain algorithm as unquestionable gospel since you probably know better what is actually in stable. The only thing I sort of dislike is the `healthIncreases` null/non-null swap in the simulation pass but it's not really anything major.

iiSaLMaN:
I think this closes https://github.com/ppy/osu/issues/5946.

smoogipoo:
@bdach fyi: this doesn't follow stable's algorithm. After discussions we've settled on trying to define HP a bit more formally than stable, as the minimum percentage of HP reached during gameplay for a perfect play.

The concept and execution is similar, however stable only adjusts in one direction - decreasing drain to make it so the player can't ever fail.

Will remove the null/non-null swap, probably best to until we deem it's an issue.

osu.Game.Tests/Gameplay/TestSceneDrainingHealthProcessor.cs
```diff
@@ -0,0 +1,149 @@
...
+{
+    [HeadlessTest]
+    public class TestSceneDrainingHealthProcessor : OsuTestScene
+    {
+        private Bindable<bool> breakTime;
+        private HealthProcessor processor;
+        private ManualClock clock;
+
+        [Test]
+        public void TestInitialHealthStartsAtOne()
```
peppy: Please add tests covering rewind support

peppy:
I changed the drain to start from the first hitobject's starttime rather than `GameplayStartTime` (more closely matches stable). Seems good apart for missing tests.

peppy (approved):

PR #7365, opened 2019-12-26
Use mania-specific judgement sprites for mania

### LINKED ISSUE #7324: Mania should use mania-specific judgement sprites

mcendu:
Fixes #7324

Prereqs:
- [x] ppy/osu-resources#83

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
@@ -25,28 +25,33 @@ namespace osu.Game.Rulesets.Mania.Skinning
             switch (component)
             {
                 case GameplaySkinComponent<HitResult> resultComponent:
-                    switch (resultComponent.Component)
-                    {
-                        case HitResult.Miss:
-                            return this.GetAnimation("mania-hit0", true, true);
+                    return getResult(resultComponent);
+            }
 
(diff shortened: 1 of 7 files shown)


mcendu (commented):
Legacy-like animation still needs to be implemented.

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
```diff
@@ -0,0 +1,62 @@
...
+
+                        case HitResult.Good:
+                            return this.GetAnimation("mania-hit200", true, true);
+
+                        case HitResult.Great:
+                            return this.GetAnimation("mania-hit300", true, true);
+
+                        case HitResult.Perfect:
+                            return this.GetAnimation("mania-hit300g", true, true);
+                    }
```
bdach: Nitpicky, but maybe have this as a separate method?

bdach (commented):
Source itself looks reasonable to me.

peppy:
I think this is fine as a first step. Animation speed isn't correct and animation doesn't match 1:1, but this can be done as a follow-up.

peppy (dismissed):

peppy (approved):

peppy:
I also think the size needs adjustment. Will need an accuracy pass, comparing playfield with stable (via screenshots or video).

PR #7366, opened 2019-12-26
Update overlay headers layout in line with the new web design

EVAST9919:
Depends on:
- [x] https://github.com/ppy/osu/pull/7350
- [x] Resourses bump

Control updates coming with the future pr's.

Screenshot_3
Screenshot_4

### DIFF
--- a/osu.Game/Overlays/OverlayHeader.cs
+++ b/osu.Game/Overlays/OverlayHeader.cs
@@ -78,6 +78,7 @@ namespace osu.Game.Overlays
                     {
                         RelativeSizeAxes = Axes.X,
                         AutoSizeAxes = Axes.Y,
+                        Depth = -float.MaxValue,
                         Children = new Drawable[]
                         {
                             controlBackground = new Box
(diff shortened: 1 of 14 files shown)


osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -3,67 +3,109 @@
...
+            });
+
+            if (CreateContent != null)
+                flow.Add(CreateContent);
         }
 
         protected abstract Drawable CreateBackground();
 
-        protected abstract Drawable CreateContent();
+        protected virtual Drawable CreateContent => null;
```
peppy: i'm so confused. why is this no longer a method?

osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -3,67 +3,109 @@
...
-                {
-                    Margin = new MarginPadding { Top = cover_height },
-                    RelativeSizeAxes = Axes.X,
-                    AutoSizeAxes = Axes.Y,
-                    Child = CreateContent()
                 }
-            };
+            });
+
+            if (CreateContent != null)
```
peppy: you are calling `CreateContent` twice. that will construct it twice.

peppy (changes requested):

smoogipoo (dismissed):

smoogipoo (approved):

PR #7367, opened 2019-12-26
Adjust scrolling container object lifetime to account for origin choice

### LINKED ISSUE #6889: Taiko notes appear lately on screen
**Describe the bug:** 

Taiko notes appear (fade in) when their *center* reaches the edge of the field, in contrast to stable where the notes appear entirely.

Lazer: https://streamable.com/1sc9v
Stable: https://streamable.com/47c2v

Slow motion effect was applied by me, not the game.

**osu!lazer version:** 2019.1113.0

bdach:
Resolves #6889.

# Summary

Visual inspection of taiko gameplay has shown that hitobjects appeared on screen only when the origin of the hitobject came into the bounds of the screen, instead of appearing when any visible part of the hitobject came into the screen bounds.

This behaviour was due to lifetime calculation being based on the origin of the hitobject and not taking into account the actual object dimensions. Adjust the lifetime start of the hitobject by subtracting the time needed to show the part of the hitobject that should already be visible on screen when the origin comes into frame.

# Remarks

To showcase the effects of this change I modified the `ScrollingHitObjects` test scene to have scrolling containers for all four directions. Here are slowed-down videos that display how the scrolling looks:

* [before commit `193e41f`](https://streamable.com/vh5z8),
* [after commit `193e41f`](https://streamable.com/250x6).

The choice of `new Vector2(75 / 4.0f)` as the origin was intentionally asymmetrical to catch out any potential bugs I might have made when introducing this change. I wasn't sure if I wanted to introduce actual assertions since it's mostly a visual change, but I can add some if it is deemed necessary.

Additionally, [here's a video taken inside the `TaikoPlayfield` test scene](https://streamable.com/c1ucu) showcasing the effect in actual gameplay as well.


bdach:
Incidental CI test failure addressed at #7368.

peppy:
Code looks solid, but I don't know how to read the test scene. With the default settings (not the ones in your video) it's hard to tell the issue is fixed, and there's no way to set the correct settings to ensure we are 100% accurate with the new implementation.

Maybe add a new test method / step in that test scene which sets the settings in such a way that this can be examined.

bdach:
Should be better now, after some mishaps. I've split off the old controls into one test and added the settings used to reproduce the bug in the other.

Unfortunately the commit order now makes it so you can't just check out the modified tests before the fix, but I didn't want to interactive rebase & force push just for that. I did check that the tests should still show the wrong lifetime by cherrypicking the fix out.

peppy (approved):
looks good!

PR #7371, opened 2019-12-27
Lock user adjusted difficulty settings when changing beatmap

### LINKED ISSUE #7345: An "override" option for Difficulty Adjust mod
**Describe the new feature:** 
Currently, the mod updates all it's parameters every time the beatmap changes, but I could see this being a problem for people that want to practice, for example.. AR8 only. They would have to change the AR slider every time they change the map. This could be implemented by having an "override" checkbox in the Customizations tab. 
Let me know if this is not the direction the project is heading towards, and I'll close this.

peppy:
Closes https://github.com/ppy/osu/issues/7345

### DIFF
--- a/osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
+++ b/osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
@@ -5,6 +5,7 @@ using osu.Game.Beatmaps;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics.Sprites;
 using System;
+using System.Collections.Generic;
 using osu.Game.Configuration;
 
 namespace osu.Game.Rulesets.Mods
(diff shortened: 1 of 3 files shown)


osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -64,8 +65,31 @@ public void ReadFromDifficulty(BeatmapDifficulty difficulty)
...
+        {
+            bindable.UnbindEvents();
+
+            userChangedSettings.TryAdd(bindable, false);
+
+            // users generally choose a difficulty setting and want it to stick across multiple beatmap changes.
+            // we only want to value transfer if the user hasn't changed the value previously.
+            if (!userChangedSettings[bindable])
+            {
+                bindable.Value = bindable.Default = beatmapDefault;
```
smoogipoo: Probably want the default to be set in any case? I'm imagining someone selecting a new beatmap and deciding they don't want their custom settings, so they just click the default value indicator.

smoogipoo (changes requested):

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
peppy: oops, good call.

smoogipoo (approved):
👍

PR #7380, opened 2019-12-28
Update overlay headers controls in line with the web design

EVAST9919:
Screenshot_2
Screenshot_3


osu.Game/Graphics/UserInterface/BreadcrumbControl.cs
```diff
@@ -15,17 +15,18 @@ namespace osu.Game.Graphics.UserInterface
     public class BreadcrumbControl<T> : OsuTabControl<T>
     {
         private const float padding = 10;
-        private const float item_chevron_size = 10;
+        private readonly float itemChevronSize;
```
peppy: just make it virtual with a default like we do everywhere else. then override it if/when needed.

peppy (changes requested):

osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -5,14 +5,16 @@
...
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.UserInterface;
 using osu.Game.Graphics.UserInterface;
 using osuTK.Graphics;
 
 namespace osu.Game.Overlays
 {
-    public abstract class OverlayHeader : Container
+    public abstract class OverlayHeader<TModel> : Container
```
peppy: what on earth is `TModel`? This needs documentation and potentially better naming.

peppy: on top of that, nothing uses this class. is there a point to splitting it out as you have done?

peppy (changes requested):

osu.Game/Overlays/OverlayHeader.cs
EVAST9919: There's a `BreadcrumbControlOverlayHeader` and `TabControlOverlayHeader`, which use this class. (If it what you mean)
I'll add xmldoc

osu.Game/Graphics/UserInterface/BreadcrumbControl.cs
```diff
@@ -15,14 +15,15 @@ namespace osu.Game.Graphics.UserInterface
     public class BreadcrumbControl<T> : OsuTabControl<T>
     {
         private const float padding = 10;
-        private const float item_chevron_size = 10;
 
-        protected override TabItem<T> CreateTabItem(T value) => new BreadcrumbTabItem(value)
+        protected virtual float ItemChevronSize => 10;
```
peppy: why is this even at `BreadcrumbControl`? The only usage has you passing it to a class you are constructing. So why not just specify it within that class?

peppy (changes requested):

osu.Game/Graphics/UserInterface/BreadcrumbControl.cs
EVAST9919: Strip size (which is in `BreadcrumbControl`) depends on the icon size

EVAST9919: That strip is not preset in `OverlayBreadcrumbControl`, but might affect possible future implementations

peppy: I've fixed this for you. But then in continued review, it doesn't even match the god damn design. And this part isn't required for it to match.

Care to put in a bit more effort? There should be no underline on the breadcrumb display.

EVAST9919: Once you told me that ingame elements should be consistent. I mean, when once I tried to implement custom links, which matched web design - you told me to use already implemented. That's why I didn't touch this control much.

peppy: If they are to be consistent, the line should be removed from all instances since at the moment it looks like a bug. But at that point why even have two classes?

peppy:
I've brought this to a rough point it can be merged.

Something to consider going forward is that the legacy tab controls all should be updated to the latest design, removing the need for a "strip" to ever exist in the item.

peppy (approved):

PR #7381, opened 2019-12-28
Move mania stage hint to its own class

mcendu:


peppy:
Is this a prereq for skinning support?

mcendu:
Mostly yes

osu.Game.Rulesets.Mania/UI/Components/StageHint.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using osu.Framework.Graphics.Containers;
+using osu.Game.Graphics;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public abstract class StageHint : CompositeDrawable, IHasAccentColour
+    {
+        public abstract Color4 AccentColour { get; set; }
```
mcendu: Change to virtual as legacy stage hints wouldn't be colored.

osu.Game.Rulesets.Mania/UI/Components/StageHint.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics.Containers;
+using osu.Game.Graphics;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public abstract class StageHint : CompositeDrawable, IHasAccentColour
```
peppy: not sure this base implementation need exist if it's this empty. if accent colour is the only usage, you can just null-cast is to pass in the colour:

```csharp
if (hitTarget is IhasAccentColour colouredHitTarget)
	colouredHitTarget.AccentColour = AccentColour;
```

peppy (changes requested):

peppy:
Not sure about the "stage hint" terminology. Where does this come from?

GSculerlor:
If this PR is targeting for skin support, then it should be referring to `mania-stage-hint` (source from [wiki](https://osu.ppy.sh/help/wiki/Skinning/osu%21mania)). But on stable, `mania-stage-hint` is for full-width playfield, not per column

osu.Game.Rulesets.Mania/UI/Components/ColumnHitObjectArea.cs
```diff
@@ -78,21 +60,86 @@ public Color4 AccentColour
...
+
+            [BackgroundDependencyLoader]
+            private void load(IScrollingInfo scrollingInfo)
+            {
+                direction.BindTo(scrollingInfo.Direction);
+                direction.BindValueChanged(dir =>
+                {
+                    Anchor anchor = dir.NewValue == ScrollingDirection.Up ? Anchor.TopLeft : Anchor.BottomLeft;
+
+                    hitTargetBar.Anchor = hitTargetBar.Origin = anchor;
```
peppy: not sure doing the full bindable flow here is required (could just pass on the anchor from self to children) but should be fine for now.

peppy (approved):

PR #7382, opened 2019-12-28
Update star rating in song select statistics when changing mods/mod settings

huoyaoyuan:
Thanks to #7303 , the implementation becomes very simple.
HardRock:
image
DoubleTime:
image

But, I'm afraid we are on the wrong way of beatmap details processing. There should be a centralized "Playable state of beatmap", in contrast of "original beatmap", and all the displaying should simply read from the playable state.

Also, there needs a redesign of beatmap info UI.

### DIFF
--- a/osu.Game/Screens/Select/Details/AdvancedStats.cs
+++ b/osu.Game/Screens/Select/Details/AdvancedStats.cs
@@ -88,7 +88,7 @@ namespace osu.Game.Screens.Select.Details
 
             ScheduledDelegate debounce = null;
 
-            foreach (var mod in mods.NewValue.OfType<IApplicableToDifficulty>())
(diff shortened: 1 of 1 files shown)


peppy:
At very least, the colouring needs to change for SR. Orange on red is not a good look.

huoyaoyuan:
image
Current implementation is not flexible enough. EZ makes the blue into green.

EVAST9919:
No longer required after https://github.com/ppy/osu/pull/9637 merge

huoyaoyuan:
Sure. But I wish to be mentioned when there's something going to replace my work, and I can take part in the discussion.

peppy:
does this have functionality not in master still in terms of display?

smoogipoo:
It doesn't.

@huoyaoyuan sorry about that - I forgot this PR existed.

PR #7388, opened 2019-12-29
Fix crashing TestSceneMedalOverlay

EVAST9919:
Also fixes the scale of the medals since they became 2x web-side


peppy:
Did you test this? It changes the display logic.

EVAST9919:
Not sure which part are you referring to

peppy:
You're allowing the drawable to complete loading before the medal does. The animation starts potentially before the medal is visible, as a result.

EVAST9919:
I pushed a possible fix, but not sure why `Show()` can't be triggered. (Moving `Show()` out of the callback works)

peppy:
`LoadComponentAsync` schedules the callback to self. In your case, self is not present so will not display. I've updated this to work as you'd expect.

peppy (approved):

PR #7389, opened 2019-12-29
Fix footer button content not correctly being centered

### LINKED ISSUE #6953: Mods button content is not centred when there are mods selected
When there are mods selected, the right margin becomes shorter, making the contents look off-centre.

Untitled

vruss:
Fixes #6953

I fixed it by applying a right margin to the entire FillFlowContainer instead of just the multiplier text.
Please let me know if this is not the intended way of solving these kinds of issues. 

With multiplier
Without multiplier


peppy:
image
 Doesn't look even?

vruss:
> 
> 
> image
> Doesn't look even?

Yes I couldn't figure out how to view the margins using the visual debug tool so I made a quick pull request to get a discussion going. I will adjust the values until they are centered.

Joehuu (changes requested):
Mods button has one more pixel without mods.

bdach (dismissed):
Also still looks uneven when mods are shown, eyeballing it on my screen (pixels are real pixels as reported by `mspaint`, not necessarily what the draw visualiser shows):

image

To access the draw visualiser please press <kbd>Ctrl</kbd>+<kbd>F1</kbd>, after making sure there are no programs that could be intercepting that hotkey (i.e. parsec, among others)

I would suggest refactoring the drawable hierarchy so that the *mods* text is under the fillflow to make applying margins easier. Right now it's beside it.

vruss:
How should it center it considering it's parent is tilting to the right? If I were to center it in the middle and meassure the pixels, it would look like it's too far to the right I think.

Also how did you measure the number of pixels without eyeballing it?

bdach:
> How should it center it considering it's parent is tilting to the right? If I were to center it in the middle and meassure the pixels, it would look like it's too far to the right I think.

You should be able to use `FooterButton`'s `SHEAR_WIDTH` to compensate:
https://github.com/ppy/osu/blob/005a4928385c811f57d4a39cecc2a921eb2022e9/osu.Game/Screens/Select/FooterButton.cs#L20-L22

> Also how did you measure the number of pixels without eyeballing it?

Launched Paint and drew a few lines. The "eyeballing" part was mostly because it's a rough measurement. Also note that due to usages of `ScalingContainer`s higher up in the hierarchy the numbers you put in as drawable dimensions aren't actual pixels - see following picture:

image

The *proportion* to the parent scaling container should be preserved, however.

vruss:
Thank you for the help! 

> I would suggest refactoring the drawable hierarchy so that the _mods_ text is under the fillflow to make applying margins easier. Right now it's beside it.

Just to make sure I understood you right. You suggest making the _FooterButton_ text a child of the fillflow container?

bdach:
> Just to make sure I understood you right. You suggest making the _FooterButton_ text a child of the fillflow container?

Yeah, I think it would be cleaner, since then you wouldn't have to come up with a magic number for the right margin, just use one common number and offset it by the shear width (if I'm thinking correctly you might have to add half of it on the left and subtract half of it on the right, but I'd recommend not taking my word on that and double-checking).

vruss:
bild

Is it possible for the child class _FooterButtonMods_ to modify and add it's content to the parent class _FooterButton_ container property? Instead of using the ```Add(new FillFlowContainer``` method.

bdach:
Upon closer inspection I'm not too sure myself what the best course of action is here. I suppose `FooterButton`'s `TextContainer` is exposed as protected, but it has a fixed size which won't work too well here. Maybe it should be wrapped in another `FillFlowContainer` in the base, which then would be exposed as protected and then you could add to it in the deriving `FooterButtonMods`; not sure.

vruss:
> 
> 
> Upon closer inspection I'm not too sure myself what the best course of action is here. I suppose `FooterButton`'s `TextContainer` is exposed as protected, but it has a fixed size which won't work too well here. Maybe it should be wrapped in another `FillFlowContainer` in the base, which then would be exposed as protected and then you could add to it in the deriving `FooterButtonMods`; not sure.

I've begun refactoring the _FooterButton_ as you suggested but I've run into two problems. 
1. The mod text is no longer centered even though it's anchor points are centered.
2. The mod text box is pushing the mods icons far away. But if I remove the fixed size of the text, then the entire _FooterButton_ shrinks when no mods are selected.

bild

bdach:
Yeah, making the sprite text that large won't help, the anchor won't do anything. I was thinking more along the lines of:

```
FooterButton
  Container
    FillFlowContainer  // expose in FooterButton as ButtonContentContainer?
      Container  // currently exposed as TextContainer
        OsuSpriteText  // currently exposed as SpriteText
      FooterModDisplay
      OsuSpriteText  // mod difficulty multiplier text
```

If the fillflow is exposed in `FooterButton` you can `Add()` directly to it instead of to `FooterButtonMods` itself. I'd also move all of the outside padding of the content to that fillflow.

peppy:
This seems to still not be centering correctly. Was it intended to be?

bdach:
I think I have this *looking* OK now:

image

although there's two hacks in the code.

* Hack no. 1 is this offset:
  https://github.com/ppy/osu/blob/f130e48c9e05ba37d348feac2a541edfff533f14/osu.Game/Screens/Select/FooterButton.cs#L126-L132
  It's applied due to a opposite-sheared container inside a sheared container:

  image

  It'll probably have to stay whatever we do.
* Hack no. 2 is this:
  https://github.com/ppy/osu/blob/f130e48c9e05ba37d348feac2a541edfff533f14/osu.Game/Screens/Select/FooterButtonRandom.cs#L30-L31
  I'm forcing the texts to always be present to avoid width flickering of `FooterButtonRandom`, which internally does fades on two `SpriteText`s. Their presence (or lack thereof) when alpha hits 0 breaks my method of keeping the padding in `UpdateAfterChildren()`.

The `UpdateAfterChildren()` thing is done to keep the padding the same after the mod buttons expand. However, if we're fine with enforcing a set, not-necessarily-ideal value (say, hardcoded, somewhere between 30-40) and having the padding change after mods are selected, then this all could get a bit simpler.

vruss:
Thank you @bdach for helping me with this pull-request. I didn't know how to finish, nor do I have time to work on this atm.

peppy (dismissed):
This solution feels pretty good

PR #7394, opened 2019-12-29
Add context menu on direct panels

Joehuu:
- [x] Depends on https://github.com/ppy/osu/pull/7393 (as context menus on `SearchableListOverlay`s are broken rn)

Just one menu item for now, which just opens beatmap info (same as left clicking panel).

Other fixes:
- inherits `OsuClickableContainer` for code quality, but also for hover and click sounds


peppy (approved):

peppy (dismissed):
no harm in this.

peppy (approved):

PR #7398, opened 2019-12-30
Remove generic editor beatmap

smoogipoo:
Removing the generic editor beatmap so that the beatmap can eventually be moved to the `Editor` itself in an upcoming PR.


osu.Game/Screens/Edit/EditorBeatmap.cs
```diff
@@ -54,70 +54,71 @@ public BeatmapInfo BeatmapInfo
...
 
-        private void trackStartTime(T hitObject)
+        private void trackStartTime(HitObject hitObject)
         {
             startTimeBindables[hitObject] = hitObject.StartTimeBindable.GetBoundCopy();
             startTimeBindables[hitObject].ValueChanged += _ =>
             {
                 // For now we'll remove and re-add the hitobject. This is not optimal and can be improved if required.
-                beatmap.HitObjects.Remove(hitObject);
+                ((IList)beatmap.HitObjects).Remove(hitObject);
```
peppy: looked at this PR a few times and still not really sure about this solution... but let's go with it for now, since it's quite localised.

peppy (approved):

osu.Game/Screens/Edit/EditorBeatmap.cs
smoogipoo: I'm willing to accept any alternative. One possible solution is to use `List<HitObject>`, but you cannot have co-variant types with non-interfaces so there can't also be a `List<THitObject>`.

PR #7407, opened 2019-12-31
Add automatic pagination to DirectOverlay

aemino:
Adds automatic pagination to osu!direct so that scrolling down the list of beatmaps triggers automatic loading of the next page of beatmap results.

This is my first time working with the codebase of osu!/osu!framework, so I'm sure that I've made some mistakes or written something that would be considered non-idiomatic.

Constants worth (re)considering: page load debounce (currently set to 500ms) and distance-from-scroll-end (currently leniency = 500).

Also, this change can cause a lot of panels to be loaded at once; perhaps panels that are not visible should be removed.

Lastly, this change relies on a framework change: exposing `ScrollContainer.ScrollableExtent`. I wanted to hold off on making a PR to change that until the semantics of the feature have been vetted.


osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -179,6 +179,39 @@ private void updateResultCounts()
                                     "Tag".ToQuantity(ResultAmounts.Tags);
         }
 
+        private IEnumerable<DirectPanel> createPanels(PanelDisplayStyle displayStyle, IEnumerable<BeatmapSetInfo> sets) {
```
bdach: This would normally get picked up by CI, but since right now it's not even compiling due to missing framework changes, I'll just point this out now - we generally use BSD style braces instead of K&R, so opening brace goes on a new line here.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -232,21 +251,49 @@ protected override void PopIn()
         private SearchBeatmapSetsRequest getSetsRequest;
 
         private readonly Bindable<string> currentQuery = new Bindable<string>(string.Empty);
+        private int currentPage = 1;
+        private bool isLastPageLoaded = false;
```
bdach: Admittedly I'm uneasy about all this state landing in here. Feels like the overlay might get a little bloaty. This could maybe be factored out to a paged request handling component? Purely spitballing:

```csharp
class BeatmapSetPager
{
   public event PageFetched;

   // returns true if page present, false if not
   public bool FetchNextPage() { ... }
   public void ResetToFirstPage() { ... }
}
```

Does that sound feasible?

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -232,21 +251,49 @@ protected override void PopIn()
...
+            
+            currentPage++;
+            updateSearch();
+
+            getSetsRequest.Success += response =>
+            {
+                addPageDebounce = Scheduler.AddDelayed(() =>
+                {
+                    addPageDebounce = null;
+                }, 500);
```
bdach: This feels dangerous/race-prone. I'm picturing a situation where the response is handled on a worker thread and the scheduling thread overwrites the outer assignment with the inner one. Or the other way around, really.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -232,21 +251,49 @@ protected override void PopIn()
...
+        private void queueAddPage()
+        {
+            if (getSetsRequest != null)
+                return;
+
+            if (addPageDebounce != null)
+                return;
+            
+            if (isLastPageLoaded)
+                return;
```
bdach: I don't honestly see much value to splitting the guards out like that, but that's subjective.

bdach (changes requested):
I think this has some way to go still, reading through it I felt that this could use some improvement. Thanks for the contribution, though, I think this is wanted and can be definitely smoothed out.

osu.Game/Overlays/DirectOverlay.cs
aemino: I was thinking that this might be an issue, so I'm glad that you picked up on it. I'm thinking that I'll take your suggestion of making a separate paging component and try to handle this in there.

aemino: Also, do you have any suggestions on handling this more elegantly? The main purpose of this whole debounce system is so that multiple page loads aren't triggered at once. In theory, what I really "want" to wait for is for the new `DirectPanel`s to be added so that `IsScrolledToEnd` is false. Am I best sticking with a more thread-safe debounce, or is there some other mechanism that I could use here?

bdach: I mean I guess the debounce would be fine if it were guaranteed that both assignments are done on the scheduler thread. The obvious way to do this seems to me to be

```csharp
                Schedule(() => addPageDebounce = Scheduler.AddDelayed(() =>
                {
                    addPageDebounce = null;
                }, 500));
```

which however looks ugly (the innards might fare better as an extracted method) and also like it might catch fire due to nested schedule calls (although I *think* it should be fine, `SquirrelUpdateManager` also schedules a delayed task inside of a scheduled call).

FWIW I think the debounce call should stay within the overlay since that's presentation logic and belongs there.

aemino:
Sorry for the delay on this one! I've attempted to refactor out the paging logic into `BeatmapSetPager`. Let me know if this design works or if it needs to be changed again. Thanks :)

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -244,54 +249,120 @@ protected override void PopIn()
...
+            if (page > 1)
+            {
+                BeatmapSets = BeatmapSets.Concat(sets);
+                addPanels(Filter.DisplayStyleControl.DisplayStyle.Value, sets);
+            }
+            else
+            {
+                BeatmapSets = sets;
+                recreatePanels(Filter.DisplayStyleControl.DisplayStyle.Value);
+            }
```
aemino: Should this go on the scheduler thread?

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -225,28 +249,122 @@ protected override void PopIn()
...
+                    {
+                        var sets = response.BeatmapSets.Select(r => r.ToBeatmapSet(rulesets)).ToList();
+
+                        if (sets.Count <= 0) IsLastPageFetched = true;
+
+                        PageFetch?.Invoke(currentPage, sets);
+                        
+                        getSetsRequest = null;
+                        currentPage++;
+                    });
```
bdach: Looks fragile due to the fire-and-forget `Task.Run`. What happens if any of the operations fail? I'm worried they'll fail silently.

Why's this in a task in the first place? I don't think any of this is inherently costly.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -225,28 +249,122 @@ protected override void PopIn()
...
+
+            public bool IsLastPageFetched { get; private set; } = false;
+            public bool IsFetching => getSetsRequest != null;
+
+            public BeatmapSetPager(RulesetStore rulesets)
+            {
+                this.rulesets = rulesets;
+            }
+
+            public SearchBeatmapSetsRequest FetchNextPage(string query, RulesetInfo ruleset, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
```
bdach: This operation will be wrong if somebody switches the criteria between `FetchNextPage` calls, since it won't reset the page counter. The criteria should probably come in through the ctor (so instances will be single-filter-use-only), or be set with a separate method that will call `Reset()`.

osu.Game/Overlays/DirectOverlay.cs
bdach: That would probably be correct, yes, to prevent mutating UI from a non-UI thread.

bdach (changes requested):

osu.Game/Overlays/DirectOverlay.cs
aemino: It was just a task originally, so I just thought to maintain the status quo. I'll remove it.

osu.Game/Overlays/DirectOverlay.cs
aemino: The query is now provided in the constructor.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -179,6 +179,41 @@ private void updateResultCounts()
...
+        {
+            if (BeatmapSets == null)
+                return;
+
+            var newPanels = createPanels(displayStyle, sets);
+
+            LoadComponentsAsync(newPanels, p =>
+            {
+                if (panels == null)
+                    return;
```
bdach: Guaranteed non-null if I'm not missing anything.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -225,28 +246,139 @@ protected override void PopIn()
...
         }
 
-        private SearchBeatmapSetsRequest getSetsRequest;
-
         private readonly Bindable<string> currentQuery = new Bindable<string>(string.Empty);
 
         private ScheduledDelegate queryChangedDebounce;
+        private ScheduledDelegate addPageDebounce;
         private PreviewTrackManager previewTrackManager;
+        private BeatmapSetPager beatmapSetPager;
```
bdach: I'd maybe annotate with `[CanBeNull]`.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -225,28 +246,139 @@ protected override void PopIn()
...
+
+        private class BeatmapSetPager
```
bdach: Maybe a separate file for this? Blows up the size here kind of unnecessarily.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -225,28 +246,139 @@ protected override void PopIn()
...
+                this.rulesets = rulesets;
+
+                this.query = query;
+                this.ruleset = ruleset;
+                this.searchCategory = searchCategory;
+                this.sortCriteria = sortCriteria;
+                this.sortDirection = sortDirection;
+            }
+
+            public SearchBeatmapSetsRequest FetchNextPage()
```
bdach: Now that I'm reading this a bit more carefully I'm not sure about this API choice to return a request. It does a bit of a WTF thing if a request is in-flight (returns a `null`). Honestly I'd consider making the pager queue API requests itself and making this `void`. Or `bool`, at most, to get rid of some null-checks, and return `false` if there's an ongoing request.

Also this component is technically not thread-safe since you could get races between reads of `getSetsRequest` and the assignment, therefore getting >1 in-flight request where the last one gets assigned. Easiest (but maybe not best) way to fix is to add an exclusive lock on that critical section. I don't expect too much starvation, debounce logic would make hitting this race probably nigh impossible too, but let's do things properly.

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -260,28 +392,36 @@ private void updateSearch()
...
-            API.Queue(getSetsRequest);
+            queueAddPage();
+        }
+
+        protected override void Update()
+        {
+            base.Update();
+
+            if (panels == null)
+                return;
```
bdach: Seems impossible too.

bdach (changes requested):
Shaping up better now. Mostly nitpicks left, except for one major suggestion to the API of the pager. Sorry for the iterative process, it's tough (or sometimes not possible) to pick everything out at once.

aemino:
No worries. I was fully aware when I submitted this patch that this would be an iterative process. I appreciate all of the reviews!

osu.Game/Overlays/DirectOverlay.cs
aemino: I think that the original idea with this check had to do with the fact that `recreatePanels` temporarily sets `panels` to null before recreating them in a `LoadComponentAsync` call. The only time I could see an issue here is if `recreatePanels` was called as a result of the display style being updated while adding additional panels. Do you think the guard should be dropped?

osu.Game/Overlays/DirectOverlay.cs
aemino: Where should the file go?

osu.Game/Overlays/DirectOverlay.cs
aemino: I share your concern about this function as well. I just wasn't sure when I was writing it what the proper way to lower the `API` provider into this component would be, and forgot to make a note about it. What's the best way to do this?

osu.Game/Overlays/DirectOverlay.cs
iiSaLMaN: in it's own file, `BeatmapSetPager.cs`

osu.Game/Overlays/DirectOverlay.cs
aemino: Also, I suppose a lock would fix the race condition you mention. I'll try to think of an alternate solution, but this is probably the most C# I've ever written, so my odds of coming up with something better are low. :smile:

osu.Game/Overlays/DirectOverlay.cs
aemino: I was asking more about the directory, but I suppose you're implying that it should go in `osu.Game/Overlays`

osu.Game/Overlays/DirectOverlay.cs
bdach: So I did miss something. Would be nice to get rid of it, but I suppose it can stay.

osu.Game/Overlays/DirectOverlay.cs
bdach: `osu.Game/Overlays/Direct` is probably best. If needed, can move up to `osu.Game/Online` somewhere.

osu.Game/Overlays/DirectOverlay.cs
bdach: I'd go with constructor injection. With the current instance-per-filter-set implementation you'd probably have to pass it to the ctor manually. You could also make the pager an always-present visual component with no content and use `[Resolved]` injection, but that's slightly uglier IMO.

osu.Game/Overlays/Direct/BeatmapSetPager.cs
```diff
@@ -0,0 +1,87 @@
...
+                    if (sets.Count <= 0) IsLastPageFetched = true;
+
+                    PageFetch?.Invoke(currentPage, sets);
+
+                    getSetsRequest = null;
+                    currentPage++;
+                };
+
+                API.Queue(getSetsRequest);
+            }
```
aemino: Not sure if this is correct. Can't lock on the request before this section because it may be null.

aemino (commented):
All right, I think that I've made the requested changes, although I may have not done the locking part correctly, so let me know if that's not what you wanted. Sorry for the delay on this one--I've been busy with other things.

osu.Game/Overlays/Direct/BeatmapSetPager.cs
aemino: I suppose I could also make a dummy object and use that to lock the whole method if desired.

bdach: This locking is definitely not correct. This doesn't really do what it appears to be doing (it takes the lock for the duration of subscribing to the event and queueing a request, and actually doesn't keep hold of it *inside* the response callback), and it doesn't prevent the race either. The actual race is slightly higher, in between the read of `getSetsRequest` and assignment to it after the null-check.

But honestly now that I look at this code again after a while the chance of this racing is probably so slim that it might not really be worth checking for. Probably fine to ignore that but mention in the xmldoc that the method is not thread-safe so that users tread carefully.

aemino: I think that it's quite unlikely for that to happen as well. I'll just remove the lock and add a comment like you suggested.

bdach:
There are some compile errors reported by appveyor here, it seems.

aemino:
Yes, there is a framework change necessary for this PR. I'll make a PR to osu-framework with the changes when I get a chance.

aemino:
In addition to the `ScrollContainer` change, there is another change that I remembered will be required: `BeatmapListingOverlay` uses `SearchBeatmapSetsRequest`, which needs to be updated to provide a page number. Should it just default to the first page for now?

peppy:
As a heads up, the osu!direct overlay is soon being decommissioned, replaced by `BeatmapListingOverlay`. It may be better than this PR is closed and the function is reimplemented there.

Also, please look at using the `cursor` logic in the response rather than manually specifying a page number – it is more reliable.

aemino:
I thought as much. I'll take a look at reimplementing this when I get a chance.

aemino:
Closing in favor of #8762.

PR #7409, opened 2019-12-31
Move definition of column colors etc. into a new type

mcendu:
Prereq for skinning mania.


osu.Game.Rulesets.Mania/UI/Components/EvenColumn.cs
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public class EvenColumn : ColumnType
```
mcendu: `ColumnTypeEven` might be better.

osu.Game.Rulesets.Mania/UI/Components/ColumnType.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public abstract Color4 ColumnColour { get; }
+        public virtual float DefaultColumnWidth => Column.COLUMN_WIDTH;
+
+        public virtual string NoteTexture => "mania-note1";
+        public virtual string HoldTexture => "mania-note1L";
+        public virtual string HoldHeadTexture => "mania-note1H";
+        public virtual string HoldTailTexture => "mania-note1T";
+        public virtual string KeyTexture  => "mania-key1";
+        public virtual string KeyDownTexture => "mania-key1D";
```
mcendu: These values does not point to the actual textures but the name of the textures.

Another name could be better.

osu.Game.Rulesets.Mania/UI/Components/EvenColumn.cs
iiSaLMaN: As far as I've seen, derived class name mostly comes before inherited class name, sometimes in-between and can be shortened if not misleading:
    - `FooType : Type`
    - `DrawableFooType : DrawableType`

osu.Game.Rulesets.Mania/UI/Components/ColumnType.cs
iiSaLMaN: perhaps `...TextureName`?

osu.Game.Rulesets.Mania/UI/Components/OddColumnType.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public class OddColumnType : ColumnType
```
iiSaLMaN: I don't like how this column type is inconsistent conpared to other types, how about abstracting the texture name properties and override here?

frenzibyte (changes requested):

osu.Game.Rulesets.Mania/UI/Column.cs
```diff
@@ -101,22 +100,25 @@ public Column(int index)
...
-                isSpecial = value;
+                columnType = value;
 
-                Width = isSpecial ? special_column_width : COLUMN_WIDTH;
+                AccentColour = columnType.ColumnColour;
+                Width = columnType.DefaultColumnWidth;
             }
         }
 
+        public bool IsSpecial { get; set; }
```
bdach: Do we even want this to still be a thing? Maybe it shouldn't exist anymore, or at least be a get-only convenience property:
```csharp
public bool IsSpecial => columnType is SpecialColumnType;
```
I'm worried if this survives it'll be bug-prone (two sources of truth for one piece of info).

osu.Game.Rulesets.Mania/UI/Components/SpecialColumnType.cs
```diff
@@ -0,0 +1,22 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public class SpecialColumnType : ColumnType
+    {
+        public override Color4 ColumnColour => new Color4(0, 48, 63, 255);
+        public override float DefaultColumnWidth => 70;
```
bdach: I'd expose and reuse
https://github.com/ppy/osu/blob/81d2487f65f827a35b124719507d641e1da07285/osu.Game.Rulesets.Mania/UI/Column.cs#L25

bdach (changes requested):

osu.Game.Rulesets.Mania/UI/Components/SpecialColumnType.cs
mcendu: Is there any reason to still put it in `Column`?

bdach: IMO the two constants should be side by side for easy searching and access.

osu.Game.Rulesets.Mania/UI/Components/ColumnType.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Mania.UI.Components
+{
+    public abstract class ColumnType
```
peppy: `ColumnType` is a bad name for this. `ColumnInfo` maybe?

Also please add full xmldoc for the class and all properties.

peppy (changes requested):

smoogipoo:
I'm not sure this is the way we want to go about this. When considering skinning, most of the values here pertain only to legacy values (which will all be consolidated within `ManiaLegacySkinTransformer` (see the other similar classes for reference)).

Rather than giving columns a type, how about deriving the column to eventually result in an inheritance structure like:
```
Column:
    - LegacyColumn:
        - LegacyEvenColumn
        - LegacyOddColumn
        - LegacySpecialColumn
    - DefaultColumn:
        - DefaultEvenColumn
        - DefaultOddColumn
        - DefaultSpecialColumn
```

Then, all the `mania-noteX` members will only appear within `LegacyColumn`.

mcendu:
Superseded by #7431 (thanks @smoogipoo for the hint).

PR #7414, opened 2020-01-01
Fix File Sharing on iOS 13

### LINKED ISSUE #7049: Allow imports via Files.app iOS
**ATTENTION**: Plan to implement myself.
Related: #6525, #5265
I noticed there was a lack of a way to import Osu! skins and beatmaps, I propose that the Osu! files structure on iOS be exposed to the files app so that people can move in skins manually either from download or otherwise. Alternatively, a mobile module for exposing the relevant dirs could be made.

dtsdao:
Fixes #7049 
A patch to fix the disability to import files via the "Share" menu.
Refer: https://forums.developer.apple.com/thread/118932

### DIFF
--- a/osu.iOS/Info.plist
+++ b/osu.iOS/Info.plist
@@ -14,15 +14,15 @@
 	<string>0.1.0</string>
 	<key>LSRequiresIPhoneOS</key>
 	<true/>
-	<key>LSSupportsOpeningDocumentsInPlace</key>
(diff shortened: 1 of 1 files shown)


swoolcock:
Looks alright, will test when I get a chance.

peppy (approved):
works as expected, thanks!

PR #7385, opened 2019-12-28
Update BeatmapSetOverlay header in line with the web design

EVAST9919:
Depends on:
- [x] https://github.com/ppy/osu/pull/7380
- [ ] https://github.com/ppy/osu/pull/7418

Screenshot_1
Screenshot_1

PR #7418, opened 2020-01-01
Implement OverlayRulesetSelector component

EVAST9919:
Split from https://github.com/ppy/osu/pull/7385


osu.Game/Overlays/OverlayRulesetTabItem.cs
```diff
@@ -0,0 +1,101 @@
...
+            set
+            {
+                if (accentColour == value)
+                    return;
+
+                accentColour = value;
+
+                updateState();
+            }
+        }
```
bdach: This seems to be essentially identical to:
```csharp
private Bindable<Color4> accentColour;
public Color4 AccentColour
{
    get => accentColour.Value;
    set => accentColour.Value = value;
}
```
and then having a `BindValueChanged` in `LoadComplete`. I know you just moved it from its previous location but if we're touching this anyway then we might as well remove the reinvented wheel.

osu.Game/Overlays/OverlayRulesetTabItem.cs
```diff
@@ -0,0 +1,101 @@
...
+
+        protected Action<Color4> OnStateUpdated;
```
bdach: I'm not sure this should be a thing. Would rather have, say, a second bindable maybe?

bdach (commented):
I think I'm mostly fine with this at its current state now.

bdach:
Is this in a ready state? Hard to tell since there's been new changes since my last review of this (which seem fine).

EVAST9919:
Yes, it is. I just added colour schemes support since https://github.com/ppy/osu/pull/7541 has been merged.

osu.Game/Overlays/OverlayRulesetTabItem.cs
```diff
@@ -0,0 +1,100 @@
...
+
+        protected override void OnActivated() => updateState();
+
+        protected override void OnDeactivated() => updateState();
+
+        private void updateState()
+        {
+            Text.Font = Text.Font.With(weight: Active.Value ? FontWeight.Bold : FontWeight.Medium);
+
+            currentColour.Value = IsHovered || Active.Value
```
peppy: you can't write statements like this. it's impossible to read. split it out more.

osu.Game/Overlays/OverlayRulesetSelector.cs
```diff
@@ -0,0 +1,51 @@
...
+using osu.Framework.Graphics.UserInterface;
+using osu.Game.Graphics;
+using osu.Game.Rulesets;
+using osuTK;
+using osuTK.Graphics;
+using System.Linq;
+
+namespace osu.Game.Overlays
+{
+    public class OverlayRulesetSelector : RulesetSelector
```
peppy: this class is non-abstract but not being used anywhere directly except for a test. is this planned to be used somewhere in the future?

peppy (changes requested):

osu.Game/Overlays/OverlayRulesetSelector.cs
EVAST9919: Will be used in `RankingsOverlay`

peppy (approved):

PR #7425, opened 2020-01-02
Implement missing API code for channel read state

Craftplacer:


osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -445,6 +445,14 @@ protected override Task Poll()
             return tcs.Task;
         }
 
+        public void MarkChannelAsRead(Message message)
```
peppy: how have you tested this? it's not called from anywhere.

Craftplacer: I added this so it could be used later on, if you want I could implement this into ChatOverlay to actually set channels as read.

peppy: It's very hard to merge something if the thing being added has no way of testing it. And it shows that the code has probably been written without a single check as to whether it works.

Craftplacer: Can you suggest me what type of test to implement? This is API code so I'm unsure how to test such.

peppy: API testing is still somewhere we are lacking. It's best you implement usage of the function in the same PR as adding it. Feel free to split this into its own PR if that makes more sense.

Craftplacer: See commit https://github.com/ppy/osu/pull/7425/commits/7cc388b5abc56b5d49d9fabe9596db1701b22641

Craftplacer: It has also has been tested successfully with kind @mouzedrift
https://streamable.com/qapuj

Craftplacer:
Can I ask about the review status of this PR?
An usage case has been added and tested locally successfully.

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -445,6 +445,15 @@ protected override Task Poll()
...
+        {
+            var channel = JoinedChannels.First(c => c.Id == message.ChannelId);
```
bdach: Probably better written as
```suggestion
            var channel = JoinedChannels.Single(c => c.Id == message.ChannelId);
```
so that it fails loudly if there are ever duplicate channel ids.

peppy:
The review status is that it has not yet been reviewed.

osu.Game/Online/Chat/Channel.cs
```diff
@@ -36,6 +36,11 @@ private long[] userIds
         /// </summary>
         public readonly SortedList<Message> Messages = new SortedList<Message>(Comparer<Message>.Default);
 
+        /// <summary>
+        /// Contains all the messages that weren't read by the user.
+        /// </summary>
+        public IEnumerable<Message> UnreadMessages => Messages.Where(m => LastReadId < m.Id);
```
bdach: Just to be extra sure, is it documented anywhere that message IDs are temporally sequential (as in lower ID ⇔ earlier message)? Poking web sources it does seem to be the case but I'd like to be 100% on that.

bdach (changes requested):
I'm quite out of touch with web and API details but if `LastReadId` is to be used meaningfully for reads on lazer's side [it should also probably be bumped every time a message is sent](https://github.com/ppy/osu-web/blob/1ee2dc1021fdc26a4b35b3eb4bea4f8a2bea1b04/app/Http/Controllers/Chat/Channels/MessagesController.php#L146). I suppose that might be considered out of scope for this pull but I figured it was worth pointing out.

I haven't ran the code yet, just read it.

osu.Game/Online/Chat/Channel.cs
peppy: it will 100% be the case (primary key of table).

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -445,6 +445,15 @@ protected override Task Poll()
             return tcs.Task;
         }
 
+        public void MarkChannelAsRead(Message message)
```
iiSaLMaN: This is definitely requiring a xmldoc.

frenzibyte (changes requested):
Does `ChannelManager.MarkChannelAsRead` mark all unread messages as read in channel of the given message, if yes, then this should not take the message as a parameter, but rather the channel of this message.

osu.Game/Online/Chat/ChannelManager.cs
Craftplacer: I didn't know this LINQ method existed, I will try to use them in the future.

osu.Game/Online/Chat/ChannelManager.cs
Craftplacer: Resolved with https://github.com/ppy/osu/pull/7425/commits/50e357a79930b0980fb67fddafc21c65712a0c49

Craftplacer:
> Does `ChannelManager.MarkChannelAsRead` mark all unread messages as read in channel of the given message, if yes, then this should not take the message as a parameter, but rather the channel of this message.

I changed now the method parameters to make more sense, it will mark the whole channel as read unless specified and it will fail if the user specified a channel and a message apart from each other.

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -445,6 +446,27 @@ protected override Task Poll()
...
+        /// <see>(see <paramref name="message"/> for more information)</see>
+        /// </summary>
+        /// <param name="channel">The channel that will be marked as read</param>
+        /// <param name="message">The message where it will be read up to. If <see langword="null"/> was provided, the latest message of the <paramref name="channel"/> will be read.</param>
+        public void MarkChannelAsRead(Channel channel, Message message = null)
+        {
+            if (message == null)
+                message = channel.Messages.Last();
+            else
+                Debug.Assert(channel.Id == message.Id, "Provided channel and message owner channel aren't equal.");
```
peppy: i highly doubt this path is tested (the comparison is wrong). if it's not tested/used just remove it and remove the message parameter for now.

peppy (changes requested):

osu.Game/Online/Chat/ChannelManager.cs
Craftplacer: I'm blind, I notice this way too late

Craftplacer: Resolved with https://github.com/ppy/osu/pull/7425/commits/2ea1367a881cab6e5e9eed787971a6f5ca22f989

bdach (commented):
Source looks saner, and runs without crashing, although I'd say the requests to mark channels as read are a bit gratuitous, as there could be less of them:

image

If there have been no new messages in the channel then IMO requests to mark them as read should not be sent. I'd consider handling that case as a guard & early return in `MarkChannelAsRead`.

peppy (approved):
seems ok

peppy:
Added a catch for the case there's no messages in a channel (would have linq crashed previously).

PR #7426, opened 2020-01-02
Make discord status text in line with stable during gameplay

Joehuu:
The default text is changed to "Playing solo" for consistent terminology.

### DIFF
--- a/osu.Game/Users/UserActivity.cs
+++ b/osu.Game/Users/UserActivity.cs
@@ -52,7 +52,29 @@ namespace osu.Game.Users
                 Ruleset = ruleset;
             }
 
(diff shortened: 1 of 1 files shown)


iiSaLMaN:
I kinda disagree with this solution as this property cannot be expanded to custom rulesets but is hardcoded. How about a virtual property on `Ruleset` where you can retrieve the playing status from.

peppy:
As above, this was intentionally *not* added in such a way because it is not extensible. Rulesets should expose a `PlayingVerb` or similar, which can be used for this display.

Should be properly documented.

Joehuu:
Superseded by https://github.com/ppy/osu/pull/7430.

PR #7431, opened 2020-01-03
Make Column abstract

mcendu:
Supersedes #7409. Facilitates the process of implementing skinning for mania.


osu.Game.Rulesets.Mania/UI/Column.cs
```diff
@@ -31,109 +27,52 @@ public class Column : ScrollingPlayfield, IKeyBindingHandler<ManiaAction>, IHasA
...
             Width = COLUMN_WIDTH;
+        }
 
-            background = new ColumnBackground { RelativeSizeAxes = Axes.Both };
-
-            Container hitTargetContainer;
-
-            InternalChildren = new[]
+        public static Column CreateColumnFromType(ColumnType type, int index)
+            => type switch
```
bdach: I don't think we have a consensus on switch expressions yet. Might be better suited as a plain old switch.

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
```diff
@@ -0,0 +1,127 @@
...
+                    RelativeSizeAxes = Axes.X,
+                    Height = ManiaStage.HIT_TARGET_POSITION,
+                },
+                background,
+                topLevelContainer = new Container { RelativeSizeAxes = Axes.Both }
+            };
+
+            topLevelContainer.Add(explosionContainer.CreateProxy());
+
+            Direction.BindValueChanged(dir =>
```
mcendu: Duplicate code.

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
```diff
@@ -0,0 +1,127 @@
...
+                hitObjectArea.AccentColour = value;
+            }
+        }
+
+        protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
+        {
+            var dependencies = new DependencyContainer(base.CreateChildDependencies(parent));
+            dependencies.CacheAs<IBindable<ManiaAction>>(Action);
+            return dependencies;
+        }
```
mcendu: Useless override.

osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
```diff
@@ -20,6 +21,18 @@ public struct StageDefinition
         /// </summary>
         /// <param name="column">The 0-based column index.</param>
         /// <returns>Whether the column is a special column.</returns>
-        public bool IsSpecialColumn(int column) => Columns % 2 == 1 && column == Columns / 2;
+        public bool IsSpecialColumn(int column)
+            // TODO: Inline this method.
+            => Columns % 2 == 1 && column == Columns / 2;
+
+        public ColumnType TypeOfColumn(int column)
```
mcendu: Tests can be added to test this method, but as the original code is not covered by tests...

peppy:
If you think you can add tests for anything, we will not say no. Generally I try and develop by writing tests before making a change or addition.

bdach (commented):
I think this is mostly fine in its current state. The only thing I'm sort of not too hot on is the existence of `public virtual IsSpecial` in `Column` and the derived, but I don't see a nicer solution right now. Traits or product types from other languages would be nice for this, but in current C# I think it's as good as it gets.

osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
```diff
@@ -16,10 +17,18 @@ public struct StageDefinition
...
 
         /// <summary>
-        /// Whether the column index is a special column for this stage.
+        /// Get the <see cref="ColumnType"/> for a given column index.
         /// </summary>
         /// <param name="column">The 0-based column index.</param>
-        /// <returns>Whether the column is a special column.</returns>
-        public bool IsSpecialColumn(int column) => Columns % 2 == 1 && column == Columns / 2;
+        /// <returns>The type of the column.</returns>
+        public ColumnType TypeOfColumn(int column)
```
smoogipoo: Prefer `GetColumnType`.

I'm not 100% on this being here due to skin-specific special styles, but let's go with it for now.

osu.Game.Rulesets.Mania/UI/Column.cs
```diff
@@ -31,109 +26,60 @@ public class Column : ScrollingPlayfield, IKeyBindingHandler<ManiaAction>, IHasA
...
+        protected Column(int index)
         {
             Index = index;
 
             RelativeSizeAxes = Axes.Y;
             Width = COLUMN_WIDTH;
+        }
 
-            background = new ColumnBackground { RelativeSizeAxes = Axes.Both };
+        public static Column CreateColumnFromType(ColumnType type, int index)
```
smoogipoo: Don't like this being `static` and part of `Column`. Should be virtual (or abstract) and part of `ManiaStage` instead.

osu.Game.Rulesets.Mania/UI/Column.cs
```diff
@@ -31,109 +26,60 @@ public class Column : ScrollingPlayfield, IKeyBindingHandler<ManiaAction>, IHasA
...
                     Bottom = dir.NewValue == ScrollingDirection.Down ? NotePiece.NOTE_HEIGHT / 2 : 0
                 };
 
-                keyArea.Anchor = keyArea.Origin = dir.NewValue == ScrollingDirection.Up ? Anchor.TopLeft : Anchor.BottomLeft;
+                KeyArea.Anchor = KeyArea.Origin = dir.NewValue == ScrollingDirection.Up ? Anchor.TopLeft : Anchor.BottomLeft;
             }, true);
         }
 
-        public override Axes RelativeSizeAxes => Axes.Y;
+        public virtual bool IsSpecial => false;
```
smoogipoo: What's this used for? Couldn't you use the stage definition directly?

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
```diff
@@ -0,0 +1,104 @@
...
+namespace osu.Game.Rulesets.Mania.UI
+{
+    public class DefaultColumn : Column, IHasAccentColour
+    {
+        private readonly ColumnBackground background;
+        private readonly ColumnKeyArea keyArea;
+        private readonly ColumnHitObjectArea hitObjectArea;
+
+        // Pull up more stuff when needed.
+        protected override Container HitTargetContainer { get; }
```
smoogipoo: The column should have some basic layout associated with it, such as the positioning of the hit target and associated containers.

I'd prefer that the shared structure (hit target + children, key area, and top-level) were moved into the base class. The `DefaultColumn` would then inject its background wherever it needs to go.

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
```diff
@@ -0,0 +1,104 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Adds a DrawableHitObject to this Playfield.
+        /// </summary>
+        /// <param name="hitObject">The DrawableHitObject to add.</param>
+        public override void Add(DrawableHitObject hitObject)
+        {
+            hitObject.AccentColour.Value = AccentColour;
```
smoogipoo: Should be in the base class.

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
```diff
@@ -0,0 +1,104 @@
...
+            base.Add(hitObject);
+        }
+
+        public override void AddHitExplosion(DrawableHitObject judgedObject)
```
smoogipoo: This should only be in the base class. The hit explosion should be a skinnable drawable itself.

smoogipoo (changes requested):

osu.Game.Rulesets.Mania/UI/DefaultColumn.cs
mcendu: Put it here as legacy notes don't have accent colors.

smoogipoo: Accent colour is an opt-in thing - you won't be able to use a lazer `DrawableNote` along with a legacy `Column`, and a `LegacyDrawableNote` simply won't use implement the accent colour.

mcendu:
Superseded by #8513

PR #7432, opened 2020-01-03
Implement ability to create OverlayHeader with no TabControl

EVAST9919:
Also adds ability to use any type for `TabControlOverlayHeader`.


osu.Game.Tests/Visual/UserInterface/TestSceneOverlayHeaders.cs
```diff
@@ -0,0 +1,174 @@
...
+using osu.Framework.Allocation;
+using osu.Framework.Graphics.Textures;
+using osu.Game.Graphics.UserInterface;
+using osu.Game.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osuTK.Graphics;
+
+namespace osu.Game.Tests.Visual.UserInterface
+{
+    public class TestSceneOverlayHeaders : OsuTestScene
```
peppy: Please just call it `TestSceneOverlayHeader` to match conventions (you are testing `OverlayHeader` not `OverlayHeaders`)

osu.Game/Overlays/ControllableOverlayHeader.cs
```diff
@@ -0,0 +1,46 @@
...
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Graphics.UserInterface;
+using osu.Game.Graphics;
+using osuTK.Graphics;
+
+namespace osu.Game.Overlays
+{
+    /// <typeparam name="T">The type of item to be represented by tabs in <see cref="TabControl{T}"/>.</typeparam>
+    public abstract class ControllableOverlayHeader<T> : OverlayHeader
```
peppy: please add full xmldoc. also reconsider the naming – this contains a `TabControl` so why is this class not called `TabControlOverlayHeader` rather than an implementation?

peppy (changes requested):

osu.Game/Overlays/ControllableOverlayHeader.cs
EVAST9919: Well, `TabControlOverlayHeader` is already used. And honestly I have no idea how to rename it. If we will use `TabControlOverlayHeader` for this one - how should I name the original one?

EVAST9919: Nevermind, I figured that out

osu.Game/Overlays/TabControlOverlayHeader.cs
```diff
@@ -54,10 +75,10 @@ protected override TabItem<string> CreateTabItem(string value) => new OverlayHea
 
             private class OverlayHeaderTabItem : OverlayTabItem
             {
-                public OverlayHeaderTabItem(string value)
+                public OverlayHeaderTabItem(T value)
                     : base(value)
                 {
-                    Text.Text = value;
+                    Text.Text = value.ToString().ToLowerInvariant();
```
bdach: Not sure about the invariance in the long haul (localisation is what I'm worried about).

osu.Game/Graphics/UserInterface/OsuTabControl.cs
```diff
@@ -59,23 +67,22 @@ public OsuTabControl()
...
 
-        private Color4 accentColour;
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            accentColour.BindValueChanged(OnAccentColourChanged, true);
+        }
 
-        public Color4 AccentColour
+        protected virtual void OnAccentColourChanged(ValueChangedEvent<Color4> colour)
```
peppy: Make `AccentColour` protected and bind each usage rather than making a virtual event. the same as done everywhere else in the game.

peppy (changes requested):

osu.Game/Graphics/UserInterface/OsuTabControl.cs
```diff
@@ -22,6 +22,22 @@ namespace osu.Game.Graphics.UserInterface
...
+            set
+            {
+                AccentColourBindable.Value = value;
+
+                if (Dropdown is IHasAccentColour dropdown)
+                    dropdown.AccentColour = value;
+                foreach (var i in TabContainer.Children.OfType<IHasAccentColour>())
+                    i.AccentColour = value;
+            }
+        }
```
bdach: This is half-good, half-not-really now. Bindable is a plus, but what's the point if you end up propagating the value down instead anyways? This could be much nicer if `IHasAccentColour.AccentColour` was an `IBindable<Color4>`.

peppy:
instead of bindable, why not just virtual the property? feels like what you have is overkill since it's only used local to the class.

peppy:
the property, not a new method.

peppy (approved):

PR #7434, opened 2020-01-04
Add looping animation to slider reverse arrows

### LINKED ISSUE #7374: Add bounce animation to reverse arrow
Reverse arrow is bouncing to the beat in the stable version, so I think it should be doing so in osu!lazer as well. Maybe add an settings toggle to deal with this?
Personally I am used to seeing the reverse arrow bouncing and the animation makes maps a little easier to read for me.

Wieku:
Fixes #7374

I'm not sure about changing the animDuration. On very fast repeating sliders (like in The Big Black) previous values caused arrows to disappear almost instantly.


frenzibyte (changes requested):
Have you checked `BeatSyncedContainer`, will make this code pretty much simple.

Wieku:
Originally reverse arrows don't use slider timing point BPM but they pulse at constant 200BPM.

Wieku:
If I remember there were arguments about having reverse arrows pulsing with the song or at constant 200bpm, some people wanted to have pulse disabled. So maybe we can add two checkboxes to settings called `Enable slider reverse arrow pulse` and `Match arrow pulse to beatmap bpm`? I want to hear more voices about that.

peppy:
Settings and configuration for animations will come in the future as a skin drawable level setting. Implement a sane default to begin with.

Wieku:
Yeah, I removed the settings and kept repeat points synced to timing points. The only one concern I have with this is that inheriting `BeatSyncedContainer` causes repeat point to eat 1Kb memory more.

smoogipoo:
don't worry about memory consumption.

peppy:
I've added some changes to make the animation occur more frequently (to closer match stable but still beatsync).

peppy (dismissed):

peppy:
I've also removed some of the extra complexity of the animation you had (it was unnecessary and incorrectly animating in the case of the `Math.Min` - just let it play as it is).

peppy (dismissed):

Wieku:
That complexity was added to have reverse arrow smoothly scale to it's original size as slider ball gets close and to allow it to normally scale up to `1.5` after being hit. Without that check reverse arrow could scale up to `1.95`. That `if (Clock.CurrentTime < repeatPoint.StartTime)` check was added after some testing on my side.

Wieku:
I made a comparison:
* [How it looked before changes](https://streamable.com/omuq5)
* [How it looks now](https://streamable.com/2dql1)
* [How it looks when I change `timingPoint.BeatLength` to `TimeUntilNextBeat`](https://streamable.com/zr9o0)
* [Previous one, but we keep time and limiting clauses](https://streamable.com/0kaw9)

peppy:
The reverse arrow *should* scale up to 1.95. In lazer each sprite, object animation has a purpose. One should not change to make another look better. If we decide it doesn't look good, we should be changing the actual animation, not stopping it at a certain arbitrary time.

I think it looks fine.

smoogipoo (dismissed):

smoogipoo (approved):

Wieku:
In the 3rd video you can see that during some fade outs random reverse arrows suddenly get really big and I think it doesn't look well. It's visible during gameplay and probably players gonna report it as a glitch (that's why I mentioned I added those checks after playing for some time, it was looking way better and more responsive).

peppy:
It's normal. The reverse arrow is moving; the hitobject is exploded; they are additive.

Wieku:
After some more testing it is visible on slower maps. So we can leave it like that. But I think at least we should change `timingPoint.BeatLength` to `TimeUntilNextBeat` to have smoother animation.

peppy:
You don't want `TimeUntilNextBeat` but rather the final beatlength. If you use `TimeUntilNextBeat` and the game was to apply the transform at 50ms to the next beat it would do a full scale in 50ms, looking very unnatural.

Wieku:
Right, I forgot about that. So move `double beatLength` as a protected property or just sum `TimeUntilNextBeat` with `TimeSinceLastBeat` because the sum of this too is basically the final beatlength?

peppy:
I'd probably expose that in `BeatSyncedContainer`, but I'd rather that was done separately (the scope of changes on that class is already too large for this PR). If you do add it, it'd be nice to have test coverage of divisor and other untested pieces of `BeatSyncedContainer` too.

PR #7435, opened 2020-01-04
Remove unused LaneGlowPiece and GlowPiece

mcendu:


bdach:
Usages seem to have been removed back in 56ea1c1, so seems like this change *should* be ok. Rider also reports no usages for both.

Maybe this PR should come with enabling CI inspections for unused classes/members? Not sure if that's something that has been tried before.

smoogipoo (approved):
Let's just get this in as it is for now.

Issue #7417, opened 2020-01-01
Better offline support


bdach:
Opening as an epic issue to group together other outstanding issues.

There are a few places where offline isn't supported in the best way (blank overlays, endless spinners...) Below is a (currently non-exhaustive) list of known issues associated with being offline:

- [x] Loading animations not stopping (#933)
- [x] Non-functional chat (#3024)
- [x] Non-functional direct (#4550)
- [x] Non-functional changelog (reported as crashing in #7416, crashes no longer)
- [x] Non-functional room browser (#8803)

Latest general direction seems to be [blocking access to online-only views if possible](https://github.com/ppy/osu/issues/933#issuecomment-308322500) and [displaying placeholder text similar to leaderboards if not](https://github.com/ppy/osu/issues/4550#issuecomment-530269254), but that might have changed in the meantime so I'd appreciate corrections if it has.

mcendu:
I am not sure why we use epic issues instead of enabling projects.

peppy:
Projects are useful for a standard agile flow, which this is not.

Game4all:
Should all online overlays which require the user to be logged in exhibit this same behaviour ? If yes, logic could certainly be abstracted in a shared overlay class.

iiSaLMaN:
Regarding the implementation of this, I have an idea that could work out and hopefully make offline support pretty straightforward for upcoming online API components, by implementing a container that handles whether to show the content or the login placeholder:
```
SomeAPIComponent:
    OnlineContainer:
        Box

// Box will show up if user is logged in, else the placeholder (implemented inside OnlineContainer) will show up.

peppy:
Closing this as all of the relevant issues have been addressed via PRs or are covered via other issues.

PR #7438, opened 2020-01-04
Implement a login placeholder

Game4all:
# Summary
- This PR moves the Placeholder class and the MessagePlaceholder classes to a dedicated namespace (as placeholders aren't specific to leaderboards anymore) and also implements a placeholder asking for login which opens login overlay on click.
- This component is planned to be reused for upcoming PRs to refine user experience for indicating that login is required to access certain online-only views (#7417)

image

### DIFF
--- a/osu.Game/Online/Leaderboards/Leaderboard.cs
+++ b/osu.Game/Online/Leaderboards/Leaderboard.cs
@@ -151,7 +151,7 @@ namespace osu.Game.Online.Leaderboards
                         break;
 
                     case PlaceholderState.NotLoggedIn:
-                        replacePlaceholder(new MessagePlaceholder(@"Please sign in to view online leaderboards!"));
+                        replacePlaceholder(new LoginPlaceholder(@"view online leaderboards!"));
                         break;
 
                     case PlaceholderState.NotSupporter:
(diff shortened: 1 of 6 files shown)


osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -150,7 +151,7 @@ protected PlaceholderState PlaceholderState
                         break;
 
                     case PlaceholderState.NotLoggedIn:
-                        replacePlaceholder(new MessagePlaceholder(@"Please sign in to view online leaderboards!"));
+                        replacePlaceholder(new LoginPlaceholder(@"view online leaderboards!"));
```
huoyaoyuan: Concating message strings is not friendly for localisation. Leave here for further consideration.

Game4all: Removed the string concatenation from the constructor and put it back like it was before. Should be fine for now

peppy:
I was going to suggest making this control derive from OsuAnimatedButton to get the correct on-hover effect, but it turns out that due to the `Placeholder` parent class it's not so simple. Should be considered for a future iteration.

peppy (approved):

PR #7449, opened 2020-01-05
Update visible triangles colour when dark or light colour is changed

### LINKED ISSUE #7308: Changing Triangles colours do not affect existing particles
**Describe the bug:** Modifying colours of `Triangles` do not affect existing triangle particles shown on screen.

The code I used to change triangles colours is by implementing `IHasAccentColour` and set `ColourLight`, `ColourDark` respectively.
```csharp
public class HeaderButton : TriangleButton, IHasAccentColour
{
    private Color4 accentColour;

    public Color4 AccentColour
    {
        get => accentColour;
        set
        {
            accentColour = value;

            BackgroundColour = accentColour;
            Triangles.ColourLight = accentColour.Lighten(0.8f);
            Triangles.ColourDark = accentColour;
        }
    }
}

// button.FadeAccent(colours.Red, 500, Easing.InOutExpo);
```

**Screenshots or videos showing encountered issue:** https://streamable.com/ss12q

Wieku:
Fixes #7308


osu.Game/Graphics/Backgrounds/Triangles.cs
```diff
@@ -177,7 +199,17 @@ protected virtual TriangleParticle CreateTriangle()
         /// Creates a shade of colour for the triangles.
         /// </summary>
         /// <returns>The colour.</returns>
-        protected virtual Color4 CreateTriangleShade() => Interpolation.ValueAt(RNG.NextSingle(), ColourDark, ColourLight, 0, 1);
+        protected virtual Color4 CreateTriangleShade(float shade) => Interpolation.ValueAt(shade, colourDark, colourLight, 0, 1);
```
peppy: this should be moved inside the triangle itself, along with the random shade implementation.

osu.Game/Graphics/Backgrounds/Triangles.cs
```diff
@@ -264,6 +296,12 @@ protected struct TriangleParticle : IComparable<TriangleParticle>
             /// </summary>
             public Vector2 Position;
 
+            /// <summary>
+            /// The colour shade of the triangle.
+            /// This is needed for colour recalculation of visible triangles when <see cref="ColourDark"/> or <see cref="ColourLight"/> is changed.
+            /// </summary>
+            public float ColourShade;
```
peppy: does not need to be `public`. private should do.

osu.Game/Graphics/Backgrounds/Triangles.cs
```diff
@@ -29,8 +29,29 @@ public class Triangles : Drawable
...
-        public Color4 ColourLight = Color4.White;
-        public Color4 ColourDark = Color4.Black;
+        private Color4 colourLight = Color4.White;
+
+        public Color4 ColourLight
+        {
+            get => colourLight;
+            set
+            {
+                colourLight = value;
```
peppy: check equality and return early to avoid needless calls.

peppy (changes requested):

osu.Game/Graphics/Backgrounds/Triangles.cs
Wieku: So remove that function completely?

peppy: depends how it's being used in the `virtual` context, if at all.

peppy:
Sorry, I didn't realise `TriangleParticle` was a struct. The generation can stay where it was in that case (I will undo your last commit).

peppy (approved):

PR #7450, opened 2020-01-06
Fix osu!direct global action not being bound to a key

Joehuu:
Probably wasn't meant to be removed in https://github.com/ppy/osu/commit/18645fcf297325daa0391619775a4becbb538f4c#diff-4576bb7cce7105ff5b6aef3dd37e62ddL29.


peppy (approved):
oops

Joehuu:
I think this messed up the key bindings, so it needs a migration. Can't do that at the moment though.

iiSaLMaN:
Which key bindings are you referring to? the global key bindings seem to be working fine on latest master.

Joehuu:
Global key bindings.

Maybe it's because I was switching between `master` and release, or when I was implementing a new key binding for now playing... Lemme test on clean install -> latest master.

Joehuu:
> when I was implementing a new key binding for now playing

Yep, it was this.

PR #7458, opened 2020-01-06
Allow scrolling through DimmedLoadingLayer

EVAST9919:


smoogipoo:
Is there a reason for this?

peppy:
Definitely need context on where/why/what this is fixing.

EVAST9919:
Example: you can't scroll beatmap overlay while you are looking at scores and decided to change some parameters and waiting for new scores loaded. The same thing can be applied to rankings overlay: you can't scroll it while waiting for stuff to be loaded. 
To be clear: by unability to scroll mean you can't use scroll wheel while hovering loading layer.

bdach:
To clarify: by *scrolling through* you mean something like the following?
```
ParentScrollingContainer      // this will be scrollable with this PR while the child is being loaded
    ChildScrollingContainer   // this is being loaded
        DimmedLoadingLayer
```
I'd say this needs a test scene regardless.

EVAST9919:
Currently this layer blocks any input when visible, and that's expected. But I can't scroll beatmap overlay, while scores are loading and I'm hovering this layer. And in the future this will became a problem for other overlays too. So that's why I decided to allow scrolling.
Screenshot_1

peppy:
Have checked all scenarios where this is used, and believe this is better/more correct feeling behaviour.

peppy (approved):

Issue #4924, opened 2019-06-05
Skin-level settings


peppy:
Right now, there are quite a few pull requests coming through which implement new customisation options which don't fit at a game-wide level. They also don't fit well in ruleset-specific options, as those are generally considered player settings which you would never want overridden by a skin.

This will add databased skin settings, with user customisation via a UI. Note that this does not encompass a full skin editor (ie. sprite changing) but focuses on settings specifically – things you would find in `skin.ini` historically.

- [x] Depends on #4921
- [x] Depends on #4922

jorolf:
What about a way to sync these settings between skins? Having to switch them everytime you change the skin can be a bit repetitive.

peppy:
User overrides will not be attached to specific skins. You will also eventually be able to modify skins though, should you wish to change skin-specific settings. I have all bases covered.

peppy:
Deprioritising (editor is higher priority for now).

peppy:
The scope of this has changed slightly. The new settings will be part of the skin editor (exposed as a property dialog and using similar automatic setting logic as what we use for mods). Hoping to get to this in the next few months.

peppy:
This has since been implemented 🎉

PR #7459, opened 2020-01-06
Add the skin setting "Color the sliderballs according to the color of the sliders"

Azn9:
This is only a small visual modification, but which can be useful for some people. For example, in some beatmaps, the slider colors are very dark so the sliderball and the followcircle arn't easily viewable.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/SkinSection.cs
+++ b/osu.Game/Overlays/Settings/Sections/SkinSection.cs
@@ -64,7 +64,7 @@ namespace osu.Game.Overlays.Settings.Sections
                 },
                 new SettingsCheckbox
                 {
-                    LabelText = "Color the sliderballs according to the color of the sliders",
+                    LabelText = "Use combo colour as tint for slider ball",
                     Bindable = config.GetBindable<bool>(OsuSetting.ColourSliderBalls)
                 },
             };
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/Settings/Sections/SkinSection.cs
```diff
@@ -62,6 +62,11 @@ private void load(OsuConfigManager config, SkinManager skins)
                     LabelText = "Beatmap hitsounds",
                     Bindable = config.GetBindable<bool>(OsuSetting.BeatmapHitsounds)
                 },
+                new SettingsCheckbox
+                {
+                    LabelText = "Color the sliderballs according to the color of the sliders",
```
bdach: I'd follow stable text:
```suggestion
                    LabelText = "Use combo colour as tint for slider ball",
```

bdach (changes requested):
This doesn't work correctly if the setting is toggled in the middle of gameplay since there's no `BindValueChanged` on the added bindable and `ApplySkin` triggers only on skin changes.

bdach:
Also an open question if we even want to bring this over from stable in lieu of #4924 (might allow editing the skin setting instead of globally enabling/disabling tint).

Azn9:
A parameter depending on the current skin which would be activated or deactivated then? So be able to modify `OsuSkinConfiguration.AllowSliderBallTint`?

bdach:
> A parameter depending on the current skin which would be activated or deactivated then? So be able to modify `OsuSkinConfiguration.AllowSliderBallTint`?

That's what I'm thinking of, yes. Although I'm definitely not at liberty here to make an executive decision about that myself.

peppy:
Thanks for the contribution, but this is not the direction we're going in. As mentioned above, this would be a setting at the skin level, not global. The only global settings that should be present for skins are very high level decisions for the users, like "never/always use skin hitsounds".

PR #7460, opened 2020-01-06
Allow CommentsContainer refetch comments using a method

EVAST9919:
Split from https://github.com/ppy/osu/pull/6716


osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -121,12 +119,24 @@ private void load()
...
+        public void ShowComments(CommentableType type, long id)
         {
+            this.type = type;
+            this.id = id;
+            Sort.TriggerChange();
+        }
+
+        private void resetComments()
+        {
+            if (id == default)
```
huoyaoyuan: - using `default` for primitive type (long) looks bad
- will anywhere accept 0 for valid id? should it be `long?` ?
- if it's reset to show nothing, clear should be invoked

EVAST9919: Looks like reset is a bad name. Should be `refetch` instead (?)
Yeah, nullable long is a better thing to use

EVAST9919: Just a side note: we will never use this method manually in-game: all the id's will be used from overlays themselves, so it's not really about allowing or disallowing to use 0 for valid id. That's why I changed my code to use `long?` instead of checking `id==0`. But will see, hot it goes.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -121,21 +119,33 @@ private void load()
 
         protected override void LoadComplete()
         {
-            Sort.BindValueChanged(onSortChanged, true);
+            Sort.BindValueChanged(_ => refetchComments());
```
peppy: why is the `true` gone?

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: No reason to call it right away because we have no parameters anyway, I guess?

peppy: what?

EVAST9919: I mean, we have no reason to trigger `refetchComments() ` on load completed if we have no type nor id to use (I hope I explained it right now)

peppy: aren't you making this to be a usable component?

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -121,21 +119,33 @@ private void load()
...
         }
 
-        private void onSortChanged(ValueChangedEvent<CommentsSortCriteria> sort)
+        /// <param name="type">The type of resource to get comments for.</param>
+        /// <param name="id">The id of the resource to get comments for.</param>
+        public void ShowComments(CommentableType type, long id)
+        {
+            this.type = type;
+            this.id = id;
+            Sort.TriggerChange();
```
peppy: this should not trigger an arbitrary bindable. call the method directly.

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: true was there because we already have all the parameters in place (passed in the ctor) and we wanted to fetch comments right away when the container has been loaded. Now we don't have any of these so I removed it.

peppy: So what happens if the `Show()` method is called before `LoadComplete()`?

EVAST9919: Crash due to not loaded api

EVAST9919: I see: we can add `return` if not loaded and add true here, thanks for good example

peppy:
As an aside, an ID of 0 is never used by any database engine. Not required to make a change here but just noticed your reasoning for using nullable so thought I'd mention it.

peppy (approved):

PR #7464, opened 2020-01-07
Remove unused variable on DifficultyIcon

GSculerlor:
Found it when looking at this file. Seems like leftover from first implementation of difficulty tooltip


bdach:
Looks ok to me. Property introduced in 515534c and remained after switch to new style in 50046d5.

I looked into enabling `UnusedType` and `UnusedMembers` r# inspections. Unfortunately they're not immediately useful right now due to catching 66 errors in total, some of which were unused getters and test code.

bdach (approved):

PR #7466, opened 2020-01-07
Make ChannelTabControl switch to the next channel upon closing an active one

### LINKED ISSUE #7465: ChannelTabControl not handling channel closing correctly
Right now after closing a currently active channel, [ChannelTabControl](https://github.com/ppy/osu/blob/master/osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs) always switches to the channel on its left (the **previous** one). However, as indicated by the comments in the code, the desired behavior would be for it to always switch to the **next** one (on its right side), only exception being if it was the last channel on the list.

This is also mentioned by ppy/osu-framework#3152, though as I'm not experienced enough to investigate this issue framework-wise, I wanted to focus on this specific scenario here.

This problem seems to be caused by conflicting `RemoveChannel` and `tabCloseRequested` functions. 

`tabCloseRequested` seems to launch first after closing a channel, switching the tab if the closed one was currently active. However, `TabContainer.IndexOf(tab)` returns the index counting from the right side of the channel list, making the tab selecting part underneath work incorrectly (it assumes counting from the left):

```csharp
if (tab == SelectedTab && totalTabs > 1)
     // Select the tab after tab-to-be-removed's index, or the tab before if current == last
     SelectTab(TabContainer[currentIndex == totalTabs ? currentIndex - 1 : currentIndex + 1]);
``` 

After this, `RemoveChannel`'s `Current.Value == channel` check is never gonna be true due to `tabCloseRequested` changing the tab beforehand. Even if it did execute, it seems to only set the `Current` value instead of using `SelectTab()`, so it seems kinda obsolete.

I can fix this, but I'm not sure which of these functions is desired to handle the switching and if the order of items being reversed (as indicated by the `IndexOf` call) is intended.

thewildtree:
Fixes #7465 

- [x] Crash needs to be resolved
- [ ] Depends on https://github.com/ppy/osu-framework/issues/3169
- [ ] Depends on [additional framework-side changes](https://github.com/ppy/osu/pull/7466/#issuecomment-574454511)

There's still one issue - current implementation crashes if I close the last visible channel while there are more channels opened, hidden in the dropdown menu. 

It seems like moving the logic to [`RemoveChannel`](https://github.com/TheWildTree/osu/blob/aac9e70f3fb52d4ebb9751bf49e89aeec8b8adca/osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs#L79) (putting it after the [`RemoveItem`](https://github.com/TheWildTree/osu/blob/aac9e70f3fb52d4ebb9751bf49e89aeec8b8adca/osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs#L81) call) would solve it, as the next channel should become visible by then, but that could still crash if the next channel still didn't have enough space to appear for some reason. Or maybe it should just switch to the previous channel instead, like it does if we're closing the last available channel - what would be the better solution here?


bdach (dismissed):
I'm not sure I'm fine with this resolution. It feels like papering over what is arguably a framework bug (`IndexOf` not returning the index counting from the left).

Additionally this behaviour, as well as the scenario you describe to be crashing and which needs a resolution, should be tested.

TheWildTree:
It doesn't look good for me either, hence why I asked about it in the linked issue. I'll try to work on tests,  maybe in the meanwhile others will share their insights here.

smoogipoo:
This is because `OsuTabControl` defines a custom tab flow that has a reverse-ChildID comparer, whereas `FlowContainer` always compares by non-reversed `ChildID`. I'm a little confused as to why `OsuTabControl` is doing this in the first place since it's a change that goes back 2 years (https://github.com/ppy/osu/pull/2175), and we may be able to remove it or consider an alternative.

Going forward it would be really nice if we could remove the whole custom comparer thing for `InternalChildren` / `AliveInternalChildren`, and maintain a separate custom-ordered list for draw-time. But that sort of change is quite a ways off and needs special memory/performance considerations.

peppy:
How do you get into the crash state? A very thin window?

peppy:
@TheWildTree rather than updating the branch with master, it would be appreciated if you can answer the question.

TheWildTree:
Sorry, I didn't mean to push that here, my lack of knowledge about git shows.

The crash that's mentioned in the first post happens if you have enough channels opened for the dropdown menu to show, if you then close the last channel visible on the tab list while it's active, the game will crash, because the index here will be out of bounds:

```csharp
SelectTab(TabContainer.TabItems.ElementAt(currentIndex == (totalChannels - 1) ? currentIndex - 1 : currentIndex + 1));
```
The second possible crash I mentioned was purely hypothetical and I was a bit busy so I haven't gotten around to testing that yet, sorry.

TheWildTree:
I moved the switching logic to `RemoveChannel()` (where it was probably originally intended to be), but it doesn't work as expected - `RemoveItem()` call doesn't update the layout, so after removing the last (rightmost) visible tab while there are channels in the dropdown, the first channel from there is still not visible until after `RemoveChannel()` finishes. That leaves it to either fall back to switching to previous channel in that case, or :

- Using `SwitchTab(1)`, but that won't work since the tab we want to switch to is not yet visible (it's still in the dropdown)
- Using 'SelectTab()`, but that won't work for the same reason (TabItems for channels from the dropdown doesn't seem to be accessible from anywhere, or I'm blind and can't see the obvious way to achieve that)
- Using `Current.Value`, but that feels... ugly, because I suppose `SwitchTab`/`SelectTab` exist for a reason

I've temporairly tried using the third option, like this:

```csharp
public void RemoveChannel(Channel channel)
{
    if (Current.Value == channel)
    {
        var itemsList = Items.ToList();
        var dropdownItemsCount = itemsList.Except(VisibleItems).Count();
        var currentIndex = itemsList.IndexOf(channel);
        var isNextTabSelector = itemsList.ElementAt(currentIndex + 1) == selectorTab.Value;

        // If the tab being closed is the last one visible and the '+' is in the dropdown menu
        // the '+' is always gonna be the next tab, even if there are other channels in there
        if (dropdownItemsCount > 1 && isNextTabSelector)
        {
            // this is very temporary and ugly
            Current.Value = Items.ElementAt(currentIndex + 2);
            // this would prevent selectorTab from popping up in case the switching goes wrong
            selectorTab.Active.Value = false;
        }
        else
            SwitchTab(isNextTabSelector ? -1 : 1);
    }
    RemoveItem(channel);
}
```

 and it worked about half of the time, but the other half results in the **channel that's being 'removed' ending up at the bottom of the dropdown** and the channel selector popping up.

I've tried looking into it, and it seems that the order of the dropdown entries in `TabControl.Items` is different from what's displayed - here's a comparison:

image
(dropdown matches the order in which these were added via the selector)

So, as ugly (and bug-prone, as visible) switching with `Current.Value` is, it would work fine if it wasn't for those inconsistencies, but I'm not skillled enough to investigate this further, as it seems to be a framework bug. The '+' also always sits 1st in the dropdown, but I guess that might be intended.

Overall, to make the forward switching possible in this particular case, the visible tab layout would have to be updated right after the `RemoveItem()` call, instead of after `RemoveChannel()` finishes, so we could easily use `SwitchTab` as it was intended to. Or, to make this 'ugly' version work, a fix for the items order is needed.

TheWildTree:
Implemented this in a slightly different way, I'm not sure if using `UpdateSubTree()` is the proper way to force the tab layout to refresh before switching, but it works. I'll write test as soon as possible, but I don't have much experience with that, so please give me some time.

Also found three bugs in the meantime:

- the order of the dropdown entries in `TabControl.Items` is sometimes different from what's displayed - here's a comparison:

image
              (dropdown matches the order in which these were added via the selector)

- the selectorTab.Active stays true even if it's not actually in use:
image

- sometimes an already closed channel pops back up on the list after closing another one, but I haven't found a way to consistently reproduce it. Might be related to the first bug.

osu.Game.Tests/Visual/Online/TestSceneChatOverlay.cs
```diff
@@ -96,6 +100,58 @@ public void TestCloseChannelWhileSelectorClosed()
...
+                    chatOverlay.ChannelTabControl.RemoveChannel(VisibleChannels.Last());
+                    chatOverlay.ChannelTabControl.UpdateSubTree();
+                    return true;
+                }
+                else return false;
+            });
+
+            // Closing the last channel with dropdown no longer present
+            AddStep("Close last when selector next", () => chatOverlay.ChannelTabControl.RemoveChannel(VisibleChannels.Last()));
+            AddAssert("Channel changed to previous", () => channelManager.CurrentChannel.Value == VisibleChannels.Last());
```
TheWildTree: In the previous check, I needed to explicitly save the next channel for comparison, since it's in the dropdown menu, but here I check by just assuming that the previous channel will become the last one after removing the current last channel. Is that okay or should this have more explicit checks?

peppy: Should be fine.

peppy:
I've tidied up your tests and logic, removing the `UpdateSubTree` calls and a lot of the complexity overhead.

Note that https://github.com/ppy/osu-framework/issues/3169 will need to be fixed for this to work, but I see this as the correct way forward.

bdach (commented):
This is a lot better in its current state to me. Won't approve outright until the framework bump & CI pass, but at least code-wise looks good to me 👍

TheWildTree:
Thanks @peppy, your simplified version + the framework-side change make much more sense than the running around in circles I did. 

Small fixes were needed though:

- removed the reverse-ChildID comparer from `OsuTabControl.TabContainer` ([mentioned earlier](https://github.com/ppy/osu/pull/7466#issuecomment-572363139)) since your framework-side change assumes non-reversed item order. As far as I see, in-game `TabControl`s show no differences after this change.
- had to de-simplify the logic a bit, because your version has made one edge-case fail

bdach:
> had to de-simplify the logic a bit, because your version has made one edge-case fail.

That edgecase should then be tested, if it isn't already.

TheWildTree:
It is already being tested, that's how I noticed it was broken after the changes.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
```diff
@@ -88,7 +88,13 @@ public void RemoveChannel(Channel channel)
                 if (nextTabIsSelector && items.Count == 2)
                     SelectTab(selectorTab);
                 else
-                    SwitchTab(nextTabIsSelector ? -1 : 1);
+                {
+                    // If there are any channels in the dropdown (one item more means only selectorTab being there, so we need two)
```
peppy: i can’t understand this

peppy:
i think you are mistaken. are you actually checking with a local framework checkout and my framework PR?

TheWildTree:
yes, I checked this with your framework PR. the reasoning behind this is that if you're closing the last visible channel (while there are more channels in the dropdown), the next tab is **always** gonna be the selectorTab, because it's always first in the dropdown. Your version didn't account for that:

```csharp
SwitchTab(nextTabIsSelector ? -1 : 1);
```

and kept switching to the previous channel in that case.

osu.Game/Overlays/Chat/Tabs/ChannelTabControl.cs
TheWildTree: poor wording on my part, sorry.

below, the 'VisibleItems.Count() < Items.Count()' is a check for the dropdown having any items, but only 1 element difference means it's the `selectorTab` in there and not an actual channel. our edge-case only happens if we *do* have channels there, hence the check for one more element.

smoogipoo:
The change to removing the custom comparer has caused the shadows in the chat tab control to be flipped.

Master:
image

This branch:
image

Will hold off merging until the above is fixed.

There are a few things that I think need to be fixed/defined:
1. The reverse-order comparer shouldn't be done for general `OsuTabControl`. It's a very special case for chat specifically, and should be moved locally.
2. Tab controls should switch in the order that tabs are displayed to the user. Essentially this means flowing in the order of `FlowingChildren`.
3. It's not trivial to use `FlowingChildren` directly, two problems are presented:
    1. `FlowingChildren` doesn't contain the dropdown children, could be fixed via a manual concatenation, BUT:
    2. Tab controls may not have dropdowns, and it seems perfectly sane to be able to switch to tabs even if they're not displayed. It would be nice if we could use `FlowingChildren`, but it enumerates only `IsPresent = true` children. Perhaps something we should consider is changing `FlowContainer` to internalise its `FlowingChildren` lookup, and consider presence as a post-processing step inside the child position computation.

TheWildTree:
Shadows fixed by readding the reverse-order comparer (but only to ChannelTabControl) + inversing the SwitchTab direction to account for that. I'm not sure whether the override I added is a good way to do that, if it should have its own xmldoc, or maybe it should be done in a different way?

smoogipoo:
I previously mentioned that it should use the order in which items are flowed on screen, so the `SwitchTab` override shouldn't be required. It depends on framework changes.

TheWildTree:
Sorry, I misunderstood that part. Removed.

TheWildTree:
The logic in this PR is completely broken since the ppy/osu-framework#3183 change + needs other framework-side changes which I don't have time to work on atm. 

I think it'll be better to close this for now, I'll PR some smaller changes separately (moving the reverse-order comparer to ChannelTabControl, for example) and come back to this issue when we have the framework side of things figured out.

PR #7470, opened 2020-01-08
Fix track speed not being reset after exiting the editor

### LINKED ISSUE #7468: Track speed modification in editor is applied globally
Connecting it to the epic: #6686

When we slow down the song in the editor to 50% or 75% and we exit the editor, we are able to play a map with that speed: https://streamable.com/sncb9

Source: https://osu.ppy.sh/comments/1004130

Wieku:
Fixes #7468

### DIFF
--- a/osu.Game/Screens/Edit/Components/PlaybackControl.cs
+++ b/osu.Game/Screens/Edit/Components/PlaybackControl.cs
@@ -5,6 +5,8 @@ using System.Linq;
 using osuTK;
 using osuTK.Graphics;
 using osu.Framework.Allocation;
+using osu.Framework.Audio;
+using osu.Framework.Bindables;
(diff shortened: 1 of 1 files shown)


smoogipoo:
This isn't the correct direction for this. Since `WorkingBeatmap` has now been moved to `Editor`, we can add a `GameplayClockContainer` to the editor and have the tabs bound to `UserPlaybackRate`.

There's a lot that this PR isn't doing, such as not removing the bindable on exit, for which `Editor` would have to call `GameplayClockContainer.StopUsingBeatmapClock()` as per what's done in `Player`.

Basically just look towards `Player` for implementation references.

Wieku:
Never thought about this, I will look into this later

PR #4523, opened 2019-03-24
Add intensive benchmark

peppy:
Still requires a bit more work before mergeable, but is quite useful as-is (used for #4520 testing)

PR #7474, opened 2020-01-08
Add benchmark project

huoyaoyuan:
The same purpose with ppy/osu-framework#3038 . To benchmark osu! specific stuff.

I can run it with VS and VSCode, but my Rider says it can't build solution. May need investigation.

##### About beatmap and storyboard parsing benchmarks
I choose https://osu.ppy.sh/beatmapsets/261911#osu/736300 as a famous storyboard-full map.
The memory performance is terrible now.

```
BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i7-8700K CPU 3.70GHz (Coffee Lake), 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=3.1.100
  [Host]     : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT
  DefaultJob : .NET Core 3.1.0 (CoreCLR 4.700.19.56402, CoreFX 4.700.19.56404), X64 RyuJIT
```

|                    Method |       Mean |     Error |    StdDev |      Gen 0 |     Gen 1 |    Gen 2 | Allocated |
|-------------------------- |-----------:|----------:|----------:|-----------:|----------:|---------:|----------:|
|    BenchmarkLegacyBeatmap |   4.531 ms | 0.0877 ms | 0.0939 ms |   500.0000 |  250.0000 |        - |   3.01 MB |
| BenchmarkLegacyStoryboard | 100.009 ms | 0.5818 ms | 0.4858 ms | 10200.0000 | 2200.0000 | 600.0000 |  57.86 MB |

### DIFF
--- a/osu.Game.Benchmarks/BenchmarkBeatmapParsing.cs
+++ b/osu.Game.Benchmarks/BenchmarkBeatmapParsing.cs
@@ -1,0 +1,37 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.IO;
+using BenchmarkDotNet.Attributes;
+using osu.Framework.IO.Stores;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.Formats;
+using osu.Game.IO;
+using osu.Game.IO.Archives;
+using osu.Game.Resources;
+
+namespace osu.Game.Benchmarks
+{
+    public class BenchmarkBeatmapParsing : BenchmarkTest
+    {
+        private readonly MemoryStream beatmapStream = new MemoryStream();
+
+        public override void SetUp()
+        {
+            using (var resources = new DllResourceStore(OsuResources.ResourceAssembly))
+            using (var archive = resources.GetStream("Beatmaps/241526 Soleily - Renatus.osz"))
+            using (var reader = new ZipArchiveReader(archive))
+                reader.GetStream("Soleily - Renatus (Gamu) [Insane].osu").CopyTo(beatmapStream);
+        }
+
+        [Benchmark]
+        public Beatmap BenchmarkBundledBeatmap()
+        {
+            beatmapStream.Seek(0, SeekOrigin.Begin);
+            var reader = new LineBufferedReader(beatmapStream); // no disposal
+
+            var decoder = Decoder.GetDecoder<Beatmap>(reader);
+            return decoder.Decode(reader);
+        }
+    }
+}
--- a/osu.Game.Benchmarks/osu.Game.Benchmarks.csproj
+++ b/osu.Game.Benchmarks/osu.Game.Benchmarks.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+</Project>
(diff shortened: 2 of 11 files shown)


peppy:
Will finally give me a place to put by [old benchmark PR](https://github.com/ppy/osu/pull/4523).

osu.Game.Benchmarks/osu.Game.Benchmarks.csproj
```diff
@@ -0,0 +1,19 @@
...
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.12.0" />
+    <PackageReference Include="nunit" Version="3.12.0" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.16.0" />
```
smoogipoo:
```suggestion
    <PackageReference Include="NUnit3TestAdapter" Version="3.15.1" />
```

See: https://youtrack.jetbrains.com/issue/RIDER-37769 . Hinders our workflow for now.

osu.Game.Benchmarks/BenchmarkBeatmapParsing.cs
```diff
@@ -0,0 +1,37 @@
...
+            {
+                var decoder = Decoder.GetDecoder<Beatmap>(reader);
+                return decoder.Decode(reader);
+            }
+        }
+
+        [Benchmark]
+        public Storyboard BenchmarkLegacyStoryboard()
+        {
+            using (var stream = Assembly.GetExecutingAssembly().GetManifestResourceStream(typeof(Program), "Resources.MitiS & MaHi - Blu (Speed Up Ver.) (Ashasaki).osb"))
```
smoogipoo: We generally don't include full beatmaps/storyboards for copyright reasons. Would prefer if the ones already included in the game/test resources were used instead. Eventually we would construct beatmaps/storyboards on a per-case basis and drill down these benchmarks.

smoogipoo (changes requested):

osu.Game.Benchmarks/BenchmarkBeatmapParsing.cs
peppy: Maybe worthwhile checking out the way I did it in the PR i mentioned. It generates a beatmap with enough content to be an intensive benchmark without using existing beatmaps.

huoyaoyuan: I'm fine to auto generate beatmaps and storyboards, with several points:
- I need to know specification of legacy beatmap files to generates text that's to be parsed from.
- Patterns coverage.
- A handful entry to run benchmark on ranked maps.

smoogipoo: 1) Create a beatmap and encode it with a `LegacyBeatmapEncoder`. It doesn’t support storyboards right now, though. I’d be fine with omitting it until that’s implemented.
2) Should be made on a per-case basis, like decoding a slider, or decoding a hitcircle, or a hitsound. As above you can just place many of these and encode them to have an “intensive” test.
3) Would love to see something like the beatmap conversion tests, where you can add a new test case based on the file name. That should be enough to then be able to separate out the problematic part and make a custom test for it.

huoyaoyuan:
I tried to write some simple random generation but fails. Keeping it simple now because my purpose was benchmarking other stuff.

Rewritten the history so that the actual beatmap doesn't appear in history. Running on the beatmap bundled in resources instead.

peppy (approved):
seems fine for now

PR #7479, opened 2020-01-08
Implement CountryFilter component for RankingsOverlay

EVAST9919:
Screenshot_1

### DIFF
--- a/osu.Game/Overlays/Rankings/CountryPill.cs
+++ b/osu.Game/Overlays/Rankings/CountryPill.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Graphics.UserInterface;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.Sprites;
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/Rankings/CountryPill.cs
```diff
@@ -0,0 +1,149 @@
...
+            ClearTransforms();
+            AutoSizeAxes = Axes.X;
+            this.FadeIn(duration, Easing.OutQuint);
+        }
+
+        public void Collapse()
+        {
+            ClearTransforms();
+            AutoSizeAxes = Axes.None;
+            this.ResizeWidthTo(0, duration, Easing.OutQuint);
```
bdach: Is this necessary? It doesn't seem to be to me, and it ticks me off how this never gets restored in `Expand()`.

bdach (commented):
Can't see too much fault with this; should be mostly fine.

osu.Game/Overlays/Rankings/CountryPill.cs
EVAST9919: It's restored by itself when setting `AutosizeAxes = Axes.X` including the fact that we have `AutosizeDuration` and `AutosizeEasing` in the ctor.
We already have some UI components with similar logic (`PlayerSettings` for example)

EVAST9919: If by "necessary" you mean this Collapse/Expand animation then no, it's not. I was just thinking about what kind of animation we can apply here. Result seems pretty good to me, though.

bdach: No, I just meant the resize call. Unless I screwed something up I didn't see any visual difference after removing it.

EVAST9919: I just checked: there's a difference after removing it. May be it's not really noticeable.

smoogipoo:
Required changes:
1. Using `BindableWithCurrent`.
2. Making things not inherit from `Container` or `CircularContainer`.

smoogipoo (approved):

bdach (approved):

PR #7488, opened 2020-01-10
Implement SpotlightSelector component for RankingsOverlay

EVAST9919:


osu.Game/Online/API/Requests/Responses/APISpotlight.cs
```diff
@@ -0,0 +1,34 @@
...
+        public bool ModeSpecific;
+
+        [JsonProperty(@"start_date")]
+        public DateTimeOffset StartDate;
+
+        [JsonProperty(@"end_date")]
+        public DateTimeOffset EndDate;
+
+        [JsonProperty(@"participant_count")]
+        public int? ParticipiantCount;
```
bdach:
```suggestion
        public int? ParticipantCount;
```

osu.Game/Online/API/Requests/GetSpotlightsRequest.cs
```diff
@@ -0,0 +1,20 @@
...
+using osu.Game.Online.API.Requests.Responses;
+
+namespace osu.Game.Online.API.Requests
+{
+    public class GetSpotlightsRequest : APIRequest<SpotlightsArray>
+    {
+        protected override string Target => "spotlights";
+    }
+
+    public class SpotlightsArray
```
bdach: Sort of misleading name, I'd go for
```suggestion
    public class SpotlightsCollection
```

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        private void onSelectionChanged(ValueChangedEvent<APISpotlight> spotlight)
+        {
+            startDateColumn.Value = dateToString(spotlight.NewValue.StartDate);
+            endDateColumn.Value = dateToString(spotlight.NewValue.EndDate);
+        }
+
+        private string dateToString(DateTimeOffset date) => $"{date.Year}-{date.Month:D2}-{date.Day:D2}";
+
+        private class InfoCoulmn : FillFlowContainer
```
bdach:
```suggestion
        private class InfoColumn : FillFlowContainer
```

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -0,0 +1,165 @@
...
+            api.Queue(request);
+        }
+
+        private void onSelectionChanged(ValueChangedEvent<APISpotlight> spotlight)
+        {
+            startDateColumn.Value = dateToString(spotlight.NewValue.StartDate);
+            endDateColumn.Value = dateToString(spotlight.NewValue.EndDate);
+        }
+
+        private string dateToString(DateTimeOffset date) => $"{date.Year}-{date.Month:D2}-{date.Day:D2}";
```
bdach: That's a very roundabout way of writing
```suggestion
        private string dateToString(DateTimeOffset date) => date.ToString("yyyy-MM-dd");
```

bdach (changes requested):
Visually the dropdown header text could use more padding from the sides, although I do know that it's not something immediately accessible or modifiable.

bdach (dismissed):
I don't think I have any issues with this at its current state anymore.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -0,0 +1,165 @@
...
+            background.Colour = colours.GreySeafoam;
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            SelectedSpotlight.BindValueChanged(onSelectionChanged);
+        }
+
+        public void FetchSpotlights()
```
smoogipoo: The selector should not be doing the web request. It should:
1. Take in a set of spotlights.
2. Handle selecting a spotlight.
3. Return a spotlight.

The web request would be part of the overlay, and in this case the test scene providing the spotlights (the test scene should provide a custom set of spotlights).

smoogipoo (changes requested):

osu.Game/Overlays/Rankings/SpotlightSelector.cs
EVAST9919: Got it

smoogipoo:
Not sure where you got `participant_count` from but it's not returned by the API. It, along with the maps, will be returned in a separate request.

I've updated code to reflect these changes (remove said columns), and made a few refactorings.

@peppy quick check plz

EVAST9919:
@smoogipoo it's true that this will be returned with a separate request, but it's still a part of the spotlight
Screenshot_1

bdach (dismissed):
Looks better to me for sure. I kind of got used to some test scenes being tightly coupled to API at this point, but I shouldn't have done so as this is the more correct approach.

bdach (approved):

smoogipoo (approved):
Seems good

Issue #7489, opened 2020-01-10
Missing rounded corners in status selection dropdown menu


recapitalverb:
**Describe the bug:** There are missing rounded corners in the status selection dropdown menu.

**Screenshots or videos showing encountered issue:** 
![Screenshot_20200110_215945](https://user-images.githubusercontent.com/41869184/72163082-b54dcb80-33f5-11ea-82c6-2c0f5b6a8b32.png)

**osu!lazer version:** master branch commit f745d746666031b785abf88c0b0d8062f0103694

**Logs:**
[database.log](https://github.com/ppy/osu/files/4046258/database.log)
[network.log](https://github.com/ppy/osu/files/4046260/network.log)
[performance.log](https://github.com/ppy/osu/files/4046261/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4046262/runtime.log)

smoogipoo:
This is by design, but the rounded style should match every other dropdown in the game:

![image](https://user-images.githubusercontent.com/1329837/72166786-3d8b9b00-340d-11ea-99f5-a545c438bb3b.png)

I.e. The corners should be larger and it should be inset.

PR #7491, opened 2020-01-11
Fix user status dropdown having no padding around items

Joehuu:
Fixes https://github.com/ppy/osu/issues/7489.


EVAST9919:
I don't think this is what this issue is about.
The issue is that the menu itself don't have rounded corners

Joehuu:
In https://github.com/ppy/osu/issues/7489#issuecomment-573094514, I thought smoogipoo said the non-rounded menu was the design, and the fix was making the padding the same on all dropdowns.

recapitalverb:
Yeah, I should have edited my issue to correctly address the real issue.

peppy (approved):

smoogipoo:
Could apply a similar removal to the corner radius (and margin?). It's smaller here than in other dropdowns iirc.

Joehuu:
Yep, there are all kinds of properties being changed in these dropdowns. Will standardize them.

PR #7492, opened 2020-01-11
Implement SongTicker component in main menu

### LINKED ISSUE #7274: Add song ticker to main menu
**Describe the new feature:** (This is something that's in stable, but not yet in lazer.) A ticker that would show up briefly (or indefinitely, if the user prefers) as a new song starts playing in the main menu.

image

EVAST9919:
Fixes https://github.com/ppy/osu/issues/7274

Screenshot_3


osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -226,15 +236,17 @@ public override void OnSuspending(IScreen next)
...
-            this.MoveTo(new Vector2(-800, 0), FADE_OUT_DURATION, Easing.InSine);
+            buttonsContainer.MoveTo(new Vector2(-800, 0), FADE_OUT_DURATION, Easing.InSine);
 
             sideFlashes.FadeOut(64, Easing.OutQuint);
         }
 
         public override void OnResuming(IScreen last)
         {
             base.OnResuming(last);
+            songTicker.Hide();
```
peppy: this shouldn't be necessary

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -263,6 +275,7 @@ public override bool OnExiting(IScreen next)
 
             buttons.State = ButtonSystemState.Exit;
             this.FadeOut(3000);
+            songTicker.Hide();
```
peppy: nor should this

osu.Game/Screens/Menu/SongTicker.cs
```diff
@@ -0,0 +1,72 @@
...
+namespace osu.Game.Screens.Menu
+{
+    public class SongTicker : Container
+    {
+        private const int fade_duration = 800;
+
+        [Resolved]
+        private Bindable<WorkingBeatmap> beatmap { get; set; }
+
+        private readonly Bindable<WorkingBeatmap> workingBeatmap = new Bindable<WorkingBeatmap>();
```
peppy: this is not how bindables work. you do not need this field.

osu.Game/Screens/Menu/SongTicker.cs
```diff
@@ -0,0 +1,72 @@
...
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            workingBeatmap.BindTo(beatmap);
+            workingBeatmap.BindValueChanged(onBeatmapChanged);
+        }
+
+        private void onBeatmapChanged(ValueChangedEvent<WorkingBeatmap> working)
+        {
+            if (working.NewValue?.Beatmap == null)
```
peppy: workingbeatmap can't ever be null (check game/gamebase implementation)

osu.Game/Screens/Menu/SongTicker.cs
```diff
@@ -0,0 +1,72 @@
...
+                return;
+
+            var metadata = working.NewValue?.Metadata;
```
peppy: remove null coalesce. it's not valid as above, and also having it present makes the next two lines 100% incorrect.

peppy (changes requested):

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -69,12 +69,22 @@ public class MainMenu : OsuScreen
...
+        private ParallaxContainer buttonsContainer;
+        private SongTicker songTicker;
+
         [BackgroundDependencyLoader(true)]
         private void load(DirectOverlay direct, SettingsOverlay settings, OsuConfigManager config, SessionStatics statics)
         {
             holdDelay = config.GetBindable<float>(OsuSetting.UIHoldActivationDelay);
             loginDisplayed = statics.GetBindable<bool>(Static.LoginOverlayDisplayed);
 
+            AddInternal(songTicker = new SongTicker
```
iiSaLMaN: I don't like how the children are added in splitted way, makes the code unnecessarily long, mabye add this to the `AddRangeInternal` below.

frenzibyte (changes requested):

osu.Game/Screens/Menu/SongTicker.cs
iiSaLMaN: is this due to `SongTicker` never getting removed?

osu.Game/Screens/Menu/MainMenu.cs
EVAST9919: song ticker can be triggered even in song select screen. So if you are in there, select another beatmap and go back to main menu - you will see the song ticker. That's why I'm hiding it here

peppy: it shouldn't be processing changes when not at the menu.

osu.Game/Screens/Menu/MainMenu.cs
EVAST9919: And here It can be visible too

osu.Game/Screens/Menu/MainMenu.cs
peppy: and how does it even become visible again then, after returning from song select?

osu.Game/Screens/Menu/SongTicker.cs
EVAST9919: In one pr someone told me to use local bindable rather than using `BindValueChanged` for resolved one.

osu.Game/Screens/Menu/SongTicker.cs
EVAST9919: saw that in the MusicController. May be I misunderstood something

osu.Game/Screens/Menu/SongTicker.cs
peppy: https://github.com/ppy/osu-framework/blob/master/osu.Framework/Allocation/ResolvedAttribute.cs#L115

osu.Game/Screens/Menu/SongTicker.cs
peppy: it's unnecessary there too. see https://github.com/ppy/osu/blob/master/osu.Game/OsuGameBase.cs#L208

osu.Game/Screens/Menu/MainMenu.cs
EVAST9919: No. It should be below `ExitConfirmOverlay`

osu.Game/Screens/Menu/MainMenu.cs
EVAST9919: Good question. Exiting song select without that line:
nice

peppy: As long as you understand it's a question you need to answer. The `Hide` as it stands is not correct.

EVAST9919: Got it, investigating

EVAST9919: Simplest (but not sure if correct) solution: I just added `AllowUpdates` flag to song ticker.

iiSaLMaN: You're probably looking for `mainMenu.IsCurrentScreen()`, or even more simpler, checking if the ticker `IsPresent`.

EVAST9919: `IsPresent` will not work because even on the main screen we are triggering changes while fully invisible

iiSaLMaN: Right, in that case, pass the menu and check `IsCurrentScreen()`.

EVAST9919: I personally don't think passing a screen for the component is a good idea. What if (for example) it will be implemented somewhere else? There may be other conditions.

iiSaLMaN: That'll be added later, currently it is used once at the main menu and it should not update when it's out of the screen, no need to make it that complex in my opinion.

osu.Game/Screens/Menu/MainMenu.cs
EVAST9919: That's why this exactly one is required
https://streamable.com/jecpp

peppy (approved):

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -107,8 +96,28 @@ private void load(DirectOverlay direct, SettingsOverlay settings, OsuConfigManag
...
+                    Action = () =>
+                    {
+                        if (holdDelay.Value > 0)
+                            confirmAndExit();
+                        else
+                            this.Exit();
+                    }
+                });
+            }
+
```
Joehuu: You broke `ButtonSystem`'s `OnPressed` when moving this.

peppy: aaaaaaaa i forgot to push 
image

PR #7494, opened 2020-01-11
Fix default button absorbing drag scroll on settings

Joehuu:


peppy (approved):
Not sure why this was present

PR #7496, opened 2020-01-11
Update localisable text usages

huoyaoyuan:
Depends on:
- [ ] ppy/osu-framework#3157
- [ ] Update framework

Humanizer requires some further attention. Others are fine.


osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs
```diff
@@ -256,16 +256,18 @@ public ScoreBeatmapMetadataContainer(BeatmapInfo beatmap)
...
+                        $"{beatmap.Metadata.Title ?? beatmap.Metadata.TitleUnicode} [{beatmap.Version}] ",
+                        $"{beatmap.Metadata.TitleUnicode ?? beatmap.Metadata.Title} [{beatmap.Version}] "),
                     Font = OsuFont.GetFont(size: 14, weight: FontWeight.SemiBold, italics: true)
                 },
                 new OsuSpriteText
                 {
                     Anchor = Anchor.BottomLeft,
                     Origin = Anchor.BottomLeft,
-                    Text = "by " + new LocalisedString((beatmap.Metadata.ArtistUnicode, beatmap.Metadata.Artist)),
+                    // TODO: we don't have the capacity to localise "by " now
```
huoyaoyuan, line 268: Localising these are quite complex because of language orders.

For example, in English it's "created by XXX", and ignoring the verb is acceptable.
But in Chinese, the order is "by XXX created", and ignoring the verb is non-natural.

`string.Format` with changing the position of `{0}` placeholders are designed for this, but not enough, because our artist titles have variations.

Use two separated translatable entries `Sentence1Prefix` and `Sentence1Suffix` is very verbose, and cannot handle scenarios with multiple parameters.

huoyaoyuan: I do have an idea for supporting this at framework level, but I think it should be done in a following pass.

PR #7501, opened 2020-01-11
Add global action for toggling now playing overlay

Joehuu:
Added the new global action to the bottom of enum to avoid migrations. If organizing is needed, I'll try to do it. If not, the comments in the enum can be removed.


peppy (approved):
Looks to work correctly

PR #7510, opened 2020-01-12
Fix song select showing incorrect key count

### LINKED ISSUE #7502: Key amount value for mania beatmaps in beatmap info is always zero
Screenshot_1
Screenshot_3

thewildtree:
Fixes #7502.

This previously had rounding in place, but since mania mapsets can't have decimal values there, and for maps from other modes Circle Size is shown as the title, it doesn't seem to be necessary.

### DIFF
--- a/osu.Game/Screens/Select/Details/AdvancedStats.cs
+++ b/osu.Game/Screens/Select/Details/AdvancedStats.cs
@@ -118,17 +118,9 @@ namespace osu.Game.Screens.Select.Details
                     mod.ApplyToDifficulty(adjustedDifficulty);
             }
 
-            //mania specific
-            if ((Beatmap?.Ruleset?.ID ?? 0) == 3)
-            {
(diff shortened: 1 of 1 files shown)


peppy:
I'm a bit curious why the rounding was going wrong. Did you investigate how this happened?

TheWildTree:
It wasn't going wrong per se, it's just that if `adjustedDifficulty` was null, the `adjustedValue` always became 0 by default:

`(int)MathF.Round(adjustedDifficulty?.CircleSize ?? 0))`

Proper way to have rounding here would be to first check if `adjustedDifficulty` has value and only then try to round the numbers.

peppy:
Right, I see it now thanks.

peppy (approved):

TheWildTree:
This always showed 0, not only when difficulty mods were applied.

PR #7522, opened 2020-01-13
Prevent SliderHead from moving along the slider

### LINKED ISSUE #7521: Slider input does not match osu!stable
There are some major differences that I've noticed when comparing lazer with stable:

- Slider head moves together with the sliderball, while on stable it stays in place with sliderball moving independently. Indirectly mentioned in #6088, almost certainly **causes replay inconsistencies** vs. stable, especially on sliders with high SV. This is best visible when sliderhead doesn't get hit at all (lazer left, stable right, full example [**here**](https://gfycat.com/ifr/DiligentMarvelousGuillemot)):

    image

    By the way, judgement sprites seem to be displayed 'behind' the sliders, stable has them on top (as seen above)

- ~~Slider circle range seems to be a bit different, same replay on lazer gives different results than stable. [**Video comparison here**](https://streamable.com/7jqj2).~~ Fixed.

Fixing these, especially the first one, should result in getting rid of major replay inconsistencies on maps with high SVs, [like this one](https://youtu.be/h3SkCJtSU3E?t=463) for example, since right now if a slider has very high SV and gets hit at least a bit late (even still in the 300 hitwindow), in lazer the head might move before getting hit, resulting in a miss.

thewildtree:
Partially fixes #7521.

Brings `SliderHead` behavior into line with stable (see [this video](https://www.youtube.com/watch?v=eqhPDrPV5Ac)).

No other changes seem to be necessary, though it won't hurt to consider modifying the slider snaking-out behavior - currently it starts no matter if `SliderHead` is hit, which in some cases (low OD + high SV for example) might look weird ([like this](https://user-images.githubusercontent.com/27722894/72269741-5de17280-3624-11ea-8ff6-2de7d16fe1f6.jpg)).

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderHead.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderHead.cs
@@ -32,17 +32,6 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             pathVersion.BindValueChanged(_ => updatePosition(), true);
         }
 
-        protected override void Update()
-        {
-            base.Update();
-
-            double completionProgress = Math.Clamp((Time.Current - slider.StartTime) / slider.Duration, 0, 1);
-
-            //todo: we probably want to reconsider this before adding scoring, but it looks and feels nice.
-            if (!IsHit)
-                Position = slider.CurvePositionAt(completionProgress);
-        }
-
         public Action<double> OnShake;
 
         protected override void Shake(double maximumLength) => OnShake?.Invoke(maximumLength);


peppy:
Please see the comment you removed here. It is intended to be the way it is.

TheWildTree:
It seemed to me that it was a leftover from before any scoring was in place, like the comment suggested.

This seems weird to be intentional, since it heavily changes the way you play (especially fast) sliders vs. stable, bringing a lot of replay inconsistencies as well as just different feel during gameplay. 

It can effectively shorten the sliderhead hitwindow by half, because if you don't hit it before it starts moving, you now have your hitcircle literally running away from you.

I'd really like to ask you to reconsider it.

peppy:
It does not change the hitwindow, and adds new gameplay mechanics as the target becomes moving if you are too late. We are undecided, but it shouldn't be changed for now. See the todo, which suggests it will be reconsidered at a later stage.

peppy:
Also your change is not correct for cases where the user has snaking turned on, where the slider head would have become detached from the slider.

Wieku:
He asked for propositions in the first comment how to handle that scenario, he probably planned to fix that in the next commit

TheWildTree:
Wieku is right, that's exactly why I mentioned it in the first comment. My idea would be to make the snaking-out start after hitting the sliderhead and smooth it out, so there won't be any detachment visible.

At least fix the todo because we already have scoring in place (not final, but it is there), which makes it misleading.

Also I'd still argue that this mechanic is not a good thing - **hitobjects should stay where the mapper placed them**.

peppy:
If you do that (smoothing out) the visuals will not match the game mechanics, which is a no-go.

Sliders are moving hitobjects. Lazer adds accuracy. Part of this addition is that if you hit late, the hitbox changes. This is intended for now.

Further discussion can occur at a later point, or in the issue thread.

TheWildTree:
They would match, because both the hitcircle and the sliderball would be where they're supposed to, only thing 'smoothed out' would be the snaking-out of the slider body, which after hitting the head, would quickly (say same speed as slider snaking-in) catch up to the sliderball.

It's your decision though and let's leave that for now.

PR #7523, opened 2020-01-13
Remove duplicate condition in TournamentFont

### LINKED ISSUE #7519: Duplicate check in TournamentFont#GetWeightString
**Describe the bug:** In the class osu.Game.Tournament.TournamentFont, the GetWeightString method performs the same check twice. See [this line](https://github.com/ppy/osu/blob/f3bb47ffce87fb1077e6627d002c9add03ecf583/osu.Game.Tournament/TournamentFont.cs#L64).
```cs
// Only exo has an explicit "regular" weight, other fonts do not
if (weight == FontWeight.Regular && family != GetFamilyString(TournamentTypeface.Aquatico) && family != GetFamilyString(TournamentTypeface.Aquatico))
    weightString = string.Empty;
```

**osu!lazer version:** master branch commit f3bb47ffce87fb1077e6627d002c9add03ecf583

recapitalverb:
Fixes #7519.

### DIFF
--- a/osu.Game.Tournament/TournamentFont.cs
+++ b/osu.Game.Tournament/TournamentFont.cs
@@ -61,7 +61,7 @@ namespace osu.Game.Tournament
             string weightString = weight.ToString();
 
             // Only exo has an explicit "regular" weight, other fonts do not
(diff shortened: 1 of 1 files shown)


peppy:
This is going to conflict with the owc branch, I believe, but I'll fix that there.

PR #7526, opened 2020-01-13
Remove duplicate returns in KeyCounterAction

recapitalverb:
This should not affect semantics at all - just makes what `OnPressed/Released` do a bit clearer.


recapitalverb:
Ouch, should have done a quick `dotnet format`.

smoogipoo:
We have discussed this in the past. The matter of inverting conditions is, in many cases, a subjective opinion.
I personally advocate for it, but I try not to force it unless there's a concrete readability argument.

Based on that, I don't believe this is substantial enough to warrant change.

Issue #7527, opened 2020-01-14
Can't navigate volume meters with left and right arrows


Joehuu:
Needed to fully control volume overlay with keyboard.

PR #7528, opened 2020-01-14
Allow changing volume in song select with arrow keys when pressing alt

Joehuu:
Volume overlay still needs navigation with left and right arrows for more usability, tracked in https://github.com/ppy/osu/issues/7527.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -412,6 +412,9 @@ namespace osu.Game.Screens.Select
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
+            if (e.AltPressed)
+                return base.OnKeyDown(e);
+
             int direction = 0;
             bool skipDifficulties = false;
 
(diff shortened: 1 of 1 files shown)


bdach:
Have you made sure this doesn't break non-English keyboard layouts/special char entry? I think this conditional might be too general.

Joehuu:
I'm not sure how to check that, but if someone notices a bug, we'll reconsider the change.

bdach:
Some layouts use <kbd>Alt</kbd> as a modifier to printable keys in order to insert special characters. I'll check whenever I get the chance, but if you want to, you could try one of those (US International should have some I believe).

If it turns out to be a problem, we could either go for a more targeted check or a platform action.

peppy:
Shouldn't affect arrow keys, I don't think.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -436,7 +436,7 @@ protected override bool OnKeyDown(KeyDownEvent e)
                     break;
             }
 
-            if (direction == 0)
+            if (direction == 0 || e.AltPressed)
```
peppy: It's unfortunate that this is how we have to do it, but I can't find a better way. Ultimately, we would want `OnkeyDown` and `OnPressed` to both be handled at the same time in a single tree iteration so we could correctly handle this at the receptor.

I think at very least, the early return should be moved to the top of this method, and commented with reasoning as to why it's done (so we can hopefully implement in a better way in the future).

peppy (changes requested):

peppy (approved):

PR #7532, opened 2020-01-14
Implement editor beatmap saving

### LINKED ISSUE #7531: Editor should save beatmaps in legacy (.osu) format

smoogipoo:
Resolves https://github.com/ppy/osu/issues/7531

Very initial/basic implementation. Doesn't support adding or removing files, but that would be done through similar APIs.


peppy:
Exiting editor switches beatmaps, but as discussed IRL we can fix that in a future PR.

peppy (approved):

PR #7537, opened 2020-01-15
Make iconless mods display acronyms

### LINKED ISSUE #6966: No icon for "Touch device" pseudo-mod
**Describe the bug:** 
If player beat their personal highscore in a beatmap via a touch device in osu!stable, the placeholder icon (?) with the caption "Touch Device" appears instead of that shown on osu!web.

**osu!lazer version:** commit [`053e38c`](https://github.com/ppy/osu/commit/053e38c95fc97e60c4d3ca0098da929e22955a60) compiled with debug

### LINKED ISSUE #7529: Display `Acronym` instead of question mark when mod does not have icon
Seems easy to implement, and might look better.

**Proposal designs of the feature:** See title. Use `OsuFont.Numeric` to display.

mcendu:
Closes #7529

Partial resolution to #6966 (icons for mania key mods; seems like font weight is not right).

### DIFF
--- a/osu.Game/Rulesets/Mods/ModDaycore.cs
+++ b/osu.Game/Rulesets/Mods/ModDaycore.cs
@@ -12,7 +12,7 @@ namespace osu.Game.Rulesets.Mods
     {
         public override string Name => "Daycore";
         public override string Acronym => "DC";
-        public override IconUsage Icon => FontAwesome.Solid.Question;
+        public override IconUsage? Icon => FontAwesome.Solid.Question;
         public override string Description => "Whoaaaaa...";
 
         private readonly BindableNumber<double> tempoAdjust = new BindableDouble(1);
--- a/osu.Game/Rulesets/Mods/ModDaycore.cs
+++ b/osu.Game/Rulesets/Mods/ModDaycore.cs
@@ -12,7 +12,7 @@ namespace osu.Game.Rulesets.Mods
     {
         public override string Name => "Daycore";
         public override string Acronym => "DC";
-        public override IconUsage? Icon => FontAwesome.Solid.Question;
+        public override IconUsage? Icon => null;
         public override string Description => "Whoaaaaa...";
 
         private readonly BindableNumber<double> tempoAdjust = new BindableDouble(1);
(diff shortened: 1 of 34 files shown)


osu.Game/Rulesets/Mods/ModDaycore.cs
```diff
@@ -12,7 +12,7 @@ public abstract class ModDaycore : ModHalfTime
     {
         public override string Name => "Daycore";
         public override string Acronym => "DC";
-        public override IconUsage Icon => FontAwesome.Solid.Question;
+        public override IconUsage? Icon => null;
```
iiSaLMaN: redundant

bdach:
I'm starting to wonder if `ModIcon` wouldn't fare better as a `ModelBackedDrawable`. Your current implementation looks quite close already, although not overly obviously (it's not a straightforward change, but more of a major refactor).

osu.Game/Rulesets/UI/ModIcon.cs
```diff
@@ -56,15 +62,44 @@ public ModIcon(Mod mod)
...
                     Shadow = true,
                 },
+                modAcronym = new OsuSpriteText
+                {
+                    Origin = Anchor.Centre,
+                    Anchor = Anchor.Centre,
+                    Colour = OsuColour.Gray(84),
+                    Alpha = 0,
+
+                    Font = OsuFont.Numeric.With(null, size - 57.5f),
```
smoogipoo:
```suggestion
                    Font = OsuFont.Numeric.With(null, 22f),
```
.5 isn't significant, also prefer to have things defined - would like the `size - 35` below to be simplified to `45` also.

osu.Game/Rulesets/UI/ModIcon.cs
```diff
@@ -56,15 +62,44 @@ public ModIcon(Mod mod)
                     Icon = OsuIcon.ModBg,
                     Shadow = true,
                 },
+                modAcronym = new OsuSpriteText
+                {
+                    Origin = Anchor.Centre,
+                    Anchor = Anchor.Centre,
+                    Colour = OsuColour.Gray(84),
+                    Alpha = 0,
+
```
smoogipoo:
```suggestion
```

smoogipoo (changes requested):
Just a few nits

smoogipoo:
Definitely could do with a refactor to MBD but it's not super necessary for now though, imo.

smoogipoo (approved):
seems ok for now

PR #7538, opened 2020-01-15
Decouple blueprint container to allow usage in timeline

peppy:
This is a first pass to see if this is the direction we want to go.


smoogipoo:
Applied some minor refactorings. Further refactorings can be done as discussed (separate container should not be a thing, the base `selectionContainer` should be replaced instead) better after https://github.com/ppy/osu-framework/pull/3181 is merged.

Otherwise, I'm happy to merge this as it is for now.

smoogipoo (dismissed):

smoogipoo:
Oh also testcase failure that I haven't resolved :p

PR #7540, opened 2020-01-15
Hyperlink beatmap author names

### LINKED ISSUE #7518: Mapper usernames should be hyperlinked consistently in song select, beatmap overlay
**Describe the new feature:** 

Mapper usernames should be hyperlinked consistently in all places including song select and beatmap overlay.

In direct, mapper is hyperlinked:
image

It is not in song select:
image

It is not in beatmap details overlay:
image

LastExceed:
Closes #7518 
currently the new hyperlink in song select is bugged (cause and effect explained in the issue above). I decided not fix that as part of this PR because I consider it a separate bug and we haven't decided yet how we want to deal with it.

For now I kept the fonts as they were, but shouldn't the author name be **bold** instead of _italic_ to match the other fields (and also all other occurrences of "mapped by xy" throughout the game) ?


mcendu (changes requested):

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -341,23 +342,23 @@ private string getBPMRange(IBeatmap beatmap)
...
+                    }).With(d =>
                     {
-                        Text = metadata.Author.Username,
-                        Font = OsuFont.GetFont(weight: FontWeight.Bold, size: 15),
-                    }
+                        d.AutoSizeAxes = Axes.Both;
+                        d.AddText("mapped by ");
+                        d.AddUserLink(metadata.Author, text => text.Font = text.Font.With(weight: FontWeight.Bold));
+                    })
                 };
```
mcendu: Should remove the `new[]` wrapper if my change is accepted.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -341,23 +342,23 @@ private string getBPMRange(IBeatmap beatmap)
                 return $"{bpmMin:0}-{bpmMax:0} (mostly {beatmap.ControlPointInfo.BPMMode:0})";
             }
 
-            private OsuSpriteText[] getMapper(BeatmapMetadata metadata)
+            private LinkFlowContainer[] getMapper(BeatmapMetadata metadata)
```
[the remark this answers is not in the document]
LastExceed: good idea, done.

LastExceed:
```csharp
AddAssert("check author", () => infoWedge.Info.MapperContainer.Children.OfType<OsuSpriteText>().Any(s => s.Text == $"{ruleset.ShortName}Author"));
```
the test fails because `LinkFlowContainer` contains a `DrawableLinkCompiler`. How should I deal with this?

bdach:
```csharp
            AddAssert("check author", () => infoWedge.Info.MapperContainer.ChildrenOfType<LinkFlowContainer>().Single()
                                                     .Children.OfType<OsuSpriteText>().Any(s => s.Text == $"{ruleset.ShortName}Author"));
```
seems to work, no? It is a bit of a mouthful, sure, and coupled to implementation details, but it was the case already, so it's not *worse*...

LastExceed:
ready for review

bdach:
I'm unsure if manually adding the links as done here is what we want, instead of creating links through `MessageFormatter` like [`DrawableRecentActivity`](https://github.com/ppy/osu/blob/5d7c40da6e8bdbf1c67e30b03a94142ae8d24480/osu.Game/Overlays/Profile/Sections/Recent/DrawableRecentActivity.cs#L51-L53) or [`DrawableComment`](https://github.com/ppy/osu/blob/a614e334059ea1ddb688652c9ad1239c8aa4b22f/osu.Game/Overlays/Comments/DrawableComment.cs#L216-L217) already do. It would certainly get rid of some code duplication in here.

LastExceed:
For this implementation I used existing occurrences of the "mapped by" link as reference: [`DirectGridPanel`](https://github.com/ppy/osu/blob/a28934f9f2b69c7317dbdaf5a36534e0410fe96c/osu.Game/Overlays/Direct/DirectGridPanel.cs#L123-L132) [`DirectListPanel`](https://github.com/ppy/osu/blob/a28934f9f2b69c7317dbdaf5a36534e0410fe96c/osu.Game/Overlays/Direct/DirectListPanel.cs#L170-L184) [`DrawableMostPlayedBeatmap`](https://github.com/ppy/osu/blob/a28934f9f2b69c7317dbdaf5a36534e0410fe96c/osu.Game/Overlays/Profile/Sections/Historical/DrawableMostPlayedBeatmap.cs#L91-L100)
I agree there is some code duplication going on here, but I don't see how using `MessageFormatter` instead of `AddUserLink()` would help with that, it would only get more verbose. We could use a factory function for "mapped by" `LinkFlowContainer`s, but I'm not even sure that's worth doing for such a small snippet

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
```diff
@@ -49,7 +50,16 @@ private void updateDisplay()
 
             fields.Children = new Drawable[]
             {
-                new Field("mapped by", BeatmapSet.Metadata.Author.Username, OsuFont.GetFont(weight: FontWeight.Regular, italics: true)),
+                new LinkFlowContainer(s =>
```
peppy: Please make this into its own class called `MapperLink` or similar, since it exists in two places with the same implementation. the constructor can take the font size and the metadata object.

peppy (changes requested):

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
LastExceed: Sure, can do. It's actually not just 2, but 5 places now: 2 new that i implemented and 3 existing that i took for reference (linked above). I'll use just the author object since that seems a lot more intuitive

LastExceed:
The tests are failing and I don't know why. I reverted to the last commit where CI reported green light but they're still failing when I run them locally. I then re-cloned the whole thing to make sure I didn't just mess up my git: 
```
git clone https://github.com/LastExceed/osu --recurse-submodules
cd osu
git checkout clickable-users
git reset 36f6736 --hard
```
and ran the `osu! (Tests)` configuration in Rider, yet the tests are still failing. 

I'm gonna need some help here.

peppy:
This may be an unrelated failure, will re-run to test.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -261,7 +263,8 @@ private void load(LocalisationManager localisation)
                                 Margin = new MarginPadding { Top = 10 },
                                 Direction = FillDirection.Horizontal,
                                 AutoSizeAxes = Axes.Both,
-                                Children = getMapper(metadata)
+                                //Child = getMapper(metadata)
```
bdach: Remove comment.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -261,7 +263,8 @@ private void load(LocalisationManager localisation)
...
+                                Child = string.IsNullOrEmpty(metadata.Author?.Username) ? new LinkFlowContainer() : new MapperLink(metadata.Author, 15)
```
bdach: * Can use `Empty()` instead of a new link flow. Will need updating test assertions.
* This usage actually has a big problem in that the link doesn't change colour when hovered. This is due to

  https://github.com/ppy/osu/blob/9be3d7a16171ca4dbe15b0ad3689c4f18f35ae2c/osu.Game/Screens/Select/BeatmapInfoWedge.cs#L156

  which means the buffered container won't be redrawn every frame.

  I wouldn't recommend turning it off as it'd probably be a significant performance regression. Instead we might want to consider splitting out the text layer to be above the pre-rendered buffered background of the wedge.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapInfoWedge.cs
```diff
@@ -105,7 +107,8 @@ private void testBeatmapLabels(Ruleset ruleset)
             AddAssert("check version", () => infoWedge.Info.VersionLabel.Text == $"{ruleset.ShortName}Version");
             AddAssert("check title", () => infoWedge.Info.TitleLabel.Text == $"{ruleset.ShortName}Source — {ruleset.ShortName}Title");
             AddAssert("check artist", () => infoWedge.Info.ArtistLabel.Text == $"{ruleset.ShortName}Artist");
-            AddAssert("check author", () => infoWedge.Info.MapperContainer.Children.OfType<OsuSpriteText>().Any(s => s.Text == $"{ruleset.ShortName}Author"));
+            AddAssert("check author", () =>
+                infoWedge.Info.MapperContainer.ChildrenOfType<LinkFlowContainer>().Single().ChildrenOfType<OsuSpriteText>().Any(s => s.Text == $"{ruleset.ShortName}Author"));
```
bdach: Suppose this one's my brainfart, but can simplify to
```suggestion
                infoWedge.Info.MapperContainer.ChildrenOfType<OsuSpriteText>().Any(s => s.Text == $"{ruleset.ShortName}Author"));
```

bdach (changes requested):
One big problem and other nitpicks. FWIW, I wanted to fix up the nitpicks for you, but my push attempt got rejected, so you might have the *Allow edits from maintainers* checkbox unchecked.

~~Also needs merge conflict resolution.~~

LastExceed:
> you might have the Allow edits from maintainers checkbox unchecked

oops my bad, try pushing now @bdach

bdach:
Yup, works now. Doesn't change the fact that the buffered container issue affecting link hover needs to be resolved.

LastExceed:
the merge looks fine to me

bdach:
To expound on the buffered background thing a little more:

`BeatmapInfoWedge` internally uses `BufferedWedgeInfo`, which is a `BufferedContainer`. As described in the xmldocs, a buffered container is first drawn to a texture, and only after that drawn to the screen. Sometimes, when it is known that the contents of the container aren't really changing that much, it can be worth caching the rendered texture entirely, and that's what `CacheDrawnFrameBuffer = true` does. It is especially helpful in scenarios where there are a lot of effects applied, as is the case in the wedge.

This means that for every beatmap change, the wedge contents are rendered to a texture once. In turn, this breaks any and all hover color transforms, as the `MapperLink` you've added is *inside* the `BufferedWedgeInfo` component, and therefore won't be updated after the initial render to texture.

The simplest way that comes to my mind when it comes to resolving this, is to split `BufferedWedgeInfo` into two "layers", as follows:

```
BufferedWedgeBackground   // contains the background of the wedge itself,
                          // with all the glow effects, etc.
WedgeInfoText             // container with all of the labels in front of the background
    TitleLabel
    ArtistLabel
    MapperContainer
    InfoLabelContainer
```

This way, the expensive background rendering can still be done once per beatmap, and the extracted `WedgeInfoText` foreground layer will allow for `MapperContainer`'s link to change colour on hover. Mind you, this *will* probably result in at least *a little* drop in performance, however if I were to guess, re-rendering the entire wedge every frame by just setting `CacheDrawnFrameBuffer = false` would probably be even worse. This is just me theorycrafting though, to be certain I would have to benchmark.

I would probably submit such a change as a separate PR myself, and get that one merged first, before this.

EVAST9919:
Superseded by https://github.com/ppy/osu/pull/12488

PR #7541, opened 2020-01-15
Implement colour schemes for OverlayHeader

EVAST9919:
In the future will be applied to the whole overlay

Before:
before

After:
after

Web:
web

### DIFF
--- a/osu.Game/Graphics/OsuColour.cs
+++ b/osu.Game/Graphics/OsuColour.cs
@@ -66,6 +66,7 @@ namespace osu.Game.Graphics
 
         public Color4 ForOverlayElement(OverlayColourScheme colourScheme, float saturation, float lightness, float opacity = 1) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, opacity));
 
+        // See https://github.com/ppy/osu-web/blob/4218c288292d7c810b619075471eaea8bbb8f9d8/app/helpers.php#L1463
         private static float getBaseHue(OverlayColourScheme colourScheme)
         {
             float hue;
--- a/osu.Game/Graphics/OsuColour.cs
+++ b/osu.Game/Graphics/OsuColour.cs
@@ -3,6 +3,7 @@
 
 using System;
 using osu.Game.Beatmaps;
+using osuTK;
 using osuTK.Graphics;
 
 namespace osu.Game.Graphics
@@ -61,6 +62,24 @@ namespace osu.Game.Graphics
                 case DifficultyRating.ExpertPlus:
                     return useLighterColour ? Gray9 : Gray0;
             }
+        }
+
+        public Color4 ForOverlayElement(OverlayColourScheme colourScheme, float saturation, float lightness, float opacity = 1) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, opacity));
+
+        private static float getBaseHue(OverlayColourScheme colourScheme)
+        {
+            var hue = colourScheme switch
+            {
+                OverlayColourScheme.Red => 0,
+                OverlayColourScheme.Pink => 333,
+                OverlayColourScheme.Orange => 46,
+                OverlayColourScheme.Green => 115,
+                OverlayColourScheme.Purple => 255,
+                OverlayColourScheme.Blue => 200,
+                _ => 0,
+            };
+
+            return hue / 360f;
         }
 
         // See https://github.com/ppy/osu-web/blob/master/resources/assets/less/colors.less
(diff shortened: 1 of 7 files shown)


osu.Game/Graphics/OsuColour.cs
```diff
@@ -63,6 +64,24 @@ public Color4 ForDifficultyRating(DifficultyRating difficulty, bool useLighterCo
...
+            var hue = colourScheme switch
+            {
+                OverlayColourScheme.Red => 0,
+                OverlayColourScheme.Pink => 333,
+                OverlayColourScheme.Orange => 46,
+                OverlayColourScheme.Green => 115,
+                OverlayColourScheme.Purple => 255,
+                OverlayColourScheme.Blue => 200,
+                _ => 0,
+            };
```
bdach: Switch expressions aren't consensus yet.

osu.Game/Graphics/OsuColour.cs
```diff
@@ -63,6 +64,24 @@ public Color4 ForDifficultyRating(DifficultyRating difficulty, bool useLighterCo
...
+        {
+            var hue = colourScheme switch
+            {
+                OverlayColourScheme.Red => 0,
+                OverlayColourScheme.Pink => 333,
+                OverlayColourScheme.Orange => 46,
+                OverlayColourScheme.Green => 115,
+                OverlayColourScheme.Purple => 255,
+                OverlayColourScheme.Blue => 200,
+                _ => 0,
```
bdach: Out-of-range values should throw instead of quietly falling back to `Red`.

EVAST9919: Is it even possible to set out-of-range value here?

iiSaLMaN: yes, by bugs.

bdach: Enums are just really integer-type aliases, so you can forcibly cast any integer value and it'll get silently accepted but not get a nice printable name. Throwing also makes it easy to spot errors if somebody forgets to modify this switch, but adds a new enum value and uses it elsewhere.

EVAST9919: Oh, I see

osu.Game/Graphics/OsuColour.cs
```diff
@@ -63,6 +64,45 @@ public Color4 ForDifficultyRating(DifficultyRating difficulty, bool useLighterCo
             }
         }
 
+        public Color4 ForOverlayElement(OverlayColourScheme colourScheme, float saturation, float lightness, float opacity = 1) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, opacity));
+
+        private static float getBaseHue(OverlayColourScheme colourScheme)
```
peppy: add a comment referencing https://github.com/ppy/osu-web/blob/4218c288292d7c810b619075471eaea8bbb8f9d8/app/helpers.php#L1463 as the source for these

peppy (changes requested):

osu.Game/Graphics/OsuColour.cs
```diff
@@ -63,6 +64,46 @@ public Color4 ForDifficultyRating(DifficultyRating difficulty, bool useLighterCo
...
+
+        // See https://github.com/ppy/osu-web/blob/4218c288292d7c810b619075471eaea8bbb8f9d8/app/helpers.php#L1463
+        private static float getBaseHue(OverlayColourScheme colourScheme)
+        {
+            float hue;
+
+            switch (colourScheme)
+            {
+                default:
+                    throw new ArgumentException(@"Used colourScheme has no hue value!");
```
iiSaLMaN: The exception message doesn't seem to be descriptive enough:
```suggestion
                    throw new ArgumentException($@"{colourScheme} colour scheme does not provide a hue value in {nameof(getBaseHue)}.");
```

frenzibyte (changes requested):
Otherwise, this seems to be ready-for-merge, good job bringing colour scheming to the overlays, that cleaned up code a bit. :+1:

osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -16,24 +18,19 @@ public abstract class OverlayHeader : Container
...
-        }
+        private readonly ScreenTitle title;
 
         protected float BackgroundHeight
         {
             set => background.Height = value;
         }
 
-        protected OverlayHeader()
+        protected OverlayColourScheme ColourScheme { get; private set; }
```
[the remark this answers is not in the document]
EVAST9919: Why? We don't want to allow to change this elsewhere. However we can use a readonly field, but not sure if we want a field here.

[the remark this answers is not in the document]
iiSaLMaN: well, fyi, making a property get-only means it can only be set in-line statically `private Type A { get; } = 1;` or at constructor. (which is what you're doing), not allowing to be set publicly.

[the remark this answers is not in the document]
EVAST9919: Ah that's what you mean, sure then

peppy (approved):
looks good for now

PR #7546, opened 2020-01-16
Add a container type to easily hide online content when not online

Game4all:
# Summary

- This PR implements the `OnlineViewContainer`, a container for displaying online-only views where the user login is required.
- It supports hiding its content and showing a login placeholder (implemented in #7438) if the local user is not logged in
- It also displays a loading animation when the user is connecting to game servers.
- This component is planned to be reused to refine UX in online-only overlays (Social, Direct, Chat) instead of displaying an empty overlay


osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -151,7 +151,7 @@ protected PlaceholderState PlaceholderState
                         break;
 
                     case PlaceholderState.NotLoggedIn:
-                        replacePlaceholder(new LoginPlaceholder());
+                        replacePlaceholder(new LoginPlaceholder(@"Please sign in to view online leaderboards!"));
```
iiSaLMaN: This case should probably be removed and inherit `OnlineViewContainer`.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,79 @@
...
+                placeholderContainer = new Container
+                {
+                    RelativeSizeAxes = Axes.Both,
+                    Alpha = 0,
+                    Child = placeholder = new LoginPlaceholder(placeholder_message)
+                },
+            };
+        }
+
+        public void APIStateChanged(IAPIProvider api, APIState state)
```
iiSaLMaN: Should be `virtual` to allow components that inherit this and also want `IOnlineComponent` features to override this.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,79 @@
...
+            }
+            else
+            {
+                placeholderContainer.FadeOut(transform_time / 2, Easing.OutQuint);
+                content.FadeIn(transform_time, Easing.OutQuint);
+            }
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(IAPIProvider api)
```
iiSaLMaN: I think this should be a protected property
```csharp
[Resolved]
protected IAPIProvider API { get; set; }
```

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private OnlineViewContainer onlineView;
+        private Box box;
+
+        public TestSceneOnlineViewContainer()
+        {
+            Child = new Container
+            {
+                RelativeSizeAxes = Axes.Both,
+                Child = onlineView = new OnlineViewContainer(@"Please login to view dummy test content")
```
mcendu: Why not “sign in”?

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,79 @@
...
+using osu.Game.Online.API;
+using osu.Game.Online.Placeholders;
+
+namespace osu.Game.Online
+{
+    /// <summary>
+    /// A <see cref="Container"/> for dislaying online content who require a local user to be logged in.
+    /// Shows its children only when the local user is logged in and supports displaying a placeholder if not.
+    /// </summary>
+    public class OnlineViewContainer : Container, IOnlineComponent
```
iiSaLMaN: Should obviously be `abstract`.

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
iiSaLMaN: One thing I was aware from this is the placeholder message becoming inconsistent across the overlays by simply one of them having `log in` instead of `sign in`, I think the string should be concatenated somewhere inside the `LoginPlaceholder` in a way that the developer only need to pass reason to log in: 
```
OnlineViewContainer(@"dummy test content") -> $@"Please sign in to view {actionMessage}" -> "Please sign in to view dummy test content"
```

Game4all: This was in the initial implementation of the login placeholder but has been reverted to passing the whole string because string concatenation was deemed bad for localization.

iiSaLMaN: I mean, it will be concatenated before locallizing, at the constructor of `LoginPlaceholder`.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,79 @@
...
+            {
+                placeholderContainer.FadeOut(transform_time / 2, Easing.OutQuint);
+                content.FadeIn(transform_time, Easing.OutQuint);
+            }
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(IAPIProvider api)
+        {
+            api.Register(this);
```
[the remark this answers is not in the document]
Game4all: I don't see in what this is a memory leak, this is being done in several places without leaking any memory (see `FullscreenOverlay`)

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
Game4all: Sounds fine in that case.

osu.Game/Online/OnlineViewContainer.cs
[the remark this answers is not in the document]
iiSaLMaN: Well, It doesn't leak there because it's an overlay, which I believe never removed and re-created, but this is a container, although as a rule of thumb, you should unregister the component from any place it's registered at disposal.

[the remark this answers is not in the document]
Game4all: You're completely right , my bad.

osu.Game/Online/Leaderboards/Leaderboard.cs
Game4all: Should be kept for a follow-up PR covering usages  imo.

osu.Game/Online/OnlineViewContainer.cs
Game4all: Not sure about that one, there are place where we may want to use the container as is, without need to subclass

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+    /// </summary>
+    public class OnlineViewContainer : Container, IOnlineComponent
+    {
+        private readonly Container content;
+        private readonly Container placeholderContainer;
+        private readonly Placeholder placeholder;
+
+        private const int transform_time = 300;
+
+        protected override Container<Drawable> Content => content;
```
bdach: This and the underlying field is a bit confusing. An improperly-written derived class could override this get-only protected property to return something different, but have to call the base ctor, in which case `content` is initialised in the private field in the base, but not accessible to the derived, therefore creating one dangling container reference per instance. You probably want a protected auto property here.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+                },
+            };
+        }
+
+        public virtual void APIStateChanged(IAPIProvider api, APIState state)
+        {
+            switch (state)
+            {
+                case APIState.Offline:
+                case APIState.Connecting:
```
bdach: Are we sure we want the offline placeholder visible during connecting, instead of a loading animation of sorts?

osu.Game/Online/Leaderboards/Leaderboard.cs
bdach: In this case, why is this change even here? It seems orthogonal and unrelated.

osu.Game/Online/Placeholders/LoginPlaceholder.cs
```diff
@@ -14,15 +14,15 @@ public sealed class LoginPlaceholder : Placeholder
...
+        public LoginPlaceholder(string actionMessage)
         {
             AddIcon(FontAwesome.Solid.UserLock, cp =>
             {
                 cp.Font = cp.Font.With(size: TEXT_SIZE);
                 cp.Padding = new MarginPadding { Right = 10 };
             });
 
-            AddText(@"Please sign in to view online leaderboards!");
+            AddText(actionMessage);
```
iiSaLMaN: I meant the concatenation would reside here:
`AddText($"Please sign in to {actionMessage}!");`

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,115 @@
...
+            base.LoadComplete();
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            API?.Unregister(this);
+            base.Dispose(isDisposing);
+        }
+
+        protected enum PlaceholderStatus
```
iiSaLMaN: Also as a rule in `enum` names, they should not be plural (`Status` is plural), you probably want `State`.

osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -166,6 +161,7 @@ protected PlaceholderState PlaceholderState
         }
 
         protected Leaderboard()
+            : base("view online leaderboards!")
         {
             InternalChildren = new Drawable[]
```
iiSaLMaN: Have you tested this correctly, it seems that all internal children of the `OnlineViewContainer` got ripped out.

osu.Game/Online/Leaderboards/PlaceholderState.cs
```diff
@@ -11,7 +11,6 @@ public enum PlaceholderState
         Unavailable,
         NoneSelected,
         NoScores,
-        NotLoggedIn,
```
iiSaLMaN: `PlaceholderState` used for `Leaderboard<>` should probably be renamed to `LeaderboardPlaceholderState` for now.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,115 @@
...
+        {
+            Offline,
+            Online,
+            Connecting,
```
iiSaLMaN: The namings in this enum seem out-of-scope, what you probably want here is:
 - `Visible`
 - `Hidden`
 - `Loading`

with xmldocs shortly explaining what these status mean to the container.

iiSaLMaN: Reason for this is to expand the container into supporting more than just 1 (not logged in) placeholder.

frenzibyte (changes requested):
Seems like `Leaderboard<>` needs much work to inherit `OnlineViewContainer` correctly and use it nicely, I advocate you revert the inheritance and remove any change to it and leave it as a follow-up PR.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,115 @@
...
+                    Schedule(() => UpdatePlaceholderVisibility(PlaceholderStatus.Offline));
+                    break;
+
+                case APIState.Online:
+                    Schedule(() => UpdatePlaceholderVisibility(PlaceholderStatus.Online));
+                    break;
+            }
+        }
+
+        protected void UpdatePlaceholderVisibility(PlaceholderStatus status)
```
iiSaLMaN: Why not use `APIState` directly?

osu.Game/Online/Placeholders/LoginPlaceholder.cs
peppy: This will not localise well. Fine as-is.

osu.Game/Online/OnlineViewContainer.cs
peppy: why not just use apistate

osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -166,6 +161,7 @@ protected PlaceholderState PlaceholderState
         }
 
         protected Leaderboard()
+            : base("view online leaderboards!")
```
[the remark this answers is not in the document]
peppy: This shouldn't be done in the first place as it can't be localised. The full message should be constructed for each usage.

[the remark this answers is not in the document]
iiSaLMaN: Correct me if i'm wrong, but shouldn't the localising happen inside `LocalisedString`? 
Which means `SpriteText.Text` is the property that should receive the full message. (which should happen in `LoginPlaceholder`)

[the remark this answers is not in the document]
peppy: It's out of scope of this PR, but have a think how you would generate such localisation strings and how (with the logic you suggest) that can't be done statically.

Game4all:
Will revert any changes made to leaderboard class and apply them in a separate PR as it requires much more work into it as stated by @iiSaLMaN.

osu.Game/Online/Leaderboards/Leaderboard.cs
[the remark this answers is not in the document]
Game4all: So, also reverting string concatenation in constructor

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
```diff
@@ -0,0 +1,70 @@
...
+                            Anchor = Anchor.Centre,
+                            Origin = Anchor.Centre,
+                        }
+                    }
+                }
+            };
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
```
iiSaLMaN: why not a test method
```csharp
[Test]
public void TestVisibilityBehaviour()
{
    ...
}
```

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
```diff
@@ -0,0 +1,70 @@
...
+        {
+            AddStep("set status to offline", () => ((DummyAPIAccess)API).State = APIState.Offline);
+
+            AddAssert("children are hidden", () => !onlineView.Children.First().Parent.IsPresent);
```
iiSaLMaN: I advocate you expose `View` instead of playing with the sub-tree to retrieve it.

also, checking is present is not the best way to determine if a drawable is hidden, it may have a `override IsPresent => true` or custom values because it's a manager or something, check `Alpha` directly instead.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,101 @@
...
+        /// <summary>
+        /// Updates the visibility of the placeholder for this <see cref="OnlineViewContainer"/> accordingly to the passed <see cref="APIState"/>.
+        /// <list type="bullet">
+        /// <item> Passing <see cref="APIState.Online"/> will cause the online-only view to display.</item>
+        /// <item> Passing <see cref="APIState.Offline"/> will cause the placeholder to display.</item>
+        /// <item> Passing <see cref="APIState.Connecting"/> or <see cref="APIState.Failing"/> will cause the loading animation to display.</item>
+        /// </list>
+        /// </summary>
+        /// <param name="status">The new API state</param>
+        protected void UpdatePlaceholderVisibility(APIState status)
```
iiSaLMaN: Is there a reason this is protected? I'm not sure if it's fine for a deriver to be able to manipulate the placeholder visibility.

frenzibyte (changes requested):
Seems OK mostly except for a thing, the rest is all on the test. (the place I missed in the last review)

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
Game4all: Actually, since View is a normal container, checking if drawable is visible through IsPresent should be safe

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -18,7 +18,7 @@ public class OnlineViewContainer : Container, IOnlineComponent
     {
         private readonly Container placeholderContainer;
         private readonly Placeholder placeholder;
-        private readonly LoadingAnimation loading;
+        internal readonly LoadingAnimation LoadingAnimation;
```
iiSaLMaN: Doesn't need to be that exposed, what you're looking for is making this protected and expose it on test sub-class with `public new LoadingAnimation => base.LoadingAnimation`.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
iiSaLMaN: Well, using `IsPresent` is technically not safe because you're not checking alpha, but you're checking a property that has alpha in it's calculation, I find it more arbitrary.

Game4all: Searching for IsPresent in Tests yields results where that property is being used to check if a drawable is displayed on screen. So it should be fine to use it to check for drawable visibility
Capture

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// </summary>
+    public class OnlineViewContainer : Container, IOnlineComponent
+    {
+        private readonly Container placeholderContainer;
+        private readonly Placeholder placeholder;
+        protected readonly LoadingAnimation LoadingAnimation;
+
+        private const int transform_time = 300;
+
+        protected Container View { get; }
```
peppy: combine with content.

`protected override Container<Drawable> Content { get; }`

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+                    Child = placeholder = new LoginPlaceholder(placeholderMessage)
+                },
+                LoadingAnimation = new LoadingAnimation
+                {
+                    Alpha = 0,
+                }
+            };
+        }
+
+        public virtual void APIStateChanged(IAPIProvider api, APIState state) => updatePlaceholderVisibility(state);
```
peppy: combine with private method below. it's only called from here.

osu.Game/Online/Leaderboards/Leaderboard.cs
peppy: seems unrelated to this implementation.

peppy (changes requested):

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+using osu.Game.Online.API;
+using osu.Game.Online.Placeholders;
+
+namespace osu.Game.Online
+{
+    /// <summary>
+    /// A <see cref="Container"/> for dislaying online content who require a local user to be logged in.
+    /// Shows its children only when the local user is logged in and supports displaying a placeholder if not.
+    /// </summary>
+    public class OnlineViewContainer : Container, IOnlineComponent
```
peppy: should be abstract.

peppy (changes requested):

osu.Game/Online/Leaderboards/Leaderboard.cs
iiSaLMaN: it is actually, the `LoginPlaceholder` constructor now has a message parameter.

osu.Game/Online/OnlineViewContainer.cs
Game4all: Reverting suggested behaviour as it causes the container to not exhibit desired behaviour (Content doesn't hide) when `Content` property is overriden in sub-class

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        protected OnlineViewContainer(string placeholderMessage)
+        {
+            InternalChildren = new Drawable[]
+            {
+                content = new Container
+                {
+                    RelativeSizeAxes = Axes.Both,
+                },
+                placeholderContainer = new Container
```
iiSaLMaN: any reason why this placeholder is added in a container? (tested removing it and animating everything directly in `placeholder`, no issues.)

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,90 @@
...
+                }
+            };
+        }
+
+        public virtual void APIStateChanged(IAPIProvider api, APIState state)
+        {
+            switch (state)
+            {
+                case APIState.Offline:
+                    content.FadeOut(150, Easing.OutQuint);
```
iiSaLMaN: This is limiting this container into just fading out on placeholder appearance, would be much better to make this customizable (by inheriting `VisibilityContainer` and setting visibility here) and add default behaviour here (fading in & out) with custom usages included in tests. (use grid container to show them both)

frenzibyte (changes requested):
Tested, seems pretty good otherwise.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+            };
+        }
+
+        public virtual void APIStateChanged(IAPIProvider api, APIState state)
+        {
+            switch (state)
+            {
+                case APIState.Offline:
+                    PopOut();
```
iiSaLMaN: That's a very dangerous way to do it (visibility state is not updated), you should be calling `Hide()` and `Show()` instead.

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
```diff
@@ -0,0 +1,95 @@
...
+                    {
+                        Text = "dummy online content",
+                        RelativeSizeAxes = Axes.Both,
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                    }
+                };
+            }
+
+            protected override void PopOut() => Content.FadeOut(150, Easing.OutQuint);
```
iiSaLMaN: another way to show that it's customizable is by creating a `BufferedContainer` and blur it here.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
Game4all: Simply indicating in class XMLDoc that PopIn() and PopOut() can be overridden to customize fade behaviour should be sufficient imo

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,89 @@
...
+using osu.Game.Graphics.UserInterface;
+using osu.Game.Online.API;
+using osu.Game.Online.Placeholders;
+
+namespace osu.Game.Online
+{
+    /// <summary>
+    /// A <see cref="Container"/> for dislaying online content who require a local user to be logged in.
+    /// Shows its children only when the local user is logged in and supports displaying a placeholder if not.
+    /// <see cref="PopIn"/> and <see cref="PopOut"/> methods can be overriden to customize fade in/out behaviour
```
iiSaLMaN:
```suggestion
    /// <see cref="PopIn"/> and <see cref="PopOut"/> methods can be overriden to customize visibility animation.
```

peppy: why is this necessary in the first place? that should be documented on `VisibilityContainer`, not this one.

osu.Game/Online/OnlineViewContainer.cs
peppy: i don't understand.

Game4all: Actually, nevermind it can be merged back. 
Care just needs to be taken to call `base.Content` when customizing the visibility animations of a drawable inheriting `OnlineViewContainer` and also overriding `Content` property. Else it would
cause the online view to still be displayed even if it should be hidden (as seen with a refactor of `Leaderboard<>` class inheriting `OnlineViewContainer`)

image

peppy: The component should be made in a way that this is not possible. Also, the change proposed here doesn't affect this – it could still happen with your current code.

Game4all:
I did a bit of cleanup in this PR commit history by removing commits of my past attempts at making Leaderboard derive from OnlineViewContainer. Should be ready for a re-review.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+                    FadeContentOut(TransformationTarget);
+                    LoadingAnimation.Show();
+                    placeholder.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
+                    break;
+            }
+        }
+
+        protected abstract void FadeContentOut(Drawable content);
+
+        protected abstract void FadeContentIn(Drawable content);
```
iiSaLMaN: The naming seems to be unnecessarily specific (content may not be faded, but blurred or darkened), how about `PopContentIn` and `PopContentOut`, or `AnimateHide` and `AnimateShow`?

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// Shows its children only when the local user is logged in and supports displaying a placeholder if not.
+    /// </summary>
+    public abstract class OnlineViewContainer : Container, IOnlineComponent
+    {
+        private readonly Placeholder placeholder;
+        protected readonly LoadingAnimation LoadingAnimation;
+
+        protected const double TRANSFORM_TIME = 300.0;
+
+        internal readonly Container TransformationTarget;
```
iiSaLMaN: why `internal` (if for testing purposes, never expose a property in normal classes for testing purposes).

iiSaLMaN: would be better to rename this to `ViewTarget`.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestSceneOnlineViewContainer.cs
```diff
@@ -0,0 +1,93 @@
...
+            AddStep("set status to connecting", () => ((DummyAPIAccess)API).State = APIState.Connecting);
+
+            AddAssert("children are hidden", () => !onlineView.TransformationTarget.IsPresent);
+            AddAssert("loading animation is visible", () => onlineView.LoadingAnimation.IsPresent);
+
+            AddStep("set status to failing", () => ((DummyAPIAccess)API).State = APIState.Failing);
+
+            AddAssert("children are hidden", () => !onlineView.TransformationTarget.IsPresent);
+            AddAssert("loading animation is visible", () => onlineView.LoadingAnimation.IsPresent);
+        }
```
bdach: I'd split this testcase in two (`TestConnectingStateVisibility` and `TestFailingStateVisibility`). You could even make all four test cases into a parameterised test for brevity but I guess that's just extra points.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,92 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Graphics.UserInterface;
+using osu.Game.Online.API;
+using osu.Game.Online.Placeholders;
+
+namespace osu.Game.Online
+{
+    /// <summary>
+    /// A <see cref="Container"/> for dislaying online content which require a local user to be logged in.
```
bdach:
```suggestion
    /// A <see cref="Container"/> for displaying online content which require a local user to be logged in.
```

osu.Game/Online/OnlineViewContainer.cs
bdach: `protected` with exposition in `TestSceneOnlineViewContainer` could work, but then it's kind of not great how you can do things with `TransformationTarget` on your own and *also* define `FadeContent{In,Out}`. Maybe should be `private` instead, and the tests should check the `Box` itself with `ChildrenOfType<>`. Not 100% sure either way.

iiSaLMaN: right, `TransformationTarget` should not be visible not even to the sub-class, checking *content* of the `TestOnlineViewContainer` is probably the best, I advocate for that.

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -21,7 +21,7 @@ public abstract class OnlineViewContainer : Container, IOnlineComponent
 
         protected const double TRANSFORM_TIME = 300.0;
 
-        internal readonly Container TransformationTarget;
+        private readonly Container viewTarget;
         protected override Container<Drawable> Content { get; }
```
iiSaLMaN: why not extract the backing-field of this property and using it as the "view target", adding a second layer seems redundant.
```csharp
```

frenzibyte (changes requested):

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+                    PopContentOut(viewContent);
+                    LoadingAnimation.Show();
+                    placeholder.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
+                    break;
+            }
+        }
+
+        protected abstract void PopContentOut(Drawable content);
+
+        protected abstract void PopContentIn(Drawable content);
```
iiSaLMaN: Why did this become abstract? Most of the subclasses may use basic fading animation, that's what `virtual` is mostly useful for.

frenzibyte (changes requested):

osu.Game/Online/OnlineViewContainer.cs
bdach: it always was abstract as far as I can see, FWIW

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+                    PopContentOut(viewContent);
+                    LoadingAnimation.Show();
+                    placeholder.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
+                    break;
+            }
+        }
+
+        protected virtual void PopContentOut(Drawable content) => content.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
+
+        protected virtual void PopContentIn(Drawable content) => content.FadeIn(TRANSFORM_TIME, Easing.OutQuint);
```
bdach: Strange how this is the only place that uses `TRANSFORM_TIME` instead of this halved. Did you really mean this, or did you want

```suggestion
        protected virtual void PopContentIn(Drawable content) => content.FadeIn(TRANSFORM_TIME / 2, Easing.OutQuint);
```

here too?

bdach (commented):
Save from one arbitrary time value I think this is ready to go. Let's get this in and move on.

osu.Game/Online/OnlineViewContainer.cs
iiSaLMaN: it was overridden when it inherited `VisibilityContainer`.

osu.Game/Online/OnlineViewContainer.cs
peppy: i believe this was intentional. generally feels good to fade out at double the speed of the related fade in

osu.Game/Online/OnlineViewContainer.cs
```diff
@@ -0,0 +1,88 @@
...
+                case APIState.Failing:
+                case APIState.Connecting:
+                    PopContentOut(viewContent);
+                    LoadingAnimation.Show();
+                    placeholder.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
+                    break;
+            }
+        }
+
+        protected virtual void PopContentOut(Drawable content) => content.FadeOut(TRANSFORM_TIME / 2, Easing.OutQuint);
```
peppy: not sure these need to exist until they have a use case, but i guess no harm..

peppy:
@Game4all i've made a few code quality and a few more important fixes (like potentially failing test steps). please check my commits for future contributions!

peppy (approved):

PR #5442, opened 2019-07-23
Update profile scores in line with the web

EVAST9919:
Screenshot_1

PR #7554, opened 2020-01-18
Update profile scores in line with the web design

EVAST9919:
Supersedes https://github.com/ppy/osu/pull/5442

Screenshot_1


osu.Game/Overlays/Profile/Sections/ProfileItemBackground.cs
```diff
@@ -0,0 +1,49 @@
...
+        protected override IEnumerable<Drawable> EffectTargets => new[] { background };
+        protected override Container<Drawable> Content => content;
+
+        private readonly Box background;
+        private readonly Container content;
+
+        public ProfileItemBackground()
+        {
+            RelativeSizeAxes = Axes.Both;
+            Enabled.Value = true; //manually enabled, because we have no action
```
bdach: Not sure how I feel about this from a UX standpoint. Yes, I understand it's to bring it with parity in respect to web, but it kind of breaks user expectations (it has a hover state and hover *and click* sounds, but clicking it doesn't actually do anything).

bdach (commented):
I suppose this is mostly ok?

osu.Game/Overlays/Profile/Sections/ProfileItemBackground.cs
EVAST9919: Valid point. Removing sounds might work, but I'll wait for ppy's response on this.

EVAST9919: Actually, yes. In the current implementation there are no sounds, so I'll change that

recapitalverb:
The pp display shown in the screenshot seems to be pushed too much to the right.

smoogipoo (approved):
Lgtm, just some minor styling fixes.

PR #7562, opened 2020-01-19
Fix order and naming of Difficulty Adjust mod settings

### LINKED ISSUE #7387: Order/Naming of Difficulty Adjust settings are different from beatmap details
**Describe the bug:** 
The order of the settings and the names of the settings are different between the beatmap details listed in song select and the settings panel for Difficulty Adjust. This can cause some confusion when looking between the settings and the beatmap details.

Details in song select:
image
Difficulty Adjust settings:
image

**osu!lazer version:** 
2019.1227.0

**Logs:** 
Not applicable.

protheory8:
Fixes issue #7387 by changing visible names of settings and by changing order of settings to [Circle Size, HP Drain, Accuracy, Approach Rate].

I used the solution of the user ICantDrive, because it seems good enough for now.

### DIFF
--- a/osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
@@ -20,7 +20,7 @@ namespace osu.Game.Rulesets.Osu.Mods
             Value = 5,
         };
 
-        [SettingSource("Drain Rate", "Override a beatmap's set HP.")]
+        [SettingSource("HP Drain", "Override a beatmap's set HP.")]
         new public BindableNumber<float> DrainRate { get; } = new BindableFloat
         {
             Precision = 0.1f,
@@ -30,7 +30,7 @@ namespace osu.Game.Rulesets.Osu.Mods
             Value = 5,
         };
 
-        [SettingSource("Overall Difficulty", "Override a beatmap's set OD.")]
+        [SettingSource("Accuracy", "Override a beatmap's set Accuracy.")]
         new public BindableNumber<float> OverallDifficulty { get; } = new BindableFloat
         {
             Precision = 0.1f,
--- a/osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
@@ -20,6 +20,26 @@ namespace osu.Game.Rulesets.Osu.Mods
             Value = 5,
         };
 
+        [SettingSource("Drain Rate", "Override a beatmap's set HP.")]
+        new public BindableNumber<float> DrainRate { get; } = new BindableFloat
+        {
+            Precision = 0.1f,
+            MinValue = 1,
+            MaxValue = 10,
+            Default = 5,
+            Value = 5,
+        };
+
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
```diff
@@ -32,17 +52,21 @@ public class OsuModDifficultyAdjust : ModDifficultyAdjust
 
         protected override void TransferSettings(BeatmapDifficulty difficulty)
         {
-            base.TransferSettings(difficulty);
+            // base.TransferSettings(difficulty);
```
bdach: Commented code should either be uncommented, or deleted. At current I don't know what the reason for suppressing this is, and it's likely an incorrect thing to do too.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -26,7 +26,7 @@ public abstract class ModDifficultyAdjust : Mod, IApplicableToDifficulty
 
         public override Type[] IncompatibleMods => new[] { typeof(ModEasy), typeof(ModHardRock) };
 
-        [SettingSource("Drain Rate", "Override a beatmap's set HP.")]
+        [SettingSource("HP Drain", "Override a beatmap's set HP.")]
```
bdach: What is the reason behind these name changes?

osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
```diff
@@ -20,6 +20,26 @@ public class OsuModDifficultyAdjust : ModDifficultyAdjust
             Value = 5,
         };
 
+        [SettingSource("HP Drain", "Override a beatmap's set HP.")]
+        new public BindableNumber<float> DrainRate { get; } = new BindableFloat
```
bdach: These properties should not be duplicated.

bdach (changes requested):
I think this solution is incorrect - as [this comment mentions](https://github.com/ppy/osu/issues/7387#issuecomment-569521110) a more robust one is needed. I'm envisioning an `Order` property on the `SettingSource` attribute that then would be [used in `SettingSourceExtensions.CreateSettingsControls`](https://github.com/ppy/osu/blob/f3ec98d85f5fe0a5a925ca2f72ddbf7d55b59e7f/osu.Game/Configuration/SettingSourceAttribute.cs) to appropriately order the settings controls.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
smoogipoo: To make them consistent with the beatmap detail area. But that place should be changed instead - the mods were already named correctly.

ProTheory8:
Thank you, I will try to implement your idea or come up with another solution.

osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
ProTheory8: Got it.

ProTheory8:
Should I make a new PR when I'm done or should I add commits to this one?

peppy:
Please make a new PR. This one is completely incorrect in its direction.

PR #7572, opened 2020-01-20
Apply OnDrag, OnDragEnd and OnMouseUp method signature refactorings

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3190

### DIFF
--- a/osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
+++ b/osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
@@ -177,17 +177,19 @@ namespace osu.Game.Overlays.KeyBinding
             return true;
         }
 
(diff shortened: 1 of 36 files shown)


osu.Game/Overlays/KeyBinding/KeyBindingRow.cs
```diff
@@ -313,14 +315,6 @@ public ClearButton()
                 Size = new Vector2(80, 20);
             }
 
-            protected override bool OnMouseUp(MouseUpEvent e)
-            {
-                base.OnMouseUp(e);
-
-                // without this, the mouse up triggers a finalise (and deselection) of the current binding target.
-                return true;
```
smoogipoo: Tested this vs public build, there doesn't seem to be a difference.

peppy (approved):

PR #7577, opened 2020-01-20
Make Difficulty Adjust mod customisation menu open automatically

### LINKED ISSUE #7514: Difficulty adjustment panel should open automatically
**Describe the new feature:** 

Difficulty adjustment panel should open automatically when adding a mod that supports adjustments, and "Customize" button should be removed because it is completely redundant.

The fact that a user adds Difficulty Adjust mod already signifies that they wish to drag sliders that adjust the beatmap. It is inconvenient to click a button to reveal what should already be visible.

protheory8:
Closes #7514

Other mods need to overwrite `RequiresConfiguration` property to make customisation menu open automatically on selection.


osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -474,6 +474,7 @@ private void modButtonPressed(Mod selectedMod)
             {
                 if (State.Value == Visibility.Visible) sampleOn?.Play();
                 DeselectTypes(selectedMod.IncompatibleMods, true);
+                if (selectedMod.RequiresConfiguration) ModSettingsContainer.Alpha = 1;
```
mcendu: Simply doing this can cause a bug in which the first click of the customization button would have no effect.

Also strongly recommend testing.

mcendu (changes requested):
Just reading the code, without actually testing.

osu.Game/Overlays/Mods/ModSelectOverlay.cs
ProTheory8: I'm probably missing something here, but if I remember correctly customize button executes `ModSettingsContainer.Alpha = ModSettingsContainer.Alpha == 1 ? 0 : 1`, so that shouldn't happen?

iiSaLMaN: From a code readability prespective, since there is more than one `if` statement here, would be better to add empty line between each line.

frenzibyte (changes requested):

osu.Game/Overlays/Mods/ModSelectOverlay.cs
iiSaLMaN: Seems correct to me on testing.

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -60,6 +60,12 @@ public abstract class Mod : IMod, IJsonSerializable
         [JsonIgnore]
         public virtual bool Ranked => false;
 
+        /// <summary>
+        /// Returns true if this mod requires configuration. If true, opens mod customisation menu every time user enables mod.
```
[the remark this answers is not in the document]
ProTheory8: Thank you

osu.Game/Overlays/Mods/ModSelectOverlay.cs
ProTheory8: Resolved, I guess?

peppy:
As mentioned, please add a test to `TestSceneModSettings`

ProTheory8:
What should I exactly test?

peppy:
- Add a test mod with the new boolean.
- Select the test mod.
- Ensure the customisation panel opened.

ProTheory8:
Thank you.

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -53,6 +55,16 @@ public void TestButtonShowsOnModAlreadyAdded()
             AddAssert("button enabled", () => modSelect.CustomiseButton.Enabled.Value);
         }
 
+        [Test]
+        public void TestCustomisationOpensOnModSelect()
+        {
+            createModSelect();
+
+            AddStep("open", () => modSelect.Show());
```
mcendu: Should we have to check that customization is not opened before the selection?

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -25,6 +25,8 @@ public class TestSceneModSettings : OsuTestScene
 
         private readonly Mod testCustomisableMod = new TestModCustomisable1();
 
+        private readonly Mod testCustomisableAutoOpenMod = new TestModCustomisable3();
```
mcendu:
```suggestion
        private readonly Mod testCustomisableAutoOpenMod = new TestModCustomisable2();
```

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -98,7 +110,8 @@ public override IEnumerable<Mod> GetModsFor(ModType type)
                         return new Mod[]
                         {
                             new TestModCustomisable1(),
-                            new TestModCustomisable2()
+                            new TestModCustomisable2(),
+                            new TestModCustomisable3()
```
mcendu:
```suggestion
```

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
ProTheory8: OK, I'll do that

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
ProTheory8: Thank you, but I'm confused, why? `TestModCustomisable2` doesn't override `RequiresConfiguration`. Should I remove Customisable3 and override property in Customisable2 instead?

mcendu: Currently `TestModCustomizable2` has no use, so I proposed that.

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -53,6 +55,17 @@ public void TestButtonShowsOnModAlreadyAdded()
             AddAssert("button enabled", () => modSelect.CustomiseButton.Enabled.Value);
         }
 
+        [Test]
+        public void TestCustomisationOpensOnModSelect()
```
bdach: This test case needs more work:

* Need to also test that deselecting all mods after selecting with customisation hides the settings
* Running this test twice will cause it to fail the second time since the customisation mod remains selected between test runs

Could you describe what sort of issues you're having with the running tests so we can help resolve them? It's not really a good idea to write tests blindly without being able to run them, kind of defeats the purpose of the entire endeavour.

bdach (changes requested):

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
ProTheory8: So interestingly enough, tests work fine on Windows and don't work on Linux for me, unless I'm doing it wrong.
To run the tests I used `dotnet test "/home/protheory8/osu/osu.Game.Tests/osu.Game.Tests.csproj"` and the same command on Windows with a different path.

The output on Linux:
`Test run for /home/protheory8/osu/osu.Game.Tests/bin/Debug/netcoreapp3.1/osu.Game.Tests.dll`
`(.NETCoreApp,Version=v3.1)`
`Microsoft (R) Test Execution Command Line Tool Version 16.3.0`
`Copyright (c) Microsoft Corporation.  All rights reserved.`

`Starting test execution, please wait...`

`A total of 1 test files matched the specified pattern.`
`  ! TestParity("Soleily - Renatus (Gamu) [Insane].osu")`
`  ! TestParity("Within Temptation - The Unforgiving (Armin) [Marathon].osu")`

`The active test run was aborted. Reason: Test host process crashed`

`Test Run Aborted.`
`Total tests: Unknown`
`     Passed: 77`
`    Skipped: 2`
` Total time: 20.5240 Seconds`

smoogipoo: If there's no exception provided, you may want to run with the `-v diag` option.

bdach: I recall something in `libbassfx` segfaulting on me when trying to run tests headless on Linux (outside of test browser), but never tried to diagnose in detail. You can check `dmesg` output to confirm.

smoogipoo (approved):
Seems ok

ProTheory8:
Do I still need to improve the test?

peppy:
Would be great, yeah. Looks like that was missed before merge.

PR #7580, opened 2020-01-21
Replace xK mods with single customizable mod

### LINKED ISSUE #7576: Key conversion mode should use customization menu
The way it works right now is really clunky (Maybe put it into the difficulty adjust)

mcendu:
Fixes #7576


peppy:
I'm a bit unsure about this being the correct direction. Will have to have a think about it.

peppy:
Closing this for now. While on the surface this looks fine, it's going to leave a lot to be considered in legacy handling (scores, diffcalc, etc.)

We can reapproach this once the game is done!

PR #7582, opened 2020-01-21
Allow parsing hex colour codes with alpha

raoits:
added the capability to use rrggbbaa and rgba hex codes with `OsuColour`.
i tried to make it work the same as css hex codes.
[ccs notation](https://drafts.csswg.org/css-color/#hex-notation)
i made a quick example of a 8 diget hex code with the osu logo
image image
i dont know if it will extually be usefull lateron.

edit: now committed from separate branch

### DIFF
--- a/osu.Game/Graphics/OsuColour.cs
+++ b/osu.Game/Graphics/OsuColour.cs
@@ -36,6 +36,20 @@ namespace osu.Game.Graphics
                         Convert.ToByte(hex.Substring(2, 2), 16),
                         Convert.ToByte(hex.Substring(4, 2), 16),
                         255);
+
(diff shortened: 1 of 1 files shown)


frenzibyte (approved):

bdach (approved):
The verified github checkmark doesn't instill too much confidence, but it *looks* correct, so I suppose not much reason to reject...

swoolcock:
This should be at framework-level, once we’ve decided which path to take in removing osuTK `Color4`.

peppy:
Looks fine for now, to complete the existing implementation. We can decide what happens to it in the future.

PR #7591, opened 2020-01-23
Improve song select display on ultrawide displays (or when UI scale is set low)

peppy:
Before:

image

After:

image

### DIFF
--- a/osu.Game/Screens/Select/BeatmapInfoWedge.cs
+++ b/osu.Game/Screens/Select/BeatmapInfoWedge.cs
@@ -33,7 +33,7 @@ namespace osu.Game.Screens.Select
     {
         private const float shear_width = 36.75f;
 
-        private static readonly Vector2 wedged_container_shear = new Vector2(shear_width / SongSelect.WEDGED_CONTAINER_SIZE.Y, 0);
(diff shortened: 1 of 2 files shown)


power9maker:
Would it be possible to make a slider in settings to adjust size of this?

peppy:
There is a slider already (UI Scale).

power9maker:
Yes I know, but I mean a slider for how far apart songs (on the right) and song description (on the left) are. So you can go back to old style if you want to.

peppy:
No, this is unnecessary.

power9maker:
Just checked this and discovered you can't absolute scroll with right click on this black part
AbsoluteScroll

peppy:
That sounds correct. Right click scroll should only apply within the carousel confines IMO.

power9maker:
On stable you can right click basically anywhere to absolute scroll, while on lazer you have to aim between this black box I highlighted on screen in previous comment and the actuall songs. This area on my 21:9 isn't that big, which makes absolute scrolling less convenient and slower to use. 
Idk what others think about it, but that's what I do.

peppy:
We'll likely improve the absolute scroll mechanic with a better scroll bar in the future.

smoogipoo (approved):
Seems to look okay.

PR #7592, opened 2020-01-23
Add TotalCommentsCounter to CommentsContainer

EVAST9919:


osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -133,6 +135,7 @@ public void ShowComments(CommentableType type, long id)
             if (!IsLoaded)
                 return;
 
+            commentCounter.Current.Value = 0;
```
iiSaLMaN: Should probably reside in `clearComments`. (where the comments amount would really become 0)

frenzibyte (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: Looks bad visually. `clearComments` called on every sort change

iiSaLMaN: Then there should probably be an in-line comment saying that.

PR #7593, opened 2020-01-23
Minor cleanups for Legacy Storyboard/Beatmap decoder

turbedi:
Mostly cosmetical, except for the changes in `LegacyStoryboardDecoder.handleEvents()`.
This way we only call `Substring` once at the end of the loop and not after every iteration. I will test the actual difference in a few minutes.


osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
```diff
@@ -300,13 +300,13 @@ private void handleEvent(string line)
             switch (type)
             {
                 case LegacyEventType.Background:
-                    string bgFilename = split[2].Trim('"');
-                    beatmap.BeatmapInfo.Metadata.BackgroundFile = bgFilename.ToStandardisedPath();
+                    string bgFilename = split[2];
+                    beatmap.BeatmapInfo.Metadata.BackgroundFile = CleanFilename(bgFilename);
```
smoogipoo: Put onto one line

osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
```diff
@@ -300,13 +300,13 @@ private void handleEvent(string line)
...
                     break;
 
                 case LegacyEventType.Video:
-                    string videoFilename = split[2].Trim('"');
-                    beatmap.BeatmapInfo.Metadata.VideoFile = videoFilename.ToStandardisedPath();
+                    string videoFilename = split[2];
+                    beatmap.BeatmapInfo.Metadata.VideoFile = CleanFilename(videoFilename);
```
smoogipoo: Same here

osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
```diff
@@ -65,12 +64,14 @@ protected override void ParseLine(Storyboard storyboard, Section section, string
...
+            var lineSpan = line.AsSpan();
 
-            while (line.StartsWith(" ", StringComparison.Ordinal) || line.StartsWith("_", StringComparison.Ordinal))
+            while (lineSpan.StartsWith(" ", StringComparison.Ordinal) || lineSpan.StartsWith("_", StringComparison.Ordinal))
             {
+                lineSpan = lineSpan.Slice(1);
                 ++depth;
-                line = line.Substring(1);
             }
+            line = line.Substring(depth);
```
smoogipoo: Maybe just `lineSpan.ToString()`.

smoogipoo (changes requested):

peppy:
@turbedi make sure to also [make inspectcode happy](https://ci.appveyor.com/project/peppy/osu/builds/30312143/messages).

osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
turbedi: I tried both, and using `Substring` seems to be a bit faster (difference is within margin of error though). Should I still change to `lineSpan.ToString()`?

turbedi:
I tested it with the Storyboard of this beatmap: https://osu.ppy.sh/beatmapsets/470977#osu/1006822
Allocations went from 430.54 MB to 418.15 MB.

osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
bdach: If `Substring` doesn't have a significant performance advantage over even thousands of executions I'd change for the sake of readability.

peppy (approved):

peppy (approved):

PR #7595, opened 2020-01-24
Use OverlayColourScheme for CommentsContainer

EVAST9919:


peppy:
Let's change the way this whole colour stuff works before it gets too ingrained into the codebase.

- `OverlayHeader` should cache an `OverlayColourProvider` class. Each overlay should do the `base(OverlayColourScheme scheme)` call (what you currently have in `OverlayHeader`) and the base FullscrenOverlay or whatever will do the caching of this object for children use.
- All section HSL logic should be moved out of OsuColour into that class. It can derive from `OsuColour` if necessary.
- Rather than passing down the enum like you are right now, each usage location should BDL that class.

Also, saturation and lightness values are well defined and limited. Please remove all hardcoded values and expose the constants from `OverlayColourProvider` in a similar way to the colours are defined in `OsuColour`. See [wiki](https://github.com/ppy/osu-web/wiki/Colour-System) for reference (C1, C2, L1, L2 etc).

You're welcome to do this in another PR if that works better.

EVAST9919:
Yeah, I'll make a separate pr for this

EVAST9919:
Will be easier to rewrite it from scratch with remaining changes.

PR #7597, opened 2020-01-24
Implement OverlayColourProvider component

EVAST9919:


osu.Game/Overlays/OverlayColourProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            float hue;
+
+            switch (colourScheme)
+            {
+                default:
+                    throw new ArgumentException($@"{colourScheme} colour scheme does not provide a hue value in {nameof(getBaseHue)}.");
+
+                case OverlayColourScheme.Red:
+                    hue = 0;
```
peppy: just change these to `return` statements and inline the `/ 360f`. i think it will be easier to read.

osu.Game/Overlays/OverlayColourProvider.cs
```diff
@@ -0,0 +1,90 @@
...
+    public class OverlayColourProvider
+    {
+        private readonly OverlayColourScheme colourScheme;
+
+        public OverlayColourProvider(OverlayColourScheme colourScheme)
+        {
+            this.colourScheme = colourScheme;
+        }
+
+        private Color4 convert(float saturation, float lightness) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, 1));
```
peppy: convert isn't a great method name (it's not really converting). call it `getColourFromHSL` or even just `getColour`.

osu.Game/Overlays/FullscreenOverlay.cs
```diff
@@ -41,6 +39,23 @@ protected FullscreenOverlay()
             };
         }
 
+        private DependencyContainer dependencies;
+
+        protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent) =>
```
peppy: just use `[Cached]` instead of this. much simpler.

peppy (changes requested):

osu.Game/Overlays/FullscreenOverlay.cs
EVAST9919: I thought about it but looks like I'm missing something because I can't cache like  that since colourScheme isn't static
Аннотация 2020-01-26 123510

EVAST9919: Nevermind

peppy (approved):

PR #7620, opened 2020-01-26
Make daycore and nightcore rate slider also adjust pitch

mcendu:
Naive implementation currently; Should behave as e.g. one octave higher than base pitch when tempo is doubled.


peppy:
It was intentionally chosen to not adjust this.

PR #7624, opened 2020-01-26
Fix visual inconsistency in BreadcrumbControl

EVAST9919:


smoogipoo:
In the future, please add tests for regressions. It takes 5 minutes and it will help ensure this doesn't happen further down the line.

smoogipoo (approved):

PR #7625, opened 2020-01-26
Prevent mods from affecting Key Amount in osu!mania

### LINKED ISSUE #7610: Easy mod for osu!mania halves key amount stat
**Describe the bug:** 
If you use the Easy modifier in osu!mania mode, it seems to halve all stats - including key count (which it should not). This doesn't affect gameplay, just the key count number under the beatmap details.

With the Easy modifier enabled:
Easy mod enabled
Without the Easy modifier enabled:
Without easy mod

**osu!lazer version:** 
2020.118.0
(self-compiled on Ubuntu 19.10)

**Logs:** 
...

thewildtree:
Fixes #7610.

### DIFF
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModEasy.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModEasy.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Mods;
 
 namespace osu.Game.Rulesets.Mania.Mods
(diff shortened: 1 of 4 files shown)


peppy:
Hmm, not sure about this as a solution. It's basically a hack on a hack (circle size shouldn't be used for key count in the first place, optimally).

Requires further discussion, but leave open for now.

TheWildTree:
I agree. This also creates a weird scenario where if the current ruleset is mania, but the beatmap is from another mode, stats might be misleading:

image

Because, unlike stable, `AdvancedStats` changes descriptions based on beatmap's ruleset ID instead of the currently chosen mode.

smoogipoo:
Hmm, in the interim we're going to need something like this. Going forward we'll want per-mode beatmap settings, but that's still quite a way off I think.

peppy:
@TheWildTree please check my [alternative implementation](https://github.com/ppy/osu/pull/7652).

Reasoning for doing it this way: it keeps the hacky logic in one place so it's easy to fix in the future.

TheWildTree:
Heh, I asked about solving this exactly your way in the issue discussion and got recommended otherwise - but yeah, it's kinda hacky either way, so your solution might be better for a temporary one.

bdach:
My bad, genuinely thought that this way would be better. Different strokes and all that, I suppose.

TheWildTree:
Not blaming you, I thought that'd be better as well, happens:')

PR #7634, opened 2020-01-27
Add beat ticks to editor timeline

peppy:
image

Degrades editor performance while scrolling / playing back, but this is fine for now.


osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBeatLineDisplay.cs
```diff
@@ -0,0 +1,90 @@
...
+                        {
+                            Colour = colour,
+                            Height = height,
+                        });
+
+                        Add(new PointVisualisation(t)
+                        {
+                            Colour = colour,
+                            Anchor = Anchor.BottomLeft,
+                            Origin = Anchor.BottomLeft,
```
smoogipoo: Maybe bottom-centre origin these so the marker always lands exactly on the tick.

smoogipoo (approved):

PR #7635, opened 2020-01-27
Add combined accuracy meters

### LINKED ISSUE #7292: Combined error and colour score meter

mcendu:
Resolves #7292

### DIFF
--- a/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
+++ b/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
@@ -74,13 +74,15 @@ namespace osu.Game.Screens.Play.HUD
                 return;
 
             foreach (var c in Children)
+            {
                 foreach (var meter in c.Children)
                     meter.OnNewJudgement(result);
+            }
         }
 
         private void typeChanged(ValueChangedEvent<ScoreMeterType> type)
         {
-            Children.ForEach(c => c.ForEach(c => c.FadeOut(fade_duration, Easing.OutQuint)));
+            Children.ForEach(c => c.ForEach(meter => meter.FadeOut(fade_duration, Easing.OutQuint)));
 
             if (hitWindows == null)
                 return;
--- a/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
+++ b/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
@@ -115,16 +115,16 @@ namespace osu.Game.Screens.Play.HUD
                     createColour(true);
                     break;
 
+                case ScoreMeterType.CombinedBoth:
+                    createCombined(false);
+                    createCombined(true);
+                    break;
+
                 case ScoreMeterType.CombinedLeft:
                     createCombined(false);
                     break;
 
                 case ScoreMeterType.CombinedRight:
-                    createCombined(true);
-                    break;
-
-                case ScoreMeterType.CombinedBoth:
-                    createCombined(false);
                     createCombined(true);
                     break;
             }
--- a/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
+++ b/osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
@@ -10,7 +10,6 @@ using osu.Game.Configuration;
 using osu.Game.Rulesets.Judgements;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Screens.Play.HUD.HitErrorMeters;
-using osuTK;
 
 namespace osu.Game.Screens.Play.HUD
 {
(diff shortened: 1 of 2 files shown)


EVAST9919:
As I mentioned in https://github.com/ppy/osu/pull/7317 it would be better to make two separate dropdowns: one for position, another for type.
Like "position: left, right, (top, bottom in the future)" and "type: HitError, Colour, Combined"
(just IMO)

mcendu:
I am not sure how to automatically convert current selection to other schemes.

EVAST9919:
> I am not sure how to automatically convert current selection to other schemes.

Well, this can be made in a separate pr then, so no worries

osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
```diff
@@ -116,12 +153,21 @@ private void createColour(bool rightAligned)
...
+        }
+
+        private void createCombined(bool rightAligned)
+        {
+            createBar(rightAligned);
+            createColour(rightAligned);
         }
 
-        private void completeDisplayLoading(HitErrorMeter display)
+        private void completeDisplayLoading(HitErrorMeter display, bool rightAligned)
```
bdach: `rightAligned` seems not too future-proof seeing that there are already requests for the meters to come back as bottom-center aligned (#5932). Enum would be more readable, too.

osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
```diff
@@ -53,12 +74,15 @@ private void onNewJudgement(JudgementResult result)
                 return;
 
             foreach (var c in Children)
-                c.OnNewJudgement(result);
+            {
+                foreach (var meter in c.Children)
+                    meter.OnNewJudgement(result);
+            }
```
bdach: I'm not too hot on this. Can we not reuse meter instances on both sides and just update the single ones?

bdach (commented):
I feel like this is a precarious piece of code that could get ugly. Maybe we want something like:
```csharp
[Flags]
enum ScoreMeterType
{
    // types
    None = 0,
    ErrorBar = 1 << 0,
    Colour = 1 << 1,
    // space for other types

    // locations - starting shift offset arbitrary
    Left = 1 << 8,
    Right = 1 << 9,
    Bottom = 1 << 10
}
```
so then you can inspect the meter type flags, create the meter instances, then inspect the meter location flags, add them wherever necessary, and done. (Or even two separate enums, although I'm not sure that's not *overly* verbose) I'm slightly concerned this is going to become switch city otherwise.

bdach:
Oh, and tests may be nice. They're not going to look too clever (probably going to boil down to inspecting if the drawables are where they need to), but who knows what can regress in the future.

peppy:
Alignment of these elements (and all UI elements) will be handled by a higher level component as part of the skinning/customisation engine. For now, positioning of individual elements should not be considered or implemented. Always having these drawn at the same location is fine for now.

mcendu:
> I feel like this is a precarious piece of code that could get ugly. Maybe we want something like:
> 
> ```cs
> [Flags]
> enum ScoreMeterType
> {
>     // types
>     None = 0,
>     ErrorBar = 1 << 0,
>     Colour = 1 << 1,
>     // space for other types
> 
>     // locations - starting shift offset arbitrary
>     Left = 1 << 8,
>     Right = 1 << 9,
>     Bottom = 1 << 10
> }
> ```
> 
> so then you can inspect the meter type flags, create the meter instances, then inspect the meter location flags, add them wherever necessary, and done. (Or even two separate enums, although I'm not sure that's not _overly_ verbose) I'm slightly concerned this is going to become switch city otherwise.

My concern here is that we need to exclude the individual flags from being listed by the options, and I am not sure whether there are any means to achieve this.

osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
mcendu: IMO this looks readable and can avoid duplicating code

bdach (commented):
If we're not considering positioning, maybe we should just axe the error bar location settings now and just stay with the types for the foreseeable future?

osu.Game/Screens/Play/HUD/HitErrorDisplay.cs
bdach: I meant the double update - if you have one meter type on both sides you get two instances that update independently. Could be nice to have just one, but displayed in two places, but that's a nitpick really.

peppy:
We'll probably still need the existing logic to some degree... I think?

This is a weird case because it isn't just repositioning but also duplicating, which is something i'm not sure we will support in the UI editor, but maybe we can/should?

smoogipoo:
The settings have been removed, and this is now supported by the skin editor since https://github.com/ppy/osu/pull/12845

PR #7636, opened 2020-01-27
Make CommentsContainer use OverlayColourProvider

EVAST9919:


osu.Game/Overlays/Comments/TotalCommentsCounter.cs
```diff
@@ -33,7 +34,7 @@ public TotalCommentsCounter()
                 Spacing = new Vector2(5, 0),
                 Children = new Drawable[]
                 {
-                    new OsuSpriteText
+                    text = new OsuSpriteText
```
peppy: Is there a reason for keeping all this drawable init in the ctor, rather than moving to bdl and assigning colours directly?

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -114,9 +110,10 @@ public CommentsContainer()
         }
 
         [BackgroundDependencyLoader]
-        private void load()
+        private void load(OverlayColourProvider colourProvider)
         {
-            background.Colour = colours.Gray2;
+            background.Colour = colourProvider.Background5;
```
peppy: pretty sure you can do the same refactor here. generally drawable hierarchies should be constructed in BDL not the constructor, because it allows for better async support.

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: I wasn't sure about that, will apply that in my future pr's as well

peppy (approved):

PR #7637, opened 2020-01-27
Refactor background creation in OverlayHeader

EVAST9919:


peppy:
Is there ever going to be a foreseeable case where the background is not a `OverlayHeaderBackground`? If so, we might want to consider using `virtual string BackgroundImage => string.Empty` instead of `CreateBackground`.

EVAST9919:
Well, player overlay is already an exception

peppy (approved):

PR #7639, opened 2020-01-27
Add top rank to the beatmap carousel

santerinogelainen:
I noticed that osu!lazer is missing ranks on the beatmap carousel itself. This is my proposal for a simple implementation of the feature. It currently shows the top result only from the local scores.

Example of how it will show in game:

image


osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
```diff
@@ -57,8 +60,26 @@ private void load(SongSelect songSelect, BeatmapManager manager, BeatmapSetOverl
                 editRequested = songSelect.Edit;
             }
 
-            if (manager != null)
-                hideRequested = manager.Hide;
+            if (beatmapManager != null)
+                hideRequested = beatmapManager.Hide;
+
+            var starCounterChildren = new List<Drawable>();
```
bdach: I'd add `starCounter` lower down alongside the entire hierarchy using initializers and only do `Insert(0, ...)` on the `DrawableRank` conditionally if there is a rank to display. Just seems to be more consistent with how things are done elsewhere.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
```diff
@@ -134,6 +157,14 @@ private void load(SongSelect songSelect, BeatmapManager manager, BeatmapSetOverl
             };
         }
 
+        private ScoreInfo getTopScore(ScoreManager scoreManager, int beatmapID, int rulesetID)
+        {
+            return scoreManager
+                .QueryScores(s => !s.DeletePending && s.Beatmap.ID == beatmapID && s.Ruleset.ID == rulesetID)
+                .OrderByDescending(s => s.TotalScore)
+                .FirstOrDefault();
+        }
```
bdach: Have you checked the performance implications of this? sqlite is in-memory, so it shouldn't be horrible, but I'm a bit concerned this will tank performance for tens of thousands of beatmaps even more than what we already have (and yes, people are actually importing that many). May be wise to go with some sort of caching component that will do a `GroupBy` and compute all of this in one shot instead of querying one by one, but that needs benchmarking.

Not necessarily an immediate concern but I imagine it's a potential issue.

bdach (commented):
Mind adding tests for this?

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
smoogipoo: Yeah this is one of those cases that should probably only be loaded when the drawable is displayed.

Furthermore, this shows e.g. SS ranks from osu! mode even if the user has osu!mania mode selected in the toolbar. Splitting this out into its own component that loads in a lazy manner will also help with responding to the currently-loaded ruleset.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
```diff
@@ -134,6 +157,14 @@ private void load(SongSelect songSelect, BeatmapManager manager, BeatmapSetOverl
             };
         }
 
+        private ScoreInfo getTopScore(ScoreManager scoreManager, int beatmapID, int rulesetID)
+        {
+            return scoreManager
+                .QueryScores(s => !s.DeletePending && s.Beatmap.ID == beatmapID && s.Ruleset.ID == rulesetID)
```
smoogipoo: As per the other comment, this should display scores from the user's selected ruleset rather than the beatmap's ruleset.

smoogipoo (changes requested):

osu.Game/Scoring/ScoreManager.cs
```diff
@@ -56,12 +56,44 @@ protected override ScoreInfo CreateModel(ArchiveReader archive)
             }
         }
 
+        public IEnumerable<ScoreInfo> GetScoresForBeatmapSet(int beatmapSetID, int rulesetID)
+        {
+            var scores = GetAllUsableScoresEnumerable()
+                .Where(s => !s.DeletePending && s.Beatmap.BeatmapSetInfoID == beatmapSetID && s.RulesetID == rulesetID);
```
bdach: What's the point of extracting `GetAllUsableScoresEnumerable` (which kind of lies with its name) if you're checking `DeletePending` the second time in the `Where` anyways? Sure, it'll probably be optimised by EF, but it *is* kind of sloppy.

osu.Game/Scoring/ScoreManager.cs
```diff
@@ -56,12 +56,44 @@ protected override ScoreInfo CreateModel(ArchiveReader archive)
...
+
+        public IEnumerable<ScoreInfo> GetScoresForBeatmap(int beatmapID, int rulesetID)
+        {
+            var scores = GetAllUsableScoresEnumerable()
+                .Where(s => !s.DeletePending && s.BeatmapInfoID == beatmapID && s.RulesetID == rulesetID);
+            scores = FilterByUser(scores);
+            return scores.ToList();
+        }
+
+        public IQueryable<ScoreInfo> FilterByUser(IQueryable<ScoreInfo> scores)
```
bdach: This is really confusing API design. A `ScoreManager` method, that *gets scores* from the outside, and uses the inner API provider (which up to now was used only to *download* scores), and then checks the *user*?

Should accept the *user* instead, not accept any scores, and *then* return a queryable.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -64,6 +65,8 @@ public class BeatmapCarousel : CompositeDrawable
         /// </summary>
         public bool BeatmapSetsLoaded { get; private set; }
 
+        public bool SelectedBeatmapSetRanksLoaded => selectedBeatmapSet?.RanksAreValid ?? true;
```
bdach: This shouldn't be exposed outwardly IMO.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -354,6 +357,25 @@ private void select(CarouselItem item)
...
+        {
+            foreach (var set in beatmapSets.Where(x => x.RanksAreValid))
+            {
+                set.InvalidateRanks();
+            }
+        }
+
+        public void UpdateRanks(int beatmapID, IEnumerable<ScoreInfo> scores)
+        {
+            var beatmapSet = beatmapSets.FirstOrDefault(s => s.Beatmaps.Any(b => b.Beatmap.ID == beatmapID));
```
```suggestion
            var beatmapSet = beatmapSets.SingleOrDefault(s => s.Beatmaps.Any(b => b.Beatmap.ID == beatmapID));
```

osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs
```diff
@@ -77,5 +82,7 @@ public override int CompareTo(FilterCriteria criteria, CarouselItem other)
         }
 
         public override string ToString() => Beatmap.ToString();
+
+        public void UpdateRank(ScoreInfo score) => rank.Value = score?.Rank;
```
bdach: Very weird - you're exposing a readonly `Rank` bindable that can't be changed and then a method to *mutate* the bindable. Just make the bindable publicly settable?

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -4,14 +4,20 @@
...
+using osu.Game.Scoring;
 using osu.Game.Screens.Select.Filter;
 
 namespace osu.Game.Screens.Select.Carousel
 {
     public class CarouselBeatmapSet : CarouselGroupEagerSelect
     {
+        private readonly Cached ranks = new Cached();
+
+        public bool RanksAreValid => ranks.IsValid;
```
bdach: Don't think this should be publicly exposed either.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
```diff
@@ -115,25 +124,62 @@ private void load(SongSelect songSelect, BeatmapManager manager, BeatmapSetOverl
...
+                {
+                    RelativeSizeAxes = Axes.None,
+                    Size = new Vector2(40, 20),
+                    Scale = new Vector2(0.8f)
+                });
+            }
+            else if (rank != null && score.HasValue)
+            {
+                rank.Rank = score.Value;
+            }
```
bdach: Now that I look at this logic I think `UpdateableRank` should just accept a null rank and not draw anything if it is passed a null. This is too much.

bdach (changes requested):
Obvious notes from a read-through of the code. Haven't ran.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -446,6 +464,34 @@ void run()
             }
         }
 
+        private IEnumerable<ScoreInfo> getScoresForBeatmapSet(int beatmapSetID)
+        {
+            if (api.IsLoggedIn && api.LocalUser.Value != null)
```
bdach: Invert and return empty list first to reduce nesting.

osu.Game/Scoring/ScoreManager.cs
```diff
@@ -56,12 +57,26 @@ protected override ScoreInfo CreateModel(ArchiveReader archive)
...
+        public IQueryable<ScoreInfo> GetLocalScoresForUser(User user)
+        {
+            var scores = GetAllUsableScoresEnumerable();
+
+            if (user != null)
+            {
+                scores = scores.Where(s => s.UserID.HasValue && s.UserID == user.Id);
+            }
+
+            return scores;
```
bdach: I'm not sure returning all scores for null user makes too much sense.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -26,6 +30,41 @@ public CarouselBeatmapSet(BeatmapSetInfo beatmapSet)
...
+            {
+                updateRanksForBeatmap(beatmap, scores);
+            }
+
+            ranks.Validate();
+        }
+
+        public void UpdateRanksForBeatmap(int beatmapID, IEnumerable<ScoreInfo> scores)
+        {
+            var beatmap = Beatmaps.FirstOrDefault(b => b.Beatmap.ID == beatmapID);
```
bdach: Should also probably be a `SingleOrDefault` that I didn't point out last time.

bdach (dismissed):

osu.Game/Online/Leaderboards/UpdateableRank.cs
```diff
@@ -7,23 +7,47 @@
...
+            switch (RelativeSizeAxes)
+            {
+                case Axes.Y:
+                    Width = Rank.HasValue ? DrawHeight * 2 : 0;
+                    break;
+
+                case Axes.X:
+                    Height = Rank.HasValue ? DrawWidth / 2 : 0;
+                    break;
+            }
```
bdach: This is not a good solution to the auto-sizing problem. I'd rather just revert to the previous version rather than allow this in.

osu.Game/Scoring/ScoreManager.cs
```diff
@@ -56,12 +57,22 @@ protected override ScoreInfo CreateModel(ArchiveReader archive)
...
+
         protected override IEnumerable<string> GetStableImportPaths(Storage stableStorage)
             => stableStorage.GetFiles(ImportFromStablePath).Where(p => HandledExtensions.Any(ext => Path.GetExtension(p)?.Equals(ext, StringComparison.OrdinalIgnoreCase) ?? false));
 
         public Score GetScore(ScoreInfo score) => new LegacyDatabasedScore(score, rulesets, beatmaps(), Files.Store);
 
-        public List<ScoreInfo> GetAllUsableScores() => ModelStore.ConsumableItems.Where(s => !s.DeletePending).ToList();
+        public List<ScoreInfo> GetAllUsableScores() => GetAllUsableScoresEnumerable().ToList();
+
+        public IQueryable<ScoreInfo> GetAllUsableScoresEnumerable() => ModelStore.ConsumableItems.Where(s => !s.DeletePending);
```
bdach: No reason for this to be `public` as far as I can tell? Might as well just do
```csharp
        public List<ScoreInfo> GetAllUsableScores() => getAllUsableScores().ToList();

        private IQueryable<ScoreInfo> getAllUsableScores() => ModelStore.ConsumableItems.Where(s => !s.DeletePending);
```

bdach (commented):
As far as the code is concerned, only a couple of minor nits remaining. The performance aspects of this change are yet to be presented in detail, however, and if I'm to review the performance implications myself it might take a long while until I've done that and am comfortable in approving the merge of this.

peppy:
Is there a reason all the score lookups and changes can't be done inside the `DrawableRank` or the `DrawableCarouselBeatmap`, rather than propagating over `SongSelect`, `BeatmapCarousel` and `CarouselBeatmapSet`?

santerinogelainen:
@peppy Originally I had the score fetching in DrawableCarouselBeatmap, but performance could have potentially been horrible if the scores were constantly being fetched as noted by others. I changed it so that the scores fetch only when the SelectedBeatmapSet is changed and if they have not been loaded yet. I tried many iterations of different types of implementations, fetching scores in BeatmapCarousel, DrawableCarouselBeatmapSet, DrawableCarouselBeatmap etc. but ended up with this since SongSelect already had selected RulesetInfo information and ScoreManager being injected into it in load(). If they are done in DrawableCarouselBeatmap etc. then the tests should also run in a clean game host for the score importation.

peppy:
We definitely can't keep the flow you have, so a first step would be moving the full retrieval inside `DrawableBeatmapCarousel` and we'll take things from there with regards to performance. I believe a good first step would be putting the rank display inside a DelayedLoadWrapper with a small delay, so it only loads after being on screen for a certain period of time.

Also, to avoid making changes to `UpdateableRank` itself (and other classes), you could subclass it to `BeatmapLocalScoreRank` (or some better name) which takes in the `BeatmapInfo` and performs all lookup, delay and display logic locally.

peppy (changes requested):
as proposed

santerinogelainen:
@peppy Here is the new implementation. I had the most trouble with thinking of a solution for when the user changes the ruleset. I made it so that it just reloads all the (loaded) ranks again. I think it could be done better in terms of performance but wasn't able to figure out a good solution. Any ideas?

osu.Game/Online/Leaderboards/TopLocalRank.cs
```diff
@@ -0,0 +1,78 @@
...
+            {
+                RelativeSizeAxes = Axes.Both
+            };
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(ScoreManager scores, IBindable<RulesetInfo> ruleset, IAPIProvider api)
+        {
+            this.scores = scores;
+            this.ruleset = ruleset;
```
smoogipoo: This causes the bindable to be copied, potentially resulting in a memory leak. You want to instead create a local bindable and use `.BindTo()`, or use `[Resolved]` which handles this for you.

osu.Game/Online/Leaderboards/TopLocalRank.cs
```diff
@@ -0,0 +1,78 @@
...
+            this.api = api;
+
+            FetchAndLoadTopScore();
```
smoogipoo: Why not have this component handle all local top score retrievals? Could this component and `CarouselBeatmapRank` be combined?

osu.Game/Online/Leaderboards/TopLocalRank.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        public Action<ScoreInfo> ScoreLoaded;
+
+        public TopLocalRank(BeatmapInfo beatmap)
+        {
+            this.beatmap = beatmap;
+
+            RelativeSizeAxes = Axes.Both;
+
+            InternalChild = rank = new UpdateableRank(null)
```
smoogipoo: I'm not sure about having this nesting of `DelayedLoadWrapper`s. One in `UpdateRank` (via it being a `ModelBackedDrawable`), and one in `CarouselBeatmapRank`.

If these two classes could be combined, the one in `CarouselBeatmapRank` could be removed and I'd feel safer about it.

osu.Game/Online/Leaderboards/TopLocalRank.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            ScoreLoaded?.Invoke(score);
+        }
+
+        private ScoreInfo fetchTopScore()
+        {
+            if (scores == null || beatmap == null || ruleset?.Value == null || api?.LocalUser.Value == null)
+                return null;
+
+            return scores.GetAllUsableScores()
```
smoogipoo: Use `scores.Query()` instead.

osu.Game/Screens/Select/Carousel/CarouselBeatmapRank.cs
```diff
@@ -0,0 +1,68 @@
...
+            if (score.BeatmapInfoID == beatmap.ID)
+            {
+                rank.FetchAndLoadTopScore();
+            }
+        }
+
+        private void scaleDisplay(ScoreInfo score)
+        {
+            if (score != null)
+                Width = rank_size * 2;
```
smoogipoo: Not sure what the point of this is really... Loads will not take a considerable amount of time, and this only increases complexity of implementation.

smoogipoo (changes requested):

osu.Game/Online/Leaderboards/TopLocalRank.cs
santerinogelainen: I changed this to `.QueryScores`. Is there a way to get the top score using only `.Query`? (`.OrderByDescending(s => s.TotalScore).FirstOrDefault()`)

osu.Game/Screens/Select/Carousel/CarouselBeatmapRank.cs
santerinogelainen: Since DrawableRank has RelativeSizeAxes = Axes.Both, the parents need to be sized accordingly. The rank is displayed before the starcounter, and if there is no top rank for the beatmap then it would create empty space between the starcounter and difficultyicon. This was just to scale the width, so that if there is no rank then the width would be 0 also. But I removed this logic in the new commit and changed it to Hide() / Show().

osu.Game/Online/Leaderboards/TopLocalRank.cs
smoogipoo: As you've done it is fine, point is to use a query. Previously it would retrieve/allocate all scores in the database and then match on them, whereas this translates into an SQL query that returns a few rows.

smoogipoo (approved):

smoogipoo:
I've applied a few refactorings. Of particular importance is the disposal override, and the scheduling of the rank update - which could otherwise cause changes to the hierarchy from a background thread (import) thread.

@peppy I think this PR is fine for now, please check.

Not super sure about the IsPresent override, but I can't think of a better way to do it - it needs to become not present to be removed from the flow layout.

peppy:
Also not hugely confident over the `IsPresent` usage but let's see how it survives in production.

PR #7642, opened 2020-01-28
Fix beat snap implementation being incorrect

peppy:
Was previously not accounting for the offset/time of the timing point being referenced.

This changed the signature of BeatSnap to be easier to understand. We may want to change `GetSnappedDurationFromDistance` to return a final time value rather than a duration-from-reference for cleanliness, unsure about this.


smoogipoo (dismissed):
seems ok

smoogipoo (approved):

PR #7643, opened 2020-01-28
Distance snap grid correct colouring

### LINKED ISSUE #7633: Distance snap grid display colours are incorrectly zeroed
They should match the timeline colouring, but currently use the previous placement object as a zero index instead of the contained control point.

peppy:
Makes colouring correctly match point-in-time colours, rather than based on the offset from previous placement.

- [x] Depends on https://github.com/ppy/osu/pull/7634.
- Closes https://github.com/ppy/osu/issues/7633


smoogipoo (approved):
seems ok

PR #7645, opened 2020-01-28
Allow OsuSliderBar tooltip to show as percentage as needed

### LINKED ISSUE #7640: OsuSliderBar Tooltip should only use percentage whenever necessary
Currently the `OsuSliderBar` automatically changes it's tooltip to a percentage when the minimum value is greater than or equal to -1 and maximum value is equal to 1. There are some cases such as https://github.com/ppy/osu/pull/7222 where we want to display the exact value.

The main approach is to use a flag (a boolean) to tell the slider bar when to show the percentage. We can either calculate the percentage regardless of the minimum and maximum values or still limit the minimum and maximum values to -1 and 1 respectively.

Opening this as an issue for visibility and discussion on how to approach said issue.

LeNitrous:
Resolves https://github.com/ppy/osu/issues/7640

Allows `OsuSliderBar` to display it's tooltip as percentage as needed and will work regardless of the minimum and maximum values set.


osu.Game/Graphics/UserInterface/OsuSliderBar.cs
```diff
@@ -35,6 +35,7 @@ public class OsuSliderBar<T> : SliderBar<T>, IHasTooltip, IHasAccentColour
         private readonly Container nubContainer;
 
         public virtual string TooltipText { get; private set; }
+        public bool DisplayAsPercentage { get; set; }
```
huoyaoyuan: Should updating `DisplayAsPercentage` trigger a change of tooltip? This is quite verbose and unlikely to be used, but keeps consistency.

osu.Game/Screens/Play/PlayerSettings/VisualSettings.cs
```diff
@@ -27,12 +27,18 @@ public VisualSettings()
                 {
                     Text = "Background dim:"
                 },
-                dimSliderBar = new PlayerSliderBar<double>(),
+                dimSliderBar = new PlayerSliderBar<double>
+                {
+                    DisplayAsPercentage = true
```
huoyaoyuan: Should `PlayerSliderBar` use percentage as default?

LeNitrous: Would probably leave as is to be consistent with it's base class.

osu.Game/Graphics/UserInterface/OsuSliderBar.cs
LeNitrous: I highly doubt there would be cases this property would be changed at run time and the tooltip text updating is handled here anyways https://github.com/ppy/osu/blob/a3cfeb08d41d3b4b8235b6422fc35c8d5cf16b0d/osu.Game/Graphics/UserInterface/OsuSliderBar.cs#L111
Regardless, it's for the sake of consistency

peppy: please xmldoc

osu.Game/Overlays/Settings/Sections/Audio/VolumeSettings.cs
```diff
@@ -17,10 +17,10 @@ private void load(AudioManager audio, OsuConfigManager config)
         {
             Children = new Drawable[]
             {
-                new SettingsSlider<double> { LabelText = "Master", Bindable = audio.Volume, KeyboardStep = 0.01f },
-                new SettingsSlider<double> { LabelText = "Master (window inactive)", Bindable = config.GetBindable<double>(OsuSetting.VolumeInactive), KeyboardStep = 0.01f },
-                new SettingsSlider<double> { LabelText = "Effect", Bindable = audio.VolumeSample, KeyboardStep = 0.01f },
-                new SettingsSlider<double> { LabelText = "Music", Bindable = audio.VolumeTrack, KeyboardStep = 0.01f },
+                new SettingsSlider<double> { LabelText = "Master", Bindable = audio.Volume, KeyboardStep = 0.01f, DisplayAsPercentage = true },
```
peppy: reformat to be on new line per object initialiser (same applies to following lines).

peppy (changes requested):

osu.Game/Graphics/UserInterface/OsuSliderBar.cs
```diff
@@ -169,11 +174,13 @@ private void updateTooltipText(T value)
...
             {
                 double floatValue = value.ToDouble(NumberFormatInfo.InvariantInfo);
-                double floatMinValue = CurrentNumber.MinValue.ToDouble(NumberFormatInfo.InvariantInfo);
                 double floatMaxValue = CurrentNumber.MaxValue.ToDouble(NumberFormatInfo.InvariantInfo);
 
-                if (floatMaxValue == 1 && floatMinValue >= -1)
-                    TooltipText = floatValue.ToString("P0");
+                if (DisplayAsPercentage)
+                {
+                    double percentage = floatValue / floatMaxValue;
```
bdach: Dicey for negative `floatValue`, and kind of nonsensical for negative `floatMaxValue` (even if mathematically correct). Maybe should throw something in those cases?

LeNitrous: I'm not sure when there would be a use case where the value and max value will both be negative but I think it's still safe enough that it won't warrant an exception. If there is, maybe it would be enough to make another PR of it.

peppy (approved):

PR #7651, opened 2020-01-28
Add test scene for slider snaking animation

mcendu:


peppy:
It would be useful if this scene also had a test where the slider didn't snake (covering all four combinations), since currently you wouldn't be able to test non-snaking sliders if you have your settings set to default (where snaking *is* already enabled).

mcendu:
Seems like that `TestSceneSlider` does not snake.

peppy:
As mentioned, this is because you have snaking disabled in your game-wide settings, and currently the visual test browser uses those.

Try changing your settings.

peppy:
If that's not the case, then settings in *that test* should be added to toggle snaking, rather than making a new test scene for this.

mcendu:
Closing in favor of #8489

PR #7652, opened 2020-01-28
Fix key count being incorrectly adjusted by hard/easy mods

### LINKED ISSUE #7610: Easy mod for osu!mania halves key amount stat

peppy:
- Alternative to and closes https://github.com/ppy/osu/pull/7625.
- Closes https://github.com/ppy/osu/issues/7610.

### DIFF
--- a/osu.Game/Screens/Select/Details/AdvancedStats.cs
+++ b/osu.Game/Screens/Select/Details/AdvancedStats.cs
@@ -117,9 +117,20 @@ namespace osu.Game.Screens.Select.Details
                     mod.ApplyToDifficulty(adjustedDifficulty);
             }
 
-            // Account for mania differences
-            firstValue.Title = (Beatmap?.Ruleset?.ID ?? 0) == 3 ? "Key Amount" : "Circle Size";
-            firstValue.Value = (baseDifficulty?.CircleSize ?? 0, adjustedDifficulty?.CircleSize);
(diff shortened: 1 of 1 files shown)


smoogipoo:
You'll still need to modify the mods to not affect the CS.

peppy:
How come?

smoogipoo:
Because the issue is that it changes the number of keys:

https://github.com/ppy/osu/blob/99d53982621739402831d94fc1b300a63eb571a6/osu.Game.Rulesets.Mania/Beatmaps/ManiaBeatmapConverter.cs#L48

smoogipoo:
Turns out this isn't actually the case...

https://github.com/ppy/osu/blob/99d53982621739402831d94fc1b300a63eb571a6/osu.Game/Rulesets/Mods/IApplicableToDifficulty.cs#L21

Veeeeeery sketchy though.

smoogipoo (approved):

PR #7655, opened 2020-01-28
Add placeholder for no comments case in CommentsContainer

EVAST9919:


osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
```diff
@@ -48,6 +48,7 @@ public TestSceneCommentsContainer()
             AddStep("Airman comments", () => comments.ShowComments(CommentableType.Beatmapset, 24313));
             AddStep("Lazer build comments", () => comments.ShowComments(CommentableType.Build, 4772));
             AddStep("News comments", () => comments.ShowComments(CommentableType.NewsPost, 715));
+            AddStep("Beatmap with no comments", () => comments.ShowComments(CommentableType.Beatmapset, 1291));
```
peppy: CleanShot 2020-01-29 at 11 08 35

beatmap has a comment.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -60,6 +62,27 @@ private void load(OverlayColourProvider colourProvider)
                             Sort = { BindTarget = Sort },
                             ShowDeleted = { BindTarget = ShowDeleted }
                         },
+                        noCommentsPlaceholder = new Container
```
peppy: why not make this its own component and load it into content just like you load comments, rather than making it an always-present-in-hierarchy element?

peppy (changes requested):

smoogipoo:
image

There's now a comment on this beatmap too. If you haven't realised yet, this needs to be changed to a value that correctly has no comments. Either a non-existing beatmap (e.g. `0`), or a custom test that provides no comments.

EVAST9919:
Well, I added `TestSceneCommentsPage` which have this type of test

smoogipoo:
Then perhaps remove the test altogether.

osu.Game/Online/API/Requests/Responses/CommentBundle.cs
```diff
@@ -47,17 +47,25 @@ public List<Comment> Comments
...
-                        c.IsVoted = true;
+                    Comments.ForEach(c =>
+                    {
+                        if (v == c.Id)
+                            c.IsVoted = true;
+                    });
                 });
-            });
+            }
         }
```
bdach: This setter has a potential bug. Let's say you do:

```csharp
bundle.UserVotes = new List<long> { 1, 2, 3 };
bundle.UserVotes = new List<long> { 4, 5 };
// comments w/ ID 1, 2, 3 will remain with IsVoted = true after set
```
which might not have been a problem when this was private, but is more of a problem now that you're exposing it publicly.

bdach (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -154,6 +155,7 @@ private void clearComments()
         {
             currentPage = 1;
             deletedChildrenPlaceholder.DeletedCount.Value = 0;
+            moreButton.Show();
```
peppy: why was this added in this PR? it doesn't look like a `Hide` call was added anywhere.

peppy (changes requested):

osu.Game/Overlays/Comments/CommentsContainer.cs
EVAST9919: It was missing because moreButton can be hidden at line 187 (`moreButton.FadeTo(response.HasMore ? 1 : 0);`)

EVAST9919: Anyway I changed it a little to make more sense

peppy (approved):

PR #7663, opened 2020-01-29
Fix presenting a beatmap from a different ruleset not working

### LINKED ISSUE #6908: Clicking imported beatmap notification from menu doesn't select correct beatmap
It transitions to song select but most times it doesn't select the correct beatmap. Likely due to the first selection logic that's changed with recent song select changes.

https://streamable.com/t0u48 (sorry for the huge video)

Needs tests.

peppy:
This also fixes a potential double-filter on entering song select.

Closes https://github.com/ppy/osu/issues/6908

@bdach i'd be interested in your review on this to see if the follow-ability of the song select code is any better as a result of these changes. It's still nowhere near perfect (I do have plans for how we can further improve), but I want to make sure I haven't regressed readability.


bdach (dismissed):
Admittedly it took me a better part of an hour to get a full understanding of what the fix is, but that might have just been time I needed to regain the context necessary for this again. I can however see how this makes sense, and after adding some debug log entries it's a bit clearer what the control flow is (I should have probably done that earlier).

All in all I'd say the change's a net positive when it comes to readability in my book, thanks for following up on this - I'm not sure I would have been able to conjure a fix like this 👍

peppy:
I should probably have mentioned what the fix was. Basically the initial filter was not being passed to the carousel early enough, causing the carousel to thing the beatmap is unavailable, and thus trying to present an empty beatmap (with the previous ruleset).

The change ensures the filter is applied immediately when song select isn't current. This was the intention, but it ended up being inside a `Schedule` call, which broke this functionality.

The rest of the changes in this PR are auxiliary fixes to allow removing the schedule altogether. This involved fixing a double-firing event in the `run()` portion of `performSelection`  when the ruleset was changed. A bit hard to explain but I think you may have figured the broken flow out in your debugging / breakpointing.

bdach:
Yup, your description of the fix is consistent with what I figured out on my own, so I think this is good to go.

bdach (approved):

smoogipoo (approved):

PR #7667, opened 2020-01-29
Fix editor being accessible for multiplayer song select

### LINKED ISSUE #7665: Editor is accessible via the multiplayer song select
**Describe the bug:** Going to multiplayer and right-clicking a difficulty in multiplayer allows you to access the editor.

image

**osu!lazer version:** master

**Logs:** N/A

peppy:
Closes https://github.com/ppy/osu/issues/7665.

We probably want to hide the context menu, but that's not really simple to do based on the current structure of things.


bdach (commented):
Looks straightforward on its own. Unsure if we want to have tests for this yet?

peppy:
I think it's failsafe enough that a test would be over-the-top for now.

smoogipoo (approved):

bdach (approved):

PR #7668, opened 2020-01-29
Update profile recent activities in line with the web design

EVAST9919:
Screenshot_2
Screenshot_3


TheWildTree:
A few differences I noticed that web has compared to your version:

- Bold + italic on player names
- Italic on mapset / diff names
- Date font isn't italic
- Score rank icon is a bit too big compared to the text ([comparison](https://user-images.githubusercontent.com/27722894/73387379-c3925780-42d0-11ea-99b2-1d2826bcaad9.png))
- Spacing between lines seems a bit bigger on web (or maybe the font / rank icon is just smaller, not sure)

Screenshot for comparison:
image

EVAST9919:
I believe font styles should be handled by MesageFormatter at some point (something like markdown support)

peppy:
A side-by-side comparison should be indistinguishable, so I agree that the above issues should be fixed for this PR. You can already style font correctly using a `TextFlowContainer`.

EVAST9919:
May be it's overkill, but now it's 1:1

TheWildTree:
I know this is nitpicky, but if we're going for 100% design parity with osu-web then: 

- spacing between lines should be *slightly* bigger

- could we make links properly coloured, so they match the overall scheme?

- the mode names in brackets should match osu-web unless that's an intentional difference

EVAST9919:
Not sure about links since they are blue everywhere and we may want to leave them consistent. Will see peppy's response on that

smoogipoo:
Keep the links blue for now.

osu.Game/Online/API/Requests/Responses/APIRecentActivity.cs
```diff
@@ -41,9 +41,38 @@ private string scoreRank
...
+                    case "taiko":
+                        Mode = "osu!taiko";
+                        return;
+
+                    case "fruits":
+                        Mode = "osu!catch";
+                        return;
+                }
+            }
+        }
```
bdach: This should not be a thing. The string before was *somewhat* acceptable, but now you have mixed in two magic strings in this class, one representing data returned from API and the other representing data *presentation*.

I'd either use an enum with `[Description]`s or use `RulesetInfo.ShortName` (higher up, not here directly) to get an actual ruleset object with a proper name.

bdach (dismissed):

EVAST9919:
Sorry for force-push. I broke this branch by pulling from my master one.

peppy:
Regarding link colouring, it's fine to implement the overlay-specific colours in its own PR if that is now feasible to do, falling back to default blue if no overlay colour scheme is available.

peppy (approved):

PR #7670, opened 2020-01-29
Bring UserProfileOverlay colour scheme in line with web

### LINKED ISSUE #7669: User profile page design issues
I know these are present mostly due to different parts of the UI being updated separately, but I couldn't find any existing issues related to this, so l guess we can track it here.

- Header, best performance / first place ranks and most played beatmaps listings use the new colour scheme, while the background, section tab control and section titles use the old one:

    image

- Section tab control behaves weirdly if we choose a section that's impossible to completely scroll to:

    

    Note that this isn't consistent, it might or might not happen depending on your aspect ratio / UI scale.

- Tabs 'flicker' while scrolling quickly past them, which looks weird when skipping sections:

    

- First tab is active right after opening the profile page, while on web it activates only when you scroll to the first section

- There are slight spacing / design differences near the tab switching section:
    
    lazer: 
    image
    web:
    image

Also, I'm curious, why is the color scheme for this page green on lazer, but pink on the web?

thewildtree:
Partially fixes #7669. Makes use of the `OverlayColourProvider`. Manually checked (probably) every colour variant on the web implementation, should match it 99%, with exceptions being the link sprites in the bottom part of the header.

Before: 

image

After:

image

### DIFF
--- a/osu.Game/Overlays/Profile/Header/Components/OverlinedInfoContainer.cs
+++ b/osu.Game/Overlays/Profile/Header/Components/OverlinedInfoContainer.cs
@@ -12,6 +12,11 @@ namespace osu.Game.Overlays.Profile.Header.Components
 {
     public class OverlinedInfoContainer : CompositeDrawable
     {
+        /// <summary>
+        /// The amount of space between the overline and the underneath text.
+        /// </summary>
+        private const float line_bottom_margin = 2;
+
         private readonly Circle line;
         private readonly OsuSpriteText title;
         private readonly OsuSpriteText content;
@@ -44,6 +49,7 @@ namespace osu.Game.Overlays.Profile.Header.Components
                     {
                         RelativeSizeAxes = Axes.X,
                         Height = 2,
+                        Margin = new MarginPadding { Bottom = line_bottom_margin }
                     },
                     title = new OsuSpriteText
                     {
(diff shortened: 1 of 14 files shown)


EVAST9919:
I was working on it, but okay

TheWildTree:
Sorry, I wouldn't do it if I had known beforehand.

EVAST9919:
Nah, it's okay, no worries

EVAST9919:
By the way, I believe at this point you can use pink colour scheme since colour schemes are already in place

TheWildTree:
True, I actually made the changes with pink scheme enabled for the ease of comparing to the websute, just wasn't sure if we want to change that permanently. I'm not on my PC anymore, will push that change tomorrow.

osu.Game/Overlays/Profile/Header/Components/OverlinedInfoContainer.cs
```diff
@@ -12,6 +12,11 @@ namespace osu.Game.Overlays.Profile.Header.Components
 {
     public class OverlinedInfoContainer : CompositeDrawable
     {
+        /// <summary>
+        /// The amount of space between the overline and the underneath text.
+        /// </summary>
+        private const float line_bottom_margin = 2;
```
bdach: No reason to introduce single-use constants. Using `2` directly where needed would also save on the xmldoc.

TheWildTree: honestly I just found another class that used a margin this way, but you're right, it's a bit overboard for a single use, will change

smoogipoo (dismissed):
lgtm!

smoogipoo (dismissed):

PR #4987, opened 2019-06-10
Linkify beatmap metadata displays

JackLee9355:
Addresses #4870. Did not implement a link for opening the BeatmapSetOverlay due to not knowing about the details button in the context menu at the time.

PR #7674, opened 2020-01-30
Linkify metadata in beatmap details and overlay

### LINKED ISSUE #4870: Adding links to beatmap info wedges, sources, and tags
**Describe the new feature:** 
I am hoping to make more of the beatmap information open internal links. To be exact I would like the sources and tags to open the search function. And for the info wedge to open the beatmap listing page.

I will be able to handle this if it gets approval.

My idea for the sources and tags are pretty straight forward.
source and tags Make those blue clickable links. My only concern with this is the text color will not match that of the descriptions. 

The first thought for the beatmap listing follows the same idea. 
title Make the title clickable. I fear this would suffer the same issue as the tags but much worse. As an alternative, and what I think would look better, I propose making the entire wedge clickable and link to the listing page.

LeNitrous:
Closes https://github.com/ppy/osu/issues/4870 supersedes https://github.com/ppy/osu/pull/4987.

Makes beatmap tags and source clickable that opens direct to search for that query.


osu.Game/Online/Chat/MessageFormatter.cs
```diff
@@ -279,7 +279,8 @@ public enum LinkAction
         JoinMultiplayerMatch,
         Spectate,
         OpenUserProfile,
-        Custom
+        Custom,
+        OpenDirectWithSearch
```
iiSaLMaN: should be above `Custom` to organize it better. ("otherwise / else / custom" enum should be last)

osu.Game/OsuGame.cs
```diff
@@ -310,6 +314,12 @@ private void load(FrameworkConfigManager frameworkConfig)
         /// <param name="beatmapId">The beatmap to show.</param>
         public void ShowBeatmap(int beatmapId) => waitForReady(() => beatmapSetOverlay, _ => beatmapSetOverlay.FetchAndShowBeatmap(beatmapId));
 
+        /// <summary>
+        /// Show Direct with a query.
+        /// </summary>
+        /// <param name="query">The query to search for.null</param>
+        public void ShowDirectWithSearch(string query) => waitForReady(() => direct, _ => direct.ShowWithSearch(query));
```
iiSaLMaN: to be consistent with other functions naming-side, this should be renamed to `SearchBeatmap` or similar.

osu.Game/Online/Chat/MessageFormatter.cs
iiSaLMaN: this should be renamed too.

osu.Game/Overlays/BeatmapSet/Info.cs
```diff
@@ -162,12 +185,12 @@ public MetadataSection(string title)
                 {
                     header = new OsuSpriteText
                     {
-                        Text = title,
+                        Text = this.type.ToString(),
```
iiSaLMaN: `this.` should not be used if unnecessary.

frenzibyte (changes requested):
I really don't like how there is almost an exact duplicate of `MetadataSection` in `BeatmapSet/Info.cs`, there should be one **public** class providing all the metadata section logic.

osu.Game/Online/Chat/MessageFormatter.cs
LeNitrous: is `SearchBeatmap` good?

iiSaLMaN: `SearchBeatmapSet`. (renamed my comment)

osu.Game/Overlays/BeatmapSet/Info.cs
mcendu: Does CI allows this?

bdach (changes requested):
For some reason CI didn't even start for the latest commit here, but I imagine it'd fail anyway due to whitespace issues & redundant imports; please run InspectCode locally and fix them up.

Definitely agreed that the fragments that are duplicated should not be, an extraction of at least a helper method is warranted.

LeNitrous:
I've done local tests before pushing and there were no issues whatsoever. AppVeyor is complaining about something regarding collisions.

osu.Game/Overlays/BeatmapSet/Info.cs
```diff
@@ -147,13 +151,32 @@ public string Text
...
+
                     textFlow.Clear();
-                    textFlow.AddText(value, s => s.Font = s.Font.With(size: 14));
+                    static void format(SpriteText t) => t.Font = t.Font.With(size: 14);
+
+                    switch(type)
+                    {
+                        case MetadataType.Tags:
+                            foreach (string tag in value.Split(" "))
+                                textFlow.AddLink(tag + " ", LinkAction.OpenDirectWithSearch, tag, null, format);
```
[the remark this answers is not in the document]
LeNitrous: is it fine to do something like this?
```c#
foreach (string tag in value.Split(" "))
{
    textFlow.AddLink(tag, LinkAction.SearchBeatmapSet, tag, null, format);
    textFlow.AddText(" ", format);
}
```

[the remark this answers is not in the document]
bdach: Certainly better, although you'd technically be adding a superfluous space at the end this way. To be perfect you probably want to change to a manual indexed `for` loop and break out of it between those lines when `index == Length - 1`.

bdach:
There seemed to be some ongoing changes to appveyor today, judging from the fact they've changed their webhook message format on discord, so it might have been some deployment mishap on their part. Should hopefully go away next time you push.

osu.Game/Overlays/BeatmapSet/Info.cs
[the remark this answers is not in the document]
LeNitrous: I'll be adding a whitespace instead since spacing would look more natural that way.

osu.Game/Overlays/BeatmapSet/Info.cs
```diff
@@ -147,13 +151,38 @@ public string Text
...
+                            break;
+
+                        case MetadataType.Source:
+                            textFlow.AddLink(value, LinkAction.SearchBeatmapSet, value, null, format);
+                            break;
+
+                        default:
+                            textFlow.AddText(value, format);
+                            break;
+                    }
```
bdach: Still think this could be extracted to a common place. Moreover, you could move `MetadataType` somewhere common too, since it's used by both places but placed under `BeatmapDetails`.

bdach (changes requested):
Still some whitespace issues remaining here.

peppy:
@LeNitrous Do you have any interest in finishing off this PR? I'm not sure there's that much more work to do to get it usable.

peppy:
I've fixed all the remaining issues with this PR. Need one extra review for sanity checks.

peppy (dismissed):

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -89,6 +89,12 @@ private void load()
             };
         }
 
+        public void ShowWithSearch(string query)
+        {
+            filterControl.Search(query);
+            Show();
+        }
```
frenzibyte, line 96: Currently performing a search from this method does not reset the current states of the filter control, this does not match osu!web's behaviour and sounds wrong in general.

frenzibyte (changes requested):
A bit weird that `MetadataSection` handles the spacing between itself and other sections, but not gonna bother much on that aspect.

osu.Game/Overlays/BeatmapListingOverlay.cs
bdach: I've looked into it and it's not as straightforward as you would imagine, for several reasons:

* No robust way to iterate over all bindables in the search control (you have to manually `.Clear()`/`.SetDefault()` them one by one)
* The former is also blocked by `IBindableList` not implementing `IHasDefault`
* `BeatmapSearchMultipleSelectionFilterRow` is written in a one-way manner (UI interactions change `Current`, but `Current` changes do not change UI state).

So I'd leave as a follow-up/prerequisite.

bdach:
I've fixed up the trivialities here. The point about other filters not being reset is salient but not easy to address here as stated above, so approving - that part can be tackled later on.

bdach (approved):

frenzibyte (approved):

PR #7676, opened 2020-01-30
Fix negative replay frames being played back incorrectly

### LINKED ISSUE #6477: Replay playback plays frames before gameplay start in unintended way
**Describe the bug:** 
On very start of the replay there's a random key smash and a small playfield jump.
Happens for replays only, autoplay is fine.
https://streamable.com/wd37g
**osu!lazer version:** 
2019.1011.0
**Logs:** 
Not needed, I guess.

peppy:
It turns out osu-stable inserts a frame at 0ms before potentially correct negative user frames. The logic we had would read this frame as valid, correctly seek back in time, then read the negative user frames (with positive `diff`) also valid, leading to out-of-order frames being played back.

CleanShot 2020-01-30 at 14 57 08

Closes https://github.com/ppy/osu/issues/6477.


smoogipoo (dismissed):

osu.Game/Scoring/Legacy/LegacyScoreParser.cs
```diff
@@ -234,8 +236,14 @@ private void readLegacyReplay(Replay replay, StreamReader reader)
                 }
 
                 var diff = Parsing.ParseFloat(split[0]);
+
                 lastTime += diff;
 
+                if (i == 0 && diff == 0)
+                    // osu-stable adds a zero-time frame before potentially valid negative user frames.
+                    // we need to ignroe this.
```
TheWildTree: typo

smoogipoo (approved):

PR #7678, opened 2020-01-30
Refactor performFromMenu to work with multiple screen targets

peppy:
- Changes behaviour to search up the screen stack, rather than only checking current screen.
- Adds ability to specify multiple valid target types rather than only one.
- Removes dead code.
- [x] Will add tests


osu.Game/OsuGame.cs
```diff
@@ -441,53 +441,42 @@ private void updateModDefaults()
         private ScheduledDelegate performFromMainMenuTask;
 
         /// <summary>
-        /// Perform an action only after returning to the main menu.
+        /// Perform an action only after returning to a specific screen specification.
```
bdach:
```suggestion
        /// Perform an action only after returning to a specific screen as indicated by <paramref name="validScreens"/>.
```

osu.Game/OsuGame.cs
```diff
@@ -441,53 +441,42 @@ private void updateModDefaults()
...
         /// Eagerly tries to exit the current screen until it succeeds.
         /// </summary>
         /// <param name="action">The action to perform once we are in the correct state.</param>
-        /// <param name="taskName">The task name to display in a notification (if we can't immediately reach the main menu state).</param>
-        /// <param name="targetScreen">An optional target screen type. If this screen is already current we can immediately perform the action without returning to the menu.</param>
-        /// <param name="bypassScreenAllowChecks">Whether checking <see cref="IOsuScreen.AllowExternalScreenChange"/> should be bypassed.</param>
-        private void performFromMainMenu(Action action, string taskName, Type targetScreen = null, bool bypassScreenAllowChecks = false)
+        /// <param name="validScreens">An optional collection of valid screen types. If any of these screens are already current we can immediately perform the action immediately, else the first valid parent will be made current before performing the action. <see cref="MainMenu"/> is used if not specified.</param>
```
bdach:
```suggestion
        /// <param name="validScreens">An optional collection of valid screen types. If any of these screens are already current we can perform the action immediately, else the first valid parent will be made current before performing the action. <see cref="MainMenu"/> is used if not specified.</param>
```

bdach (dismissed):
Besides some typically annoying xmldoc nits looks like something I thought about submitting ages ago, so I'm all for it 👍

bdach (dismissed):

smoogipoo (approved):

PR #7677, opened 2020-01-30
Decouple ModSelectOverlay from global SelectedMods

peppy:
Fixes potential issues when a child screen of song select changes mods (`ReplayPlayerLoader` in a future PR).

PR #7679, opened 2020-01-30
Return to song select after viewing a replay from the leaderboard

### LINKED ISSUE #5232: Wrong transition when going back from PlayerLoader/gameplay/result-screen of a replay
**Describe the bug:** Pressing `Esc` (or the Back/Home button in some scenarios) while in the PlayerLoader/gameplay/_after-gameplay result-screen_ of a replay doesn't send you to the `PlaySongSelect` as expected, but it sends you to the `MainMenu` ~(keeping the mods applied). This cause the bouncing animation for the icons to "break" with mods like DoubleTime~ (this part has already been fixed).

**Screenshots or videos showing encountered issue:** https://streamable.com/etwd6

**osu!lazer version:** 2019.704.0

peppy:
Definitively resolves #5232.

Previously the game would forcefully return to the main menu to play a replay, which made for a very bad user experience when viewing a replay from the song select leaderboard.
- [x] Depends on #7677.
- [x] Depends on #7678.
- [x] Will attempt to add a test, but no guarantee.


bdach:
Not approving just yet since you might have a test coming, but looks solid otherwise. I marked this as a resolution to #5232.

smoogipoo (approved):

PR #7683, opened 2020-01-30
Fix chat test intermittently failing

peppy:
Was throwing exception instead of returning false due to LINQ Single() call.


bdach (approved):
Passes my `[Repeat(10000)]` stress test, so definitely looks more solid 👍

smoogipoo (approved):

PR #7685, opened 2020-01-30
Add extra check to attempt to track down cause of test failures

peppy:
End goal is to fix regressed test failures that we cannot figure out the cause of (https://ci.appveyor.com/project/peppy/osu/builds/30461381/tests).

Likely a completely separate scene leaving the game bindable in a bad state.

### DIFF
--- a/osu.Game/Tests/Visual/ScreenTestScene.cs
+++ b/osu.Game/Tests/Visual/ScreenTestScene.cs
@@ -29,8 +29,9 @@ namespace osu.Game.Tests.Visual
 
         protected void LoadScreen(OsuScreen screen)
         {
-            if (Stack.CurrentScreen != null)
+            while (Stack.CurrentScreen != null)
                 Stack.Exit();
+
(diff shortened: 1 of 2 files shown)


osu.Game/Tests/Visual/ScreenTestScene.cs
```diff
@@ -29,8 +29,9 @@ protected ScreenTestScene()
 
         protected void LoadScreen(OsuScreen screen)
         {
-            if (Stack.CurrentScreen != null)
+            while (Stack.CurrentScreen != null)
                 Stack.Exit();
```
bdach: I'm pretty sure `TestScenePause` is causing this to enter an infinite loop due to `ValidForResume = true` at the end of some tests, which makes `Exit()` not unwind the stack.

peppy: good catch

peppy:
Reopening a new PR with a more resilient logic.

PR #7686, opened 2020-01-30
Fix cosmetic issues in UserProfileOverlay

thewildtree:
Corrects small issues that I missed in the recolor PR.

Before:
image

After:
image

Fixes part of #7669.


osu.Game/Overlays/Profile/Header/BottomHeaderContainer.cs
```diff
@@ -131,7 +131,11 @@ private void updateDisplay(User user)
             tryAddInfo(FontAwesome.Brands.Discord, user.Discord);
             tryAddInfo(FontAwesome.Brands.Skype, user.Skype, @"skype:" + user.Skype + @"?chat");
             tryAddInfo(FontAwesome.Brands.Lastfm, user.Lastfm, $@"https://last.fm/users/{user.Lastfm}");
-            tryAddInfo(FontAwesome.Solid.Link, websiteWithoutProtcol, user.Website);
+            tryAddInfo(FontAwesome.Solid.Link, websiteWithoutProtocol, user.Website);
+
+            // Hide the container to prevent adding unnecessary padding if it has no children other than the NewLine
+            if (!bottomLinkContainer.Children.Skip(1).Any())
+                bottomLinkContainer.Hide();
```
bdach: This check is really brittle since you're only checking the count. I'd say this at least needs to check if the only child element is a `NewLineContainer` for this to be solid. You could also refactor `tryAddInfo` slightly and do something like:

```csharp
bool anyInfoAdded = false;
anyInfoAdded |= tryAddInfo(FontAwesome.Solid.MapMarker, user.Location);
anyInfoAdded |= tryAddInfo(OsuIcon.Heart, user.Interests);
// ...
if (!anyInfoAdded)
    bottomLinkContainer.Hide();

private bool tryAddInfo(IconUsage icon, string content, string link = null)
{
    if (string.IsNullOrEmpty(content)) return false;
    // existing body
    return true;
}
```

bdach (dismissed):

osu.Game/Overlays/Profile/Header/BottomHeaderContainer.cs
TheWildTree: Did this by checking if the only child is a `NewLineContainer`, since it looks less cluttered IMO. Let me know if that's good enough.

bdach: I'm fine with the type check, I'm less fine with the double negation you now have in there. Reads better as
```csharp
if (bottomLinkContainer.Children.All(child => child is NewLineContainer))
```
IMO.

TheWildTree: believe me or not, I was wondering how I could improve it and completely forgot that `.All()` exists, looks like I need a bit of sleep. fixed

TheWildTree:
Temporarily recoloured the `RankGraphTooltip` to a neutral shade of gray that kind of matches osu-web. In the future this should use the `OverlayColourProvider`, but that requires some changes to how tooltips are being handled behind the scenes. I'll open an issue about this later.

osu.Game/Overlays/Profile/Header/Components/RankGraph.cs
```diff
@@ -270,7 +270,8 @@ public RankGraphTooltip()
             [BackgroundDependencyLoader]
             private void load(OsuColour colours)
             {
-                background.Colour = colours.GreySeafoamDark;
+                // To match osu-web, background and text should both be coloured using OverlayColourProvider
```
peppy: need to explain why it can't be, or at least link to an issue explaining why.

osu.Game/Overlays/Profile/Header/BottomHeaderContainer.cs
```diff
@@ -131,7 +132,14 @@ private void updateDisplay(User user)
             tryAddInfo(FontAwesome.Brands.Discord, user.Discord);
             tryAddInfo(FontAwesome.Brands.Skype, user.Skype, @"skype:" + user.Skype + @"?chat");
             tryAddInfo(FontAwesome.Brands.Lastfm, user.Lastfm, $@"https://last.fm/users/{user.Lastfm}");
-            tryAddInfo(FontAwesome.Solid.Link, websiteWithoutProtcol, user.Website);
+            tryAddInfo(FontAwesome.Solid.Link, websiteWithoutProtocol, user.Website);
+
+            // Hide the container to prevent adding unnecessary padding if it has no children other than the NewLineContainer
```
peppy: i'd prefer `tryAddInfo` didn't add anything in the case of null/empty string, rather than this check.

peppy (changes requested):

osu.Game/Overlays/Profile/Header/BottomHeaderContainer.cs
bdach: This could be accomplished if `tryAddInfo` was made to return `bool` [as I proposed earlier](https://github.com/ppy/osu/pull/7686#discussion_r373139469), and the addition of the new line was guarded by checking the `anyInfoAdded` flag.

TheWildTree: that check would still happen before attempting to add second-line info, resulting in the same situation we have on master if no further info was added afterwards.

imo the check has to stay, I could maybe change it to make use of the proposed `anyInfoAdded` instead of the type check, but it would still happen at the end and work virtually the same way with more clutter.

bdach: I see - the hide is necessary due to the parent fill flow setting vertical spacing.

There's one more benefit to the `anyInfoAdded` variant, however - the children check could be replaced by just checking `anyInfoAdded`, which is probably less surprising than checking children types.

I don't know about clutter, we use similar patterns in other places ([example](https://github.com/ppy/osu/blob/52b66fdc167059ca0336dde474e2ee84073d04b1/osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs#L24-L61)).

peppy: I've pushed a better (IMO) solution to this, please check it to make sure you agree.

bdach: I was initially opposed but after re-reading it a few times and considering alternatives it grew on me, so no objections here 👍

TheWildTree: Seems okay to me as well, except for the spacing difference now:

image

That's due to increased spacing in the combined container, here's how it looks on web (previous version here matched it):

image

Can this be done 1:1 without splitting the containers to change the newline spacing?

TheWildTree: After reverting to the 2-container setup and checking, it was actually not the NewLine's fault for adding the extra spacing - even if its not added at all, the spacing is there because `bottomLinkContainer` exists no matter what, which makes the parent `FillFlowContainer` add the specified spacing between its children.

To bring back proper spacing I think it would be best to go back to returning `bool` from `tryAddInfo` and just hide the bottom container if no info was added.

bdach: I think you could also keep 30a5835, remove spacing on the merged `linkContainer`, and insert *two* newlines if necessary after the first line. Sorta hacky, though. Not sure which is best.

TheWildTree: That sounds a bit too hacky for me. I decided to revert peppy's changes and do it [the way you proposed](https://github.com/ppy/osu/pull/7686#discussion_r373139469) at the very beginning.

Should be good to go, IMO.

bdach: I've fixed up a couple of remaining nits in this area, but other than that I think so too.

smoogipoo (approved):

PR #7687, opened 2020-01-30
Fix Alt+number shortcuts for tabs in chat overlay

### LINKED ISSUE #7608: Alt+[number] combination, in Chat Overlay, starts from number 2
**Describe the bug:** (Use the image below as the example for this explanation)

When I press <kbd>Alt</kbd>+<kbd>2</kbd>, it opens `#announce`; <kbd>Alt</kbd>+<kbd>3</kbd> opens `#italian`, etc.
So it seems like it's skipping the number 1.

chat

**osu!lazer version:** 2020.118.0

bdach:
Resolves #7608.

- [x] Depends on #7683 (didn't have to, I just wanted to get the inevitable merge conflict out of the way).

# Summary

Add test for <kbd>Alt</kbd>+[<kbd>0</kbd>-<kbd>9</kbd>] key shortcuts and resolve the regression by checking the channel type at the point of enumerating channels.

# Remarks

Since I didn't want to recreate the entire container for a single test, I just made the pre-existing tests take in 10 channels instead of the previous 3 and adjusted data-dependent logic appropriately. I'm aware that this introduces interdependence to the test cases, but I don't think it's that bad. I can separate them out if that's deemed preferable.

### DIFF
--- a/osu.Game/Overlays/ChatOverlay.cs
+++ b/osu.Game/Overlays/ChatOverlay.cs
@@ -323,8 +323,7 @@ namespace osu.Game.Overlays
         {
             var channel = ChannelTabControl.Items
                                            .Where(tab => !(tab is ChannelSelectorTabItem.ChannelSelectorTabChannel))
-                                           .Skip(index)
-                                           .FirstOrDefault();
+                                           .ElementAtOrDefault(index);
             if (channel != null)
                 ChannelTabControl.Current.Value = channel;
         }
--- a/osu.Game/Overlays/ChatOverlay.cs
+++ b/osu.Game/Overlays/ChatOverlay.cs
@@ -321,8 +321,11 @@ namespace osu.Game.Overlays
 
         private void selectTab(int index)
         {
-            var channel = ChannelTabControl.Items.Skip(index).FirstOrDefault();
-            if (channel != null && !(channel is ChannelSelectorTabItem.ChannelSelectorTabChannel))
+            var channel = ChannelTabControl.Items
+                                           .Where(tab => !(tab is ChannelSelectorTabItem.ChannelSelectorTabChannel))
+                                           .Skip(index)
+                                           .FirstOrDefault();
+            if (channel != null)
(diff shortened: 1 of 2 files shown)


swoolcock:
Tab-switching on macOS is generally done with cmd+number, so this should be a `PlatformAction`, or accept both.

bdach:
That's fair, it's definitely something I can follow up on, it might be better suited as a separate PR, though. (And I'd appreciate help testing it as a Mac-less person!)

osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -321,8 +321,11 @@ protected override void OnDragEnd(DragEndEvent e)
 
         private void selectTab(int index)
         {
-            var channel = ChannelTabControl.Items.Skip(index).FirstOrDefault();
-            if (channel != null && !(channel is ChannelSelectorTabItem.ChannelSelectorTabChannel))
+            var channel = ChannelTabControl.Items
+                                           .Where(tab => !(tab is ChannelSelectorTabItem.ChannelSelectorTabChannel))
+                                           .Skip(index)
+                                           .FirstOrDefault();
```
huoyaoyuan: `ElementAtOrDefault` should be enough

peppy: wow, that exists?!

huoyaoyuan: according to the document, it existed from the start of LINQ

peppy (approved):

bdach:
As for the suggestion to switch to <kbd>Cmd</kbd>/<kbd>Ctrl</kbd>+[<kbd>0</kbd>-<kbd>9</kbd>] - I just remembered that <kbd>Ctrl</kbd>+[<kbd>1</kbd>-<kbd>4</kbd>] are already used for switching rulesets in the top bar, so I'm going to leave this as-is for now to avoid conflicts.

PR #7692, opened 2020-01-31
Ensure OsuScreen leases are taken out synchronously

peppy:
We recently had some [very persistent test failures](https://ci.appveyor.com/project/peppy/osu/builds/30476210/tests) where bindables were altered when they shouldn't be (song select making changes while already in a leased state). The cause turns out to be the fact that leases were taken out in `CreateChildDependencies`, which is an **async context**.

Being hugely unsafe, this PR changes the flow to take out leases at the point of `Push` instead. This is now enforced, such that a screen cannot be preloaded via `LoadComponentAsync` if it requires leases (via `DisallowExternalBeatmapRulesetChanges`).


smoogipoo (approved):
Seems to work fine

PR #7695, opened 2020-01-31
Make use of ElementAtOrDefault() when possible

turbedi:


osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -321,7 +321,7 @@ protected override void OnDragEnd(DragEndEvent e)
 
         private void selectTab(int index)
         {
-            var channel = ChannelTabControl.Items.Skip(index).FirstOrDefault();
+            var channel = ChannelTabControl.Items.ElementAtOrDefault(index);
```
peppy: will conflict with existing PR

turbedi: Whoops, forgot about that. Will change it back in a sec.

peppy (approved):

PR #7703, opened 2020-02-01
Apply precision when determining bar colour in difficulty statistics

### LINKED ISSUE #7700: Difficulty Adjust wrongly adjusts Difficulty Statistics colours
**Describe the bug:** Even when at default difficulty value, the difficulty statistics show that it's red. Some difficulties do not show this behaviour. The values are all the same at both the statistics area and the adjustment panel.

**Screenshots or videos showing encountered issue:** https://streamable.com/kffcm

**osu!lazer version:** local debug at c904ae485

**Logs:** Not needed. This behaviour is not logged, I believe.

bdach:
Resolves #7700.

# Summary

In some cases, applying the Difficulty Adjust mod without actually changing any of the settings previously caused the bar in song select beatmap details to appear red/blue instead of staying white.

This was caused by not accounting for floating-point imprecisions when determining bar colour in `AdvancedStats`. To resolve, first check equality with tolerance, and only then apply blue/red colours if that equality check does not hold.

# Remarks

I wanted to test this properly, but modifying the existing test in `TestSceneBeatmapDetails` felt wrong, and so I went ahead and I factored it out to a separate test scene and exposed what I needed to test. Unfortunately this makes tests 90% of the diff. The actual fix is in 0bfadfb.

I also covered some other recent code (i.e. the mania "Key Size" text) with the tests.

### DIFF
--- a/osu.Game/Screens/Select/Details/AdvancedStats.cs
+++ b/osu.Game/Screens/Select/Details/AdvancedStats.cs
@@ -16,6 +16,7 @@ using System.Collections.Generic;
 using osu.Game.Rulesets.Mods;
 using System.Linq;
 using osu.Framework.Threading;
+using osu.Framework.Utils;
 using osu.Game.Configuration;
 using osu.Game.Overlays.Settings;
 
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Select/Details/AdvancedStats.cs
```diff
@@ -122,32 +124,33 @@ private void updateStatistics()
...
         {
             private const float value_width = 25;
             private const float name_width = 70;
 
             private readonly float maxValue;
             private readonly bool forceDecimalPlaces;
             private readonly OsuSpriteText name, valueText;
-            private readonly Bar bar, modBar;
+            private readonly Bar bar;
+            public readonly Bar ModBar;
```
peppy: As an aside, rather than exposing, you can use `ChildrenOfType<>()`. I'm neutral on whether this should be used in this case, up to you.

bdach: I thought about it, since that's the way it was originally done, but it felt brittle due to relying on the implementation detail of order of addition. I do however agree that this exposition is a bit sketchy as well. Let me think this over a little bit.

peppy: Fair, let's go with this for now.

peppy (approved):

PR #7704, opened 2020-02-01
Match osu-stable follow circle behaviour and size

thewildtree:
Brings `FollowCircle` behaviour in line with stable by introducing `InternalFollowCircle` (not sure about this name though) and making it scale instantly to `2.4f`, so the input check matches stable, while the visual part scales to `2f` on its own as it previously did.

Fixes part of #7521.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
@@ -24,6 +24,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 
         private readonly Slider slider;
         public readonly Drawable FollowCircle;
+        public readonly Drawable InternalFollowCircle;
         private readonly DrawableSlider drawableSlider;
 
         public SliderBall(Slider slider, DrawableSlider drawableSlider = null)
@@ -38,6 +39,13 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 
             Children = new[]
             {
+                InternalFollowCircle = new CircularContainer
+                {
+                    Origin = Anchor.Centre,
+                    Anchor = Anchor.Centre,
+                    RelativeSizeAxes = Axes.Both,
+                    Alpha = 0
+                },
                 FollowCircle = new FollowCircleContainer
                 {
                     Origin = Anchor.Centre,
@@ -95,7 +103,9 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 
                 tracking = value;
 
-                FollowCircle.ScaleTo(tracking ? 2f : 1, 300, Easing.OutQuint);
+                // Cursor input is checked against the internal circle, which scales instantly to match osu-stable behaviour
+                InternalFollowCircle.ScaleTo(tracking ? 2.4f : 1f);
+                FollowCircle.ScaleTo(tracking ? 2f : 1f, 300, Easing.OutQuint);
                 FollowCircle.FadeTo(tracking ? 1f : 0, 300, Easing.OutQuint);
             }
         }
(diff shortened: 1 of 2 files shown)


frenzibyte (changes requested):
~~That's a little bit of a hacky way to do this, what you probably want to do is create that circular container inside the `FollowCircleContainer` **internally**.~~

TheWildTree:
If done that way, would it let me scale the internal circle to a size bigger than the visible one (its parent)?

bdach:
I think it would if you specified a relative `Size` bigger than 1 (1.2 in this case, I suppose). I wouldn't say the current solution is *hacky* per se, but encapsulation is better, I agree.

TheWildTree:
Encapsulating it doesn't make sense to me, since they have to scale separately (one instantly, one over time) - if done as proposed, scaling will always be wonky for one of the circles (depending on which one we'd make the parent). Unless there is a way to decouple the scaling part, of course, but that seems counterproductive in a parent-child relationship.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
```diff
@@ -38,6 +39,13 @@ public SliderBall(Slider slider, DrawableSlider drawableSlider = null)
 
             Children = new[]
             {
+                InternalFollowCircle = new CircularContainer
```
iiSaLMaN: The name of this drawable is definitely misleading given what it does, how about `TrackingArea` or `FollowingArea`, with an in-line comment explaining why it's there. (and reference to osu-stable)

frenzibyte (changes requested):
On second thought, this may actually be fine currently FWIW. also if possible, would be nice to add tests following the slider with the cursor beyond the follow circle to ensure no issues with this.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
```diff
@@ -38,6 +39,14 @@ public SliderBall(Slider slider, DrawableSlider drawableSlider = null)
 
             Children = new[]
             {
+                // This is separate from the visible followcircle to ensure consistent internal tracking area (needed to match osu-stable)
+                TrackingArea = new CircularContainer
+                {
+                    Origin = Anchor.Centre,
+                    Anchor = Anchor.Centre,
+                    RelativeSizeAxes = Axes.Both,
+                    Alpha = 0
```
peppy: this shouldn't be required

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
```diff
@@ -24,6 +24,7 @@ public class SliderBall : CircularContainer, ISliderProgress, IRequireHighFreque
 
         private readonly Slider slider;
         public readonly Drawable FollowCircle;
+        public readonly Drawable TrackingArea;
```
peppy: i think this can be private

peppy (changes requested):

TheWildTree:
Added tests checking for the tracking area radius. Please let me know if that's the right way to test this - I'm a bit curious about tracking dropping on hitobjectradius * 1.2f (so theoretically on the edge of tracking area which is scaled to 2.4f), I thought it's gonna drop tracking only after passing that point (so at >1.2f), but I guess that's not the case.

TheWildTree:
Testing this behaves differently in visual testing than it does in CLI. In visual browser, cursor `circle_radius * 1.2f` away from the center was enough to drop tracking, while in CLI it needed >1.2f to drop. 

I don't know which one was behaving correctly, so for now I decreased the precision of tests to not trigger those edge cases, as they seem to be out of scope for this PR.

peppy:
I think that is fine for the time being. We will likely to a precision pass once we are closer to requiring this from replay playback / recording.

peppy:
Probably important to note (though I don't think it requires a comment) that the tests are using `hit object radius * 1.2`, which is only correct because it's using default CS5

peppy (approved):

smoogipoo:
What about our prior discussion where we concluded that the user should always see their input matching with something on screen? With this, not only does it not scale up but it also doesn’t match the visual representation anymore.

peppy:
You raise a good point, completely missed that this is actually going against that.

We may want to just fix the size / tracking region to match for the time being, and revert this fix until it can be applied in a legacy-only way.

PR #7707, opened 2020-02-01
Fix mod select overlay overflowing toolbar at max ui scale

### LINKED ISSUE #7369: Mod select overlay overflows at max ui scale
Just want to make sure this is right. First, we can make it relative size (not 100% though). We can then share the resize implementation in `ChatOverlay`, so portrait players won't be stuck with the higher height.

osu_2019-12-26_16-55-20

Joehuu:
Fixes https://github.com/ppy/osu/issues/7369. Uses grid container instead of what I proposed (making it resizeable). May be good to do in the future, when more mods are introduced.

For the triangles, all other uses which resize (direct, etc.) have relative axes on both. From what I know, the densities of those are wrong when opened too.

tI9Oacf3ai


bdach (dismissed):
Unfortunately this doesn't seem to be working properly, which you've even demonstrated in the PR description. When increasing UI scale mods can overflow the screen width and be partially or even fully off screen ([see video](https://streamable.com/moxnm)).

peppy:
Seems that some mods and text in the dialog end up off screen now?

Joehuu:
Not sure how you missed the toolbar in the PR title, but this just fixes the height of the overlay to not exceed the toolbar height. The issue you mentioned already exists in master and may just be another `FillFlowContainer` problem.

bdach:
Are you sure? You had submitted #6700 before, which seemed to be a fix for that same issue - doesn't that mean that this regressed somewhere?

bdach:
Upon closer inspection this seems to always have been broken. The flow brokenness will most likely end up being another case of ppy/osu-framework#223; sorry for the confusion.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -221,23 +221,37 @@ private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, S
 
             if (ShowFooter)
             {
-                AddRangeInternal(new[]
+                AddRangeInternal(new Drawable[]
                 {
-                    FooterPanels = new Container
+                    new GridContainer // used for max width implementation
```
peppy: max height?

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -221,23 +221,37 @@ private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, S
...
-                        AutoSizeAxes = Axes.Y,
-                        Margin = new MarginPadding { Bottom = Footer.HEIGHT },
-                        Children = new Drawable[]
+                        RelativeSizeAxes = Axes.Both,
+                        RowDimensions = new[]
                         {
-                            BeatmapOptions = new BeatmapOptionsOverlay(),
-                            ModSelect = new ModSelectOverlay
+                            new Dimension(),
+                            new Dimension(GridSizeMode.Relative, 1f, maxSize: 560),
```
peppy: need to know where 560 comes from, via a shared constant or otherwise.

peppy (dismissed):

smoogipoo (approved):
Seems to look okay 👍

PR #7710, opened 2020-02-02
Adjust profile scores to closer match osu-web

recapitalverb:
Screenshot_20200202_174641

### DIFF
--- a/osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs
+++ b/osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Linq;
@@ -23,8 +23,6 @@ namespace osu.Game.Overlays.Profile.Sections.Ranks
     public class DrawableProfileScore : CompositeDrawable
     {
         private const int performance_width = 80;
-        private const int content_padding_left = 10;
-        private const int content_padding_right = 30;
 
         protected readonly ScoreInfo Score;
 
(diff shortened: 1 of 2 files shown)


TheWildTree:
While you're at it, you could correct the colouring as well, because currently background colours are off when compared to the website:

Screenshot_20200202-124002_Chrome

Should be easy using `OverlayColourProvider`.

osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs
```diff
@@ -23,7 +23,8 @@ namespace osu.Game.Overlays.Profile.Sections.Ranks
     public class DrawableProfileScore : CompositeDrawable
     {
         private const int performance_width = 80;
-        private const int content_padding = 10;
+        private const int content_padding_left = 10;
+        private const int content_padding_right = 30;
```
bdach: These are now single-use constants that can be inlined.

osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs
```diff
@@ -142,17 +143,21 @@ private void load()
                         {
                             new Box
                             {
+                                Anchor = Anchor.TopRight,
+                                Origin = Anchor.TopRight,
                                 RelativeSizeAxes = Axes.Both,
-                                Size = new Vector2(1, 0.5f),
+                                Size = new Vector2(1.06f, 0.5f),
```
bdach: How was this size determined? Maybe it should be absolute?

recapitalverb: Nice catch. I agree it's better to make this absolute, because changing `performance_width` and `DrawableProfileScore.Height` shows that using relative widths isn't very robust. Currently working on a solution.

recapitalverb:
The failing test (https://ci.appveyor.com/project/peppy/osu/builds/30518465/tests) seems odd - locally testing this turns out fine. What should I do in this case?

bdach:
Seems wholly unrelated to the changes here, so you can leave this as-is. If you feel like it you could also try to reproduce the failure and find its cause, but even if you do find a fix, it's probably better to submit it separately.

recapitalverb:
These tests tell me that my code style is wrong, but it seems that I have also messed up modifiers for the `performance_background_width`. Will find a solution when possible.

bdach:
FYI you can check your code style by running `InspectCode.ps1` if on a Windows platform. Looks to me that you can also make that field `private static readonly` (it cannot be `const` due to the call to `MathF.Tan`).

PR #7712, opened 2020-02-02
Fix key counter ignoring visibility setting on HUD visibility change

### LINKED ISSUE #7705: Disabled key overlay always shows after re-enabling in-game interface
https://streamable.com/xsv6o

**osu!lazer version:** 
2020.130.0

frenzibyte:
Closes #7705 

- The underlying issue behind this is that the `HUDOverlay` changes visibility of `KeyCounterDisplay` based on its visibility value, which causes the key counter to be displayed regardless of the configuration, the solution I picked is whenever the key counter visibility configuration has changed, all key counters inside `KeyCounterDisplay` will be affected by the value instead of the container itself, which means displaying the `KeyCounterDisplay` will not display its internals unless they're visible by the configuration.


osu.Game/Screens/Play/KeyCounterDisplay.cs
```diff
@@ -100,7 +101,9 @@ public Color4 KeyUpTextColor
             }
         }
 
-        private void updateVisibility() => this.FadeTo(Visible.Value || configVisibility.Value ? 1 : 0, duration);
+        private void updateVisibility() =>
+            // Change visibility of all key counters internally to isolate from showing them by fading in this container.
+            Children.ForEach(k => k.FadeTo(Visible.Value || ConfigVisibility.Value ? 1 : 0, duration));
```
peppy: make a container rather than applying this to individual children.

osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
```diff
@@ -68,14 +75,58 @@ public void TestExternalHideDoesntAffectConfig()
...
 
-        private void createNew(Action<HUDOverlay> action = null)
+        [Test]
+        public void TestChangeHUDVisibilityOnHiddenKeyCounter()
+        {
+            createNew();
+
+            AddStep("set keycounter visible false", () =>
+            {
+                hudOverlay.KeyCounter.ConfigVisibility.Value = false;
```
peppy: set the config, not the bindable. make the bindable private again.

osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
```diff
@@ -68,14 +75,58 @@ public void TestExternalHideDoesntAffectConfig()
...
+
+            public TestHUDOverlay()
+                : base(null, null, null, Array.Empty<Mod>())
+            {
+                // Add any key just to display the key counter visually.
+                KeyCounter.Add(new KeyCounterKeyboard(Key.Space));
+            }
+        }
+
+        private class TestKeyCounterDisplay : KeyCounterDisplay
```
peppy: unnecessary

osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
```diff
@@ -68,14 +75,58 @@ public void TestExternalHideDoesntAffectConfig()
...
-                Child = hudOverlay = new HUDOverlay(null, null, null, Array.Empty<Mod>());
+                Child = hudOverlay = new TestHUDOverlay();
 
                 action?.Invoke(hudOverlay);
             });
         }
+
+        private class TestHUDOverlay : HUDOverlay
+        {
+            public new TestKeyCounterDisplay KeyCounter => (TestKeyCounterDisplay)base.KeyCounter;
```
peppy: unnecessary

osu.Game/Screens/Play/KeyCounterDisplay.cs
```diff
@@ -42,15 +43,15 @@ public override void Add(KeyCounter key)
...
         {
-            config.BindWith(OsuSetting.KeyOverlay, configVisibility);
+            config.BindWith(OsuSetting.KeyOverlay, ConfigVisibility);
         }
 
         protected override void LoadComplete()
         {
             base.LoadComplete();
 
             Visible.BindValueChanged(_ => updateVisibility());
```
peppy: i'd rename this to `AlwaysVisible` and add xmldoc explaining what it does.

peppy (dismissed):

osu.Game/Screens/Play/KeyCounterDisplay.cs
bdach: Also I'm really confused by the logic - isn't
```csharp
private void updateVisibility() => this.FadeTo(Visible.Value && configVisibility.Value ? 1 : 0, duration);
```
what we really want? How's this change fixing the problem?

iiSaLMaN: @peppy I was unsure at first, that way would cause a huge change to the `KeyCounterDisplay` (inherit `CompositeDrawable` and make the fill flow private), but if you think it's better that way, will do.

peppy: yes, this is how it should be done. it's only a few lines of change.

osu.Game/Screens/Play/KeyCounterDisplay.cs
bdach: You've renamed the other one that's bound to the setting, but I'd also rename `Visible` to `HasReplayLoaded` because that's the only actual case when this is bound to:

https://github.com/ppy/osu/blob/aa1daa0ad50205ea96da37b358e71820c04e41da/osu.Game/Screens/Play/Player.cs#L222

Removes all doubts about what is happening here.

peppy: i was actually asking for `Visible` to be renamed to `AlwaysVisible` but hey.

iiSaLMaN: I think @peppy's way is better here since the key counter needs to be visible when constructed, which will look weird with current names.
 - `alwaysShow = false (default by config)`. 
 - `HasReplayLoaded = false (default by DrawableRuleset)`.

@peppy's way:
 - `configVisibility - false (default by config)`.
 - `AlwaysVisible - true (should commonly be true)`.

bdach: I don't personally care that much, either's fine at this point compared to the previous version.

osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
bdach: I'm not sure this and the other comments are actually addressed? Can avoid exposing with `ChildrenOfType`.

smoogipoo (approved):
seems fine

PR #7714, opened 2020-02-03
Make slider tracking match what is on screen

peppy:
As discussed [here](https://github.com/ppy/osu/pull/7704#issuecomment-581211367), for the time being we want what happens on screen to match input 1:1 on lazer. This means, for the time being we will not be matching legacy behaviour in cases like this for the time being.

- Reverts a portion of #7704, but keeps the new tests and slider ball scale fix (2.4 instead of 2.0)


smoogipoo (approved):

TheWildTree:
I originally wanted to do it like this in the original PR, but gameplay with the 2.4-scale followcircle felt weird, more cluttered I guess.

How about a small change to the default followcircle, maybe make it slightly more transparent to increase readibility while still displaying it in the proper scale?

smoogipoo:
please start a separate issue about that

PR #7719, opened 2020-02-03
Update BeatmapOverlay header in line with the web design

EVAST9919:
Screenshot_1

I tried to make changes as little as possible.

### DIFF
--- a/osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
+++ b/osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
@@ -18,8 +18,6 @@ namespace osu.Game.Overlays.BeatmapSet
     {
         public readonly Bindable<BeatmapSetInfo> BeatmapSet = new Bindable<BeatmapSetInfo>();
 
-        public override bool PropagatePositionalInputSubTree => Enabled.Value && !Active.Value && base.PropagatePositionalInputSubTree;
-
         [Resolved]
         private OverlayColourProvider colourProvider { get; set; }
 
@@ -72,17 +70,6 @@ namespace osu.Game.Overlays.BeatmapSet
                 countContainer.FadeTo(beatmapsCount > 0 ? 1 : 0);
 
                 Enabled.Value = beatmapsCount > 0;
-            }, true);
-
-            Enabled.BindValueChanged(enabled =>
-            {
-                if (enabled.NewValue)
-                {
-                    UpdateState();
-                    return;
-                }
-
-                AccentColour = colourProvider.Foreground1;
             }, true);
         }
     }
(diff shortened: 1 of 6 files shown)


osu.Game/Overlays/OverlayRulesetTabItem.cs
```diff
@@ -64,29 +64,33 @@ public OverlayRulesetTabItem(RulesetInfo value)
...
-        protected override void OnDeactivated() => updateState();
+        protected override void OnDeactivated() => UpdateState();
 
-        private void updateState()
+        protected void UpdateState()
         {
             text.Font = text.Font.With(weight: Active.Value ? FontWeight.Bold : FontWeight.Medium);
+
+            if (!Enabled.Value)
+                return;
```
bdach: I don't understand what's going on with `UpdateState()` - why is it exposed as protected and why does it have this guard in the middle?

osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
```diff
@@ -3,143 +3,87 @@
...
+                    UpdateState();
+                    return;
+                }
 
-        protected override bool OnHover(HoverEvent e)
-        {
-            updateState();
-            return false;
+                AccentColour = colourProvider.Foreground1;
+            }, true);
```
bdach: I'm also really confused by this part too. I can't tell what this is trying to achieve or why `AccentColour` is set here, and I don't think I should be digging in this code to find out.

osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
```diff
@@ -3,143 +3,87 @@
...
+        {
+            base.LoadComplete();
 
             BeatmapSet.BindValueChanged(setInfo =>
             {
                 var beatmapsCount = setInfo.NewValue?.Beatmaps.Count(b => b.Ruleset.Equals(Value)) ?? 0;
 
                 count.Text = beatmapsCount.ToString();
-                count.Alpha = beatmapsCount > 0 ? 1f : 0f;
+                countContainer.FadeTo(beatmapsCount > 0 ? 1 : 0);
```
bdach: Kind of a strange change, as there doesn't seem to be any functional difference between `FadeTo` without explicit `duration` and directly setting `Alpha`.

bdach (dismissed):

osu.Game/Overlays/BeatmapSet/BeatmapRulesetTabItem.cs
EVAST9919: Just easier to read imo, nothing special

osu.Game/Overlays/OverlayRulesetTabItem.cs
EVAST9919: Item deselecting happens after disabling, so we don't want to update the colour here. Tricky stuff, but my best thoughts so far, will revisit it

EVAST9919: Okay, should be fixed

smoogipoo (approved):
lgtm 👍

PR #7721, opened 2020-02-03
Add {ScoreInfo,UserStatistics}.Accuracy

### LINKED ISSUE #7718: Accuracy formatting is inconsistent
**Describe the bug:** `ScoreInfo.Accuracy` is formatted in different inconsistent ways throughout osu!lazer.

In some places, it is formatted using `$"{value:0.00%}"`, for example: https://github.com/ppy/osu/blob/0759752bb40abce02a0ca228927b3fcfce24cdad/osu.Game/Overlays/Profile/Sections/Ranks/DrawableProfileScore.cs#L185

In some other places, `$"{value:P2}"` is used: https://github.com/ppy/osu/blob/0759752bb40abce02a0ca228927b3fcfce24cdad/osu.Game/Screens/Ranking/Results.cs#L214

This line does it correctly, in line with stable and web, though I would use `P2` instead of `0.00%`: https://github.com/ppy/osu/blob/0759752bb40abce02a0ca228927b3fcfce24cdad/osu.Game/Screens/Select/LocalScoreDeleteDialog.cs#L35

I think it is best if we use the solution above, but pasting `string.Format(...)` anywhere we need to format accuracy seems repetitive. Should we:
- create an extension method on type `double` named `FormatAccuracy`? This makes it rather awkward as it would stutter:
```cs
score.Accuracy.FormatAccuracy(); // repeats "Accuracy"
```
- extend type `double` to `Accuracy` and either add a `Format` or `ToString` method?

**osu!lazer version:** master branch commit 0759752bb40abce02a0ca228927b3fcfce24cdad

recapitalverb:
Closes #7718

### DIFF
--- a/osu.Game/Scoring/ScoreInfo.cs
+++ b/osu.Game/Scoring/ScoreInfo.cs
@@ -32,7 +32,7 @@ namespace osu.Game.Scoring
         public double Accuracy { get; set; }
 
         [JsonIgnore]
-        public string DisplayAccuracy => Accuracy == 1 ? "100%" : $"{Accuracy:P2}";
+        public string DisplayAccuracy => Accuracy == 1 ? "100%" : $"{Accuracy:0.00%}";
 
         [JsonProperty(@"pp")]
         public double? PP { get; set; }
--- a/osu.Game/Scoring/ScoreInfo.cs
+++ b/osu.Game/Scoring/ScoreInfo.cs
@@ -31,6 +31,9 @@ namespace osu.Game.Scoring
         [Column(TypeName = "DECIMAL(1,4)")]
         public double Accuracy { get; set; }
 
+        [JsonIgnore]
+        public string DisplayAccuracy => Accuracy == 1 ? "100%" : $"{Accuracy:P2}";
+
         [JsonProperty(@"pp")]
         public double? PP { get; set; }
 
--- a/osu.Game/Scoring/ScoreInfo.cs
+++ b/osu.Game/Scoring/ScoreInfo.cs
@@ -13,6 +13,7 @@ using osu.Game.Rulesets;
 using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Users;
+using osu.Game.Utils;
 
 namespace osu.Game.Scoring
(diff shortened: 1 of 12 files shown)


osu.Game/Scoring/ScoreInfo.cs
```diff
@@ -31,6 +31,9 @@ public class ScoreInfo : IHasFiles<ScoreFileInfo>, IHasPrimaryKey, ISoftDelete,
         [Column(TypeName = "DECIMAL(1,4)")]
         public double Accuracy { get; set; }
 
+        [JsonIgnore]
+        public string DisplayAccuracy => Accuracy == 1 ? "100%" : $"{Accuracy:P2}";
```
bdach: `:P2` will actually insert a space before the percent sign in the invariant culture (so i.e. `62.94 %`), which is super strange IMO, and at least inconsistent with the special check for `1`. Therefore I'd say it'd be better to explicitly get rid of that behaviour:
```suggestion
        public string DisplayAccuracy => Accuracy == 1 ? "100%" : $"{Accuracy:0.00%}";
```

osu.Game/Users/UserStatistics.cs
```diff
@@ -43,6 +43,9 @@ private int rank
         [JsonProperty(@"hit_accuracy")]
         public decimal Accuracy;
 
+        [JsonIgnore]
+        public string DisplayAccuracy => $"{Accuracy:0.00%}";
```
peppy: This should probably match the same implementation as the other one. At which point it may be best to call an extension method for formatting in both cases (but leave the `DisplayAccuracy` properties for accessibility).

recapitalverb: Should I put the extension method under `osu.Game.Extensions`?

smoogipoo: `osu.Game.Utils`

recapitalverb:
> was kinda weirded out by the decimal case, but it seems to look fine

`UserStatistics.Accuracy` ranges from 0 to 100 instead of 0 to 1 for *some* reason, so I was weirded out too. Ended up with duplicated code and comments.

PR #7725, opened 2020-02-04
Adjust TopScoreStatisticsSection to closer match web design

recapitalverb:
Screenshot_20200204_195805


recapitalverb:
One thing about the order of the HitResults columns.

https://github.com/recapitalverb/osu/blob/5030b66092ff9e28c006c5ee025bf5881c65997d/osu.Game/Overlays/BeatmapSet/Scores/TopScoreStatisticsSection.cs#L101

This is very clumsy and results in unsorted columns:
Screenshot_20200204_200610

I could only come up with an ugly-looking solution:
```cs
var sortedHitResults = value.Statistics.Keys.ToList();
sortedHitResults.Sort();
sortedHitResults.Reverse();
statisticsColumns.ChildrenEnumerable = sortedHitResults.Select(key => createStatisticsColumn(key, value.Statistics[key]));
```
Any suggestions?

TheWildTree:
just a heads up, separator should use `Background3` colour and be 2px tall to exactly match osu-web

https://github.com/ppy/osu/blob/5030b66092ff9e28c006c5ee025bf5881c65997d/osu.Game/Overlays/BeatmapSet/Scores/TopScoreStatisticsSection.cs#L132-L137

https://github.com/ppy/osu/blob/5030b66092ff9e28c006c5ee025bf5881c65997d/osu.Game/Overlays/BeatmapSet/Scores/TopScoreStatisticsSection.cs#L143-L147

recapitalverb:
With the colours suggested it definitely looks a lot better, thanks @TheWildTree !
I also added some padding to the right of `DrawableTopScore`. I think this is close enough - only thing left to do is to sort the columns from `HitResults`.
Screenshot_20200204_215816

TheWildTree:
noticed a few more differences!

- spacing is not even between top and bottom side of the separator:
image
web:
image

- separator width should match text:
image
web:
image

- separator height is 2px on web, but on lazer it looks off to me, it actually looks closer with 1.5 height (sorry for not checking this earlier) - though it might be due to some UI scaling issues, because sometimes 2 looks the same as 1.5, it's kinda random:

    image image image
    1.5 vs. 2 in worst case scenario vs. web

recapitalverb:
I have already noticed the artifacts on the separator while scrolling, it seems like a framework-side issue as it also happens in other places in osu!lazer. Also 2px is fine imo, I agree it's just scaling issues.

bdach:
I fixed the wrong ordering by doing an `.OrderByDescending`. You were close, but probably didn't find that linq method.

By the way, your first commit in this PR is incorrect. You shouldn't merge changes from upstream to your fork by opening PRs back; see [*Syncing a fork*](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) for instructions. I attempted to fix that up by rebasing but screwed up somehow since github picked up that I committed the rebased commits somehow, so that's why the double force-push, sorry :(

recapitalverb:
I see, thanks a lot!
By the way, I tried fixing the incorrect separator width issue but that quickly turned into a mess; I think it is best to implement something like `Container.MinimumWidth` framework-side, which should make a lot of things, including implementing `InfoColumn` correctly, a lot easier.

bdach:
`GridContainer` is the closest thing to minimum width support framework-side.

PR #7726, opened 2020-02-04
Add API to get rankings data for selected spotlight

EVAST9919:


osu.Game.Tests/Visual/Online/TestSceneRankingsTables.cs
```diff
@@ -68,6 +68,7 @@ protected override void LoadComplete()
             AddStep("Mania scores", () => createScoreTable(new ManiaRuleset().RulesetInfo));
             AddStep("Taiko country scores", () => createCountryTable(new TaikoRuleset().RulesetInfo));
             AddStep("Catch US performance page 10", () => createPerformanceTable(new CatchRuleset().RulesetInfo, "US", 10));
+            AddStep("Osu 271 spotlight table", () => createSpotlightTable(new OsuRuleset().RulesetInfo, 271));
```
bdach: Could do with a more descriptive name.

osu.Game/Users/UserStatistics.cs
```diff
@@ -71,21 +71,21 @@ private int rank
...
+            public int? SPlus;
 
             [JsonProperty(@"s")]
             public int S;
 
             [JsonProperty(@"a")]
             public int A;
 
-            public int this[ScoreRank rank]
+            public int? this[ScoreRank rank]
```
bdach: I'm unsure if doing this makes sense if it's going to have to get fixed up to 0 in places anyway (as it is being here in `UserBasedTable`).

EVAST9919: hmm, yeah, we can just make a null check in return

smoogipoo (approved):
this is fine for now, spotlights are... temporary? as far as I'm hearing

PR #7728, opened 2020-02-04
Allow guest users to see the comments in CommentsContainer

EVAST9919:


bdach (dismissed):
This can't be even directly tested right now. The fetch itself could be (by opening the game, logging out, and opening the test browser), but the refetch on logout cannot.

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -113,6 +113,7 @@ private void load(OverlayColourProvider colourProvider)
 
         protected override void LoadComplete()
         {
+            api.LocalUser.BindValueChanged(_ => refetchComments());
```
iiSaLMaN: if I'm not missing something here, that's a memory leak?

frenzibyte (changes requested):

EVAST9919:
@bdach I can just a test step with `LocalUser.TriggerChange()`, but not sure about that

osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
```diff
@@ -37,23 +39,29 @@ public class TestSceneCommentsContainer : OsuTestScene
...
             {
                 scroll.Clear();
-                scroll.Add(comments = new CommentsContainer());
+                scroll.Add(comments = new TestCommentsContainer());
             });
         }
+
+        private class TestCommentsContainer : CommentsContainer
+        {
+            public new Bindable<User> User => base.User;
```
iiSaLMaN: no need to expose a bindable bound to a source you can easily retrieve.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestSceneCommentsContainer.cs
smoogipoo: i'm ok with this for now

smoogipoo (approved):

PR #7724, opened 2020-02-04
Update BeatmapSetOverlay to match web design

EVAST9919:
Screenshot_1

PR #7730, opened 2020-02-04
Adjust minor BeatmapSetOverlay details to better match osu-web

thewildtree:
Follows #7724, aims to bring the `BeatmapSetOverlay` even closer to its osu-web counterpart. These changes are mostly small details, but they definitely can affect the UX.

A few things worth noting:
- the scoreboard font size change from mentioned PR has been reversed - while 12 technically matches osu-web, scaling in lazer is a bit different, making 14 bring a more familiar look here
- scoreboard had wrong order of columns just like `TopScoreStatisticsSection` (as mentioned in #7725), I applied a similar solution, but I'm not sure if it's the proper one here. maybe instead of this we should sort them before they get passed to any UI components?
- added some checks to hide ratings / show placeholder instead of success rate if beatmap has no leaderboard, but I'm not sure if its done the right way
- wanted to add the green colour to the combo if it was the max possible, same as web has, but currently I believe there's no way to check beatmaps maxcombo in this scope

image

image


bdach:
Please use `.\InspectCode.ps1` to find stylistic issues instead of relying on CI to spot them for you.

TheWildTree:
Sorry, I couldn't run InspectCode because of some issues with PowerShell on my PC. Will start doing that as soon as I solve them. Also sorry for the CI-fix commit spam, I'm no longer at my PC and github web edit only allows one file per commit.

TheWildTree:
Fixed few more details I found, I'm gonna leave the `AdvancedStats` bars background like it is even though it doesn't match osu-web since that part is reused elsewhere and that's a very small detail anyway.

Other than that, no more changes planned.

osu.Game/Utils/FormatUtils.cs
```diff
@@ -7,18 +7,18 @@ public static class FormatUtils
...
-        /// Omits all decimal places when <paramref name="accuracy"/> equals 1d.
         /// </summary>
         /// <param name="accuracy">The accuracy to be formatted</param>
+        /// <param name="alwaysShowDecimals">Whether to show decimal places if <paramref name="accuracy"/> equals 1d</param>
         /// <returns>formatted accuracy in percentage</returns>
-        public static string FormatAccuracy(this double accuracy) => accuracy == 1 ? "100%" : $"{accuracy:0.00%}";
+        public static string FormatAccuracy(this double accuracy, bool alwaysShowDecimals = false) => accuracy == 1 && !alwaysShowDecimals ? "100%" : $"{accuracy:0.00%}";
 
         /// <summary>
         /// Turns the provided accuracy into a percentage with 2 decimal places.
```
bdach: Kind of no longer true. I'm overall not too happy about this `alwaysShowDecimals` flag. Why `0.00%` but not `100.00%`? At this point if we're centralising we should centralise, potential discrepancies with web be damned.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -65,6 +66,10 @@ public IReadOnlyList<ScoreInfo> Scores
                 for (int i = 0; i < value.Count; i++)
                     backgroundFlow.Add(new ScoreTableRowBackground(i, value[i]));
 
+                // Ensure correct column order
+                foreach (ScoreInfo score in value)
+                    score.Statistics = score.Statistics.OrderByDescending(pair => pair.Key).ToDictionary(pair => pair.Key, pair => pair.Value);
```
bdach: Shouldn't be mutating `ScoreInfo` IMO. Would be best if `ScoreInfo` exposed a read-only ordered view of rank-count pairs.

osu.Game/Utils/FormatUtils.cs
TheWildTree: If I understand correctly, this wasn't true beforehand either - `100%` never had decimals shown before this change, so, if anything, it brings that closer to truth if the flag is set to `true`.

> Why `0.00%` but not `100.00%`? 

Good question to ask the original author of this method.

bdach: Honestly feels like wasting time going back and forth on significant digits. Should pick an opinionated choice and stick with it.

TheWildTree: Agreed, I also think we should just stick to 2 decimal places everywhere, as it'ss the most consistent option.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
TheWildTree: That would indeed be a good way to do it. 

I think I'll split this as well as the `FormatAccuracy` changes into separate PRs, since their scope is beyond just the beatmapset overlay.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -116,7 +121,7 @@ private Drawable[] createContent(int index, ScoreInfo score)
                 new OsuSpriteText
                 {
                     Margin = new MarginPadding { Right = horizontal_inset },
-                    Text = score.DisplayAccuracy,
+                    Text = score.Accuracy.FormatAccuracy(alwaysShowDecimals: true),
```
peppy: this shouldn't be happening. we added the `DisplayAccuracy` string for a reason.

peppy: as mentioned elsewhere, please split these changes into a separate PR for further discussion. it's not clear why this is being done in this PR currently.

TheWildTree:
Removed mentioned parts, will create separate PRs for them later.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -190,7 +190,7 @@ private class HeaderText : OsuSpriteText
             public HeaderText(string text)
             {
                 Text = text.ToUpper();
-                Font = OsuFont.GetFont(size: 10, weight: FontWeight.Bold);
+                Font = OsuFont.GetFont(size: 12, weight: FontWeight.Bold);
```
smoogipoo: osu-web uses size 10 here.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -23,7 +23,7 @@ public class ScoreTable : TableContainer
     {
         private const float horizontal_inset = 20;
         private const float row_height = 25;
-        private const int text_size = 12;
+        private const int text_size = 14;
```
smoogipoo: osu-web uses size 12 here

osu.Game/Overlays/BeatmapSet/Info.cs
```diff
@@ -110,6 +112,14 @@ public Info()
                                     RelativeSizeAxes = Axes.Both,
                                     Padding = new MarginPadding { Top = 20, Horizontal = 15 },
                                 },
+                                unrankedPlaceholder = new OsuSpriteText
+                                {
+                                    Anchor = Anchor.Centre,
+                                    Origin = Anchor.Centre,
+                                    Alpha = 0,
+                                    Text = "Unranked beatmap",
+                                    Font = OsuFont.GetFont(size: 13)
```
smoogipoo: osu-web uses size 12 here

osu.Game/Overlays/BeatmapSet/Scores/TopScoreUserSection.cs
```diff
@@ -96,13 +96,14 @@ public TopScoreUserSection()
                             {
                                 Anchor = Anchor.CentreLeft,
                                 Origin = Anchor.CentreLeft,
-                                Font = OsuFont.GetFont(size: 10, weight: FontWeight.Bold)
+                                Font = OsuFont.GetFont(size: 12)
```
smoogipoo: osu-web uses size 10 here

smoogipoo (changes requested):
In general, always match osu-web sizings

TheWildTree:
size 10 is unreadable due to the font rendering issues

in general, if you take a look at the rest of the overlay, most elements use fonts a few px bigger than web and I followed that direction - it seems correct because matching osu-web 1:1 here looks way more off (I suppose due to some scaling behaviour I can't properly explain) and font sizes like 10px are unreadable due to how lazer renders them in overlays.

smoogipoo:
Match the font sizes. We can uniformly apply a scale at a later point if needed, inside `OsuFont` itself.

TheWildTree:
Then we'd have to change every single overlay element / font since they're mostly sized a tad bigger than osu-web.

smoogipoo:
That's correct.

TheWildTree:
Alright, I'll try to work on it, I'm not sure how that's gonna end up looking though.

peppy:
It will look fine. All metrics should match (font sizes, widths, paddings, margins). Overall adjustments can be applied later on when we decide how large or small we want things to display.

The important part is that everything matches on a numeric and relative basis so it is easy to confirm correctness without manually pixel checking (aka what you've probably been doing until now).

TheWildTree:
Went back to osu-web font size values. I'll make a separate PR soon which will aim to bring all the dimensions / sizes 1:1 with osu-web.

bdach:
I fixed up a few remaining nits; looks fine otherwise.

osu.Game/Overlays/BeatmapSet/Header.cs
```diff
@@ -134,7 +134,7 @@ public Header()
                                                 {
                                                     title = new OsuSpriteText
                                                     {
-                                                        Font = OsuFont.GetFont(size: 37, weight: FontWeight.SemiBold, italics: true)
+                                                        Font = OsuFont.GetFont(size: 37.5f, weight: FontWeight.SemiBold, italics: true)
```
TheWildTree: If the artist text uses osu-web font size, I think this should too

bdach: Good point, I got confused and looked at web's `line-height` instead of `font-size`. Will fix up in a sec.

bdach (approved):
Think this PR has run its course - I'd say let's get it in for now. Accuracy with web can be iterated on.

smoogipoo (approved):

PR #7731, opened 2020-02-05
Added Spun Out mod

tk421bsod:


peppy:
Please stop doing what you are doing

TK421bsod:
My bad. I'll leave it open.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
-    public class OsuModSpunOut : Mod
+    public class OsuModSpunOut : Mod, IUpdatableByPlayfield, IApplicableToDrawableRuleset<OsuHitObject>
     {
         public override string Name => "Spun Out";
         public override string Acronym => "SO";
-        public override IconUsage? Icon => OsuIcon.ModSpunout;
-        public override ModType Type => ModType.DifficultyReduction;
-        public override string Description => @"Spinners will be automatically completed.";
+        public override IconUsage? Icon => OsuIcon.ModSpunOut;
+        public override ModType Type => ModType.DifficultyIncrease;
```
bdach: I'm pretty sure Spun Out is not supposed to be a difficulty increase mod.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+        public override string Description => @"Spinners will be automatically completed";
+        public override Type[] IncompatibleMods => base.IncompatibleMods.Append(typeof(OsuModAutopilot)).ToArray();
         public override double ScoreMultiplier => 0.9;
-        public override bool Ranked => true;
-        public override Type[] IncompatibleMods => new[] { typeof(ModAutoplay), typeof(OsuModAutopilot) };
```
bdach: As far as I can tell `ModAutopilot` was erroneously dropped here as well.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+        private List<OsuReplayFrame> replayFrames;
+        private int currentFrame;
+        public void Update(Playfield playfield)
+        {
+            bool requiresHold = false;
+            inputManager.AllowUserCursorMovement = true;
+            foreach (var drawable in playfield.HitObjectContainer.AliveObjects)
+            {
+                inputManager.AllowUserCursorMovement = true;
```
bdach: These repeated assignments of the same values to the same variable are not necessary.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+                double time = playfield.Time.Current;
+                if (!(drawable is DrawableOsuHitObject osuHit))
+                {
+                    inputManager.AllowUserCursorMovement = true;
+                    return;
+                }
+                if (osuHit is DrawableSpinner)
+                {
+                    inputManager.AllowUserCursorMovement = false;
+                    requiresHold |= (osuHit is DrawableSpinner);
```
bdach: Repeated condition check.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+                    if (currentFrame == replayFrames.Count - 1)
+                    {
+                        return;
+                    }
+                    else if (Math.Abs(replayFrames[currentFrame + 1].Time - time) <= Math.Abs(replayFrames[currentFrame].Time - time))
```
bdach: I'm not sure I'm following this check, mind explaining? In fact I think this could have a bug in it.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+                return;
+            wasHit = hitting;
+            var state = new ReplayState<OsuAction>
+            {
+                PressedActions = new List<OsuAction>()
+            };
+            if (hitting)
+            {
+                state.PressedActions.Add(wasLeft ? OsuAction.LeftButton : OsuAction.RightButton);
+                wasLeft = !wasLeft;
```
bdach: What's the purpose of this? Does this mean that the held key will alternate every replay frame?

bdach (changes requested):

mcendu:
Hijacking the input manager looks like a hack to me.

smoogipoo:
hijacking like done here is the proper way to do it

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,86 @@
...
+        private int currentFrame;
+        public void Update(Playfield playfield)
+        {
+            bool requiresHold = false;
+            inputManager.AllowUserCursorMovement = true;
+            foreach (var drawable in playfield.HitObjectContainer.AliveObjects)
+            {
+                inputManager.AllowUserCursorMovement = true;
+                double time = playfield.Time.Current;
+                if (!(drawable is DrawableOsuHitObject osuHit))
```
mcendu: I suppose we will never reach such a situation?

bdach:
After checking out this PR to see if it's salvageable, it is unfortunately my opinion that fixing this current implementation would require almost as much work as starting from scratch. To recap:

* Code style is not adhered to. Errors indicated by CI have not been addressed.
* The mod has the wrong icon (see mod icons on web for the proper one), is in the wrong section if aiming for parity with stable (should be in `Automation`) and still has the wrong mod incompatibility settings.
* During gameplay the mouse cursor (I don't mean the manager, but the actual mouse cursor) is hijacked directly and crudely. While the spinner is being auto-spun, the user has no perception of where on the screen they're currently pointing. Moreover, when control is handed back to the user, it is handed back to the current absolute position of the mouse cursor, causing disorienting jumps.

  Compare this behaviour with stable, where the user can visually see and track the mouse cursor the whole way through, while the spinner auto-spins *on its own*.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
mcendu: Please resolve this

peppy:
Closing in favour of #7764.

PR #7732, opened 2020-02-05
Make EditorBeatmap a drawable component

peppy:
Allows for scheduling and accessing DI components with ease. Also moves `UpdateHitObjects` here so it can be used more readily by other components (like the timeline).


bdach (approved):
seems to work, code looks better, so tentative approve here 👍

smoogipoo (approved):

PR #7733, opened 2020-02-05
Add spotlight selector to RankingsOverlay

EVAST9919:
Screenshot_3


osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -36,50 +39,60 @@ public IEnumerable<APISpotlight> Spotlights
...
-                            Children = new Drawable[]
+                            dropdown = new SpotlightsDropdown
                             {
-                                startDateColumn = new InfoColumn(@"Start Date"),
-                                endDateColumn = new InfoColumn(@"End Date"),
+                                Anchor = Anchor.TopCentre,
+                                Origin = Anchor.TopCentre,
+                                RelativeSizeAxes = Axes.X,
+                                Current = Current,
+                                Depth = -float.MaxValue
```
bdach: I feel like this shouldn't be explicitly set for the dropdown. Dropdowns should by their nature have the dropdown content at the forefront of the screen always (can't think of any reasonable exceptions). Which is not to say this code is *wrong*, I just don't want this popping up with every dropdown usage - that could just mean the current default is not chosen correctly.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -88,17 +101,24 @@ private void load(OverlayColourProvider colourProvider)
             background.Colour = colourProvider.Dark3;
         }
 
-        protected override void LoadComplete()
+        public void ShowInfo(APISpotlight spotlight, int mapCount)
```
bdach: This might as well have accepted the entire `GetSpotlightRankingsResponse` and saved a method argument.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -184,26 +215,51 @@ private Drawable createTableFromResponse(APIRequest request)
...
+                                RelativeSizeAxes = Axes.X,
+                                Spacing = new Vector2(10),
+                                Children = spotlightRequest.Result.BeatmapSets.Select(b => new DirectGridPanel(b.ToBeatmapSet(rulesets))
+                                {
+                                    Anchor = Anchor.TopCentre,
+                                    Origin = Anchor.TopCentre,
+                                }).ToList()
+                            }
+                        }
+                    };
```
bdach: Kinda feel this should be a method at least, if not a component.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -22,18 +26,24 @@ public class RankingsOverlay : FullscreenOverlay
...
         private readonly BasicScrollContainer scrollFlow;
-        private readonly Container tableContainer;
+        private readonly Container contentContainer;
         private readonly DimmedLoadingLayer loading;
         private readonly Box background;
+        private readonly RankingsOverlayHeader header;
 
         private APIRequest lastRequest;
         private CancellationTokenSource cancellationToken;
+        private GetSpotlightsRequest spotlightsRequest;
```
bdach: I'm not sure how I feel about this being all the way up here. I'd prefer it to be hidden somewhere below because now you have `lastRequest` and `spotlightsRequest` next to each other.

bdach (commented):
Code looks ok, albeit awkward in places. I feel there's some improvement to be gained here still.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
EVAST9919: In most cases dropdown is a part of autosized drawable

EVAST9919: At this point I don't even remember, why I set it there. Will revisit and reply here

osu.Game/Overlays/Rankings/SpotlightSelector.cs
EVAST9919: Please, explain further, I don't really understand what are you trying to tell, sorry

osu.Game/Overlays/Rankings/SpotlightSelector.cs
bdach: I checked - it's because of the dropdown list getting covered by the neighbouring `FillFlowContainer` here when the dropdown is opened. The list should always have min depth when opened IMO.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
bdach: In the overlay you're doing:
```csharp
header.SpotlightSelector.ShowInfo(spotlightRequest.Result.Spotlight, spotlightRequest.Result.BeatmapSets.Count);
```
while you could just do:
```csharp
header.SpotlightSelector.ShowInfo(spotlightRequest.Result);
```
and make `ShowInfo` only accept a single `GetSpotlightRankingsResponse`.

EVAST9919: Now I got it, good catch

bdach (dismissed):
I think this can go in as-is for now. The depth issue should probably be covered in a separate PR.

smoogipoo:
I'm able to replicate a crash in `TestFlagScopeDependency`:

1. Allow all the steps to complete.
2. Click the "spotlights" tab.

Crashes with:

```
[runtime:error] 2020-02-07 03:20:06: An unhandled error has occurred.
[runtime:error] 2020-02-07 03:20:06: System.NullReferenceException: Object reference not set to an instance of an object.
[runtime:error] 2020-02-07 03:20:06: at osu.Game.Overlays.RankingsOverlay.createScopedRequest() in /Users/smoogipoo/Repos/osu/osu.Game/Overlays/RankingsOverlay.cs:line 195
[runtime:error] 2020-02-07 03:20:06: at osu.Game.Overlays.RankingsOverlay.loadNewContent() in /Users/smoogipoo/Repos/osu/osu.Game/Overlays/RankingsOverlay.cs:line 167
[runtime:error] 2020-02-07 03:20:06: at osu.Framework.Threading.ScheduledDelegate.RunTask()
[runtime:error] 2020-02-07 03:20:06: at osu.Framework.Threading.Scheduler.Update()
```

smoogipoo (changes requested):
As mentioned.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -100,20 +110,31 @@ protected override void LoadComplete()
                 if (Country.Value != null)
                     Scope.Value = RankingsScope.Performance;
 
-                Scheduler.AddOnce(loadNewContent);
+                if (Scope.Value != RankingsScope.Spotlights)
+                    Scheduler.AddOnce(loadNewContent);
```
bdach: I am *very* not fine with this way of fixing the problem. What the crash scenario demonstrated is that the overlay can be forced to fire off requests to fetch spotlight ranks while not being in a ready state. At the point of the nullref being thrown neither `header.Spotlight` nor `header.Spotlights` were initialised to sane values.

This check is essentialy monkey-patching this edge case. I had to spend a good 10 minutes to understand why this is making the crash go away, and it is due to:

* spotlights changing `Country.Value` to `null`,
* which fires this `Country.BindValueChanged` subscriber with `RankingsScope.Spotlights` and not-ready state of not having any spotlights at all.

I feel like letting this remain at this state is just asking for trouble further down the line. In my opinion the proper direction to go forward with this is to go back on the decision to add the spotlight selector to the header and make a separate component for spotlight rankings, which will have:

* the spotlight dropdown along with the extra content,
* the actual ranking table,
* and the beatmap covers in the footer.

This component would be solely responsible for firing its own request and making sure it's in a sane state to start filtering, therefore encapsulating the spotlight logic and removing clutter from the top-level overlay.

bdach (changes requested):

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: Refactored

smoogipoo (approved):

smoogipoo (approved):
Looks good from me. @bdach feel free to merge.

bdach (approved):
Looks good. `RankingsOverlay` needs more work, but it's a good start. In the end I think all of the tabs should be extracted to their own separate layouts.

PR #7734, opened 2020-02-05
Add the ability to extend hold notes (spinners / sliders etc.) via timeline

peppy:
Also improves the visuals of timeline blueprints.

- Some objects may not update correctly just yet in the playfield (visually)
- [x] Depends on https://github.com/ppy/osu/pull/7732

screencast 2020-02-05 19-45-23


osu.Game.Rulesets.Catch/Objects/JuiceStream.cs
```diff
@@ -110,7 +110,11 @@ protected override void CreateNestedHitObjects()
             }
         }
 
-        public double EndTime => StartTime + this.SpanCount() * Path.Distance / Velocity;
+        public double EndTime
+        {
+            get => StartTime + this.SpanCount() * Path.Distance / Velocity;
+            set => throw new System.NotImplementedException();
```
bdach: If I'm understanding this right, using this setter is directly incorrect - so I'd throw `NotSupportedException` or `InvalidOperationException` (`NotImplemented` suggests it'll be implemented at some point)

bdach (commented):
Works well, code looks pretty good too. There were a couple of places I thought things could be a little nicer (the throwing setters being probably the most notable one) but I think this is ok for a starting point 👍

osu.Game.Rulesets.Catch/Objects/JuiceStream.cs
smoogipoo: `NotSupportedException` is probably more adequate here.

peppy: I was actually mistaken; we can already support this. Will implement correctly.

peppy: Ah, I left it out for simplicity. So let's go with `NotSupportedException` for now.

smoogipoo (approved):

PR #7743, opened 2020-02-06
Make ScreenshotManager a Component

UselessToucan:
see https://github.com/ppy/osu-framework/pull/3248

### DIFF
--- a/osu.Game/Graphics/ScreenshotManager.cs
+++ b/osu.Game/Graphics/ScreenshotManager.cs
@@ -9,7 +9,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Audio;
 using osu.Framework.Audio.Sample;
 using osu.Framework.Bindables;
-using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics;
 using osu.Framework.Input;
 using osu.Framework.Input.Bindings;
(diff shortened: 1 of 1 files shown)


osu.Game/Graphics/ScreenshotManager.cs
```diff
@@ -22,7 +22,7 @@
 
 namespace osu.Game.Graphics
 {
-    public class ScreenshotManager : Container, IKeyBindingHandler<GlobalAction>, IHandleGlobalKeyboardInput
+    public class ScreenshotManager : Drawable, IKeyBindingHandler<GlobalAction>, IHandleGlobalKeyboardInput
```
iiSaLMaN: What you really want it to be is a `Component`:
```suggestion
    public class ScreenshotManager : Component, IKeyBindingHandler<GlobalAction>, IHandleGlobalKeyboardInput
```

frenzibyte (changes requested):

osu.Game/Graphics/ScreenshotManager.cs
UselessToucan: Yes, you are correct. I forgot about this class. :(

smoogipoo (approved):

frenzibyte (approved):

PR #7745, opened 2020-02-06
Fix lifetime calculation in overlapping scroll algorithm

### LINKED ISSUE #7436: osu!taiko finisher shows up too late
https://streamable.com/z45h5

**Beatmap:** 
https://osu.ppy.sh/beatmapsets/1003952#taiko/2101202 (last diff)

**osu!lazer version:** 
latest build on linux (self compiled)

bdach:
Resolves #7436.

# Summary

Changes to lifetime calculation in scrolling rulesets introduced in #7367, which aimed to account for the distance between hit objects' origin and its edge entering the scrolling area, fixed some issues with hit objects appearing abruptly, but also regressed some other scenarios. 

Upon investigation, the regression was localised to the overlapping scroll algorithm. The reason for this was two-fold:

* The previous code used `TimeAt()` to calculate the time of travel from the hit object's edge to its origin. For other algorithms, that time can be accurately reconstructed, because they don't have periods of time where there are multiple hit objects scrolling at different velocities.

  That invariant does not hold for the overlapping algorithm, therefore it is possible for different values to be technically correct for `TimeAt()`. However, the only value that matters for the adjustment is the one that's indicated by the control point that applies to the hit object origin, which can be uniquely identified.

* Additionally, the offset returned (even if correct) was applied externally to the hit object's start time and passed to `GetDisplayStartTime()`. In the overlapping algorithm, the choice of control point used in `GetDisplayStartTime()` is important, since the value of the speed multiplier is read within. Externally rewinding the hit object's start time meant that in some cases the speed multiplier of the *previous* control point is applied, which led to hit objects appearing too late if the scrolling rate decreased.

Because of the above, modify `GetDisplayStartTime()` to take the offset into account in all algorithms, and apply the adjustment correctly inside of them. The constant and sequential algorithms needed no
adjustment from the previous logic, since:

* the constant algorithm disregarded control points, and 
* the sequential algorithm would effectively rewind to `time = 0`, calculate the absolute distance from `time = 0` to the hit object start, apply the origin offset *to the absolute distance*, and then convert back to time, applying all control points in sequence. Due to this it was impossible for control points to get mixed up while calculating.

As for the overlapping algorithm, the high-level logic is as follows: 

* The distance that the origin has to travel is the length of the scroll plus the distance from the origin to the object edge.
* The above distance divided by the scroll length gives the relative scroll lengths that the object has to travel.
* As one relative scroll length takes one time range, the relative travel length multiplied by the time range gives the absolute travel time of the object origin.
* Finally, the control point multiplier applicable at origin time is applied to the whole travel time.

Correctness of the above is demonstrated by visual tests added before and headless unit tests of the algorithms themselves. The sequential scroll algorithm was not covered by unit tests, and remains uncovered due to floating-point inaccuracies that need to be addressed separately.

# Remarks

The visual test scene was changed to add steps demonstrating the regression, and additionally to better distinguish the hit objects by assigning them random colours.

I'm not happy with adding new parameters to `GetDisplayStartTime`, but I didn't want to expose control points externally as it defeats the entire point of the algorithms, and the original method without them had no meaningful users. The old behaviour can always be recovered by passing a zero adjustment and any non-zero scroll width - I can make that a default interface implementation if deemed necessary, but again - I'm unconvinced we're going to need it.

Here are some demonstration videos for the change:

* test scene ([before](https://streamable.com/jvr6g), [after](https://streamable.com/i31lj)),
* gameplay at the [map reported as affected](https://osu.ppy.sh/beatmapsets/1003952#taiko/2101202) ([before](https://streamable.com/8sz1b), [after](https://streamable.com/np0me)).

### DIFF
--- a/osu.Game/Rulesets/UI/Scrolling/Algorithms/IScrollAlgorithm.cs
+++ b/osu.Game/Rulesets/UI/Scrolling/Algorithms/IScrollAlgorithm.cs
@@ -6,15 +6,33 @@ namespace osu.Game.Rulesets.UI.Scrolling.Algorithms
     public interface IScrollAlgorithm
     {
         /// <summary>
-        /// Given a point in time, computes the time at which it enters the time range.
+        /// Given a point in time associated with an object's origin
+        /// and the spatial distance between the edge and the origin of the object along the scrolling axis,
+        /// computes the time at which the object initially enters the time range.
         /// </summary>
-        /// <remarks>
-        /// E.g. For a constant time range of 5000ms, the time at which t=7000ms enters the time range is 2000ms.
-        /// </remarks>
-        /// <param name="time">The point in time.</param>
+        /// <example>
+        /// Let's assume the following parameters:
+        /// <list type="bullet">
+        ///     <item><paramref name="originTime"/> = 7000ms,</item>
+        ///     <item><paramref name="offset"/> = 100px,</item>
+        ///     <item><paramref name="timeRange"/> = 5000ms,</item>
+        ///     <item><paramref name="scrollLength"/> = 1000px</item>
+        /// </list>
+        /// and a constant scrolling rate.
+        /// To arrive at the end of the scrolling container, the object's origin has to cover
+        /// <code>1000 + 100 = 1100px</code>
+        /// so that the edge starts at the end of the scrolling container.
+        /// One scroll length of 1000px covers 5000ms of time, so the time required to cover 1100px is equal to
+        /// <code>5000 * (1100 / 1000) = 5500ms,</code>
+        /// and therefore the object should start being visible at
+        /// <code>7000 - 5500 = 1500ms.</code>
+        /// </example>
+        /// <param name="originTime">The time point at which the object origin should enter the time range.</param>
+        /// <param name="offset">The spatial distance between the object's edge and its origin along the scrolling axis.</param>
         /// <param name="timeRange">The amount of visible time.</param>
-        /// <returns>The time at which <paramref name="time"/> enters <paramref name="timeRange"/>.</returns>
-        double GetDisplayStartTime(double time, double timeRange);
+        /// <param name="scrollLength">The absolute spatial length through <paramref name="timeRange"/>.</param>
+        /// <returns>The time at which the object should enter the time range.</returns>
+        double GetDisplayStartTime(double originTime, float offset, double timeRange, float scrollLength);
 
         /// <summary>
         /// Computes the spatial length within a start and end time.
--- a/osu.Game.Tests/ScrollAlgorithms/ConstantScrollTest.cs
+++ b/osu.Game.Tests/ScrollAlgorithms/ConstantScrollTest.cs
@@ -18,12 +18,21 @@ namespace osu.Game.Tests.ScrollAlgorithms
         }
 
         [Test]
-        public void TestDisplayStartTime()
+        public void TestPointDisplayStartTime()
         {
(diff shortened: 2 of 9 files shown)


osu.Game/Rulesets/UI/Scrolling/Algorithms/IScrollAlgorithm.cs
```diff
@@ -6,15 +6,33 @@ namespace osu.Game.Rulesets.UI.Scrolling.Algorithms
...
+        ///     <item><paramref name="scrollLength"/> = 1000px</item>
+        /// </list>
+        /// and a constant scrolling rate.
+        /// To arrive at the end of the scrolling container, the object's origin has to cover
+        /// <code>1000 + 100 = 1100px</code>
+        /// so that the edge starts at the end of the scrolling container.
+        /// One scroll length of 1000px covers 5000ms of time, so the time required to cover 1100px is equal to
+        /// <code>5000 * (1100 / 1000) = 5500ms,</code>
+        /// and therefore the object should start being visible at
+        /// <code>7000 - 5500 = 1500ms.</code>
```
smoogipoo: Hmm, this example is a little bit hard to understand, although the correct result is still arrived at. Specifically the line `the origin has to cover 1100px` to me assumes that the edge of the hitobject is already positioned at the edge of the scroll area, but this is not true as they are separated by 300 more pixels (`= (7000 - 5000) * pixels-per-ms - offset`).

```
Consider the following:
                          
< < < < < < < < < < < < < < < < < < < < < < < < (Scroll direction)
         Scroll Area              Object
[--------------------------]      [-----o-----]
      TimeRange = 5000                Origin
       Distance = 1000            Offset = 100px
                                    Time = 7000

Pixels-per-ms (ppms)     = 1000 / 5000 = 0.2
Distance-to-origin (dto) = ppms * 7000 = 1400
Distance-to-edge (dte)   = dto - 100 = 1300

Visible-at = dte * 1 / ppms - TimeRange = 1500
```

Am I missing something? Because this also means that the implementation of `OverlappingScrollAlgorithm` should instead be:

`(originTime - TimeRange - offset * timeRange / scrollLength) / controlPoint.Multiplier)`

That is to say, take the origin time delta, subtract the offset duration, and finally factor in the time-range manipulation control point.

bdach: As far as I can tell, both lines of reasoning are correct. The point of possible confusion is, that when I wrote the 1100px number, what I meant is that the hit object will travel that much in *screen space*, not absolutely from `time = 0`. This is why my assumption of the edge being at the start of the scroll wasn't an incorrect one to make.

The only thing that's not quite right in your formula is that you've divided too much by the control point multiplier (it should not affect the origin time). In fact, if you rearrange your formula with that fixed, you actually get the same result:
```
originTime - TimeRange / controlPoint.Multiplier - offset * timeRange / scrollLength / controlPoint.Multiplier =
(extract common factor)
originTime - TimeRange / controlPoint.Multiplier * (1 + offset / scroll length) =
(move division by scrollLength outside inner bracket)
originTime - TimeRange / controlPoint.Multiplier * (scrollLength + offset) / scrollLength
```
at which point differs from my formula only with the order of operations (I do multiplications first in code so that divisions don't introduce floating point inaccuracies when dividing early).

Please let me know how I can make this clearer.

smoogipoo: Ok thanks! I had a feeling it was just my seeing the issue in a different way :)

smoogipoo (approved):

PR #7753, opened 2020-02-07
Show current placement blueprint in timeline

peppy:
image

### DIFF
--- a/osu.Game.Rulesets.Mania/Edit/Blueprints/ManiaPlacementBlueprint.cs
+++ b/osu.Game.Rulesets.Mania/Edit/Blueprints/ManiaPlacementBlueprint.cs
@@ -56,7 +56,7 @@ namespace osu.Game.Rulesets.Mania.Edit.Blueprints
 
         protected override void OnMouseUp(MouseUpEvent e)
         {
-            EndPlacement();
+            EndPlacement(true);
             base.OnMouseUp(e);
         }
 
--- a/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBlueprintContainer.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBlueprintContainer.cs
@@ -3,6 +3,7 @@
 
 using System;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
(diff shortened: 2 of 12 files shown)


osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBlueprintContainer.cs
```diff
@@ -43,6 +51,29 @@ protected override void LoadComplete()
...
+                    placementBlueprint = null;
+                }
+            }
+            else
+            {
+                placementBlueprint = CreateBlueprintFor(obj.NewValue);
+
+                placementBlueprint.Colour = Color4.MediumPurple;
+
+                SelectionBlueprints.Add(placementBlueprint);
```
smoogipoo: So what actually removes the placement from the selection blueprints in this case? I see the one above, which removes it if it's been set to null, but what about when the user cycles through their tools?

peppy: on tool change placement is ended with commit=false

osu.Game.Rulesets.Mania/Edit/Blueprints/ManiaPlacementBlueprint.cs
```diff
@@ -56,7 +56,7 @@ protected override bool OnMouseDown(MouseDownEvent e)
 
         protected override void OnMouseUp(MouseUpEvent e)
         {
-            EndPlacement();
+            EndPlacement(true);
```
bdach: As a side note this doesn't necessarily match the behaviour in the standard editor as `BeginPlacement` is called in `OnMouseDown` here (so it is only then that the placement is shown in the mania editor, instead of on mouse movement). Out of scope here, since at a quick glance it'd require some higher-level placement logic changes, and you probably know that, but I figured I'd say it just in case.

bdach (dismissed):
Looks OK to me. Played with it a bit, seems to work well in the standard editor 👍

osu.Game.Rulesets.Mania/Edit/Blueprints/ManiaPlacementBlueprint.cs
peppy: the mania editor right now needs a pass to fix its behaviour (aiming to get osu! mode editor perfected first).

as such we haven't tested mania at all yet, so i'm surprised it actually works!

smoogipoo (approved):

PR #7755, opened 2020-02-07
Fix spinner placement blueprint in multiple ways

peppy:
- Correct left/right click handling
- Updating values more frequently for timeline placement display
- Disallow negative duration

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Spinners/SpinnerPlacementBlueprint.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Spinners/SpinnerPlacementBlueprint.cs
@@ -1,13 +1,14 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osu.Framework.Graphics;
+using System;
 using osu.Framework.Input.Events;
 using osu.Game.Rulesets.Edit;
 using osu.Game.Rulesets.Osu.Edit.Blueprints.Spinners.Components;
 using osu.Game.Rulesets.Osu.Objects;
 using osu.Game.Rulesets.Osu.UI;
 using osuTK;
+using osuTK.Input;
 
 namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Spinners
 {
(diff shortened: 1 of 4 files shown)


osu.Game/Rulesets/Edit/HitObjectComposer.cs
```diff
@@ -49,8 +49,9 @@ public abstract class HitObjectComposer<TObject> : HitObjectComposer, IPlacement
         [Resolved]
         private IBeatSnapProvider beatSnapProvider { get; set; }
 
+        protected ComposeBlueprintContainer BlueprintContainer;
```
smoogipoo:
```suggestion
        protected ComposeBlueprintContainer BlueprintContainer { get; private set; }
```

osu.Game.Rulesets.Osu/Edit/Blueprints/Spinners/SpinnerPlacementBlueprint.cs
```diff
@@ -29,22 +30,29 @@ protected override void Update()
...
+                if (e.Button != MouseButton.Right)
+                    return false;
+
                 HitObject.EndTime = EditorClock.CurrentTime;
                 EndPlacement();
             }
             else
             {
-                isPlacingEnd = true;
-                piece.FadeTo(1f, 150, Easing.OutQuint);
```
smoogipoo: Intentionally removed?

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Edit/Blueprints/Spinners/SpinnerPlacementBlueprint.cs
peppy: I actually removed it to figure out what it was doing, and never noticed a difference. It already starts at alpha 1, no?

smoogipoo: Should be 0.5? (ctor)

peppy: ah, i see. i guess revert it if you think it looks better – removal was definitely not intentional.

smoogipoo: Added it back since without it there is no visual feedback that you're placing.

smoogipoo (approved):

PR #7756, opened 2020-02-07
Seek to previous object's end time on successful placement

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/7753 (for easier merge).


bdach:
I'm not sure what it is precisely (I figured you guys might know better since you're more familiar with editor code, so I didn't investigate further), but `GetEndTime()` seems to not be working quite right in some cases.

See [this video](https://streamable.com/1qx2d) - I placed a slider and then got seeked to a little bit *before* the slider end. This seems to vary by beatmap too sometimes, as I had another case where I think it seeked past the end.

smoogipoo:
Very weird indeed, but it isn't caused by this PR so I'm going to go ahead with merging.

bdach (approved):

PR #7758, opened 2020-02-07
Receive historical monthly user playcounts from API

EVAST9919:


osu.Game/Users/User.cs
```diff
@@ -203,6 +203,21 @@ public class UserAchievement
             public int ID;
         }
 
+        [JsonProperty("monthly_playcounts")]
```
smoogipoo: @nanaya how am I supposed to be reading the osu-web docs? It's extremely inconsistent for `User` to the point that I don't trust the docs to even be telling me the correct json properties.

This maps to `$monthlyPlaycounts`:

https://github.com/ppy/osu-web/blob/c353e7386b3b265ae0ed555d64874afbf52c41d5/app/Models/User.php#L75

The one below maps to `$replaysWatchedCounts`:

https://github.com/ppy/osu-web/blob/c353e7386b3b265ae0ed555d64874afbf52c41d5/app/Models/User.php#L97

The `rankHistory` one above maps to `$rankHistories`:

https://github.com/ppy/osu-web/blob/c353e7386b3b265ae0ed555d64874afbf52c41d5/app/Models/User.php#L93

I know that we should be using lower_snake_case for all json properties, but given that there already is the inconsistency in `rankHistory`, it's very hard to know whether these are correct being snake-case themselves...

nanaya: That's because it's not the documentation for json response...

(try checking UserTransformer in Transformers directory instead)

smoogipoo: I see... Thanks. I guess I should assume `rankHistory` is wrong then since it's not included in any shape/form in there?

smoogipoo (approved):

osu.Game/Users/User.cs
nanaya: looks like that one missed the snake case cleanup a while back because of the way it's currently added to the response.

bdach (approved):

Issue #7027, opened 2019-12-01
Carets in text input should blink to currently playing music


recapitalverb:
**Describe the new feature:** I think the title is self-explanatory - the cursor (seen in the image below) should blink to the beat of the currently playing song.
![image](https://user-images.githubusercontent.com/41869184/69909612-0c907180-1430-11ea-8840-55d7765cf64f.png)

GSculerlor:
Afaik caret loop duration is hardcoded on framework side (See https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/UserInterface/TextBox.cs#L380)

bdach:
Doing this might not look great with high BPM tracks. Would need to try and see.

HoLLy-HaCKeR:
> Doing this might not look great with high BPM tracks. Would need to try and see.

Could be scaled down similar to how CtB bananas are handled: keep dividing by 2 until you get to a reasonable number

MightyHelper:
I'm not sure how this would look so I'll suggest it:
If the current playing song is in a Kiai section, then make the text `jump` in time with the music. ( could be in time with the caret.)
By `jump` I mean, increase the thickness ( kinda like bold but **not half as much** )
I suggest making it togglable if it gets too annoying on any specific map/song.

bdach:
Personally I'm quite opposed to messing with the text thickness, I'm 99% sure it'll be too distracting.

KaiHicks:
> 
> 
> > Doing this might not look great with high BPM tracks. Would need to try and see.
> 
> Could be scaled down similar to how CtB bananas are handled: keep dividing by 2 until you get to a reasonable number

What about maps with a high prime BPM? In those cases, it would probably be best to just revert to the standard blink-rate.

InvisibleSymbol:
I think you all are overthinking this, as there already is an implementation of this behaviour in lazer, see https://github.com/ppy/osu/blob/master/osu.Game.Tests/Visual/UserInterface/TestSceneBeatSyncedContainer.cs. Pretty sure this can also be applied here.

PR #7761, opened 2020-02-07
Make the caret blink to the beat

jorolf:
- Closes https://github.com/ppy/osu/issues/7027

### DIFF
--- a/osu.Game/Graphics/UserInterface/OsuTextBox.cs
+++ b/osu.Game/Graphics/UserInterface/OsuTextBox.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Allocation;
+using osu.Framework.Audio.Track;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Graphics.UserInterface;
@@ -9,6 +10,9 @@ using osu.Game.Graphics.Sprites;
 using osuTK.Graphics;
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Input.Events;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Graphics.Containers;
+using osuTK;
 
 namespace osu.Game.Graphics.UserInterface
 {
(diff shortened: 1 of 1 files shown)


bdach:
Code looks good to me. The blinking caret looks really cool when there is a track playing, but when there is no track it can take a while for the next blink. (Weirdly enough it'll decrease the rate of blinking as time goes on - not sure what that's about)

Maybe we want to expose

https://github.com/ppy/osu/blob/e308845fafde6bf15bf66fef5e1618b6efbc2000/osu.Game/Graphics/Containers/BeatSyncedContainer.cs#L49

to make this blink a bit faster by default? Not sure.

osu.Game/Graphics/UserInterface/OsuTextBox.cs
```diff
@@ -59,5 +63,47 @@ protected override void OnFocusLost(FocusLostEvent e)
...
+            SelectionColour = SelectionColour,
+        };
+
+        private class BeatCaret : BasicCaret
+        {
+            private bool hasSelection;
+
+            public BeatCaret()
+            {
+                AddInternal(new CaretBeatSyncedContainer(this));
```
peppy: not too keen on passing self into child. instead of this, i'd propose making this caret derive `Caret` not `BasicCaret`, moving the logic we use from the basic implementation across and in the process putting the box inside the beat synced portion (so no passing of `this` is required).

peppy (changes requested):

peppy:
@bdach i can't repro the "getting slower" part fwiw. can you confirm this is truly a thing?

bdach:
I'm pretty sure I am getting some irregular beating at song select when changing filters so that they don't match any beatmaps. Here's [a video](https://streamable.com/a56te), and here's a screenshot from Vegas in which I roughly bookmarked all keyframes in which the caret alpha goes to `0.7f`:

image

Here are the roughly-measured frame times of each blink (taken from vegas timeline, at 60 FPS recording speed):

```
00:00.31
00:01.02 (+0.31)
00:01.32 (+0.30)
00:02.04 (+0.32)
00:02.35 (+0.31)
00:03.08 (+0.33)
00:03.47 (+0.39) - might be bogus since measured during filtering
00:04.46 (+0.59)
00:07.46 (+3.00)
00:08.47 (+1.01)
00:11.47 (+3.00)
00:12.47 (+1.00)
00:17.47 (+5.00)
00:18.47 (+1.00)
00:23.47 (+5.00)
```
Based on this it appears the blinking is missing some beats. Unsure if that's due to incorrect use of `BeatSyncedContainer`, or a bug in the container itself.

peppy:
@bdach Can reproduce under the scenario you mention. Working on a fix (is a `BeatSyncedContainer` level issue, I believe).

peppy:
I think this should be fine to go. The default rate is a bit slow, but I do like that it is in sync with the logo. We could decide to increase this rate a bit going forward (if someone wants to give this a try, please do so in a new PR).

peppy (approved):

PR #7762, opened 2020-02-07
Implement UserHistoryGraph component

EVAST9919:
Depends on:
- [x] https://github.com/ppy/osu/pull/7758


osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,235 @@
...
+
+        public ITooltip GetCustomTooltip() => GetTooltip();
+
+        public object TooltipContent => GetTooltipContent();
+
+        protected abstract UserGraphTooltip GetTooltip();
+
+        protected abstract object GetTooltipContent();
+
+        protected class RankChartLineGraph : LineGraph
```
bdach: Needs to be renamed since this is now in the abstract. Maybe `HoverableLineGraph`? `UserLineGraph`?

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,90 @@
...
+                updateValues(value);
+            }
+        }
+
+        private readonly string tooltipCounterName;
+
+        public UserHistoryGraph(string tooltipCounterName)
+        {
+            this.tooltipCounterName = tooltipCounterName;
+        }
```
bdach: Not sure how I feel about passing this in the ctor. Probably better suited as a 
```csharp
protected abstract string TooltipCounterName { get; }
```

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+
+            Data = values.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToArray();
+
+            if (values.Length > 1)
+            {
+                Graph.DefaultValueCount = Data.Length;
+                Graph.Values = Data.Select(x => (float)x.Value);
+                Graph.FadeIn(FADE_DURATION, Easing.Out);
+            }
```
bdach: This chunk of code is almost shared with `RankGraph`. The only difference is how `Values` is calculated, which can be solved with a `Func<TValue, float>`.

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        private void updateValues(UserHistoryCount[] values)
+        {
+            if (values == null || !values.Any())
+            {
+                Graph.FadeOut(FADE_DURATION, Easing.Out);
+                Data = null;
+                return;
+            }
```
bdach: This is also almost duplicated verbatim in the other graph.

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,235 @@
...
+                return new Vector2(index / (float)(DefaultValueCount - 1), y);
+            }
+        }
+
+        protected abstract class UserGraphTooltip : VisibilityContainer, ITooltip
+        {
+            protected readonly OsuSpriteText Counter, BottomText;
+            private readonly Box background;
+
+            protected UserGraphTooltip(string topText)
```
bdach: Same thing applies here, `topText` could be an abstract property.

bdach (dismissed):

bdach:
I've done some refactoring on this to remove the code duplication, and I think it's looking pretty alright now.

There's a slight semantics change. The old rank graph would show a tooltip even if the graph data was `null` - it'd not show the graph, but instead jerry-rig the tooltip to show the top rank, but due to how the logic was implemented, it'd use the wrong date, so I removed that case entirely as broken. I don't think that's what web does anyways.

peppy:
Is it intended that this isn't used outside of tests?

bdach:
I think it might have been (integration will come in a separate PR) but I'm not sure - @EVAST9919, can you confirm/deny?

EVAST9919:
Yes, it's a part of another component, which will come in a separate pr.

bdach (dismissed):
Noticed a few other minor nits on the re-review. Fixed them up since I had the branch local anyways.

Bump approving but double check if possible please 😅

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,276 @@
...
+                return new Vector2(index / (float)(DefaultValueCount - 1), y);
+            }
+        }
+
+        protected abstract class UserGraphTooltip : VisibilityContainer, ITooltip
+        {
+            protected readonly OsuSpriteText Counter, BottomText;
+            private readonly Box background;
+
+            protected abstract string TooltipCounterName { get; }
```
peppy: comment. what is this?

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,57 @@
...
+        {
+            set
+            {
+                if (value == null)
+                {
+                    Data = null;
+                    return;
+                }
+
+                Data = value.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToArray();
```
peppy:
```suggestion
                Data = value?.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToArray();
```

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,276 @@
...
+            if (data.Length > 1)
+                graph.UpdateBallPosition(e.MousePosition.X);
+
+            return base.OnMouseMove(e);
+        }
+
+        protected override void OnHoverLost(HoverLostEvent e)
+        {
+            if (data.Length > 1)
+                graph.HideBar();
```
peppy: <=1
>1
>1

confusing. choose one and stick to it.

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,276 @@
...
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Input.Events;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
+using osu.Game.Graphics.UserInterface;
+using osuTK;
+
+namespace osu.Game.Overlays.Profile
+{
+    public abstract class UserGraph<TKey, TValue> : Container, IHasCustomTooltip
```
peppy: xmldoc this class. xmldoc all abstract classes *completely*.

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,276 @@
...
+    {
+        protected const float FADE_DURATION = 150;
+
+        private readonly UserLineGraph graph;
+        private KeyValuePair<TKey, TValue>[] data;
+        private int dataIndex;
+
+        protected UserGraph()
+        {
+            data = Array.Empty<KeyValuePair<TKey, TValue>>();
```
peppy: i don't understand why internally this thing makes this non-nullable while handling incoming nulls. usually we support null or don't the whole way, not at point of `set`. i'd either make `data` nullable or stop supporting null as a valid input.

osu.Game.Tests/Visual/Online/TestSceneUserHistoryGraph.cs
```diff
@@ -0,0 +1,122 @@
...
+            {
+                RelativeSizeAxes = Axes.X,
+                Height = 200,
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+            });
+
+            var values = new[]
+            {
+                new UserHistoryCount
```
peppy: makes all these one-liners

peppy (changes requested):

osu.Game/Overlays/Profile/UserGraph.cs
bdach: that one was me - think there was a place where null was passed in (might have been a test) so I wanted to do some defensive programming but it does make sense to stick to one convention

peppy: fine to do this, but i'd move to the field definition and clearly mark as `[NotNull]`

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -83,7 +88,7 @@ protected override void OnHoverLost(HoverLostEvent e)
 
         private void redrawGraph()
         {
-            if (data.Length == 0)
```
bdach:
```csharp
data?.Length > 1
!data?.Any() ?? true
```
No consistency still.

osu.Game/Overlays/Profile/UserGraph.cs
bdach: There are still undocumented protected members (from the more important ones - `UserGraph.Data`, `UserGraph.GetDataPointHeight`). Rest is self-explanatory to a degree I suppose.

osu.Game/Overlays/Profile/UserGraph.cs
```diff
@@ -0,0 +1,284 @@
...
+            }
+        }
+
+        protected abstract class UserGraphTooltip : VisibilityContainer, ITooltip
+        {
+            protected readonly OsuSpriteText Counter, BottomText;
+            private readonly Box background;
+
+            /// <summary>
+            /// Text which will be shown near the <see cref="Counter"/>.
```
bdach: This xmldoc is very nondescript and doesn't really say what this *does*. Better:
```suggestion
            /// Text describing the value being plotted on the graph, which will be displayed as a prefix to the value of <see cref="Counter"/>.
```

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using static osu.Game.Users.User;
+
+namespace osu.Game.Overlays.Profile.Sections.Historical
+{
+    public abstract class UserHistoryGraph : UserGraph<DateTime, long>
```
bdach: I just noticed this while scanning for implementors of `UserGraph` - is there a good reason this is abstract?

osu.Game.Tests/Visual/Online/TestSceneUserHistoryGraph.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        private class TestGraph : UserHistoryGraph
+        {
+            protected override UserGraphTooltip GetTooltip() => new TestTooltip();
+
+            private class TestTooltip : HistoryGraphTooltip
+            {
+                protected override string TooltipCounterName => "Test Counter";
+            }
+        }
```
bdach: I see no reason for this to exist if `UserHistoryGraph` is made non-abstract.

bdach (dismissed):

osu.Game/Overlays/Profile/UserGraph.cs
bdach: github bugged out as it seemingly doesn't like review comments of a range of commits, but the second variant is there in the current diff.

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
EVAST9919: Need to check my follow-up pr: there was a reason, but at this point I'm not sure. If there's no reason - will remove.

EVAST9919: Tooltip is the only reason. We have two `UserHistoryGraph`s in the overlay with different tooltips
Screenshot_1
Screenshot_2

bdach: Overriding an entire class to change some text seems like overkill. They're semantically a little different but can be always split if necessary. I'd advocate for publicly settable property instead of having an overridable property but at this point there's so many conventions I'm not going to insist on it.

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,60 @@
...
+    {
+        public UserHistoryCount[] Values
+        {
+            set => Data = value?.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToArray();
+        }
+
+        /// <summary>
+        /// Text describing the value being plotted on the graph, which will be displayed as a prefix to the value in the <see cref="HistoryGraphTooltip"/>
+        /// </summary>
+        public string TooltipCounterName { get; set; } = @"Plays";
```
EVAST9919: I guess at this point it's better to make it just a ctor parameter, but I'm not sure

bdach:
@EVAST9919 I've done some further changes here. Please review when you have the time.

EVAST9919:
I have no complains with that, thanks

peppy:
@EVAST9919 is there a follow-up PR to this which actually makes use of it in the profile overlay?

osu.Game.Tests/Visual/Online/TestSceneUserHistoryGraph.cs
```diff
@@ -0,0 +1,57 @@
...
+                new UserHistoryCount { Date = new DateTime(2010, 6, 1), Count = 20 },
+                new UserHistoryCount { Date = new DateTime(2010, 7, 1), Count = 20000 },
+                new UserHistoryCount { Date = new DateTime(2010, 8, 1), Count = 30 },
+                new UserHistoryCount { Date = new DateTime(2010, 9, 1), Count = 50 },
+                new UserHistoryCount { Date = new DateTime(2010, 10, 1), Count = 2000 },
+                new UserHistoryCount { Date = new DateTime(2010, 11, 1), Count = 2100 }
+            };
+
+            AddStep("Set fake values", () => graph.Values = values);
+            AddStep("Set more values", () => graph.Values = moreValues);
```
nbvdkamp, line 52: Since the history data that is retrieved from web is sparse (only months where plays are recorded are returned) I think UserHistoryGraph should add the months in between with 0 values like web does.
```suggestion

            var gapValues = new[]
            {
                new UserHistoryCount { Date = new DateTime(2010, 5, 1), Count = 10 },
                new UserHistoryCount { Date = new DateTime(2010, 8, 1), Count = 20 },
                new UserHistoryCount { Date = new DateTime(2010, 11, 1), Count = 15 }
            };

            AddStep("Set fake values", () => graph.Values = values);
            AddStep("Set more values", () => graph.Values = moreValues);
            AddStep("Set values with gaps", () => graph.Values = gapValues);
```

peppy (approved):
Seems okay.

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
```diff
@@ -0,0 +1,62 @@
...
+using static osu.Game.Users.User;
+
+namespace osu.Game.Overlays.Profile.Sections.Historical
+{
+    public class UserHistoryGraph : UserGraph<DateTime, long>
+    {
+        [CanBeNull]
+        public UserHistoryCount[] Values
+        {
+            set => Data = value?.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToArray();
```
nbvdkamp, line 17: This adds 0 values for months with missing data
```suggestion
            set 
            {
                if (value == null)
                {
                    Data = null;
                    return; 
                }

                List<KeyValuePair<DateTime, long>> data = value?.Select(v => new KeyValuePair<DateTime, long>(v.Date, v.Count)).ToList();

                int toSkip = 0;
            
                // Insert missing months with value 0 for display
                for (int i = 0; i < data.Count(); i++)
                {
                    i += toSkip;
                    if (i == data.Count() - 1)
                        break;

                    DateTime current = data[i].Key;
                    DateTime next = data[i + 1].Key;

                    List<KeyValuePair<DateTime, long>> toInsert = new List<KeyValuePair<DateTime, long>>();

                    for (int j = 1; DateTime.Compare(current.AddMonths(j), next) < 0; j++)
                        toInsert.Add(new KeyValuePair<DateTime, long>(current.AddMonths(j), 0));

                    data.InsertRange(i + 1, toInsert);
                    toSkip = toInsert.Count;
                }

                Data = data.ToArray();
            }
```

nbvdkamp (changes requested):

osu.Game.Tests/Visual/Online/TestSceneUserHistoryGraph.cs
EVAST9919: 0 values will be handled in another place (my plan initially)

nbvdkamp: Ah ok then you can disregard this

osu.Game/Overlays/Profile/Sections/Historical/UserHistoryGraph.cs
peppy: Hard to say where this responsibility should be. But as this PR is already huge let's keep this for a follow-up.

EVAST9919: If you are interested in my approach (will be added in another pr):
Аннотация 2020-11-11 052216

PR #7763, opened 2020-02-07
Make accuracy formatting more consistent

thewildtree:
Currently, `FormatAccuracy` omits all decimal places if the accuracy equals 100%. This behaviour is consistent with neither osu-web nor osu-stable, as you can see here:

stable:
image

web:
image

lazer on master:
image

Additionally, lazer itself isn't consistent in this matter since the in-game HUD (`PercentageCounter`) doesn't use `FormatAccuracy` and happily displays both decimal places when the accuracy equals 100%:
image

As discussed [here](https://github.com/ppy/osu/pull/7730#discussion_r375457851), it doesn't really make sense to go back and forth on the amount of decimal places just because the acc value hits 100, because why `0.00%` (or any other `.00%` value) but not `100.00%`?

This PR aims to make `FormatAccuracy` consistent with what has been the standard formatting in osu! for years, making it match osu-web / osu-stable and be more consistent throughout the whole game.

### DIFF
--- a/osu.Game/Graphics/UserInterface/PercentageCounter.cs
+++ b/osu.Game/Graphics/UserInterface/PercentageCounter.cs
@@ -3,6 +3,7 @@
 
 using System;
 using osu.Framework.Allocation;
+using osu.Game.Utils;
 
 namespace osu.Game.Graphics.UserInterface
 {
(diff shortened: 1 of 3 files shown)


bdach (approved):
Missed one spot which I fixed up, looks good otherwise 👍

peppy (approved):

PR #7764, opened 2020-02-08
Implement Spun Out mod

mcendu:
Alternative to / closes #7731


osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
```diff
@@ -73,6 +73,8 @@ public bool Complete
             }
         }
 
+        public bool AutoSpin { get; set; } = false;
```
peppy: `HitObject` implementations (and their drawables) should not know about mods. You will need to move this logic to inside your mod class.

peppy (changes requested):

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    [TestFixture]
+    public class TestSceneSpinnerSpunOut : TestSceneSpinner
+    {
+        public override IReadOnlyList<Type> RequiredTypes => base.RequiredTypes.Concat(new[] { typeof(OsuModSpunOut) }).ToList();
+
+        [SetUp]
+        public void SetUp() => Schedule(() =>
+        {
+            SelectedMods.Value = new[] { new OsuModSpunOut() };
```
mcendu: Spinner automatically spins in game but does not in this test.

peppy:
Please avoid pushing until you're done with your changes locally. There's no reason to push each commit individually (it will needlessly trigger CI and notifications).

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -18,5 +23,33 @@ public class OsuModSpunOut : Mod
...
+        public void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> drawables)
+        {
+            foreach (var hitObject in drawables)
+            {
+                if (hitObject is DrawableSpinner spinner)
+                {
+                    spinner.Disc.Trackable = false;
+                    spinner.Disc.OnUpdate += d =>
+                    {
+                        if (d is SpinnerDisc s)
```
mcendu: Looks like a redundant check, but...

bdach:
Seems to work properly in the game, although with two footnotes:

* Spin rate reported is 238 (vs 286 in stable) - could impact extra points awarded for spins,
* When the spinner is autospinning holding down any key will make the spinner background light up like if normal input was received ([see video](https://streamable.com/taitk)).

Will move on to looking at the code now.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -18,5 +23,33 @@ public class OsuModSpunOut : Mod
...
+                        {
+                            if (s.Valid)
```
bdach: Can merge conditions:
```csharp
if (d is SpinnerDisc s && s.IsValid)
```

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
bdach: Why is this required? Is this just a weird null check? I don't see how this could be false otherwise.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
bdach: It doesn't spin because `frameDelay` is always 0 inside the mod implementation. I assume you need an actual backing clock to test this properly.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
```diff
@@ -0,0 +1,23 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using NUnit.Framework;
+using osu.Game.Rulesets.Osu.Mods;
+
+namespace osu.Game.Rulesets.Osu.Tests
+{
+    [TestFixture]
+    public class TestSceneSpinnerSpunOut : TestSceneSpinner
```
bdach: This test scene is broken because the way it's written, the `[SetUp]` necessary for the tests to even make sense is the *last* step.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
```diff
@@ -73,6 +73,9 @@ public bool Complete
             }
         }
 
+        public bool Valid => spinner.StartTime <= Time.Current && spinner.EndTime > Time.Current;
+        public bool Trackable { get; set; } = true;
```
bdach: I'm not sure this is the proper way to go. It solves the problem, but is yet another flag to take into account inside of the spinner, and doesn't really make sense semantically (what's the meaning of a spinner being *trackable* or not?). It'd be cleaner if the disc could stay unmodified at all.

Not sure what the right answer I'd be looking here is. Manual `TriggerEvent`s, maybe? Some kind of proxying input manager that would only send events to the spinner (which I'm not even sure is viable)? I have no idea myself, so this comment isn't a change request *per se*.

bdach (dismissed):
Still think this should be set:
```csharp
public override ModType Type => ModType.Automation;
```

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
bdach: Also it should actually be *asserting* something.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
huoyaoyuan:
```csharp
if (d is SpinnerDisc { IsValid: true } s)
```
Just a hint for new syntax.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
mcendu: AFAIK the issue of `[SetUp]` being the last step in test browser is also present in `TestSceneSpinnerHidden`.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
mcendu: Fixed by 596f4f7d2e6c62c7a4c3fa43ce161c5b71c388d8

mcendu:
> Seems to work properly in the game, although with two footnotes:
> 
> * Spin rate reported is 238 (vs 286 in stable) - could impact extra points awarded for spins,

I've chose to take ["half of Auto's 477 spins per minute"](https://osu.ppy.sh/help/wiki/Game_Modifiers#spun-out) at face value (which is 238) instead of 286, as the former looks easier to implement.
https://github.com/ppy/osu/blob/e78d94d4692d02ed5843e5f23882c660e23099d1/osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs#L37

> * When the spinner is autospinning holding down any key will make the spinner background light up like if normal input was received ([see video](https://streamable.com/taitk)).

Thanks, done.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
bdach: Possible fallacies of pre-existing code are no excuse for fallacies of new code.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,41 @@
...
+        public void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> drawables)
+        {
+            foreach (var hitObject in drawables)
+            {
+                if (hitObject is DrawableSpinner spinner)
+                {
+                    spinner.Disc.Enabled = false;
+                    spinner.Disc.OnUpdate += d =>
+                    {
+                        if (d is SpinnerDisc s && s.Valid)
```
bdach: I'd still like clarification as to why this type check is necessary.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -177,7 +177,7 @@ protected override void CheckForResult(bool userTriggered, double timeOffset)
         protected override void Update()
         {
             Disc.Tracking = OsuActionInputManager?.PressedActions.Any(x => x == OsuAction.LeftButton || x == OsuAction.RightButton) ?? false;
-            if (!SpmCounter.IsPresent && Disc.Tracking)
+            if (!SpmCounter.IsPresent && (Disc.Tracking || !Disc.Enabled))
```
bdach: The `Enabled` thing would be pretty good if not for this. In the context of this PR this change seems fine, but when reading the source at face value it makes no sense why the SPM counter should be visible when disabled. This means that the spinner *still* essentially is aware of the mod, but in a round-about way (through a nebulous property).

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
```diff
@@ -0,0 +1,23 @@
...
+    {
+        public override IReadOnlyList<Type> RequiredTypes => base.RequiredTypes.Concat(new[] { typeof(OsuModSpunOut) }).ToList();
```
iiSaLMaN:
```csharp
        public override IReadOnlyList<Type> RequiredTypes => new[]
        {
            typeof(OsuModSpunOut)
        };
```

frenzibyte (changes requested):

iiSaLMaN:
Currently, the changes around rotating the spinner in `SpinnerDisc` and `DrawableSpinner` made them not making sense at all, but they're the less complex way, seems OK for what it's worth.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerSpunOut.cs
bdach: Unconvinced this suggestion is correct, since applying this change might not pick up types required by the base scene.

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
mcendu: Null check to suppress CI warning.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
```diff
@@ -126,7 +115,30 @@ protected override void Update()
                     .FadeTo(tracking_alpha, 250, Easing.OutQuint);
             }
 
-            this.RotateTo(currentRotation / 2, validAndTracking ? 500 : 1500, Easing.OutExpo);
+            this.RotateTo(currentRotation / 2, 500, Easing.OutExpo);
```
bdach: What's the reason for this change?

bdach (commented):
I'm actually mostly fine with this now. I fixed up the tests for you and done some slight refactoring. At this point I want to clarify one more thing.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerDisc.cs
mcendu: I can’t find any scenarios other than spun out that would make the spinner turn without being valid and tracking, so removing as a redundant condition.

mcendu:
How do we derive the speed of stable's spun out from auto's speed?

osu.Game.Rulesets.Osu/Mods/OsuModSpunOut.cs
```diff
@@ -2,21 +2,48 @@
...
+                }
+            }
+        }
+
+        private void autoSpin(Drawable drawable)
+        {
+            if (drawable is DrawableSpinner spinner)
+            {
+                if (spinner.Disc.Valid)
+                    spinner.Disc.Rotate(180 / MathF.PI * (float)spinner.Clock.ElapsedFrameTime / 40);
```
cl8n:
```suggestion
                    spinner.Disc.Rotate(180 / MathF.PI * (float)spinner.Clock.ElapsedFrameTime * 0.03f);
```
3/5 the speed of auto, 30 radians per second. i'll fix wiki

cl8n (changes requested):

peppy:
I was hoping this to be a simple merge but the approach was wrong in multiple ways. @mcendu please check my changes and ask if you have any questions. You were basically making too many changes to (and assumptions about) the hitobject itself, which would almost definitely break in the future.

A further improvement may be to move the rotation method to `DrawableSpinner` itself and have `OsuModSpunOut` not know about the disc at all, but I think this can come later.

@smoogipoo requesting review, specifically regarding the `DrawableHitObject` change.

peppy (dismissed):

bdach:
FWIW I looked through the new changes and agree that they're a considerable improvement that I admittedly did not think of when reviewing. I'll attempt to draw conclusion from this for the future.

smoogipoo (approved):

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -38,6 +38,19 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
...
         public IReadOnlyList<DrawableHitObject> NestedHitObjects => nestedHitObjects.IsValueCreated ? nestedHitObjects.Value : (IReadOnlyList<DrawableHitObject>)Array.Empty<DrawableHitObject>();
 
+        /// <summary>
+        /// Whether this object should handle any user input events.
+        /// </summary>
+        public bool HandleUserInput { get; set; } = true;
+
+        public override bool HandlePositionalInput => HandleUserInput;
+
+        public override bool HandleNonPositionalInput => HandleUserInput;
```
smoogipoo: I think these are unnecessary and serve as an anti-optimisation: https://github.com/ppy/osu-framework/wiki/Handling-input-events#controlling-whether-input-is-to-be-handled

smoogipoo (changes requested):

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
peppy: ah yeah, wasn't sure about those. free to remove.

smoogipoo (approved):

PR #7765, opened 2020-02-08
Avoid throwing exceptions on DrawableRuleset load interruption

### LINKED ISSUE #7759: OperationCancelledException when exiting gameplay
**Describe the crash:** 
Sometimes when exiting from gameplay to the song select screen while debugging, the game will throw the following exception:

image

Game works fine after resuming in the IDE and this is not visible in logs, but I think it's still not something that should be happening

**osu!lazer version:** 
latest master as of 07/02

peppy:
Closes https://github.com/ppy/osu/issues/7759.

### DIFF
--- a/osu.Game/Rulesets/UI/DrawableRuleset.cs
+++ b/osu.Game/Rulesets/UI/DrawableRuleset.cs
@@ -214,11 +214,14 @@ namespace osu.Game.Rulesets.UI
         {
             foreach (TObject h in Beatmap.HitObjects)
             {
-                cancellationToken?.ThrowIfCancellationRequested();
+                if (cancellationToken?.IsCancellationRequested == true)
+                    return;
+
                 addHitObject(h);
             }
 
-            cancellationToken?.ThrowIfCancellationRequested();
+            if (cancellationToken?.IsCancellationRequested == true)
+                return;
 
             Playfield.PostProcess();
 


osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -214,11 +214,14 @@ private void loadObjects(CancellationToken? cancellationToken)
...
             {
-                cancellationToken?.ThrowIfCancellationRequested();
+                if (cancellationToken?.IsCancellationRequested == true)
+                    return;
+
                 addHitObject(h);
             }
 
-            cancellationToken?.ThrowIfCancellationRequested();
+            if (cancellationToken?.IsCancellationRequested == true)
```
TheWildTree: no need for `== true` :')

peppy: have you tried this? i think you'll find there is.

TheWildTree: I only looked at the code, my bad if that's correct

huoyaoyuan: 2 points:
- For this expression that results in `bool?`, `== true` is meaningful.
- But the usage of `CancellationToken?` is unnecessary.  `CancellationToken.None` and `default(CancellationToken)` both mean "will never be canceled", and widely used in BCL.

I'll send a PR for this.

peppy: You're welcome to look into that, but I believe it is nullable for a reason (to do with DI).

bdach (approved):
Looks safe enough, judging by the call site of `loadObjects`.

smoogipoo:
By merging this you have to consider the possible repercussions from exiting rather than throwing - throwing will interrupt the entire load async call, up to the point it started. Exiting will not do this.

I don’t like this on that principle.

The mentioned issue is a debugger exception. It is irrelevant, the user should ignore exceptions they don’t want.

peppy:
Can you explain how the repercussions would affect us in this case? I checked the complete flow and it looks safe. I was trying to find best practices for this, since this is a case we are 100% in control of the async flow (*we* are the call site).

I'm all for avoiding throwing exceptions we don't need to. For the issue you mention to affect us, it would mean that the framework loading was unaware of the cancellation token's presence, which I think we are guaranteed to not be the case?

peppy:
On further discussion it seems safest to throw here after all. Reasoning is that `DrawableRuleset` is not sealed – a custom ruleset could be doing more execution in the BDL method which would continue to run. By throwing an exception we can be sure we exit as early as possible.

PR #7768, opened 2020-02-08
Minor cleanups for legacy beatmap decoders

turbedi:
Replaces some string.StartsWith(string, StringComparison.Ordinal) calls with ring.StartsWith(char) , when only one char is compared. Possible since .NET-Standard 2.1

And another LegacyStoryboardDecoder.handleEvents() cleanup, saves some MB of allocations.

### DIFF
--- a/osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
@@ -64,15 +64,16 @@ namespace osu.Game.Beatmaps.Formats
         private void handleEvents(string line)
         {
             var depth = 0;
-            var lineSpan = line.AsSpan();
 
-            while (lineSpan.StartsWith(" ", StringComparison.Ordinal) || lineSpan.StartsWith("_", StringComparison.Ordinal))
+            foreach (char c in line)
             {
-                lineSpan = lineSpan.Slice(1);
-                ++depth;
+                if (c == ' ' || c == '_')
+                    depth++;
+                else
+                    break;
             }
 
-            line = lineSpan.ToString();
+            line = line.Substring(depth);
 
             decodeVariables(ref line);
 
--- a/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
@@ -64,7 +64,7 @@ namespace osu.Game.Beatmaps.Formats
                 hitObject.ApplyDefaults(this.beatmap.ControlPointInfo, this.beatmap.BeatmapInfo.BaseDifficulty);
         }
 
-        protected override bool ShouldSkipLine(string line) => base.ShouldSkipLine(line) || line.StartsWith(" ", StringComparison.Ordinal) || line.StartsWith("_", StringComparison.Ordinal);
+        protected override bool ShouldSkipLine(string line) => base.ShouldSkipLine(line) || line.StartsWith(' ') || line.StartsWith('_');
 
         protected override void ParseLine(Beatmap beatmap, Section section, string line)
         {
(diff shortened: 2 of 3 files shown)


peppy:
When making alloc changes, provding BenchmarkDotNet memory analysis (`[MemoryDiagnoser]`) results is very appreciated (so we don't have to double-check ourselves).

bdach:
Could include the benchmark in the PR as part of the recently-added [benchmark project](https://github.com/ppy/osu/tree/master/osu.Game.Benchmarks) too, so that anyone can just run it at their convenience. There's already a beatmap parsing benchmark in there, but it seems to be more CPU-oriented.

turbedi:
Tested with the storyboard of [this beatmap](https://osu.ppy.sh/beatmapsets/470977).

master:
|                     Method |    Mean |    Error |   StdDev |      Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|--------------------------- |--------:|---------:|---------:|-----------:|-----------:|----------:|----------:|
| BenchmarkStoryboard | 1.257 s | 0.0064 s | 0.0059 s | 72000.0000 | 20000.0000 | 1000.0000 | 427.92 MB |

This PR:
|                     Method |    Mean |    Error |   StdDev |      Gen 0 |      Gen 1 |     Gen 2 | Allocated |
|--------------------------- |--------:|---------:|---------:|-----------:|-----------:|----------:|----------:|
| BenchmarkStoryboard | 1.227 s | 0.0076 s | 0.0067 s | 71000.0000 | 20000.0000 | 1000.0000 | 418.15 MB |

I didn't include the benchmark in this PR, because there aren't any storyboard files under osu.Game.Tests/Resources, [refering to this comment from the benchmark project PR](https://github.com/ppy/osu/pull/7474#discussion_r364533606).

bdach:
Mind linking the benchmark source as a gist or something at least, then?

turbedi:
> 
> 
> Mind linking the benchmark source as a gist or something at least, then?

https://gist.github.com/turbedi/eaf4a6e99d0fa6f49acb1f2c2f917b41

peppy (approved):

osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
```diff
@@ -64,15 +64,16 @@ protected override void ParseLine(Storyboard storyboard, Section section, string
...
             {
-                lineSpan = lineSpan.Slice(1);
-                ++depth;
+                if (c == ' ' || c == '_')
+                    depth++;
+                else
+                    break;
             }
 
-            line = lineSpan.ToString();
```
huoyaoyuan: `TrimStart(' ', '_')`

huoyaoyuan (commented):
Overlaps with planned spanify optimization, but good to change now.

osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
peppy: have you benchmarked this? it's all much of a muchness because this isn't even a hot path, but probably worth knowing for future reference.

huoyaoyuan: No, but I believe the `TrimStart<T>(this ReadOnlySpan<T> span, ReadOnlySpan<T> trimElements) where T : IEquatable<T>` overload should have a good performance.

turbedi: I benchmarked it, allocations went from 418.15 MB to 436.69 MB, so even worse than before this PR.

huoyaoyuan: You can store the param array into a static field, which won't increase allocation.

turbedi: True, but I think the code looks cleaner this way.

smoogipoo (approved):

PR #7773, opened 2020-02-09
Improve mod settings tests

protheory8:
Now test checks if Customisation menu closes after deselecting mod.


bdach:
In a freak accident, the CI failure was in the exact same test modified here, although as far as I can tell it was a pre-existing fallacy of the test. The mod overlay could be forced to attempt to select mods while not being in a ready state, which could be replicated when running tests headless with `[Repeat(10000)]`.

I've pushed in a fix for that (could no longer replicate the failure under the same stress test), and additionally renamed the test changed originally to better reflect its purpose. Should be all good now, but I'm not approving as I touched this in a major way.

peppy (approved):

PR #7776, opened 2020-02-09
Fix BeatSyncContainer failing at song select

peppy:
Due to an unfortunate clash of sane defaults, `BeatSyncContainer` could skip beats while displaying the dummy beatmap at song select. This is due to its default beat length being 1,000ms, which aligns with `DummyWorkingBeatmap`'s virtual track length of 1,000ms. As this is looped at song select, there's a race condition where the BeatSyncContainer would incorrectly fire multiple or no times.

Hard to test this one, so no additional tests are provided.

Closes issue pointed out at https://github.com/ppy/osu/pull/7761#issuecomment-583772201.

### DIFF
--- a/osu.Game/Beatmaps/WorkingBeatmap.cs
+++ b/osu.Game/Beatmaps/WorkingBeatmap.cs
@@ -59,7 +59,7 @@ namespace osu.Game.Beatmaps
             switch (lastObject)
             {
                 case null:
-                    length = excess_length;
+                    length = 0;
                     break;
 
                 case IHasEndTime endTime:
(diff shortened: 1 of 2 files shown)


bdach:
I can confirm that the problem goes away with this PR applied, but the changes are a bit in over my head. Nothing in the code jumps out to me, but I'd have to get familiar with `BeatSyncedContainer` to be able to approve this with reasonable confidence, so I think I'll defer on this one.

smoogipoo (approved):

smoogipoo:
As mentioned I'm a little bit sketched out on the underlying issue being circumvented rather than being fixed, but we've concluded it's unlikely for such cases to be reached naturally.

PR #7779, opened 2020-02-09
Add tooltips with precise dates to beatmap set overlay

### LINKED ISSUE #7777: Dates inside BeatmapSetContainer don't have tooltip
**Describe the bug:** All dates inside the `BeatmapSetContainer` don't have a tooltip that should show their "long version" like in the web version.

|  | Lazer  | Web |
| -------- | ------------- | ------------- |
| submission/rank status date | image | image |
| score date | image | image |

**osu!lazer version:** 2020.208.0

Malox10:
Closes #7777 

Field constructor is intentionally duplicated because the old one becomes obsolete as soon as #7540 is merged.


osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
```diff
@@ -50,19 +51,19 @@ private void updateDisplay()
...
+                new Field("submitted", online.Submitted, OsuFont.GetFont(weight: FontWeight.Bold))
                 {
                     Margin = new MarginPadding { Top = 5 },
                 },
             };
 
             if (online.Ranked.HasValue)
             {
-                fields.Add(new Field("ranked on", online.Ranked.Value.ToString(@"MMMM d, yyyy"), OsuFont.GetFont(weight: FontWeight.Bold)));
+                fields.Add(new Field("ranked", online.Ranked.Value, OsuFont.GetFont(weight: FontWeight.Bold)));
```
bdach: This is not consistent with web - dropping the `ToString()` means that the date will get automatically humanised. See screenshots in the original issue.

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
```diff
@@ -50,19 +51,19 @@ private void updateDisplay()
             fields.Children = new Drawable[]
             {
                 new Field("mapped by", BeatmapSet.Metadata.Author.Username, OsuFont.GetFont(weight: FontWeight.Regular, italics: true)),
-                new Field("submitted on", online.Submitted.ToString(@"MMMM d, yyyy"), OsuFont.GetFont(weight: FontWeight.Bold))
+                new Field("submitted", online.Submitted, OsuFont.GetFont(weight: FontWeight.Bold))
```
bdach: Also not consistent with the web date format.

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
```diff
@@ -126,6 +127,25 @@ public Field(string first, string second, FontUsage secondFont)
...
+                {
+                    new OsuSpriteText
+                    {
+                        Text = $"{first} ",
+                        Font = OsuFont.GetFont(size: 13)
+                    },
+                    new DrawableDate(second)
+                    {
+                        Font = secondFont.With(size: 13)
+                    }
```
bdach: The appearance of these elements doesn't match web design.

bdach (changes requested):
Aside from the following comments, this PR does not fully resolve the linked issue as it does not add a tooltip to rank 1/personal best score panels lower down in the beatmap set overlay.

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
Malox10: Wouldn't it make more sense to instead change web to the humanised form to be consistent with all other occurrences of time stamps?
Capture
Capture2
Capture3
Capture4

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
Malox10: I'm not quite sure what's wrong here. Can you elaborate?

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
bdach: I imagine this might have been an intentional choice. Note that humanised dates are used in lists of things, where it's more readable to have a human-readable offset rather than a precise date. In the case of beatmap metadata, however, it is more useful to see the full date outright without having to hover over it or trying to do mental math. I am however not sure.

cc @arflyte - could you confirm/deny if this was intentional?

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
bdach: OK, I highlighted the wrong place, but what I meant to point out is the difference in typography - web tooltip has date in white, bold font, time + offset in regular, slightly dimmed colour, while lazer-side it's all regular white text.

Malox10: This would require a framework change for how tooltips work which is beyond the scope of this PR.

bdach: I don't think that's correct. Custom tooltips are already supported - check the `IHasCustomTooltip` interface and its lazer-side implementations.

Now changing `DrawableDate` to use it *might* warrant a separate PR, but I'm not overly convinced it does.

Malox10: You're right. I switched to `IHasCustomTooltip` now.

Malox10:
I used [`RankGraphTooltip`](https://github.com/ppy/osu/blob/804d210250a7b2b430da59fab7f55a6e2cd40394/osu.Game/Overlays/Profile/Header/Components/RankGraph.cs#L218) for reference. Not sure if the colors are right, but they will eventually be supplied by `OverlayColourProvider` anyway.

osu.Game/Graphics/DrawableDate.cs
```diff
@@ -75,6 +78,74 @@ private void updateTimeWithReschedule()
...
+                };
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(OsuColour colours)
+            {
+                // Temporary colour since it's currently impossible to change it without bugs (see https://github.com/ppy/osu-framework/issues/3231)
+                // If above is fixed, this should use OverlayColourProvider
+                background.Colour = colours.Gray1;
+                timeText.Colour = colours.GreyCyanLighter;
```
bdach: Since `DrawableDate` is used in more places than just the beatmap set overlay, I think while we can't yet use `OverlayColourProvider` here we should use a neutral shade of grey for now.

osu.Game/Graphics/DrawableDate.cs
```diff
@@ -75,6 +78,74 @@ private void updateTimeWithReschedule()
...
+
+            protected override void PopIn() => this.FadeIn(200, Easing.OutQuint);
+            protected override void PopOut() => this.FadeOut(200, Easing.OutQuint);
+
+            public bool SetContent(object content)
+            {
+                if (!(content is DateTimeOffset date))
+                    return false;
+
+                dateText.Text = string.Format($"{date:d MMMM yyyy}") + " ";
```
bdach: Can inline the space in the `string.Format` and save a string concatenation.

bdach (commented):
A couple of nits left to iron out and we still need feedback from @arflyte re: the date format, but otherwise looks pretty good now.

osu.Game/Graphics/DrawableDate.cs
bdach: OK I just totally misread that - the `string.Format` doesn't make much sense as you're doing string interpolation inside of it and not passing args. Remove the call and put the space inside the interpolated string.

osu.Game/Overlays/BeatmapSet/AuthorInfo.cs
peppy: It should be matching, so short date format is best here.

peppy: actually let's go with relative for now. it feels better.

peppy (approved):

PR #7780, opened 2020-02-10
Fix order and naming of Difficulty Adjust sliders

### LINKED ISSUE #7387: Order/Naming of Difficulty Adjust settings are different from beatmap details

voidedWarranties:
Fixes #7387 by adding parameters `OrderMode` and `OrderPosition` to `SettingSource`
Options exist to put a setting first, by relative order using `OrderPosition`, in no particular order, and last.
All settings without any `OrderMode` will be put after the settings ordered numerically but before the one marked as last.
Also made all slider labels consistent with the website.


osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -20,22 +21,33 @@ namespace osu.Game.Configuration
     [AttributeUsage(AttributeTargets.Property)]
     public class SettingSourceAttribute : Attribute
     {
+        public const int FIRST = 0;
+        public const int ORDERED_RELATIVE = 1;
+        public const int UNORDERED = 2;
+        public const int LAST = 3;
```
smoogipoo: Should be an enum, not consts.

smoogipoo (changes requested):

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -20,22 +21,36 @@ namespace osu.Game.Configuration
     [AttributeUsage(AttributeTargets.Property)]
     public class SettingSourceAttribute : Attribute
     {
+        public enum OrderMode
+        {
+            FIRST,
+            ORDERED_RELATIVE,
+            UNORDERED,
+            LAST
```
bdach: Enum members in C# are usually Pascal-cased (`First`, `OrderedRelative`...)

bdach (commented):
I have a feeling this is overengineered. I'm not sure four ordering modes are necessary here. I'd just keep `OrderPosition` and maybe make it an `int?`, so that you can skip the order. Maybe a bit questionable what happens when you get some `null`s and some non-`null`s, but I'd just pick an option (probably ordered first) and document it with an xmldoc.

osu.Game/Configuration/SettingSourceAttribute.cs
voidedWarranties: Will note enum casing for future contributions

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -24,18 +29,21 @@ public class SettingSourceAttribute : Attribute
 
         public string Description { get; }
 
-        public SettingSourceAttribute(string label, string description = null)
+        public int OrderPosition { get; }
+
+        public SettingSourceAttribute(string label, string description = null, int orderPosition = -1)
```
smoogipoo: How about making this nullable instead, and allowing negative ordering?

It's pretty arbitrary that the base mod has orders `1, 2` and the derived mods have orders `0, 3`. Meaning if there was a mod that has multiple settings before the `1`, it couldn't do it due to the `> -1` limitation.

smoogipoo (changes requested):

osu.Game/Configuration/SettingSourceAttribute.cs
voidedWarranties: I attempted in the past to make the `int` an `int?` but C# doesn't support nullables as attribute arguments. Not sure if there is any workaround to this

voidedWarranties: I think the best solution for the feedback I've received so far is to add back a parameter `Ordered` but this time only as a `bool` which serves as an indication of whether the setting should be sorted (rather than using a value of -1). Would this suffice?

smoogipoo: I see, didn't realise there was such a limitation. Maybe a second constructor?

voidedWarranties: In order for `orderPosition` to be a required parameter (in a separate constructor), `description` has to be required as well. Should I make it required or set a default value for `orderPosition`?

Or, I could move `orderPosition` before `description`, but I think that would be weird

osu.Game.Rulesets.Catch/Mods/CatchModDifficultyAdjust.cs
```diff
@@ -10,7 +10,7 @@ namespace osu.Game.Rulesets.Catch.Mods
 {
     public class CatchModDifficultyAdjust : ModDifficultyAdjust
     {
-        [SettingSource("Fruit Size", "Override a beatmap's set CS.")]
+        [SettingSource("Circle Size", "Override a beatmap's set CS.", 0)]
```
bdach: I don't think this is a proper change. We've already fixed up `Circle SIze` -> `Key Amount` for mania in song select (albeit with a hack of sorts), so if we've gone that way then we should probably also fix up `Fruit Size` there as well instead of changing the mod this time.

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -16,6 +17,10 @@ namespace osu.Game.Configuration
     /// An attribute to mark a bindable as being exposed to the user via settings controls.
     /// Can be used in conjunction with <see cref="SettingSourceExtensions.CreateSettingsControls"/> to automatically create UI controls.
     /// </summary>
+    /// <remarks>
+    /// All controls with OrderPosition set to an int greater than 0 will be placed first in ascending order.
```
bdach:
```suggestion
    /// All controls with <see cref="OrderPosition"/> set will be placed first in ascending order.
```

osu.Game/Configuration/SettingSourceAttribute.cs
```diff
@@ -16,6 +17,10 @@ namespace osu.Game.Configuration
...
+    /// All controls with no OrderPosition will come afterward in default order.
```
bdach:
```suggestion
    /// All controls with no <see cref="OrderPosition"/> will come afterward in default order.
```

osu.Game.Rulesets.Catch/Mods/CatchModDifficultyAdjust.cs
voidedWarranties: peppy says [here](https://discordapp.com/channels/188630481301012481/188630652340404224/676283363430236161) that all naming should be consistent with web. Do I change it back anyway?

bdach: Ah, fair call then, missed that. Disregard for now.

peppy:
I've added constants just so we can track the cross-class references. A bit ugly but I think the added safety makes it worth it.

peppy (approved):

PR #7783, opened 2020-02-10
Bypass song select filter to show externally changed beatmap temporarily

### LINKED ISSUE #7774: Changing beatmap while song select has text filter causes desync
image

- Enter song select
- Enter a filter which matches no beatmap
- Change beatmap from playlist
- Delete filter

Observe:

- Wedge showing no beatmap selected
- Carousel showing one beatmap selected
- Music overlay showing another beatmap selected

peppy:
image

- Closes https://github.com/ppy/osu/issues/7774


smoogipoo (approved):
Seems to work fine

PR #7784, opened 2020-02-10
Fix disposal-related errors by making WorkingBeatmap non-disposable

smoogipoo:
While debugging https://github.com/ppy/osu/issues/7774, it's possible to encounter an `AggregateException` in `WorkingBeatmap.Beatmap` instead of the expected `TaskCanceledException`. This is actually [expected](https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-cancellation), and we should be catching aggregate exceptions.

However while doing that I also noticed that we are returning disposed beatmaps from the working beatmap cache of `BeatmapManager`. This is all kinds of wrong, so I've also made `WorkingBeatmap` non-disposable.

In doing so, I've replaced `Dispose()` with a `CancelAsyncLoad()` method, which should make the entire load process a lot more safe:
- If load -> cancel -> load, the previous task is cancelled, a new load task is started, and the beatmap is allowed to load under the new task.
- If load -> load complete -> cancel, the task isn't cleared and the beatmap can be reused.

Related:
- `ClockBackedTestWorkingBeatmap` and `WaveformTestBeatmap` now use finalisers to dispose the track stores.

Finally, all methods log sentry-loggable errors before returning nulls. In every case except the async beatmap task cancellation, an exception is unexpected and we should be alerted.

Finally, I've also added a null-check to `LogoVisualisation`, because that's another area where I've encountered a null-ref in the same trigger scenario. Ideally we want to mark all these properties as `[CanBeNull]` and add local null-checks in each case, but that goes pretty deep.

### DIFF
--- a/osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
+++ b/osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
@@ -36,8 +36,9 @@ namespace osu.Game.Beatmaps
                     using (var stream = new LineBufferedReader(store.GetStream(getPathForFile(BeatmapInfo.Path))))
                         return Decoder.GetDecoder<Beatmap>(stream).Decode(stream);
                 }
-                catch
+                catch (Exception e)
                 {
+                    Logger.Error(e, "Beatmap failed to load");
                     return null;
                 }
             }
@@ -59,8 +60,9 @@ namespace osu.Game.Beatmaps
                 {
                     return textureStore.Get(getPathForFile(Metadata.BackgroundFile));
                 }
-                catch
+                catch (Exception e)
                 {
+                    Logger.Error(e, "Background failed to load");
                     return null;
                 }
             }
@@ -74,8 +76,9 @@ namespace osu.Game.Beatmaps
                 {
                     return new VideoSprite(textureStore.GetStream(getPathForFile(Metadata.VideoFile)));
                 }
-                catch
+                catch (Exception e)
                 {
+                    Logger.Error(e, "Video failed to load");
                     return null;
                 }
             }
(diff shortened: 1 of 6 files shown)


osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
```diff
@@ -36,8 +36,9 @@ protected override IBeatmap GetBeatmap()
                     using (var stream = new LineBufferedReader(store.GetStream(getPathForFile(BeatmapInfo.Path))))
                         return Decoder.GetDecoder<Beatmap>(stream).Decode(stream);
                 }
-                catch
+                catch (Exception e)
                 {
+                    Logger.Error(e, "Beatmap failed to load");
```
peppy: Intentionally using `Logger.Error` here? I'm not against it, but we probably want to change the output target.

peppy:
I don't believe this is a regression as I can repro on master, but after scrolling through all of song select, the `WorkingBeatmaps` stat stays very high (even after waiting and forcing GC).

At main menu, going through the music controller one song at a time (using next) will slowly recover the count.



Probably worth double-checking whether it can/should be fixed in this PR.

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
smoogipoo: I think we want to be notified of these errors since they tend to be pretty critical. Keep in mind there already is a secondary catch inside the beatmap decoder itself which catches general parsing errors without reporting to us:

https://github.com/ppy/osu/blob/0f1d30c55ee2c627f680d98a0b284946d7055769/osu.Game/Beatmaps/Formats/LegacyDecoder.cs#L47-L54

peppy: Okay, let's go with it for now.

PR #7785, opened 2020-02-10
Add ability to load long comment trees in CommentsContainer (Alternative)

EVAST9919:
Alternative to https://github.com/ppy/osu/pull/7775
Both works the same, but in this one `CommentsPage` handle all the logic


EVAST9919:
oops, wrong branch

PR #7786, opened 2020-02-10
Add ability to load long comment trees in CommentsContainer

EVAST9919:
Alternative to https://github.com/ppy/osu/pull/7775
Both works the same, but in this one CommentsPage handle all the logic


osu.Game/Overlays/Comments/CommentsPage.cs
```diff
@@ -52,18 +62,60 @@ private void load(OverlayColourProvider colourProvider)
...
+        private DrawableComment createCommentWithReplies(Comment comment, CommentBundle commentBundle)
+        {
+            var drawableComment = createDrawableComment(comment);
+
+            var replies = commentBundle.Comments.Where(c => c.ParentId == comment.Id);
+
+            if (replies.Any())
+            {
+                replies.ForEach(c => c.ParentComment = comment);
+                drawableComment.OnLoadComplete += _ => drawableComment.AddReplies(replies.Select(reply => createCommentWithReplies(reply, commentBundle)));
```
bdach: IMO this is *very* dangerous and insidious recursion. It's nice in concept, but I'd be incredibly hesitant to let this be.

Such recursion in `OnLoadComplete` introduces a branching factor. Say you have a comment tree, with 1 parent comment, which is 5 levels deep, and each parent comment has 3 child comments. That tree then has (3<sup>6</sup> - 1)/2 = 364 comments, which is not that much.

However, for each child comment you're launching a linear scan of `commentBundle`, which'd mean 364 linear scans of the entire comment set. Moreover, I'd have to check how the threading with `OnLoadComplete` is done, but if it's done using worker threads on the thread pool, you could end up saturating the pool at the lower levels of the tree (since the further down you go, the more times the recursion fires). Level 0 of the tree only has the root node, level 1 - 3 nodes, level 2 - 9, level 3 - 27, and so on up to level 5 = 243 nodes. Or, if it's done on the draw thread, you'll starve that threads with `OnLoadComplete` callbacks.

bdach (changes requested):

osu.Game/Overlays/Comments/CommentsPage.cs
EVAST9919: In theory I wanted to not use `OnLoadComplete` at all, but we can't use `AddReplies` if the drawable isn't loaded since it has `LoadAsync` function.

What about scanning `commentBundle` - it would be nice if all the comments were already ordered in tree-like manner. But may be they already are - I don't know yet (will look at it). Anyway I'll revisit all that system

EVAST9919: Also, is it worth to work on? I mean, is handling all that logic in `CommentsPage` (excluding this case) is good enough?

EVAST9919: Just a side note: we receiving maximum 50 comments at once (thanks to web api), so may be it's not a big deal?
(I'll remove `LoadComplete` anyway, already have some thoughts)

bdach:
I think that for this to be handled correctly, it is imperative that the tree structure be reconstructed, so that updates can be as local and as minimal as necessary. In my view, the perfect solution would involve two data structures: a normal tree, and a dictionary (or anything associative that has performant by-key lookup).

Let's say that upon the starting load of comments, we have the following situation:

Untitled Diagram

Five comments, in a tree on the left (solid arrows represent parent-child relationships), with a look-up table done on the dictionary on the right (dashed arrows represent by-key lookups). Now, assume we have received a new bundle of comments.

Untitled Diagram(1)

Unlucky for us, the bundle is (a) not a complete tree, so we need to append nodes, and (b) not ordered topologically, by which I mean that we might possibly get a child comment before a parent comment. Now I don't know whether that situation is possible with the data web returns, but it's better not to assume.

Let's start with the first comment. We don't have its parent anywhere, so let's create an orphaned node and treat it as a top-level comment for now, but remember that it's supposed to have a parent somewhere.

Untitled Diagram(2)

Onto comment 202. First, let's look at all orphaned nodes that are supposed to have parents - aha, there's the parent for 201, so we can make it a child of 202. Moreover, 202 itself is also a child of 104 which we've already seen, so we can make it its child, yielding this structure:

Untitled Diagram(3)

If we follow this procedure, we are able to reconstruct the whole tree from the bottom up. The pseudocode would be:

```
nodeDictionary = new Dictionary<int, Comment>();
topLevelNodes = new List<Comment>();
orphanedNodes = new List<Comment>();

foreach (var comment in bundle)
{
    if (comment.ParentId is null)
        topLevelNodes.Add(comment);

    foreach (var orphan in orphanedNodes)
        if (orphan.ParentId == comment.Id)
            comment.ChildComments.Add(orphan);

    if (!nodeDictionary.Contains(comment.ParentId))
        orphanedNodes.Add(comment);
    else
        nodeDictionary[comment.ParentId].ChildComments.Add(comment);
}
```

This should be correct, pretty good in performance, and most importantly, iterative and not recursive. I'd recommend encapsulating this algorithm in a separate class, which can have an `Add(CommentBundle)` method that calls the pseudocode above.

peppy:
As an aside, without checking the contents of this pull request or the conversation above, did you check how this is implemented at `osu-web`'s side as a sanity check? Rather than reinventing the wheel, doing things in the same way (assuming it's sane) seems like the best way forward.

bdach:
I know next to nothing about react and stores, but web does seem to be doing something [remarkably](https://github.com/ppy/osu-web/blob/f62fdb23ac3e08af26f9b5e1f2f4842525085877/resources/assets/lib/stores/ui-state-store.ts) [similar](https://github.com/ppy/osu-web/blob/f62fdb23ac3e08af26f9b5e1f2f4842525085877/resources/assets/lib/stores/comment-store.ts) to the solution I suggested. Mind you, they don't have the `orphanedNodes` thing since I assume API *must* then return topologically-sorted comments, but the dual lookup table-tree structure *is* there.

smoogipoo:
I've refactored/adjusted a few things:
1. Removed orphaned comments being made into top-level comments.
2. Removed the duplicate web request - they're the same endpoint.
3. Fixed initial comments being loaded with a `null` parent_id. Turns out that `null` essentially returns the comments listing (e.g. https://osu.ppy.sh/comments). For the initial request you want to use `parent_id = 0`.
4. Fixed included comments not being used at all, along with the properties for the comments not being set (`User`, `IsVoted`).

I've kept orphaning only as a fail-safe for the case where they _may_ be returned out of order by osu-web.

@peppy @bdach I consider this PR complete

smoogipoo (approved):

smoogipoo:
As a further refactoring that can be done post-merge, `CommentsPage` shouldn't exist. All of that functionality should be combined into the main comments listing.

PR #7795, opened 2020-02-11
Implement CommentEditor component

EVAST9919:
Screenshot_1


osu.Game/Overlays/Comments/CancellableCommentEditor.cs
```diff
@@ -0,0 +1,71 @@
...
+                            Text = @"Cancel"
+                        }
+                    }
+                };
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(OverlayColourProvider colourProvider)
+            {
+                IdleColour = colourProvider.GetColour(0.5f, 0.45f);
```
peppy: this shouldn't be a thing. we don't want to be getting raw values ever. if you want to change brightness, use the `Darken` or `Lighten` extension methods.

osu.Game/Overlays/OverlayColourProvider.cs
```diff
@@ -16,28 +16,28 @@ public OverlayColourProvider(OverlayColourScheme colourScheme)
...
+        public Color4 Foreground1 => GetColour(0.1f, 0.6f);
+        public Color4 Background1 => GetColour(0.1f, 0.4f);
+        public Color4 Background2 => GetColour(0.1f, 0.3f);
+        public Color4 Background3 => GetColour(0.1f, 0.25f);
+        public Color4 Background4 => GetColour(0.1f, 0.2f);
+        public Color4 Background5 => GetColour(0.1f, 0.15f);
+        public Color4 Background6 => GetColour(0.1f, 0.1f);
 
-        private Color4 getColour(float saturation, float lightness) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, 1));
+        public Color4 GetColour(float saturation, float lightness) => Color4.FromHsl(new Vector4(getBaseHue(colourScheme), saturation, lightness, 1));
```
peppy: should be reverted to `private`

peppy (dismissed):

bdach (dismissed):
Tests with assertions should be added for:

* committing text via button (check that text is committed and that the button is loading),
* committing text via keyboard (as above),
* trying to commit text via keyboard when button is disabled (making sure it doesn't do anything),
* pressing the cancel button in the cancellable variant (making sure it fires the action).

bdach (commented):
Did a touch-up pass on this, looks ok to me otherwise. Not approving since I touched.

smoogipoo (approved):
Looks fine to me

PR #7800, opened 2020-02-12
Fix editor hit objects displaying incorrectly after StartTime change

### LINKED ISSUE #7769: Editor: hitcircle/approachcircle starts glitching when you move it in the timeline
sliders dont have this bug.

https://streamable.com/h3vtc

**osu!lazer version:** 
2020.208.0

peppy:
- Closes https://github.com/ppy/osu/issues/7769.
- [x] Depends on https://github.com/ppy/osu-framework/pull/3256
- [x] Depends on https://github.com/ppy/osu-framework/pull/3257
- [x] Depends on framework update


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -116,7 +116,7 @@ protected override void LoadComplete()
             HitObject.DefaultsApplied += onDefaultsApplied;
 
             startTimeBindable = HitObject.StartTimeBindable.GetBoundCopy();
-            startTimeBindable.BindValueChanged(_ => updateState(ArmedState.Idle, true));
+            startTimeBindable.BindValueChanged(_ => updateState(State.Value, true));
```
bdach: Threw me in a loop since I thought this was a no-op for a sec, but I missed the `force: true` so all good.

bdach (approved):
Looks correct as far as I can tell. Also tested and can confirm it seems to fix the issue 👍

smoogipoo (approved):

PR #7805, opened 2020-02-13
Update placement blueprint's position more often

peppy:
Fixes blueprint not updating in time (timeline) or position (compose) when the mouse isn't being moved.

- [x] Depends on https://github.com/ppy/osu/pull/7803 (only for clean merge)

Before:

image

After:

image


bdach (approved):
Looks simple enough to feel correct; had the desired effect upon running. Perhaps a bit of a shame it has to be done in `Update()` but I can't think of an elegant way to only update when necessary since at least two simultaneous conditions would have to be involved, so probably best to stick with this unless it's deemed an issue in the future.

smoogipoo (approved):

PR #7807, opened 2020-02-13
Fix not being able to seek using scroll wheel in timeline while playing track

peppy:


osu.Game/Screens/Edit/Compose/Components/Timeline/ZoomableScrollContainer.cs
```diff
@@ -104,8 +109,15 @@ protected override void LoadComplete()
         protected override bool OnScroll(ScrollEvent e)
         {
             if (e.IsPrecise)
+            {
+                // can't handle scroll correctly while playing.
+                // the editor will handle this case for us.
+                if (editorClock.IsRunning)
+                    return false;
```
bdach: It's slightly unclear to me as to why this check is done only in the `IsPrecise` case. I'm not even really sure what that flag means, as I'd have to dig in source (xmldocs are either not present or obtuse).

peppy: `IsPrecise` denotes that we are on a high precision trackpad (surface or apple laptops) where they will send absolute floating point scroll data, rather than int “notches”.

Editor has a different scroll implementation for devices with this support, else the `Editor` itself will handle

smoogipoo (dismissed):

smoogipoo (approved):

PR #7809, opened 2020-02-13
Allow selecting/playing a specific difficulty using the beatmapset difficulty icons

voidedWarranties:
`LClick`ing a beatmap difficulty icon will send you directly to that difficulty rather than the easiest/last played one.
~~`Alt+LClick`ing a beatmap difficulty will immediately play that difficulty.~~

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -463,6 +463,46 @@ namespace osu.Game.Tests.Visual.SongSelect
             AddAssert("Selected beatmap has not changed", () => songSelect.Carousel.SelectedBeatmap.ID == previousID);
         }
 
+        [Test]
+        public void TestDifficultyIconSelecting()
+        {
+            int? previousID = null;
+            addRulesetImportStep(0);
+            createSongSelect();
+
+            AddStep("Store current ID", () => previousID = songSelect.Carousel.SelectedBeatmap.ID);
+            AddStep("Click on a difficulty", () =>
+            {
+                InputManager.MoveMouseTo(songSelect.Carousel.ChildrenOfType<DrawableCarouselBeatmapSet.FilterableDifficultyIcon>()
+                                                   .First(icon => icon.Item.Beatmap != songSelect.Carousel.SelectedBeatmap));
+
+                InputManager.PressButton(MouseButton.Left);
+                InputManager.ReleaseButton(MouseButton.Left);
+            });
+            AddAssert("Selected beatmap changed", () => songSelect.Carousel.SelectedBeatmap.ID != previousID);
+
+            AddStep("Filter some difficulties", () => songSelect.Carousel.Filter(new FilterCriteria
+            {
+                BPM = new FilterCriteria.OptionalRange<double>
+                {
+                    Min = songSelect.Carousel.SelectedBeatmapSet.MaxBPM,
+                    IsLowerInclusive = true
+                }
+            }));
+            AddUntilStep("Wait for filter", () => songSelect.Carousel.ChildrenOfType<DrawableCarouselBeatmapSet.FilterableDifficultyIcon>().Any(icon => icon.Item.Filtered.Value));
+
+            AddStep("Store current ID", () => previousID = songSelect.Carousel.SelectedBeatmap.ID);
+            AddStep("Click on a filtered difficulty", () =>
+            {
+                InputManager.MoveMouseTo(songSelect.Carousel.ChildrenOfType<DrawableCarouselBeatmapSet.FilterableDifficultyIcon>()
+                                                   .First(icon => icon.Item.Filtered.Value));
+
+                InputManager.PressButton(MouseButton.Left);
+                InputManager.ReleaseButton(MouseButton.Left);
+            });
+            AddAssert("Selected beatmap has not changed", () => songSelect.Carousel.SelectedBeatmap.ID == previousID);
+        }
+
         private void addRulesetImportStep(int id) => AddStep($"import test map for ruleset {id}", () => importForRuleset(id));
 
         private void importForRuleset(int id) => manager.Import(createTestBeatmapSet(getImportId(), rulesets.AvailableRulesets.Where(r => r.ID == id).ToArray())).Wait();
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -466,41 +466,67 @@ namespace osu.Game.Tests.Visual.SongSelect
         [Test]
         public void TestDifficultyIconSelecting()
         {
-            int? previousID = null;
             addRulesetImportStep(0);
             createSongSelect();
 
-            AddStep("Store current ID", () => previousID = songSelect.Carousel.SelectedBeatmap.ID);
+            DrawableCarouselBeatmapSet set = null;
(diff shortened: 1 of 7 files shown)


osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -80,7 +80,7 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
 
         protected override BackgroundScreen CreateBackground() => new BackgroundScreenBeatmap(Beatmap.Value);
 
-        protected BeatmapCarousel Carousel { get; private set; }
+        public BeatmapCarousel Carousel { get; private set; }
```
peppy: this should not be done.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -210,12 +216,31 @@ public class FilterableDifficultyIcon : DifficultyIcon
         {
             private readonly BindableBool filtered = new BindableBool();
 
-            public FilterableDifficultyIcon(CarouselBeatmap item)
+            private readonly SongSelect songSelect;
+            private readonly BeatmapInfo info;
+
+            public FilterableDifficultyIcon(CarouselBeatmap item, SongSelect songSelect)
```
peppy: screens shouldn't be passed around like this. you could have DI'd this locally rather than in the parent (although read on – that is also not what we want).

peppy (changes requested):

peppy:
There are two ways you can go about this flow:

1.

- Move the DI for `SongSelect` local to the usage, similar to `DrawableCarouselBeatmap`.
- make a method on `SongSelect` similar to `FinaliseSelection` which performs the selection.

2.

- Instead DI in `OsuGame` and use `PresentBeatmap`. Note that you will need to make it support a `BeatmapInfo` as an override, since right now it's done on a set basis.


I'm not sure which of these solutions is better. Generally I'd recommend the second but `DrawableCarousel**` classes are already using the first, so I would be okay with it. If you're interested, I'd suggest trying the second and see how it works flow-wise.

voidedWarranties:
Attempted to implement solution 2 - certainly an improvement, but unsure if this is what you were looking for

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -210,12 +211,40 @@ public class FilterableDifficultyIcon : DifficultyIcon
...
+                info = item.Beatmap;
+            }
+
+            protected override bool OnClick(ClickEvent e)
+            {
+                if (!filtered.Value)
+                {
+                    game.PresentBeatmap(info);
+
+                    if (e.AltPressed)
```
peppy: this seems like weird / unnecessary behaviour.

voidedWarranties: Should the `Alt` functionality be removed?

peppy: Yeah, it's new behaviour which is weird and should be added as its own proposal, separate from this.

peppy:
You'll want to employ `(canBeNull: true)` to fix tests.

voidedWarranties:
Not sure if `PresentBeatmap` is the best route after all, as using this will result in strange side effects such as playing the incorrect sample and restarting the song when switching difficulties. Would creating a selection function in `SongSelect` be more efficient?
If so, should `OsuGame` be reverted to how it was before?

peppy:
Probably, yes.

voidedWarranties:
I think I found a solution which allows this functionality to work in testing. Is this ok?

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -210,12 +211,23 @@ public class FilterableDifficultyIcon : DifficultyIcon
         {
             private readonly BindableBool filtered = new BindableBool();
 
+            private readonly Action select;
```
peppy: i'd rather you stored the `CarouselBeatmap` here

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -210,12 +211,23 @@ public class FilterableDifficultyIcon : DifficultyIcon
...
                 filtered.TriggerChange();
+
+                select = item.Select;
+            }
+
+            protected override bool OnClick(ClickEvent e)
+            {
+                select?.Invoke();
+
+                return base.OnClick(e);
```
peppy: return `true` since you're handling the click.

osu.Game/OsuGame.cs
```diff
@@ -446,7 +446,7 @@ private void updateModDefaults()
         /// </summary>
         /// <param name="action">The action to perform once we are in the correct state.</param>
         /// <param name="validScreens">An optional collection of valid screen types. If any of these screens are already current we can perform the action immediately, else the first valid parent will be made current before performing the action. <see cref="MainMenu"/> is used if not specified.</param>
-        protected void PerformFromScreen(Action<IScreen> action, IEnumerable<Type> validScreens = null)
+        public void PerformFromScreen(Action<IScreen> action, IEnumerable<Type> validScreens = null)
```
peppy: revert

peppy (changes requested):

peppy:
Have you tried just using `CarouselBeatmap.State = Selected`?

peppy:
Needs a filtered check

image

Also it would be pretty good to have tests covering this, since we have quite comprehensive tests for `PlaySongSelect`.

osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
```diff
@@ -463,6 +463,46 @@ public void TestHideSetSelectsCorrectBeatmap()
...
+            AddStep("Store current ID", () => previousID = songSelect.Carousel.SelectedBeatmap.ID);
+            AddStep("Click on a difficulty", () =>
+            {
+                InputManager.MoveMouseTo(songSelect.Carousel.ChildrenOfType<DrawableCarouselBeatmapSet.FilterableDifficultyIcon>()
+                                                   .First(icon => icon.Item.Beatmap != songSelect.Carousel.SelectedBeatmap));
+
+                InputManager.PressButton(MouseButton.Left);
+                InputManager.ReleaseButton(MouseButton.Left);
+            });
+            AddAssert("Selected beatmap changed", () => songSelect.Carousel.SelectedBeatmap.ID != previousID);
```
bdach: This assertion is a bit strange; IMO it should be checking that the beatmap corresponding to the actual item clicked is selected, as that's the desired behaviour here really.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -211,7 +211,7 @@ public class FilterableDifficultyIcon : DifficultyIcon
         {
             private readonly BindableBool filtered = new BindableBool();
 
-            private readonly CarouselBeatmap item;
+            public readonly CarouselBeatmap Item;
```
bdach: Not too sure what to feel about this - ideally this exposition as `public` shouldn't be necessary, but it might be somewhat difficult to test without it (would most likely have to depend on beatmap item order in the carousel matching the icon order).

peppy (approved):

PR #7811, opened 2020-02-13
Improve extensibility of mod display expansion

smoogipoo:


peppy (approved):

peppy:
Weird that `AlwaysContracted` still shows an appear animation, but that's how it was so let's leave for now.

PR #7814, opened 2020-02-13
Remove the concept of a "current" item from multiplayer rooms

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/7812

Stepping stone. The test scene is broken but it always was, it will be removed along with the panel in a future PR, so /shrug.


peppy (approved):
i thiiiiiiiiink this seems ok

PR #7817, opened 2020-02-13
Make RankingsOverlay available in-game

EVAST9919:


osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -135,6 +135,9 @@ public void ShowCountry(Country requested)
 
         private void loadNewContent()
         {
+            if (ruleset.Value == null)
+                return;
```
bdach: When can this be the case?

EVAST9919: For some reason ruleset is null on first `Show()` call. I guess, not the best way to fix that - just first thought.

osu.Game/OsuGame.cs
```diff
@@ -600,6 +600,7 @@ protected override void LoadComplete()
             //overlay elements
             loadComponentSingleFile(direct = new DirectOverlay(), overlayContent.Add, true);
             loadComponentSingleFile(social = new SocialOverlay(), overlayContent.Add, true);
+            var rankingsOverlay = loadComponentSingleFile(new RankingsOverlay(), overlayContent.Add, true);
```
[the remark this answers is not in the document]
EVAST9919: Same as for changelog overlay. Not used somewhere else, only locally

osu.Game/Overlays/RankingsOverlay.cs
bdach: Even if this is the correct fix for this, "for some reason" does not inspire much confidence. The proper way to fix code issues is to investigate, understand and *then* settle on a fix.

peppy: It's pretty obvious why this is happening from a 5 second glance at the code. Even if you can't see it by looking at the code, one breakpoint should be enough to discern what's going on.

EVAST9919:
Since the overlay has too many issues no reason to leave this pr open until resolved

EVAST9919:
Reopening since all the related issues have been resolved

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -130,7 +130,7 @@ public void ShowCountry(Country requested)
 
             Show();
 
-            Country.Value = requested;
+            Schedule(() => Country.Value = requested);
```
bdach: What was the error that required adding this schedule? Mind giving a reproduction scenario?

Without that kind of information I have to go look on my own, which takes time that I can't spend on getting to review other things. I also can't just not look at all since if I don't look I can't judge if the fix is proper, since I can't know what this is fixing without looking.

EVAST9919: Open the game => open your profile => select your flag to trigger rankings overlay. Happens only on first load so looks like we are triggering country change when the overlay isn't fully loaded (that's how I see it)

smoogipoo: Can you describe the issue that this was causing? If it's attempting to change components that aren't loaded, it should rather be the responsibility of those components to handle it properly inside `LoadComplete()` rather than accepting changes during async load.

smoogipoo: Nevermind, I've investigated myself and the issue is:

`RankingsOverlay` binds value changed to `Country` in `LoadComplete`, which schedules loading new content. This schedule is run prior to children getting updated, and the web request created requires the ruleset.

The ruleset is a bindable that's held by a `TabControl` nested in the header. `TabControl` updates its current item only on `LoadComplete`. For the header and the nested `TabControl`, `LoadComplete` won't happen unless the overlay is visible (standard update optimisations).

So essentially `LoadComplete` of children is happening one frame too late - everything is loaded but you're relying on the state of children to determine logic in the overlay whereas it should be the other way around.

The proper fix here is for the overlay to contain a ruleset bindable of its own, which the header is bound to. For now, we'd like the ruleset bindable to be bound to the global game ruleset (effectively tying the toolbar and the overlay rulesets together). Depending on how we want to go about it, it would be a simple change to e.g. split them apart to bypass bindable leasing, etc, in the future. The important part is that the overlay should ensure that its own bindables are in the correct state at the point where it needs to use them.

peppy: Fixed for the time being by binding to the global ruleset. Long game plan is to *not* rely on children's bindables to decide values. the overlay should be in control of the bindable and only pass them down.

peppy (approved):

PR #7820, opened 2020-02-13
Update design of username and country name in RankingsOverlay

### LINKED ISSUE #7819: User and country names should be italic in the RankingsOverlay
lazer:
osu!_3nwl3w8mca
osu!_ESuRTo22wv
web:
chrome_HGMYamt2q1
chrome_YX20yJ545r

the *filtered by country* country name shouldnt become italic though.

EVAST9919:
Fixes https://github.com/ppy/osu/issues/7819

Screenshot_1
Screenshot_1


osu.Game/Overlays/Rankings/Tables/CountriesTable.cs
```diff
@@ -63,5 +60,20 @@ public CountriesTable(int page, IReadOnlyList<CountryStatistics> rankings)
                 Text = $@"{item.Performance / Math.Max(item.ActiveUsers, 1):N0}",
             }
         };
+
+        private class CountryName : OsuSpriteText
+        {
+            public CountryName(Country country)
+            {
+                Font = OsuFont.GetFont(size: 12, italics: true);
+                Text = $@"{country.FullName}";
```
bdach: Not too much benefit from interpolating a single property. Unless you wanted the coercion to empty string logic, at which point

```suggestion
                Text = country.FullName ?? string.Empty;
```

is probably way easier to process. The interpolation would probably get optimised out anyway but it looks very strange.

bdach (commented):
Small enough and almost correct. Not approving since it's not yet decided if this is a valid change, or we want to remove italics in web instead (pending discussion in issue)

bdach (approved):
Looks simple enough, I suppose.

smoogipoo (approved):
country names are supposed to be links and thus blue until `LinkFlowContainer` is updated to use overlay colours, but i suppose this is fine for now

PR #7827, opened 2020-02-14
Use Span for OsuColour.FromHex

turbedi:
Not a hot-path, so it shouldn't matter too much, but ¯\\_(ツ)_/¯.

Benchmark for 1000 FromHex calls with random hex strings:
|              Method |      Mean |    Error |   StdDev |   Gen 0 | Gen 1 | Gen 2 | Allocated |
|-------------------- |----------:|---------:|---------:|--------:|------:|------:|----------:|
| BenchmarkFromHexOld | 134.13 us | 0.705 us | 0.660 us | 31.0059 |     - |     - |   97544 B |
| BenchmarkFromHexNew |  80.99 us | 0.296 us | 0.262 us |       - |     - |     - |         - |

### DIFF
--- a/osu.Game/Graphics/OsuColour.cs
+++ b/osu.Game/Graphics/OsuColour.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Globalization;
 using osu.Game.Beatmaps;
 using osuTK.Graphics;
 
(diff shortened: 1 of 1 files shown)


peppy:
This is pointless

smoogipoo:
It’s fine. Probably more correct to parse rather than convert anyway.

smoogipoo (approved):

PR #7834, opened 2020-02-14
Add the ability to select multiple beatmaps in multiplayer song select

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/7833


peppy (approved):
Not 100% on testing every scenario, but looks good for now.

PR #7837, opened 2020-02-14
Use Resolved attribute instead of BackgroundDependencyLoader wherever possible

recapitalverb:


osu.Game/Online/API/APIAccess.cs
```diff
@@ -21,7 +22,9 @@ namespace osu.Game.Online.API
 {
     public class APIAccess : Component, IAPIProvider
     {
-        private readonly OsuConfigManager config;
+        [Resolved]
```
smoogipoo: This is also passed in in the ctor, and I think I'd rather keep that for now.

osu.Game/Online/Chat/ExternalLinkOpener.cs
```diff
@@ -13,15 +13,17 @@ namespace osu.Game.Online.Chat
 {
     public class ExternalLinkOpener : Component
     {
-        private GameHost host;
-        private DialogOverlay dialogOverlay;
+        [Resolved(CanBeNull = true)]
```
smoogipoo: Null-permission can be removed

osu.Game/Overlays/Settings/Sections/General/LoginSettings.cs
```diff
@@ -201,7 +201,9 @@ private class LoginForm : FillFlowContainer
             private TextBox username;
             private TextBox password;
             private ShakeContainer shakeSignIn;
-            private IAPIProvider api;
+
+            [Resolved]
```
smoogipoo: Should be nullable

bdach:
FWIW, this is another thing that could be enforced with a Roslyn analyser if that ever happens.

PR #7841, opened 2020-02-15
Update inspectcode version and fix new issues

peppy:


bdach (commented):
Looks good, although I had to manually delete the old InspectCode version from `build/tools/Addins` for `InspectCode.ps1` to pick up the new version of `ReSharper.CommandLineTools`. Not sure it's something we should be handling though.

smoogipoo (approved):

PR #7848, opened 2020-02-15
Replace hashcode override with local equality comparer

smoogipoo:
The hashcode override was only used in the de-duping process, but it causes issues with rearrangeable beatmap list.

This is probably why c# requires hashcode fields to be readonly :p

### DIFF
--- a/osu.Game/Online/Multiplayer/Room.cs
+++ b/osu.Game/Online/Multiplayer/Room.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using Newtonsoft.Json;
 using osu.Framework.Allocation;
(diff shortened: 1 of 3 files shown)


bdach:
Yeah, using mutable fields in `GetHashCode()` causes problems. Aside from failed lookups like the one this PR fixes it can also cause memory leaks, since some of the items in the underlying hashtable might be no longer reachable after insertion and mutation (since it remains in the old hash bucket in the container, but its hash is now different).

bdach (dismissed):

osu.Game/Online/Multiplayer/Room.cs
```diff
@@ -124,15 +125,13 @@ public void CopyFrom(Room other)
                 var localItem = Playlist.FirstOrDefault(i => i.BeatmapID == item.BeatmapID);
 
                 if (localItem != null)
-                {
                     item.Beatmap.Value.Metadata = localItem.Beatmap.Value.Metadata;
-                }
             }
 
-            foreach (var removeableItem in Playlist.Except(other.Playlist).ToArray())
+            foreach (var removeableItem in Playlist.Except(other.Playlist, new PlaylistEqualityComparer()).ToArray())
```
peppy: i'm thinking we just want to clear the playlist and re-add all, unless every item matches. this current logic is going to potentially change the order of the items incorrectly on an incoming change (can't happen in current scenarios, but may in the future).

i think i'd prefer to not use the equality comparer or `Except`, and just rewrite this whole section into a local linq implementation. and at that point, we can use `.Equals` and move the equality back into `PlaylistItem` itself (as `IEquatable<PlaylistItem>`).

peppy (changes requested):

osu.Game/Online/Multiplayer/Room.cs
smoogipoo: Fair call.

peppy (approved):
much nicer. we can optimise later if required.

PR #7854, opened 2020-02-15
Display a login placeholder in direct when user isn't logged in

### LINKED ISSUE #4550: osu!direct when a user isn't logged in
**Describe the feature:**
osu!direct when a user isn't logged in doesn't show anything even click search button. I think it should show something to tell the user not logged in.

I created a sample. 
Screenshot

Game4all:
# Summary

- Fixes #4550

- This PR adds a login placeholder to the osu!direct overlay through the use of an OnlineViewContainer.
Since the direct overlay inherits `SearchableListOverlay` and only online overlays derive from this class, I chose to implement it in `SearchableListOverlay` as it allows every overlay inheriting from this class to inherit the functionality (social overlay now benefits of this)

- The displayed placeholder message can be customized through the abstract `SearchableListOverlay.LoginPlaceholder` property.

- The `SearchableListOverlay` filter controls are now disabled when the user is offline.

## Todo
- [x] Make use of `OnlineViewContainer` in `SearchableListOverlay`
- [x] Disable filter input when offline

## Visual Demo

[Avalaible here](https://streamable.com/ytcej)

### DIFF
--- a/osu.Game/Overlays/DirectOverlay.cs
+++ b/osu.Game/Overlays/DirectOverlay.cs
@@ -39,6 +39,7 @@ namespace osu.Game.Overlays
 
         protected override SearchableListHeader<DirectTab> CreateHeader() => new Header();
         protected override SearchableListFilterControl<DirectSortCriteria, BeatmapSearchCategory> CreateFilterControl() => new FilterControl();
+
         protected override string LoginPlaceholder => @"Please sign in to browse direct!";
 
         private IEnumerable<BeatmapSetInfo> beatmapSets;
--- a/osu.Game/Overlays/DirectOverlay.cs
+++ b/osu.Game/Overlays/DirectOverlay.cs
@@ -39,6 +39,7 @@ namespace osu.Game.Overlays
 
         protected override SearchableListHeader<DirectTab> CreateHeader() => new Header();
         protected override SearchableListFilterControl<DirectSortCriteria, BeatmapSearchCategory> CreateFilterControl() => new FilterControl();
+        protected override string LoginPlaceholder => @"Please sign in to browse direct!";
 
         private IEnumerable<BeatmapSetInfo> beatmapSets;
 
--- a/osu.Game/Overlays/SearchableList/DisplayStyleControl.cs
+++ b/osu.Game/Overlays/SearchableList/DisplayStyleControl.cs
@@ -17,6 +17,8 @@ namespace osu.Game.Overlays.SearchableList
         public readonly SlimEnumDropdown<T> Dropdown;
         public readonly Bindable<PanelDisplayStyle> DisplayStyle = new Bindable<PanelDisplayStyle>();
 
+        public override bool HandleNonPositionalInput => !DisplayStyle.Disabled;
+
         public DisplayStyleControl()
         {
             AutoSizeAxes = Axes.Both;
@@ -53,6 +55,13 @@ namespace osu.Game.Overlays.SearchableList
             };
 
             DisplayStyle.Value = PanelDisplayStyle.Grid;
+            DisplayStyle.DisabledChanged += disabled =>
+            {
+                if (disabled)
+                    Dropdown.Current.Disabled = true;
+                else
+                    Dropdown.Current.Disabled = false;
+            };
         }
 
         private class DisplayStyleToggleButton : OsuClickableContainer
@@ -81,7 +90,12 @@ namespace osu.Game.Overlays.SearchableList
 
                 bindable.ValueChanged += Bindable_ValueChanged;
                 Bindable_ValueChanged(new ValueChangedEvent<PanelDisplayStyle>(bindable.Value, bindable.Value));
-                Action = () => bindable.Value = this.style;
+                Action = () =>
+                {
+                    if (bindable.Disabled) return;
+
+                    bindable.Value = this.style;
+                };
             }
 
             private void Bindable_ValueChanged(ValueChangedEvent<PanelDisplayStyle> e)
--- a/osu.Game/Overlays/SearchableList/DisplayStyleControl.cs
+++ b/osu.Game/Overlays/SearchableList/DisplayStyleControl.cs
@@ -55,13 +55,7 @@ namespace osu.Game.Overlays.SearchableList
             };
 
             DisplayStyle.Value = PanelDisplayStyle.Grid;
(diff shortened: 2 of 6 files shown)


osu.Game/Overlays/SearchableList/SearchableListOverlay.cs
```diff
@@ -125,5 +130,13 @@ protected override void PopOut()
...
             Filter.Search.HoldFocus = false;
         }
+
+        private class PanelContainer : OnlineViewContainer
+        {
+            public PanelContainer(string placeholderMessage)
+                : base(placeholderMessage)
+            {
+            }
+        }
```
iiSaLMaN: I thought `OnlineViewContainer` would only be used for inheritance. but after I've given this usage a couple of reads, I would argue that this container should not be abstract at all.

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -33,6 +33,7 @@ public class SocialOverlay : SearchableListOverlay<SocialTab, SocialSortCriteria
 
         protected override SearchableListHeader<SocialTab> CreateHeader() => new Header();
         protected override SearchableListFilterControl<SocialSortCriteria, SortDirection> CreateFilterControl() => new FilterControl();
+        protected override string LoginPlaceholder => @"Please sign in to browse social!";
```
iiSaLMaN: (spacing please)
```suggestion

        protected override string LoginPlaceholder => @"Please sign in to browse social!";
```

osu.Game/Overlays/DirectOverlay.cs
```diff
@@ -39,6 +39,7 @@ public class DirectOverlay : SearchableListOverlay<DirectTab, DirectSortCriteria
 
         protected override SearchableListHeader<DirectTab> CreateHeader() => new Header();
         protected override SearchableListFilterControl<DirectSortCriteria, BeatmapSearchCategory> CreateFilterControl() => new FilterControl();
+        protected override string LoginPlaceholder => @"Please sign in to browse direct!";
```
iiSaLMaN: (also here)
```suggestion

        protected override string LoginPlaceholder => @"Please sign in to browse direct!";
```

frenzibyte (changes requested):

TheWildTree:
> I chose to implement it in SearchableListOverlay as it allows every overlay inheriting from this class to inherit the functionality

`SearchableListOverlay` itself shouldn't be tied to any online-related code IMO, as it might be used for non-online stuff in the future. 

For me, you should either create a separate, online-tied version of `SearchableListOverlay` deriving from it and use that for direct / social overlays, or implement the placeholder logic directly inside of them.

The second option would be better because it also makes the placeholder able to hide the search / filter parts of the overlays, which IMO would look much better.

peppy:
I think it's fine hiding only the results area, but special attention should be given to how changes to the search criteria work.

Game4all:
> `SearchableListOverlay` itself shouldn't be tied to any online-related code IMO, as it might be used for non-online stuff in the future.

A look at `SearchableListOverlay` inheritance tree shows that it derives from `FullscrenOverlay` which in turn implements `IOnlineComponent`, which somehows indicate that it is intended to be used in an online context. So I doubt it will be used in a non-online context (please correct me if I'm wrong) thus it is pointless to me to abstract online behaviour in another class.

bdach:
I'm not too sure if we're not going to want to apply this to the `BeatmapListingOverlay` which is being developed in parallel to this instead, as I don't know if it's intended to be a replacement for direct.

peppy:
@bdach it is intended to replace the direct overlay. i'm not sure why it wasn't a refactor instead of a new class, to be honest (maybe @EVAST9919 can answer that).

that being said, as this component is likely going to disappear maybe it's a good opportunity to merge this and see how it behaves before it gets replaced?

EVAST9919:
I just wanted to replace everything at once (to make sure we still have fully working overlay) while implementing component by component.

osu.Game/Overlays/SearchableList/DisplayStyleControl.cs
```diff
@@ -53,6 +55,7 @@ public DisplayStyleControl()
             };
 
             DisplayStyle.Value = PanelDisplayStyle.Grid;
+            DisplayStyle.DisabledChanged += disabled => Dropdown.Current.Disabled = disabled;
```
peppy: i'd argue this control doesn't even need to be disabled.

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -34,6 +34,8 @@ public class SocialOverlay : SearchableListOverlay<SocialTab, SocialSortCriteria
         protected override SearchableListHeader<SocialTab> CreateHeader() => new Header();
         protected override SearchableListFilterControl<SocialSortCriteria, SortDirection> CreateFilterControl() => new FilterControl();
 
+        protected override string LoginPlaceholder => @"Please sign in to browse social!";
```
peppy: this sounds really weird.

just make it "Please sign in to view content" for all cases for now.

peppy (changes requested):

peppy:
This whole thing feels weird. The "disabled" stuff should probably not be required and the `OnlineViewContainer ` should be similar to (or share a parent class with) `ProcessingOverlay`.

osu.Game/Overlays/SearchableList/SearchableListFilterControl.cs
```diff
@@ -116,6 +119,22 @@ protected SearchableListFilterControl()
...
+                    Tabs.Current.Disabled = true;
+                    DisplayStyleControl.DisplayStyle.Disabled = true;
+                    Search.Current.Disabled = true;
+                }
+                else
+                {
+                    Tabs.Current.Disabled = false;
+                    DisplayStyleControl.DisplayStyle.Disabled = false;
+                    Search.Current.Disabled = false;
+                }
```
turbedi:
```
Tabs.Current.Disabled = disabled.NewValue;
DisplayStyleControl.DisplayStyle.Disabled = disabled.NewValue;
Search.Current.Disabled = disabled.NewValue;
```

Game4all:
I'm going to close this to work on a refactor of `OnlineViewContainer` first as I agree the container behaviour isn't as great as I imagined.

PR #7855, opened 2020-02-16
Fix transform mod not being applied correctly

### LINKED ISSUE #7851: Transform mod effects not applying
**Describe the bug:** 
~~When transform mod is selected along with any other mod the effects of transform mod do not apply and the circles do not move~~ I was mistaken and the mod seems to just not work in general
**osu!lazer version:** 
2020.215.0
**Logs:** 
[runtime.log](https://github.com/ppy/osu/files/4208896/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4208897/updater.log)
[database.log](https://github.com/ppy/osu/files/4208898/database.log)
[network.log](https://github.com/ppy/osu/files/4208899/network.log)
[performance.log](https://github.com/ppy/osu/files/4208900/performance.log)

peppy:
Closes https://github.com/ppy/osu/issues/7851.

### DIFF
--- a/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
@@ -38,6 +38,7 @@ namespace osu.Game.Rulesets.Osu.Mods
                 case DrawableSliderHead _:
                 case DrawableSliderTail _:
                 case DrawableSliderTick _:
+                case DrawableRepeatPoint _:
                     return;
 
                 default:
--- a/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics.Sprites;
 using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Osu.Objects;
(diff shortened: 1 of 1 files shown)


osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
```diff
@@ -27,26 +28,39 @@ internal class OsuModTransform : Mod, IApplicableToDrawableHitObjects
...
+
+        private void applyTransform(DrawableHitObject drawable, ArmedState state)
+        {
+            switch (drawable)
             {
-                var hitObject = (OsuHitObject)drawable.HitObject;
+                case DrawableSliderHead _:
+                case DrawableSliderTail _:
+                case DrawableSliderTick _:
+                    return;
```
bdach: Should also include `DrawableRepeatPoint` here.

bdach (commented):
Don't know if it's supremely relevant, but at a rough look `applyTransform` ends up being called twice, once on `loadComplete` of the hit object and once on hit. Adding a guard to early-return if `state != ArmedState.Idle` seems to not affect the fix but is possibly a little hacky.

osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
peppy: good catch. that's actually the one i was trying to fix with these exceptions, too...

bdach (approved):

smoogipoo (approved):

PR #7857, opened 2020-02-16
Implement BeatmapSearchTextBox component

EVAST9919:
Screenshot_1
Screenshot_2


osu.Game/Overlays/BeatmapListing/BeatmapSearchTextBox.cs
```diff
@@ -0,0 +1,76 @@
...
+            FadeEdgeEffectTo(0, fade_duration, Easing.OutQuint);
+        }
+
+        protected override Caret CreateCaret() => base.CreateCaret().With(c => ((OsuCaret)c).CaretColour = Color4.Black);
+
+        protected override Color4 SelectionColour => Color4.Gray;
+
+        private OsuSpriteText placeholder;
+
+        protected override SpriteText CreatePlaceholder() => placeholder = new OsuSpriteText
```
iiSaLMaN: https://github.com/ppy/osu-framework/blob/2c2ebf790584daca76d5a792f26717f3c6e1e4bc/osu.Framework/Graphics/UserInterface/TextBox.cs#L562

osu.Game/Graphics/UserInterface/OsuTextBox.cs
```diff
@@ -100,6 +100,12 @@ public OsuCaret()
 
             public Color4 SelectionColour { get; set; }
 
+            public Color4 CaretColour
+            {
+                get => beatSync.Colour;
+                set => beatSync.Colour = value;
+            }
```
iiSaLMaN: This would cause blending as you're changing the colour of an internal child of the caret while colour transformations below are applied to the caret, what you really want to do is something like this:
```csharp
private Color4 caretColour;

public Color4 CaretColour
{
    get => caretColour;
    set
    {
        if (caretColour == value)
            return;

        caretColour = value;
        if (!HasSelection) // move HasSelection to here.
            this.FadeColour(value, 200, Easing.Out);
    }
}

...
// in DisplayAt(position, null):
this.FadeColour(CaretColour, 200, Easing.Out); // (instead of Color4.White)
```
same should be applied for `SelectionColour` with the exception of fading colour if there's a selection.

frenzibyte (changes requested):

peppy:
Just as a heads up, I wouldn't worry too much about matching this design as it will probably change soon.

EVAST9919:
Let's wait for update then and will see

osu.Game/Overlays/BeatmapListing/BeatmapSearchTextBox.cs
```diff
@@ -0,0 +1,74 @@
...
+            FadeEdgeEffectTo(0, fade_duration, Easing.OutQuint);
+        }
+
+        protected override Caret CreateCaret() => base.CreateCaret().With(c => ((OsuCaret)c).CaretColour = Color4.Black);
+
+        protected override Color4 SelectionColour => Color4.Gray;
+
+        protected override SpriteText CreatePlaceholder() => new OsuSpriteText
+        {
+            Font = OsuFont.GetFont(weight: FontWeight.Light, italics: true),
```
bdach: Should have size 20; currently is of size 17.5:

image

Web CSS:

image

osu.Game/Overlays/BeatmapListing/BeatmapSearchTextBox.cs
```diff
@@ -0,0 +1,74 @@
...
+            Shadow = false
+        };
+
+        protected override Drawable GetDrawableCharacter(char c) => new OsuSpriteText
+        {
+            Font = OsuFont.GetFont(weight: FontWeight.Light, italics: true),
```
bdach: Should be of size 20, is of size 16:

image

bdach (changes requested):
Long text overlaps the search button, instead of staying clear with a 10px margin like web-side:

| this PR | web |
| :-: | :-: |
| image | image |

Selected text has incorrect colours and should use `OverlayColourProvider` for theming:

| this PR | web |
| :-: | :-: |
| image | image |

EVAST9919:
Currently we can't override SelectionColour with Resolved stuff such as `OsuColour` and `OverlayColourProvider` (framework fix required)

osu.Game/Overlays/BeatmapListing/BeatmapSearchTextBox.cs
EVAST9919: Fixed

osu.Game/Overlays/BeatmapListing/BeatmapSearchTextBox.cs
EVAST9919: Fixed

EVAST9919:
Also we can't adjust left side margin since we have no container to apply mask to in framework-side

bdach:
The way I see it there's not much reason for the search icon to be inside of the text box, could be beside it wrapped in a container. The way I see it the text overflowing the icon should be resolved *somehow* as it's a big UX/usability issue.

peppy:
Prefer better UX over matching design in all cases. We can tweak as a final pass.

EVAST9919:
> The way I see it there's not much reason for the search icon to be inside of the text box, could be beside it wrapped in a container. The way I see it the text overflowing the icon should be resolved _somehow_ as it's a big UX/usability issue.

I fixed the icon overlapping, I was talking about left margin when textbox is full
Screenshot_1

bdach:
Right, I misunderstood - that part I'd consider negligible.

EVAST9919:
Closing as already existing `SearchTextBox` looks better for it's usage

PR #7858, opened 2020-02-16
Implement BeatmapSearchFilter component

EVAST9919:
Screenshot_1


osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
```diff
@@ -0,0 +1,99 @@
...
+
+            public FilterTabItem(T value)
+                : base(value)
+            {
+                AutoSizeAxes = Axes.Both;
+                AddRangeInternal(new Drawable[]
+                {
+                    text = new OsuSpriteText
+                    {
+                        Font = OsuFont.GetFont(size: 13, weight: FontWeight.Regular),
```
bdach: Font size should be configurable if advanced search criteria are to be moved to the game too, as they have size 10 instead of 13:

image

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
```diff
@@ -0,0 +1,99 @@
...
+                updateState();
+            }
+
+            protected override void OnActivated() => updateState();
+
+            protected override void OnDeactivated() => updateState();
+
+            private void updateState()
+            {
+                text.Font = text.Font.With(weight: Active.Value ? FontWeight.Medium : FontWeight.Regular);
```
bdach: As far as I can tell web-side the selected item does not change font weight.

bdach (dismissed):
Component should probably also include the 100px filter header:

image

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
EVAST9919: True, but I wanted to make it more clear that an item is selected. Will remove

EVAST9919:
> Component should probably also include the 100px filter header:

I wanted to make a separate component which will have the header and the filter, but looks like there's no reason to do so

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
EVAST9919: Fixed

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
EVAST9919: Fixed

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilter.cs
```diff
@@ -0,0 +1,97 @@
...
+
+        protected override TabFillFlowContainer CreateTabFlow() => new TabFillFlowContainer
+        {
+            AutoSizeAxes = Axes.Both,
+            Spacing = new Vector2(10),
+        };
+
+        protected class FilterTabItem : TabItem<T>
+        {
+            protected virtual float TextSize() => 13;
```
bdach: A little un-C#-like to make this a method instead of a getter-only property...?

```suggestion
            protected virtual float TextSize => 13;
```

(requires fixing the override too)

bdach:
I'm not sure how adding the header here would be harder. Having two components is a bit much in my opinion, as `BeatmapSearchFilter` doesn't have too much of a reason to live on its own, so to speak. In my view it'd be better as a private class within the actual filter row component.

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilterRow.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public BeatmapSearchFilterRow(string headerName)
+        {
+            AutoSizeAxes = Axes.Y;
+            RelativeSizeAxes = Axes.X;
+            AddInternal(new FillFlowContainer
+            {
+                AutoSizeAxes = Axes.Y,
+                RelativeSizeAxes = Axes.X,
+                Spacing = new Vector2(0, 7),
```
bdach: If we're going to end up supporting that mode, should be

```suggestion
                Spacing = new Vector2(0, 5),
```

image

bdach (commented):
The 0.3 sizing mode doesn't look very good at all:

image

It's sort of what web does but on a row-by-row basis, which makes discrepancies possible and they don't look very good at all. As there are no known ways (to me) in the framework to do responsive design in the same manner CSS does, I'd just stick with the normal layout and possibly display a dropdown overflow.

Some design feedback could be nice here possibly (@arflyte?)

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilterRow.cs
EVAST9919: Hmm, I made it 7 because the header has 5px bottom margin and the filter has 2px top margin
Screenshot_1
Screenshot_2

EVAST9919:
Also just a side note: I tested it ingame and with the current values even on highest possible UI scale each row have enough space to be a single line

EVAST9919:
Screenshot_1

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilterRow.cs
bdach: As far as I can tell in devtools those margins shouldn't be added up as they both apply from the text baseline downwards, so the 5 actually overrides the 2.

EVAST9919: Screenshot_1
Looks like they don't

EVAST9919:
Well, making the filter floating is not a big deal, the main problem is combine it with a header

EVAST9919:
For now I see the dropdown as a best solution for now, will implement

EVAST9919:
Should be a good temporary solution
Screenshot_3

smoogipoo:
I think this is fine for now. Let's get this in and iterate on it later.

smoogipoo (approved):

PR #7871, opened 2020-02-17
Show skinnable test skin names and autosized component sizes

peppy:
- [x] Depends on #7870

image


bdach (approved):
Looks pretty useful 👍
I'm imagining it might be useful to be able to turn these off in some cases, but that's a maybe - that bridge is probably better crossed if we ever get to it.

smoogipoo (approved):

PR #7873, opened 2020-02-17
Make osu!mania's TestSceneDrawableJudgement use shared SkinnableTestScene

mcendu:
Prereqs:
- [x] #7870


peppy:
Have you tested this? I don't believe it will find the correct resources (they need to be added to the mania tests project).

mcendu:
Aside from the default legacy skin, all other four shows the lazer skin. None of the three custom skins contain sprites for mania `DrawableJudgement`s.

bdach:
This could perhaps use some custom resources, so 4 out of the 5 skins don't look samey, but I suppose it doesn't *regress* anything as the existing test had only two (lazer & custom) anyway.

peppy (approved):

PR #7879, opened 2020-02-17
Update osu!catch test scenes to show skinnable versions

peppy:
- Adds some new fruits test resources
- Converts a few test scenes across to show skinnable versions
- Adds accessibility to already created `SkinnableTestScene` drawables to apply further steps
- [x] Depends on https://github.com/ppy/osu/pull/7871


bdach (approved):
Got a bit spooked by the file count, but most of it is resources and the rest looks simple enough 👍

smoogipoo (approved):

bdach:
Just noticed that there doesn't seem to be any resources for bananas in `old-skin`. Should not have taken the blob file list at face value when reviewing I suppose...

PR #7881, opened 2020-02-17
Implement osu!catch fruit skinning support

### LINKED ISSUE #2224: Catch fruit rotation combo colouring doesn't match stable
It would seem that catch is supposed to go through combo colours rather than provide its own colours. Also fruit types switch on each object rather than each combo.

### LINKED ISSUE #3967: Add skinning support for osu!catch fruits
**Describe the feature:** Skinning Elements for osu!catch

**Designs:**
screenshot242
screenshot243
screenshot244

peppy:
- Closes https://github.com/ppy/osu/issues/3967
- Closes https://github.com/ppy/osu/issues/2224
- Fixes combo colour progressing incorrectly
- Adds fruit and droplet skinning
- [x] Add pulp support
- [x] Figure out why everything is double in size
- [x] Fix dynamic recompilation on `TestSceneFruitObjects`
- [x] Depends on https://github.com/ppy/osu/pull/7879
- [x] Depends on https://github.com/ppy/osu/pull/7896
- [x] Depends on https://github.com/ppy/osu/pull/7898


peppy:
This is now ready to go. Still some more changes to come but they will be PR'd separately.

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -221,9 +221,9 @@ private void beginTrail()
             /// Add a caught fruit to the catcher's stack.
             /// </summary>
             /// <param name="fruit">The fruit that was caught.</param>
-            public void Add(DrawableHitObject fruit)
+            public void PlaceOnPlate(DrawableHitObject fruit)
             {
-                float ourRadius = fruit.DrawSize.X / 2 * fruit.Scale.X;
+                float ourRadius = ((DrawableCatchHitObject)fruit).DisplayRadius;
```
bdach: Looks like the arg type can be scoped down to `DrawableCatchHitObject` and the cast removed without any other alterations.

bdach (commented):
I had a quick look through the diff. Looks mostly sane, although I didn't really pay much attention to the various sizing changes and adjustments that are kind of all over the place and hard to keep track of on a read-through alone. I might give this a second re-review over the sizings themselves when the prereqs come in (if I can make it in time).

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
peppy: nice catch

osu.Game.Rulesets.Catch/Skinning/CatchLegacySkinTransformer.cs
```diff
@@ -0,0 +1,58 @@
...
+                return null;
+
+            switch (catchSkinComponent.Component)
+            {
+                case CatchSkinComponents.FruitApple:
+                case CatchSkinComponents.FruitBananas:
+                case CatchSkinComponents.FruitOrange:
+                case CatchSkinComponents.FruitGrapes:
+                case CatchSkinComponents.FruitPear:
+                    var lookupName = catchSkinComponent.Component.ToString().Underscore().Hyphenate();
```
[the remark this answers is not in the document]
peppy: Oh interesting. I saw this but must have misread the documentation (a lot of the other methods expect input to already be underscores to actually work).

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
```diff
@@ -90,7 +90,7 @@ protected override void ApplyDefaultsToSelf(ControlPointInfo controlPointInfo, B
 
             TimePreempt = (float)BeatmapDifficulty.DifficultyRange(difficulty.ApproachRate, 1800, 1200, 450);
 
-            Scale = 1.0f - 0.7f * (difficulty.CircleSize - 5) / 5;
+            Scale = (1.0f - 0.7f * (difficulty.CircleSize - 5) / 5) / 2;
```
bdach: I'm not sure why this is halved, but my working hypothesis is "it just needs to be".

osu.Game.Rulesets.Catch/Skinning/CatchLegacySkinTransformer.cs
```diff
@@ -0,0 +1,58 @@
...
+                case CatchSkinComponents.FruitPear:
+                    var lookupName = catchSkinComponent.Component.ToString().Kebaberize();
+                    if (GetTexture(lookupName) != null)
+                        return new LegacyFruitPiece(lookupName);
+
+                    break;
+
+                case CatchSkinComponents.Droplet:
+                    if (GetTexture("fruit-drop") != null)
+                        return new LegacyFruitPiece("fruit-drop") { Scale = new Vector2(0.8f) };
```
bdach: I assume the 0.8 here is "for stable reasons" too.

bdach (commented):
I looked at the visuals and spacings some more. I noticed some things that I'm going to point out below, but I'm aware they might be deliberate changes.

In general the lazer skin fruits have changed some (not hugely). Not sure if it's an actual issue that needs to be addressed. Don't think it looks any better or worse than before, just mostly _slightly different_. Objects are slightly larger (including juice streams) and the pulps are more spaced out. Screenshots:

| `master` | this PR |
| :-: | :-: |
| image | image |

In general this looks mergeable to me but I'm not approving since I have zero clue about what stable does or doesn't do. Feel free to ignore my questions too.

osu.Game.Rulesets.Catch/Skinning/CatchLegacySkinTransformer.cs
peppy: yep.

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
peppy: it's now 1:1 with the osu! ruleset implementation, but yeah, "because it needs to be" is about right.

osu.Game.Rulesets.Catch/Objects/Drawable/DrawableDroplet.cs
```diff
@@ -2,32 +2,30 @@
...
         private void load()
         {
-            AddInternal(pulp = new Pulp { Size = Size });
-
-            AccentColour.BindValueChanged(colour => { pulp.AccentColour = colour.NewValue; }, true);
+            ScaleContainer.Child = new SkinnableDrawable(
+                new CatchSkinComponent(CatchSkinComponents.Droplet), _ => new Pulp
+                {
+                    Size = Size / 4,
+                    AccentColour = { Value = Color4.White }
```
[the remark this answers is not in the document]
peppy: this is actually a mistake

osu.Game.Rulesets.Catch/Objects/Drawable/DrawableCatchHitObject.cs
```diff
@@ -15,11 +19,34 @@ public abstract class PalpableCatchHitObject<TObject> : DrawableCatchHitObject<T
     {
         public override bool CanBePlated => true;
 
+        protected Container ScaleContainer;
```
smoogipoo:
```suggestion
        protected Container ScaleContainer { get; private set; }
```

osu.Game.Rulesets.Catch/Objects/Drawable/DrawableTinyDroplet.cs
```diff
@@ -10,7 +10,12 @@ public class DrawableTinyDroplet : DrawableDroplet
         public DrawableTinyDroplet(TinyDroplet h)
             : base(h)
         {
-            Size = new Vector2((float)CatchHitObject.OBJECT_RADIUS) / 8;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Scale /= 2;
```
smoogipoo:
```suggestion
            ScaleContainer.Scale /= 2;
```

smoogipoo (changes requested):

smoogipoo (approved):
lgtm

PR #7888, opened 2020-02-18
Fix osu!catch fruit exploding/dropping multiple timed is skin is changed during explode animation

### LINKED ISSUE #7880: Changing skin while playing osu!catch results in stuck fruit
The catcher is likely getting incorrectly recycled, leaving fruit unsure what to do.

peppy:
`updateState` is forcefully called on skin change. This would clear our custom explode or drop transforms, so instead we should use the `ApplyCustomUpdateState` mechanism to ensure they are reapplied correctly.

- Closes https://github.com/ppy/osu/issues/7880.


osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -425,10 +418,26 @@ public void Explode()
...
 
+                removeFromPlatWithTransform(fruit, f =>
+                {
+                    f.MoveToY(f.Y - 50, 250, Easing.OutSine).Then().MoveToY(f.Y + 50, 500, Easing.InSine);
+                    f.MoveToX(f.X + originalX * 6, 1000);
+                    f.FadeOut(750);
+                });
+            }
+
+            private void removeFromPlatWithTransform(DrawableHitObject fruit, Action<DrawableHitObject> action)
```
smoogipoo: `removeFromPlateWithTransform`

smoogipoo (changes requested):

smoogipoo (approved):

PR #7893, opened 2020-02-18
Refactor RankingsOverlay to improve edge-case conditions

EVAST9919:
Fixes:
- No disposal handling
- Null ruleset exception on first `Show()` call


smoogipoo (approved):
Was a little concerned about the removal of the `, true)` on the bindings, but they're only internal so it's fine.

PR #7896, opened 2020-02-19
Move DrawableHitObject accent colour update duties to ruleset

peppy:
As it turns out, only the osu! ruleset uses traditional combo colour mapping. In order to allow osu!catch to work correctly, it needs to override the previously forced `AccentColours`.

While none of our current rulesets use the same combo colour (based on combo index), there is a high chance that custom rulesets may want to. Rather than having them reimplement, they can choose to use the `proposedColour` or not.

### Custom ruleset DrawableHitObject implementations should manage their own AccentColour

If you were relying on osu! doing this for you until now, the fix is simple:

```csharp
public class DrawableAwesomeHitObject : DrawableHitObject<AwesomeHitObject>
{
    ...

    protected override void UpdateComboColour(Color4 proposedColour, IReadOnlyList<Color4> comboColours) => AccentColour.Value = proposedColour;
}
```


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -336,23 +336,32 @@ protected sealed override void SkinChanged(ISkinSource skin, bool allowFallback)
...
         }
 
-        private void updateAccentColour()
+        private void updateComboColour()
         {
             if (HitObject is IHasComboInformation combo)
             {
                 var comboColours = CurrentSkin.GetConfig<GlobalSkinColours, IReadOnlyList<Color4>>(GlobalSkinColours.ComboColours)?.Value;
-                AccentColour.Value = comboColours?.Count > 0 ? comboColours[combo.ComboIndex % comboColours.Count] : Color4.White;
+                UpdateComboColour(comboColours?.Count > 0 ? comboColours[combo.ComboIndex % comboColours.Count] : Color4.White, comboColours);
```
bdach: The fallback to `Color4.White` in the case of null/no combo colours could be a problem if someone wanted to provide own colours only in those cases. Doing this makes it so "there are no colours to suggest" and "the colour white is suggested" are indistinguishable to the ruleset developer. Would go nullable there.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -336,23 +336,32 @@ protected sealed override void SkinChanged(ISkinSource skin, bool allowFallback)
...
         }
 
+        /// <summary>
+        /// Called when a combo colour change is proposed.
+        /// </summary>
+        /// <param name="proposedColour">The proposed combo colour, based off the combo index.</param>
+        /// <param name="comboColours">A list of combo colours provided by the beatmap or skin. Can be null if not available.</param>
+        protected virtual void UpdateComboColour(Color4 proposedColour, IReadOnlyList<Color4> comboColours)
+        {
+        }
```
bdach: It's probably because I don't fully know where you plan taking this forward, but I'm not feeling this API too much. I'd probably prefer a pure function myself (return `Color4` and then assign it where it needs to go).

I understand *why* the default implementation is empty (if it fell back to the old behaviour, there'd be no way of undoing it), and I'm guessing it's done like this to allow ruleset creators to change more than just one accent colour...?

peppy: we could feasibly leave the default implementation in there and cancel it out with no-base-call, although not sure how i feel about that since we try to avoid relying on base calls for externally exposed APIs.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
peppy: sounds fine too, will wait for confirmation from @smoogipoo.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: The proposal is to make this return a Color4.

```csharp
DHO:
    protected virtual Color4 GetComboColour(IReadOnlyList<Color4> comboColours) => comboColours?.Count > 0 ? comboColours[combo.ComboIndex % comboColours.Count] : Color4.White;

DCHO:
    protected override Color4 GetComboColour(IReadOnlyList<Color4> comboColours) => comboColour?.Count > 0 ? comboColours[hitObject.IndexInBeatmap % comboColours.Count] : Color4.White;
```

Though as I mentioned yesterday, the whole `AccentColour` stuff feels super shocking to me. It's publicly exposed and because of that it loses all meaning in the context of an `IHasComboInformation` object - whatever accent colour you set on a `DHO` may or may not get replaced by the combo colours based on certain variables including timing.

So based on that it is my opinion that this shouldn't be a pure method unless `AccentColour` itself was only privately mutable, and that has its own issues to resolve too such as how do you set the accent colour of mania hitobjects, what should the value of the exposed colour be if there is no combo, or should it even _exist_ inside `DHO` (which again, I mentioned yesterday).

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: Agree with making nullable.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
peppy: I'm also fine with moving `AccentColour` into each ruleset if we decide that's the correct direction.

bdach: Good point, the fact of `AccentColour` being public makes my proposal entirely unviable.

Separating per ruleset would be probably pretty good, as it eliminates the problem of trying to come up with a good abstraction for everything. I'm not even sure how a generic solution for all rulesets could look like and I'd probably have to sit down for a couple of hours to see if one is even viable.

Not sure it's something that needs to be done right now, though. Could roll with this as a solution for now until it's no longer tenable.

smoogipoo: Let's go with that then, make it a pure method and then we can consider moving it out into a `DrawableHitObjectWithComboColour` at a later point. Important part is getting the API into a steady state, because anything further from there won't be too hard to change.

smoogipoo (approved):
This is fine as a temporary measure

PR #7897, opened 2020-02-19
Fix indices in beatmap not being transferred to children

peppy:


bdach:
The propagation to children looks correct, but I'm not seeing the off-by-one in the old version...?

image

Mind you, I wholeheartedly agree that it *is* confusing and that's why I avoid postfix increment/decrement like the plague.

peppy:
as it turns out, the off by one fix here here https://github.com/ppy/osu/pull/7881/files#diff-91c1076d523ad5caeadd9641c8596b48R51

bdach (approved):

smoogipoo (approved):

PR #7899, opened 2020-02-19
Refactor overlined displays to support different sizing modes

smoogipoo:
Going forward we'll need these components (particularly the participants list) to support x-axis autosize.

This is maybe going a bit overboard all in the name of code-reuse, but it works well.


osu.Game/Screens/Multi/Components/OverlinedDisplay.cs
```diff
@@ -83,5 +94,25 @@ private void load(OsuColour colours)
...
+
+            grid.AutoSizeAxes = Axes.None;
+            grid.RelativeSizeAxes = Axes.None;
+            grid.AutoSizeAxes = AutoSizeAxes;
+            grid.RelativeSizeAxes = ~AutoSizeAxes;
+
+            Content.AutoSizeAxes = Axes.None;
+            Content.RelativeSizeAxes = Axes.None;
+            Content.AutoSizeAxes = grid.AutoSizeAxes;
+            Content.RelativeSizeAxes = grid.RelativeSizeAxes;
```
bdach: Maybe worth mentioning here (as a comment) that the `None` assignments aim to clear any sizings to make sure no exceptions about invalid sizing combinations get thrown?

Also could do compound assignment here:
```csharp
grid.AutoSizeAxes = grid.RelativeSizeAxes = Axes.None;
```
which maybe makes the intent a bit clearer but breaks the nice symmetry so it's mostly personal opinion.

bdach (changes requested):
`TestSceneOverlinedPlaylist` needs a relative size specification:
```csharp
                Child = new OverlinedPlaylist(false)
                {
                    RelativeSizeAxes = Axes.Both
                }
```
as it's 0x0 right now.

peppy (approved):

PR #7901, opened 2020-02-19
Add osu!catch droplet rotation animation

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/7881 for structural changes.


bdach:
Is there a way to visually check how this looks? I don't see any as far as I can see (test scenes use default circular sprite which is kind of hard to be able to tell if this works by, as it's perfectly symmetrical in regards to rotation; can't see any with a skin).

peppy:
Will be visible with resources update on legacy skin.

peppy:
...and with now present test steps

smoogipoo (dismissed):

smoogipoo (approved):

PR #7902, opened 2020-02-19
Rename drawable namespace to avoid clashes with framework class

peppy:
In line with other three rulesets.

- [x] Depends on https://github.com/ppy/osu/pull/7881 for structural changes.


bdach (commented):
Looks like an obvious merge after the prereq.

smoogipoo (approved):

PR #7909, opened 2020-02-19
Add a global multiplayer background header

smoogipoo:
I'm not too hot on my handling of the screen change events, but I also don't want to really want to make it too complicated while there are only 2 screens to work with.

(along with https://github.com/ppy/osu/pull/7900):

Lounge:
image

Match:
image


osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -62,43 +65,68 @@ public class Multiplayer : OsuScreen, IOnlineComponent
...
+                                    {
+                                        RelativeSizeAxes = Axes.Both,
+                                        Width = 1.25f,
+                                        Masking = true,
+                                        Children = new Drawable[]
+                                        {
+                                            new HeaderBackgroundSprite
+                                            {
+                                                RelativeSizeAxes = Axes.X,
+                                                Height = 400 // Keep a static height so the header doesn't change as it's resized
```
bdach: Annoying comment suggestion, but alas:
```suggestion
                                                Height = 400 // Keep a static height so the header doesn't change as it's resized between subscreens
```
IMO a few too many mental shortcuts in the current version as I got confused a smidge when looking at what this is doing in-game.

bdach (commented):
Don't think the switching logic is that ugly myself - not the prettiest, sure, but probably good enough for the foreseeable future.

PR #7911, opened 2020-02-19
Split out pulp formations into individual classes

### LINKED ISSUE #2790: Construction of catch drawable objects should be moved to separate classes
Currently all the catch hitobjects are created in one common `DrawableFruit.createPulp()` method. This is backwards to everything else in the entire game. Instead of this, there should be separate classes e.g.:

`DrawableRaspberry`
`DrawablePineapple`
`DrawableGrape`
`DrawableBanana`

With each class defining its own hierarchy.

peppy:
- Closes https://github.com/ppy/osu/issues/2790
- [x] Depends on https://github.com/ppy/osu/pull/7902


bdach (commented):
Also looks simple enough to be good to go after prereqs (both visually and code-wise).

smoogipoo (approved):

PR #7912, opened 2020-02-19
Implement BeatmapListingOverlay

EVAST9919:
For now only in tests

### DIFF
--- a/osu.Game/Online/API/Requests/SearchBeatmapSetsResponse.cs
+++ b/osu.Game/Online/API/Requests/SearchBeatmapSetsResponse.cs
@@ -2,12 +2,20 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using Newtonsoft.Json;
 using osu.Game.Online.API.Requests.Responses;
 
 namespace osu.Game.Online.API.Requests
 {
     public class SearchBeatmapSetsResponse : ResponseWithCursor
     {
+        [JsonProperty("beatmapsets")]
         public IEnumerable<APIBeatmapSet> BeatmapSets;
+
+        [JsonProperty("error")]
+        public string Error;
+
+        [JsonProperty("total")]
+        public int Total;
     }
 }
--- a/osu.Game/Online/API/Requests/SearchBeatmapSetsResponse.cs
+++ b/osu.Game/Online/API/Requests/SearchBeatmapSetsResponse.cs
@@ -12,9 +12,6 @@ namespace osu.Game.Online.API.Requests
         [JsonProperty("beatmapsets")]
         public IEnumerable<APIBeatmapSet> BeatmapSets;
 
-        [JsonProperty("error")]
-        public string Error;
-
         [JsonProperty("total")]
         public int Total;
     }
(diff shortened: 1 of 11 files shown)


osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -0,0 +1,288 @@
...
+            var sortCriteria = sortControl.Current;
+            var sortDirection = sortControl.SortDirection;
+
+            searchSection.Query.BindValueChanged(query =>
+            {
+                sortCriteria.Value = string.IsNullOrEmpty(query.NewValue) ? DirectSortCriteria.Ranked : DirectSortCriteria.Relevance;
+                sortDirection.Value = SortDirection.Descending;
+
+                Scheduler.AddOnce(updateSearch);
+            });
```
bdach: This querying should be debounced (see: current direct implementation). The way you're currently doing it, you're pretty much sending a new request per each keystroke of the user, which is going to end up with a lot of cancelled requests at best, or end up hitting the server with multiple requests per single query at worst, killing performance.

The actual search should be delayed slightly.

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -0,0 +1,288 @@
...
+                sortControl.SortDirection.Value);
+
+            getSetsRequest.Success += response => Schedule(() => recreatePanels(response));
+
+            API.Queue(getSetsRequest);
+        }
+
+        private void recreatePanels(SearchBeatmapSetsResponse response)
+        {
+            var hasError = !string.IsNullOrEmpty(response.Error);
```
bdach: As far as I can tell from web sources [this is impossible](https://github.com/ppy/osu-web/blob/a02e1cb7dc352f3ee40fecb220d135d726140747/app/Http/Controllers/BeatmapsetsController.php#L106-L111). A non-null `error` in the response will throw a 504 error code, which means it won't raise a `Success` event, but a `Fail` one.

Mind you, arguably the field shouldn't be there at all, but that's for web to handle.

osu.Game/Online/API/Requests/SearchBeatmapSetsResponse.cs
```diff
@@ -2,12 +2,20 @@
...
 
 namespace osu.Game.Online.API.Requests
 {
     public class SearchBeatmapSetsResponse : ResponseWithCursor
     {
+        [JsonProperty("beatmapsets")]
         public IEnumerable<APIBeatmapSet> BeatmapSets;
+
+        [JsonProperty("error")]
+        public string Error;
```
bdach: As per the other comment, should be removed, as the contents of this will never be relevant if the response is successful.

bdach (dismissed):

smoogipoo:
A few annoyances here:
1. There's a flicker when new panels are loaded in: https://drive.google.com/open?id=1FAVjdCMbJlq648Zwax4KrYIkp4MsLq4x
2. The default category should be "Has Leaderboard" (can probably be addressed in a future PR).

smoogipoo:
I've adjusted the height transition to be instant, since the autosize duration felt weird.

smoogipoo (approved):
lgtm

PR #7913, opened 2020-02-19
Highlight max combo on beatmap leaderboards

### LINKED ISSUE #7862: Max combo's aren't green highlighted when someone has a full combo on it

lazer:
osu!_DfZGXP1u0U
web:
chrome_fKY8i83t8i

**osu!lazer version:** 
commit e80c310ab29030fd4d6c5d300a98f8f51c68067c

**Used beatmap:** https://osu.ppy.sh/beatmapsets/999868#osu/2091238

thewildtree:
Fixes #7862 

Made `MaxCombo` nullable because there were situations where `max_combo` returned from the server was null and caused the beatmap overlay to load infinitely.

dotnet_2020-02-19_17-53-26

### DIFF
--- a/osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
+++ b/osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
@@ -146,7 +146,8 @@ namespace osu.Game.Overlays.BeatmapSet.Scores
                 new OsuSpriteText
                 {
                     Text = $@"{score.MaxCombo:N0}x",
-                    Font = OsuFont.GetFont(size: text_size)
+                    Font = OsuFont.GetFont(size: text_size),
+                    Colour = score.MaxCombo == score.Beatmap.MaxCombo ? highAccuracyColour : Color4.White
                 }
             });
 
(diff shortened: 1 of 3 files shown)


osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -146,7 +146,8 @@ private Drawable[] createContent(int index, ScoreInfo score)
                 new OsuSpriteText
                 {
                     Text = $@"{score.MaxCombo:N0}x",
-                    Font = OsuFont.GetFont(size: text_size)
+                    Font = OsuFont.GetFont(size: text_size),
+                    Colour = score.MaxCombo == score.Beatmap.MaxCombo ? highAccuracyColour : Color4.White
```
bdach: Better null handling is required (as evidenced by tests failing).

bdach (changes requested):

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
TheWildTree: woops, fixed

bdach (approved):
At a closer look `TestSceneScoreContainer` should probably provide the previously-null beatmap, but since API docs are incomplete I can't even really tell if the beatmap is mandatory or not in the actual web responses without digging in PHP sources (which is risky at best for me).

Maybe better to stay on the safe side for now and potentially clean up when `ScoreContainer` is fixed up.

smoogipoo (approved):
Hmm... I'm a little suspicious of the null still, but I've confirmed with the osu-web team and it can be null if it's awaiting difficulty processing, etc.

Issue #7916, opened 2020-02-19
Decouple ScoresContainer from API


bdach:
`ScoresContainer` is too tightly coupled to API, rendering `TestSceneScoresContainer` unusable without letting it fire off requests. It should be able to display content without.

Looks like it might be in order for a general refactor as well since this many fields is probably too many and on some of them I have zero clue as to what they're supposed to do or be (`background`, `content`):

![image](https://user-images.githubusercontent.com/20418176/74863346-0b475480-534e-11ea-9e61-4a2a8aaf8c65.png)

PR #7919, opened 2020-02-20
Support null leaderboard position

smoogipoo:
Can be merged before, but not after a deploy with https://github.com/ppy/osu-web/pull/5645 .


osu.Game/Overlays/BeatmapSet/Scores/ScoresContainer.cs
```diff
@@ -70,7 +71,10 @@ protected APILegacyScores Scores
                 topScoresContainer.Add(new DrawableTopScore(topScore));
 
                 if (userScoreInfo != null && userScoreInfo.OnlineScoreID != topScore.OnlineScoreID)
```
peppy: should handle null (is local user top score which could have null pos)

smoogipoo:
Done, and I've added a test but it can't really be tested right now since the test scene is broken (https://github.com/ppy/osu/issues/7916). I've tested in-game though:

image

peppy (approved):

PR #7922, opened 2020-02-20
Improve visual appearance of ProcessingOverlay

peppy:
Previously it would dim a full area, which could sometimes look very bad (multiplayer lounge room display). Now a dim target can be chosen, allowing only dimming content when more appropriate.

image

Also adds a test scene for this component.

### DIFF
--- a/osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
+++ b/osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
@@ -80,7 +80,7 @@ namespace osu.Game.Graphics.UserInterface
 
             if (State.Value == Visibility.Visible)
             {
-                // ensure we don't leave the targetin a bad state.
+                // ensure we don't leave the target in a bad state.
                 dimTarget?.FadeColour(Color4.White, transition_duration);
             }
         }
--- a/osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
+++ b/osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
@@ -6,20 +6,27 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input.Events;
+using osuTK;
 using osuTK.Graphics;
 
 namespace osu.Game.Graphics.UserInterface
 {
     /// <summary>
-    /// An overlay that will consume all available space and block input when required.
+    /// An overlay that will show a loading overlay and completely block input to an area.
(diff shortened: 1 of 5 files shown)


osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
```diff
@@ -28,29 +35,54 @@ private void load()
...
+            dimTarget?.FadeColour(Color4.White, transition_duration, Easing.OutQuint);
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (State.Value == Visibility.Visible)
+            {
+                // ensure we don't leave the targetin a bad state.
```
bdach: classic comment nitpick
```suggestion
                // ensure we don't leave the target in a bad state.
```

bdach (commented):
That's a horrible diff for reading, but it's github's fault. It doesn't fare well with indent changes, Rider was much better. Code looks good save for a classic annoying comment suggestion.

As a general note I don't know how I feel about the various loading animations we have in the game on a design language level. We've got a couple of slightly differing variants (this - spinner in a solid box, `DimmedLoadingLayer` - spinner in front of a, well, dimmed layer) and I'm not feeling the consistency too much.

That's something that can be iterated on I guess, and also probably requires design feedback from a designer (which I am not, and my attempts at summoning one have had little effect up to now 😛)

osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
```diff
@@ -28,29 +35,54 @@ private void load()
...
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (State.Value == Visibility.Visible)
+            {
+                // ensure we don't leave the target in a bad state.
+                dimTarget?.FadeColour(Color4.White, transition_duration);
```
smoogipoo: Should be `Easing.OutQuint` no?

smoogipoo (changes requested):

osu.Game/Graphics/UserInterface/ProcessingOverlay.cs
peppy: ah, good point. i though this wasn't important as we are already disposed but that logic is back-to-front

peppy:
@bdach yeah, we are still waiting on a more custom loading animation, but for the time being i *think* we only have two display variants. This one should probably be preferred going forward.

smoogipoo (approved):

PR #7924, opened 2020-02-20
Implement sorting by genre and language in BeatmapListingOverlay

EVAST9919:


osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -16,15 +16,19 @@ public class SearchBeatmapSetsRequest : APIRequest<SearchBeatmapSetsResponse>
         private readonly BeatmapSearchCategory searchCategory;
         private readonly DirectSortCriteria sortCriteria;
         private readonly SortDirection direction;
+        private readonly BeatmapSearchGenre genre;
+        private readonly BeatmapSearchLanguage language;
         private string directionString => direction == SortDirection.Descending ? @"desc" : @"asc";
 
-        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
+        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending, BeatmapSearchGenre genre = BeatmapSearchGenre.Any, BeatmapSearchLanguage language = BeatmapSearchLanguage.Any)
```
bdach: 7 params is getting a bit much. My ultrawide monitor has the horizontal scrollbar of half the size of the window when this is open. Lines should be split at least, but I'd _really_ advocate passing in an actual structure here:

```csharp
public class SearchBeatmapSetsParameters
{
    public string Query { get; set; }
    public RulesetInfo Ruleset { get; set; }
    public BeatmapSearchCategory SearchCategory { get; set; }
    public DirectSortCriteria SortCriteria { get; set; }
    public SortDirection SortDirection { get; set; }
    public BeatmapSearchGenre Genre { get; set; }
    public BeatmapSearchLanguage Language { get; set; }

    public BeatmapSetSearchParameters(string query, RulesetInfo ruleset)
    {
        Query = query;
        Ruleset = ruleset;

        SearchCategory = BeatmapSearchCategory.Any;
        SortCriteria = DirectSortCriteria.Ranked;
        SortDirection = SortDirection.Descending;
        Genre = BeatmapSearchGenre.Any;
        Language = BeatmapSearchLanguage.Any;
    }
}
```
Some of these initialisers might be a bit gratuitous as the default would be the same but probably better to be explicit than not.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
...
         Mine,
     }
+
+    public enum BeatmapSearchGenre
+    {
+        Any,
+        Unspecified,
+
+        [Description("Video Game")]
+        Game,
```
bdach: I don't think the names of these enums matter when mapping requests, so I'd just keep casing as consistent with the descriptions as possible.
```suggestion
        VideoGame,
```

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
...
+        Anime,
+        Rock,
+        Pop,
+        Other,
+        Novelty,
+
+        [Description("Hip Hop")]
+        Hiphop = 9,
```
bdach:
```suggestion
        HipHop = 9,
```

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
         [Description("My Maps")]
         Mine,
     }
+
+    public enum BeatmapSearchGenre
```
bdach: Seeing the gap at id 8 I'd probably prefer this enum to have explicit values assigned everywhere instead of just for 9.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
...
+        French = 7,
+        German,
+        Italian = 11,
+        Japanese = 3,
+        Korean = 6,
+        Spanish = 10,
+        Swedish = 9,
+        Instrumantal = 5,
+        Other = 1
+    }
```
bdach: This looks super messy and in general prone to error (I had to stop and think why `German` had no value - it's because by accident it's after `French` which have both adjacent `language_id` and `display_order` web-side).

I'd rather [follow what web does entirely](https://github.com/ppy/osu-web/blob/ef8964813fa1d30e97cde9f0e7e117649b5cd9f4/database/seeds/ModelSeeders/MiscSeeder.php#L51-L64) by ordering _by value_ in the enum, and then adding some mechanism for display order (probably best to mirror directly and implement an `[Order(int)]` attribute).

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
...
+        [Description("Hip Hop")]
+        Hiphop = 9,
+        Electronic
+    }
+
+    public enum BeatmapSearchLanguage
+    {
+        Any,
+        English = 2,
+        Chilnese = 4,
```
bdach:
```suggestion
        Chinese = 4,
```

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +73,38 @@ public enum BeatmapSearchCategory
...
+        French = 7,
+        German,
+        Italian = 11,
+        Japanese = 3,
+        Korean = 6,
+        Spanish = 10,
+        Swedish = 9,
+        Instrumantal = 5,
```
bdach:
```suggestion
        Instrumental = 5,
```

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -154,10 +154,60 @@ protected override void LoadComplete()
 
             searchSection.Ruleset.BindValueChanged(_ => queueUpdateSearch());
             searchSection.Category.BindValueChanged(_ => queueUpdateSearch());
+            searchSection.Genre.BindValueChanged(_ => queueUpdateSearch());
+            searchSection.Language.BindValueChanged(_ => queueUpdateSearch());
             sortCriteria.BindValueChanged(_ => queueUpdateSearch());
             sortDirection.BindValueChanged(_ => queueUpdateSearch());
```
bdach: I'm not sure if it isn't time to make this a separate component. Could also leverage the proposed `SearchBeatmapSetsParameters` class to externally expose just that one bindable and then bind to it here and fire requests at will, encapsulating all of the search fields in the new component.

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -154,10 +154,60 @@ protected override void LoadComplete()
...
+            var currentQuery = searchSection.Query.Value;
+
+            if (currentQuery != tag)
+            {
+                setDefaultSearchValues();
+                searchSection.Query.Value = tag;
+            }
+
+            Show();
+        }
```
bdach: I'm not sure what the point of these is either. This is another place that could be vastly simplified with the introduction of `SearchBeatmapSetsParameters`.

bdach (changes requested):

EVAST9919:
@bdach Everything has been addressed. (But not sure about quality of enum attributes, working with them for the first time)

bdach:
Structurally alone, the changes done here are not enough for me to feel comfortable approving a merge of this. In particular the biggest pain points for me are:

* the request parameters being half in `BeatmapListingSearchSection`, half in `BeatmapListingSortTabControl`,
* the sheer fact that `{Set,Show}{Tag,Genre,Language}` exist (are we going to end up adding those sort of methods and passing data through two layers for each possible param we add further down the line?)
* the cascade of `BindValueChanged`s in `BeatmapListingSearchSection` recreating the whole param structure each time.

I don't think it's immediately fixable in this PR. The ideal solution I'm picturing is something like depicted on the following diagram.

Untitled Diagram(5)

Biggest upsides for me:

* centralises parameter collection into one component,
* removes *all* online logic from `BeatmapListingOverlay` (the overlay only has flowing data in it, the component encapsulates online),
* therefore immediately making `BeatmapListingOverlay` testable without online (all you need to test it offline is to have a mock that will subscribe to bindable changes and spit out `MatchingBeatmaps`),
* allows setting of individual parameters (by having them exposed as protected, or heck, even public maybe, in the components closest to them) - but at that point that might not even be necessary for anything anymore.

Might need a second opinion of the core team on this one since it's a non-trivial architectural proposition.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -62,4 +81,78 @@ public enum BeatmapSearchCategory
...
+        public OrderAttribute(int order)
+        {
+            Order = order;
+        }
+    }
+
+    [AttributeUsage(AttributeTargets.Enum)]
+    public class HasOrderedElementsAttribute : Attribute
+    {
+    }
```
bdach: Does not belong here & might be useful elsewhere later. I'd move to somewhere common - `osu.Game.Utils` maybe?

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilterRow.cs
```diff
@@ -79,9 +81,30 @@ public BeatmapSearchFilter()
...
                 {
-                    foreach (var val in (T[])Enum.GetValues(typeof(T)))
+                    if (Attribute.GetCustomAttribute(type, typeof(HasOrderedElementsAttribute)) != null)
+                    {
+                        var enumValues = Enum.GetValues(type).Cast<T>().ToArray();
+                        var enumNames = Enum.GetNames(type);
+
+                        int[] enumPositions = Array.ConvertAll(enumNames, n =>
+                        {
+                            var orderAttr = (OrderAttribute)type.GetField(n).GetCustomAttributes(typeof(OrderAttribute), false)[0];
```
bdach: Could get all fields at once to save on reflection runtime. If an `OrderAttribute` is missing the numeric enum value could be used to order.

osu.Game/Overlays/BeatmapListing/BeatmapSearchFilterRow.cs
```diff
@@ -79,9 +81,30 @@ public BeatmapSearchFilter()
...
+                            return orderAttr.Order;
+                        });
+
+                        Array.Sort(enumPositions, enumValues);
```
bdach: You're silently relying on the fact that `enumPositions` has items in the same order as `enumValues`, which as far as I can see is not documented.

bdach: The whole sorting should be a global helper to go along with `OrderAttribute`.

bdach (changes requested):
In addition to the above:

EVAST9919:
Refactored with proposed logic in mind (hopefully I understood it correctly).
Removed methods which out-of-scope of this pr.

peppy:
I've tidied up the remaining review issues but not yet reviewed the full PR. @smoogipoo can you take a look at just my last commit and make sure you're okay with this direction before we go further?

bdach:
FWIW, on a rapid skim-through that's pretty much exactly the way I envisioned the attribute code should look like.

osu.Game/Utils/OrderAttribute.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            if (Attribute.GetCustomAttribute(type, typeof(HasOrderedElementsAttribute)) == null)
+                return items;
+
+            return items.OrderBy(i =>
+            {
+                if (type.GetField(i.ToString()).GetCustomAttributes(typeof(OrderAttribute), false).FirstOrDefault() is OrderAttribute attr)
+                    return attr.Order;
+
+                return 0;
```
peppy, line 32: should probably throw

smoogipoo (commented):
I generally disagree with the notion of ordering an enum value, for similar reasons that I disagree with humanizing enum values for webrequest purposes.

I'll bring it up again if this bites us, because it's pretty self-contained as it is.

smoogipoo (dismissed):

peppy (approved):
i think this can suffice for now...

PR #7951, opened 2020-02-22
Add CommentsContainer to BeatmapSetOverlay

### LINKED ISSUE #7943: Show comments in beatmap overlay
In case it was forgotten

EVAST9919:
Closes https://github.com/ppy/osu/issues/7943


bdach:
I made one small change (the `.HasValue` check was a little redundant, could have just negated it instead of doing `!= true` since it's a `bool`). I replaced it with pattern matching; seems good otherwise.

smoogipoo (approved):

osu.Game/Overlays/BeatmapSet/BeatmapSetLayoutSection.cs
```diff
@@ -0,0 +1,29 @@
...
+using osu.Framework.Extensions.Color4Extensions;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Effects;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Game.Overlays.BeatmapSet
+{
+    public class BeatmapSetLayoutSection : Container
```
peppy: please document in the future. it's really hard to tell why these kinds of rogue classes exist otherwise.

PR #7953, opened 2020-02-22
A mod that forces the player to alternate

### LINKED ISSUE #3773: A mod to only allow alternating
I wanted to switch to full-alt playstyle in taiko while I'm still relatively new, so I modified the game to only allow alternating hands. I was wondering if that kind of feature would be useful as an official mod because usually switching to full-alt is quite painful. Also I've heard always alternating in std helps you with accuracy or something like that so the mod could work for std as well.

voidedWarranties:
Implements/Closes #3773 
Both standard and taiko have an option to fail when a repeat key is pressed
taiko has options to allow repeats when the color changes and after drumrolls (set to the most forgiving option by default)

Probably needs a new name/description


bdach:
Accessing key bindings in mods seems like the incorrect direction to take. I can't say if it's feasible for sure, but you should look at using action receptors instead (relevant classes: `ActionReceptor` and its example usages, in particular `KeyCounterDisplay.Receptor`).

voidedWarranties:
When using `OnPressed` to block keypresses, the `KeyBindingContainer` is still able to receive the event and sliderends are still scored (giving a 50). Is there any way to prevent this?
image
Also, is there any way to prevent other `IKeyBindingHandler`s (ie. the key counter) from receiving the event?

bdach:
Perhaps what you're looking for is to hijack the ruleset-specific input managers (ie. `OsuInputManager`) and use the last pressed button info from the receptor to suppress UI events that you don't want to fire in all other receptors. I'm not sure changes to receptors that would allow suppression of propagation to other receptors are a good direction.

As a side thought, the option to fail if the wrong key is used is equivalent to combining the mod with Sudden Death.

recapitalverb:
> As a side thought, the option to fail if the wrong key is used is equivalent to combining the mod with Sudden Death.
IMO not really, with Sudden Death it fails on misses, while enabling the option to fail on wrong key means it only fails if you hit a circle with the wrong key so you can still continue when you miss the circle entirely.

bdach:
You're right, that is slightly different behaviour.

LanTenDragon:
Depending on how you see it, this mod can be considered as a ``Difficulty Increase`` mod as well.

voidedWarranties:
> Perhaps what you're looking for is to hijack the ruleset-specific input managers (ie. `OsuInputManager`) and use the last pressed button info from the receptor to suppress UI events that you don't want to fire in all other receptors.

`UIEvent` has no information on what actions are pressed (without dealing with keybindings) until the base event in `KeyBindingContainer` has already been called, and knowing what action is pressed in that event probably defeats the purpose of using a receptor
Would it be better to move code dealing with bindings to the `RulesetInputManager`s and passing the actions into `BlockCondition`?

bdach:
IMO bindings should stay in `DatabasedKeyBindingContainer`. It's an abstraction that works well and leaking bindings into the `RulesetInputManager`s breaks layering.

Maybe injecting a custom `KeyBindingContainer` is what you're after. Again, I can't say for sure without actually trying to essentially rewrite the PR.

voidedWarranties:
In this case, would it be helpful to create a separate PR with something like `IApplicableToInputManager` which allows the mod to create its own implementation of the `RulesetInputManager`? Something like this could probably be used for other mod proposals as well.

voidedWarranties:
Closed on accident. Apologies

voidedWarranties:
All outside usages of keybindings has been removed, but not sure if this is what you're looking for

pavlukivan:
I'll remake the taiko version, please don't merge yet

pavlukivan:
That said, I really don't like this behavior. It would be way better to miss instead of failing, even if it requires adding new handlers.

osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -153,10 +163,55 @@ protected virtual RulesetKeyBindingContainer CreateKeyBindingContainer(RulesetIn
 
         public class RulesetKeyBindingContainer : DatabasedKeyBindingContainer<T>
         {
+            public event Func<T?, List<T>, bool> HandleBindings;
```
bdach: * This event signature is confusing. There is no reason to split between the single action nullable case and the list-of-all-actions collection case.
* By convention events are usually named `OnXYZ`. In this case I'd probably suggest `OnRulesetAction`.

osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -38,6 +40,14 @@ protected override InputState CreateInitialState()
 
         private readonly Container content;
 
+        public event Func<T?, List<T>, bool> HandleBindings
+        {
+            add => ((RulesetKeyBindingContainer)KeyBindingContainer).HandleBindings += value;
+            remove => ((RulesetKeyBindingContainer)KeyBindingContainer).HandleBindings -= value;
+        }
+
+        public void RetractLastBlocked() => ((RulesetKeyBindingContainer)KeyBindingContainer).RetractLastBlocked();
```
bdach: These changes should not reside here publicly if at all avoidable. You should try to extract a ruleset input manager implementation specific to your mod that contains the blocking behaviour and expose it there only.

osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -153,10 +163,55 @@ protected virtual RulesetKeyBindingContainer CreateKeyBindingContainer(RulesetIn
...
+                {
+                    var pressedActions = combos.Select(c => c.GetAction<T>()).ToList();
+
+                    InputKey? key = null;
+                    if (e is KeyDownEvent kb)
+                        key = KeyCombination.FromKey(kb.Key);
+                    else if (e is MouseDownEvent mouse)
+                        key = KeyCombination.FromMouseButton(mouse.Button);
+                    else if (e is JoystickButtonEvent joystick)
+                        key = KeyCombination.FromJoystickButton(joystick.Button);
```
bdach: I'm not sure why the re-mapping to ruleset actions is done here. Can't you just use the public `KeyBindingContainer.PressedActions`?

osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -153,10 +163,55 @@ protected virtual RulesetKeyBindingContainer CreateKeyBindingContainer(RulesetIn
...
+
+                    var single = combos.Find(c => c.KeyCombination.Keys.Any(k => k == key))?.GetAction<T>();
+
+                    var blocked = HandleBindings?.GetInvocationList().Cast<Func<T?, List<T>, bool>>().ToList().FindAll(f => f.Invoke(single, pressedActions)).Any();
```
bdach: This is borderline unreadable. This needs to be split more.

osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -153,10 +163,55 @@ protected virtual RulesetKeyBindingContainer CreateKeyBindingContainer(RulesetIn
...
+                }
+
+                return base.Handle(e);
+            }
+
+            public void RetractLastBlocked()
+            {
+                if (lastBlocked != null)
+                    base.Handle(lastBlocked);
+            }
```
bdach: Suppressing an input just to retract that suppressing later feels backwards. This is too generic (as you need this only for the taiko implementation) and it shows.

osu.Game.Rulesets.Taiko/Mods/TaikoModEnforceAlternate.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            if (blocked.Any())
+                inputManager.RetractLastBlocked();
+
+            return false;
+        }
+
+        private bool handleBindings(TaikoAction? single, List<TaikoAction> pressedActions)
+        {
+            if (!healthProcessor.IsBreakTime.Value)
```
bdach: Invert & early-return to reduce nesting.

osu.Game.Rulesets.Taiko/Mods/TaikoModEnforceAlternate.cs
```diff
@@ -0,0 +1,162 @@
...
+            {
+                var rims = pressedActions.FindAll(a => a == TaikoAction.LeftRim || a == TaikoAction.RightRim);
+                var centres = pressedActions.FindAll(a => a == TaikoAction.LeftCentre || a == TaikoAction.RightCentre);
```
bdach: Splitting this into lists feels awkward as you have at most 4 actions you care about at one given frame. Flag enums or a struct would be better and most likely faster. It would simplify a lot of the neighbouring code, too.

bdach (changes requested):
The ideal amount of changes to the base components (input managers, etc.) that should be made for a mod implementation is zero. I suppose to implement @pavlukivan's variant you want to be hooking into `DrawableRuleset.OnNewResult` to change hits into misses. I think that combined with a receptor to read inputs and a custom `KeyBindingContainer` might yield the cleanest solution achievable.

osu.Game/Rulesets/UI/RulesetInputManager.cs
voidedWarranties: `PressedActions` is propagated after the event in `RulesetKeyBindingContainer` is called (after `base.Handle`)

voidedWarranties:
@pavlukivan Since nothing in this PR is final yet (including outside of the taiko mod), I would prefer if you post what problems you have with it

For now, I've decided to not implement assigning misses on repeat keys because the player is already almost certain to miss the note (unless they are able to react and press the correct key during the hitwindow). But, it could be made an option

As mentioned above, (as far as I can tell) creating a custom `RulesetInputManager`/`KeyBindingContainer` would break compatibility with any other (future) mods modifying them as well, even if these mods could logistically be combined.

pavlukivan:
On the second thought, I don't have that much time, so let me document the behavior here instead of implementing it:

There's 2 main playstyles - kddk and ddkk. There should be a switch in the settings, defaulting to kddk.

Behavior depending on playstyle (note that the settings listed here are playstyle-specific):
## kddk
There are two groups of keys, each group has 1 rim and 1 center (L rim+L center; R rim+R center). For any sequence of notes without big notes, swells or long breaks, the key group must change on each hit. After big note there are two possible options (toggleable in settings) (still, preferably treat the big note as a small note if the note wasn't hit with two keys) - 1. reset everything and allow any key after the note 2. treat the big note as two small notes - that is, the *next* note after the big note must be hit with the key group that the key *before* the big note does NOT belong to. Another setting can toggle whether drumroll ticks should be treated as small notes. If the setting is disabled, drumrolls should be treated as breaks and reset allowed keys. For swells, there are two acceptable ways of hitting them (rr=right rim, lc=left center...) - 1. rr-lc-rr-lc... or lr-rc-lr-rc... 2. lr-lc-rr-rc-lr-lc-rr-rc... or lc-lr-rc-rr-lc-lr-rc-rr.., anything else should always be banned.

## kkdd
This one is quite close to the implemented behavior - notes are divided by color, for each color the key must change on every note, with a setting to allow any key after a color change. Swells' behavior should be same as on kddk. For drumrolls, theres can be a setting that allows to treat the drumroll ticks as small notes - however, it's more lenient because the allowed key depend on whether the used key corresponds for a center or a rim. Similarly, for big notes there should be a toggleable option to treat them as two small notes if they are hit with two keys.

pavlukivan:
Also, I meant adding some additional bindings to update judgements before they are applied to the game code, not binding to something that already exists

voidedWarranties:
I think out of these things, the things I haven't implemented are

- [x] Reset after large note
- [ ] Swell behavior
- [ ] Reset after break
- [ ] kddk and kkdd distinction

When it comes to kddk and kkdd, is there any need for logical distinction between these? From what I can tell, the current logic applies to kkdd, but instead of alternating hands you alternate fingers.

> however, it's more lenient because the allowed key depend on whether the used key corresponds for a center or a rim.

I'm not sure what you mean by this - if the allowed key depended on color rather than left/right, wouldn't this block most keypresses?

Also, how do those who use mouse play the game? Or is there nobody who actually does this

pavlukivan:
> When it comes to kddk and kkdd, is there any need for logical distinction between these? From what I can tell, the current logic applies to kkdd, but instead of alternating hands you alternate fingers.

It is absolutely necessary - kddk alternating focuses on using wrists to the fullest by changing hand on every note, while kkdd alternating simply reduces finger strain.

> if the allowed key depended on color rather than left/right, wouldn't this block most keypresses?

I mean that you can either press it with "the next allowed center" or "the next allowed rim", not just a single "next allowed key"

> Also, how do those who use mouse play the game? Or is there nobody who actually does this

Mouse only has 2 buttons, taiko is a 4-button game.

pavlukivan:
> For now, I've decided to not implement assigning misses on repeat keys because the player is already almost certain to miss the note (unless they are able to react and press the correct key during the hitwindow)

You can bypass it by doing stuff like doubletapping.

voidedWarranties:
Taiko logic and features need some serious work, and tests need to be recreated under #8066. Closing for now, might open a new PR if I decide to revisit.

PR #7954, opened 2020-02-22
Implement the ability to leave a new reply to existing comments in CommentsContainer

EVAST9919:


bdach (changes requested):
IMO `NewReplies` in general should not be necessary. I see no reason to distinguish replies from "new replies".

The bigger problem here is that the comment editor basically does the equivalent of a fire-and-forget operation. `OnCommit` fires that `PostReplyRequested` action, which schedules an API request to post the comment and then appends the result from API to the comment page, but at that point the editor no longer knows anything about the operation it aimed to do, so it remains open. You could probably try to pass it in the action but I'd be against it too as it breaks layering.

If you ask me the editor should receive necessary context (commentable type and parent comment ID), fire the request itself, then on success close itself and raise an event that it has posted a comment, which then the page will handle cleanly without having the split. That will also allow for cleaner error handling if we ever decide we want to present an error message locally in the editor somewhere.

EVAST9919:
The main reason why I have 'NewReplies' is because they should only appear on very top when loaded (when `Replies` will appear in bottom)

bdach:
> The main reason why I have 'NewReplies' is because they should only appear on very top when loaded (when `Replies` will appear in bottom)

That doesn't justify the two separate lists. At most the order of insertion should be affected (doing `Insert(0, ...)` instead of `Add()`. The two lists make everything twice as difficult to understand and lead to code duplication.

bdach:
FWIW, on the other side to be honest I find web's behaviour in that regard kind of counter-intuitive. In threaded comments I would expect new comments to also get added last. It's super weird that they get added first web-side, especially when after refresh they move to last anyways.

EVAST9919:
Well, they move to last only if sorted by top/old, but that's what we have.
Anyway will revisit addition.
Not sure why we want to `CommentEditor` do the online stuff, though. In my implementation it will not be interactable while new comment is loading (because it's in loading state which means we can't double-post or something)

bdach:
I don't see what the lack of interactability of the editor has to do where the request is done. It should still be possible to block it while the request is in-flight even in the editor?

Also I'm not necessarily saying that it should be added to the editor itself, could do an `EditorContainer` or something of the sort. Mostly what I want to achieve is to have the posting logic as encapsulated as possible. Right now we have:

* `replyEditor.OnCommit`
* firing `DrawableComment`'s `PostReplyRequested`
* which is then actually sending the request
* on success of which `AppendComments()` is called.

Instead you could have

* `replyEditor.OnCommit`
* firing `EditorContainer`'s one single method that sends the request, receives it, unblocks the editor, and raises an event with the response comment
* which then you could pick up higher up and just operate on data instead.

I'd have to look how the code would look like, but purely spitballing the second variant sounds simpler in execution and understandng.

EVAST9919:
Response should be handled by `CommentsPage` anyway, because it helps to avoid duplicates. Imagine we have 2 pages of unloaded replies sorted by old (oldest first), so after posting a reply and loading unloaded one (by clicking on "show replies") we will receive our posted comment too at some point. 
Probably I just not fully understand the logic and you already included that case in your suggestion

EVAST9919:
Also since reply uses async loading so we should probably unlock the editor only when it's loaded (like in my case, but you are suggesting unloading it after receiving the response(?))

bdach:
> Response should be handled by `CommentsPage` anyway, because it helps to avoid duplicates. 

It doesn't need to handle the *actual* web response. It needs to handle its *contents*, sure, and your way of handling duplicates is fine. I'm suggesting constraining API calls to as narrow a scope as possible, and returning a comment bundle as early as possible. Operate on the data, keep flow control simple.

> Also since reply uses async loading so we should probably unlock the editor only when it's loaded (like in my case, but you are suggesting unloading it after receiving the response(?))

I'm not sure I follow what the problem is here.

bdach:
Seeing that we're at odds with the opinions on the direction that should be taken here we probably need a third opinion.

smoogipoo:
I would also prefer that it was encapsulated further. It's probably fine to unblock immediately after receiving the response - async load shouldn't take long for these components.

EVAST9919:
Posting comments is cool, but it'd be better to make a cleanup first (at least remove the CommentsPage, as mentioned [here](https://github.com/ppy/osu/pull/7786#issuecomment-589603329). Closing for now.

PR #7961, opened 2020-02-22
Rework issue templates

bdach:
# Summary

Add log file locations to issue templates which mention their existence. Android log location purposefully omitted to discourage reports.

Also add a mention of the issue templates in README.md, for how much that's worth.


bdach:
[Following discord discussion](https://discordapp.com/channels/188630481301012481/188630652340404224/680799584959594496) I also removed the mobile issue template as it doesn't really seem to be working that much in practice.

Let's treat this PR as a place of discussion about the templates and their usefulness altogether. While I agree that they are a bit annoying to use by devs, they have a non-negligible positive impact on the average quality of an issue and save us at least a bit of triaging work.

peppy (approved):

PR #7967, opened 2020-02-22
Reduce hit error display performance overhead

peppy:
Due to the number of lines (and an unnecessarily long transform time) the hit error display could end up using quite a bit of CPU time. This is especially noticeable when fast-forwarding a replay, as they run in game time while the replay is running many times faster than real-time.

image


bdach (approved):
Can confirm this gets back some performance (a few frames per second on centipede). Code looks good too 👍

smoogipoo (approved):

PR #7968, opened 2020-02-22
Improve gameplay performance via follow point renderer optimisations

### LINKED ISSUE #7949: long Beatmaps significantly lower performance
The longer the beatmap is the lower the initial performance is.
In addition to this, over the course of a map of any length, performance goes downhill from the beginning to the end.

In the Screenshots taken for comparison you can see that the longer the map goes on, the time update needs to work is increased, resulting in sleeptime in draw.
Also the longer the map is the intial work time for update is longer the longer the map is.
This holds true as long as you compare beatmaps with a similiar notedensity in total.
In general: The more notes total the beatmap has makes performance worse.
Autoplay was taken because its easier to screenshot, the same happens when playing normally.

[short map](https://osu.ppy.sh/beatmapsets/433506#osu/960539) beginning:
short initial
at the end:
short end
[long map](https://osu.ppy.sh/beatmapsets/70744#osu/203315) beginning:
long initial
at the end:
long end
[Marathon map](https://osu.ppy.sh/beatmapsets/296952#osu/674188) beginning:
marathon initial
at the end:
marathon end

osu!lazer version 2020.221.0

PC Specs:
intel i7-4770
Msi Nvidia GTX 1070 (primary | main monitor)
Msi Nvidia GTX 760 (secondary | 2cnd and 3rd monitor)
16GB RAM

peppy:
Before:

image

After:

image

Closes #7949.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs
@@ -20,6 +20,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Connections
         private const int spacing = 32;
         private const double preempt = 800;
 
+        public override bool RemoveWhenNotAlive => false;
+
         /// <summary>
         /// The start time of <see cref="Start"/>.
         /// </summary>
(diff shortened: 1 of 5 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointRenderer.cs
```diff
@@ -74,6 +72,8 @@ private void addConnection(FollowPointConnection connection)
                 FollowPointConnection previousConnection = connections[index - 1];
                 previousConnection.End = connection.Start;
             }
+
+            AddInternal(connection);
```
bdach: I presume this was moved to the end of the method to ensure connection ends were initialised before proceeding, right?

bdach (dismissed):
Yeah, this is the big perf win. Went up from 9-10 fps (constant, from the very start) on centipede to full 240 on the starting spinner and never went below 25-30 on my machine. (I should really be getting more on this machine, which makes me a little worried about its lifespan) Anyway, looks good and very fast 🚀

osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointRenderer.cs
peppy: Generally with `LifetimeManagementContainer` you want to set the lifetime *before* adding to avoid an extra update. I checked this and it doesn't seem to matter where this specific call is (since it's all loaded in BDL anyway) but it feels like the more correct place to put it so I left it at the end.

osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs
```diff
@@ -131,10 +139,13 @@ private void refresh()
...
                     fp.MoveTo(pointEndPosition, osuEnd.TimeFadeIn, Easing.Out);
                     fp.Delay(fadeOutTime - fadeInTime).FadeOut(osuEnd.TimeFadeIn);
-                }
 
-                fp.Expire(true);
+                    finalTransformEndTime = fadeOutTime + osuEnd.TimeFadeIn;
+                }
             }
+
+            LifetimeStart = startTime;
```
smoogipoo: I’m not convinced. Just reading the code you can tell fp.LifetimeStart is going to be different from LifetimeStart since the preempt isn’t included. Have you actually tested this for visual correctness?

I suspect what you want to do is actually use the minimum and maximum lifetimes of the children. All of this stuff is too raw and fragile against these transforms potentially changing (and they will change to match stable at some point).

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs
peppy: Yeah, this should probably be done in a similar way to `finalTransformEndTime` (based on the first the first `fadeInTime`).

We can't use the old code – it was calling `Expire` from BDL, using the game clock instead of audio clock and resulting in completely warped lifetimes (lasting hundreds of seconds).

bdach: I jumped the gun here, it does look visually incorrect. Will remember to check better next time.

bdach:
While I won't attempt to re-review this since I've already been wrong once before, I tested the latest version with the new commit and it seems to be both fast and visually correct now.

smoogipoo (approved):

PR #7969, opened 2020-02-22
Fix scale of loading animation on player loading screen

### LINKED ISSUE #7942: Wrong scaling on loading animation during map loading
**Describe the bug:** 

Happens on Windows and iOS.

F8D6AB86-83A6-4B87-B05E-C477AE4C48C7

**osu!lazer version:** 
iPad Pro 12.9
Windows 10
UI scaling at 0%

peppy:
Uses new design (feels better to me).

image

Also adjusts the rotation cycle to make the animation more dynamic (previously would only rotate to two "nodes").

Closes #7942.


bdach (approved):
I like the way it looks now better as well 👍

smoogipoo:
Merged in https://github.com/ppy/osu/pull/7968

PR #7971, opened 2020-02-22
Hide pp display on leaderboards if map is qualified or loved

thewildtree:
On web, the pp value is always shown on the leaderboards, even if a map is loved or qualified. Lazer receives null values from the API in those cases (or loses the 0 value somewhere along the road, haven't checked it too precisely), so it doesn't show any number there. 

This PR hides the pp display if a given beatmap is not ranked.

image


bdach:
While this PR on its own does what it says correctly, I kind of disagree with the web status quo. Performance points in the context of loved/qualified maps are just not valid and displaying zeroes does not seem to bring much value. If anything, in my opinion the columns should be hidden outright on loved and qualified maps, and displayed only for ranked maps.

peppy:
I'd rather hide the column.

TheWildTree:
Hid the column instead as suggested. Here's how it looks now:

image

bdach:
Looks good, but I'm not overly sold on the implementation. Right now it works correctly because it relies on web implementation details (getting `null` on qualified/loved maps). It feels like a weak point since if a score makes it in by some freak accident that *does* have a PP value it'll be displayed with the column and look wrong.

Maybe `ScoreTable` should have a parameter/property that determines if the performance point column should be displayed, which would be set to `beatmapInfo.Status == BeatmapSetOnlineStatus.Ranked` higher up. Not too certain that is the more correct way to go, may be a little bit paranoid.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -65,7 +67,9 @@ public IReadOnlyList<ScoreInfo> Scores
                 for (int i = 0; i < value.Count; i++)
                     backgroundFlow.Add(new ScoreTableRowBackground(i, value[i], row_height));
 
-                Columns = createHeaders(value[0]);
+                isBeatmapRanked = value.First().Beatmap.Status == BeatmapSetOnlineStatus.Ranked;
```
bdach: This is not entirely what I was looking for. There is nothing stopping users from providing a list of scores from different beatmaps (while there's no situation I can think of right now that would necessitate that, it isn't out of the question), at which point the `First()` would be wrong. I'd honestly prefer a plain, explicit ctor parameter or property.

TheWildTree: Fair point, I haven't thought about that scenario at all. Will change.

osu.Game/Overlays/BeatmapSet/Scores/ScoreTable.cs
```diff
@@ -65,11 +65,13 @@ public IReadOnlyList<ScoreInfo> Scores
                 for (int i = 0; i < value.Count; i++)
                     backgroundFlow.Add(new ScoreTableRowBackground(i, value[i], row_height));
 
-                Columns = createHeaders(value[0]);
+                Columns = createHeaders(value.First());
                 Content = value.Select((s, i) => createContent(i, s)).ToArray().ToRectangular();
             }
         }
 
+        public bool IsBeatmapRanked { get; set; }
```
bdach: I'd rename to
```suggestion
        public bool ShowPerformanceColumn { get; set; }
```
It gets the point across more directly and has the nice effect of explicitly expressing at usage site that performance points are only shown for ranked maps.

osu.Game/Overlays/BeatmapSet/Scores/ScoresContainer.cs
```diff
@@ -59,11 +59,12 @@ protected APILegacyScores Scores
                 }
 
                 var scoreInfos = value.Scores.Select(s => s.CreateScoreInfo(rulesets)).ToList();
+                var topScore = scoreInfos.First();
 
                 scoreTable.Scores = scoreInfos;
+                scoreTable.IsBeatmapRanked = topScore.Beatmap.Status == BeatmapSetOnlineStatus.Ranked;
```
bdach: This doesn't actually work properly for ranked maps (screenshot below from `TestSceneBeatmapSetOverlay`, `Online` step):

image

Columns are created in the setter of `Scores` one line above, so this has no effect at the point of being set to `true` here. This indicates faulty API design as now you have to call things in the right order to have them work. The solution here is to remove `Scores` and `IsBeatmapRanked` and merge them into a method:
```csharp
public void DisplayTable(IEnumerable<ScoreInfo> scores, bool showPerformanceColumn);
```
Also please add tests for that behaviour (checking for presence of the column or lack thereof) while you're at it.

bdach (changes requested):

osu.Game/Overlays/BeatmapSet/Scores/ScoresContainer.cs
TheWildTree: Added something similar to what you proposed - please check if I've done it correctly.

Regarding tests - those would be best suited in the `ScoreContainer` test scene which in its current state is entirely broken as noted in #7916, so it kinda hard to test anything there.

bdach (approved):
Considered adding a scene for `ScoreTable` only, but at that point I'd have to add a scene for `TopScoreStatisticsSection` for 100% coverage so I decided against it. Tested manually, let's get it in for now, we can sort out testing whenever `ScoresContainer` is dealt with hopefully 👍

PR #7972, opened 2020-02-23
Add "Scroll to top" button to overlays

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu/pull/8471
nice


bdach:
I really like this addition. It feels like something that has been missing UX-wise for a long time and it's nice to have it.

Code looks good as far as I can tell too, even though the diff is ugly as usual when indentation changes are involved. I tweaked the test scene a little (check out the commit msg for the what and why). Other than that, great work 👍

peppy:
Why is this not using `OsuAnimatedButton`?

peppy:
I guess it's a bit custom and `OsuAnimatedButton` provides too much.

This should be addressed though (the button should likely fade as soon as clicked, not later):

image

EVAST9919:
I guess, I should appear again, if right after clicking you manually scroll down?

peppy:
yep, sounds fine.

peppy:
- No unit testing added to cover new case
- Lacking manual testing

image

you can do better than this

EVAST9919:
Should be good now

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                Anchor = Anchor.BottomRight,
+                Origin = Anchor.BottomRight,
+                Margin = new MarginPadding(20),
+                Action = () =>
+                {
+                    if (animaionIsRunning)
+                        return;
+
+                    animaionIsRunning = true;
```
bdach: Is never set back to `false`, leading to a scenario in which you could leak mouse down events to background drawables (because `OnMouseDown` yields to background drawables if this is true). See bdach/osu@7fdafef and [this video](https://streamable.com/l5qgx).

Reproduction steps:
* Scroll down to make button visible
* Click button
* Do not scroll or drag container, hold left mouse over button (underneath the scrolling container)
* Button will take the left mouse down event
* Scroll or drag container, hold left mouse over button
* Button will not take the left mouse down event

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,173 @@
...
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Game.Overlays
+{
+    public class OverlayScrollContainer : OsuScrollContainer
+    {
+        public ScrollToTopButton Button { get; }
+
+        private bool animaionIsRunning;
```
bdach: Typo in name (should be at least `animationIsRunning`). Old name was probably better, but I'd ideally want to see this entirely removed.

bdach (changes requested):
It is not.

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,155 @@
...
+                            {
+                                Anchor = Anchor.Centre,
+                                Origin = Anchor.Centre,
+                                Size = new Vector2(15),
+                                Icon = FontAwesome.Solid.ChevronUp
+                            }
+                        }
+                    });
+
+                    TooltipText = @"Scroll to top";
```
bdach:
```suggestion
                    TooltipText = "Scroll to top";
```

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,155 @@
...
+                {
+                    content.ScaleTo(0.75f, 2000, Easing.OutQuint);
+                    return base.OnMouseDown(e);
+                }
+
+                protected override void OnMouseUp(MouseUpEvent e)
+                {
+                    content.ScaleTo(1, 1000, Easing.OutElastic);
+                    base.OnMouseUp(e);
+                }
```
bdach: Please explicitly return `true` or `false` as much as possible when overriding UI handlers. Now `OnClick` returns `true` as the base impl in `ClickableContainer` does so, but `OnMouse{Up,Down}` don't because they go down to `Drawable.HandleUIEvent`.

bdach (changes requested):
This should have been split over 2 PRs at least (introduction & integration). This diff is very large.

I noticed one more potential issue ([video](https://streamable.com/smboi)): if you try to use the mouse to drag past the end of a scrollable container that does not normally scroll, the button shows up while it probably shouldn't? Not sure.

This is not a thorough review, more like a first pass. The `OverlayScrollContainer` by itself looks mostly fine but I want to look at usages closer, in particular at `UserProfileOverlay` (to see if we can get rid of the `CreateScrollContainer` override).

EVAST9919:
> I noticed one more potential issue ([video](https://streamable.com/smboi)): if you try to use the mouse to drag past the end of a scrollable container that does not normally scroll, the button shows up while it probably shouldn't? Not sure.

I'm pretty sure there's a flag to check this kind of situation (when content height < parent container height). Will investigate.

osu.Game/Overlays/OverlayScrollContainer.cs
EVAST9919: We should call `base.OnClick` because action handled in there
https://github.com/ppy/osu-framework/blob/20cf599582d5ec2e9cd1bfe7231e8352ba7efe0c/osu.Framework/Graphics/Containers/ClickableContainer.cs#L26

EVAST9919:
Will open a new simplified pr withoud excessive logic such as overlays refactoring which is out of context.

PR #7973, opened 2020-02-23
Fix hitobjects with unknown lifetimes by enforcing non-null judgement

peppy:
We've seen multiple cases where `DrawableHitObject` are stuck in the lifetime management container due to not implementing a judgement (meaning they are never "hit" or "missed"). To avoid this going forward `CreateJudgement()` must be implemented and return a non-null judgement.

- Fixes BananaShower and JuiceStreams in osu!catch.
- Makes HitObject abstract and cleans up convert HitObject implementations.

### All HitObjects must implement CreateJudgement and return non-nullable Judgement

Judgements are used to control lifetime and ensure hitobjects are expired correctly. For hitobjects which are not used in gameplay (ie. convert types), `ConvertHitObject` can be used. For hitobjects which don't want to provide a score, combo or health adjustment for a judgement, a default implementation `IgnoreJudgement` is available.

Existing usages can safely add the following line for compatibility:

```csharp
public override Judgement CreateJudgement() => new IgnoreJudgement();
```


osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -144,9 +144,10 @@ protected virtual void CreateNestedHitObjects()
 
         /// <summary>
         /// Creates the <see cref="Judgement"/> that represents the scoring information for this <see cref="HitObject"/>.
-        /// May be null.
+        /// Used to decide on drawable object lifetimes.
         /// </summary>
-        public virtual Judgement CreateJudgement() => null;
+        [NotNull]
+        public abstract Judgement CreateJudgement();
```
smoogipoo: I'd much rather prefer this was `=> new Judgement()` and the `HitWindows` maintained below as in the previous version.

osu.Game/Rulesets/Objects/ConvertHitObject.cs
```diff
@@ -0,0 +1,18 @@
...
+
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
+
+namespace osu.Game.Rulesets.Objects
+{
+    /// <summary>
+    /// A hit object only used for conversion, not actual gameplay.
+    /// </summary>
+    public class ConvertHitObject : HitObject
```
smoogipoo: As above, not really a fan of this. What's the reasoning here?

smoogipoo: As discussed before:
1. Make `abstract`.
2. Move to legacy namespace.
3. `internal`ise.
4. Make `HitObject` non-abstract again, give it sane default implementations.

smoogipoo (changes requested):

osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -144,9 +144,10 @@ protected virtual void CreateNestedHitObjects()
 
         /// <summary>
         /// Creates the <see cref="Judgement"/> that represents the scoring information for this <see cref="HitObject"/>.
-        /// May be null.
+        /// Used to decide on drawable object lifetimes.
```
smoogipoo:
```suggestion
```

Really unnecessary. It's used to score the hitobject primarily, lifetime only occurs as a consequence of doing so.

smoogipoo (changes requested):

smoogipoo (approved):

PR #7974, opened 2020-02-23
Fix bar lines in osu!taiko and osu!mania not correctly being cleaned up

peppy:
- [x] Depends on #7973


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -265,7 +262,7 @@ private void updateState(ArmedState newState, bool force = false)
                     }
                 }
 
-                if (state.Value != ArmedState.Idle && LifetimeEnd == double.MaxValue)
+                if (state.Value != ArmedState.Idle && LifetimeEnd == double.MaxValue || HitObject.HitWindows == null)
```
bdach: I thought [the goal was to get rid of `null` `HitWindows`](https://github.com/ppy/osu/pull/6445)? What's different here? Kind of getting hard to track what's going on with those.

peppy: It's a bit of a weird one, but `DrawableHitObjects` may still have null `HitWindows` (happens if they are never `AppliedDefaults`), which I believe is by design. Will definitely be discussing the direction of that with @smoogipoo.

The previous change was specific to `HitObjects` and more of a clean-up effort. I thought it may have been usable to also fix barlines, but due to how they are processed (or not), they need to be handled slightly differently. I actually don't think it is a prereq for this PR, so may rebase away from that depending on further discussion.

smoogipoo: I think this is fine for now...

smoogipoo (approved):

PR #7978, opened 2020-02-23
Add ability to disable blur during breaks

### LINKED ISSUE #7495: Add the ability to "unblur" the background during gameplay breaks
**Describe the new feature:**
Similar to "Lighten playfield during breaks", but this one undoes the blur effect set on the gameplay background to a certain percentage (0% blur would be my suggestion).

The feature could be added as a toggle right under the "Lighten playfield during breaks" toggle or, for clear differentiation between settings, have each toggle to be put under the background adjustment slider in correlation (e.g. playfield lighten toggle under the "Background dim" slider and background unblur toggle under the "Background blur" slider).

protheory8:
Closes #7495
I also renamed `lightenDuringBreaks` to `disableEffectsDuringBreaks`, not sure if that was a right choice.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
@@ -34,8 +34,8 @@ namespace osu.Game.Overlays.Settings.Sections.Gameplay
                 },
                 new SettingsCheckbox
                 {
-                    LabelText = "Lighten playfield during breaks",
-                    Bindable = config.GetBindable<bool>(OsuSetting.LightenDuringBreaks)
+                    LabelText = "Disable background effects during breaks",
+                    Bindable = config.GetBindable<bool>(OsuSetting.DisableEffectsDuringBreaks)
                 },
                 new SettingsCheckbox
                 {
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimContainer.cs
@@ -19,12 +19,12 @@ namespace osu.Game.Tests.Visual.Background
 
         private readonly BindableBool isBreakTime = new BindableBool();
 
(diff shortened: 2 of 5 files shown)


bdach:
On top of the adjustments made to pre-existing dim tests, the blur level should also be tested analogously.

protheory8:
Can't figure out how to do that, what file do I need to change?

bdach:
You might end up having to write your own test scene for `BackgroundScreenBeatmap` to test this behaviour. I think it'll be very similar to what `TestSceneUserDimContainer` does.

protheory8:
I don't really remember what this PR was about, should I close it?

protheory8:
So I tried to implement the test, but they all fail because `disableEffectsDuringBreaks` in `BackgroundScreenBeatmap` is `null` for some reason

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -34,8 +34,8 @@ private void load(OsuConfigManager config)
                 },
                 new SettingsCheckbox
                 {
-                    LabelText = "Lighten playfield during breaks",
-                    Bindable = config.GetBindable<bool>(OsuSetting.LightenDuringBreaks)
+                    LabelText = "Disable background effects during breaks",
```
peppy, line 37: doesn't this sound wrong? lighten isn't lightening to 0, so there's still "effects" being applied.

peppy (changes requested):

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
bdach: I'm not even sure how to name this better so that it's both accurate and understandable.

peppy: Not 100% sure about having these both as one setting either, but if we are going to do it this way how about "Make background more visible during breaks"?

smoogipoo:

PR #7982, opened 2020-02-24
Fix potential crash when clicking on show more button in comments

EVAST9919:
The reason is that we can receive already existing comments not only in `IncludedComments` list but also in the `Comments` list.

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneCommentsPage.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneCommentsPage.cs
@@ -13,6 +13,8 @@ using osu.Framework.Bindables;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics;
 using osuTK;
+using JetBrains.Annotations;
+using NUnit.Framework;
 
 namespace osu.Game.Tests.Visual.Online
 {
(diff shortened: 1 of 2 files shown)


bdach (changes requested):
I don't like the unnecessary code duplication this introduces. I'd rather go for a
```csharp
foreach (var child in bundle.Comments.Concat(bundle.IncludedComments))
```
or a local function.

bdach:
I moved the inline comment into a more-suitable place (looks better near the actually relevant conditional).

In general the change looks ok and obvious enough, but when submitting PRs for bugs/crashes I'd generally like to see a test reproducing the crash scenario to make sure it doesn't regress again. I would have submitted one but I'd have to expose things in `CommentsPage` to do so and I don't want to interfere with your other PRs so I abstained from doing this.

bdach (dismissed):
approving even though I pushed here, since I only made one cosmetic change

peppy:
Can a test be added showing this crash?

EVAST9919:
Added

smoogipoo:
Just confirmed with the osu-web, this is fine for now. The proper solution is to use cursors, but the documentation for that is neither available (right now) nor does the logic around it fit into our existing cursor response usages (`ResponseWithCursor`).

smoogipoo (approved):

PR #7983, opened 2020-02-24
Implement UserListToolbar component

EVAST9919:
Screenshot_2


osu.Game.Tests/Visual/UserInterface/TestSceneUserListToolbar.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+            });
+
+            Add(new FillFlowContainer
+            {
+                AutoSizeAxes = Axes.Both,
+                Direction = FillDirection.Vertical,
+                Spacing = new Vector2(0, 5),
```
bdach: Not sure if this shouldn't be
```suggestion
                Spacing = new Vector2(0, 10),
```
Each sort tab item on web has 5px margin individually, which comes to 10px spacing between items altogether.

osu.Game/Overlays/Comments/CommentsHeader.cs
```diff
@@ -11,6 +11,8 @@
 using osuTK;
 using osu.Framework.Input.Events;
 using osu.Game.Graphics.Sprites;
+using System.ComponentModel;
+using Container = osu.Framework.Graphics.Containers.Container;
```
bdach: It's a nitpick but I'd prefer this the other way round:
```csharp
using Description = System.ComponentModel.Description;
```
or even use the fully qualified name.

osu.Game/Overlays/OverlayPanelDisplayStyleControl.cs
```diff
@@ -0,0 +1,97 @@
...
+            });
+        }
+
+        protected override TabFillFlowContainer CreateTabFlow() => new TabFillFlowContainer
+        {
+            AutoSizeAxes = Axes.Both,
+            Direction = FillDirection.Horizontal
+        };
+
+        private class PanelDisplayTabItem : TabItem<OverlayPanelDisplayStyle>, IHasTooltip
```
bdach: These should probably have `HoverClickSounds` for UX consistency, to indicate interactability.

bdach (commented):
The *Rank* button does not have a ruleset icon like web does:

image

This is generally probably going to be a bigger problem if we're aiming for 100% parity with web as the ruleset is taken from the user's selected primary ruleset on their profile. I'd be fine with it being a bindable for now if we decide to include it and populate later when that information is available from API (if it isn't already).

osu.Game/Overlays/OverlayPanelDisplayStyleControl.cs
EVAST9919: oops, true

EVAST9919:
Also the icon require some refactoring in the `OverlaySortTabControl` and pr will be more confusing.

bdach:
OK, if that is the case then we can go back to that after the eventual refactor.

osu.Game.Tests/Visual/UserInterface/TestSceneUserListToolbar.cs
EVAST9919: You selected wrong file. In `UserListToolbar` it has 10px spacing

bdach (approved):
Looks good for now.

smoogipoo (approved):

PR #7993, opened 2020-02-25
Adjust overlay header elements to match osu-web

thewildtree:
Sizing of the `OverlayHeader` and its elements was quite off from what can be seen on osu-web. This PR brings its values as close to the web counterpart as possible.

master:
image

this PR:
image

web:
image

Please keep in mind that this PR might make things look a bit *worse* for now since it still needs a font sizing fix to look exactly like osu-web. Until then, to mantain proper sizing and prevent things from looking *too* off, I added line-height compensations (margins) to some of the text elements (similar to the ones from `BeatmapSetOverlay` resize PR) that will be safe to remove when the font sizing fix comes.

That being said, here's a comparison with a temporary font sizing fix applied (web on the left, lazer on the right):

image

### DIFF
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -21,7 +21,7 @@ namespace osu.Game.Graphics.UserInterface
         private const float text_padding = 17.5f; // 15px padding + 2.5px compensation for line-height
 
         private SpriteIcon iconSprite;
-        protected readonly OsuSpriteText TitleText, PageText;
+        private readonly OsuSpriteText titleText, pageText;
         protected readonly Circle Separator;
 
         protected IconUsage Icon
@@ -37,18 +37,18 @@ namespace osu.Game.Graphics.UserInterface
 
         protected string Title
         {
-            set => TitleText.Text = value;
+            set => titleText.Text = value;
         }
 
         protected string Section
         {
-            set => PageText.Text = value;
+            set => pageText.Text = value;
         }
 
         public Color4 AccentColour
         {
-            get => PageText.Colour;
-            set => PageText.Colour = value;
+            get => pageText.Colour;
+            set => pageText.Colour = value;
         }
 
         public Color4 SeparatorColour
@@ -81,7 +81,7 @@ namespace osu.Game.Graphics.UserInterface
                             t.Origin = Anchor.Centre;
                             t.Margin = new MarginPadding { Horizontal = 5 }; // compensates for osu-web sprites having around 5px of whitespace on each side
                         }),
-                        TitleText = new OsuSpriteText
+                        titleText = new OsuSpriteText
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
@@ -96,7 +96,7 @@ namespace osu.Game.Graphics.UserInterface
                             Colour = Color4.Gray,
                             Margin = new MarginPadding { Top = 3 } // compensation for osu-web using a font here making the circle appear a bit lower
                         },
-                        PageText = new OsuSpriteText
+                        pageText = new OsuSpriteText
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -2,11 +2,13 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Graphics.Sprites;
+using osu.Game.Overlays;
 using osuTK;
 using osuTK.Graphics;
 
@@ -16,11 +18,12 @@ namespace osu.Game.Graphics.UserInterface
     {
         public const float ICON_WIDTH = ICON_SIZE + spacing;
 
-        public const float ICON_SIZE = 25;
-        private const float spacing = 6;
-        private const int text_offset = 2;
+        public const float ICON_SIZE = 30; // osu-web uses 40, but the images there aren't cropped edge-to-edge
+        private const float spacing = 10;
+        private const float text_padding = 17.5f; // 15px padding + 2.5px compensation for line-height
 
         private SpriteIcon iconSprite;
+        private readonly Circle circle;
         private readonly OsuSpriteText titleText, pageText;
 
         protected IconUsage Icon
@@ -72,31 +75,38 @@ namespace osu.Game.Graphics.UserInterface
                         {
                             t.Anchor = Anchor.Centre;
                             t.Origin = Anchor.Centre;
+                            t.Margin = new MarginPadding { Horizontal = 5 }; // compensates for osu-web sprites having around 5px of whitespace on each side
                         }),
                         titleText = new OsuSpriteText
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
                             Font = OsuFont.GetFont(size: 20, weight: FontWeight.Bold),
-                            Margin = new MarginPadding { Bottom = text_offset }
+                            Margin = new MarginPadding { Vertical = text_padding }
                         },
-                        new Circle
+                        circle = new Circle
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
-                            Size = new Vector2(4),
-                            Colour = Color4.Gray,
+                            Size = new Vector2(5),
+                            Margin = new MarginPadding { Top = 3 } // compensation for osu-web using a font here making the circle appear a bit lower
                         },
                         pageText = new OsuSpriteText
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
                             Font = OsuFont.GetFont(size: 20),
-                            Margin = new MarginPadding { Bottom = text_offset }
+                            Margin = new MarginPadding { Vertical = text_padding }
                         }
                     }
                 },
             };
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(OverlayColourProvider colourProvider)
+        {
+            circle.Colour = colourProvider.Foreground1;
         }
     }
 }
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -51,6 +51,12 @@ namespace osu.Game.Graphics.UserInterface
             set => PageText.Colour = value;
         }
 
+        public Color4 SeparatorColour
+        {
+            get => Separator.Colour;
+            set => Separator.Colour = value;
+        }
+
         protected virtual Drawable CreateIcon() => iconSprite = new SpriteIcon
         {
             Size = new Vector2(ICON_SIZE),
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -4,7 +4,6 @@
 using System;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
-using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Graphics.Sprites;
 using osuTK;
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -7,7 +7,6 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Graphics.Sprites;
-using osu.Game.Overlays;
 using osuTK;
 using osuTK.Graphics;
 
--- a/osu.Game/Graphics/UserInterface/ScreenTitle.cs
+++ b/osu.Game/Graphics/UserInterface/ScreenTitle.cs
@@ -21,8 +21,7 @@ namespace osu.Game.Graphics.UserInterface
         private const float text_padding = 17.5f; // 15px padding + 2.5px compensation for line-height
 
         private SpriteIcon iconSprite;
-        private readonly OsuSpriteText titleText, pageText;
-        protected readonly Circle Separator;
+        private readonly OsuSpriteText titleText, pageText, separator;
 
         protected IconUsage Icon
         {
@@ -53,8 +52,8 @@ namespace osu.Game.Graphics.UserInterface
 
         public Color4 SeparatorColour
         {
-            get => Separator.Colour;
-            set => Separator.Colour = value;
+            get => separator.Colour;
+            set => separator.Colour = value;
         }
 
         protected virtual Drawable CreateIcon() => iconSprite = new SpriteIcon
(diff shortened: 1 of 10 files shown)


osu.Game/Overlays/OverlayTabControl.cs
```diff
@@ -32,6 +33,20 @@ public override Color4 AccentColour
...
+            get => textColour;
+            set
+            {
+                textColour = value;
+
+                foreach (var i in TabContainer.Children.OfType<OverlayTabItem>())
+                    i.TextColour = value;
+            }
+        }
+
```
TheWildTree: I had to separate the text colour from the bar colour, but I'm not sure if this is the right way to do it.

peppy:
`OverlayTabControl` is confusing to me. It's derived in two places:

- `OverlayHeaderTabControl`, which sets `TextColour` but *not* `AccentColour`
- `ProfileTabControl`, which sets `AccentColour` but *not* `TextColour`

Is this correct? Also, is there a reason the colour specifications can't be moved local to `OverlayTabControl` itself, if it is to be used in overlays (it could get the colour provider locally then).

TheWildTree:
It's confusing indeed, but kind of makes sense:

- `OverlayHeaderTabControl` is the one we're restyling here; it's the tab control directly under the screen title and its text colour is different from its accent (bar) colour, 
- `ProfileTabControl` is only used in the profile overlay as the section controller ([this one](https://user-images.githubusercontent.com/27722894/75374291-47d5fb80-58cc-11ea-88ec-31fa803f6c4f.png)), it doesn't need any restyling so no changes here. I'll rename it to better reflect what it does, though, as it confused me as well.

I haven't noticed that earlier though and accidentally applied changes to the `ProfileTabControl` as well, will fix.

TheWildTree:
Adjusted the content X-axis margin - osu-web uses 50 everywhere except for the user profile page (where it uses 70) while lazer used 70 in all cases. 

Moved the base constant to the `OverlayHeader` for potential reference from other components and added the ability to adjust that margin in case we ever need to set a different one (like in `UserProfileOverlay` now).

osu.Game/Graphics/UserInterface/ScreenTitle.cs
```diff
@@ -72,27 +79,29 @@ protected ScreenTitle()
...
-                        new Circle
+                        Separator = new Circle
                         {
                             Anchor = Anchor.Centre,
                             Origin = Anchor.Centre,
-                            Size = new Vector2(4),
+                            Size = new Vector2(5),
                             Colour = Color4.Gray,
+                            Margin = new MarginPadding { Top = 3 } // compensation for osu-web using a font here making the circle appear a bit lower
                         },
```
bdach: I checked and it should be possible to fully mirror web implementation here:

```csharp
                        Separator = new OsuSpriteText
                        {
                            Anchor = Anchor.Centre,
                            Origin = Anchor.Centre,
                            Font = OsuFont.GetFont(size: 20),
                            Colour = Color4.Gray,
                            Margin = new MarginPadding { Vertical = text_padding },
                            Text = "\u2022"
                        },
```

osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -15,9 +15,18 @@ public abstract class OverlayHeader : Container
     {
         private readonly Box titleBackground;
         private readonly ScreenTitle title;
+        private readonly Container content;
 
         protected readonly FillFlowContainer HeaderInfo;
 
+        public const float CONTENT_X_MARGIN = 50;
+
+        public virtual float HorizontalMargin
```
bdach: Should be `protected`, get-only and used directly in the ctor. Basically make `CONTENT_X_MARGIN` into a protected get-only property.

osu.Game/Graphics/UserInterface/ScreenTitle.cs
```diff
@@ -36,18 +37,24 @@ protected IconUsage Icon
...
-            set => pageText.Colour = value;
+            get => PageText.Colour;
+            set => PageText.Colour = value;
+        }
+
+        public Color4 SeparatorColour
+        {
+            get => Separator.Colour;
+            set => Separator.Colour = value;
         }
```
bdach: Do these need to be a thing? Can't we just inject `OverlayColourProvider` locally?

osu.Game/Overlays/OverlayRulesetTabItem.cs
```diff
@@ -53,6 +53,8 @@ public OverlayRulesetTabItem(RulesetInfo value)
                         Origin = Anchor.Centre,
                         Anchor = Anchor.Centre,
                         Text = value.Name,
+                        Font = OsuFont.GetFont(size: 18),
+                        ShadowColour = new Color4(0, 0, 0, 0.75f)
```
bdach: Slightly more readable:
```suggestion
                        ShadowColour = Color4.Black.Opacity(0.75f)
```

osu.Game/Overlays/UserProfileOverlay.cs
```diff
@@ -149,30 +151,32 @@ private void userLoadComplete(User user)
...
                 TabContainer.Origin |= Anchor.x1;
             }
 
-            protected override TabItem<ProfileSection> CreateTabItem(ProfileSection value) => new ProfileTabItem(value)
+            protected override TabItem<ProfileSection> CreateTabItem(ProfileSection value) => new ProfileSectionTabItem(value)
             {
-                AccentColour = AccentColour
+                // This needs to be set here manually because TabControl attempts to propagate these values to its children before these tabs are added
+                AccentColour = AccentColour,
+                TextColour = TextColour
```
bdach: Can you explain further? I'm kind of confused at why this is necessary.

osu.Game/Overlays/OverlayTabControl.cs
```diff
@@ -55,19 +79,23 @@ protected class OverlayTabItem : TabItem<T>, IHasAccentColour
...
+                    if (textColour == value)
                         return;
 
-                    accentColour = value;
-                    Bar.Colour = value;
-
+                    textColour = value;
                     updateState();
                 }
             }
```
bdach: Again, can we not apply these locally with `OverlayColourProvider`?

osu.Game/Overlays/OverlayTabControl.cs
```diff
@@ -40,12 +56,20 @@ protected OverlayTabControl()
             AddInternal(bar = new Box
             {
                 RelativeSizeAxes = Axes.X,
+                Margin = new MarginPadding { Top = -1 },
```
bdach: Not sure what this is fixing but this is a hack. Probably not worth the one pixel.

osu.Game/Overlays/OverlayTabControl.cs
bdach: Ideally this shouldn't exist.

osu.Game/Screens/Multi/Header.cs
```diff
@@ -71,11 +72,20 @@ private void load(OsuColour colours)
 
         private class MultiHeaderTitle : ScreenTitle
         {
+            private const int text_offset = 2;
```
bdach: Single-use constant (can be inlined), and again, not sure why this is required.

bdach (changes requested):

osu.Game/Screens/Multi/Header.cs
TheWildTree: This was copied over from the previous version of `ScreenTitle`, it was used there to align the text to match the desired design I suppose? Can remove if not necessary anymore.

osu.Game/Overlays/UserProfileOverlay.cs
TheWildTree: At the moment these tabs are created, its parent `TabControl` has already attempted to set its children `AccentColour` / `TextColour` like here:
https://github.com/ppy/osu/blob/6b284fb3e74fbfce50c96e1096563360b95ba9cd/osu.Game/Graphics/UserInterface/OsuTabControl.cs#L29-L37

 This happens because they're being set when the `TabControl` is created, while the tabs in `ProfileSectionTabControl` are added a bit later (when their data is loaded up from the API):

https://github.com/ppy/osu/blob/6b284fb3e74fbfce50c96e1096563360b95ba9cd/osu.Game/Overlays/UserProfileOverlay.cs#L139-L149

`AccentColour` was already being set this way here, so I just added the `TextColour` which acts pretty much in the same manner.

bdach: That is immensely broken & yet another argument to just move `OverlayColourProvider` down to the tabs themselves.

osu.Game/Graphics/UserInterface/ScreenTitle.cs
TheWildTree: For some reason I thought we can't use unicode chars like this here (don't know why, haven't even tried tbh), thanks

osu.Game/Overlays/OverlayTabControl.cs
TheWildTree: Without this, the coloured bar is displayed below the background box:
image

while on web it's inside it:
image

osu.Game/Screens/Multi/Header.cs
bdach: If nobody knows why it's there let's just remove it.

osu.Game/Overlays/OverlayTabControl.cs
bdach: This is probably the proper way to get rid of this:
```csharp
            AddInternal(bar = new Box
            {
                RelativeSizeAxes = Axes.X,
                Height = 2,
                Anchor = Anchor.BottomLeft,
                Origin = Anchor.BottomLeft
            });
```

FWIW, same thing probably goes for the expanding bar below:
```csharp
                    Bar = new ExpandingBar
                    {
                        Anchor = Anchor.BottomCentre,
                        Origin = Anchor.Centre,
                        ExpandedSize = 7.5f,
                        CollapsedSize = 0,
                    },
```

osu.Game/Graphics/UserInterface/ScreenTitle.cs
TheWildTree: This one has to be a thing as long as `MultiHeaderTitle` exists in the current form since the multiplayer room screen doesn't have an OCP higher up in the hierarchy.

osu.Game/Overlays/UserProfileOverlay.cs
TheWildTree: I moved the text colour change down to the tab items but left `AccentColour` as is, since that's being set by the `OsuTabControl` higher up in the hierarchy and I don't think I want to change that.

osu.Game/Overlays/OverlayHeader.cs
TheWildTree: Sorry, I don't think I understood this properly, could you elaborate? 

I don't agree that `CONTENT_X_MARGIN` should be made into anything `protected` since it might be used by other components outside the `OverlayHeader` or any classes deriving from it (see current `UserProfileOverlay.CONTENT_X_MARGIN and how it's used by other parts of the overlay). Unless this can be achieved in a different way?

We also need the header to let us customize this value in case an overlay needs a different one - like the `UserProfileOverlay` does in this PR - that's why I made `HorizontalMargin` like this. I do indeed feel like there's a better way to do that, but I don't have any concrete ideas right now.

osu.Game/Overlays/OverlayTabControl.cs
TheWildTree: The bar fix you suggested does work, but the `ExtendedBar` is still 1px too low using these values:

image

osu.Game/Overlays/OverlayHeader.cs
bdach: OK, those consts can stay for now I suppose, although they feel super wrong. Here I mainly have an issue with exposition as public. IMO would be better to invert flow:

```csharp
protected virtual float HorizontalPadding => CONTENT_X_MARGIN;
```

and then instead at `content` creation pass `new MarginPadding { Horizontal = HorizontalPadding }`.

osu.Game/Overlays/UserProfileOverlay.cs
bdach: A half-change feels worse than the existing state of affairs to be honest. Let's bring this back as it was.

osu.Game/Overlays/OverlayTabControl.cs
bdach: I suppose we can bring back the -1 but I'd make the 2 height into a `public const EXPANDED_TAB_HEIGHT` and reference that instead by doing `-EXPANDED_TAB_HEIGHT / 2` so it's obvious what it does.

osu.Game/Graphics/UserInterface/ScreenTitle.cs
bdach: Alright. Could be worse all things considered.

osu.Game/Overlays/OverlayTabControl.cs
TheWildTree: Except the line we're seeing is using `Height = 1` as seen here:

https://github.com/ppy/osu/blob/6b284fb3e74fbfce50c96e1096563360b95ba9cd/osu.Game/Overlays/TabControlOverlayHeader.cs#L76-L85

I'll change this to be more readable (only the `ProfileSectionTabControl` uses `Height = 2` because having it in the base class is a bit misleading.

bdach: Yeah I got confused because I looked at `OverlayTabControl` instead of `OverlayHeaderTabControl` by mistake. Should be half regardless.

TheWildTree: Half of 1 doesn't really help us here though.

bdach: Does it not? Since sizes are floats I thought it might.

TheWildTree: We need a `-1` margin, `-0.5` looks just ever-so-slightly off at higher UI scales:

image image
(-0.5 on the left)

bdach: OK, sure.

osu.Game/Overlays/OverlayHeader.cs
TheWildTree: Changed, please check if it's better now.

bdach:
It's become apparent that `OsuTabControl`'s `AccentColour` brokenness and the overall byzantine hierarchy of the overlay tab controls needs urgent addressing, so I'm marking that as blocked for now and will be attempting to slowly fix it over the coming days.

osu.Game/Overlays/OverlayTabControl.cs
bdach: By the way, since that's important - *why* is it necessary exactly to separate the colours? Because if I do things with `AccentColour` then it'll probably be necessary to consider how this should behave too.

TheWildTree: See osu-web: in header tabs, separate colours are used for the text and for the bar below it. Previous lazer implementation didn't account for that and just used `AccentColour` for both.

Plus it's not consistent, header tabs use colours from the themed pallete while, for example, the profile section tabs just use plain white.

bdach: Actually yeah, it has a chance of working. If the `IAccentColourProvider` is injectable it can also be provided in BDL which is like 100% better since then there's one central place for overall colouring. Or even we could entirely disregard it and just use `OverlayColourProvider` exclusively directly in the overlay tabs.

TheWildTree: Second option sounds much easier (and actually doable in a few lines of code I think), that is if dropping `AccentColour` in its entirety despite deriving from `OsuTabControl` would be acceptable. If we went in that direction, we'd also have to make the bar below the tabs coloured in the same way, but that's a non-issue really.

bdach: It's not a good idea to ignore the interface members. Way I see it `AccentColour` on the tab containers needs to be fixed anyway as it's broken. If we want to disregard it entirely here than some higher-order refactoring needs to be made (decouple `AccentColour` entirely by extracting another abstract tab control) but that might paradoxically create even more owrk.

TheWildTree:
Closing for now, I'll split changes not related to the `TabControl` mess into different PRs to further work on them (design changes, mostly).

PR #7995, opened 2020-02-25
Fix crashes in catch upon ending combo with banana shower or juice stream

bdach:
Fixes a regression from #7973.

# Summary

Since introduction of `IgnoreJudgement` and its usage in `JuiceStream` and `BananaShower` the hard cast in `CatcherArea` that was used to check if the drawable hit object should cause the fruits on the plate to explode at the end of combo caused a hard crash instead, since `IgnoreJudgement` was no longer deriving from `CatchJudgement`.

Replace the hard cast with a soft pattern-matched cast.

# Remarks

One test scene was hijacked and another (very primitive) one written to ensure this does not regress again, as a smoke test of sorts. I shortened the duration of the stream on the banana shower scene to make it go a little faster since it'll actually run as part of CI now.

The ignore judgements can be safely, well, ignored in the context of determining whether an explosion should be shown, since the last nested hit objects [also have `LastInCombo` set to true](https://github.com/ppy/osu/blob/ae86fa367c4835eea328f89142dcddcfaebcd1f6/osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapProcessor.cs#L41-L42).

I both ran autoplay smoke tests on the other rulesets to check if anything else crashes and searched through source to make sure there were no other suspicious casts, and I found none.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneBananaShower.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneBananaShower.cs
@@ -29,6 +29,12 @@ namespace osu.Game.Rulesets.Catch.Tests
         {
         }
 
+        [Test]
+        public void TestBananaShower()
+        {
+            AddUntilStep("player is done", () => !Player.ValidForResume);
+        }
+
         protected override IBeatmap CreateBeatmap(RulesetInfo ruleset)
         {
             var beatmap = new Beatmap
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):

peppy:
Is there a reason the existing TestSceneAutoJuiceStream wasn't used for this?

bdach:
I figured it was too long for CI to wait for.

PR #5093, opened 2019-06-21
Fix position of gradient in beatmap carousel when scaling UI

nyquillerium:
Fixes https://github.com/ppy/osu/issues/4968

The issue was caused by https://github.com/ppy/osu-framework/issues/223. Changing this to a normal container with fixed relative positioning is an optimization compared to before anyway, so this is a more permanent solution for this issue.

Tested using a modified `TestSceneBeatmapCarousel`

PR #8006, opened 2020-02-26
Apply osu!-side invalidation changes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3307

The only interesting case is `SectionsContainer`, where I've chosen to invalidate when any child changes. This hasn't caused any issues in my testing, but if it proves to cause any problems we can fix this via better defining `ExpandableHeader` and do something similar to `GridContainer` (propagate the invalidation upwards manually).

Also reverts the changes in https://github.com/ppy/osu/pull/5093


Joehuu:
Can revert the changes made in https://github.com/ppy/osu/pull/5093.

PR #8010, opened 2020-02-26
Combine CommentsPage logic into CommentsContainer

EVAST9919:


osu.Game.Tests/Visual/Online/TestSceneOfflineCommentsContainer.cs
```diff
@@ -3,90 +3,52 @@
...
-        }
+        });
 
         [Test]
-        public void TestEmptyBundle()
+        public void TestLocalCommentBundle()
         {
-            AddStep("Create page", () => createPage(getEmptyCommentBundle()));
-            AddAssert("Dictionary length is 0", () => commentsPage?.DictionaryLength == 0);
-        }
```
bdach: What was the point of having these tests if they're going to get removed with this refactoring? The first one in particular was a *regression test*. This is unacceptable.

bdach (dismissed):

osu.Game/Overlays/Comments/CommentsContainer.cs
```diff
@@ -153,39 +156,44 @@ private void getComments()
...
-                Sort = { BindTarget = Sort },
-                Type = { BindTarget = type },
-                CommentableId = { BindTarget = id }
-            }, loaded =>
+                content.Add(new NoCommentsPlaceholder());
+                moreButton.Hide();
+                return;
+            }
+
+            var topLevelComments = AppendComments(response);
```
smoogipoo: Really not feeling this. Some comments are async-loaded while others parts aren't.

The loading part itself should be moved inside `AppendComments`. The other stuff done here can exist outside of the load.

smoogipoo (changes requested):

bdach (commented):
otherwise looks ok I suppose

osu.Game.Tests/Visual/Online/TestSceneOfflineCommentsContainer.cs
```diff
@@ -211,16 +178,18 @@ private void createPage(CommentBundle commentBundle)
...
-            {
-            }
-
             public new void AppendComments([NotNull] CommentBundle bundle) => base.AppendComments(bundle);
 
             public int DictionaryLength => CommentDictionary.Count;
+
+            public void ShowComments(CommentBundle bundle)
+            {
+                CommentCounter.Current.Value = 0;
```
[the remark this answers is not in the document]
EVAST9919: `ClearComments` called even if we are switching the sort which looks a bit weird. (when it jumps from `number` to 0 and then to this exact number again)

gagahpangeran:
What is the state of this PR?

I want to add pinned comment support and looks like this structure is easier to use.

smoogipoo (approved):
This can be refactored further, but I think it's fine to get in for the time being, so progress can be made in areas such as pinning.

PR #8021, opened 2020-02-27
Remove legacy DrawableHitObject state management

peppy:


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -221,54 +221,35 @@ private void addNested(DrawableHitObject hitObject)
...
+                    state.Value = newState;
                 }
-
-                if (state.Value != ArmedState.Idle && LifetimeEnd == double.MaxValue || HitObject.HitWindows == null)
-                    Expire();
             }
-            else
-                state.Value = newState;
 
-            UpdateState(newState);
```
smoogipoo: This should remain.

smoogipoo (changes requested):

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: Nevermind, was obsolete.

smoogipoo (approved):

PR #8031, opened 2020-02-27
Fix double-dimming in changelog stream badge area

### LINKED ISSUE #7998: The changelog overlay should just dimmer the selected category when hovering
lazer:
j55DoF8Hln
web:
aOEDOWI2tn

**osu!lazer version:** 
commit 1cb91b24278ff0815033cd17cfad718136a9a852

EVAST9919:
Fixes https://github.com/ppy/osu/issues/7998
nice

### DIFF
--- a/osu.Game/Graphics/UserInterface/ExpandingBar.cs
+++ b/osu.Game/Graphics/UserInterface/ExpandingBar.cs
@@ -13,17 +13,17 @@ namespace osu.Game.Graphics.UserInterface
     /// </summary>
     public class ExpandingBar : Circle
     {
-        private bool isCollapsed;
+        private bool expanded = true;
 
-        public bool IsCollapsed
+        public bool Expanded
         {
-            get => isCollapsed;
+            get => expanded;
             set
             {
-                if (value == isCollapsed)
+                if (value == expanded)
                     return;
 
-                isCollapsed = value;
+                expanded = value;
                 updateState();
             }
         }
@@ -83,19 +83,21 @@ namespace osu.Game.Graphics.UserInterface
             updateState();
         }
 
-        public void Collapse() => IsCollapsed = true;
+        public void Collapse() => Expanded = false;
 
-        public void Expand() => IsCollapsed = false;
+        public void Expand() => Expanded = true;
 
         private void updateState()
         {
-            float newSize = IsCollapsed ? CollapsedSize : ExpandedSize;
-            Easing easingType = IsCollapsed ? Easing.Out : Easing.OutElastic;
+            float newSize = expanded ? ExpandedSize : CollapsedSize;
+            Easing easingType = expanded ? Easing.OutElastic : Easing.Out;
 
             if (RelativeSizeAxes == Axes.X)
                 this.ResizeHeightTo(newSize, 400, easingType);
             else
                 this.ResizeWidthTo(newSize, 400, easingType);
+
+            this.FadeTo(expanded ? 1 : 0.5f, 100, Easing.OutQuint);
         }
     }
 }
--- a/osu.Game/Overlays/Changelog/ChangelogHeader.cs
+++ b/osu.Game/Overlays/Changelog/ChangelogHeader.cs
@@ -92,7 +92,7 @@ namespace osu.Game.Overlays.Changelog
                     AutoSizeAxes = Axes.Y,
                     Padding = new MarginPadding
                     {
-                        Horizontal = 85,
+                        Horizontal = 65,
                         Vertical = 20
                     },
                     Child = Streams = new UpdateStreamBadgeArea()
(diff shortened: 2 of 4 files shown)


EVAST9919:
First two commits weren't necessary, though, I just think that `TabControl` itself shouldn't store background and padding.

osu.Game/Overlays/Changelog/UpdateStreamBadge.cs
```diff
@@ -112,21 +103,23 @@ private void updateState()
...
             }
             else
             {
                 expandingBar.Collapse();
-                fadeContainer.FadeTo(0.5f, transition_duration);
+                expandingBar.FadeTo(0.5f, transition_duration, Easing.OutQuint);
             }
 
-            text.FadeTo(externalDimRequested && !IsHovered ? 0.5f : 1, transition_duration);
+            text.FadeTo(IsHovered || (Active.Value && !externalDimRequested) || allHighlighted ? 1 : 0.5f, transition_duration, Easing.OutQuint);
```
peppy: i have no idea what this condition is specifying. it's not legible

osu.Game/Overlays/Changelog/UpdateStreamBadge.cs
```diff
@@ -112,21 +103,23 @@ private void updateState()
             // Expand based on the local state
             bool shouldExpand = Active.Value || IsHovered;
 
+            bool allHighlighted = SelectedTab.Value == null && !externalDimRequested;
+
             // Expand based on whether no build is selected and the badge area is hovered
-            shouldExpand |= SelectedTab.Value == null && !externalDimRequested;
+            shouldExpand |= allHighlighted;
 
             if (shouldExpand)
```
peppy: i would probably just make a separate if section for `allHighlighted` rather than try and squeeze them in existing cases.

peppy (changes requested):
as proposed

osu.Game/Overlays/Changelog/ChangelogHeader.cs
```diff
@@ -72,7 +82,21 @@ private void showBuild(ValueChangedEvent<APIChangelogBuild> e)
...
+                {
+                    RelativeSizeAxes = Axes.Both
+                },
+                new Container
+                {
+                    RelativeSizeAxes = Axes.X,
+                    AutoSizeAxes = Axes.Y,
+                    Padding = new MarginPadding
+                    {
+                        Horizontal = 85,
```
bdach: Pretty sure this should be
```suggestion
                        Horizontal = 65,
```
<details>
<summary>Devtools screenshots</summary>

image

image
</details>

The paddings don't stack.

bdach (commented):
I can't see any immediately obvious issue with the dimming logic at its current state. Perhaps `externalDimRequested` could be renamed to `anyStreamHovered` to make code more obvious but it's a very minor nitpick so I'm not even going to argue for it.

peppy:
I've basically rewritten the conditionals and fixed some other weird code along the way.

peppy (approved):

osu.Game/Graphics/UserInterface/ExpandingBar.cs
```diff
@@ -13,17 +13,17 @@ namespace osu.Game.Graphics.UserInterface
...
             {
-                if (value == isCollapsed)
+                if (value == expanded)
                     return;
 
-                isCollapsed = value;
+                expanded = value;
                 updateState();
             }
         }
```
bdach: Now that you pointed this out this is basically a poor man's bindable, maybe we want to make it an actual one?

osu.Game/Overlays/Changelog/UpdateStreamBadge.cs
```diff
@@ -100,34 +100,41 @@ protected override void OnHoverLost(HoverLostEvent e)
...
+                    return;
 
-        public void DisableDim()
-        {
-            externalDimRequested = false;
-            updateState();
+                userHoveringArea = value;
+                updateState();
+            }
         }
```
bdach: Also could be a bindable?

peppy: it could, but i didn't see much point. it's very local.

osu.Game/Graphics/UserInterface/ExpandingBar.cs
peppy: still not sure there's much benefit to be had by converting these to bindables, but maybe? @smoogipoo thoughts?

smoogipoo: Very hard to say in these sort of situations. We've discussed properties vs bindables many times before and the best conclusion was "it's situational". I'm generally fine with leaving properties instead of bindables, unless there's more complex logic such as recursive property setting.

smoogipoo (approved):

osu.Game/Graphics/UserInterface/ExpandingBar.cs
bdach: Cool, I'll keep that guideline in mind going forward, thanks!

PR #8038, opened 2020-02-27
Adjust rankings overlay elements to better match osu-web

thewildtree:
## Summary

Brings spacing and overall style of the `RankingsOverlay` and its elements in line with osu-web. Just like other PRs of this kind, it will benefit greatly from a future font sizing fix.

Also brings back the usage of `LinkFlowContainer` to the country rankings while keeping country names non-italic, as mentioned in #7820.

Two things I'm not too sure about are the implementation of `GradeColumns` (in osu-web, they use different padding than the rest of the columns, so I had to separate them somehow) and the way I added spacing between table rows, but I think there currently isn't way to do this in a cleaner way without framework-side changes.

## Screenshots
osu-web vs. lazer with this PR:

image

Visible differences left: flags corner radius / shadows (out of scope for this PR), side content margins (being fixed in another PR), selector box-shadow / border (the desired 2px border causes ugly rendering issues in lazer because of masking), numbers not being humanized (not worth the hassle for now), links not being coloured. Will fix what I can in future PRs.


osu.Game/Overlays/Rankings/Tables/CountriesTable.cs
```diff
@@ -62,35 +61,19 @@ public CountriesTable(int page, IReadOnlyList<CountryStatistics> rankings)
...
-            private void load(OverlayColourProvider colourProvider)
-            {
-                IdleColour = colourProvider.Light2;
-                HoverColour = colourProvider.Content2;
+                AutoSizeAxes = Axes.X;
+                RelativeSizeAxes = Axes.Y;
+                TextAnchor = Anchor.CentreLeft;
 
-                Action = () => rankings?.ShowCountry(country);
+                AddLink(country.FullName ?? string.Empty, () => rankings?.ShowCountry(country));
```
TheWildTree: Now that I look at this, wouldn't it be better if it only added a link when the `country.FullName` has a value? I kinda mindlessly copied the text value from the previous version and didn't notice that until now, I'm also not sure if it can ever realistically be null (haven't checked).

TheWildTree:
This now depends on #7993 being merged to avoid conflicts since `SpotlightSelector` and `CountryFilter` needed changes to make use of the side margin fix from there.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -50,7 +51,7 @@ public IEnumerable<APISpotlight> Spotlights
         public SpotlightSelector()
         {
             RelativeSizeAxes = Axes.X;
-            Height = 100;
+            Height = 155;
```
bdach: I kind of disagree with setting this explicitly of making this `AutoSize`d along Y and just making the measurements below match. This is not an explicitly-set width in css.

osu.Game/Overlays/Rankings/SpotlightSelector.cs
```diff
@@ -162,11 +164,33 @@ private class SpotlightsDropdown : OsuDropdown<APISpotlight>
...
                 menu.BackgroundColour = colourProvider.Background5;
-                AccentColour = colourProvider.Background6;
+                Padding = new MarginPadding { Vertical = 20 };
+            }
+
+            private class SpotlightsDropdownHeader : OsuDropdownHeader
+            {
+                public SpotlightsDropdownHeader()
+                {
+                    Height = 48;
```
bdach: This probably shouldn't be explicit too as it's computed on web. This means it'll look slightly wrong in the short run but will hopefully go back to being correct when font sizes are corrected for.

osu.Game/Overlays/Rankings/Tables/CountriesTable.cs
bdach: It would.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -47,10 +48,11 @@ private void load()
             {
                 RelativeSizeAxes = Axes.Both,
                 Depth = 1f,
-                Margin = new MarginPadding { Top = row_height }
+                Margin = new MarginPadding { Top = row_height + row_spacing },
+                Spacing = new Vector2(0, row_spacing),
             });
 
-            rankings.ForEach(_ => backgroundFlow.Add(new TableRowBackground()));
+            rankings.ForEach(_ => backgroundFlow.Add(new TableRowBackground(row_height)));
```
bdach: Not much point passing this in the ctor if you have already-public properties to work with:
```suggestion
            rankings.ForEach(_ => backgroundFlow.Add(new TableRowBackground { Height = row_height }));
```

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -84,31 +86,36 @@ private void load()
...
+            public HeaderText(string text, string highlighted, IEnumerable<string> gradeColumns)
             {
                 this.highlighted = highlighted;
 
                 Text = text;
                 Font = OsuFont.GetFont(size: 12);
-                Margin = new MarginPadding { Horizontal = 10 };
+
+                var isGrade = gradeColumns.Contains(text);
+                Margin = new MarginPadding { Vertical = 5, Horizontal = isGrade ? 20 : 10 };
```
bdach: The header text should know nothing about grade columns. Set the margin higher-up explicitly however you need it.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -35,7 +36,7 @@ protected RankingsTable(int page, IReadOnlyList<TModel> rankings)
             AutoSizeAxes = Axes.Y;
 
             Padding = new MarginPadding { Horizontal = horizontal_inset };
-            RowSize = new Dimension(GridSizeMode.Absolute, row_height);
+            RowSize = new Dimension(GridSizeMode.Absolute, row_height + row_spacing);
```
bdach: This is hacky as all heck. I know that this is all you have to work with that maybe we need framework-level changes here to allow grid row/column spacing.

bdach (changes requested):

osu.Game/Overlays/Rankings/SpotlightSelector.cs
TheWildTree: This has to be explicitly set because autosizing causes the dropdown to change the height of the whole selector area when expanded (not to mention other layout issues):

osu.Game/Overlays/Rankings/SpotlightSelector.cs
TheWildTree: Removed this and temporarily added padding to the text sprite to compensate for the size differences (like in other adjustment PRs).

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
TheWildTree: Done, also changed the highlight logic a bit to only pass a bool instead of the whole highlighted column string.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
TheWildTree: I created an issue framework-side for that (ppy/osu-framework#3330).

I know this is hacky, but I don't feel it's *that* bad - maybe it could stay as a temporary solution until we have proper framework-side support (which I don't think is gonna happen too quickly, I presume there are more important issues to work on)

bdach: Let's get it in for now, but with sights set at potentially replacing it in the future.

TheWildTree:
Removed the dependency because it was only added to avoid merge conflicts with the aforementioned PR - since it's blocked now, that shouldn't be an issue anymore.

osu.Game/Overlays/Rankings/Tables/CountriesTable.cs
```diff
@@ -62,35 +61,20 @@ public CountriesTable(int page, IReadOnlyList<CountryStatistics> rankings)
...
-            private void load(OverlayColourProvider colourProvider)
-            {
-                IdleColour = colourProvider.Light2;
-                HoverColour = colourProvider.Content2;
+                AutoSizeAxes = Axes.X;
+                RelativeSizeAxes = Axes.Y;
+                TextAnchor = Anchor.CentreLeft;
 
-                Action = () => rankings?.ShowCountry(country);
+                if (country.FullName != null)
```
bdach:
```suggestion
                if (!string.IsNullOrEmpty(country.FullName))
```
More robust that way.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -68,13 +70,20 @@ private void load()
 
         protected abstract Drawable[] CreateAdditionalContent(TModel item);
 
-        protected override Drawable CreateHeader(int index, TableColumn column) => new HeaderText(column?.Header ?? string.Empty, HighlightedColumn());
+        protected override Drawable CreateHeader(int index, TableColumn column)
+        {
+            var title = column?.Header ?? string.Empty;
+            var isHighlighted = HighlightedColumn() == title;
+            var isGrade = GradeColumns().Contains(title);
```
bdach: I don't think these are acceptable. Leaving aside the mind-boggling design of `HighlightedColumn()` which existed previously, `GradeColumns()` is similarly strange.

* More idiomatic to have it as a property and not a method.
* More importantly, the premise that a generic rankings table should know what is or isn't a grade column is flawed and doing table-specific styling based on that is also flawed to me. Table-specific stylings should be applied in table-specific places. This implementation tries to be way too generic with what it does, therefore introducing 2 levels of inheritance for 4 usage cases and ending up being unreadable in the process:
```
RankingsTable
  CountriesTable
  UserBasedTable
    PerformanceTable
    ScoresTable
```

IMO we can't proceed with this code without de-generifying it.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -68,13 +70,20 @@ private void load()
...
+
+            return new HeaderText(title, isHighlighted) { Margin = new MarginPadding { Vertical = 5, Horizontal = isGrade ? 20 : 10 } };
```
bdach: This long line is too long, when it could be split across object initialiser braces.

bdach (changes requested):
Spotlight styling has changed in the meantime at ppy/osu-web@ecd4845: 

image

(note alignment). Column spacing is also wrong (is 15, should be 20).

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
bdach: As an aside, the header colours are actually flipped right now:
image

TheWildTree: Moved the grade-column parts to `UserBasedTable` as that's where they're used (my bad for not noticing that in the first place), I'm far from being sure that this is the right way to do it, but I didn't have any better ideas.

Header colours fixed.

bdach (dismissed):
Not too happy with `HiglightedColumn` and `GradeColumns` still existing but I suppose this is fine for now. Can be refactored at some later date.

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
```diff
@@ -68,13 +70,19 @@ private void load()
 
         protected abstract Drawable[] CreateAdditionalContent(TModel item);
 
-        protected override Drawable CreateHeader(int index, TableColumn column) => new HeaderText(column?.Header ?? string.Empty, HighlightedColumn());
+        protected virtual string HighlightedColumn => @"Performance";
```
smoogipoo: Can't have this be a string. How about index instead?

smoogipoo (changes requested):

osu.Game/Overlays/Rankings/Tables/RankingsTable.cs
bdach: FWIW that was there before (check a few lines down, it was a method before). I'd have to check but index might not be feasible since these tables are inheritance hell and like two or three tables used the same base class for some reason, and I think index of this column might be different between them. Ideally I'd say these tables should be separated entirely (even if by cost of some code duplication) but that is not immediately trivial.

smoogipoo: If that's the case I'd like to see that happen first, since that would also probably change how this PR looks. As it is I'm concerned localisation will immediately break this, also it seems a little fragile w.r.t. capitalisation, spelling, etc.

TheWildTree: I agree that this mechanism isn't the best (same with `GradeColumns`), but I couldn't really come up with a solution to the inheritance problem that'd be any kind of acceptable so far.

Right now some changes seem to be planned for the rankings section ([see here](https://github.com/ppy/osu-web/pull/5889#issuecomment-614512400)), so I think this whole PR has to wait for them anyway. I'll try to work on these goddamn tables in the meantime.

peppy: This is still an issue, but I'm going to look past it and get this change merged as it is generally improving the code quality around this area (which was already pretty shocking).

Note that this isn't adding a string but just moving it up from lower in the class, where it still exists on `master`.

peppy (approved):
Seems non-offensive

PR #8042, opened 2020-02-28
Properly colour links in FullscreenOverlays

### LINKED ISSUE #7822: LinkFlowContainer should be able to use OverlayColourProvider.
Just a tracking issue since it's planned (as mentioned in https://github.com/ppy/osu/pull/7668).

thewildtree:
Fixes #7822.

Small change code-wise but brings a big improvement in the overall look and feel of `FullscreenOverlay`s. Only changes the idle colour, links still turn yellow when hovering.

### Beatmap score display:

before:
image

after:
image

### Recent plays:

before:
image

after:
image


osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -31,10 +32,16 @@ public DrawableLinkCompiler(IEnumerable<Drawable> parts)
             Parts = parts.ToList();
         }
 
+        [Resolved(CanBeNull = true)]
+        private OverlayColourProvider colourProvider { get; set; }
+
+        [Resolved]
+        private OsuColour colours { get; set; }
```
EVAST9919: Why these have been moved outside of `load()`?

EVAST9919: Also mapper name colour in Most played beatmaps in UserOverlay is a bit off

TheWildTree: ColourProvider had to be moved out because it can be null (unless you can do that directly in `load()`?) and OsuColour got moved out for consistency, but you're right, it could've stayed where it was

EVAST9919: you can make stuff nullable in load (check `AppendComments` in `CommentsPage.cs`)

TheWildTree: thanks, I should've checked that earlier. i'll fix that and will look for any other colours being off due to this PR.

bdach: FWIW there's been a PR merged recently that replaced BDL injection with `[Resolved]` wherever possible (#7837). I would like to avoid going back and forth on this stylistic choice.

TheWildTree: yup, they aren't needed anywhere outside the `load()` so i'll move them back there.

TheWildTree:
Added an option to make `DrawableLinkCompiler` ignore the colour provider because there might be places where we want to keep the links in blue (only occurence I've spotted so far was the `DirectGridPanel`). 

Also adjusted the most played section text - it's now in line with other parts, although still a bit off vs. osu-web because it uses a different text colour for the link there and we can't do that in lazer for now. I think it can stay like that as long as it's consistent with other links.

TheWildTree:
Sorry for the force-push, turns out the most played section was a bit off colour-wise so I decided to move the change mentioned above to a separate adjustment PR to avoid merge conflicts here.

bdach:
A revert would have been fine, but at least the force doesn't seem to have broken CI (it builds wrong commits sometimes when forcing - I recall it had something to do with it building the latest commit time-wise instead of `BRANCH_HEAD`).

bdach:
I kind of dislike the notion of `IgnoreColourProvider` being a property here and my nebulous reasoning is that it feels "too specific". If we want to be able to switch between having theming or no theming in places (which is arguable from a UX/design standpoint, but that is not my domain), I feel like there should be two separate components - `LinkFlowContainer` should mainly stay as-is and a more specific one (`ThemedLinkFlowContainer`? `OverlayLinkFlowContainer`?) would then definitively take in the colour provider and use it without having any extra switching logic inside.

TheWildTree:
My original idea was similar, but it was said in the linked issue to keep this simple. It was without taking into account that we'd need to switch between themed and non-themed versions though, so I guess in this situation a specific, themed `LinkFlowContainer` won't be a bad idea. Will see what I can do.

bdach:
I mean I would agree with that to the point of a5c004e where that flag wasn't there, but now that we have an injected provider that overrides the colour and then a flag that undoes the override, it seems not too great to me. Let's double check our opinions here first though - @smoogipoo, @peppy, thoughts?

TheWildTree:
I thought about this for a while and in my opinion, a separate class for this would indeed be a good idea. Seeing as osu-web uses different link colours in some cases (mostly to make the overall design look more uniform), we are also in need of something that would let us match that lazer-side.

I agree with your initial proposal - `LinkFlowContainer` should IMO stay as-is, while a new one (`ColouredLinkFlowContainer`?) should handle all the non-blue link cases: it would allow us to supply a specific colour to use for the link, or in case of no colour provided (which would mostly be the case) fall back to using `OverlayColourProvider`s `Light2`.

smoogipoo:
I agree with the notion of separating it out into a separate class. Either `ColouredLinkFlowContainer` or `OverlayLinkFlowContainer` would be fine imo.

TheWildTree:
One more thing - should the yellow colour on hover be kept in all cases or should we allow to customize that as well?

bdach:
I'd say the overlay-themed links need to follow theme at all times, and that includes hover colour.

TheWildTree:
The version I just pushed is probably *far* from perfect - but I couldn't come up with anything better without external help, so I decided to push it here and slowly improve it with the help from more experienced people. 

It does what it's supposed to, though - once you replace a `LinkFlowContainer` with its `Coloured` version, it uses the `OverlayColourProvider` by default while also letting you supply a custom idle / hover colour - I tried my best to make the colour switching work at any moment, both before you add links and after that.

It's not hooked up anywhere yet - I'll do that once it's decided whether this is a good way to implement the desired behaviour.

osu.Game/Online/Chat/ColouredDrawableLinkCompiler.cs
```diff
@@ -0,0 +1,46 @@
...
+            set => base.IdleColour = idleColour = value;
+        }
+
+        private Color4 hoverColour;
+
+        public new Color4 HoverColour
+        {
+            get => hoverColour;
+            set => base.HoverColour = hoverColour = value;
+        }
```
TheWildTree: Not too sure about the whole colour setting logic - I didn't want to make these fields from `OsuHoverContainer` public, so I decided to overwrite them this way. Is there a better way to achieve what we need here?

bdach:
> It does what it's supposed to, though - once you replace a `LinkFlowContainer` with its `Coloured` version, it uses the `OverlayColourProvider` by default while also letting you supply a custom idle / hover colour - I tried my best to make the colour switching work at any moment, both before you add links and after that.

I'll come back and do a proper review later, but this is entirely unnecessary. Customisation of themed links with public properties should not be allowed, period, unless there is a case of links having different hover/idle colours in different places.

TheWildTree:
It kinda is, though - web has (or as least the old design had) a couple places where links were coloured a bit differently. Should we just ignore that then?

bdach:
That is extremely annoying. I've already struggled with this a bit in the other PR where ideally I'd have no setters at all and just specify the shades (members of `OverlayColourProvider`) to use declaratively. What I'd like to see in that case is something like this:
```csharp
new ColouredLinkFlowContainer
{
    IdleColour = colourProvider => colourProvider.Light2,
    HoverColour = colourProvider => colourProvider.Light4
}
```
I kind of dislike those being of type `Func<OverlayColourProvider, Color4>` though. Ideally you'd just have an enum. But that's a separate PR for that to be possible.

TheWildTree:
That would be great because it'd allow custom colouring while still strictly adhering to the `OverlayColourProvider` scheme, but just as you said, that's out of scope for this one.

I mentioned the non-OCP colouring in one of my earlier comments and implemented that because no one explicitly told me not to. It felt a bit annoying for me too, but oh well - this kind of customization might turn out useful in the future if it goes through.

One thing I'm thinking about is maybe combining the `IdleColour` and `HoverColour` properties into one - it would use the provided colour in the idle state and apply something like `.Lighten()` when hovering to ensure consistent colour differences.

peppy:
I think the code in its current state looks fine for now. Not against the proposed direction (Func with colour provider) as a future excursion.

peppy:
@TheWildTree please go ahead and hook things up again 👍

TheWildTree:
Current version turned out to be not-so-fine, I ran into the same colour assignment logic problem as we did a while ago in the `OsuTabControl` (#8117), resulting in the idle / hover colours not being properly assigned if links are added after the colour assignment happens (which is quite often).

I did kind of solve that (by passing current colours to the `ColouredDrawableLinkCompiler` when creating it), but another issue popped up: the base `DrawableLinkCompiler` kept setting the `IdleColour` to blue in its BDL, overriding whatever colour we set when constructing our coloured variant of that class.

My solution to that feels hacky but I couldn't come up with anything better.

I also tried to experiment with using `Bindable`s here, which might've been a cleaner solution to the first problem described above, but couldn't get them to work as intended.

Hooked up the `ColouredLinkFlowContainer` wherever it was needed, with rankings table being an exception since it's being dealt with in another PR and I'll add changes there to avoid merge conflicts.

osu.Game/Graphics/Containers/ColouredLinkFlowContainer.cs
```diff
@@ -0,0 +1,54 @@
...
+        public Color4 HoverColour
+        {
+            get => hoverColour;
+            set
+            {
+                hoverColour = value;
+                foreach (var linkCompiler in InternalChildren.OfType<ColouredDrawableLinkCompiler>())
+                    linkCompiler.HoverColour = value;
+            }
+        }
```
bdach: Yeah, definitely not a fan of these.

> I also tried to experiment with using `Bindable`s here, which might've been a cleaner solution to the first problem described above, but couldn't get them to work as intended.

What was the problem?

TheWildTree: Couldn't get colour value propagation to work properly, though that was probably just my lack of experience with bindables, and I didn't want to spend too much time on something that I wasn't sure will work. I'll try to approach that again later.

bdach: After a short look at the situation the protected-ness of `{Idle,Hover}Colour` is the main problem here. If the base classes do defaults in BDL then there's really no good way to reliably override them. Maybe what is desired here is something that makes the default more explicit, i.e.

```csharp
class OsuHoverContainer
{
        [Resolved]
        private OsuColour colours { get; set; };

        private Color4? hoverColour;

        protected Color4 HoverColour
        {
            get => hoverColour ?? colours.Yellow;
            set => hoverColour = value;
        }

}
```

and the same thing for `IdleColour`. Might be better off as a separate PR, though?

PR #8044, opened 2020-02-28
Implement an online-only view layer

Game4all:
# Summary
This implements a `OnlineViewLayer` class geared towards use within online overlays by providing them a diming layer blocking interactions with the overlay contents and displays a centered login placeholder if the local user isn't logged in.

## How it looks with the current online overlays:
Beatmap Listing | Dashboard | Rankings
-----------------|------------|----------
listing | dashboard | rankings

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneOnlineViewLayer.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneOnlineViewLayer.cs
@@ -16,7 +16,7 @@ using osuTK.Graphics;
 namespace osu.Game.Tests.Visual.Online
 {
     [TestFixture]
-    public class TestSceneOnlineViewLayer : ManualInputManagerTestScene
+    public class TestSceneOnlineViewLayer : OsuTestScene
     {
         public override IReadOnlyList<Type> RequiredTypes => new[]
         {
@@ -25,11 +25,7 @@ namespace osu.Game.Tests.Visual.Online
 
         private readonly Container con;
 
-        private readonly OsuButton button;
-
         private readonly TestOnlineViewLayer view;
-
-        private bool inputBlocked;
 
         public TestSceneOnlineViewLayer()
         {
@@ -48,13 +44,13 @@ namespace osu.Game.Tests.Visual.Online
                                 RelativeSizeAxes = Axes.Both,
                                 Colour = Color4.Blue,
                             },
-                            button = new OsuButton
+                            new OsuButton
                             {
                                 Anchor = Anchor.Centre,
                                 Origin = Anchor.Centre,
                                 Size = new Vector2(500, 500),
                                 Text = "Click me!",
-                                Action = () => { inputBlocked = true; }
+                                Action = () => { }
                             }
                         }
                     },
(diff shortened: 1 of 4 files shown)


osu.Game.Tests/Visual/Online/TestSceneOnlineViewLayer.cs
```diff
@@ -0,0 +1,108 @@
...
+            AddAssert("content is dimmed", () => con.Colour != Color4.White);
+            AddAssert("loading animation is not visible", () => !view.LoadingSpinner.IsPresent);
+        }
+
+        [Test]
+        public void TestConnectingStateVisibility()
+        {
+            AddStep("set status to connecting", () => ((DummyAPIAccess)API).State = APIState.Connecting);
+            AddWaitStep("wait for animation to finish", 5);
+            AddAssert("content is dimmed", () => con.Colour != Color4.White);
```
bdach: Might also be worth checking that the button underneath is not interactable.

osu.Game/Online/OnlineViewLayer.cs
```diff
@@ -0,0 +1,112 @@
...
+        public override bool HandleNonPositionalInput => false;
+
+        protected override bool Handle(UIEvent e) => true;
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Children = new Drawable[]
+            {
+                placeholder = new LoginPlaceholder(placeholderMessage)
```
bdach: Doesn't really make sense to have a `LoginPlaceholder` be used here, in a general `OnlineViewLayer` - should be either abstracted away to just a `Placeholder` or renamed to reflect the new more-general purpose.

osu.Game.Tests/Visual/Online/TestSceneOnlineViewLayer.cs
```diff
@@ -0,0 +1,108 @@
...
+                    view = new TestOnlineViewLayer("Please sign in to view content!", con)
+                }
+            };
+        }
+
+        [Test]
+        public void TestOfflineStateVisibility()
+        {
+            AddStep("set status to offline", () => ((DummyAPIAccess)API).State = APIState.Offline);
+            AddWaitStep("wait for animation to finish", 5);
```
bdach: Probably more reliable to use an `UntilStep`. No need to expose the placeholder either, can use `ChildrenOfType<>`.

osu.Game/Online/OnlineViewLayer.cs
```diff
@@ -0,0 +1,112 @@
...
+
+namespace osu.Game.Online
+{
+    /// <summary>
+    /// A layer which displays on top of content which require a local user to be logged in.
+    /// Blocks input to the underlying content, dims the optional target content and displays a login placeholder when the user isn't logged in.
+    /// </summary>
+    public class OnlineViewLayer : VisibilityContainer, IOnlineComponent
+    {
+        protected LoadingSpinner LoadingSpinner;
```
bdach: If this is exposed just for the test, might as well use `ChildrenOfType<>` for that too instead and not expose.

bdach (changes requested):
Out of curiosity, have you tried to apply this in more places and is changing to a layer a conclusion you've arrived to after trying that? I'm a bit worried that we'll flip-flop between the two variants.

Game4all:
It may be better as you pointed out to let both exist as variants.

bdach:
If that ends up being the case they should share logic as much as possible, but then I don't think they *both* should be permitted to exist if it's at all avoidable.

Game4all:
The only place where I found an usage for OnlineViewLayer for now is for online overlays where it is preferable to dim content instead of hiding it. 
The logic could have been embedded in shared a class if all online-overlays inherited a same class but since the ChatOverlay doesn't, it needs to exist as a variant to OVC.

peppy:
@Game4all i know this is a bit out of date now and didn't really get reviewed in time. We now have an `OnlineView<>` class for the new dashboard setup which can potentially become one of these. Did you want to give it a shot and see if it fits well?

Game4all:
Sounds fine, will give it a shot to try integrating offline behaviour handling. It get the feeling that this may basically supersede what this PR proposed if it turns to work well. (Although it may require quite a bit of refactoring to make use of it in all online-only overlays)

Game4all:
@peppy After messing a bit with `OverlayView<>`, I get the feeling that `OverlayView<>` can definitely supersede this PR contents if it were to have the offline logic handling added directly to it, altough it may require tweaking transitions. Thus, I'm closing this and will start from scratch implementing the behaviour proposed by this PR in `OverlayView<>`.

PR #8047, opened 2020-02-28
Adjust most played beatmaps section to better match osu-web

thewildtree:
Background colour was off (on osu-web it uses different values than top scores / first place ranks) + the author text was a bit dimmed due to weird colouring method (spotted [here](https://github.com/ppy/osu/pull/8042#discussion_r385720687)).

before:
image

after:
image


osu.Game/Overlays/Profile/Sections/ProfileItemContainer.cs
```diff
@@ -16,12 +16,19 @@ public class ProfileItemContainer : Container
...
-
         private readonly Box background;
         private readonly Container content;
 
+        private Color4 idleColour;
+
+        public Color4 IdleColour
+        {
+            get => idleColour;
+            set => idleColour = background.Colour = value;
```
bdach: This compound assignment is sort of wrong compared to the previous implementation as the setter will overwrite hover colour if called while the item is hovered.

osu.Game/Overlays/Profile/Sections/ProfileItemContainer.cs
```diff
@@ -16,12 +16,19 @@ public class ProfileItemContainer : Container
...
+        }
+
+        public Color4 HoverColour { get; set; }
```
bdach: I'm not sure I like these being public too much, the component is kind of tightly coupled to the theme so just being able to set any colour feels off. In general `DrawableMostPlayedBeatmap` is awkwardly written imo, as the `ProfileItemContainer` should be the top-level component there (or even, it should inherit from it). I'd envision something like the following:

```
DrawableMostPlayedBeatmap : ProfileItemContainer
    GridContainer
        Column (GridSizeMode.Absolute, width = cover_width)
            UpdateableBeatmapSetCover
        Column (GridSizeMode.Distributed)
            FillFlowContainer
                MostPlayedBeatmapMetadataContainer
                LinkFlowContainer
        Column (GridSizeMode.AutoSize)
            PlayCountText
```

Then you can make these `protected`. I'm aware this is a little bit out there in terms of work required but I believe it'll lead to cleaner code.

bdach (changes requested):

osu.Game/Overlays/Profile/Sections/ProfileItemContainer.cs
TheWildTree: Solved by checking for the container being hovered (and adding similar check to `HoverColour` as well) - hopefully that's enough.

osu.Game/Overlays/Profile/Sections/ProfileItemContainer.cs
TheWildTree: Solved this in a bit simpler way - added a `MostPlayedBeatmapContainer` deriving from the `ProfileItemContainer` (replacing its existing usage here) and set its colours in its own BDL, allowing mentioned fields to be set to protected. 

I know the whole component is still in an awkward state, but I'd wanna avoid rewriting it if possible (for now).

bdach (approved):
I've made a slight change in the background colour application - having three places do the same (however simple) logic felt like a step backwards, so I refactored it into a method. Aware that this might mean there are unnecessary assignments, but it's a trade-off. Other than that looks good for now 👍

PR #8050, opened 2020-02-28
Fix crash when reaching results screen on single threaded execution mode

### LINKED ISSUE #8043: Can't complete beatmaps on mobile
Version: 2020.228.0
https://i.imgur.com/zHcnm2X.jpg

After finishing beatmap there is no score screen, osu just freezes.

peppy:
Resolves #8043.


bdach (approved):
Tested as working on Android & works in my head, for however much that's worth.

smoogipoo (approved):

PR #8051, opened 2020-02-29
Replace SocialOverlay with DashboardOverlay

### LINKED ISSUE #6939: "Show Online" button in social browser
**Describe the new feature:** 
In the Social Browser in osu!stable, it only shows online players.
I feel like this needs to be an option, if not planned already in osu!lazer's social browser, as i always message friends and its slightly annoying to have to search for that "online" tag.

image
Just a checkbox around here, i guess

### LINKED ISSUE #7391: Sorting by friend rank in Social browser is broken
**Describe the bug:** 
When sorting by rank in the friends tab, it's still sorting by Alphabet

Screenshot

**osu!lazer version:** 
Local debug - current master branch

**Logs:**

### LINKED ISSUE #7533: Searching in social browser doesn't work
**Describe the bug:** 
Searching in social browser doesn't work. When typing something into the search box, the player list "refreshes", but it doesn't change according to the search input.

The initial player list looks like this
osu_2020-01-14_19-23-33
Typing someone's name in the search box, the player list refreshes but doesn't change (and not showing the searched player)
osu_2020-01-14_19-25-58

**osu!lazer version:**
2020.112.0

**Logs:** 
[logs.zip](https://github.com/ppy/osu/files/4058858/logs.zip)

### LINKED ISSUE #7941: Social browser should match web friends page
In line with DirectOverlay -> BeatmapListingOverlay changes

EVAST9919:
Closes https://github.com/ppy/osu/issues/7941
Closes https://github.com/ppy/osu/issues/6939
Closes https://github.com/ppy/osu/issues/7533 (since being replaced)
Closes https://github.com/ppy/osu/issues/7391


osu.Game/Overlays/Home/HomeOverlayHeader.cs
```diff
@@ -0,0 +1,49 @@
...
+            protected override Drawable CreateIcon() => new ScreenTitleTextureIcon(@"Icons/rankings");
+        }
+    }
+
+    public enum HomeOverlayTabs
+    {
+        Dashboard,
+        Friends,
+
+        [Description(@"forum subscriptions")]
```
bdach: If `.ToLower()` is going to end up being called anyways then might as well have proper casing here.

osu.Game/Overlays/TabControlOverlayHeader.cs
```diff
@@ -90,7 +92,7 @@ private class OverlayHeaderTabItem : OverlayTabItem
                 public OverlayHeaderTabItem(T value)
                     : base(value)
                 {
-                    Text.Text = value.ToString().ToLower();
+                    Text.Text = ((value as Enum)?.GetDescription() ?? value.ToString()).ToLower();
```
bdach: This should be (a) probably split to a method, since it's getting unreadable, and something like this even could go miles for readability:
```csharp

public string FormatHeaderText(object value)
{
    string baseText;

    switch (value)
    {
        case string stringValue:
            baseText = stringValue;
            break;
        case Enum enumValue:
            baseText = enumValue.GetDescription();
            break;
        default:
            baseText = enumValue.ToString();
            break;
    }

    return baseText.ToLower();
}
```
and I'd even go further to say (b) this needs to be made a global extension helper as the same sort of logic is already popping up in other places (see `HomeTitle`).

bdach (changes requested):
This feels too early for being merged. There's next to no actual content in the overlay.

EVAST9919:
> This feels too early for being merged. There's next to no actual content in the overlay.

Well, yes, but there's nothing wrong to have it in tests(?)

osu.Game/Overlays/TabControlOverlayHeader.cs
EVAST9919: Yeah, I thought to move it to some kind of a helper class as well, but wasn't sure

EVAST9919: May be it's a good idea to just move it to `TabControlOverlayHeader`. It will be used only once there

EVAST9919:
Also we only need `FriendsLayout` to be completed and then we will replace the `SocialOverlay`. And after that we can implement other layouts one by one and unlock them when ready.

bdach:
The progression feels backwards. Should be going bottom-up (smallest components first) and not top-down IMO, but that might just be me.

EVAST9919:
I agree, I will implement them one-by-one, but it's easier for me to see how it will looks and feels like in the overlay itself right away rather than adjust them later after the implementation.

EVAST9919:
Will reopen after https://github.com/ppy/osu/pull/8108 merge

EVAST9919:
Oh, actually it's the dashboard overlay now, will rename

smoogipoo (changes requested):
Please split out the friends layout part of this into its own PR. I've done a full pass over this but there's quite a lot of changes and it's quite chunky to read.

peppy:
I've done a basic merge of this for you, but there's plenty of further changes required which I'm going to leave to you:

- Because of the way you've structured these "Layouts" (please use the term "Display" going forward) they are not online components, and do not correctly handle API state changes. You need to add this functionality.
- Keep in mind that this applies to all your "layout" classes. At which point if you are following along you'd realise it's basically trying to be a FullscreenOverlay without being one. Maybe this is okay, but something to keep in mind.
- There's a loading spinner at dashboard. But then you also have on in the friend display. I'd recommend removing the spinner from the dashboard.
- The loading spinner on friends only displays when going from nothing to something. seems a bit wrong? Is this due to layout issues?

EVAST9919:
I accidentally broke this branch... Will apply requested changes in the new one.

peppy:
how?

EVAST9919:
I had a local outdated commit and accidentally pushed it. So some magic happened and now I can't see your latest changes to this branch even after full branch re-pull

EVAST9919:
Nevermind, I figured this out. Only with force-push, sorry

EVAST9919:
> * Because of the way you've structured these "Layouts" they are not online components, and do not correctly handle API state changes. You need to add this functionality.

This logic handled by the overlay itself (https://github.com/ppy/osu/pull/8051/files#diff-2ba431d9d74d076aa7dc226deb3b12a6R81, https://github.com/ppy/osu/pull/8051/files#diff-2ba431d9d74d076aa7dc226deb3b12a6R97). Is it bad?

EVAST9919:
Actually, my statement above is stupid, because now we have half of online logic in the overlay and half in the display. So the question is: is it better to handle all the online logic by the overlay and make displays full-offline (I mean, like in rankings, when all the requests are handled by the overlay and displays just accept the data) or the overlay should be only just a "container which only selects the needed display" and everything online will be handled by displays?

bdach:
(This might veer a little general, but it's beginning to be a general discussion, so I suppose it's not entirely out of place?)

I've had this idea for an overarching structure of overlays that could probably be best explained with the following diagram:

Untitled Diagram(8)

Easier said than done to achieve most likely, but I think this is a pattern that could end up working for us.

EVAST9919:
Refactored. Now all the online logic is handled in the overlay itself. (Grabbed some logic from the rankings with small improvements). Feels way better and have more sense now!

osu.Game.Tests/Visual/Online/TestSceneDashboardOverlay.cs
```diff
@@ -17,6 +17,8 @@ public class TestSceneDashboardOverlay : OsuTestScene
             typeof(DashboardOverlay),
             typeof(DashboardOverlayHeader),
             typeof(FriendDisplay),
+            typeof(DashboardDisplay),
+            typeof(DashboardDisplay<>),
```
peppy: the generic one here is not required (it's in the same file)

osu.Game/Overlays/Dashboard/Friends/DashboardDisplay.cs
```diff
@@ -0,0 +1,49 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Online.API;
+
+namespace osu.Game.Overlays.Dashboard.Friends
+{
+    public abstract class DashboardDisplay : CompositeDrawable
```
peppy: can spotlights use this class? i would hope so.

also xmldoc all public classes also xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classesalso xmldoc all public classes

what is this supposed to be? when should it be used? why would i use it?

EVAST9919: I forgot to remove it, my bad.

EVAST9919: Oh, wait, I did

peppy:
@EVAST9919 requires conflict resolution

EVAST9919:
Resolved

osu.Game.Tests/Visual/Online/TestSceneFriendDisplay.cs
```diff
@@ -24,31 +27,50 @@ public class TestSceneFriendDisplay : OsuTestScene
 
         protected override bool UseOnlineAPI => true;
 
+        [Resolved]
```
bdach, line 30: Should be marked as
```suggestion
        [Resolved(CanBeNull = true)]
```
(well, either that, or the null checks in tests should be removed).

osu.Game/Overlays/Dashboard/Friends/FriendDisplay.cs
```diff
@@ -193,32 +163,33 @@ private List<User> getUsersInCurrentGroup()
             }
         }
 
-        private void addContentToPlaceholder(Drawable content)
+        private void addTableToPlaceholder(FillFlowContainer<UserPanel> table)
```
bdach, line 166: This name had me quite confused for a second; if I'm understanding the intent correctly I'd rename to something like `replacePreviousContent`.

osu.Game/Overlays/Dashboard/Friends/FriendDisplay.cs
```diff
@@ -193,32 +163,33 @@ private List<User> getUsersInCurrentGroup()
...
             }
 
-            itemsPlaceholder.Add(currentContent = content);
-            currentContent.FadeIn(200, Easing.OutQuint);
+            itemsPlaceholder.Add(currentTable = table);
+            currentTable.FadeIn(200, Easing.OutQuint);
         }
 
-        private FillFlowContainer createTable(List<User> users)
+        private FillFlowContainer<UserPanel> createTable()
```
bdach, line 182: ...and this I'd call `createUserPanels()`.

osu.Game/Overlays/DashboardOverlay.cs
```diff
@@ -0,0 +1,179 @@
...
+            lastRequest = request;
+
+            if (request == null)
+            {
+                loadDisplay(null);
+                return;
+            }
+
+            request.Success += () => Schedule(() => loadDisplay(createDisplayFromResponse(request)));
+            request.Failure += _ => Schedule(() => loadDisplay(null));
```
bdach, line 127:
```suggestion
            request.Failure += _ => Schedule(() => loadDisplay(Empty()));
```
although all things considered this probably isn't the best UX (but I suppose this can be addressed in a separate PR further down the line).

osu.Game/Overlays/DashboardOverlay.cs
```diff
@@ -0,0 +1,179 @@
...
+        private void loadDisplay(Drawable display)
+        {
+            scrollFlow.ScrollToStart();
+
+            if (display == null)
+            {
+                content.Clear();
+                loading.Hide();
+                return;
+            }
```
bdach, line 141: As above - if `null` was disallowed here and `Empty` was passed instead this logic could go away.

osu.Game/Overlays/DashboardOverlay.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        private Drawable createDisplayFromResponse(APIRequest request)
+        {
+            switch (request)
+            {
+                case GetFriendsRequest friendsRequest:
+                    return new FriendDisplay(friendsRequest.Result);
+            }
+
+            return null;
```
bdach, line 158:
```suggestion
            return Empty();
```

osu.Game/Overlays/DashboardOverlay.cs
```diff
@@ -0,0 +1,179 @@
...
+
+        private APIRequest createScopedRequest()
+        {
+            switch (header.Current.Value)
+            {
+                case HomeOverlayTabs.Friends:
+                    return new GetFriendsRequest();
+            }
+
+            return null;
```
bdach, line 169: This one I'd say should throw an `ArgumentOutOfRangeException` as it's clear that if the enum is missing, something is just not implemented (don't think any of the future tabs in this overlay will be fully offline).

bdach (changes requested):

osu.Game/Overlays/DashboardOverlay.cs
EVAST9919: I need this logic though. I need to check if we have any content in `PopIn()`, so we don't create new requests if spamming social button in toolbar

bdach: Sounds like something `BindValueChanged(..., true)` could do.

EVAST9919: I added a simple bool flag to check if overlay has been fully closed before sending a new request.

peppy:
@EVAST9919 i'm not 100% sure what i'm looking at here. did you intend to structure things how @bdach asked for?

peppy:
To reiterate the general direction this should be moving towards:

- `DashboardOverlay` should not be doing any online logic
- `FriendDisplay` should inherit a new class `OverlayView` which implements basic `IOnlineComponent` (copy paste from `FullscreenOverlay` really)

This differs slightly from @bdach's diagram as it combines the view and controller, but I think this is enough for us for now (and quite a common practice).

And just to clarify, `FullscreenOverlay` is still fine to be an `IOnlineComponent`. It is still fine to handle requests when it's an overlay with *one* scope. But Dashboard has many scopes which need to be separated from each other else we are going to become huge class-wise.

EVAST9919:
Superseded by https://github.com/ppy/osu/pull/8771

PR #8055, opened 2020-02-29
Toolbar stay hides on new Screen

### LINKED ISSUE #8052: Toolbar should stay hidden when the user hides it
Pressing <kbd>CTRL</kbd> + <kbd>T</kbd> hides the toolbar (when shown), but when navigating most screens it shows back and that makes the shortcut useless in this sense.

This may be also good for stable users who think the lazer ui has many elements on screen or just use shortcuts. Most overlays on the toolbar can be accessed using shortcuts already.

Volnetiks:
Closes #8052.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -334,3 +334,6 @@ inspectcode
 
 # BenchmarkDotNet
 /BenchmarkDotNet.Artifacts
+
+# Discord Plugin Integration
+.idea/.idea.osu.Desktop/.idea/discord.xml
--- a/.idea/.idea.osu.Desktop/.idea/discord.xml
+++ b/.idea/.idea.osu.Desktop/.idea/discord.xml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="DiscordProjectSettings">
-    <option name="show" value="true" />
-  </component>
-  <component name="ProjectNotificationSettings">
-    <option name="askShowProject" value="false" />
-  </component>
-</project>
--- a/.idea/.idea.osu.Desktop/.idea/discord.xml
+++ b/.idea/.idea.osu.Desktop/.idea/discord.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="DiscordProjectSettings">
+    <option name="show" value="true" />
+  </component>
+  <component name="ProjectNotificationSettings">
+    <option name="askShowProject" value="false" />
+  </component>
+</project>
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -947,7 +947,7 @@ namespace osu.Game
 
                 if (newOsuScreen.HideOverlaysOnEnter)
                     CloseAllOverlays();
-                else
+                else if(Toolbar.State.Value == Visibility.Visible)
                     Toolbar.Show();
 
                 if (newOsuScreen.AllowBackButton)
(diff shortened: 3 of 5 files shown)


mcendu:
Please add “closes #8052”

.idea/.idea.osu.Desktop/.idea/discord.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcendu: Why add files?

Volnetiks: Uh, idk, I've never added it tbh lmao, I don't know where does it come from

Volnetiks: It comes from my discord RPC, should i remove the plugin?

peppy: you should not commit this. add it to your global .gitignore

Volnetiks: Okay, i will do this

osu.Game/Overlays/Toolbar/Toolbar.cs
```diff
@@ -79,7 +79,7 @@ private void load(OsuGame osuGame, Bindable<RulesetInfo> parentRuleset)
                         //{
                         //    Icon = FontAwesome.Solid.search
                         //},
-                        userButton = new ToolbarUserButton(),
+                        userButton = new ToolbarUserButton(this),
```
bdach: What is the meaning of these changes? You can't be doing things like this in a commit with the message of `Updated .gitignore`.

.idea/.idea.osu.Desktop/.idea/discord.xml
Volnetiks: Done, i've added it to the .gitignore, i will remove it from my fork

.gitignore
```diff
@@ -334,3 +334,6 @@ inspectcode
 
 # BenchmarkDotNet
 /BenchmarkDotNet.Artifacts
+
+# Discord Plugin Integration
+.idea/.idea.osu.Desktop/.idea/discord.xml
```
bdach: I think what @peppy intended is to edit your *global `.gitignore` on your actual local machine*, not this repository's. Git can have per-machine ignores that span all repositories.

bdach (changes requested):
Solution is incorrect, does not pass code style (as indicated by CI) and has no tests attached.

Also as a side note in general you do not want to be PRing from `master`. Make a branch instead.

osu.Game/Overlays/Toolbar/Toolbar.cs
Volnetiks: Oh, i started working on an other issue, and i've tried something, but i added it in my commit gitignore, i will change it rn

.gitignore
Volnetiks: Uh, i've never tried this, lemme check that

osu.Game/OsuGame.cs
```diff
@@ -947,7 +947,7 @@ protected virtual void ScreenChanged(IScreen current, IScreen newScreen)
 
                 if (newOsuScreen.HideOverlaysOnEnter)
                     CloseAllOverlays();
-                else
+                else if(Toolbar.State.Value == Visibility.Visible)
```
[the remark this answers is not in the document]
Volnetiks: Okay, i will try to find an other way for this issue

peppy:
As the logic in this doesn't work and you have committed many unrelated changes, I'm going to close this. Please make tests covering changed logic if you attempt this again.

PR #8059, opened 2020-02-29
Make KeyCounter not count key presses during breaks

### LINKED ISSUE #2707: Key presses during break times count up towards total
They do not on stable

### LINKED ISSUE #8033: Key overlay counts down to negative in replays when key pressed before maps start
**Describe the bug:**  
This issue is only specific to the maps that have a skippable start. When players do key presses before the point where the background fades to black *(seen with "Lighten playfield during break" setting on, but can be replicated with that setting turned off)*, the key counter will start to count down into negative values. After the background fades to black, all subsequent key presses make the key counter count up again. This issue only seems to happen in replays.

**Maps tested**
*Skippable maps (ones with issue):*
[Cry Thunder [Legend]](https://osu.ppy.sh/beatmapsets/316050#osu/704360)
[The pressure [wa's Extra]](https://osu.ppy.sh/beatmapsets/325307#osu/759638)
[Hades: The rise [Hardest]](https://osu.ppy.sh/beatmapsets/662526#osu/1402392)
[ONIGIRI FREEWAY [EXTREME RICE]](https://osu.ppy.sh/beatmapsets/996628#osu/2084629)

*Unskippable:*
[FREEDOM DiVE [FOUR DIMENSIONS]](https://osu.ppy.sh/beatmapsets/39804#osu/129891)
[Everything will freeze [Time Freeze]](https://osu.ppy.sh/beatmapsets/158023#osu/555797)
[ONIGIRI FREEWAY [PKHG x TATAN'S LOVE EXTRA]](https://osu.ppy.sh/beatmapsets/996628#osu/2161305)
[Night of Knights [The World]](https://osu.ppy.sh/beatmapsets/352570#osu/776951)

---
image
> This is one of the skippable maps during intro, key counter at `-3` and `-3`

---
**osu!lazer version:** 2020.226.0  
Tested on both Mac OS 10.15.3 and Windows 10 latest build

**Logs:**
I couldn't find a `~/.local/` folder at the moment, let me know if you do need a log for this

naoey:
Wasn't intended and not fully tested against the issue but seems to incidentally fix #8033.

Resolves #2707


bdach:
Haven't checked but might also resolve #2707?

osu.Game/Screens/Play/Player.cs
```diff
@@ -184,7 +184,7 @@ private void load(AudioManager audio, OsuConfigManager config)
             foreach (var mod in Mods.Value.OfType<IApplicableToHealthProcessor>())
                 mod.ApplyToHealthProcessor(HealthProcessor);
 
-            BreakOverlay.IsBreakTime.ValueChanged += _ => updatePauseOnFocusLostState();
+            BreakOverlay.IsBreakTime.BindValueChanged(_ => onBreakTimeChanged(), true);
```
bdach: Doesn't really make sense to discard the `ValueChangedEvent<bool>` argument if there's a read of `IsBreakTime.Value` in `onBreakTimeChanged`. Even more, I'd argue it's incorrect as it's possibly race-prone.

osu.Game.Tests/Visual/Gameplay/TestSceneAutoplay.cs
```diff
@@ -24,25 +22,31 @@ protected override Player CreatePlayer(Ruleset ruleset)
...
-            AddUntilStep("key counter reset", () => ((ScoreAccessiblePlayer)Player).HUDOverlay.KeyCounter.Children.All(kc => kc.CountPresses == 0));
+            ScoreAccessiblePlayer scoreAccessiblePlayer = null;
+
+            AddUntilStep("player loaded", () => (scoreAccessiblePlayer = (ScoreAccessiblePlayer)Player) != null);
+            AddUntilStep("score above zero", () => scoreAccessiblePlayer.ScoreProcessor.TotalScore.Value > 0);
+            AddUntilStep("key counter counted keys", () => scoreAccessiblePlayer.HUDOverlay.KeyCounter.Children.Any(kc => kc.CountPresses > 2));
+            AddStep("seek to break time", () => scoreAccessiblePlayer.GameplayClockContainer.Seek(scoreAccessiblePlayer.BreakOverlay.Breaks.First().StartTime));
+            AddUntilStep("wait for seek to complete", () =>
+                scoreAccessiblePlayer.HUDOverlay.Progress.ReferenceClock.CurrentTime >= scoreAccessiblePlayer.BreakOverlay.Breaks.First().StartTime);
+            AddAssert("test keys not counting", () => !scoreAccessiblePlayer.HUDOverlay.KeyCounter.IsCounting);
```
bdach: I'm not sure this does much. Sure, on the surface it looks correct as it tests the flag, but *the flag itself isn't tested* (as in, it's not tested that the key counter doesn't actually increment keys with `IsCounting = false` - what if the logic checking for `IsCounting` regresses?)

bdach (changes requested):
Would be nice to have a regression test for the replay case too. Shouldn't be too hard to generate - could roughly follow `TestSceneReplay` and manually generate a replay frame for a fake beatmap before it starts.

osu.Game.Tests/Visual/Gameplay/TestSceneAutoplay.cs
naoey: Not sure why I assumed this was already tested in KeyCounter

naoey:
As for the replay case, I didn't get the time to look into it too deeply but it seemed to have been caused by a clock having a rate of `-1` for a brief period at the start of a map where the replay had already recorded key presses. Still not entirely sure if that needs to be investigated separately or covered as fixed by this PR.

peppy (approved):

peppy (approved):

PR #8065, opened 2020-03-01
SD

kornkaobat:
Nuked


kornkaobat:
Linked with ppy/osu-performance/issues/108

TheWildTree:
Instead of using GitHub's editor to blindly modify the code, I'd recommend you install an actual IDE to compile and test the code before you commit. You can also use the `InspectCode.ps1` from the root directory of this repo to run the code quality checks locally instead of relying on CI.

kornkaobat:
@TheWildTree 
Capture124
Code is fully working using .NET Core 2.0.0. I tried it with the latest version of .NET but it won't compile with the latest version. Can you help me get some .osr and .osu file for me to test my pp calculations?

bdach:
This PR is intensely confused. To list off in no particular order:

* Editing code files through the github interface seems to have entirely mangled line endings (although I suspect it might have been due to copying from a local editor with different line endings), rendering the diff useless.
* The code seems to have been developed by trial and error and "compile tested" on CI. This is unacceptable.
* There is also added commented code [copy-pasted from Stack Overflow, that in parts doesn't even make sense or is valid C#](https://github.com/ppy/osu/pull/8065/files#diff-c60c222884828c0b2e604228b492403dR29-R42).
* As for this claim:
  > Code is fully working using .NET Core 2.0.0. I tried it with the latest version of .NET but it won't compile with the latest version.
  We're now on .NET Core *3.1*, so this is also not acceptable.

I'm going to go ahead and close this. If you attempt this again, please install an actual IDE, test your changes and understand the programming language you're attempting to use.

PR #8066, opened 2020-03-01
Add base test scene allowing different player initialisation per test method

### LINKED ISSUE #7867: Test scene for mod development
Right now, developing or modifying a mod is a laborious process which can't be easily iterated on via a visual test. It should be quite trivial to create a test scene for this. Should be a good task for someone looking to learn the visual testing setup.

- A new abstract test scene `TestSceneModSandbox` should be created in `osu.Game`, with one non-abstract implementation in each ruleset test project.
- It should use the existing test beatmap to show gameplay.
- It should use the existing `SelectedMods` bindable to read the current mods.
- The scene should show mod settings similar to the mod select overlay, and on adjusting a setting it should reload gameplay.
- Dynamic compilation should be tested to work.

Going forward, this would be abstracted once more to create a test scene per mod. A sample mod-specific test scene could be included as part of the initial implementation.

voidedWarranties:
All gamemodes have a generic testscene with a dummy mod

Specific mod example included for DifficultyAdjust in osu!std

- [x] Depends on https://github.com/ppy/osu/pull/8140.


smoogipoo:
I've improved this further because:

1. It needs to be automated.
2. It needs to support multiple mod settings / cases.

@peppy this is ready for review

smoogipoo (dismissed):

voidedWarranties:
I think the testing improvements are necessary, but the removal of setting adjustments outside of code makes the actual development process quite a bit more tedious
see linked issue-
> The scene should show mod settings similar to the mod select overlay, and on adjusting a setting it should reload gameplay.

should a new test case be added in the base class to allow for manual adjustment?

smoogipoo:
Manual adjustment isn't really necessary. When coding a new mod it is relatively simple to `CreateModTest(...)`. `TestSceneOsuModDifficultyAdjust` shows how this is intended to be used.

osu.Game/Tests/Visual/ModSandboxTestScene.cs
```diff
@@ -0,0 +1,107 @@
...
+        private ModTestCaseData currentTest;
+
+        protected void CreateModTest(ModTestCaseData testCaseData) => CreateTest(() =>
+        {
+            AddStep("set test data", () => currentTest = testCaseData);
+        });
+
+        public override void TearDownSteps()
+        {
+            AddUntilStep("test passed", () => currentTest?.PassCondition?.Invoke() ?? true);
```
bdach: Not sure we should be passing tests by default. Personally I'd do the opposite (fail by default and only pass if explicitly overridden by the test writer), or even throw if a pass condition wasn't specified.

bdach (commented):
Sort of disagree on one point, but looks good otherwise.

osu.Game/Tests/Visual/ModSandboxTestScene.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        protected sealed override IBeatmap CreateBeatmap(RulesetInfo ruleset) => currentTest?.Beatmap ?? base.CreateBeatmap(ruleset);
+
+        protected sealed override Player CreatePlayer(Ruleset ruleset)
+        {
+            SelectedMods.Value = SelectedMods.Value.Append(currentTest.Mod).ToArray();
+
+            var score = currentTest.Autoplay
+                ? ruleset.GetAutoplayMod().CreateReplayScore(Beatmap.Value.GetPlayableBeatmap(ruleset.RulesetInfo, SelectedMods.Value))
+                : new Score { Replay = new Replay() };
```
voidedWarranties: What issues does using the actual autoplay mod cause?

voidedWarranties:
Also, possibly belonging in a separate issue: the test player/beatmap act really strangely with a rate adjustment mod enabled
For example, with DT enabled the gameplay will suddenly speed up around the break

osu.Game/Tests/Visual/ModSandboxTestScene.cs
smoogipoo: Autoplay mod does not fails, for instance. It also doesn't allow restarting upon failure.

Both of those make SD and PF impossible to test.

osu.Game/Tests/Visual/ModSandboxTestScene.cs
smoogipoo: Changed it to `false` by default. I think this is the better of the two since an endless + failing step will allow the user to at least test their mods further.

smoogipoo:
DT issue tracked in the mentioning PR.

bdach (dismissed):

voidedWarranties:
For any mods/cases without a pass condition, will this purposely failing test also trip CI?

I've set the value of `score` when Autoplay is off to `null` because with an empty score, the player will block all input.
I've also passed the `AllowFail` value onto the player for proper fail handling in SD and PF (but with this the player will exit instead of restarting gameplay).

I still think a mod settings menu should be provided, as currently this mostly benefits mod testing/integrity rather than mod development

osu.Game/Tests/Visual/ModSandboxTestScene.cs
```diff
@@ -0,0 +1,117 @@
...
+
+namespace osu.Game.Tests.Visual
+{
+    public abstract class ModSandboxTestScene : PlayerTestScene
+    {
+        protected sealed override bool HasCustomSteps => true;
+
+        public override IReadOnlyList<Type> RequiredTypes => new[]
+        {
+            typeof(ModSandboxTestScene)
```
peppy: is this here as a requirement for inherited test scenes? sounds like something we may want to automate framework-side.

bdach:
> For any mods/cases without a pass condition, will this purposely failing test also trip CI?

It should, it's kind of the point. We want to very strongly discourage assert-less tests because they break silently and bring little value other than smoke testing (in fact they make CI run longer for little to no gain).

> I still think a mod settings menu should be provided, as currently this mostly benefits mod testing/integrity rather than mod development

There are options for changing values on-the-fly that do not require one (hot code reload, for one), and by the point of being a PR tests should have no manual input required. Even then there are ways to get around that (parameterised tests, range test steps, etc.)

smoogipoo (approved):

PR #8084, opened 2020-03-02
Fix perfect mod incorrectly failing in some scenarios

### LINKED ISSUE #8078: Perfect mod checks ignore judgements incorrectly
**Describe the bug:** 

When I got a GREAT I lose with perfect mode

https://youtu.be/Zs18Z9X8bz8

**osu!lazer version:** 

2020.229.1

**Logs:**
*please attach logs here, which are located at:*
- `%AppData%/osu/logs` *(on Windows),*
- `~/.local/share/osu/logs` *(on Linux & macOS).*
Android

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8066
- [x] new framework release

Fixes https://github.com/ppy/osu/issues/8078


osu.Game.Rulesets.Catch.Tests/TestSceneCatchModPerfect.cs
```diff
@@ -0,0 +1,54 @@
...
+        [TestCase(true)]
+        public void TestBananaShower(bool shouldMiss) => CreateHitObjectTest(new HitObjectTestCase(new BananaShower { StartTime = 1000, EndTime = 3000 }, false), shouldMiss);
+
+        [TestCase(false)]
+        [TestCase(true)]
+        public void TestFruit(bool shouldMiss) => CreateHitObjectTest(new HitObjectTestCase(new Fruit { StartTime = 1000 }), shouldMiss);
+
+        [TestCase(false)]
+        [TestCase(true)]
+        public void TestTestJuiceStream(bool shouldMiss)
```
bdach:
```suggestion
        public void TestJuiceStream(bool shouldMiss)
```

osu.Game/Tests/Visual/ModPerfectTestScene.cs
```diff
@@ -0,0 +1,68 @@
...
+                    return ScoreProcessor.HasCompleted && !HealthProcessor.HasFailed;
+
+                return HealthProcessor.HasFailed;
+            }
+        }
+
+        protected class HitObjectTestCase
+        {
+            public readonly HitObject HitObject;
+            public readonly bool FailOnMiss;
```
bdach: Do you have anything in mind for this for the future? As far as I can see this is always `true`, and I can't really see much usage opportunity in having this.

osu.Game.Rulesets.Osu.Tests/TestSceneOsuModPerfect.cs
```diff
@@ -0,0 +1,52 @@
...
+
+using NUnit.Framework;
+using osu.Game.Rulesets.Objects;
+using osu.Game.Rulesets.Objects.Types;
+using osu.Game.Rulesets.Osu.Mods;
+using osu.Game.Rulesets.Osu.Objects;
+using osu.Game.Tests.Visual;
+using osuTK;
+
+namespace osu.Game.Rulesets.Osu.Tests
```
bdach: For consistency with `TestSceneOsuModDifficultyAdjust`:
```suggestion
namespace osu.Game.Rulesets.Osu.Tests.Mods
```
Goes for all of the others if we're making this a convention.

bdach (commented):
A few nits from source. Haven't ran this with framework checkout yet but it compiles in my head.

osu.Game/Tests/Visual/ModPerfectTestScene.cs
smoogipoo: It's actually `false` in one case in `TestSceneCatchModPerfect`. Taiko should do the same, however it's a little hard to test drumroll ticks by themselves right now.

smoogipoo: Same with spinner bonus ticks when they're implemented.

bdach: Ah, my bad. Gotcha.

bdach:
CI is failing, because at the end of tests, upon gameplay completion there is an attempt to import the score:

https://github.com/ppy/osu/blob/fd16d24362ac1ed98f05e9b6942475bc9962c639/osu.Game/Screens/Play/Player.cs#L618-L625

and sqlite is eating a foreign key exception (haven't checked exactly why, but I'd bet it's due to violating constraints on either `BeatmapSetInfo` or `BeatmapInfo`).

Seeing that `ReplayPlayer` is a thing already I'd say it might be best to allow `Player` inheritors to specify their own logic on completion, defaulting to import.

peppy:
I agree with that direction. If you're already deep enough that you can easily commit the fix as proposed, you're welcome to go ahead!

bdach:
So I started to implement that but then found these and facepalmed *very* hard at my blindness:

https://github.com/ppy/osu/blob/9c5e3942adee2afe2d49394ef7081c7a622ce32c/osu.Game/Screens/Play/Player.cs#L107-L111

Sort of eh that these flags are so easy to miss but I guess if they're there already might as well set them and call it a day?

peppy:
It's weird that we do flags in three ways (ctor / settable props / overridden virtual props). Makes discovery hard.

bdach:
I passed in that flag and also fixed up another issue that would cause the tests to not do what they intended to (explanation in commit msg of c803de2). Seems good otherwise 👍

bdach (dismissed):

osu.Game/Tests/Visual/ModPerfectTestScene.cs
```diff
@@ -0,0 +1,67 @@
...
+            public bool CheckFailed(bool failed)
+            {
+                if (!failed)
+                    return ScoreProcessor.HasCompleted && !HealthProcessor.HasFailed;
+
+                return HealthProcessor.HasFailed;
+            }
+        }
+
+        protected class HitObjectTestCase
```
peppy: i still don't like this and think it's convoluted. should look at a better solution before merging (even if that means combining with the other "info" class in the parent test scene)

peppy (changes requested):

osu.Game/Tests/Visual/ModPerfectTestScene.cs
peppy: just going to go with this for now (just renamed to match `ModTestScene`). i think we can refactor/rethink as this continues to get used in more places.

PR #8090, opened 2020-03-02
Add SelectPrevious and SelectNext bindings / Change volume bindings to be truly global

### LINKED ISSUE #2869: missing hotkey for volume change while hovering over a scrollable container
the mouse wheel can be used to change the volume, unless you're hovering a scrollable container (e.g. song selection, scoreboards, social browser, direct, etc). stable features the hotkey alt+mouse wheel to change volume in these cases. This hotkey seems to be missing in lazer.

### LINKED ISSUE #8002: Alt + Up/Down combination for volume doesn't work in the pause/fail overlay
**Describe the bug:** As title: the <kbd>Alt</kbd> + <kbd>Up</kbd>/<kbd>Down</kbd> key combination for volume doesn't work in the pause/fail overlay. It only works with the mouse wheel. Not very useful since the overlay has no sound/music, but I guess it should work there (like in osu! stable).

**osu!lazer version:** 2020.225.0

peppy:
- Closes https://github.com/ppy/osu/issues/8002
- Closes https://github.com/ppy/osu/issues/2869

Both of these changes are highly interdependent so I couldn't split them out into separate PRs. This removes hacks for getting volume events to propagate upwards.

---

This resets any custom settings for volume increase / decrease hotkeys. The new defaults are Alt+MouseWheel and Alt+Up/Down. Any keys set will work *anywhere* as opposed to previously only working if nothing else handled.

This also adds new bindings for `Select Next` and `Select Previous` which can be used to customise selection of things like songs at song select and options at the pause menu.


smoogipoo (approved):
I think this is okay for now. Not 100% sold on the naming of "SelectNext" and "SelectPrevious" given their defaults, but we can change that in the future.

PR #8104, opened 2020-03-03
Add torus font

peppy:
Replaces existing usages of Aquatico with torus. Design fixes will come in follow-up PRs (for OWC 2020 series design).


osu.Game/Graphics/OsuFont.cs
```diff
@@ -65,16 +70,7 @@ public static string GetWeightString(Typeface typeface, FontWeight weight)
         /// <param name="family">The family string.</param>
         /// <param name="weight">The <see cref="FontWeight"/>.</param>
         /// <returns>The string representation of <paramref name="weight"/> in the specified <paramref name="family"/>.</returns>
-        public static string GetWeightString(string family, FontWeight weight)
-        {
-            string weightString = weight.ToString();
-
-            // Only exo has an explicit "regular" weight, other fonts do not
-            if (family != GetFamilyString(Typeface.Exo) && weight == FontWeight.Regular)
```
smoogipoo: Have you checked that this isn’t required still? I’m fairly certain this breaks Venera.

smoogipoo (changes requested):

osu.Game/Graphics/OsuFont.cs
peppy: aha, didn't see that case, looks bad. we probably want to rename this and remove "medium" altogether (or standardise to regular).

peppy: Turns out all the naming for venera was wrong (and one instance existed without a weight suffix). Should be better aligned now.

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
     public enum FontWeight
     {
-        Light,
-        Regular,
-        Medium,
-        SemiBold,
-        Bold,
-        Black
+        /// <summary>
+        /// equivalent to weight 300
```
smoogipoo:
```suggestion
        /// Equivalent to weight 300.
```

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
+        /// </summary>
+        Light = 300,
+
+        /// <summary>
+        /// equivalent to weight 400
```
smoogipoo:
```suggestion
        /// Equivalent to weight 400.
```

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
+        /// </summary>
+        Regular = 400,
+
+        /// <summary>
+        /// equivalent to weight 500
```
smoogipoo:
```suggestion
        /// Equivalent to weight 500.
```

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
+        /// </summary>
+        Medium = 500,
+
+        /// <summary>
+        /// equivalent to weight 600
```
smoogipoo:
```suggestion
        /// Equivalent to weight 600.
```

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
+        /// </summary>
+        SemiBold = 600,
+
+        /// <summary>
+        /// equivalent to weight 700
```
smoogipoo:
```suggestion
        /// Equivalent to weight 700.
```

osu.Game/Graphics/OsuFont.cs
```diff
@@ -102,15 +98,39 @@ public enum Typeface
...
+        /// </summary>
+        Bold = 700,
+
+        /// <summary>
+        /// equivalent to weight 900
```
smoogipoo:
```suggestion
        /// Equivalent to weight 900.
```

smoogipoo (changes requested):

PR #8106, opened 2020-03-03
Update tournament client to 2019 design language

peppy:
- [x] Depends on #8105
- [x] Depends on #8104
- [x] Depends on #8122

Soon to be replaced by 2020 designs. Want to get this in master first, though.


smoogipoo (approved):
I've done a very rough logic-only pass on this without checking correctness against the design.

PR #8107, opened 2020-03-03
Default replay playback settings menu to collapsed

integerrr:
First PR so please don't drill into me too hard
UX-wise I think it makes more sense to have this menu collapsed in replays/automod


bdach:
A one-line change should not take six commits (four of which are merge commits over what appears to be at least three branches). I'm not sure what you're doing with your git workflow but it's not proper.

As for the actual change I find it disputable. It could be argued that it breaks parity with stable (where replay controls are visible by default). Personally I'd at most persist the hiding during the length a play session (so if a user hides the controls, keep them hidden until next restart) using `SessionStatics`, or even between sessions, but nothing more.

peppy:
If anything it should remember the choice, as mentioned above. Please reopen as a new PR following provided advice if you attempt this.

PR #8121, opened 2020-03-04
Comment out log locations in bug report template

mcendu:
This info is only useful for reporters of issues, and people viewing the issue can find them distracting.


LittleEndu:
Same needs to be done to error template

Issue #3051, opened 2018-07-16
choosing beatmap with lef/right arrows is broken


ticotaco72:
when you rewind beatmaps with left/right arrow in beatmap solection menu, it sometimes freezes and goes to up of list instead one up/ one down

logs:
[database.log](https://github.com/ppy/osu/files/2199136/database.log)
[network.log](https://github.com/ppy/osu/files/2199137/network.log)
[performance.log](https://github.com/ppy/osu/files/2199138/performance.log)
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/2199139/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/2199140/performance-gamethread.draw.log)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/2199141/performance-gamethread.input.log)
[performance-gamethread.update.log](https://github.com/ppy/osu/files/2199142/performance-gamethread.update.log)
[runtime.log](https://github.com/ppy/osu/files/2199143/runtime.log)

video:
https://drive.google.com/open?id=18Lby7tLooPEGnz-KW5R0oVOXg_OXs7Zw

Morilli:
Could reproduce. I imagine the lag caused by the scrolling to be connected to this issue.

ticotaco72:
i have noticed that when you rewind with left arrow (to up) then after freeze you go to down of list
and similarly with right arrow

PR #8129, opened 2020-03-04
Fix DummyWorkingBeatmap's track completion attempting to change game-wide beatmap

### LINKED ISSUE #8111: Using a notification to view a beatmap while search filter wouldn't display it breaks searching
**Describe the bug:** 
After downloading a beatmap from direct, if you click on the notification that says "View this beatmap" some difficult of said beatmap is displayed no matter what, even if your current search query would filter the beatmap out. 
The result is some funcky unexpected behaviour. Sometimes the song shown to you isn't even the one you clicked the notification for. Searching is broken for some time but recovers after you clear the query. Navigating away from the selected beatmap seems to also restore search functionality.

osu_2020-03-03_21-23-40

**osu!lazer version:** 
2020.229.1

**Logs:**
[notification-bug-logs.zip](https://github.com/ppy/osu/files/4283208/notification-bug-logs.zip)

peppy:
Quite hard to test this one reliably (used to trigger a feedback loop, but not always, timing dependent), but the case where this can happen should be quite understandable.

- Closes https://github.com/ppy/osu/issues/8111
- Fixes long freezes occasionally occurring when selecting beatmaps from notifications or osu!direct.
- Has a chance of fixing https://github.com/ppy/osu/issues/3051
- [x] Depends on https://github.com/ppy/osu-framework/pull/3333

### DIFF
--- a/osu.Game.Tests/Visual/Navigation/OsuGameTestScene.cs
+++ b/osu.Game.Tests/Visual/Navigation/OsuGameTestScene.cs
@@ -97,6 +97,8 @@ namespace osu.Game.Tests.Visual.Navigation
 
             public new SettingsPanel Settings => base.Settings;
 
+            public new MusicController MusicController => base.MusicController;
+
             public new OsuConfigManager LocalConfig => base.LocalConfig;
 
             public new Bindable<WorkingBeatmap> Beatmap => base.Beatmap;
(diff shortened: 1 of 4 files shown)


peppy:
Note that the main fix for this is now the framework-side change, but this local change is still required to guard against callbacks being fired by the previous (old) track.

smoogipoo (approved):

PR #8130, opened 2020-03-04
Fix icon being outside overlay header

recapitalverb:
Follows #8103.
Screenshot_20200304_151712


smoogipoo:
The correct fix is to revert the commit and instead set only the content flow to 10:

https://github.com/ppy/osu/blob/master/osu.Game/Overlays/SearchableList/SearchableListOverlay.cs#L83

Because the change also reduced the paddings of everything else.

PR #8132, opened 2020-03-04
Update user panel design

EVAST9919:
Screenshot_1
Screenshot_2


smoogipoo:
This should immediately replace the existing user card.

EVAST9919:
I'd like to, but I have no design of these states (which are lazer-only for now)
Screenshot_2

EVAST9919:
I assume changing the icon colour and the state text will be enough?

EVAST9919:
Replaced

osu.Game/Users/UserPanel.cs
```diff
@@ -217,33 +98,105 @@ protected override void LoadComplete()
...
+                }
+                else
+                {
+                    lastVisitMessage.FadeTo(status is UserStatusOffline && User.LastVisit.HasValue ? 1 : 0);
+                    statusMessage.Text = status.Message;
+                    statusIcon.FadeColour(status.GetAppropriateColour(colours), 500, Easing.OutQuint);
+                }
+
+                return;
             }
```
bdach: This needs to be split into a method, refactored for readability (reduce nesting), and also thoroughly tested in the test scene.

bdach (changes requested):
I would request that changes from #8138 be integrated before this is merged.

osu.Game/Users/UserPanel.cs
```diff
@@ -50,164 +34,61 @@ public class UserPanel : OsuClickableContainer, IHasContextMenu
...
 
-            FillFlowContainer infoContainer;
+        [Resolved]
+        private OsuColour colours { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Masking = true;
+            BorderColour = colours.GreyVioletLighter;
```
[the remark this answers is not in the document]
EVAST9919: I wasn't sure about the colour if the panel used outside the overlay (like in login settings or tournament client)

osu.Game/Users/UserGridPanel.cs
```diff
@@ -0,0 +1,113 @@
...
+                                        Spacing = new Vector2(6),
+                                        Children = new Drawable[]
+                                        {
+                                            CreateFlag(),
+                                        }
+                                    },
+                                    CreateUsername(),
+                                }
+                            }
+                        },
```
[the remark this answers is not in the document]
EVAST9919: Will revisit this one

osu.Game/Users/UserPanel.cs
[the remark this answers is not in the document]
TheWildTree: maybe try using the `OverlayColourProvider` while keeping the `GreyVioletLighter` as a fallback colour if OCP is not accesible?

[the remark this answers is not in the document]
EVAST9919: Sounds good

EVAST9919:
Basically it's now 1:1(ish)
nice
Test scene will be expanded after https://github.com/ppy/osu/pull/8138 merge as suggested

osu.Game/Users/UserPanel.cs
```diff
@@ -50,164 +34,64 @@ public class UserPanel : OsuClickableContainer, IHasContextMenu
...
+        private OverlayColourProvider colourProvider { get; set; }
+
+        [Resolved]
+        private OsuColour colours { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            Masking = true;
+            BorderColour = colourProvider?.Light1 ?? colours.GreyVioletLighter;
```
TheWildTree: `UserListPanel` has a weird border rendering issue:

image

Not sure what's causing that, but the border has a different depending on what element it overlays (notice how it's different over the `UserCoverBackground` than over the background colour box) - probably a framework issue.

osu.Game/Users/UserPanel.cs
```diff
@@ -217,33 +101,110 @@ protected override void LoadComplete()
...
+
+        protected UpdateableFlag CreateFlag() => new UpdateableFlag(User.Country)
+        {
+            Size = new Vector2(39, 26)
+        };
+
+        protected OsuSpriteText CreateUsername() => new OsuSpriteText
         {
-            const float transition_duration = 500;
+            Font = OsuFont.GetFont(size: 20, weight: FontWeight.Bold, italics: true),
```
TheWildTree:
```suggestion
            Font = OsuFont.GetFont(size: 16, weight: FontWeight.Bold),
```

seems like the new design doesn't use italics anymore

osu.Game/Users/UserPanel.cs
```diff
@@ -217,33 +101,110 @@ protected override void LoadComplete()
...
-                this.ResizeHeightTo(height - status_height, transition_duration, Easing.OutQuint);
-            }
-            else
+                AutoSizeAxes = Axes.Both,
+                Direction = FillDirection.Vertical
+            };
+
+            var alignment = rightAlignedChildren ? Anchor.CentreRight : Anchor.CentreLeft;
+
+            statusContainer.Add(lastVisitMessage = new TextFlowContainer(t => t.Font = OsuFont.GetFont(size: 15, weight: FontWeight.SemiBold)).With(text =>
```
TheWildTree:
```suggestion
            statusContainer.Add(lastVisitMessage = new TextFlowContainer(t => t.Font = OsuFont.GetFont(size: 12, weight: FontWeight.SemiBold)).With(text =>
```

osu.Game/Users/UserPanel.cs
```diff
@@ -217,33 +101,110 @@ protected override void LoadComplete()
...
+                        Shadow = false
+                    });
+                }
+            }));
+
+            statusContainer.Add(statusMessage = new OsuSpriteText
+            {
+                Anchor = alignment,
+                Origin = alignment,
+                Font = OsuFont.GetFont(size: 17, weight: FontWeight.SemiBold)
```
TheWildTree:
```suggestion
                Font = OsuFont.GetFont(size: 14, weight: FontWeight.SemiBold)
```

thewildtree (commented):
Font size values should match osu-web 1:1 - I know this will make the panel look *less* like its web counterpart for now, but we need to have the right values there so when the framework-side font sizing fix comes in the future, we won't have to re-check and adjust them again. 

You could consider adding temporary paddings to the text elements to compensate for the line-height differences like I did in my adjustments overlay and comment them accordingly to make them easily removable when the font fix comes.

osu.Game/Users/UserPanel.cs
smoogipoo: This is because borders are actually drawn as part of the contained drawables. Meaning you should have a `RelativeSizeAxes = Axes.Both, Alpha = 0, AlwaysPresent = true` box at the highest level, in the masking container, to ensure a constant border (or even a border to be drawn in the first place).

Yeah this is an issue that we'll eventually need to resolve framework-side, but this is a fine solution for now.

osu.Game.Tests/Visual/Online/TestSceneUserPanel.cs
```diff
@@ -51,27 +62,40 @@ public class TestSceneUserPanel : OsuTestScene
...
         public void TestUserStatus()
         {
-            AddStep("online", () => peppy.Status.Value = new UserStatusOnline());
-            AddStep("do not disturb", () => peppy.Status.Value = new UserStatusDoNotDisturb());
-            AddStep("offline", () => peppy.Status.Value = new UserStatusOffline());
-            AddStep("null status", () => peppy.Status.Value = null);
+            AddStep("online", () => peppy.Status.Value = evast.Status.Value = new UserStatusOnline());
+            AddStep("do not disturb", () => peppy.Status.Value = evast.Status.Value = new UserStatusDoNotDisturb());
+            AddStep("offline", () => peppy.Status.Value = evast.Status.Value = new UserStatusOffline());
+            AddStep("null status", () => peppy.Status.Value = evast.Status.Value = null);
```
bdach: You can extract one local bindable and do `BindTo()` twice (the same way as it's done with `Activity`) to avoid compound assignment.

bdach (commented):
No major objections.

bdach (dismissed):
LGTM (for now)

smoogipoo (approved):
lgtm 👍

PR #8136, opened 2020-03-04
Centralise logic for login/retry placeholders

Game4all:
Should allow the placeholder to get correct hover and click animations as mentionned here
https://github.com/ppy/osu/pull/7438#issuecomment-571945417


bdach (dismissed):
The structure of `LoginPlaceholder` is all weird now. You have:

```
LoginPlaceholder : OsuTextFlowContainer
    LoginButton : OsuAnimatedButton
        OsuTextFlowContainer
```

It shows just how arbitrary the choice of the `Placeholder` type in `OnlineViewContainer` was. It should probably either be a non-specific `Drawable`, or be a button on its own. Not sure.

Game4all:
It seems that another placeholder implements a custom logic for making part of the placeholder clickable (`RetrievalFailurePlaceholder`), so it may be worth making a shared class for clickable placeholders to unify the behaviour.

As to fix the weird placeholder inheritance, there's a `MessagePlaceholder` class already there so I guess the `OsuTextFlowContainer` should be moved to the MessagePlaceholder class and the abstract `Placeholder` class should inherit a Container instead

bdach:
No, no baroque inheritance hierarchies, please. These are literally buttons with an icon and text. There's no reason to have three levels of inheritance for that.

Game4all:
I created a `ClickablePlaceholder` class which implements a clickable placeholder and replaced the other customs implementations of a clickable placeholder by instances of it.
LoginPlaceholder now derives from that class and is kept as a subclass as it is being used in two places (and soon three).
Also renaming the PR title as it doesn't only touch the LoginPlaceholder anymore.

Game4all:
I created a `ClickablePlaceholder` class which implements a clickable placeholder and replaced the other customs implementations of a clickable placeholder by instances of it.
LoginPlaceholder now derives from that class and is kept as a subclass as it is being used in two places (and soon three).
Also renaming the PR title as it doesn't touch only the LoginPlaceholder anymore.

peppy (approved):

PR #8141, opened 2020-03-05
Centralise creation of UpdateManagers

peppy:


osu.Game/Updater/UpdateManager.cs
```diff
@@ -34,12 +34,12 @@ protected override void LoadComplete()
...
             if (game.IsDeployedBuild && version != lastVersion)
             {
-                config.Set(OsuSetting.Version, version);
-
                 // only show a notification if we've previously saved a version to the config file (ie. not the first run).
                 if (!string.IsNullOrEmpty(lastVersion))
                     Notifications.Post(new UpdateCompleteNotification(version));
             }
+
+            config.Set(OsuSetting.Version, version);
```
smoogipoo: Aren't debug builds going to destroy the config, since this is now always set?

peppy: yes, intentionally.

smoogipoo (dismissed):

smoogipoo (approved):

PR #8151, opened 2020-03-06
Simplify tournament video construction

peppy:
Move video stream retrieval internal to `TourneyVideo`. Allows for more optional video specifications at a per-screen level even if they potentially aren't being implemented (will fallback to `main`, handled by `TournamentSceneManager`).


smoogipoo (approved):
Seems sane enough.

PR #8163, opened 2020-03-06
Fix osu! hitbox accepting input outside of circle

### LINKED ISSUE #8158: Hitcircles can be clicked outside of their borders
It seems that the hitbox extends past the edge of the physical hitcircles, meaning that some clicks that could break your combo don't actually do so and register as a hit.  See the video from the following reddit link for an example. Maybe this is intended, and the cursor hitbox isn't confined to its center, but this could be worthwhile to look into if it isn't. 

https://www.reddit.com/r/osugame/comments/fegn3b/osulazer_hitcircles_are_actually_hitsquares/

Lazer 2020.306.0 (Latest Release)

bdach:
Resolves #8158.

# Summary

Hitboxes of circle pieces in osu! seem to have unintentionally regressed with commit 8592335. The reason for the regression was that hit detection was moved from `CirclePiece` itself to a newly-introduced private `HitArea` class (now named `HitReceptor`).

As `HitArea` inherited from `Drawable`, it would return `IsHovered == true` over its entire bounding box. This meant that the hit area could wrongly pick up actions that are not within circle radius and make them into hits.

To resolve, make `HitReceptor` a `CompositeDrawable` and set its corner radius to match the circle piece. This fixes the invalid hitbox, as `IsHovered` takes radius into account.

# Remarks

Includes a test scene, so this should hopefully never regress again. I do realise it might be a little awkwardly written - I tried to not base on either `PlayerTestScene` or `SkinnableTestScene` as the first one seems too slow for what the tests aim to test and the skinnable one would probably look broken and it's not what that test class is for anyway.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneHitCircleArea.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneHitCircleArea.cs
@@ -1,0 +1,108 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Testing;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Rulesets.Osu.Objects;
+using osu.Game.Rulesets.Osu.Objects.Drawables;
+using osu.Game.Rulesets.Scoring;
+using osu.Game.Skinning;
+using osuTK;
+
+namespace osu.Game.Rulesets.Osu.Tests
+{
+    public class TestSceneHitCircleArea : ManualInputManagerTestScene
+    {
+        private HitCircle hitCircle;
+        private DrawableHitCircle drawableHitCircle;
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu.Tests/TestSceneHitCircleArea.cs
```diff
@@ -0,0 +1,108 @@
...
+            });
+            scheduleHit();
+
+            AddAssert("hit registered", () => hitAreaReceptor.HitAction == OsuAction.LeftButton);
+        }
+
+        [Test]
+        public void TestCircleMissBoundingBoxCorner()
+        {
+            AddStep("move mouse to top left corner of bounding box", () => InputManager.MoveMouseTo(hitAreaReceptor.ScreenSpaceDrawQuad.TopLeft));
```
LittleEndu: Wouldn't this test fail if the bounding box for some reason would ever be smaller than hit area? 
Shouldn't this test be specifically moving the mouse to a position just outside the hit area instead of to the position of the bounding box? Wouldn't it catch future regressions better that way?

bdach: FWIW the inside of the circle was covered in a test, and so is the outside now with b608764.

peppy (approved):
Looks like a good solution for now, nice catch.

Issue #7057, opened 2019-12-04
Replace local EFcore database backing with Realm


peppy:
Over the past five days I've dedicated a large block of time into looking at alternatives to EF Core. This is a consideration that we have been talking about internally for a while now, for the following reasons:

- We are already disabling EF change tracking as the overhead is too large and unpredictable with the way we consume database objects.
- The way we update entities is very hap-hazard, sometimes requiring a call to `store.Refresh()` to retrieve a "live" version which can be modified and updated correctly.
- Startup performance of EF core is unpredictable, especially with a large number of migrations present.
- Migrations are a bit of a pain to orchestrate, and limited by the SQLite backend

The first step was to decide whether we should be sticking with SQLite

Pro:
- We can use all the existing structures with minimal changes
- SQLite is tried and tested as far as reliability, threading and performance
- We already have the native libraries setup and working correctly
- We could return to EF core if required.

Con:
- SQLite migrations are *very* limited
- Data types provided by SQLite are very limited
- ORL libraries generally include SQLite as an afterthought and as such have long-standing bugs (ran into this with EF core, requiring manual workarounds)

## Attempt 1: [Dapper + LiteDB](https://github.com/ppy/osu/compare/master...peppy:dapper?expand=1)

I an initial effort to move forward with simplicity in mind, I attempted switching to dapper + LiteDB. LiteDB is a very lightweight (and not heavily maintained or used) project which allowed to quickly test things out without changing our models too much.

The branch above is in a mostly working state, but brought up a few crippling issues:

### All queries need to be converted to raw SQL.

While this isn't a killer, it feels like a step backwards in terms of code quality, readability and maintainability. Rider will give inspections on the SQL to ensure correctness to some extent, but this relies on the table name being present in the queries (which it isn't in my branch, to use our existing store layout).

### SQLite bugs prevail

I ran into [this](https://github.com/StackExchange/Dapper/issues/642) issue. It has a multi-year history and no sign of being fixed without local workarounds. This isn't necessarily Dapper's fault either – it's a shortcoming in the communications between the sqlite native/wrappers and the Dapper mappings which can't easily be resolved without a workaround that would affect other database engines too.

At this point I stopped looking towards dapper as it didn't feel like the correct direction, and would leave a lot of the issues that we have with EF core.

## Attempt 2: [Realm](https://github.com/ppy/osu/compare/master...peppy:realm-with-wrapper?expand=1)
[...]

PR #8167, opened 2020-03-07
Start background video playback based on provided offset

### LINKED ISSUE #6101: Video playback doesn't consider offset value
**Describe the bug:** 
Videos on most maps are out of sync with the music. Suspected reason: video offset in .osu file is not being applied correctly.

The discrepancy is especially big on [this map](https://osu.ppy.sh/beatmapsets/42532#osu/133729), which I believe is a map with a very large video offset.

https://osu.ppy.sh/beatmapsets/42532#osu/133729
https://osu.ppy.sh/beatmapsets/45404#osu/141786
https://osu.ppy.sh/beatmapsets/945049#osu/2026320

**osu!lazer version:** 
Latest 2019.913.0
iPad Pro
**Logs:**

voidedWarranties:
Fixes #6101

Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3356
- [x] #8426

Starts video playback at the correct time.

Testing:
Negative offset but `AudioLeadIn` is 0: https://osu.ppy.sh/beatmapsets/680251/#osu/1438380
Skippable intro before video begins: https://osu.ppy.sh/beatmapsets/42532#osu/133729


bdach:
I don't think persisting the offset to database is the correct approach to take, and I think that's what [one of the comments in the issue thread](https://github.com/ppy/osu/issues/6101#issuecomment-531444215) refers to:

> In fact the current method of video playback initialisation is very wrong and temporary at best (it should be read as part of the storyboard, where the information exists).

I'm pretty sure that would imply reading the offset *and filename* of the video file in `LegacyStoryboardDecoder` and refactoring code to use that data instead of DB metadata.

voidedWarranties:
Option to disable videos (without disabling storyboards) is still not reimplemented - how should individual storyboard layers be disabled?

peppy:
I completely agree with adding the video to the storyboard as a layer. For now the disable can be shared – we can look at adding more flexibility here going forward (something like a dropdown which allows `None`/`Background`/`Video`/`All`)

bdach (dismissed):
After this change, the background and the video both show on non-standard resolutions (while it used to be just the video):

osu_2020-03-08_15-04-58

FWIW, the letterboxing was already there, so that didn't regress. Could be good to have it stretch to fill the entire screen matching stable at some point though (maybe this is the right moment for it, maybe not).

osu.Game/Migrations/OsuDbContextModelSnapshot.cs
```diff
@@ -141,6 +141,8 @@ protected override void BuildModel(ModelBuilder modelBuilder)
 
                     b.Property<string>("VideoFile");
 
+                    b.Property<int>("VideoOffset");
```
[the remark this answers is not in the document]
voidedWarranties: I've already removed both `VideoOffset` and `VideoFile` from the actual `BeatmapMetadata`, should I put `VideoFile` back?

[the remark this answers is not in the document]
bdach: That is *not* what I'm saying. The column spec should not be in the db context snapshot if the migration is gone and so are the fields (which I intended to highlight by leaving a comment at this exact line).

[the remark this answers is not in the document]
voidedWarranties: Sorry, hadn't realized that I left `VideoFile` in the metadata.
When creating a migration for removing `VideoFile`, it failed because dropping columns is not supported. How should it be dealt with?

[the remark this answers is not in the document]
bdach: Huh, you're right, [sqlite doesn't support it](https://docs.microsoft.com/en-us/ef/core/providers/sqlite/limitations#migrations-limitations). Wow.

In that case I'm not too sure? We could rebuild the table altogether but that's going to require hand-writing SQL, most likely. Maybe better to just leave it unused [as we were planning to ditch EF Core and sqlite in the long run](https://github.com/ppy/osu/issues/7057)?

bdach:
The x & y position stuff should be a separate pull. This one is already big enough as it stands.

bdach:
Aside from leaving behind the now-unused `VideoFile` column in users' sqlite databases this looks to be in a pretty good state right now 👍

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -116,6 +117,14 @@ private void load(OsuConfigManager config)
             if (beatmap.BeatmapInfo.AudioLeadIn > 0)
                 startTime = Math.Min(startTime, firstHitObjectTime - beatmap.BeatmapInfo.AudioLeadIn);
 
+            // some beatmaps have no AudioLeadIn but the video starts before the first object
+            var videoLayer = beatmap.Storyboard.GetLayer(LegacyStoryLayer.Video);
+
+            var videoOffset = videoLayer.Elements.SingleOrDefault()?.StartTime;
+
+            if (videoOffset != null)
+                startTime = Math.Min(startTime, videoOffset.GetValueOrDefault());
```
smoogipoo:
```suggestion
                startTime = Math.Min(startTime, videoOffset.Value);
```

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -116,6 +117,14 @@ private void load(OsuConfigManager config)
             if (beatmap.BeatmapInfo.AudioLeadIn > 0)
                 startTime = Math.Min(startTime, firstHitObjectTime - beatmap.BeatmapInfo.AudioLeadIn);
 
+            // some beatmaps have no AudioLeadIn but the video starts before the first object
+            var videoLayer = beatmap.Storyboard.GetLayer(LegacyStoryLayer.Video);
```
smoogipoo: I'm a little bit confused as to why this is here. It should be included as part of `beatmap.Storyboard.FirstEventTime` (above)?

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -116,6 +117,14 @@ private void load(OsuConfigManager config)
             if (beatmap.BeatmapInfo.AudioLeadIn > 0)
                 startTime = Math.Min(startTime, firstHitObjectTime - beatmap.BeatmapInfo.AudioLeadIn);
 
+            // some beatmaps have no AudioLeadIn but the video starts before the first object
```
smoogipoo: This should be "have AudioLeadIn", right?

osu.Game/Storyboards/Drawables/DrawableStoryboardVideo.cs
```diff
@@ -0,0 +1,84 @@
...
+                Alpha = 0
+            });
+
+            videoClock = new ManualClock();
+            videoSprite.Clock = new FramedClock(videoClock);
+        }
+
+        protected override void Update()
+        {
+            if (clock != null && clock.CurrentTime > Video.StartTime)
```
smoogipoo: Not a fan of this. You want to give VideoSprite an `OffsetClock`, with the appropriate offset corresponding to its start time.
Then you want to fade in the video via:
```
using (videoSprite.BeginAbsoluteSequence(0))
    videoSprite.FadeIn(500)
```
Inside `LoadComplete()`.

smoogipoo (changes requested):

osu.Game/Screens/Play/GameplayClockContainer.cs
voidedWarranties: Sorry, I left these in because of the previous (non-storyboard) implementation, not realizing the logic was already there

smoogipoo:
Made some minor refactorings. Also starting at absolute time 0 is intended, since it's absolute time 0 for the post-offset-clock time.

Looks good otherwise.

smoogipoo (dismissed):

voidedWarranties:
Using the `OffsetClock` caused a regression in maps like https://osu.ppy.sh/beatmapsets/42532#osu/133729 where you can skip before the video begins - the video won't seek properly after the skip

From what I could previously tell, this was caused by the `VideoSprite` not updating when `Alpha = 0`
Previously, this was solved by setting `AlwaysPresent = true` but this is no longer viable as the video will simply run in the background before it shows up
Maybe it would be nice to have something framework-side to force-sync the video?

smoogipoo:
Wow, this is huge bug in VideoSprite itself. The logic here is correct, but VideoSprite is using clocks wrong.

bdach (dismissed):
👍

osu.Game/Storyboards/Storyboard.cs
```diff
@@ -46,14 +50,17 @@ public bool ReplacesBackground
...
+
+                return GetLayer(LegacyStoryLayer.Background).Elements.Any(e => e.Path.ToLowerInvariant() == backgroundPath);
             }
         }
 
         public DrawableStoryboard CreateDrawable(WorkingBeatmap working = null)
         {
             var drawable = new DrawableStoryboard(this);
-            drawable.Width = drawable.Height * (BeatmapInfo.WidescreenStoryboard ? 16 / 9f : 4 / 3f);
+            drawable.Width = drawable.Height * (BeatmapInfo.WidescreenStoryboard || GetLayer(LegacyStoryLayer.Video).Elements.Any() ? 16 / 9f : 4 / 3f);
```
peppy: where does this come from?

osu.Game/Beatmaps/Legacy/LegacyStoryLayer.cs
```diff
@@ -3,11 +3,12 @@
 
 namespace osu.Game.Beatmaps.Legacy
 {
-    internal enum LegacyStoryLayer
+    public enum LegacyStoryLayer
```
peppy: legacy classes should not be public. use strings for now in places that are non-legacy (ie. `Storyboard.cs`)

osu.Game/Overlays/Settings/Sections/Graphics/DetailSettings.cs
```diff
@@ -22,11 +22,6 @@ private void load(OsuConfigManager config)
                     Bindable = config.GetBindable<bool>(OsuSetting.ShowStoryboard)
```
peppy: Change the label here to "Storyboards / Videos" for now

osu.Game/Screens/Play/PlayerSettings/VisualSettings.cs
```diff
@@ -44,7 +43,6 @@ public VisualSettings()
                     Text = "Toggles:"
                 },
                 showStoryboardToggle = new PlayerCheckbox { LabelText = "Storyboards" },
```
peppy:
```suggestion
                showStoryboardToggle = new PlayerCheckbox { LabelText = "Storyboard / Video" },
```

peppy (changes requested):

osu.Game/Storyboards/Storyboard.cs
voidedWarranties: This was added to prevent the storyboard from masking the video to 4:3

peppy: It looks like masking isn't actually implemented right now (it's just limiting the size).

The correct solution here is to move the masking (sizing?) logic to `DrawableStoryboardLayer` and actually apply it (by setting `Masking = true`). Probably best done as a separate PR. Then the video layer could choose in its constructor to not be masked.

voidedWarranties: `Masking = true` is set for the entire storyboard in [DimmableStoryboard](            https://github.com/ppy/osu/blob/8fb41e4c670f1dcf1b8b27d7168d89b67e8d6f43/osu.Game/Screens/Play/DimmableStoryboard.cs#L47)

With this removed, the video will always display at full brightness (ignoring `UserDimContainer`)
It appears `DrawableStoryboardVideo` will render with the dim as long as any of its parents have masking enabled (including `DimmableStoryboard` itself)

Should this part be left as-is for now?

peppy: Aha. That masking can be removed after the above refactor is made.

Masking should not affect dim, though.

voidedWarranties: It looks like `VideoSprite` must have masking in any of its parents in order for dims to apply. Are there any special properties about either masking or the `VideoSprite` which would cause this?
The existing `DimmableVideo` implementation also has `Masking = true` set even though nothing should actually get masked away.

peppy: This shouldn't be the case and requires further investigation. We shouldn't be working around this.

voidedWarranties: Is an appropriate solution to allow `UserDimContainer` to use a `BufferedContainer`?

smoogipoo: This is probably a framework issue which I initially caught but thought it had been fixed, relating to the non-masking shader of `VideoSprite`. Will look into it and fix.

smoogipoo: Please test again using the referencing o!f PR.

voidedWarranties: Works great!
Should masking be reimplemented per-layer in a separate PR this one depends on?

peppy (dismissed):

peppy (approved):

PR #8170, opened 2020-03-08
Apply tournament client usability changes

peppy:
This contains several usability improvements that don't deserve their own PR. It's mostly code quality changes that don't belong in upcoming design refresh PRs.

Please read individual commit messages if clarification is required.

### DIFF
--- a/osu.Game.Tournament/Components/TournamentMatchChatDisplay.cs
+++ b/osu.Game.Tournament/Components/TournamentMatchChatDisplay.cs
@@ -66,6 +66,10 @@ namespace osu.Game.Tournament.Components
             }
         }
 
+        public void Expand() => this.FadeIn(300);
+
+        public void Contract() => this.FadeOut(200);
+
         protected override ChatLine CreateMessage(Message message) => new MatchMessage(message);
 
         protected class MatchMessage : StandAloneMessage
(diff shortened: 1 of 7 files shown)


osu.Game.Tournament/Components/TournamentMatchChatDisplay.cs
```diff
@@ -66,6 +66,10 @@ private void load(MatchIPCInfo ipc)
             }
         }
 
+        public void Expand() => this.FadeIn(300);
+
+        public void Contract() => this.FadeOut(200);
```
bdach: Was it intended that the Y movement logic was culled? After this PR it seems to look broken:

image

(screenshot taken in `TestSceneGameplayScreen` after jerry-rigging it to have a channel with messages).

peppy: it gets fixed in the gameplay screen update. i just removed here for simplicity.

bdach: Figured that might be the case, but it was going to be a while until I got to it. Good to know anyways 👍

bdach (approved):

smoogipoo (approved):

PR #8172, opened 2020-03-08
Add better flow logic to map pool layout when few beatmaps are present

peppy:
Previously we would toggle between 3-wide or 2-wide depending on the size of the map pool (to fit everything on screen). This automates the decision process.

- [x] Add test


peppy:
Should be ready for review now.

bdach (dismissed):
Took me a second to figure out how this works but it's tournament client so if it works, it works. Makes sense to me 👍

smoogipoo (approved):

PR #8175, opened 2020-03-08
Increase flexibility of StarCounter component

peppy:
Allows custom drawables and transitions for `StarCounter`. Am willing to add a test scene for this if required.


bdach:
Looking up and down the diff looks like a mechanical enough change but I necessarily wouldn't mind having a test.

I'd also include the [rename of `CountStars` to `Current` in `TeamScore`](https://github.com/ppy/osu/pull/8176/files#diff-368a643a1cbb58a4968b4f34705c6cc1R40) as not to break build status here, unless this won't get merged outright but instead as part of #8176.

peppy:
Ah, the reason I didn't add a test is that there is sneakily already one in `TestSceneScoreCounter`... I'll split it out.

bdach (approved):

smoogipoo (approved):
seems fine to me

PR #8176, opened 2020-03-08
Implement 2020 gameplay design

peppy:
- [x] Depends on #8175.
- [x] Depends on #8171.

Updates positioning of a lot of elements and includes changes to SongBar. Further code quality improvements can still be made here, this is an initial implementation.

This also hides scores/stars during warmup by default.

### DIFF
--- a/osu.Game.Tournament/Components/TournamentBeatmapPanel.cs
+++ b/osu.Game.Tournament/Components/TournamentBeatmapPanel.cs
@@ -27,7 +27,7 @@ namespace osu.Game.Tournament.Components
         private readonly string mods;
 
         private const float horizontal_padding = 10;
-        private const float vertical_padding = 5;
+        private const float vertical_padding = 10;
 
         public const float HEIGHT = 50;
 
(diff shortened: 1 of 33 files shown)


osu.Game.Tournament/Components/TournamentBeatmapPanel.cs
```diff
@@ -70,8 +70,8 @@ private void load(LadderInfo ladder, TextureStore textures)
                 new FillFlowContainer
                 {
                     AutoSizeAxes = Axes.Both,
-                    Anchor = Anchor.TopCentre,
-                    Origin = Anchor.TopCentre,
+                    Anchor = Anchor.CentreLeft,
+                    Origin = Anchor.CentreLeft,
```
bdach: These make the text look off:

image

A few `TopLeft`s lower down and some spacing adjustments seem to bring this to a visually good-looking state.

bdach (commented):
Tests are broken because `ladder.CurrentMatch` is null when running headless (as headless tests use different storage to test browser), therefore dying when trying to determine if it's warmup time or not by team scores. Caching a local `LadderInfo` seems too much trouble to be worth it so maybe just easier to coerce to 0 / nullcheck and call it a day.

osu.Game.Tournament/Components/TournamentBeatmapPanel.cs
peppy: Weird merge failure (seems like this was fixed on the map pool screen PR rather than the base one).

bdach (approved):
Looks all fixed 👍

smoogipoo (approved):

PR #8177, opened 2020-03-08
Implement 2020 schedule design

peppy:

### DIFF
--- a/osu.Game.Tournament.Tests/Screens/TestSceneScheduleScreen.cs
+++ b/osu.Game.Tournament.Tests/Screens/TestSceneScheduleScreen.cs
@@ -2,6 +2,8 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Game.Tournament.Components;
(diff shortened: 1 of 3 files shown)


bdach:
While clicking through this I noticed something that I felt was unintended, which was that every time I would click on a match, any match where one team won the winner box would flash white ([see video](https://streamable.com/1b8r5)). The root cause of this seems to be that `TournamentMatch` reconstructs children on every team update:

https://github.com/ppy/osu/blob/2ef8714d6713730bcb33f665842aa3aaeca94fff/osu.Game.Tournament/Screens/Ladder/Components/DrawableTournamentMatch.cs#L258-L262

The cause of the flash is that the drawable starts off with white but then fades off to its target blue colour in the winner's case, but on the normal case it just instantaneously gets gray. Setting gray outright also doesn't help since that just makes winner boxes flash gray themselves.

peppy:
@bdach yeah, it's been that way forever due to the reconstruction. I'll probably look at improving the situation (by not reconstructing the whole fillflow) in a future pass, but since this PR hasn't really regressed that logic I'm going to let it be for now.

smoogipoo (dismissed):

smoogipoo (approved):

PR #8178, opened 2020-03-08
Fix video looping not propagating when set too early in initialisation

peppy:


bdach (approved):
Haven't tested personally but makes sense 👍

smoogipoo (approved):

PR #8179, opened 2020-03-08
Implement 2020 win screen design

peppy:


bdach:
Not actually sure how to test this one, so can't check it out...

smoogipoo (approved):

PR #8180, opened 2020-03-08
Implement 2020 intro screen design

peppy:


bdach (approved):
Looks ok. Slightly annoying (for me) that the test relies on data but I punched in some and got it working, so doesn't matter ¯\\\_(ツ)\_/¯

smoogipoo (approved):

PR #8182, opened 2020-03-08
Fix hit error ticks getting out of the hit error meter

### LINKED ISSUE #8097: Hit error ticks extend past score meter
**Describe the bug:** While playing (only tested mania), I managed to get hit error ticks to extend past the score meter. I've been able to reproduce this by letting go of hold notes very early.

image

**osu!lazer version:** 2020.229.1

**Logs:** not needed, easily reproducible

Game4all:
Closes #8097.

As the ticks are relatively positioned on the Y axis, having a relative position which is negative or over 1 causes the ticks to appear out of the error meter for some gamemodes (ctb and mania).
This PR adds a clamp to ensure the relative position on Y axis is always in the range [0;1]


bdach:
This seems like a strange solution. Mind explaining *why* the position ends up being over/under 1? Is that due to hitwindow differences?

Honestly instead of clamping position I'd say we might want to employ alpha fade over the range (so fade out ticks at both ends) and not even display outside of [0,1] range.

Game4all:
The relative position on Y axis is for the hit ticks is calculated using this method:

https://github.com/Game4all/osu/blob/e21db7dbbeb124080b759957066f27d75623b28a/osu.Game/Screens/Play/HUD/HitErrorMeters/BarHitErrorMeter.cs#L227

Notes hit very early or late may (issue case) return a negative value or a value over 1 through this method as the the corresponding time offset is largely off the maximum hit window.

As for discarding ticks out of the [0; 1] instead of clamping them to [0; 1], I would agree it is better but I would like someone else input on this

smoogipoo:
The question was why can the time be earlier or later than the maximum hit window?

This happens in mania because the hold note has an applied leniency:
https://github.com/ppy/osu/blob/460bd993ca2c7afcd50b08f6914af8b9b2630c32/osu.Game.Rulesets.Mania/Objects/Drawables/DrawableHoldNoteTail.cs#L36

I think this solution is fine for now, reasoning being that we don't want to increase the hit window display or ticks wouldn't correspond to the judgement you got for normal notes, and the display doesn't show multiple hit windows (yet? ever?).

smoogipoo (approved):

PR #8192, opened 2020-03-10
Fix hyperdash not initiating correctly when juice streams are present

### LINKED ISSUE #8016: osu!catch red fruits don't work sometimes
https://streamable.com/70uzp

**osu!lazer version:** 
commit f1c97451d976c3e1d3fcd4bb9701302b188d6378

peppy:
Hyperdash wasn't working correctly with non-fruit objects (that have their judgements ignored). This makes hyperdash state decision changes only happen on fruit.

Closes #8016


smoogipoo:
Hmm I can still replicate the test failure at 200% speed, which is pretty scary.

peppy:
We probably need to change the way player tests run their virtual track.

smoogipoo (approved):

PR #8197, opened 2020-03-10
Fix osu! shaking instead of missing for early hits

smoogipoo:


peppy (dismissed):

peppy:
Might be best to add a late hitobject test as well.

peppy (approved):

PR #8198, opened 2020-03-10
Add catcher kiai/fail animation states

### LINKED ISSUE #7903: Catcher can't be animated
In stable the catcher sprite can be animated.

### LINKED ISSUE #7908: Idle/fail/kiai catcher states are not implemented
image

peppy:
- Closes https://github.com/ppy/osu/issues/7903.
- Closes #7908.

image


bdach:
As a general note, not necessarily relevant to merging this: while I was playtesting this, the very first play I did I got a major hitch at the first transition from normal to kiai sprite (looked like an I/O stall, maybe half a second at most). It's very likely not apparent on SSD drives (I use a HDD), and I have not been able to reproduce that again.

I am however still seeing these with performance logging enabled:

image

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -155,11 +155,21 @@ private void load()
...
+            private Drawable catcherSprite;
+
+            private void updateCatcher()
+            {
+                catcherSprite?.Expire();
+
+                Add(catcherSprite = createCatcherSprite().With(c =>
+                {
+                    c.Anchor = Anchor.TopCentre;
+                }));
```
bdach: Related to the comment above: maybe worth considering caching the three sprites and cycling the cached instances? It might be a bit of an issue with animated sprites, though (to make sure animation cycle is restarted upon switch).

osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
```diff
@@ -47,7 +85,8 @@ private void createCatcher(float size)
                 Child = new TestCatcherArea(new BeatmapDifficulty { CircleSize = size })
                 {
                     Anchor = Anchor.CentreLeft,
-                    Origin = Anchor.TopLeft
+                    Origin = Anchor.TopLeft,
+                    CreateDrawableRepresentation = ((DrawableRuleset<CatchHitObject>)catchRuleset.CreateInstance().CreateDrawableRulesetWith(new CatchBeatmap())).CreateDrawableRepresentation
```
bdach: That's a *bit* of a line, but it's test code so doesn't matter too much, I suppose.

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
smoogipoo: I think this is fine for now...

PR #8199, opened 2020-03-10
Add random rotation and scale factors to osu!catch bananas

### LINKED ISSUE #7905: osu!catch bananas should scale in when falling

peppy:
Matches stable for now.

- Closes https://github.com/ppy/osu/issues/7905

### DIFF
--- a/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
+++ b/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
@@ -33,10 +33,11 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
             ScaleContainer.ScaleTo(HitObject.Scale * (end_scale + random_scale_range * RNG.NextSingle()))
                           .Then().ScaleTo(HitObject.Scale * end_scale, HitObject.TimePreempt);
 
-            const float random_angle_range = 180;
+            ScaleContainer.RotateTo(getRandomAngle())
+                          .Then()
+                          .RotateTo(getRandomAngle(), HitObject.TimePreempt);
 
-            ScaleContainer.RotateTo(random_angle_range * (RNG.NextSingle() * 2 - 1))
-                          .Then().RotateTo(random_angle_range * (RNG.NextSingle() * 2 - 1), HitObject.TimePreempt);
+            float getRandomAngle() => 180 * (RNG.NextSingle() * 2 - 1);
         }
 
         private Color4 getBananaColour()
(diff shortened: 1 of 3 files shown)


osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
```diff
@@ -22,6 +23,22 @@ protected override Color4 GetComboColour(IReadOnlyList<Color4> comboColours)
...
+            const float end_scale = 0.6f;
+            const float random_scale_range = 1.6f;
+
+            ScaleContainer.ScaleTo(HitObject.Scale * (end_scale + random_scale_range * RNG.NextSingle()))
+                          .Then().ScaleTo(HitObject.Scale * end_scale, HitObject.TimePreempt);
+
+            const float random_angle_range = 180;
+
+            ScaleContainer.RotateTo(random_angle_range * (RNG.NextSingle() * 2 - 1))
+                          .Then().RotateTo(random_angle_range * (RNG.NextSingle() * 2 - 1), HitObject.TimePreempt);
```
bdach: Perhaps an opportunity for a (local?) function here in
```csharp
float GetRandomAngle() => random_angle_range * (RNG.NextSingle() * 2 - 1);
```
to avoid duplication but it's a nitpick.

osu.Game/Tests/Visual/ScreenTestScene.cs
```diff
@@ -38,12 +38,11 @@ protected ScreenTestScene()
...
         private void addExitAllScreensStep()
         {
-            AddUntilStep("exit all screens", () =>
+            AddStep("exit all screens", () =>
             {
-                if (Stack.CurrentScreen == null) return true;
+                if (Stack.CurrentScreen == null) return;
 
                 Stack.Exit();
-                return false;
```
bdach: Not a 100% about this change; it makes the tests run faster in browser but if anything ever breaks in screen exiting logic then we might not know about it (current version will only exit topmost if I understand correctly). Maybe this should unwind to the top in a manual `while` loop so that we know if it does - or maybe we just adjust step name so that this no longer claims to exit all screens?

bdach (commented):
Visually looks quite nice 👍

osu.Game/Tests/Visual/ScreenTestScene.cs
peppy: didn't mean to commit this here, will undo

smoogipoo (approved):

PR #8200, opened 2020-03-10
Add afterimage glow when entering hyperdash

### LINKED ISSUE #7906: Catcher should show lasting glow sprites at each hyperdash endpoint

peppy:
- Closes #7906


bdach (approved):
Looks good to me for now 👍

smoogipoo (approved):

ch4tie:
Currently there doesn't seem to be support for the HyperDashAfterImage line in the skin.ini file. All hyperdash afterimages are a bright red, while this was changeable in the stable version. Not entirely sure if the cause is HyperDashAfterImage since I can't locate skin files in lazer. Regardless, would like to be able to customize the color of the glow.

bdach:
That should be supported (see #8448). If it isn't, try seeing if you don't have beatmaps skin disabled, and if you don't and it still doesn't work, please open an issue.

PR #8201, opened 2020-03-10
Fix hyperdash test having a zero-length juice stream

peppy:
Also changes steps to not false-positive.


osu.Game.Rulesets.Catch.Tests/TestSceneHyperDash.cs
```diff
@@ -49,39 +53,50 @@ protected override IBeatmap CreateBeatmap(RulesetInfo ruleset)
...
+
+        private class TestJuiceStream : JuiceStream
+        {
+            public TestJuiceStream(float x)
+            {
+                X = x;
+                Path = new SliderPath(new[]
+                {
+                    new PathControlPoint(new Vector2(x, 0)),
+                    new PathControlPoint(new Vector2(x + 30, 0)),
```
bdach: Slightly confused as to why these use `x` at all but I suppose it doesn't *hurt*?

bdach (dismissed):
👍

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDash.cs
smoogipoo: It is actually wrong, it means that the path has an implicit empty linear segment from (0, 0) to (x, 0).

smoogipoo (approved):

PR #8202, opened 2020-03-10
Don't play samples on catching a tiny droplet

### LINKED ISSUE #5100: Droplets sounds too intense on some osu!catch beatmaps
The addition of droplet sounds has made quite a few maps sound really awful (about 1/4 of the handful I tested) when they sound fine on stable.  Some adjustments will need to be made or a lot of old maps will sound really bad. (ie sounds like random off beat drum hits etc just spammed throughout the song)

Will link a few examples below

peppy:
For now we'll remove these. We need to create some new sounds to put in their place, at some later date.

- Closes https://github.com/ppy/osu/issues/5100


bdach (approved):
Code looks ok, seems to be working fine in-game. Might be useful to have a new issue up for reintroducing the samples since the removal could raise a few heads after next release.

smoogipoo:
Stable doesn't play the tiny tick samples either, fwiw.

PR #8204, opened 2020-03-10
Fix osu!catch hitobjects appearing / fading too late

### LINKED ISSUE #6902: Fruits fade in too late
**Describe the bug:** 
I suffered from this from 3 months. I cannot play catch over 40 min because my eyes are very painful about identifying fruits and bananas. Fruits appear completely too late than osu!stable or lazer before 0828 ver.  
https://www.dropbox.com/s/honfalli88tsnvy/after0828_2.jpg?dl=0
https://www.dropbox.com/s/10lw6x9zfoi9uxy/after0828_3.jpg?dl=0  
https://www.dropbox.com/s/6z7s14hqtjiaz73/before0828_2.jpg?dl=0   
https://www.dropbox.com/s/htfx5lajnrn2wab/before0828_3.jpg?dl=0
**osu!lazer version:** 
0828-1113
**Logs:**

peppy:
- Closes #6902

This is to be considered a temporary fix. `ScrollingHitObjecctManager` needs work to allow extensions outside the drawable bounding box to be considered. The current method of using `OriginPosition` to do so is not working (see taiko / catch examples below)

image

image


bdach (commented):
Ouch, yeah... I didn't expect that origin wasn't necessarily going to reflect the actual dimensions of the drawable subtree. Seems like a good enough fix for now but it could be addressed better for sure.

peppy:
We were discussing that instead of using `OriginPosition`, either exposing a custom extension range (at a new `DrawableScrollingHitObject` class) or using `BoundingBox` / `ScreenSpaceDrawQuad` at `ScrollingHitObjectManager` instead.

smoogipoo (approved):

PR #8208, opened 2020-03-11
Use framework extension method for FromHex

### LINKED ISSUE #8143: OsuColour hex code parsing can be removed
It can be removed upon the merging of https://github.com/ppy/osu-framework/pull/3332.

peppy:
Closes #8143


iiSaLMaN:
Should be closing https://github.com/ppy/osu/issues/8143

smoogipoo:
We should consider making this an extension method on Color4 instead. Using `Color4Extensions` everywhere is ugly.

smoogipoo:
Turns out you can't have non-instanced extension methods... Let's go with this for now.

smoogipoo (approved):

swoolcock:
As a reminder, we will need to move away from osuTK `Color4` eventually.

PR #8211, opened 2020-03-11
Update text on disclaimer screen (and add tips)

peppy:
Just making things a little less serious, while also removing the "early development build" classification (because it really isn't any more) and adding some tips to give users something to do while waiting for the disclaimer to go away.

Tips can eventually be moved to the main menu once we remove the disclaimer completely.

### DIFF
--- a/osu.Game/Screens/Menu/Disclaimer.cs
+++ b/osu.Game/Screens/Menu/Disclaimer.cs
@@ -200,7 +200,7 @@ namespace osu.Game.Screens.Menu
                 "Try scrolling down in the mod select panel to find a bunch of new fun mods!",
                 "Most of the web content (profiles, rankings, etc.) are available natively in-game from the icons on the toolbar!",
                 "Get more details, hide or delete a beatmap by right-clicking on its panel at song select!",
-                "All delete operations are temoprary until quit. Restore accidentally deleted content from the maintenance settings!",
+                "All delete operations are temporary until exiting. Restore accidentally deleted content from the maintenance settings!",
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Menu/Disclaimer.cs
```diff
@@ -178,5 +183,35 @@ public override void OnEntering(IScreen last)
...
+                "New features are coming online every update. Make sure to stay up-to-date!",
+                "If you find the UI too large or small, try adjusting UI scale in settings!",
+                "Try adjusting the \"Screen Scaling\" mode to change your gameplay or UI area, even in fullscreen!",
+                "For now, osu!direct is available to all users on lazer. You can access it anywhere using Ctrl-D!",
+                "Seeking in replays is available by dragging on the difficulty bar at the bottom of the screen!",
+                "Multithreading support means that even with low \"FPS\" your input and judgements will be accurate!",
+                "Try scrolling down in the mod select panel to find a bunch of new fun mods!",
+                "Most of the web content (profiles, rankings, etc.) are available natively in-game from the icons on the toolbar!",
+                "Get more details, hide or delete a beatmap by right-clicking on its panel at song select!",
+                "All delete operations are temoprary until quit. Restore accidentally deleted content from the maintenance settings!",
```
smoogipoo:
```suggestion
                "All delete operations are temporary until exiting. Restore accidentally deleted content from the maintenance settings!",
```

smoogipoo (changes requested):

PR #8212, opened 2020-03-11
Update loader animation

peppy:


osu.Game/Graphics/UserInterface/LoadingSpinner.cs
```diff
@@ -19,15 +19,16 @@ public class LoadingSpinner : VisibilityContainer
...
 
         private const float spin_duration = 900;
 
         /// <summary>
         /// Constuct a new loading spinner.
         /// </summary>
         /// <param name="withBox">Whether the spinner should have a surrounding black box for visibility.</param>
-        public LoadingSpinner(bool withBox = false)
+        /// <param name="inverted">Whether colours should be inverted (black spinner instead of white).</param>
+        public LoadingSpinner(bool withBox = false, bool inverted = false)
```
bdach: Not a major fan of the `inverted` param to be quite honest, it's a little specific. If we're drawing a line and saying there is not going to be any more design variety in this component then I'd say it's at the limit of being fine, but if there are other places then I'd probably switch to properties (and maybe even have `{Foreground,Background}Colour` for the inverted case).

peppy: I'm not sure I agree with properties as I'm trying to limit the scope of how much this component can be changed (ie. we probably don't ever want something which isn't one of these two colour options).

smoogipoo (approved):

PR #8213, opened 2020-03-11
Fix osu!catch trail animating (and displaying incorrect frame)

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/8198
- [x] Depends on https://github.com/ppy/osu-framework/pull/3355


bdach:
Conflict resolution looks non-trivial. Seems like the answer is to reuse `CatcherTrailSprite` in `SetHyperDashState` (replacing existing call to `createAdditiveSprite`), but that would force duplication of code used to access current catcher texture, so seems best to push that down there.

bdach (commented):
I fixed up a merge conflict here and removed some methods which were no longer used since the resolution of the previous one.

The code looks good and like it should work fine, but I can't see a very good way of checking if it works correctly in test scenes.

smoogipoo (approved):

PR #8218, opened 2020-03-11
Limit maximum number of ticks per slider span

smoogipoo:
Helps RAM usage for some beatmaps that have many ticks + many repeats + an inner set of ticks (catch). I found that this could cause RAM usage to skyrocket otherwise.


smoogipoo:
Blocking while I assess whether this is required again.

smoogipoo:
I think we're going to go down a different path here and limit the total processing time instead.

PR #8227, opened 2020-03-11
Improve visual behaviour of osu! hidden mod

### LINKED ISSUE #8063: For "Increase visibility of first object when visual impairment mods are enabled," ignore Spinners and work after breaks.
Spinners give no intuitive indication for rhythm, approach rate, circle size, or where the map will begin its structure, making this feature nearly pointless should a map start with a spinner (which for most cases will devolve into a normally usable spinner, just without the meter). 

My proposal is to find the first non-spinner object and remove visual impairment from it- whether spinners still have their visual impairment removed as well is neutral to me- and to also enable this for after breaks, to futureproof for maps that will feature variable AR and CS which most likely will be used in different sections with different pacing.

### LINKED ISSUE #8153: First hitcircle not vanishing with HD

Screenshots:
https://imgur.com/a/Yg8nREW

osu!lazer version:
2020.306.0

Logs:

LanTenDragon:
Fixes https://github.com/ppy/osu/issues/8063
Fixes https://github.com/ppy/osu/issues/8153

The most I can do is a visual test without asserts.
I would need help for writing an assert test.


osu.Game/Rulesets/Mods/ModHidden.cs
```diff
@@ -58,5 +58,9 @@ public ScoreRank AdjustRank(ScoreRank rank, double accuracy)
         protected virtual void ApplyHiddenState(DrawableHitObject hitObject, ArmedState state)
         {
         }
+
+        protected virtual void ApplySpecialHiddenState(DrawableHitObject hitObject, ArmedState state)
+        {
+        }
```
bdach: Does this need to exist in the base? Seems like it could be local to the osu! implementation, and it's a bit weird having a *special* hidden state that isn't really used anywhere else.

bdach (commented):
Haven't looked at helping out with the test. Might try to do so in the coming days.

osu.Game/Rulesets/Mods/ModHidden.cs
peppy: no matter where this exists, if you are going to call something "special" it *needs* xmldoc

osu.Game.Rulesets.Osu/Mods/OsuModHidden.cs
```diff
@@ -34,7 +36,30 @@ public override void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> dr
                     adjustFadeIn(h);
             }
 
-            base.ApplyToDrawableHitObjects(drawables);
+            int skipCount = 0;
+
+            if (IncreaseFirstObjectVisibility.Value)
+            {
+                ++skipCount;
```
peppy: why not just `drawable.FirstOrDefault(d => !(d is Spinner))`?

osu.Game.Rulesets.Osu/Mods/OsuModHidden.cs
```diff
@@ -90,5 +110,54 @@ protected override void ApplyHiddenState(DrawableHitObject drawable, ArmedState
...
+                else
+                    return startingSpinnersCount;
+            }
+
+            return startingSpinnersCount;
+        }
+
+        private void getFadeTimes(OsuHitObject h)
+        {
+            fadeOutStartTime = h.StartTime - h.TimePreempt + h.TimeFadeIn;
```
peppy: you can't set fields in a get method and then use outside. needs some rethinking.

LanTenDragon: So it should be something akin to ``AutoProperty``?

osu.Game.Rulesets.Osu/Mods/OsuModHidden.cs
```diff
@@ -34,7 +36,30 @@ public override void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> dr
...
+                {
+                    IEnumerable<DrawableHitObject> startingSpinners = drawables.Take(spinnerCount);
+
+                    foreach (var d in startingSpinners)
+                        d.ApplyCustomUpdateState += ApplyHiddenState;
+
+                    skipCount += spinnerCount;
+                }
+
+                DrawableHitObject firstHitObject = drawables.First();
```
LanTenDragon: I have forgotten to skip the spinners, so the first drawable I was taking is actually a spinner. Weird, *was* working as intended before

peppy:
@LanTenDragon thanks for the contribution. in reviewing I decided that we probably want to do this in a simplified fashion to better allow the behaviour to be implemented in other rulesets.

this changes the visual slightly (spinners at the start will not become "hidden") but i don't think this is a huge deal, and maybe even makes more sense.

Opened an alternative as it diverges from your implementation somewhat (mainly to make it easier to compare which implementation we want to go with) https://github.com/ppy/osu/pull/8769

peppy:
Closing (replaced by #8769 & #10427).

PR #8230, opened 2020-03-11
Fix catcher showing miss sprite upon missing bananas

EVAST9919:


bdach:
Having a test for this behaviour would be nice.

EVAST9919:
We already have one `TestSceneDrawableHitObjects`

bdach:
...which doesn't have asserts, so if this ever breaks the test is useless for checking this logic.

EVAST9919:
Added

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -345,21 +345,24 @@ public bool AttemptCatch(CatchHitObject fruit)
                 if (validCatch)
                     updateState(fruit.Kiai ? CatcherAnimationState.Kiai : CatcherAnimationState.Idle);
                 else
-                    updateState(CatcherAnimationState.Fail);
+                {
+                    if (!(fruit is Banana))
+                        updateState(CatcherAnimationState.Fail);
+                }
```
peppy: can combine else and if here

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -345,21 +345,24 @@ public bool AttemptCatch(CatchHitObject fruit)
...
+                if (CurrentState == state)
                     return;
 
-                currentState = state;
+                CurrentState = state;
                 updateCatcher();
             }
 
-            private CatcherAnimationState currentState;
+            public CatcherAnimationState CurrentState;
```
peppy: private set

peppy (changes requested):

PR #8235, opened 2020-03-12
Disable adjusting volume via "select next" and "select previous" as fallbacks

### LINKED ISSUE #8169: Gameplay clicking doesn't block global volume adjust
As reported [here](https://osu.ppy.sh/comments/1042399), it would seem that binding gameplay keys to down / up arrow don't block global volume adjust correctly.

peppy:
I think it's best we just keep one set to dedicated keys to volume for now for simplicity.

Mouse wheel still works, which is the main one we want to support as a fallback option.

- Closes #8169


smoogipoo (approved):
Seems fine

PR #8238, opened 2020-03-12
Prevent editor clock from continuing beyond song end

### LINKED ISSUE #8134: Can play beyond song end in editor
**Describe the bug:** I can play beyond the song's end in the editor.

Steps to reproduce:
- Edit a song
- Play, then scrub to end of song. If the song wraps back to the beginning, pause and play again
- Observe

**Screenshots or videos showing encountered issue:** https://youtu.be/lkx13waO9es

**osu!lazer version:** 94d5cc593d9c79edce974f3839132d2a83ed7d25

**Logs:**
[database.log](https://github.com/ppy/osu/files/4287551/database.log)
[network.log](https://github.com/ppy/osu/files/4287552/network.log)
[performance.log](https://github.com/ppy/osu/files/4287553/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4287554/runtime.log)

voidedWarranties:
Fixes #8134

Behavior now matches stable


osu.Game/Screens/Edit/EditorClock.cs
```diff
@@ -37,6 +39,23 @@ public EditorClock(ControlPointInfo controlPointInfo, double trackLength, Bindab
...
+            {
+                if (!PlaybackFinished)
+                {
+                    PlaybackFinished = true;
+                    Stop();
+                    Seek(TrackLength);
+                }
+            }
+            else
+                PlaybackFinished = false;
```
bdach: This is super weirdly written. If the track is running then `false` gets assigned every frame, but when it finishes `true` gets assigned only once. It feels (a) way too complex (if it's not a bottleneck, and I'd be very surprised if it was, it should just assign every frame), and (b) if we're even doing microoptimisations like this then it should probably be the other way around since the `true` path is the (*very*) unlikely one.

bdach (changes requested):
This has non-trivial behaviour (editor clock stops upon reaching end and then restarts when it is sought to end and re-started), and as such it should probably be tested to avoid it regressing in the future.

bdach (dismissed):
I think this is fine for now 👍

smoogipoo:
This seems to behave weirdly in the actual editor. When manually seeking to the end while the track is playing, it can restart the track: https://drive.google.com/file/d/1dzD3UM8JTGvcMvkAS6OaC464AJXEJNgK

voidedWarranties:
I've fixed this behavior, and also prevented the `Timeline` from forcing the clock to seek past the end of the track.

peppy:
This needs proper testing (hopefully via a test scene with asserts). It's more broken than before, I think:

image

voidedWarranties:
I think the preview track looping feature is not correctly disabled when accessing the editor via the context songselect menu. (Most of) the brokenness shown above is present in the latest release.
This behavior works correctly when used via the main menu.

peppy:
I see. Please open an issue for that if you can. Would still appreciate test coverage.

peppy:
As an aside, the new behaviour does not match stable. stable makes specific allowances for storyboarding outside of the track end time:

image

peppy:
This PR is approaching the issue in the wrong way. To stop the editor playing beyond the end (which was intentionally done this way to test behaviour we will need in the future (see above example for storyboard editor)) the editor clock just needs to be set to coupled mode.

https://github.com/ppy/osu/blob/ebe3cf296640971fb9d95be0fc76faa1be548545/osu.Game/Screens/Edit/Editor.cs#L81

The only remaining issue after setting this (and fixing a framework bug with coupled seeking) is to decide whether to restart the track or continue playing.

peppy:
One concern of using `IsCoupled = true` is that all seek operations will be blocking on async bass operations, for a `TrackBass`. This may be an oversight or may be something we have to live with. It is one reason to potentially leave the clock decoupled, but at this point it defeats the purpose of having the `IsCoupled` toggle in the first place (it was made with the editor in mind).

bdach:
I have attempted necromancy on this branch to have it available without the merge conflicts if we choose to work on top of it, as it seems to be the better than the other one PR'd today.

That said, I'm not a huge fan of this branch's logic, and the following test that I'd expect to pass fails on this branch:

```csharp
        [Test]
        public void TestWrapWhenStoppedAtTrackEnd()
        {
            AddStep("reset clock", () => Clock.Seek(0));

            AddStep("stop clock", Clock.Stop);
            AddAssert("clock running", () => !Clock.IsRunning);

            AddStep("seek exactly to end", () => Clock.Seek(Clock.TrackLength));
            AddAssert("clock stopped at end", () => Clock.CurrentTime == Clock.TrackLength);

            AddStep("start clock again", Clock.Start);
            AddAssert("clock looped to start", () => Clock.IsRunning && Clock.CurrentTime < 500);
        }
```

Admittedly I'm not very familiar with the woes of the decoupleable clock, so I'm not sure whether I can make that target solution work. @peppy do you perhaps recall what the problem was there?

voidedWarranties:
The most recent commit moves the looping logic to `Start`, which I think is a bit less sketchy.
It also fixes the failing test above.

osu.Game/Screens/Edit/EditorClock.cs
```diff
@@ -216,7 +222,21 @@ public void SeekSmoothlyTo(double seekDestination)
...
+            playbackFinished = CurrentTime >= TrackLength;
+
+            if (playbackFinished && !playbackAlreadyStopped)
+            {
+                if (IsRunning)
+                    underlyingClock.Stop();
+
+                underlyingClock.Seek(TrackLength);
+            }
+        }
```
bdach, line 239: I'm thinking that this could be simplified further to

```csharp
        public void ProcessFrame()
        {
            underlyingClock.ProcessFrame();

            playbackFinished = CurrentTime >= TrackLength;

            if (playbackFinished)
            {
                if (IsRunning)
                    underlyingClock.Stop();

                if (CurrentTime > TrackLength)
                    underlyingClock.Seek(TrackLength);
            }
        }
```

Reasoning for why:
* Gets rid of another conditional variable
* Ensures that the time always is rolled back to `TrackLength` (I can imagine a scenario where a clock seek stops at `TrackLength` first, but then, because in the current code we have `playbackAlreadyStopped`, you could technically seek further past the track length, as that variable being `true` makes the code inside dead).

Thoughts?

voidedWarranties: Seems like a no-brainer to me

peppy:
@bdach I'd have to double check, but I think it has to do something with the editor doing `Transform` logic on the time of the clock, and if it was changed to be directly coupled, each frame would trigger a seek on the bass track, causing a considerable overhead.

In fact, I remember this being a thing at one point, so the decoupling may actually be turned on as a "workaround" for that..

peppy:
As a temporary change I think this looks fine. There's still a larger piece of the puzzle remaining, which is sharing offset/clock logic with `GameplayClock` (see #11400) at which point this logic may need to be removed or replaced.

Let's get this in for now.

osu.Game.Tests/Visual/Editing/TestSceneEditorClock.cs
```diff
@@ -0,0 +1,81 @@
...
+        [Test]
+        public void TestStopAtTrackEnd()
+        {
+            AddStep("reset clock", () => Clock.Seek(0));
+
+            AddStep("start clock", Clock.Start);
+            AddAssert("clock running", () => Clock.IsRunning);
+
+            AddStep("seek near end", () => Clock.Seek(Clock.TrackLength - 250));
+            AddUntilStep("clock stops", () => !Clock.IsRunning);
```
peppy, line 58: Super not confident this is going to hold up on CI, but I guess we'll see in time. I can't repro a failure locally.

peppy (approved):

peppy:
While testing this I'm 101% sure there are going to be further changes required, but I will PR those as follow-ups.

PR #8242, opened 2020-03-12
Apply ruleset filter in all cases (even when bypassing filter for selection purposes)

peppy:
Fixes non-current rulesets becoming incorrectly selectable during filter bypassing.


osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs
```diff
@@ -25,18 +25,18 @@ public override void Filter(FilterCriteria criteria)
...
             base.Filter(criteria);
 
+            bool match =
+                criteria.Ruleset == null ||
+                Beatmap.RulesetID == criteria.Ruleset.ID ||
+                (Beatmap.RulesetID == 0 && criteria.Ruleset.ID > 0 && criteria.AllowConvertedBeatmaps);
+
             if (Beatmap.BeatmapSet?.Equals(criteria.SelectedBeatmapSet) == true)
             {
                 // bypass filtering for selected beatmap
```
bdach:
```suggestion
                // only check ruleset equality or convertability for selected beatmap
```

bdach (commented):
Makes sense to me 👍

bdach (approved):

smoogipoo (approved):

PR #8245, opened 2020-03-12
Fallback on invalid AnimationFramerate for legacy skins

kOchirasu:
Had some skins with AnimationFramerate set to -1. This seems to be the way the stable client handles these cases (by experiment)

### DIFF
--- a/osu.Game/Skinning/LegacySkinExtensions.cs
+++ b/osu.Game/Skinning/LegacySkinExtensions.cs
@@ -61,7 +61,7 @@ namespace osu.Game.Skinning
             {
                 var iniRate = source.GetConfig<GlobalSkinConfiguration, int>(GlobalSkinConfiguration.AnimationFramerate);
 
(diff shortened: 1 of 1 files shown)


peppy (approved):

osu.Game/Skinning/LegacySkinExtensions.cs
```diff
@@ -61,7 +61,7 @@ private static double getFrameLength(ISkin source, bool applyConfigFrameRate, Te
             {
                 var iniRate = source.GetConfig<GlobalSkinConfiguration, int>(GlobalSkinConfiguration.AnimationFramerate);
 
-                if (iniRate != null)
+                if (iniRate != null && iniRate.Value > 0)
```
peppy:
```suggestion
                if (iniRate?.Value > 0)
```

peppy (changes requested):

PR #8246, opened 2020-03-12
Improve beatmap carousel / song select fallback logic pathing

peppy:
- [x] Depends on #8242

Moves fallback logic (ie. what to do when we don't match something we're trying to select) out of `BeatmapCarousel`. General clean-ups.

This does change behaviour of the *carousel* slightly, but in the context of `SongSelect` it should match.


osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -399,7 +399,7 @@ public void TestSelectingFilteredRuleset()
             AddStep("filter to ruleset 0", () =>
                 carousel.Filter(new FilterCriteria { Ruleset = rulesets.AvailableRulesets.ElementAt(0) }, false));
             AddStep("select filtered map skipping filtered", () => carousel.SelectBeatmap(testMixed.Beatmaps[1], false));
-            AddAssert("unfiltered beatmap selected", () => carousel.SelectedBeatmap.Equals(testMixed.Beatmaps[0]));
+            AddAssert("unfiltered beatmap not selected", () => carousel.SelectedBeatmap == null);
```
LittleEndu: Not really sure what this is testing. Shouldn't the line above it make a selection and this line check if something was selected? From SelectBeatmap's docstring: ``If bypassFilters is false, we will try to select another unfiltered beatmap in the same set.``

bdach: As far as I can understand it's testing what was intended, and that is that the carousel itself no longer makes any selections on its own where no item matches criteria. The xmldoc is two years old now, and needs to be trimmed to reflect the current state of affairs.

bdach (commented):
I *think* this is fine? This stuff is super finicky as usual, so I mostly trust tests passing.

bdach (approved):

smoogipoo (approved):

PR #8247, opened 2020-03-12
Add the ability to click filtered difficulty icons

peppy:
Adds the ability to click filtered difficulty icons on beatmap set panels to switch to another ruleset (or to a filtered difficulty).

- [x] Depends on https://github.com/ppy/osu/pull/8246


osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
```diff
@@ -612,6 +614,46 @@ public void TestDifficultyIconSelecting()
...
+            {
+                set = songSelect.Carousel.ChildrenOfType<DrawableCarouselBeatmapSet>().FirstOrDefault();
+                return set != null;
+            });
+
+            DrawableCarouselBeatmapSet.FilterableDifficultyIcon difficultyIcon = null;
+            AddStep("Find an icon for different ruleset", () =>
+            {
+                difficultyIcon = set.ChildrenOfType<DrawableCarouselBeatmapSet.FilterableDifficultyIcon>()
+                                    .First(icon => icon.Item.Beatmap.ID == 3);
```
bdach: Works first try on a freak accident but fails on a re-run due to silently relying on beatmap IDs. Should probably be
```suggestion
                                    .First(icon => icon.Item.Beatmap.RulesetID == 3);
```

bdach (commented):
Think this one is fairly straight-forward on its own, save for one test foible. Not sure what appveyor has done here building the second commit from last.

osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
bdach: Literally just opened the next one in chain after this and saw that this is fixed there, so...

peppy: Yeah, was wondering why this wasn't fixed when I explicitly remember fixing it.

bdach (dismissed):

smoogipoo (dismissed):

smoogipoo (approved):

PR #8248, opened 2020-03-12
Add the ability to click grouped beatmap difficulty icons

peppy:
Will select the first difficulty in the grouping. Quite useful for beatmaps with tens of difficulties.

- [x] Depends on https://github.com/ppy/osu/pull/8247


bdach (approved):
Seems straightforward enough 👍

PR #8251, opened 2020-03-12
Improve beatmap carousel's auto selection

### LINKED ISSUE #2638: Filtering beatmaps after null selection
While fixing #2264 I never looked at what stable does. In stable the selection never goes to null, as in there's always some beatmap playing/displayed in the background, which is good in one situation.

A user might mistype while trying to further filter their selection, which in lazer would result in you losing your index (for a lack of better word) after being placed to random spot after your now correct filter. In stable the selection would just go back to the previous beatmap that was playing before the filter matched no beatmaps.

Example:
* You first search ``touhou``
* Go through half your library and then want to further refine the search term
* You now search for ``touhou imperishable`` but mistype imperishable
* In stable you are greeted with ``No matches found. Hit ESC to reset.`` and after correcting your mistake you can continue browsing your library from where you left off.
* In lazer you are greeted with ``no beatmaps avalibe! please load a beatmap!`` and after correcting your mistake are sent to random spot in your library and you have to find the spot you were at before. 

I propose that instead of random selection, last valid beatmap is used, unless that beatmap is now filtered in which case keep the current behaviour. Last valid beatmap could be just the last one, could be last 50 ones and the algorithm chooses the latest.

### LINKED ISSUE #8239: Song select doesn't autoselect after entering on no-beatmap ruleset
- enter song select with a ruleset that has no beatmaps
- switch to ruleset with beatmaps
- selection doesn't autoselect

Note that after manually making a selection, this starts to work as expected. Currently no test coverage.

image

LittleEndu:
Makes beatmap carousel select from previous beatmaps when returning from nothing being selected. This selection also happens on every filter change.

* Fixes #8239
* Resolves #2638
* [x] Depends on #8248
* [x] Needs to fix failing test due to new CarouselRoot behaviour ([see this comment](https://github.com/ppy/osu/pull/8246#discussion_r391818451))
* [ ] Add tests for new carousel root behaviour


peppy:
I don't understand or see any explanation as to why more than one beatmap is stored for this functionality.

LittleEndu:
The user might filter out the very last selected beatmap but still expect the selection after new filter to not be random. I need to figure out how "present beatmap" works so that the tests I didn't expect to fail would still pass, then I'll write tests for this new functionality. Made the PR early to see if this is the kind of hard to merge code you mentioned in the issue.

peppy:
I still don't understand how a user could ever want to go to anything but the last selection

ie. why you need the list.

LittleEndu:
I would expect if I search "monstrata" for example that the last beatmap I selected mapped by monstrata would be my new selection, and not that my selection will be random if my previous selection gets filtered out. That's why I used a list instead of a single instance, or are you thinking about something else?

peppy:
that sounds super weird. i believe only the last selection should be recalled.

LittleEndu:
While writing a test for this I found out that carousel behaviour already matches stable's so the current behaviour is probably very much indented.

Behaviour after this PR would then be:
* change filter -> carousel selected nothing -> clear filter -> carousel selects last beatmap (instead of random)
* apply filter -> carousel selected something -> refine filter -> carousel selects nearest next beatmap (I thought this also used to be random, but it clearly isn't)

So there already is a test for #8239 in this PR but idk if there should also be test for #2638 and I also don't see an obvious test for selecting the next beatmap behaviour.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -694,10 +730,20 @@ private BeatmapSetInfo createTestBeatmapSetWithManyDifficulties(int id)
...
+            public TestBeatmapCarousel()
+            {
+                StartAllowingSelectionOnFiltered();
+            }
+
             public new List<DrawableCarouselItem> Items => base.Items;
 
             public bool PendingFilterTask => PendingFilter != null;
 
+            public void ForgetPreviouslySelectedSets()
```
LittleEndu: This should be renamed

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -496,6 +496,7 @@ public override void OnEntering(IScreen last)
             FilterControl.Activate();
 
             ModSelect.SelectedMods.BindTo(selectedMods);
+            Scheduler.AddDelayed(Carousel.StartAllowingSelectionOnFiltered, 250);
```
peppy: this whole thing is awkward

LittleEndu: This is meant to wait for all the automatic beatmap bindings to not change values anymore before allowing beatmap carousel to react to user changes to the filter. Every call to filter function before actually entering song select is automatic and not issued by the user.

idk if there is a better way to do this.

bdach:
I've looked at this PR 3 times in the last 3 days and each time I put reviewing it off since the changes seem completely inscrutable to me. The carousel & filtering logic is already arcane at places but this brings it up to 11. I don't understand the inline comments at all and would probably have to spend an hour minimum to be able to.

If this can't be done cleaner I'd be against bringing this behaviour back, unless it becomes viable to add back at some point in the future.

LittleEndu:
Maybe the filter functionality should be separated to it's own class, call it FilterManager or something. In the same vein, maybe carousel should only deal with displaying CarouselBeatmap instances and there should be a separate SelectionManager.

peppy:
I still think this new logic should be able to be added with basically no changes. I'll give it a try.

Issue #5090, opened 2019-06-20
Add a toggle for disabling Discord Rich Presence


Vezqi:
**Describe the new feature:** 

Adding a toggle to the settings menu, which would allow you to either disable the Rich Presence feature completely, or hide rank and username info (but still display the current game mode + map info).

**Proposal designs of the feature:** 

If a user has the 'Hide User Info' feature toggled, it would be neat if the the RPC could update with 'Anonymous' instead of the user's username.

Two toggles, which are listed below could be added to the settings menu:

→ Enable/Disable Discord Rich Presence.
→ Show/Hide User Info

peppy:
Aren't toggles for this already at discord's end? Should we really be adding our own?

Vezqi:
Discord only allows you to disable all game info. I think it would be a nice feature so that my friends can still see that I'm playing osu, but without the RPC + user info and whatnot.

Gabixel:
Maybe this can be implemented using the "Do not disturb" option in the future _(because right now it's just W.I.P.)_

![DoNotDisturb](https://user-images.githubusercontent.com/43073074/59917153-6fadbf00-9421-11e9-9dc3-5e08f34f2ea8.png)

Craftplacer:
> Aren't toggles for this already at discord's end? Should we really be adding our own?

You've done them for Windows Live Messenger, so.....

![](https://files.catbox.moe/vp23v1.png)
![](https://files.catbox.moe/l4adr6.png)

Also, the rich presence option allows me and others to "hunt down" players on osu!.

Joehuu:
Recently added in stable, so not a proposal now. Closing and making https://github.com/ppy/osu/issues/1109 track this implementation.

Joehuu:
Reopening as this hasn’t been added in the initial implementation.

PR #8258, opened 2020-03-13
Add show/hide toggle for Discord Rich Presence

jdmremi:
~~So, let's try this again~~

This PR is for a feature I requested (#5090), for allowing a user to either show their username/rank info with Discord's Rich Presence, or to hide their username/rank info (for a few reasons...). I have already acknowledged that the original issue has been closed, but gave this a shot.

I added a toggle under the 'Web' settings for allowing a user to either hide/show their info. Default value for this toggle is false.

image

Here's how it looks when a user has "Hide Discord Rich Presence information" disabled:

image

Here's how it looks when a user has "Hide Discord Rich Presence information" enabled:

image

### DIFF
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -9,6 +9,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Logging;
+using osu.Game.Configuration;
 using osu.Game.Online.API;
 using osu.Game.Rulesets;
 using osu.Game.Users;
@@ -19,15 +20,23 @@ namespace osu.Desktop
 {
     internal class DiscordRichPresence : Component
     {
-        private const string client_id = "367827983903490050";
+        public DiscordRichPresence(OsuConfigManager config)
+        {
+            this.config = config;
+        }
 
+        private const string client_id = "367827983903490050";
         private DiscordRpcClient client;
+
+        private bool isHidden;
+        private string usernameInfo;
+        private string rankInfo;
 
         [Resolved]
         private IBindable<RulesetInfo> ruleset { get; set; }
 
         private Bindable<User> user;
-
+        private OsuConfigManager config;
         private readonly IBindable<UserStatus> status = new Bindable<UserStatus>();
         private readonly IBindable<UserActivity> activity = new Bindable<UserActivity>();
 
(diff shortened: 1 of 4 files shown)


Craftplacer:
The title (Add show/hide toggle for Discord Rich Presence) is a bit misleading because it just filters information and does not disable rich presence.
I would suggest making a toggle for **actually disabling rich presence**.

peppy:
You can already disable discord rich presence from discord's end, FWIW. On a per game basis.

peppy:
I'm not sure about this change in general. I feel if you have something to hide here, you're likely doing it (life?) wrong. Could someone offer a potential use case for this that would warrant not just suggesting the user turns off rich presence at discord's end altogether?

Craftplacer:
> You can already disable discord rich presence from discord's end, FWIW. On a per game basis.

I remember being able to disable the playing status of a game in Discord (in the games panel in Activity), but with Discord deprecating its gaming platform, I couldn't find this feature again (right-/clicking on osu! in Discord doesn't show any menu and I can't hide it anymore).

> I'm not sure about this change in general. I feel if you have something to hide here, you're likely doing it (life?) wrong. Could someone offer a potential use case for this that would warrant not just suggesting the user turns off rich presence at discord's end altogether?

You can stalk the osu! username of others (PM'ing them on osu!, spectating, etc.), would be one of the cases.

Vezqi:
I don't mind hiding it completelty or by the idea I proposed. My idea comes from games like Fortnite/R6 etc. which only show maps/activity, not your username. I just moderate a few large servers and maybe I do want people to see what I'm playing, but not my username and all that. 

image

image

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -19,15 +20,23 @@ namespace osu.Desktop
 {
     internal class DiscordRichPresence : Component
     {
-        private const string client_id = "367827983903490050";
+        public DiscordRichPresence(OsuConfigManager config)
```
peppy: Rather than passing in the whole config, expose a bindable and perform the binding in `OsuGameDesktop`. Look at something like `DrawableRuleset.IsPaused` for an example of how this is done.

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -107,6 +122,11 @@ private void updateStatus()
 
         private static readonly int ellipsis_length = Encoding.UTF8.GetByteCount(new[] { '…' });
 
+        private bool checkRPC()
```
peppy: shouldn't be required after switching to bindable flow.

peppy (changes requested):

peppy:
@Vezqi are you planning on applying the review feedback to this?

Vezqi:
> @Vezqi are you planning on applying the review feedback to this?

Yes. Life is a bit hectic at the moment, so I hope to complete this relatively soon. 👍

PR #8259, opened 2020-03-13
Switch game to use new font (Torus)

peppy:
- [x] Loosely depends on https://github.com/ppy/osu-resources/pull/91

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -51,7 +51,7 @@
     <Reference Include="Java.Interop" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="ppy.osu.Game.Resources" Version="2020.304.0" />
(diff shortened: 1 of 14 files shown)


bdach (dismissed):
image

Some places are a little weird after the change (e.g. changelog author names are wrongly anchored) but those are separate pulls. (Same thing will probably apply with the introduction of Inter - I assume it's also coming, right?)

peppy:
We're not 100% on inter yet, so leaving that one a while. Torus is here to stay though, yep.

smoogipoo (dismissed):

smoogipoo:
Only issue I noticed is that the usernames in chat are supposed to be "bolditalic", but since that font isn't supported it falls back to regular. It should probably fallback to bold in that case, but I don't really see that much of a problem with regular in chat.

Other than that, lgtm.

bdach (approved):

PR #8261, opened 2020-03-13
Fix mod sprite bleeding border colour

peppy:
Fixes this:
image


smoogipoo:
For reference, this is another case of https://github.com/ppy/osu-framework/issues/1675

smoogipoo (dismissed):

PR #8262, opened 2020-03-13
Add resolution selector in tournament setup screen

peppy:


osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -97,9 +108,25 @@ private void reload()
...
                     Current = LadderInfo.Ruleset,
                 },
+                resolution = new ActionableInfo
+                {
+                    Label = "Stream area resolution",
+                    ButtonText = "Set to 1080p",
+                    Action = () =>
+                    {
+                        windowSize.Value = new Size((int)(1920 / TournamentSceneManager.STREAM_AREA_WIDTH * TournamentSceneManager.REQUIRED_WIDTH), 1080);
+                    }
```
bdach: This button action actually causes a hard freeze on Windows. From a cursory look the set goes down all the way to osuTK - `WinGLNative.Size` calls into `Functions.SetWindowPos` and never returns. Best guess at current would be that it ends up being called from the wrong thread.

bdach (dismissed):

osu.Game.Tournament/Screens/SetupScreen.cs
peppy: very weird. tested and working on macOS and windows here..

bdach: hmm, I'll dig into that a little bit more then...

peppy: If this is unsafe, it likely should be fixed framework-side.

bdach: OK, I think I can see what is going on here. The button is trying to set the window size to 2369x1080, which exceeds the physical size of both my monitors and that's probably why the hang occurs. Going to investigate a bit more, if I can't figure out if this can be fixed easily I'll open an issue about it framework-side.

smoogipoo (approved):

PR #8267, opened 2020-03-13
Fix osu! slider ball tint colour being applied to follow circle

### LINKED ISSUE #8137: Legacy skins incorrectly apply AllowSliderBallTint to sliderfollowcircle
**Describe the bug:** 
Legacy skins have two elements that follow the slider: sliderbN and sliderfollowcircle. When AllowSliderBallTint is enabled in the skin.ini, slider ball is supposed to get tinted the same color as the combo color is (in stable there is an setting to overwrite this skin behaviour) but the follow circle is supposed to stay the same color as it is. 

Screenshots using stable's default skin template where only edit is ``AllowSliderBallTint = 1``

osu_2020-03-04_20-42-39
screenshot180

**osu!lazer version:** 
2020.229.1

EVAST9919:
Fixes https://github.com/ppy/osu/issues/8137

Example with skin:
Stable:
stable
Master:
current
PR:
new

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
@@ -23,9 +23,16 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
     {
         public Func<OsuAction?> GetInitialHitAction;
 
+        public new Color4 Colour
+        {
+            get => ball.Colour;
+            set => ball.Colour = value;
+        }
+
         private readonly Slider slider;
         private readonly Drawable followCircle;
         private readonly DrawableSlider drawableSlider;
+        private readonly CircularContainer ball;
 
         public SliderBall(Slider slider, DrawableSlider drawableSlider = null)
         {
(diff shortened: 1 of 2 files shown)


frenzibyte (changes requested):
Solution seem to be OK code-wise, but what about using `AccentColour` instead?

peppy:
Can agree with that thought. Much less error-prone than using `new`.

peppy (changes requested):
Use AccentColour

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SliderBall.cs
```diff
@@ -16,16 +16,30 @@
...
+
+        public Color4 AccentColour
+        {
+            get => accentColour;
+            set
+            {
+                accentColour = value;
+                ball.Colour = value;
+            }
+        }
```
bdach: Why is this written in such a convoluted way, instead of:
```suggestion
        public Color4 AccentColour
        {
            get => ball.Colour;
            set => ball.Colour = value;
        }
```
That way (a) reads easier (it's obvious at a glance that it's a delegating property), (b) is less error-prone in case somebody ever exposes `ball` for some reason, (c) gets rid of the field.

peppy (approved):

bdach (approved):

PR #8270, opened 2020-03-13
Switch to the next chat tab upon closing an active one

thewildtree:
Supersedes #7466.

## Summary
Brings chat tab closing behaviour to what's intended - if possible, switches to the next opened chat tab upon closing an active one. If not possible (if the `ChannelSelectorTab` is the next tab), switches to the previous channel, and if there are no channels left, opens the channel selector.

Tests taken from #7466, with some small readibility improvements and more explicit checks added.

## Remarks
There are issues with other tests in the `ChatOverlay` scene.

Right now, other tests use the PM chat tabs which change their size and have different `CloseButton` appereance behaviour. Closing of these tabs is done via `clickDrawable`. A problem arises when attempting to use that on non-PM tabs, though - due to the `CloseButton` not showing up instantly on hover, using `clickDrawable` there won't work properly - we'd have to explicitly hover the chat tab and only then attempt to click the button, but that's not a good solution IMO.

(notice the first click being ignored)

Due to that, my tests use `ChannelTabControl.RemoveChannel()`, as I kind of had to use non-PM tabs for them (due to their static width). 

Should we just switch to using the `RemoveChannel()` in all cases for consistency? That would let us switch to having non-PM tabs in all tests (or at least mix them in), which kind of reflects the actual use case a bit more I think.

### DIFF
--- a/osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
+++ b/osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
@@ -39,7 +39,7 @@ namespace osu.Game.Overlays.Chat.Tabs
             public ChannelSelectorTabChannel()
             {
                 Name = "+";
-                Type = ChannelType.Temporary;
+                Type = ChannelType.System;
             }
         }
     }
--- a/osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
+++ b/osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
@@ -39,6 +39,7 @@ namespace osu.Game.Overlays.Chat.Tabs
             public ChannelSelectorTabChannel()
             {
                 Name = "+";
+                Type = ChannelType.Temporary;
             }
         }
     }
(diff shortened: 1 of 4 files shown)


osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
```diff
@@ -39,6 +39,7 @@ public class ChannelSelectorTabChannel : Channel
             public ChannelSelectorTabChannel()
             {
                 Name = "+";
+                Type = ChannelType.Temporary;
```
smoogipoo: Is there a reason why you're using the temporary type rather than a new "system" type or something like that?

smoogipoo (changes requested):
Other than that, these changes look good.

osu.Game/Overlays/Chat/Tabs/ChannelSelectorTabItem.cs
TheWildTree: No reason honestly, it just didn't cross my mind at all. Will do.

smoogipoo (approved):

TheWildTree:
So, other tests are fine to stay as they are? Thought we'd maybe want to fix the inconsistencies mentioned in the description too.

smoogipoo:
I was fine with using `RemoveChannel`, since functionally there isn't any difference. Further work can be done on the tests in a separate PR.

I think it would be good to have a test that _just_ clicks the close button. It can do it in two steps - move mouse to close button, click. As two steps, it'd avoid your issue with the button not showing up instantly.

But apart from that using `RemoveChannel` elsewhere for simplicity is fine imo.

PR #8284, opened 2020-03-15
Implement Approach Circle mod

BlauFx:
# Introduction:
This pull request introduces a new fun mod called "Approach Circle".
In the customisation menu of this mod you can change the easing type of the approach circle's animation and it's scale.

# Preview:
[https://www.youtube.com/watch?v=fvo6Il4uxb4](https://www.youtube.com/watch?v=fvo6Il4uxb4)


osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
```diff
@@ -0,0 +1,59 @@
...
+            Value = 4
+        };
+
+        public void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> drawables)
+        {
+            drawables.ForEach(drawable =>
+            {
+                drawable.ApplyCustomUpdateState += (drawableHitObj, state) =>
+                {
+                    if (!(drawableHitObj is DrawableOsuHitObject drawableOsuHitObj) || !(drawableHitObj is DrawableHitCircle hitCircle)) return;
```
bdach: Redundant check:
```suggestion
                    if (!(drawableHitObj is DrawableHitCircle hitCircle)) return;
```
A `DrawableHitCircle` *is* a `DrawableOsuHitObject` through inheritance and I'm pretty sure `HitObject` should still be publicly accessible there.

osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
```diff
@@ -0,0 +1,59 @@
...
+    internal class OsuModApproachCircle : Mod, IApplicableToDrawableHitObjects
+    {
+        public override string Name => "Approach Circle";
+        public override string Acronym => "AC";
+        public override string Description => "Never trust the approach circles...";
+        public override double ScoreMultiplier => 1;
+        public override IconUsage? Icon { get; } = FontAwesome.Regular.Circle;
+
+        [SettingSource("Easing", "Change the easing type of the approach circles.", 0)]
+        public Bindable<Easing> BindableEasing { get; } = new Bindable<Easing>();
```
bdach: This can't be a setting in its current state. The enum member names are unreadable to non-programmers. It's bad UX to display to the user things like `OutQuint`, `InSine`, etc. - use human-readable names.

bdach (changes requested):
As an honest *personal* opinion, I'm unconvinced that this needs to be a mod. It feels like a proposal made because it's easy to implement, and not necessarily something people would play. I certainly never would myself (looks visually too busy and distracting). Happy to be proved wrong if people think otherwise.

Few notes to the code that should not be subjective, though.

osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
```diff
@@ -0,0 +1,73 @@
...
+                };
+            });
+        }
+
+        internal enum ReadableEasing
+        {
+            Accelerate = 2,
+            Accelerate2 = 6,
+            Accelerate3 = 12,
+            AccelerateInAfterDeceleraingeOut = 29,
```
bdach:
```suggestion
            AccelerateInAfterDeceleratingOut = 29,
```

osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
```diff
@@ -0,0 +1,73 @@
...
+            CasualBounces = 32,
+            CasualBouncesWhileAccelerating = 24,
+            Decelerate = 1,
+            DecelerateAfterAccelerating = 8,
+            Default = 0,
+        }
```
bdach: * It's bad practice to rely directly on the underlying numerical values of the `Easing` enum. If someone adds another enum value in between the existing ones in the framework it'll silently break this mapping since the numerical values depend on member order in code. Should try to reference the values directly through casting if anything.
* The names are only *slightly* better and IMO demonstrate just how arbitrary allowing this setting for easing curve is. How does _Accelerate_ differ from _Accelerate 2_ and _Accelerate 3_? What are _casual bounces_? With other mods, you pretty much know what you're going to get with them. But this one - it's tweaking implementation details of an animation.

bdach (changes requested):

osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
TheWildTree: This would make sense if we had some sort of a visualizer (just a hitcircle visualizing the currently chosen animation) next to the setting dropdown - but I guess it's not really possible in the current mod menu, nor is it feasible to really try to implement given its scope.

peppy: Remove all numbers here and reorder so "Default" is at the top. Make a mapping function. Also what is a "casual" bounce?

osu.Game.Rulesets.Osu/Mods/OsuModApproachCircle.cs
```diff
@@ -0,0 +1,73 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Configuration;
+using osu.Game.Rulesets.Mods;
+using osu.Game.Rulesets.Objects.Drawables;
+using osu.Game.Rulesets.Osu.Objects.Drawables;
+
+namespace osu.Game.Rulesets.Osu.Mods
+{
+    internal class OsuModApproachCircle : Mod, IApplicableToDrawableHitObjects
```
peppy: can we call this something more imaginative?

peppy (changes requested):

LanTenDragon:
Just my suggestion: Let this be a mod for mappers 

Reasoning: I feel that as a player, I would play this mod once for the novelty and that's it. But for mappers, the possibilities for them are vast. This looks like it would fit just right in Aspire style maps.

smoogipoo:

peppy:
I might pick this one up, since it's pretty close to being usable and should take 5-10 minutes. Feels like a mod I'd enjoy playing.

Will re-open if I get to it.

peppy:
Closing in favour of #13514.

PR #8288, opened 2020-03-16
Implement FriendDisplay component

EVAST9919:
Split from https://github.com/ppy/osu/pull/8051


osu.Game.Tests/Visual/Online/TestSceneFriendsLayout.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void TestPopulate()
+        {
+            AddStep("Populate", () => layout.Users = getUsers());
+        }
+
+        private List<APIFriend> getUsers() => new List<APIFriend>
+        {
+            new APIFriend
+            {
+                Username = @"flyte",
```
bdach: No idea why these strings have raw string prefixes at all.

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
```diff
@@ -0,0 +1,256 @@
...
+            }
+        }
+
+        private List<APIFriend> sortUsers(List<APIFriend> unsorted)
+        {
+            switch (userListToolbar.SortCriteria.Value)
+            {
+                default:
+                case UserSortCriteria.LastVisit:
+                    return unsorted.OrderBy(u => u.LastVisit).Reverse().ToList();
```
bdach:
```suggestion
                    return unsorted.OrderByDescending(u => u.LastVisit).ToList();
```

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
```diff
@@ -0,0 +1,256 @@
...
+
+                case UserSortCriteria.Rank:
+                    return unsorted.Where(u => u.CurrentModeRank.HasValue).OrderBy(u => u.CurrentModeRank).Concat(unsorted.Where(u => u.CurrentModeRank == null)).ToList();
```
bdach:
```suggestion
                    return unsorted.OrderByDescending(u => u.CurrentModeRank.HasValue).ThenBy(u => u).ToList();
```

osu.Game/Overlays/SocialOverlay.cs
```diff
@@ -120,7 +120,7 @@ private void updateSearch()
             {
                 case SocialTab.Friends:
                     var friendRequest = new GetFriendsRequest(); // TODO filter arguments?
-                    friendRequest.Success += users => Users = users.ToArray();
+                    friendRequest.Success += users => Users = users.Select(u => (User)u).ToArray();
```
bdach:
```suggestion
                    friendRequest.Success += users => Users = users.Cast<User>().ToArray();
```

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
```diff
@@ -0,0 +1,256 @@
...
+            request = new GetFriendsRequest();
+            request.Success += response => Schedule(() => Users = response);
+            api.Queue(request);
+        }
+
+        private void recreatePanels()
+        {
+            // Don't allow any changes until we have users loaded
+            if (!usersLoaded)
+                return;
```
bdach: Why does this flag need to exist? Can the same effect be achieved by early-returning if there are no users?

osu.Game/Online/API/Requests/Responses/APIFriend.cs
```diff
@@ -0,0 +1,14 @@
...
+using Newtonsoft.Json;
+using osu.Game.Users;
+
+namespace osu.Game.Online.API.Requests.Responses
+{
+    public class APIFriend : User
+    {
+        [JsonProperty(@"current_mode_rank")]
+        public int? CurrentModeRank;
+    }
```
bdach: This inheritance feels gratuitous. Naming aside (what is an `APIFriend`?), as far as I can see web [does not return a full user here, but instead a trimmed structure with only partial fields filled out](https://github.com/ppy/osu-web/blob/4687a1e8a5c132af3ccfc63a2b90d239ec395d71/app/Http/Controllers/FriendsController.php#L46). If you try to pass this data around pretending it's a `User` it'll break somewhere where an actual `User` with the full data is required (in short, this breaks the Liskov substitution principle).

bdach (changes requested):
Where is the online code tested? I don't see a test scene or step that would hit the API.

Also, when switching between different filters, the user card status circle will flash white ([see video](https://streamable.com/mftt4)).

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
EVAST9919: Before:
my
After:
Screenshot_1

bdach: Not sure what your point is? The two last cards both have a `null` value on `LastVisit`. `OrderBy` is a stable sort (meaning that if two items have the same value of the attribute to sort by, it preserves the initial order of the items). The reason for the swap is, because you did the `Reverse` after sorting, the effects of the stable sort got reversed too. So the new behaviour is arguably the correct one.

EVAST9919: That's why we need an online test, which is currently in progress!

EVAST9919: Yeah, everything is fine

osu.Game/Online/API/Requests/Responses/APIFriend.cs
EVAST9919: Well, we have many cases like this in-game (where web does not return a full user, but we are using one)

EVAST9919: Should I just add this property to `User`?

peppy: we cannot use subclassing on models. it is not supported by realm and not the direction we want to take things.

osu.Game.Tests/Visual/Online/TestSceneFriendsLayout.cs
```diff
@@ -0,0 +1,98 @@
...
+                Country = new Country { FlagName = "BY" },
+                CoverUrl = "https://assets.ppy.sh/user-profile-covers/8195163/4a8e2ad5a02a2642b631438cfa6c6bd7e2f9db289be881cb27df18331f64144c.jpeg",
+                IsOnline = false,
+                LastVisit = DateTimeOffset.Now
+            }
+        };
+
+        private class TestFriendsLayout : FriendsLayout
+        {
+            public FriendsOnlineStatusControl StatusControl => OnlineStatusControl;
```
bdach: Not necessary to expose `OnlineStatusControl` for the tests when you can do:
```suggestion
            public FriendsOnlineStatusControl StatusControl => this.ChildrenOfType<FriendsOnlineStatusControl>().Single();
```

osu.Game/Users/User.cs
```diff
@@ -69,6 +69,9 @@ public class UserCover
         [JsonProperty(@"support_level")]
         public int SupportLevel;
 
+        [JsonProperty(@"current_mode_rank")]
+        public int? CurrentModeRank;
```
bdach: In my opinion this is even worse than `APIFriend` was - now the base user has a property that only one overlay will use.

I would bring back `APIFriend`, and actually *duplicate* the `User` fields that are returned by the API, and *only* those (so as far as I can see from web: `'cover', 'country', 'current_mode_rank', 'support_level'`, and then everything else from [`UserCompact`](https://osu.ppy.sh/docs/index.html#usercompact)). Heck, you could even make an `APIUserCompact` and then derive from *that*, adding the other fields.

I wouldn't be scared of specialised data structures for API interfacing. The concept of data transfer objects has been alive for long.

bdach (changes requested):

osu.Game/Users/User.cs
EVAST9919: IMO it's fine for now. We already have properties in `User` which have been used only in one particular case. I agree, that this should be fixed at some point, but at very least it deserves a separate pr (which I'm not gonna make since I'm not sure how to do that properly)

bdach: I am unconvinced but will defer on this if it turns out that it is only my opinion that this should be fixed here.

EVAST9919: Or probably I should make `APIFriend : APIUser` and place that property there

peppy: In RL discussions, we were talking about moving all API pieces of the user to `APIUser` and derive from that. Which is similar to what @bdach is suggesting.

ie. all serialization logic should be moved out of User and there should be a `ToUser()` function on `APIUser` / whatever, simlar to `APIBeatmapSet`'s implementation.

To clarify, there's no problem with deriving/subclassing if it's API specific models.

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
```diff
@@ -0,0 +1,252 @@
...
+    public class FriendsLayout : CompositeDrawable
+    {
+        private List<User> users = new List<User>();
+
+        public List<User> Users
+        {
+            get => users;
+            set
+            {
+                request?.Cancel();
```
bdach: Why does the setter cancel the request? As far as I can see it's only fired once in `LoadComplete`, and the setter is only used when the request *succeeds*? (Leaving aside the fact that a setter doing API logic is in itself a bit of a code smell.)

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
```diff
@@ -0,0 +1,252 @@
...
+        {
+            if (!users.Any())
+                return;
+
+            cancellationToken?.Cancel();
+
+            if (itemsPlaceholder.Any())
+                loading.Show();
+
+            var groupedUsers = onlineStatusControl.Current.Value?.Users ?? new List<User>();
```
bdach: Needs a better name like `usersInCurrentGroup`. I got confused at what it is for a second.

bdach (changes requested):

osu.Game/Overlays/Dashboard/Friends/FriendsLayout.cs
EVAST9919: It's just to make tests work.
Probably exposing a method like `Fetch` will be a better approach (?)

osu.Game/Overlays/Dashboard/Friends/FriendsOnlineStatusControl.cs
```diff
@@ -5,20 +5,19 @@
...
         public void Populate(List<User> users)
         {
-            var userCount = users.Count;
-            var onlineUsersCount = users.Count(user => user.IsOnline);
+            Clear();
 
-            AddItem(new FriendsBundle(FriendsOnlineStatus.All, userCount));
-            AddItem(new FriendsBundle(FriendsOnlineStatus.Online, onlineUsersCount));
-            AddItem(new FriendsBundle(FriendsOnlineStatus.Offline, userCount - onlineUsersCount));
+            AddItem(new FriendsBundle(FriendsOnlineStatus.All, users));
```
peppy: i don't get this. why are we making three copies of users? why is this not a filter control like every other overlay implementation? at very least, why aren't these live linq queries and instead creating copies of the list?

peppy (changes requested):

peppy (dismissed):

peppy (approved):

PR #8291, opened 2020-03-16
Use new logo name for showcase screen

ItsMestro:


peppy:
This whole class needs to be updated/replaced, but sure I guess this is not a regression.

PR #8300, opened 2020-03-16
Limit Torus font weight to bold

Joehuu:
Fixes torus fonts with `FontWeight.Black` set by changing it to `FontWeight.Bold`.

Didn't want to do fallback method because it hides the real font weight. The main font before (Exo 2) only used the black weight, so it's best removing it for now.

### DIFF
--- a/osu.Game/Graphics/OsuFont.cs
+++ b/osu.Game/Graphics/OsuFont.cs
@@ -136,10 +136,5 @@ namespace osu.Game.Graphics
         /// Equivalent to weight 700.
         /// </summary>
         Bold = 700,
-
-        /// <summary>
-        /// Equivalent to weight 900.
-        /// </summary>
-        Black = 900
     }
 }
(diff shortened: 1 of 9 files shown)


osu.Game/Graphics/OsuFont.cs
```diff
@@ -136,10 +136,5 @@ public enum FontWeight
         /// Equivalent to weight 700.
         /// </summary>
         Bold = 700,
-
-        /// <summary>
-        /// Equivalent to weight 900.
-        /// </summary>
-        Black = 900
```
peppy: not sure about removing here, but i guess it's okay for now?

peppy (approved):

osu.Game/Graphics/OsuFont.cs
smoogipoo: would've preferred to keep it for reference purposes too.

PR #8309, opened 2020-03-17
Implement the top score panel contents

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8308

image

### DIFF
--- a/osu.Game/Screens/Ranking/Expanded/ExpandedPanelTopContent.cs
+++ b/osu.Game/Screens/Ranking/Expanded/ExpandedPanelTopContent.cs
@@ -12,10 +12,17 @@ using osuTK;
 
 namespace osu.Game.Screens.Ranking.Expanded
 {
+    /// <summary>
+    /// The content that appears in the middle section of the <see cref="ScorePanel"/>.
+    /// </summary>
     public class ExpandedPanelTopContent : CompositeDrawable
     {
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Ranking/Expanded/ExpandedPanelTopContent.cs
```diff
@@ -1,15 +1,64 @@
...
+                Direction = FillDirection.Vertical,
+                Children = new Drawable[]
+                {
+                    new UpdateableAvatar(user)
+                    {
+                        Anchor = Anchor.TopCentre,
+                        Origin = Anchor.TopCentre,
+                        Size = new Vector2(80),
+                        CornerRadius = 20,
+                        CornerExponent = 2.5f,
```
smoogipoo: Just remembered I did this, not sure if you want it in like this.

peppy: which part? the corner exponent?

smoogipoo: Corner exponent, yes

peppy: seems sane enough

PR #8311, opened 2020-03-17
Implement the new results screen

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8309
- [x] https://github.com/ppy/osu/pull/8310
- [x] https://github.com/ppy/osu/pull/8305 (fixes crash on autoplay)

image

Currently timeshift won't display the leaderboard, as that will come with the implementation of contracted panels.


Nichtzzz:
The mapper name in the screen is replaced by the player name 🤔

bdach:
[yes, we are aware of this](https://github.com/ppy/osu/issues/8317)

PR #8312, opened 2020-03-17
Fade playfield to red when player health is low

### LINKED ISSUE #8095: Add fade to red on low-HP effect
**Describe the new feature:** 

On stable if you have shaders enabled your screen will start fading to red when your HP start getting low. This is useful to have when you like playing with no UI but still want to know what HP you are at.

Game4all:
Closes #8095

# Summary

This PR implements the `FaillingLayer`, a player overlay component which fades the playfield to red when the player health falls under a defined threshold.
The component is avalaible as a shared component to enable its use by rulesets interested by that functionnality.

## Changes to Player

- Every child of type `HealthDisplay` in `DrawableRuleset.Overlay` will now be bound to the ruleset's HealthProcessor on player loading
## Rulesets on which this has been applied

- osu!standard
- osu!catch
- osu!mania


LittleEndu:
This definitely needs easing. Otherwise high HP drain maps have major flashing during streams.

osu.Game/Screens/Play/HUDOverlay.cs
```diff
@@ -260,6 +262,11 @@ protected virtual ScoreCounter CreateScoreCounter() => new ScoreCounter(6)
             Margin = new MarginPadding { Top = 20 }
         };
 
+        protected virtual HealthDisplay CreateLowHealthDisplay() => new LowHealthLayer
```
bdach: Should probably have xmldoc.

osu.Game/Rulesets/Scoring/HealthProcessor.cs
```diff
@@ -26,6 +31,12 @@ public abstract class HealthProcessor : JudgementProcessor
         /// </summary>
         public readonly BindableDouble Health = new BindableDouble(1) { MinValue = 0, MaxValue = 1 };
 
+        /// <summary>
+        /// The treshold under which the current health value should be considered low.
+        /// Used by HUD elements to display visual indication that health value is low.
+        /// </summary>
+        public virtual double LowHealthTreshold => 0.20f;
```
bdach: I see little reason for this to exist here to then be propagated to all score displays (that might not even need it) instead of it being locally in `LowHealthLayer`. This seems like preemptive design. It can be extracted when there's a use-case for it.

bdach (changes requested):

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -53,6 +53,12 @@ private void load(OsuConfigManager config)
                     Keywords = new[] { "hp", "bar" }
                 },
                 new SettingsCheckbox
+                {
+                    LabelText = "Fade playfield to red when health is low",
+                    Bindable = config.GetBindable<bool>(OsuSetting.FadePlayfieldWhenLowHealth),
+                    Keywords = new[] { "hp", "playfield", "health" }
+                },
```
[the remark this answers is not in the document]
LittleEndu: It is a setting in stable under "Shaders"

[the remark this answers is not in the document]
bdach: That one is kind of legacy by itself though really? I don't imagine lazer is going to be able to run without pixel shaders anywhere, so I'd say that one really has no longer much merit for existing.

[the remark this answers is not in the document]
Game4all: It doesn't exist as a setting per se on stable but is enabled with the special effects setting. (see https://github.com/ppy/osu/issues/8095#issuecomment-593728914)

osu.Game/Screens/Play/HUDOverlay.cs
```diff
@@ -32,6 +32,7 @@ public class HUDOverlay : Container
         public readonly ScoreCounter ScoreCounter;
         public readonly RollingCounter<double> AccuracyCounter;
         public readonly HealthDisplay HealthDisplay;
+        public readonly HealthDisplay LowHealthDisplay;
```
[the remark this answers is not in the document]
Game4all: It is being exposed mainly for test purposes.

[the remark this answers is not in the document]
bdach: I guess that's fine for consistency, even though `ChildrenOfType<T>` exists.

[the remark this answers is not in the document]
Game4all: Let's go with a rename of the field though I agree the naming is weird

osu.Game/Rulesets/Scoring/HealthProcessor.cs
Game4all: Since health handling is not the same accross all rulesets, I had implemented this property to set the threshold at which the health should be considered low (the point at which the `LowHealthLayer` should start fade in)
or to completely disable the fading of the component if the ruleset's health system works instead like in taiko. Also this could be used by the `StandardHealthDisplay` to change the bar accent colour when health gets under this threshold.

bdach: Yeah I'm not sure I like this direction. If the low health layer is going to have ruleset-specific logic then that logic should reside in the rulesets themselves and not in the now-as-reusable-as-possible health processor (the goal with them was to unify health as much as possible, aside from taiko which is so different to begin with that it's unviable).

I'd even go as far as to say say that the proper place for the layers is in the ruleset-specific playfields (if one doesn't want the layer, it just doesn't add it, but if another one does, it "automagically" somehow gets the health processor and knows what its alpha needs to be based on that).

osu.Game/Screens/Play/HUD/FaillingLayer.cs
```diff
@@ -0,0 +1,48 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osu.Game.Graphics;
+
+namespace osu.Game.Screens.Play.HUD
+{
+    /// <summary>
+    /// An overlay layer on top of the playfield which fades to red when the current player health falls a certain threshold defined by <see cref="LowHealthThreshold"/>.
+    /// </summary>
+    public class FaillingLayer : HealthDisplay
```
bdach:
```suggestion
    public class FailingLayer : HealthDisplay
```
same deal in usages, etc.

osu.Game/Screens/Play/HUD/FaillingLayer.cs
```diff
@@ -0,0 +1,48 @@
...
+using System;
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Shapes;
+using osu.Game.Graphics;
+
+namespace osu.Game.Screens.Play.HUD
+{
+    /// <summary>
+    /// An overlay layer on top of the playfield which fades to red when the current player health falls a certain threshold defined by <see cref="LowHealthThreshold"/>.
```
bdach:
```suggestion
    /// An overlay layer on top of the playfield which fades to red when the current player health falls below a certain threshold defined by <see cref="LowHealthThreshold"/>.
```

osu.Game/Screens/Play/HUD/FaillingLayer.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class FaillingLayer : HealthDisplay
+    {
+        private const float max_alpha = 0.4f;
+
+        private readonly Box box;
+
+        /// <summary>
+        /// The threshold under which the current player life should be considered low and the layer should start fading in.
+        /// </summary>
+        protected virtual double LowHealthThreshold => 0.20f;
```
bdach: Not a fan of this being only-protected-get to be fair. Having to subclass just to change this seems senseless, and this component isn't really that reusable anyways (it's just a box with alpha, what's the point of conflating it with anything else?). I'd just make it a plain settable property.

osu.Game/Screens/Play/HUD/HealthDisplay.cs
```diff
@@ -3,15 +3,30 @@
...
         {
             MinValue = 0,
             MaxValue = 1
         };
+
+        /// <summary>
+        /// Bind the tracked fields of <see cref="HealthProcessor"/> to this health display.
+        /// </summary>
+        /// <param name="processor"></param>
+        public virtual void BindHealthProcessor(HealthProcessor processor)
```
bdach: This again seems to be prematurely engineered, why is this `virtual`?

bdach (dismissed):

osu.Game/Screens/Play/HUD/HealthDisplay.cs
Game4all: This was marked virtual to allow implementors of the `HealthDisplay` class to bind to other properties of the `HealthProcessor` if they need to but I guess this can be reverted to a normal method in a lack of a current usage overriding the method.

bdach: As a general rule, make things as simple as they can be when going forward with something new. Don't try to anticipate future usages, most of the time you'll fail anyway. Things can be changed when they need to be changed.

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class FailingLayer : HealthDisplay
+    {
+        private const float max_alpha = 0.4f;
+
+        private readonly Box box;
+
+        /// <summary>
+        /// The threshold under which the current player life should be considered low and the layer should start fading in.
+        /// </summary>
+        public double LowHealthThreshold = 0.20f;
```
bdach: Public fields are sort of eh for me. Would just rather have it a property but it's not anything major.

bdach (commented):
I think this looks ok at the current state, save for one small nitpick.

osu.Game/Screens/Play/HUD/FailingLayer.cs
Game4all: Yeah, it should have stayed an auto-prop but CI inspection complains about it being convertible to an get-only auto-prop so I made it a field to make CI stop complaining about it..

frenzibyte (changes requested):
Is there a reason why you specifically picked inheriting `HealthDisplay`? (I believe it was abstracted away to allow different types of `HealthDisplay`s, osu!stable's health for example). If I were to implement this, I would've done it as a normal composite-drawable component that binds to the health processor and updates its alpha whenever the health changes.

smoogipoo:
This is going to need a config setting, imo.

osu.Game.Tests/Visual/Gameplay/TestSceneFailingLayer.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        private readonly FailingLayer layer;
+
+        public TestSceneFailingLayer()
+        {
+            Child = layer = new FailingLayer();
+        }
+
+        [Test]
+        public void TestLayerFading()
```
bdach: Test lacks assertions.

osu.Game/Screens/Play/HUD/HealthDisplay.cs
```diff
@@ -3,15 +3,30 @@
...
         public readonly BindableDouble Current = new BindableDouble
         {
             MinValue = 0,
             MaxValue = 1
         };
+
+        /// <summary>
+        /// Bind the tracked fields of <see cref="HealthProcessor"/> to this health display.
+        /// </summary>
+        /// <param name="processor"></param>
```
bdach:
```suggestion
```

bdach (changes requested):
The new setting for enabling/disabling this is not tested.

peppy:
I've fixed the following issues:
- HUD elements should not be added at a ruleset level. They are separate layers.
- Check health processor type to avoid this, rather than local initialisation.
- Changed effect to be less full-screen.
- You made a base `HealthDisplay` method class but didn't use the new function for the standard health display's binding.
- All words in `LabelText` are included in keywords. All the keywords you specified were redundant.
- Tests shouldn't use `WaitStep` when `UntilStep` can be used. Also just tidied them up in general (too much copy paste).
- `Lerp` should be used instead of `ValueAt` for ongoing lerps.

Not sure if this needs to be a toggle, but I guess no harm for now.

@smoogipoo request re-review on this.

smoogipoo:
Applied a few more fixes:
* Removed now unnecessary bind inside `Player`.
* Safety for config bind order.

smoogipoo (dismissed):

smoogipoo:
@peppy requesting quick review/merge

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -73,16 +75,29 @@ private void load(OsuColour color, OsuConfigManager config)
...
 
-            // don't display ever if the ruleset is not using a draining health display.
-            if (!(processor is DrainingHealthProcessor))
+            healthProcessor = processor;
+            updateBindings();
+        }
+
+        private void updateBindings()
+        {
+            if (configEnabled == null || healthProcessor == null)
```
peppy, line 92: can we make this a `LoadState >= Ready` check instead? took me a bit of mental effort to ensure this wouldn't bind more than once.

smoogipoo: Just to confirm, you want the initial `updateBindings()` call that's currently inside `load()` to also be moved to `LoadComplete()`?

peppy: nah, fine where it is, but just would rather than conditional check here is on LoadState rather than null checks (i think it will mean the same thing at the end of the day, right?)

smoogipoo: It's not fine where it is, that condition will exactly break the scenario I intended to fix here - the potential of calling `BindHealthProcessor()` before `load()`.

peppy: `if (LoadState < Ready) return` is what we usually use in these cases, along with a call from load or loadcomplete (either is fine).

peppy:
looks fine otherwise

PR #8320, opened 2020-03-17
Fix mapper info alignment in score panel

recapitalverb:
Master: 
image

This PR:
image


bdach:
Judging from 2020 tournament designs this alignment might have been intended, but that will need to be confirmed.

smoogipoo:
Definitely a mistake on my part.

PR #8323, opened 2020-03-17
Re-use colors defined for each rank in result screen accuracy circle

Tina-otoge:
The `AccuracyCircle` is made of 6 different sections each corresponding to a rank, instead of using hardcoded hex values for the 6 different colors, make use of `OsuColour.ForRank` to keep it unified with other places where ranks are displayed.

### DIFF
--- a/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
+++ b/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
@@ -119,42 +119,42 @@ namespace osu.Game.Screens.Ranking.Expanded.Accuracy
                         new SmoothCircularProgress
                         {
                             RelativeSizeAxes = Axes.Both,
-                            Colour = Color4Extensions.FromHex("#BE0089"),
+                            Colour = OsuColour.ForRank(ScoreRank.X),
                             InnerRadius = RANK_CIRCLE_RADIUS,
                             Current = { Value = 1 }
                         },
                         new SmoothCircularProgress
                         {
                             RelativeSizeAxes = Axes.Both,
-                            Colour = Color4Extensions.FromHex("#0096A2"),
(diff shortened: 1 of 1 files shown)


bdach:
As far as I can see this changes the colours? They should be probably updated globally if anything but I am not aware what are the overall design plans for rank colours.

skielred:
The accuracy circle were using values that did not appear anywhere else, but were very similar to the current colour values that can be found in https://github.com/ppy/osu/blob/156a3db57d7841c7eb26a2d2ae76ca0a754b281a/osu.Game/Graphics/OsuColour.cs#L44-L68 and that are already used when displaying ranks in, let's say, leaderboards and user profiles.
The accuracy circle were using colours that are very similar but not equal to the currently already defined colours.
This PR makes the accuracy circle use the same colours that are already used everywhere else, instead of letting it introduce a new set of very similar colour for each ranks.

bdach:
I realise that the colours are different. What I was trying to say is that maybe they're different for a purpose and the other colours might be updated in line with the *new* values at a later point. Or they're just different because they were designed that way. Need confirmation from the team on that.

skielred:
Oh alright didn't understand it this way, thanks for re-explaining to poor me.
If this is intended then this PR is useless yeah.
If all the colours are supposed to get changed, the values in `OsuColour` should be changed to reflect that and avoid confusion like what happened to me.
Waiting for answers, should I ping someone?

bdach:
Not necessary, we should get a conclusive answer once the timezones roll around.

smoogipoo:
In the design they're supposed to appear as a darker tone of the badges, but the values in `OsuColour` are slightly off compared to the design.

The difference is minor but I'll confirm the path forward with @arflyte first.

peppy (approved):

peppy:
Let's get this in. There are further changes to the colour/animation for this component but they should be sharing colours, at least.

PR #8329, opened 2020-03-17
Fix beat divisor control selecting invalid divisors on drag end

### LINKED ISSUE #8326: Beat Divisor Control value can be set outside of valid divisors
**Describe the bug:** When dragging marker it is possible to set value and position of marker outside of valid divisors, it happens only if you start dragging > marker jump > release mouse > it will land on wherever your mouse position was. 
That leading to having all red and small divisors in beatmap editor when it set to 1/7, 1/11 etc.

https://imgur.com/a/3nm2lSl

**osu!lazer version:** 2020.317.0

I don't know how to submit a fix.
So there is a fix, in BeatDivisorControl.cs TickSliderBar class, we need to override OnDragEnd to prevent base class from changing value of the slider.
`
protected override void OnDragEnd(DragEndEvent e) {}
`

Fuewburvpoa:
Resolves #8326.

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/BeatDivisorControl.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/BeatDivisorControl.cs
@@ -281,6 +281,7 @@ namespace osu.Game.Screens.Edit.Compose.Components
 
             protected override void OnDragEnd(DragEndEvent e)
             {
+                handleMouseInput(e.ScreenSpaceMousePosition);
             }
 
             private void handleMouseInput(Vector2 screenSpaceMousePosition)
--- a/osu.Game/Screens/Edit/Compose/Components/BeatDivisorControl.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/BeatDivisorControl.cs
@@ -279,6 +279,10 @@ namespace osu.Game.Screens.Edit.Compose.Components
                 handleMouseInput(e.ScreenSpaceMousePosition);
             }
 
+            protected override void OnDragEnd(DragEndEvent e)
+            {
+            }
+
             private void handleMouseInput(Vector2 screenSpaceMousePosition)
             {
                 // copied from SliderBar so we can do custom spacing logic.
--- a/osu.Game.Tests/Visual/Editor/TestSceneBeatDivisorControl.cs
+++ b/osu.Game.Tests/Visual/Editor/TestSceneBeatDivisorControl.cs
@@ -4,7 +4,6 @@
 using System;
 using System.Collections.Generic;
 using NUnit.Framework;
-using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Game.Screens.Edit;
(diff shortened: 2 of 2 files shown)


osu.Game/Screens/Edit/Compose/Components/BeatDivisorControl.cs
```diff
@@ -279,6 +279,10 @@ protected override void OnDrag(DragEvent e)
                 handleMouseInput(e.ScreenSpaceMousePosition);
             }
 
+            protected override void OnDragEnd(DragEndEvent e)
+            {
+            }
```
bdach: While the diagnosis that the problem lies with the base implementation of `OnDragEnd` seems correct, I'm not overly convinced that silencing the event outright is a better solution than just mirroring `OnDrag`:

```suggestion
            protected override void OnDragEnd(DragEndEvent e)
            {
                handleMouseInput(e.ScreenSpaceMousePosition);
            }
```

bdach (commented):
Could be nice to have a test for this to prevent regressions, although test coverage in editor is currently not that high overall.

Fuewburvpoa:
I will try to write test for this case tomorrow.

Fuewburvpoa:
Added 2 tests.
TestBindableBeatDivisor - test BindableBeatDivisor Next() and Previous() methods.
TestMouseInput - test for #8326

I have one question, is it fine to get location of private elements by simply getting position of the object and finding offset manually? For example in this case i wanted to get position of the Marker that is private member inside private class, so instead of messing up with reflection i just got coords manually.

bdach:
@Fuewburvpoa Your attempt was pretty close! I've just pushed in an adjusted version of the test, that looks for the appropriate drawables in the hierarchy and uses their coordinates to perform the mouse movements.

The selection of the items is a little bit hacky (i.e. moving the marker by getting its inner `EquilateralTriangle` relies on implementation details of the control) but I opted not to expose any of the private inner classes in `BeatDivisorControl`.

Fuewburvpoa:
> @Fuewburvpoa Your attempt was pretty close! I've just pushed in an adjusted version of the test, that looks for the appropriate drawables in the hierarchy and uses their coordinates to perform the mouse movements.
> 
> The selection of the items is a little bit hacky (i.e. moving the marker by getting its inner `EquilateralTriangle` relies on implementation details of the control) but I opted not to expose any of the private inner classes in `BeatDivisorControl`.

I see, thank you for good example, will use it in future cases.

peppy:
@Fuewburvpoa can you explain why adding the `OnDragEnd` is fixing this? Is it due to bypassing the `base.` invocation? Always helps the review process to mention this in the PR description or commit.

bdach:
Yup, the base call in `OnDragEnd` does the normal slider bar logic of clamping position to the nearest tick which breaks the beat divisor as not all tick values are allowed in this case.

peppy (approved):

PR #8336, opened 2020-03-18
Fix perfect display showing when misses are present

### LINKED ISSUE #8319: 'Perfect' displayed wrongly on results screen with misses
**Describe the bug:** 

When a map is finished with the new results screen at the combos section there is writted "PERFECT" but i've made misses so it's not perfect

Screen

**osu!lazer version:** 

2020.317.0

**Logs:**
 Android

peppy:
Closes https://github.com/ppy/osu/issues/8319

Doesn't account for edge cases like "slider breaks" just yet.


MTGPROD:
In the future it will count slider breaks?

smoogipoo:
Yes, but we don't really have a way of doing that right now - the `MaxCombo` property of `BeatmapInfo` only comes from online sources. It'll happen after we database difficulty calculation.

PR #8337, opened 2020-03-18
Block out-of-order judgements from occurring (aka "note lock")

### LINKED ISSUE #8099: Add notelock back with adjusted values
the notelock will have 166ms of leniency which is about 1ms less than 2 1/4s or 1 1/2 on 180bpm

this means that up to 166ms after the current time in gameplay will not have notelock

i suggest this because it will prevent cheesing on lower difficulties but it will not be as annoying on streams on higher difficulties as on stable

smoogipoo:
Resolves https://github.com/ppy/osu/issues/8099

Follows the method proposed by @Karoo13 - no notelock if clicking after the first hitobject's start time or if the first hitobject has been judged.


osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -64,19 +65,42 @@ public override void Add(DrawableHitObject h)
...
+            }
 
             return result;
         }
 
+        private bool checkHittable(DrawableOsuHitObject osuHitObject)
+        {
+            var lastObject = HitObjectContainer.AliveObjects.GetPrevious(osuHitObject);
+
+            // Ensure the last object is not alive anymore, in which case always allow the hit.
```
peppy:
```suggestion
            // If there is no previous object alive, allow the hit.
```

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
```diff
@@ -16,6 +17,12 @@ public class DrawableOsuHitObject : DrawableHitObject<OsuHitObject>
         // Must be set to update IsHovered as it's used in relax mdo to detect osu hit objects.
         public override bool HandlePositionalInput => true;
 
+        /// <summary>
+        /// Whether this <see cref="DrawableOsuHitObject"/> can be hit.
+        /// If not-null, this <see cref="DrawableOsuHitObject"/> will not receive a judgement until this function returns <c>true</c>.
```
peppy:
```suggestion
        /// If non-null, judgements will be ignored (resulting in a shake) whilst the function returns false.
```

peppy (changes requested):

peppy:
Needs to be tested against aspire/other edge cases. Checking on centipede results in ~60% accuracy (100% on master). We likely need some lenience for special cases of equality.

image

Karoo13:
Don't try to merge this until you add the part which causes prior objects to be missed if you do end up going out of order. I think there are some edge cases your implementation missed, I will try to check it.

smoogipoo:
Not sure what you're referring to re: equality comparer. The equality is by reference, and the alive hitobjects are sorted by start time. Therefore retrieving the previous hitobject is equivalent to retrieving the previous hitobject by start time.

Anyway I've applied the change that causes earlier hitobjects to get missed, however I'm going to block this PR for now until more testing is done. Specifically I need automated tests for interactions between circles + sliders, but that will be dependent on https://github.com/ppy/osu/pull/8356 and https://github.com/ppy/osu/pull/8358 .

smoogipoo:
I think this is about ready to go. I'm a little bit wary of the performance, but upon quick benchmarking on centipede I found it's on the order of 0.0%.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -64,21 +67,54 @@ public override void Add(DrawableHitObject h)
 
             base.Add(h);
 
-            followPoints.AddFollowPoints((DrawableOsuHitObject)h);
+            DrawableOsuHitObject osuHitObject = (DrawableOsuHitObject)h;
```
peppy: Getting the feeling this logic should be in its own class to keep `OsuPlayfield` from getting too polluted..

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -92,6 +128,67 @@ private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
             judgementLayer.Add(explosion);
         }
 
+        /// <summary>
+        /// Misses all <see cref="OsuHitObject"/>s occurring earlier than the start time of a judged <see cref="OsuHitObject"/>.
+        /// </summary>
+        /// <param name="result">The <see cref="JudgementResult"/> of the judged <see cref="OsuHitObject"/>.</param>
+        private void missAllEarlier(JudgementResult result)
```
peppy: any reason this is taking a `JudgementResult` and not just a `HitObject`?

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -64,21 +67,54 @@ public override void Add(DrawableHitObject h)
...
 
         public override bool Remove(DrawableHitObject h)
         {
             bool result = base.Remove(h);
 
             if (result)
-                followPoints.RemoveFollowPoints((DrawableOsuHitObject)h);
+            {
+                DrawableOsuHitObject osuHitObject = (DrawableOsuHitObject)h;
+                osuHitObject.CheckHittable = null;
```
peppy: why is this required?

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -92,6 +128,67 @@ private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
...
+                return;
+
+            if (!causesNoteLockMisses(obj.HitObject))
+                return;
+
+            osuObject.MissForcefully();
+        }
+
+        /// <summary>
+        /// Whether a <see cref="HitObject"/> is contributes to note lock.
```
peppy:
```suggestion
        /// Whether a <see cref="HitObject"/> contributes to note lock.
```

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -92,6 +128,67 @@ private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
...
+        /// Future contributing <see cref="HitObject"/>s will not be hittable until the start time of the last contributing <see cref="HitObject"/> is reached.
+        /// </summary>
+        /// <param name="hitObject">The <see cref="HitObject"/> to test.</param>
+        /// <returns>Whether <paramref name="hitObject"/> causes note lock.</returns>
+        private bool contributesToNoteLock(HitObject hitObject)
```
peppy: `canBlockFutureHits`? "Whether the provided hitObject will block further blocking hitObjects from being hit"

i don't like this "contributing" or "note lock" language because it's hard to understand.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -92,6 +128,67 @@ private void onNewResult(DrawableHitObject judgedObject, JudgementResult result)
...
+            => hitObject is HitCircle || hitObject is Slider;
```
peppy: curious why this is matching on `Slider` but the one below is matching on `SliderTailCircle`. logically i'd expect this to be checking against `SliderHeadCircle`, not `Slider` itself.

peppy (changes requested):

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
smoogipoo: Because `SliderHeadCircle` isn't a part of the hitobjects list, whereas it _is_ apart of the judgements. Maybe I should comment that?

peppy: it is in nested though right? maybe i misread the logic but it looked like you were also checking nested objects.

smoogipoo: I'm only checking nested for the miss loop which happens on judgement. This is a separate loop that happens every `CheckHittable`. I could check nested in both cases, but with the added complexity I'm not sure if it's really worth it... will think about it.

peppy (approved):

PR #8342, opened 2020-03-18
Handle beatmap import from a stable installation with a custom Songs path

Game4all:
Depends on 
- [ ] ppy/osu-framework#3384

# Summary

The songs directory location of an osu!stable installation can be modified through the osu! configuration file.
This PR adds the ability to the `StableStorage` to detect if the user has modified the Songs directory location by reading the osu! stable config file.

Changes have been made to the way `ArchiveModelManager`s access to stable data, they now get access only to the data directory they require (using `Storage.GetStorageForDirectory()`) 
instead of the whole installation directory since `NativeStorage` possess functionnality to prevent path traversal outside of the base directory (blocking import in case of a custom Songs directory location)

`StableStorage` overrides `GetStorageForDirectory()` to return a `Storage` pointing to the Songs folder if the asked folder is the Songs folder.

## Remarks

Implementing automated tests for this is quite hard, so I tested the behaviour myself by creating a directory containing a few beatmaps, modifying my `BeatmapDirectory` in my osu! stable config file to point to that directory and trying importing from lazer. It **works** on my machine.

### DIFF
--- a/osu.Desktop/OsuGameDesktop.cs
+++ b/osu.Desktop/OsuGameDesktop.cs
@@ -117,7 +117,7 @@ namespace osu.Desktop
         /// </summary>
         private class StableStorage : WindowsStorage
         {
-            private const string default_songs_path = "Songs";
+            private const string songs_path = "Songs";
 
             private readonly string songsPath;
             private bool usesCustomSongPath;
@@ -132,7 +132,7 @@ namespace osu.Desktop
 
             protected override string LocateBasePath()
             {
-                static bool checkExists(string p) => File.Exists(Path.Combine(p, "osu!.exe"));
+                static bool checkExists(string p) => Directory.EnumerateFiles(p, "osu!.*.cfg").Any();
 
                 string stableInstallPath;
 
@@ -159,6 +159,22 @@ namespace osu.Desktop
                 return null;
             }
 
+            public override Storage GetStorageForDirectory(string directory)
+            {
+                if (usesCustomSongPath && directory.Equals(songs_path, StringComparison.Ordinal))
+                    return new StableSongsStorage(Path.GetFullPath(songsPath), host);
+
+                return base.GetStorageForDirectory(directory);
+            }
+
+            public override bool ExistsDirectory(string directory)
+            {
+                if (usesCustomSongPath && directory.Equals(songs_path, StringComparison.Ordinal))
+                    return Directory.Exists(songsPath);
+
+                return base.ExistsDirectory(directory);
+            }
+
             /// <summary>
             /// osu!stable <c>Songs</c> folder can be moved to another location by changing the <c>BeatmapDirectory</c> setting in the stable configuration file.
             /// This locates the <c>Songs</c> folder location for this stable installation and sets flags if the song folder location has been customized.
@@ -175,9 +191,9 @@ namespace osu.Desktop
 
                     if (line?.StartsWith("BeatmapDirectory") == true)
                     {
-                        var directory = line.Split('=')[1].TrimStart();
+                        string directory = line.Split('=')[1].TrimStart();
 
-                        if (!directory.Equals(default_songs_path, StringComparison.Ordinal))
+                        if (!directory.Equals(songs_path, StringComparison.Ordinal))
                         {
                             usesCustomSongPath = true;
                             return Path.GetFullPath(directory);
@@ -187,23 +203,7 @@ namespace osu.Desktop
                     }
                 }
 
-                return default_songs_path;
-            }
-
-            public override Storage GetStorageForDirectory(string directory)
-            {
-                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && usesCustomSongPath)
-                    return new StableSongsStorage(Path.GetFullPath(songsPath), host);
-
-                return base.GetStorageForDirectory(directory);
-            }
-
-            public override bool ExistsDirectory(string directory)
-            {
-                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && usesCustomSongPath)
-                    return Directory.Exists(songsPath);
-
-                return base.ExistsDirectory(directory);
+                return songs_path;
             }
 
             private class StableSongsStorage : WindowsStorage
--- a/osu.Desktop/OsuGameDesktop.cs
+++ b/osu.Desktop/OsuGameDesktop.cs
@@ -119,9 +119,8 @@ namespace osu.Desktop
         {
             private const string default_songs_path = "Songs";
 
-            private string songsPath;
-            private bool usesCustomSongsPath;
-            private DesktopGameHost host;
+            private readonly string songsPath;
+            private readonly DesktopGameHost host;
 
             public StableStorage(DesktopGameHost host)
                 : base(string.Empty, host)
@@ -160,10 +159,9 @@ namespace osu.Desktop
             }
 
             /// <summary>
-            /// osu! stable Songs folder can be moved to another location by changing the 'BeatmapDirectory' setting in the stable configuration file.
-            /// This locates the Songs folder location for this stable instalation and sets flags if the songs folder location has been customized.
+            /// osu!stable <c>Songs</c> folder can be moved to another location by changing the <c>BeatmapDirectory</c> setting in the stable configuration file.
+            /// This locates the <c>Songs</c> folder location for this stable installation and sets flags if the song folder location has been customized.
             /// </summary>
-            /// <returns></returns>
             private string locateSongsDirectory()
             {
                 //we get the user config file.
@@ -174,24 +172,17 @@ namespace osu.Desktop
                 {
                     var line = textReader.ReadLine();
 
-                    if (line.StartsWith("BeatmapDirectory"))
-                    {
-                        var dir = line.Split('=')[1].TrimStart();
-                        if (!dir.Equals(default_songs_path, StringComparison.Ordinal))
-                        {
-                            usesCustomSongsPath = true;
-                            return Path.GetFullPath(dir);
-                        }
-                    }
+                    if (line?.StartsWith("BeatmapDirectory") == true)
+                        return line.Split('=')[1].TrimStart();
                 }
 
-                return null;
+                return default_songs_path;
             }
 
             public override Storage GetStorageForDirectory(string directory)
             {
-                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && usesCustomSongsPath)
-                    return new StableSongsStorage(songsPath, host);
+                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && !songsPath.Equals(default_songs_path, StringComparison.Ordinal))
+                    return new StableSongsStorage(Path.GetFullPath(songsPath), host);
 
                 return base.GetStorageForDirectory(directory);
             }
--- a/osu.Desktop/OsuGameDesktop.cs
+++ b/osu.Desktop/OsuGameDesktop.cs
@@ -117,6 +117,19 @@ namespace osu.Desktop
         /// </summary>
         private class StableStorage : WindowsStorage
         {
+            private const string default_songs_path = "Songs";
+
+            private string songsPath;
+            private bool usesCustomSongsPath;
+            private DesktopGameHost host;
+
+            public StableStorage(DesktopGameHost host)
+                : base(string.Empty, host)
+            {
+                this.host = host;
+                songsPath = locateSongsDirectory();
+            }
+
             protected override string LocateBasePath()
             {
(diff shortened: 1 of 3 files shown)


osu.Desktop/OsuGameDesktop.cs
```diff
@@ -117,6 +117,19 @@ private void fileDrop(object sender, FileDropEventArgs e)
...
+            public StableStorage(DesktopGameHost host)
+                : base(string.Empty, host)
+            {
+                this.host = host;
+                songsPath = locateSongsDirectory();
+            }
+
             protected override string LocateBasePath()
             {
                 static bool checkExists(string p) => Directory.Exists(Path.Combine(p, "Songs"));
```
TheWildTree: I know this is not your code, but this check seems very fragile - it might fail if someone deletes/renames the original `Songs` folder, which isn't that unlikely to happen if we're talking about changing the beatmap directory

imo we could just check for the `osu!.exe` there instead.

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -117,9 +117,22 @@ private void fileDrop(object sender, FileDropEventArgs e)
         /// </summary>
         private class StableStorage : WindowsStorage
         {
+            private const string default_songs_path = "Songs";
+
+            private string songsPath;
+            private bool usesCustomSongsPath;
```
bdach: Why do `songsPath` and `usesCustomSongsPath` need to both exist? Can't we just start off with `songsPath == default_songs_path`, then set value of `songsPath` to the custom one if it turns out that it is different?

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,50 @@ protected override string LocateBasePath()
                 return null;
             }
 
-            public StableStorage(DesktopGameHost host)
-                : base(string.Empty, host)
+            /// <summary>
+            /// osu! stable Songs folder can be moved to another location by changing the 'BeatmapDirectory' setting in the stable configuration file.
+            /// This locates the Songs folder location for this stable instalation and sets flags if the songs folder location has been customized.
+            /// </summary>
+            /// <returns></returns>
```
bdach:
```suggestion
            /// <summary>
            /// osu!stable <code>Songs</code> folder can be moved to another location by changing the <code>BeatmapDirectory</code> setting in the stable configuration file.
            /// This locates the <code>Songs</code> folder location for this stable installation and sets flags if the song folder location has been customized.
            /// </summary>
```

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,50 @@ protected override string LocateBasePath()
...
+                    var line = textReader.ReadLine();
+
+                    if (line.StartsWith("BeatmapDirectory"))
+                    {
+                        var dir = line.Split('=')[1].TrimStart();
+                        if (!dir.Equals(default_songs_path, StringComparison.Ordinal))
+                        {
+                            usesCustomSongsPath = true;
+                            return Path.GetFullPath(dir);
+                        }
```
bdach: The comparison seems unnecessary, can just assign.

Also you should break out of the loop after reading this line to avoid reading the entire file unnecessarily.

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,50 @@ protected override string LocateBasePath()
...
+                    }
+                }
+
+                return null;
+            }
+
+            public override Storage GetStorageForDirectory(string directory)
+            {
+                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && usesCustomSongsPath)
+                    return new StableSongsStorage(songsPath, host);
```
bdach: So I suppose this allocation is the only reason that `usesCustomSongsPath` exists (to not reallocate storage in the default case) - but maybe we can just have a `Lazy<StableSongsStorage>` and return that? Not sure about thread safety of operations though, would have to check to be 100% sure.

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -560,7 +560,7 @@ private List<TFileModel> createFileInfos(ArchiveReader reader, FileStore files)
         /// <summary>
         /// Select paths to import from stable. Default implementation iterates all directories in <see cref="ImportFromStablePath"/>.
         /// </summary>
-        protected virtual IEnumerable<string> GetStableImportPaths(Storage stableStoage) => stableStoage.GetDirectories(ImportFromStablePath);
+        protected virtual IEnumerable<string> GetStableImportPaths(Storage stableStoage) => stableStoage.GetDirectories(".");
```
bdach:
```suggestion
        protected virtual IEnumerable<string> GetStableImportPaths(Storage stableStorage) => stableStorage.GetDirectories(".");
```

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -589,7 +589,9 @@ public Task ImportFromStableAsync()
                 return Task.CompletedTask;
             }
 
-            return Task.Run(async () => await Import(GetStableImportPaths(GetStableStorage()).Select(f => stable.GetFullPath(f)).ToArray()));
+            var directory = stable.GetStorageForDirectory(ImportFromStablePath);
```
bdach: It's really confusing that many other usages have started using `.` as the directory but this hasn't...?

bdach (changes requested):

osu.Game/Database/ArchiveModelManager.cs
Game4all: `ArchiveModelManager` now access only to a `Storage` to their `ImportFromStablePath` directory to import data from stable instead of a storage at the root of the stable installation
hence the usage of "." for enumerating directories in `GetStableImportPaths`

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -560,7 +560,7 @@ private List<TFileModel> createFileInfos(ArchiveReader reader, FileStore files)
         /// <summary>
         /// Select paths to import from stable. Default implementation iterates all directories in <see cref="ImportFromStablePath"/>.
```
iiSaLMaN: this should be changed now, right?

osu.Desktop/OsuGameDesktop.cs
iiSaLMaN: sounds like this needs `ThreadLocal<StableSongsStorabe>` instead?

Game4all: Why use a `Lazy` here though? This code path is reached once (when `ArchiveModelManager.ImportFromStableAsync()` is called), so I guess it should be fine to allocate directly the Storage instead of wrapping it in a Lazy.

I see your point in using a `Lazy` to cache the instance if this exact code path was to execute several times, but can we consider it really necessary  (what are the chances that the user will re-import his beatmaps two times in a row)

Unless i'm missing something?

bdach: I was just mostly confused at why that was written in this way and figured that might have been the reason. Would have to look at how heavy the allocation cost is to say if this is worth the trouble or not.

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -117,9 +117,22 @@ private void fileDrop(object sender, FileDropEventArgs e)
         /// </summary>
         private class StableStorage : WindowsStorage
         {
+            private const string default_songs_path = "Songs";
```
peppy: `default_` is wrong/unnecessary

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -117,9 +117,22 @@ private void fileDrop(object sender, FileDropEventArgs e)
...
+                : base(string.Empty, host)
+            {
+                this.host = host;
+                songsPath = locateSongsDirectory();
+            }
+
             protected override string LocateBasePath()
             {
-                static bool checkExists(string p) => Directory.Exists(Path.Combine(p, "Songs"));
+                static bool checkExists(string p) => File.Exists(Path.Combine(p, "osu!.exe"));
```
peppy: This is a regression. osu! does not always locate its executable at the data path.

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,60 @@ protected override string LocateBasePath()
                 return null;
             }
 
-            public StableStorage(DesktopGameHost host)
-                : base(string.Empty, host)
+            /// <summary>
+            /// osu!stable <c>Songs</c> folder can be moved to another location by changing the <c>BeatmapDirectory</c> setting in the stable configuration file.
+            /// This locates the <c>Songs</c> folder location for this stable installation and sets flags if the song folder location has been customized.
+            /// </summary>
+            private string locateSongsDirectory()
```
peppy: private functions go below public

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,60 @@ protected override string LocateBasePath()
...
+                var configFile = GetStream(GetFiles(".", "osu!.*.cfg").First());
+                var textReader = new StreamReader(configFile);
+
+                while (!textReader.EndOfStream)
+                {
+                    var line = textReader.ReadLine();
+
+                    if (line?.StartsWith("BeatmapDirectory") == true)
+                    {
+                        var directory = line.Split('=')[1].TrimStart();
```
peppy:
```suggestion
                        string directory = line.Split('=')[1].TrimStart();
```

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -146,9 +159,60 @@ protected override string LocateBasePath()
...
+                        break;
+                    }
+                }
+
+                return default_songs_path;
+            }
+
+            public override Storage GetStorageForDirectory(string directory)
+            {
+                if (directory.Equals(default_songs_path, StringComparison.Ordinal) && usesCustomSongPath)
```
peppy:
```suggestion
                if (usesCustomSongPath && directory.Equals(default_songs_path, StringComparison.Ordinal))
```

put bool condition first

osu.Game/Scoring/ScoreManager.cs
```diff
@@ -57,7 +57,7 @@ protected override ScoreInfo CreateModel(ArchiveReader archive)
         }
 
         protected override IEnumerable<string> GetStableImportPaths(Storage stableStorage)
-            => stableStorage.GetFiles(ImportFromStablePath).Where(p => HandledExtensions.Any(ext => Path.GetExtension(p)?.Equals(ext, StringComparison.OrdinalIgnoreCase) ?? false));
+            => stableStorage.GetFiles(".").Where(p => HandledExtensions.Any(ext => Path.GetExtension(p)?.Equals(ext, StringComparison.OrdinalIgnoreCase) ?? false));
```
peppy: this looks like a regression?

peppy (changes requested):

osu.Game/Scoring/ScoreManager.cs
Game4all: Actually it isn't.
`ArchiveModelManager` now operates on a storage pointing directly to the target data directory (`ImportFromStablePath`) instead of a Storage pointing to the root of the osu! stable installation
thus the use of "." to enumerate all directories / files in the current directory.

See this line:
	https://github.com/Game4all/osu/blob/a0fb308e8858265654c44e37b9ee3ba22b26c8e8/osu.Game/Database/ArchiveModelManager.cs#L592

osu.Desktop/OsuGameDesktop.cs
Game4all: ok alright. but checking for the Songs directory existence would fail in the case where the user changed his songs folder location and deletes the original one, right ?. So checking the existence of the stable installation by checking for the Songs directory presence is unreliable unless that directory is being recreated by the osu! stable client on every launch if it doesn't exist.  (in which case the old check can be restored)

TheWildTree: if you change the `BeatmapDirectory` to something else, it doesn't recreate the default folder

Game4all: In that case we should check for existence of files such as the user configuration file instead.

osu.Game/Scoring/ScoreManager.cs
peppy: check the code again. test it. it's a regression.

peppy:
not a huge fan of this method. why not just make a StableSongStorage and StableStorage.GetSongStorage to avoid all the hassles you're running in to? would remove the hardcoded string, framework changes and archivemodelmanager changes.

Game4all:
Closing this for now. Will reopen once changes will have been applied in the proposed direction.

PR #8343, opened 2020-03-18
Decouple user model from API structures

EVAST9919:


peppy:
Not sure the attributes of `APIUser` should be private (no real reason to make them such).

EVAST9919:
Otherwise one can use APIUser without using ToUser at all and operate on this one. 
Is this okay?

EVAST9919:
For example we could just use `APIUser` instead of `User` in `SocialOverlay`, which is not correct as far as I can tell. But probably I'm wrong.

EVAST9919:
Well, private is really an overkill. At least we want to leave them protected for easy inheritance.

peppy:
public is fine. using them as-is is fine if that's all you need. see `BeatmapInfo` vs `APIBeatmap` etc.

we convert to non-API classes when we need to store to the local database, and potentially display from *both* sources. if you're making a component which is exclusively API/online (be very careful before saying "yes" to this) then you can indeed just use `APIUser`.

osu.Game/Online/API/Requests/Responses/APIUser.cs
```diff
@@ -1,14 +1,217 @@
...
+            PlayStyles = PlayStyles,
+            PlayMode = PlayMode,
+            ProfileOrder = ProfileOrder,
+            Kudosu = Kudosu,
+            Statistics = Statistics,
+            Badges = Badges,
+            Achievements = Achievements,
+            MonthlyPlaycounts = MonthlyPlaycounts,
+            ReplaysWatchedCounts = ReplaysWatchedCounts
+        };
```
bdach: So this almost 50-line copy constructor shows the whole problem I have with this `User` stuff in general.

Why is *all of it* here? Now I realise setting this right is going to be a toothache and a half, but it's just... Almost *fifty* fields. I guarantee you some usages won't have 80% of this filled out. This is like a C WinAPI structure that you're supposed to fill half of and leave the rest blank for some unknowable reason.

IMO it needs to go. And go *properly*, not by adding another 50-property `APIUser`.

bdach (changes requested):

EVAST9919:
Closing due to the reasons described above

peppy:
Definitely needs a bit more thought. I'm not strictly against the direction of this PR, but soon after the `User` class would need to be a cut-down implementation (only things we wanted to store to the database).

Proabably best to wait until we start storing users to the local database and make the change at that point.

PR #8346, opened 2020-03-18
Reduce allocations of followpoints by reusing existing

peppy:
I'm open to alternative fixes to follow point logic, but feel we probably want to hold off on any drastic changes until we reconsider skinning and pooling. Consider this a hotfix until we get to a point we can look into that.

Before:
image

After:
image

Reduces [centipede](https://osu.ppy.sh/beatmapsets/150945#osu/372245) allocation overhead during load by 4.8gb


peppy:
May be a good idea to further investigate the cause of follow points being so aggressively refreshed during load. It does not happen during normal beatmaps so is not high priority. This fix only addresses the huge memory overhead – it does not address the remaining 2m+ object allocation patterns seen in the "After" screenshot.

smoogipoo (approved):

PR #8351, opened 2020-03-19
Show customised mod setting values in tooltip

### LINKED ISSUE #8159: Show mod settings in scores page
**Describe the new feature:** 
Being able to see what settings for the difficulty adjust mod in past scores

Tool tip showing what difficulty settings for the mod when hovering over the mod icon in scores page, currently it only says "difficulty adjust"

Liam-DeVoe:
Resolves #8159 

Will only show the setting if it's changed from the default. `DoubleTime (Speed Increase 1.42)` didn't look great so I opted for `DoubleTime (1.42x)` instead. Similarly for the other `ModRateAdjust` mods.

Screen Shot 2020-03-18 at 11 48 03 PM

Wasn't sure how to add tests for this, point me in the right direction and I'll do it, if necessary.


osu.Game.Rulesets.Catch/Mods/CatchModDifficultyAdjust.cs
```diff
@@ -30,6 +30,11 @@ public class CatchModDifficultyAdjust : ModDifficultyAdjust
             Value = 5,
         };
 
+        public override string IconTooltip => ($"{Name} ({(CircleSize.IsDefault ? "" : $"CS {CircleSize.Value}, ")}" +
+                                       $"{(DrainRate.IsDefault ? "" : $"HP {DrainRate.Value}, ")}" +
+                                       $"{(OverallDifficulty.IsDefault ? "" : $"OD {OverallDifficulty.Value}, ")}" +
+                                       $"{(ApproachRate.IsDefault ? "" : $"AR {ApproachRate.Value}")}").TrimEnd(new char[] { ',', ' ' }) + ")";
```
bdach: `string.Join` is better than `TrimEnd` for this sort of thing.

bdach (commented):
I'm not sure I'm 100% on board with this approach of having to manually set the text for every mod that has settings. Ideally either the tooltip text should autogenerate based on all setting values defined for the mod, or the manual specification of text should be mandatory (via abstract property or otherwise) to make sure no future mod with settings forgets to do so.

Oh, as for the tests, I'd let it slide for this one. Testing tooltip text seems extravagant even for me.

tybug:
> Ideally either the tooltip text should autogenerate based on all setting values defined for the mod

I tried that approach first with `mod.GetOrderedSettingsSourceProperties` but that only gives you the label of the setting ("Circle Size", "Approach Rate", etc), and converting those to acronyms seems messier, especially when you want some custom configuration, like `1.42x` instead of `Speed Increase 1.42` for DT.

Maybe using `GetOrderedSettingsSourceProperties` as the default implementation for `public virtual string IconTooltip` in `Mod` would be good, as an alternative to making it abstract?

bdach:
> I tried that approach first with `mod.GetOrderedSettingsSourceProperties` but that only gives you the label of the setting ("Circle Size", "Approach Rate", etc), and converting those to acronyms seems messier, especially when you want some custom configuration, like `1.42x` instead of `Speed Increase 1.42` for DT.

Maybe extending `SettingSourceAttribute` to allow custom formatting is the way then? Something like
```csharp
[SettingSource(
    "Circle Size",
    "Override a beatmap's set CS.",
    FIRST_SETTING_ORDER - 1,
    TooltipText = "CS: {0}"
)]
```
`Description` could then be used in conjunction with `string.Format` to do the value substitution. Mind you `SettingSource` would be getting a little big at this stage, so I'm not 100% convinced of this direction.

tybug:
That seems like an elegant solution to me. I'll implement it that way for now

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -42,6 +47,44 @@ public abstract class Mod : IMod, IJsonSerializable
...
+        {
+            get
+            {
+                List<string> attributes = new List<string>();
+                foreach ((SettingSourceAttribute attr, System.Reflection.PropertyInfo property) in this.GetOrderedSettingsSourceProperties())
+                {
+                    // use TooltipText from SettingSource if available, but fall back to Label, which has to be provided
+                    string tooltipText = attr.TooltipText ?? attr.Label + " {0}";
+                    object bindableObj = property.GetValue(this);
+                    if (bindableObj is BindableInt bindableInt && !bindableInt.IsDefault)
```
tybug: couldn't find a way to cast to a generic bindable, am I missing something? Could cast to `IBindable` but not `IBindable<T>`, the latter of which has the `Value` method.

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -42,6 +47,48 @@ public abstract class Mod : IMod, IJsonSerializable
...
+        /// <remarks>
+        /// Differs from <see cref="Name"/>, as the value of attributes (AR, CS, etc) changeable via the mod
+        /// are displayed in the tooltip.
+        /// </remarks>
+        [JsonIgnore]
+        public virtual string IconTooltip
+        {
+            get
+            {
+                List<string> attributes = new List<string>();
```
bdach: Cases like this I'd 100% of the time use `var` to not make the code Java-level redundant.
```suggestion
                var attributes = new List<string>();
```

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -42,6 +47,48 @@ public abstract class Mod : IMod, IJsonSerializable
...
+
+                foreach ((SettingSourceAttribute attr, System.Reflection.PropertyInfo property) in this.GetOrderedSettingsSourceProperties())
```
bdach: Would prefer having a `using` statement for `System.Reflection` instead of this fully qualified name for `PropertyInfo`.

osu.Game/Rulesets/Mods/Mod.cs
bdach: You can't cast to a generic type without knowing the actual type parameters. That said, I don't think casting to `Bindable<T>` is necessary, as [`Bindable<T>`'s default implementation for `ToString()` returns the value](https://github.com/ppy/osu-framework/blob/7b8a966bbddc5fc09aec38bfe9df6e7440409a9a/osu.Framework/Bindables/Bindable.cs#L377). You can check if the value is default without knowing the type param by casting to [`IHasDefaultValue`](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Bindables/IHasDefaultValue.cs) instead.

So in general you'd probably want to do something like

```csharp
var settingIsDefault = (bindableObj as IHasDefaultValue)?.IsDefault;
var settingValueText = settingIsDefault == true ? string.Empty : bindableObj.ToString();
```
Note that the `== true` comparison is not redundant in this case, as the null-coalescing operator makes `settingIsDefault` be a tri-state `bool?`. This get rid of all the ugly type checks.

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -42,6 +47,48 @@ public abstract class Mod : IMod, IJsonSerializable
...
+                        continue;
+                    }
+
+                    if (bindableObj is BindableDouble bindableDouble && !bindableDouble.IsDefault)
+                    {
+                        attributes.Add(string.Format(tooltipText, bindableDouble.Value));
+                    }
+                }
+
+                return $"{Name}{(attributes.Any() ? $" ({string.Join(", ", attributes)})" : "")}";
```
bdach: To be honest I find nested ternaries quite unreadable. This should be split.

osu.Game/Rulesets/Mods/ModEasy.cs
```diff
@@ -21,7 +21,7 @@ public abstract class ModEasy : Mod, IApplicableToDifficulty, IApplicableFailOve
         public override bool Ranked => true;
         public override Type[] IncompatibleMods => new[] { typeof(ModHardRock), typeof(ModDifficultyAdjust) };
 
-        [SettingSource("Extra Lives", "Number of extra lives")]
+        [SettingSource("Extra Lives", "Number of extra lives", "{0} lives")]
```
bdach: This one is actually tricky, as it's incorrect for a value of 1... You probably want to actually override here and use `Humanizer.ToQuantity` to pluralise properly.

osu.Game/Rulesets/Mods/ModTimeRamp.cs
```diff
@@ -26,6 +26,8 @@ public abstract class ModTimeRamp : Mod, IUpdatableByPlayfield, IApplicableToBea
         [SettingSource("Final rate", "The final speed to ramp to")]
         public abstract BindableNumber<double> FinalRate { get; }
 
+        public override string IconTooltip => $"{Name} ({InitialRate.Value}x to {FinalRate.Value}x)";
```
bdach: I feel like the mod name and the brackets shouldn't be included in this override to make sure display is standardised.

bdach (changes requested):

bdach:
To be frank now that I see half of usages be overrides and the other half be attributes I'd probably think about reverting to the old version. This feels like doing things in two ways while there could be just one.

I *would* standardise the overall tooltip text to be of the form
```
{mod name} ({settings go here})
```
but make the `settings go here` part (and only that) overridable.

tybug:
I think you're right about that. There's enough variation in the display that setting the format through `SettingSource` doesn't quite make sense. Will revert to overridable properties

tybug:
> The reflection stuff can act as a baseline but should use the mod setting name and just that IMO.

Maybe I'm misreading, but do you want a default tooltip of the form `Difficulty Adjust (Drain Rate, Overall Difficulty,)` then? Obviously it would be overridden by `ModDifficultyAdjust`, but without the values of the setting the tooltip doesn't make much sense. Would be better to be just the mod name.

(as of right now I removed all the reflection code)

tybug:
Just realized you probably meant setting name + value, lol. yeah that's a better default than none at all

osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
```diff
@@ -30,6 +31,22 @@ public class OsuModDifficultyAdjust : ModDifficultyAdjust
             Value = 5,
         };
 
+        public override string SettingDescription
+        {
+            get
+            {
+                string circleSize = CircleSize.IsDefault ? "" : $"CS {CircleSize.Value}";
+                string drainRate = DrainRate.IsDefault ? "" : $"HP {DrainRate.Value}";
+                string overallDifficulty = OverallDifficulty.IsDefault ? "" : $"OD {OverallDifficulty.Value}";
```
peppy: replace all `""` with `string.Empty`

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -52,6 +52,20 @@ public abstract class ModDifficultyAdjust : Mod, IApplicableToDifficulty
...
+        {
+            get
+            {
+                string drainRate = DrainRate.IsDefault ? "" : $"HP {DrainRate.Value}";
+                string overallDifficulty = OverallDifficulty.IsDefault ? "" : $"OD {OverallDifficulty.Value}";
+
+                string[] settings = { drainRate, overallDifficulty };
+                // filter out empty strings so we don't have orphaned commas
+                settings = Array.FindAll(settings, s => !string.IsNullOrEmpty(s));
+                return string.Join(", ", settings);
```
peppy:
```suggestion
                 return string.Join(", ", new[]
                {
                    drainRate,
                    overallDifficulty
                }.Where(s => !string.IsNullOrEmpty(s)));
```

i think this reads better.

osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
```diff
@@ -30,6 +31,22 @@ public class OsuModDifficultyAdjust : ModDifficultyAdjust
...
+                string approachRate = ApproachRate.IsDefault ? "" : $"AR {ApproachRate.Value}";
+
+                string[] settings = { circleSize, drainRate, overallDifficulty, approachRate };
```
peppy: repeated implementation for `drainRate` / `overallDifficulty`? why not use `base.` for those?

peppy (changes requested):

peppy:
Need better number formatting:

IMG_0159

tybug:
> Need better number formatting:

I meant to bring that up - I noticed it but didn't want to impose a precision on the bindable value when that should be determined by the bindable itself. This only happens on `x.1` and `x.6` values for the settings, and the true value of the bindable is actually `3.10000014`:
Screen Shot 2020-03-22 at 6 17 21 PM

The bindables are already instantiated with `0.1f` precision. Is it a bug for the setting to then have a value with precision beyond that?

iiSaLMaN:
I believe you should just cap the amount of decimal places, you're dealing with float values here.

tybug:
Maybe I don't understand `BindableNumber` and floating point numbers in general well enough, but `ModDifficultyAdjust` (eg) has a `BindableNumber<float> DrainRate` member:

```csharp
        [SettingSource("HP Drain", "Override a beatmap's set HP.", FIRST_SETTING_ORDER)]
        public BindableNumber<float> DrainRate { get; } = new BindableFloat
        {
            Precision = 0.1f,
            MinValue = 1,
            MaxValue = 10,
            Default = 5,
            Value = 5,
        };
```

It seems like a bug to me that a `BindableNumber` with a precision of `0.1f` could ever have a value not in increments of `0.1`

Not trying to waste anyones time, just want to get across what I'm saying

bdach:
Floating point math in general is funny in that you simply *cannot* represent `0.1` as a base-2 floating-point decimal number accurately. Generally what can sometimes happen (I can't say for sure if this is the case here without investigating, but it looks likely) is that if you keep adding `0.1f` to a number then after enough additions the error of representation of `0.1f` will add up to a non-negligble over/underflow that will start showing up in `ToString()` and other places.

(You might then be wondering, why calling `ToString()` on `float`s and `double`s *will* sometimes just return `0.1` - the reason is that C# will silently lie to you and perform rounding so you don't get the ugly fraction. But *only* if it deems the number to be close enough to the nearest possible representation of the decimal fraction, which might no longer be true after the error builds up significantly enough.)

Fixing this properly is a non-negligible amount of work (probably would need to actually use `decimal` which can handle base-10 fractions properly), so I'd just say to call it a day and enforce precision manually.

osu.Game.Rulesets.Osu/Mods/OsuModDifficultyAdjust.cs
tybug: no reason other than I forgot to. fixed

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
```diff
@@ -56,13 +57,14 @@ public override string SettingDescription
...
             {
-                string drainRate = DrainRate.IsDefault ? "" : $"HP {DrainRate.Value}";
-                string overallDifficulty = OverallDifficulty.IsDefault ? "" : $"OD {OverallDifficulty.Value}";
-
-                string[] settings = { drainRate, overallDifficulty };
-                // filter out empty strings so we don't have orphaned commas
-                settings = Array.FindAll(settings, s => !string.IsNullOrEmpty(s));
-                return string.Join(", ", settings);
+                string drainRate = DrainRate.IsDefault ? string.Empty : $"HP {DrainRate.Value:0.#}";
+                string overallDifficulty = OverallDifficulty.IsDefault ? string.Empty : $"OD {OverallDifficulty.Value:0.#}";
```
peppy: you probably want to use `N1` instead of `0.#`, for better localisation support

https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings

tybug:
> You might then be wondering, why calling ToString() on floats and doubles will sometimes just return 0.1

I was in fact wondering just that! Thank you for the explanation. I enforced precision to one decimal place (two for `ModRateAdjust`, though I chose to display `1.4x` instead of `1.40x`. Wasn't sure which was preferred)

peppy:
For DP count, try to match existing displays whenever possible.

osu.Game/Rulesets/Mods/ModDifficultyAdjust.cs
tybug: `N1` doesn't remove the trailing zero, is a tooltip like `Difficulty Adjust (CS 3.0) okay? The "details" tab for beatmaps doesn't show trailing zeroes either

iiSaLMaN: definitely

peppy: matching the same formatting as the details display is fine.

tybug:
realized that the mod icon at the bottom of the screen (under "mods") doesn't update its tooltip when you change settings, will have to implement that as well

peppy:
It updates now. Was a simple change/oversight in your code (see https://github.com/ppy/osu/pull/8351/commits/a6b153673e2fa61dc5c7c674a5fb6b964008063d)

peppy:
Make a few more tidy-up changes, looks good otherwise 👍 .

peppy (dismissed):

tybug:
Forgot a minor thing for `ModTimeRamp` 😅

peppy (approved):

PR #8355, opened 2020-03-19
Fix potentially invalid push in player while already exiting

### LINKED ISSUE #8352: Crash: "Cannot perform Push on a non-current screen."
**Describe the crash:** 
Game crashed after finishing map. I think I pressed esc so maybe it caused some invalid state?

None

**osu!lazer version:** 
local dev version

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/4352633/runtime.log)

**Computer Specifications:** 
See logs

peppy:
- Closes https://github.com/ppy/osu/issues/8352.

Would hope we can figure a more streamlined path of ensuring this kind of consistency, but in discussion all the possible solutions we have available are not what we want.

No test because it's a very hard one to test (requires specific score import timing which will take way too much effort to test). If it regresses another time I'll put the effort in to add proper test coverage (would need a custom `ScoreManager` and multiple mutexes).

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -387,6 +387,10 @@ namespace osu.Game.Screens.Play
 
         private void onCompletion()
         {
+            // screen may be in the exiting transition phase.
+            if (!this.IsCurrentScreen())
+                return;
+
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
Looks pretty sane to me

PR #8356, opened 2020-03-19
Fix slider ticks/repeats contributing too much to accuracy

### LINKED ISSUE #3300: The way judgements work for sliders isn't very good.
So there is a couple of issues here, so maybe this should be more than one issue, but I figured they are related enough to be under the same one.

My first problem is with slider tails. They give judgements, but these judgements don't affect combo, accuracy, or score (the only reason I can see for them to even exist is to affect the slider completion judgements). The problem I have with this is that it can lead to result screens like this one:
lazerscorescreenerror If I apply the accuracy formula to the the judgement results shown I get something different than the accuracy I achieved because there are extra greats and misses from slider tails that don't affect accuracy. This makes it appear like I got more misses, greats, or both than my accuracy actually reflects. Further, I FCed despite having 2 misses shown, which should not be able to happen. To resolve this you could simply get rid of slider end judgments, or you could make it so they don't count toward the numbers displayed on the results screen.

My second problem is that sliders are incredibly overweighted in terms of accuracy, and their accuracy is not properly tied to player skill. This is caused by the fact that not only does the slider head have it's own judgement, but so does the slider repeat, each slider tick, and the slider as a whole (the tail also has a judgement, but this doesn't affect accuracy as mentioned above). This means that every slider gives out at least 2 judgements affecting accuracy, and sometimes many, many more, which overweights it. So in a map that is 50% sliders and 50% circles, at least 67% of your accuracy will come from your performance on the sliders, and if there are hold sliders or long sliders this could easily be as high as 75% or maybe even 80%. This behavior is abusable as well because an increased slider tick rate will result in much increased accuracy for exactly the same performance. Further, kick sliders with many repeats could be used to greatly increase accuracy. For example, imagine I made a jump map with 200 circles and 100 sliders, where 10 of these sliders are hold sliders with 7 repeats each and 10 others are low velocity sliders with 45 ticks (I checked an actual map and this is reasonable at the max tick rate). In this theoretical map, there are 300 clickable objects, 100 slider completion scores, and 520 slider components, that all equally affect accuracy. This means more than _**half**_ of the accuracy on this map would come from easy to hit, unclicked objects that never give partial accuracy. If a player got 90% accuracy from the clickable objects and slider completions, but hit all of the easy to hit slider ticks and repeats, they would end up with 95.65% accuracy, which is a very significant boost. Maybe this scenario is exaggerated (though I tried to be within the bounds of reason), but similar scenarios could certainly exist to a lesser extent and mappers could strategically utilize this to inflate accuracy on farm maps, if they wanted to do so.

The way I see it only clickable objects should have an effect on accuracy, as they are the only objects that actually require any timing; however, I understand your desire to account for accuracy in following a slider (otherwise, what's the point of sliders), but this honestly doesn't seem like a good way of doing it to me. The way in stable was ok, but I understand why you might want to move to a system that holds the player more accountable for their ability to follow sliders. This a bit of a hard problem to solve without overweighting sliders, but I have a couple of ideas. The first is to get rid of slider ticks and repeats having judgements and to divide the judgements for the head and slider completion in half (that way they add to 300). The problem with this is it isn't the easiest to reconcile with the current 300, 100, 50, 0 judgement system. My second suggestion is to get rid of all of the judgements that come from the slider except the completion judgement, while modifying it's formula. Perhaps to this:
```
if (hitFraction == 1 && HeadCircle.Result.Type == HitResult.Great)  
    r.Type = HitResult.Great;
else if (hitFraction >= 0.5 && HeadCircle.Result.Type >= HitResult.Good)
    r.Type = HitResult.Good;
else if (HeadCircle.Result.Type >= HitResult.Meh)
    r.Type = HitResult.Meh;
else
    r.Type = HitResult.Miss;
```
IDK if either of these solutions would be good, but I'm sure there are many other possible solutions to the issue. At the very least slider ticks and repeats should not have a direct affect on accuracy, or at least much less weighting than clickable objects.

This got a little long & ranty, and it is maybe not the best organized. Sorry. I hope I made some sense.

peppy:
In stable, slider ticks/repeats did not contribute to accuracy. This brings them in line with that logic and also reduces the base score they give in line with stable.

- Closes https://github.com/ppy/osu/issues/3300 (mostly, can be reopened for further issues)
- Renames `RepeatPoint` to `SliderRepeat`

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -52,6 +52,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.315.0" />
-    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.317.0" />
+    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.319.0" />
   </ItemGroup>
 </Project>
--- a/osu.Game/osu.Game.csproj
+++ b/osu.Game/osu.Game.csproj
@@ -23,7 +23,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="2.2.6" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.315.0" />
-    <PackageReference Include="ppy.osu.Framework" Version="2020.317.0" />
+    <PackageReference Include="ppy.osu.Framework" Version="2020.319.0" />
     <PackageReference Include="Sentry" Version="2.1.0" />
     <PackageReference Include="SharpCompress" Version="0.24.0" />
     <PackageReference Include="NUnit" Version="3.12.0" />
--- a/osu.iOS.props
+++ b/osu.iOS.props
@@ -71,7 +71,7 @@
   </ItemGroup>
   <ItemGroup Label="Package References">
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.315.0" />
-    <PackageReference Include="ppy.osu.Framework.iOS" Version="2020.317.0" />
+    <PackageReference Include="ppy.osu.Framework.iOS" Version="2020.319.0" />
   </ItemGroup>
   <!-- Xamarin.iOS does not automatically handle transitive dependencies from NuGet packages. -->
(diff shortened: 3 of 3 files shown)


smoogipoo (dismissed):

peppy:
Somehow broke this, reopened as #8362.

PR #8358, opened 2020-03-19
Remove slider implicit judgement

peppy:
Removes judgement from slider. Due to lazer having slider head accuracy, sliders would actually contribute double the accuracy/baseScore. This reduces them to only contributing once (based off accuracy on head). Tails now give the same as a repeat, in bonus score.

Discussion welcomed (but we'll probably push this out and wait for people to complain or rejoice because it needs some time to play, adjust and reflect).

- [x] Depends on #8362.
- Closes remaining portion of #3300.


peppy:
Comparing old vs new vs scorev2 (stable):

lazer-slider-changes

huoyaoyuan:
Unrelated questions: 
- Where can I find complete documentation of scoreV2(stable)? I can't find it in wiki. Forum search result are discussions, not conclusions.
- Where is scoring system of lazer discussed?

peppy:
Documentation is above for lazer. all a work in progress

smoogipoo (approved):

PR #8368, opened 2020-03-20
Fix carousel not returning to previous selection after filter

### LINKED ISSUE #2638: Filtering beatmaps after null selection

peppy:
Alternative to and closes https://github.com/ppy/osu/pull/8251.

- Closes #2638.

This logic already existed but wasn't being executed due to `LastSelected` getting incorrectly set to null. Basically a one line change (although I had to make one more change to keep correct test behaviour).

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -333,8 +333,7 @@ namespace osu.Game.Screens.Select
             else
                 set = visibleSets.ElementAt(RNG.Next(visibleSets.Count));
 
-            var visibleBeatmaps = set.Beatmaps.Where(s => !s.Filtered.Value).ToList();
-            select(visibleBeatmaps[RNG.Next(visibleBeatmaps.Count)]);
+            select(set);
             return true;
         }
 
@@ -756,7 +755,7 @@ namespace osu.Game.Screens.Select
 
             protected override void PerformSelection()
             {
-                if (LastSelected == null)
+                if (LastSelected == null || LastSelected.Filtered.Value)
                     carousel.SelectNextRandom();
                 else
                     base.PerformSelection();
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -419,7 +419,7 @@ namespace osu.Game.Tests.Visual.SongSelect
         }
 
         [Test]
-        public void TestCarouselRootIsRandom()
+        public void TestCarouselRemembersSelection()
         {
             List<BeatmapSetInfo> manySets = new List<BeatmapSetInfo>();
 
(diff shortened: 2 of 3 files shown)


LittleEndu:
If this also addresses #8239

peppy:
This does not address #8239. Only #2638 I guess.

peppy:
Actually it looks to solve both.

peppy:
No, it definitely doesn't. #8239 requires a different solution.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -756,7 +755,7 @@ public CarouselRoot(BeatmapCarousel carousel)
 
             protected override void PerformSelection()
             {
-                if (LastSelected == null)
+                if (LastSelected == null || LastSelected.Filtered.Value)
```
LittleEndu: I would honestly consider carousel selecting the next beatmap when refining your filter a feature and wouldn't change it. I can see a similar scenario to what I laid out in #2638 happening if you refine your current filter while your currently selected map would get filtered out by your new filter.

peppy: I don't follow.

LittleEndu: No. The selection happens in SelectNextRandom or in base.PerformSelection. This change (on this line at least) is making carousel select a random beatmap when there hasn't been a selection or when the last selection is filtered.

peppy: Correct. That's the intention.

LittleEndu: I did, that's #2638. The change on this line is making the carousel select something away from beatmaps near previous selection and instead select a random one. "Fix carousel not returning to previous selection after filter" is literally the title of the PR, I fail to see how this is out of scope.

peppy: Please wait until this fix is merged and then rebase your proposed change. I think you are misunderstanding this PR (please see the provided tests).

bdach: So I think @LittleEndu has an issue with behaviour tested in `RandomFallbackOnNonMatchingPrevious` - the reasoning is that when the previously-selected beatmap is filtered, the carousel should make a "local" change (as in choose from maps in proximity to the filtered ones) instead of making a random one. Whether that's the behaviour we want in the end is disputable and sort of up to preference, so I can't say anything objective about this. Subjectively I don't really care either way myself.

bdach:
Don't have any issue with the code here, or any strong opinion on the discussion about desired behaviour.

osu.Game/Screens/Select/BeatmapCarousel.cs
peppy: Yeah, I think I get that. Not sure it's correct though, and the case I mention in his new PR (where do you draw a line) is probably what we want to consider when deciding a path forward. ie. if 30000 beatmaps are hidden from a filter, do you just select the first remaining one? seems potentially worse than random. why even have random if it's only going to select once ever?

i dunno.

smoogipoo (approved):

PR #8370, opened 2020-03-20
Fix selection not occurring when switching away from empty ruleset

### LINKED ISSUE #8239: Song select doesn't autoselect after entering on no-beatmap ruleset

peppy:
- Closes #8239.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -751,13 +751,17 @@ namespace osu.Game.Screens.Select
 
             public CarouselRoot(BeatmapCarousel carousel)
             {
+                // root should always remaing selected. if not, PerformSelection will not be called.
+                State.Value = CarouselItemState.Selected;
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -751,13 +751,17 @@ private class CarouselRoot : CarouselGroupEagerSelect
 
             public CarouselRoot(BeatmapCarousel carousel)
             {
+                // root should always remaing selected. if not, PerformSelection will not be called.
```
```suggestion
                // root should always remain selected. if not, PerformSelection will not be called.
```

bdach (commented):
Aside from an annoying typo suggestion looks fine. Song select filtering pulls are tough to review as usual but I think I see how this works and why it's not done in this particular manner.

smoogipoo (dismissed):

PR #8374, opened 2020-03-20
Don't open profile if it's Autoplay

### LINKED ISSUE #8372: Clicking on player icon in result panel with autoplay shows blank user panel
Clicking on player icon in result panel with autoplay shows blank user panel

Screenshot_1
Screenshot_2

osu!lazer version: 2020.320.1

repricable on any song

uzervlad:
Closes #8372

### DIFF
--- a/osu.Game/Users/Drawables/DrawableAvatar.cs
+++ b/osu.Game/Users/Drawables/DrawableAvatar.cs
@@ -68,7 +68,7 @@ namespace osu.Game.Users.Drawables
             if (!OpenOnClick.Value)
                 return;
 
-            if (user != null && user.Id != 1)
+            if (user?.Id > 1)
                 game?.ShowUser(user.Id);
         }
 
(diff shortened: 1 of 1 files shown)


osu.Game/Users/Drawables/DrawableAvatar.cs
```diff
@@ -68,7 +68,7 @@ private void openProfile()
             if (!OpenOnClick.Value)
                 return;
 
-            if (user != null)
+            if (user != null && user.Id != 1)
```
peppy:
```suggestion
            if (user?.Id > 1)
```

peppy (changes requested):

peppy:
While this fix is okay, a better approach may be to not link in the first place, as to avoid showing an erroneous tooltip.

peppy (approved):

osu.Game/Users/Drawables/DrawableAvatar.cs
Yash-HCL: Probably a reasonable comment would also help to understand why should be user Id > 1? If not then what is the fallback?

peppy: probably making `GuestUser` a static/readonly and referencing its ID directly.

PR #8383, opened 2020-03-20
Disallow manual combo colouring when not on beatmap skin

### LINKED ISSUE #8282: Disabling beatmap skins option does not fully disable beatmap's combo colors
**Describe the bug:** 
When the option ``Beatmap skins`` is disabled, combo color usage is wrong.

* Expected: Combo colors go to only selected skin's values
* Actual: I think it reverts to a modulo operator, so if your skin has two combo colors, every even numbered color in map skin's will be your skin's second color, every odd one your first.

lazer side: https://www.youtube.com/watch?v=6e0_x9AzkRs
expected from stable: https://www.youtube.com/watch?v=3IZHCLjyXIo

**osu!lazer version:** 
2020.312.0

frenzibyte:
Closes #8282 

Disallows manual combo colouring (aka "color haxing") when not enabling beatmap skin, this is a behaviour on osu!stable and should be a thing here to avoid selecting combo colours of a different skin to hit objects that are not meant to be coloured in that way.

### `IHasComboInformation.ComboIndex` and `IHasComboInformation.ComboIndexBindable` have changed types from `int` to `ComboIndex (struct)`
Now `ComboIndex` provides two indices within its structure, automated index which increments by `1` per new combo, and manual index which increments by `1 + ComboOffset` per new combo, to allow switching between manual combo colouring and automated combo colouring.


osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+    public readonly struct ComboIndex : IEquatable<ComboIndex>
+    {
+        /// <summary>
+        /// The automated index value of the combo in relation to the beatmap.
+        /// </summary>
+        /// <remarks>
+        /// This is separated from <see cref="Manual"/> to disallow
+        /// manual combo colouring when not on the right skin to apply.
+        /// </remarks>
+        public readonly int Automated;
```
bdach: Not too sure about these names. Would rename to `Ordinal` (or `WithoutOffset`?)...

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        /// <summary>
+        /// The manual index value of the combo, this accounts for <see cref="IHasCombo.ComboOffset"/>.
+        /// </summary>
+        public readonly int Manual;
```
bdach: ...and this one to `WithOffset`.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// The automated index value of the combo in relation to the beatmap.
+        /// </summary>
+        /// <remarks>
+        /// This is separated from <see cref="Manual"/> to disallow
+        /// manual combo colouring when not on the right skin to apply.
+        /// </remarks>
+        public readonly int Automated;
+
+        /// <summary>
+        /// The manual index value of the combo, this accounts for <see cref="IHasCombo.ComboOffset"/>.
```
bdach:
```suggestion
        /// The manual index value of the combo. Accounts for <see cref="IHasCombo.ComboOffset"/>.
```

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+
+namespace osu.Game.Rulesets.Objects.Types
+{
+    /// <summary>
+    /// Represents a manual and automated combo indices.
+    /// </summary>
+    public readonly struct ComboIndex : IEquatable<ComboIndex>
+    {
+        /// <summary>
+        /// The automated index value of the combo in relation to the beatmap.
```
bdach:
```suggestion
        /// The automated index value of the combo in relation to the beatmap. Ignores <see cref="IHasCombo.ComboOffset"/>.
```

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static ComboIndex FromInt32(int i) => new ComboIndex(i);
+
+        public bool Equals(ComboIndex other) => Automated == other.Automated && Manual == other.Manual;
+        public override bool Equals(object obj) => obj is ComboIndex other && Equals(other);
+        public override int GetHashCode() => HashCode.Combine(Automated, Manual);
+
+        public static ComboIndex Add(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated + b.Automated, a.Manual + b.Manual);
+        public static ComboIndex Subtract(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated - b.Automated, a.Manual - b.Manual);
+        public static ComboIndex Multiply(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated * b.Automated, a.Manual * b.Manual);
+        public static ComboIndex Divide(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated / b.Automated, a.Manual / b.Manual);
```
bdach: I don't think these make a lot of sense (why would adding two combo indices add them element-wise?). Do they need to exist? I don't think there is *any* sane way to define arithmetic operators for this type.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        public static ComboIndex Increment(ComboIndex ci) => ci + 1;
+        public static ComboIndex Decrement(ComboIndex ci) => ci - 1;
```
bdach: These are supremely convoluted. To recap, you're relying *both* on the implicit conversion from `int` and the half-well-defined element-wise addition of combo indices to make this work. (And even then it's questionable if the increment operator should increment both members.)

bdach (changes requested):
I think I understand the issue this aims to fix. I am unsure if this is really what stable does, or if it's a behaviour we even want to keep, as it over-complicates matters significantly. I can't see what stable does easily without reverse-engineering, so I'll defer on that matter to the team.

Some gripes with the code, however.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: I've added these such that the structure can almost act as a normal integer, such that custom implementations that need to interact / modify the combo index (say combo tests) can do so easily, for example:
```csharp
public void ApplyToHitCircle(HitCircle circle)
{
    Scheduler.AddDelayed(circle.ComboIndex += 2, 250);
}
```
since combo index is initialized as `(0, 0)`, adding 2 would do for both indices, and the combo index will become `(2, 2)` which will pick the third combo colour no matter the conditions. (as if you're adding 2 to an `int`)

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: see the other conversation above's reply, if you want to increment a specific index, you can just increment it directly `ComboIndex.Automated++`.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: +1 for `Ordinal`, my english dictionary is still small :P

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: `Manual` goes well with `Ordinal` as this index can be set to literally anything, but since it should take value from `IHasCombo.ComboOffset`, `WithOffset` sounds fair.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
bdach: Yes, I get the idea. Thing is, with the current implementation, you also allow things like
```csharp
new ComboIndex(2, 6) + new ComboIndex(1, 5)
```
which will yield `new ComboIndex(3, 11)` but what I'm trying to get across is that *it doesn't really make much sense to do that semantically*. Those operations are ill-defined and *should not be possible to do in an implicit manner*. Even worse are the division operators. I don't think being able to do division on two instances of `ComboIndex` is going to ever be defensible.

Point is, (a) those are at this stage unused, and I generally discourage from trying to future-proof as it just produces dead code that might have unknown bugs, and (b) the API for manipulation of combo indices here is not really sane in many cases. Operator overloading in general I also think should be used *very* sparingly. At most I'd allow
```csharp
public struct ComboIndex
{
    public void Add(int amount);
    public void Subtract(int amount);
}
```
but even *then* strange issues can arise (what if you try to do `new ComboIndex(4, 0).Subtract(2)`? `ComboIndex(2, -2)` isn't really sane either, is it?)

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,71 @@
...
+        public static ComboIndex FromInt32(int i) => new ComboIndex(i);
+
+        public bool Equals(ComboIndex other) => Automated == other.Automated && Manual == other.Manual;
+        public override bool Equals(object obj) => obj is ComboIndex other && Equals(other);
+        public override int GetHashCode() => HashCode.Combine(Automated, Manual);
+
+        public static ComboIndex Add(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated + b.Automated, a.Manual + b.Manual);
+        public static ComboIndex Subtract(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated - b.Automated, a.Manual - b.Manual);
+        public static ComboIndex Multiply(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated * b.Automated, a.Manual * b.Manual);
+        public static ComboIndex Divide(ComboIndex a, ComboIndex b) => new ComboIndex(a.Automated / b.Automated, a.Manual / b.Manual);
```
huoyaoyuan: Strongly disagree against method alternatives of operators.
Disagree with meaningless operations.

Proposed:
```csharp
public struct ComboIndex
{
    public static ComboIndex operator+(ComboIndex left, int count);
    public static ComboIndex operator-(ComboIndex left, int count);
}
```

iiSaLMaN: method alternatives for operators are actually inspection rules, also duplicate of what @bdach said.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: Went ahead with static `Add` and `Subtract` methods only with filled-up XMLDocs to avoid misusing it, I suppose that's the least sane way.

huoyaoyuan:
Note: some CLS compliant rules or inspection rules are inherited from 2005, at which time VB had less features. Nowadays, just considering modern C# consumption is enough.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,92 @@
...
+    public readonly struct ComboIndex : IEquatable<ComboIndex>
+    {
+        /// <summary>
+        /// The ordinal index value of the combo in relation to the beatmap. Ignores <see cref="IHasCombo.ComboOffset"/>.
+        /// </summary>
+        /// <remarks>
+        /// This is separated from <see cref="WithOffset"/> to disallow
+        /// manual combo colouring when not on the right skin to apply.
+        /// </remarks>
+        public readonly int Ordinal;
```
peppy: so is this the combo's index in the beatmap? if so, you should also look at `IHasCombo.ComboOffset` as it may be similar to what is provided here.

as it stands, this documentation/class feels very hard to understand.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// Represents a manual and automated combo indices.
+    /// </summary>
+    public readonly struct ComboIndex : IEquatable<ComboIndex>
+    {
+        /// <summary>
+        /// The ordinal index value of the combo in relation to the beatmap. Ignores <see cref="IHasCombo.ComboOffset"/>.
+        /// </summary>
+        /// <remarks>
+        /// This is separated from <see cref="WithOffset"/> to disallow
+        /// manual combo colouring when not on the right skin to apply.
```
peppy: "not on the right skin to apply"

i don't know what this means

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
iiSaLMaN: isn't `IHasCombo.ComboOffset` just the amount of combos to skip from current combo index in the processing? this index doesn't let that in its calculation: `Ordinal = lastObj.Ordinal + 1;` (which just increments by 1 per new combo, or as you said "the combo's index in the beatmap")

If osu!stable provides something much saner / straightforward than this, would be better to let me know I suppose.

bdach (commented):
No major mechanical objections to the code itself, although again, I'm not 100% sure if this is the behaviour stable has or if we want to preserve it. (We probably do, but it's a little complicated to keep track of in the way done here; maybe this logic should be somehow encapsulated in legacy beatmap decoding?)

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// The ordinal index value of the combo in relation to the beatmap. Ignores <see cref="IHasCombo.ComboOffset"/>.
+        /// </summary>
+        /// <remarks>
+        /// This is separated from <see cref="WithOffset"/> to disallow
+        /// manual combo colouring when not on the beatmap's skin.
+        /// </remarks>
+        public readonly int Ordinal;
+
+        /// <summary>
+        /// The manual index value of the combo. Accounts for <see cref="IHasCombo.ComboOffset"/>.
```
peppy: what does manual even mean here? i'd probably remove the word.

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// </summary>
+        public readonly int WithOffset;
```
peppy: `WithBeatmapOffset`

osu.Game/Rulesets/Objects/Types/ComboIndex.cs
```diff
@@ -0,0 +1,92 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Game.Rulesets.Objects.Types
+{
+    /// <summary>
+    /// Represents an <see cref="Ordinal"/> and <see cref="WithOffset"/> combo indices.
```
peppy: this isn't english nor does it describe the class. don't describe something using itself.

iiSaLMaN: I've reworded it, please check if it's good now.

iiSaLMaN:
> No major mechanical objections to the code itself, although again, I'm not 100% sure if this is the behaviour stable has or if we want to preserve it. (We probably do, but it's a little complicated to keep track of in the way done here; maybe this logic should be somehow encapsulated in legacy beatmap decoding?)

I think this is the best approach FWIW. encapsulating in legacy beatmap decoder isn't clear to me, can you elaborate further?

bdach:
The loose idea would be to check if the beatmap has a skin at parse time and then expose a `ComboIndex` or something at the legacy beatmap level if it does have a skin (so basically have it in `IHasCombo`, maybe?) and then use it only in skinning. I haven't looked in detail if that's viable, however.

iiSaLMaN:
> The loose idea would be to check if the beatmap has a skin at parse time and then expose a `ComboIndex` or something at the legacy beatmap level if it does have a skin (so basically have it in `IHasCombo`, maybe?) and then use it only in skinning. I haven't looked in detail if that's viable or if it'll look good in code, however.

so you're simply saying to move `ComboIndex.WithBeatmapOffset` to `IHasCombo` and set it on parse time, that almost sounds like what this PR does but slightly different and on parse time, which is incorrect because the combo index processing must be inside `BeatmapProcessor` for beatmap editing purposes, unless I'm mistaking?

frenzibyte:
I have another way that doesn't affect the `IHasComboInformation` interface like that, going to close this as it's gonna be a complete rewrite.

PR #8384, opened 2020-03-21
Let carousel root use default selection when previous selection is filtered

LittleEndu:
~~Currently there is no test for this behaviour. Behaviour which I think is actually intended. This test would fail if #8368 is merged ([see this comment](https://github.com/ppy/osu/pull/8368#discussion_r395732948)).~~

This PR will let carousel root use default selection strategy when previous selection is filtered instead of selecting a random beatmap. I think this is preferred, because when user doesn't explicitly randomize the selection the selection should be near the previous selection (so the user wouldn't need to scroll back near to where their previous selection was after carousel root selected a random beatmap)

Also this PR adds test cases for this behaviour.

* [x] Depends on #8370
* [x] Depends on #8368


peppy:
Please wait until other ongoing changes are merged, then propose a change (preferrably via an issue or PR with this test attached). This is a logic requirement which requires further discussion; adding a test ahead of time in order to intentionally break other PRs is generally not how we do things here.

If you are going to keep this PR open, please rebase it on top of #8370 and #8368.

LittleEndu:
> If you are going to keep this PR open, please rebase it on top of #8370 and #8368.

Should I also add the proposed change ([from the comment](https://github.com/ppy/osu/pull/8368#discussion_r395732948)) to this PR then?

peppy:
Yes. You would need to make the required logic change to satisfy your tests. If you are worried about wasted effort for a proposal change that may not be executed, open an issue instead.

Keep in mind that each change you make should be well defined. One purpose only. Make one PR per logic change and minimise the changes you are making. If you are switching logic to "select next", make sure you also cover cases such as ruleset changes, where the "next" item may be 500 or 5000 items down in the list, arbitrarily selected instead of random.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -751,13 +750,17 @@ private class CarouselRoot : CarouselGroupEagerSelect
...
+                State.ValueChanged += state => State.Value = CarouselItemState.Selected;
+
                 this.carousel = carousel;
             }
 
             protected override void PerformSelection()
             {
                 if (LastSelected == null)
-                    carousel.SelectNextRandom();
+                    carousel?.SelectNextRandom();
```
LittleEndu: When would ``carousel`` ever be null? ``CarouselRoot`` constructor is never called with null and the value of ``carousel`` is never changed.

LittleEndu: I guess I should have commented this on peppy's PR instead

peppy: you can remove the check and run in debug

peppy:
I can't agree with this change. Now switching rulesets always leave song select at the top of the list, or in a completely random (yet *not* actually random) place. This feels very bad. If we're going to use "next" logic at *very least* the "next" should be on-screen, or once we have grouping in place, in the same group.

LittleEndu:
That sounds like it shouldn't be happening. The logic here uses the same index saving logic that's already in CarouselGroup I believe. It should just select the next beatmap. Also if there are no beatmaps from the selected ruleset on screen then the next logic would never be able to select something that would be on screen.

I'll take a look at it anyway. The logic in CarouselGroup seemed weird that it only looks forwards even if looking backwards would yield a closer beatmap.

LittleEndu:
You can change CarouselGroupEagerSelect's nextToSelect to something like ``Children.OrderBy(i => Math.Abs(Children.IndexOf(i) - lastSelectedIndex)).FirstOrDefault(i => !i.Filtered.Value);`` to make it also look backwards. Sounds like the right thing to do, but I would rather change it after #8444 is merged.

This still won't change the way carousel acts on ruleset change. If you don't allow conversion then most beatmaps just won't have difficulties for other modes and carousel can't do anything but to select something that's way off screen.

LittleEndu:
I accidentally hit the merge master button here, but now I'm concerned about why the PR size dropped...

osu.Game/Screens/Select/Carousel/CarouselGroupEagerSelect.cs
```diff
@@ -101,8 +101,7 @@ private void attemptSelection()
 
         protected virtual CarouselItem GetNextToSelect()
         {
-            return Children.Skip(lastSelectedIndex).FirstOrDefault(i => !i.Filtered.Value) ??
-                   Children.Reverse().Skip(InternalChildren.Count - lastSelectedIndex).FirstOrDefault(i => !i.Filtered.Value);
+            return InternalChildren.OrderBy(i => Math.Abs(InternalChildren.IndexOf(i) - lastSelectedIndex)).FirstOrDefault(i => !i.Filtered.Value);
```
LittleEndu, line 104: ``OrderBy`` is probably wasteful. Could just get first visible item both directions and check which one is closer to last selection.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -714,7 +731,8 @@ public void TestFilteringByUserStarDifficulty()
             checkVisibleItemCount(true, 15);
         }
 
-        private void loadBeatmaps(List<BeatmapSetInfo> beatmapSets = null, Func<FilterCriteria> initialCriteria = null, Action<BeatmapCarousel> carouselAdjust = null, int? count = null, bool randomDifficulties = false)
+        private void loadBeatmaps(List<BeatmapSetInfo> beatmapSets = null, Func<FilterCriteria> initialCriteria = null, Action<BeatmapCarousel> carouselAdjust = null, int? count = null,
+                                  bool randomDifficulties = false)
```
peppy, line 735: what is this? it's never used.

osu.Game/Screens/Select/Carousel/CarouselGroupEagerSelect.cs
```diff
@@ -101,8 +101,12 @@ private void attemptSelection()
 
         protected virtual CarouselItem GetNextToSelect()
         {
-            return Children.Skip(lastSelectedIndex).FirstOrDefault(i => !i.Filtered.Value) ??
-                   Children.Reverse().Skip(InternalChildren.Count - lastSelectedIndex).FirstOrDefault(i => !i.Filtered.Value);
+            var forwards = Children.Skip(lastSelectedIndex).FirstOrDefault(i => !i.Filtered.Value);
```
peppy, line 104: This method needs documentation via comments. Also potential consideration to the fact that by resolving these (via `FirstOrDefault`) ahead of time, you are incurring a much higher performance overhead, which may be important since this is operating on all beatmaps in the carousel. The previous code was not doing this.

peppy (changes requested):

peppy:
The issue I pointed out above is still and issue (https://github.com/ppy/osu/pull/8384#issuecomment-612561107). Should we just close this PR or are you going to look to fixing it? The new behaviour is objectively worse than what is on master.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
LittleEndu: idk, is this a change from me hitting "reformat code". Rider probably doesn't like how long this line is, but line length isn't actually enforced.

osu.Game/Screens/Select/Carousel/CarouselGroupEagerSelect.cs
LittleEndu: Yeah, but if we want carousel to move in both directions then resolving ahead of time is still O(n) instead of sorting that I had here few commits before which is O(nlogn). 

And I think this is also fine because carousel root only calls this eagerly, that is when it loses selection which can only occur on filter changes.

LittleEndu:
I already answered to that comment back in April. If there is a beatmap from the next ruleset on screen it will be selected, but most of the time when switching from standard there just isn't. Also looking at this PR with #11176 also merged you can see that the selected beatmap is the nearest and doesn't look half as bad. But then again I really like the carousel animation.

peppy:
When there isn't a beatmap on screen it hits the top of the carousel every time. Previously it didn't. This cannot happen.

PR #8385, opened 2020-03-21
Move execution of completion to performUserRequestedExit

### LINKED ISSUE #8381: Interface bug after exiting the map
This bug occurs between the moments when the last note is pressed, but the results screen has not yet appeared. If at this moment you exit the map, then the interface will catch a serious visual glitch, which is not fixed by anything (only complete exit from the application).

You can repeat as many times as you like

https://www.youtube.com/watch?v=erCzCjhRCQk
osu!lazer version:
2020.320.1
Logs:
[performance.log](https://github.com/ppy/osu/files/4359602/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4359603/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4359604/updater.log)
[database.log](https://github.com/ppy/osu/files/4359605/database.log)

mcendu:
Fixes #8381. Makes sure the screen push does not collide with exiting process.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -356,6 +356,13 @@ namespace osu.Game.Screens.Play
         {
             if (!this.IsCurrentScreen()) return;
 
+            if (completionProgressDelegate != null && !completionProgressDelegate.Cancelled && !completionProgressDelegate.Completed)
+            {
+                // proceed to result screen if beatmap already finished playing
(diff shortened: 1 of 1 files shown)


peppy:
This doesn't seem correct. It needs to be in the exit flow to fix another race condition. First is figuring out why it is not working there, and then finding a better solution. May need a schedule if a push can't happen during an exit.

mcendu:
Closing as this approach does not fix the bug.

PR #8397, opened 2020-03-21
Don't scroll NowPlayingOverlay background when it's hidden

### LINKED ISSUE #8393: Now playing overlay sometimes has to scroll to correct background when opened
**Describe the bug:** 
When you open now playing overlay, it displays the background of the current beatmap. If the beatmap changes it scrolls to the correct background. This background should already be the current one when you first open the now playing overlay, but it sometimes (seemingly randomly) isn't.

https://youtu.be/d-Mov8UJjco

**osu!lazer version:** 
2020.320.1

LittleEndu:
Closes #8393

Before, NowPlayingOverlay would (almost) always scroll the background, which looks weird when first opening it. This PR makes it so that the background is not scrolled while NowPlayingOverlay is hidden, so that there aren't two transformations (overlay fading in and background moving to the side) happening at the same time when you open it later.


bdach:
An explanation what the problem was and what the fix is would be appreciated. Of course I can try to find out myself but things go faster if there is a proper reasoning in the PR description.

LittleEndu:
Description added. Thought the title and the issue the PR was fixing were descriptive enough.

LittleEndu:
New tests are going to fail. I'm unsure how you would add beatmaps with virtual tracks to beatmap manager

peppy:
The sideways movement was intentional though, as I attempted to convey in the issue thread.

osu.Game/Overlays/NowPlayingOverlay.cs
```diff
@@ -253,25 +253,43 @@ protected override void Update()
...
+                artist.Text = new LocalisedString((metadata.ArtistUnicode, metadata.Artist));
+            }
+        }
+
         private void trackChanged(WorkingBeatmap beatmap, TrackChangeDirection direction = TrackChangeDirection.None)
         {
+            if (Alpha == 0)
+            {
+                setMetadata(beatmap);
+                LoadComponentAsync(new Background(beatmap) { Depth = float.MaxValue }, newBackground =>
```
peppy: not even worth mentioning but this is a memory / performance ordeal. read comments below to know why

LittleEndu: How would one see this performance drop? Just got done clicking next track 1000x and didn't notice any increase in memory usage.

peppy: while the now playing overlay is not present, it will still be loading components in the background, but not process the callbacks (they are scheduled to itself).

they will pile up, causing memory leaks (beatmap references, as mentioned in the comment below) and also add the overhead of doing the `LoadComponentAsync` of the background component.

LittleEndu: I'm still unsure how one would intentionally hit this performance drop. I'm running the below test and don't see my memory usage going above 1.3GB 
```cs
        [Test]
        public void TestManyNext()
        {
            AddStep("Add overlay", () =>
            {
                if (!Children.ToList().Contains(nowPlayingOverlay))
                    Add(nowPlayingOverlay);
            });
            AddStep("Remove overlay", () => {
                Remove(nowPlayingOverlay);
            });
            for (int i = 0; i < 1000; i++)
            {
                currentBeatmap = Beatmap.Value;
                AddStep("next track", () => musicController.NextTrack());
                AddUntilStep("track changed", () => Beatmap.Value != currentBeatmap);
            }
        }
```

peppy: if you can't figure it out please just understand the comments and try and understand the underlying logic. you can google/youtube on reference leaks and how they can occur if you want to learn more.

osu.Game.Tests/Visual/UserInterface/TestSceneNowPlayingOverlay.cs
```diff
@@ -86,11 +102,32 @@ public void TestBackgroundScrolling(bool goNext)
...
+                if (goNext)
+                    AddStep("next track", () => musicController.NextTrack());
+                else
+                    AddUntilStep("previous track", () => musicController.PreviousTrack() != PreviousTrackResult.Restart);
+            }
+
+            step();
+            AddAssert("Background changed", () => getFirstBackground() != currentBackground);
+            step();
+            AddAssert("Previous background is present", () => currentBackground.IsPresent);
```
LittleEndu: idk why this fails even though you can clearly see the previous background still scrolling by

LittleEndu: I was assigning value I wanted to test outside a step. Got the tests working in visual now.

peppy:
I will reiterate that we do still want the animation for the time being.

LittleEndu:
I can probably fix tests failing in current master now.

peppy:

PR #8402, opened 2020-03-22
Fix test scene potentially missing dependencies

frenzibyte:
*osu!framework*'s test scenes must never be used in here to avoid not running tests from `OsuGameBase`. (which causes missing dependencies as seen [here](https://ci.appveyor.com/project/peppy/osu/builds/31612254/tests))


bdach (dismissed):
Not sure how I missed the fact that I imported the wrong one when originally writing this scene but it does make sense to use the game-local one 👍

peppy:
Ouch, let's rename that class.

peppy (approved):

PR #8406, opened 2020-03-23
Implement random mod for taiko

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8405

Was requested for stable here: https://github.com/ppy/osu-stable-issues/issues/399


osu.Game.Rulesets.Taiko/Mods/TaikoModRandom.cs
```diff
@@ -0,0 +1,27 @@
...
+        public override string Description => @"Shuffle around the colours!";
+
+        public void ApplyToBeatmap(IBeatmap beatmap)
+        {
+            var taikoBeatmap = (TaikoBeatmap)beatmap;
+
+            foreach (var obj in taikoBeatmap.HitObjects)
+            {
+                if (obj is Hit hit)
+                    hit.Type = RNG.Next(2) == 0 ? HitType.Centre : HitType.Rim;
```
peppy: did we want this to be seeded?

smoogipoo: Eventually, but mania should also be seeded yet it isn't, so I didn't do it here either.

peppy (approved):

PR #8407, opened 2020-03-23
Disable raw input toggle on all but windows

peppy:
Until we properly test and support this, let's just disable it.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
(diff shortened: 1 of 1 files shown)


bdach:
Source looks obvious enough, but if you'd like this tested on linux & android give me a day and I'll make extra sure.

bdach:
Toggle is disabled fine on android and linux, two gripes however:

* The disabled raw input toggle somehow looks different than the disabled sensitivity slider:

  2020-03-24-220256_542x249_scrot

* If someone already has raw input enabled it will stay enabled, and with this disable the hotkey is the only remaining way to disable it back manually. Possibly worth considering resetting it ourselves on relevant platforms?

Joehuu:
The dim was fixed in https://github.com/ppy/osu/pull/8408.

peppy:
Out of curiosity, is anyone relying on raw input being *available* on linux? ie. is it working as-expected for anyone?

peppy:
Going to merge this for now.

sawe871:
> Out of curiosity, is anyone relying on raw input being _available_ on linux? ie. is it working as-expected for anyone?

I was using raw input as a fix for my tablet being more floaty than normal in xfce. It was working as expected. The only issue was that the cursor would slowly drift to the top left corner of the screen, but it was manageable by keeping the cursor in motion.

bdach:
Sorry to hear; it'll definitely return when there's throughput for it, but as of now it was generating a lot of noise by provoking reports from people for whom the current implementation of raw input was broken. Hopefully will get better when we've fully transitioned to SDL.

peppy:
@sawe871 you can easily undo this change by building from source from now, so I'd recommend that short-term.

ReNForce72:
Is this issue still being worked on? I play on Mac, and I the delay without the raw input option is really rough. I've been playing a browser-based version of Osu instead and it works just fine. Just wondering when this will get resolved so I can finally play the official version of this game.

frenzibyte:
Please open a new issue, this is completely irrelevant. (and raw input is back as a "high precision mode" toggle, not sure if it's completely raw though)

PR #8409, opened 2020-03-23
Fix crash when holding a key down while entering player

### LINKED ISSUE #8168: Crash when changing filter query during transition to player loader
If you press Enter and any button on the keyboard at the same time (or with very little delay), then there are 3 options:
1. Everything will be fine.
2. The application will give an error, but will continue to work.
3. The application crashes (as I understand it, this happens if you click something immediately after clicking on Enter, and not before that)

https://youtu.be/8qSJ2gU9yiM
**osu!lazer version:** 
2020.306.0
**Logs:**
Logs are available only from complete crash.
[performance.log](https://github.com/ppy/osu/files/4302609/performance.log)
[network.log](https://github.com/ppy/osu/files/4302610/network.log)
[runtime.log](https://github.com/ppy/osu/files/4302611/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4302612/updater.log)

Logs of the error that is shown in the lazer (second option). Apparently, this error is also shown here.
[runtime.log](https://github.com/ppy/osu/files/4302618/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4302619/updater.log)
[network.log](https://github.com/ppy/osu/files/4302620/network.log)
[performance.log](https://github.com/ppy/osu/files/4302621/performance.log)

peppy:
- Closes https://github.com/ppy/osu/issues/8168

Reproduction steps:
- Make sure you have one or more beatmaps matching search for a single `/`.
- Make sure you have no beatmaps matching search for `//`.
- Hold key down `/` until one character appears in the search box.
- While keeping key held down, click (has to be a click) on the osu! logo before the next `/` appears (from key repeat).

If it doesn't crash, repeat from scratch until it does (timing is quite important).

The cause of this can be seen in [TextBox](https://github.com/ppy/osu-framework/blob/43edfc3653444793bd8268e5d5c7f66705b519e2/osu.Framework/Graphics/UserInterface/TextBox.cs#L625), where text will continue to be consumed and fire events until the next key up event. We were not protecting song select against this scenario.

Easy fix is making the textbox readonly at the moment we no longer want the filter control to apply.

No tests as hard to test.

### DIFF
--- a/osu.Game/Screens/Select/FilterControl.cs
+++ b/osu.Game/Screens/Select/FilterControl.cs
@@ -136,6 +136,8 @@ namespace osu.Game.Screens.Select
 
         public void Deactivate()
         {
+            searchTextBox.ReadOnly = true;
+
             searchTextBox.HoldFocus = false;
             if (searchTextBox.HasFocus)
                 GetContainingInputManager().ChangeFocus(searchTextBox);
@@ -143,6 +145,7 @@ namespace osu.Game.Screens.Select
 
         public void Activate()
         {
+            searchTextBox.ReadOnly = false;
             searchTextBox.HoldFocus = true;
         }
 


bdach (approved):
Could reproduce before, could not reproduce after, nice catch 👍

smoogipoo (approved):

PR #8417, opened 2020-03-23
Scroll to top when pressing escape on a specific changelog build

Joehuu:
When escaping back to the changelog listing, you are in the middle of the listing where you were last scrolled to. This puts you back up, feels better that way.

### DIFF
--- a/osu.Game/Overlays/ChangelogOverlay.cs
+++ b/osu.Game/Overlays/ChangelogOverlay.cs
@@ -33,6 +33,8 @@ namespace osu.Game.Overlays
 
         private List<APIChangelogBuild> builds;
 
+        private OsuScrollContainer scrollContainer;
+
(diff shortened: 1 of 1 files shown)


bdach:
I don't really agree with the direction of this PR. In my mind if I'm in nested scrolling views, going back out of the inner one should bring me to the same point I was in the outer one, so I can continue reading from the point at which I was on.

On that note, I actually think there is *another* place where this would be better-advised - to me it doesn't make sense that when I go from the main changelog scroll to an individual build, I end up at the *end* of the build changelog and have to manually scroll back up to read it from the start (see [this video](https://streamable.com/5rmig) for what I mean).

peppy:
Easiest solution here would be splitting each piece into its own scroll container, i guess.

Joehuu:
Closing for now as I cannot implement this properly.

PR #8441, opened 2020-03-25
Fix last seen date being visible in user panel when it shouldn't

### LINKED ISSUE #8438: "Last seen" text doesn't disappear from user panel while getting back online in song select/gameplay/editor/etc.
(Note: it works correctly in the main menu with the default "_Online_" message)

user panel gif

Here's an image of the issue:

user panel

**osu!lazer version:** 2020.325.0

EVAST9919:
Fixes https://github.com/ppy/osu/issues/8438

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneUserPanel.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneUserPanel.cs
@@ -28,7 +28,7 @@ namespace osu.Game.Tests.Visual.Online
         private readonly Bindable<UserStatus> status = new Bindable<UserStatus>();
 
         private UserGridPanel peppy;
-        private UserListPanel evast;
+        private TestUserListPanel evast;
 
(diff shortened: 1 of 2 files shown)


bdach (commented):
looks good to me, maybe could have avoided exposing `LastVisitMessage` but at cost of maintainability later so probably better to do it this way

osu.Game.Tests/Visual/Online/TestSceneUserPanel.cs
```diff
@@ -38,6 +38,9 @@ public class TestSceneUserPanel : OsuTestScene
         {
             UserGridPanel flyte;
 
+            activity.Value = null;
+            status.Value = null;
```
bdach: Would have probably taken off the `readonly` modifier and reconstructed every test but this works too I suppose.

peppy (approved):

PR #8444, opened 2020-03-25
Make beatmap carousel select recommended difficulties

### LINKED ISSUE #8612: Random always selects lowest star beatmap on beatmap carousel
I must have clicked on random almost 100 times and it always chooses the lowest star beatmap of the beatmap set. Seems to happen also on startup when you have yet to choose a song.

osu!lazer version: 2020.404.0

LittleEndu:
Early PR. Haven't made tests for this and this needs a way to find out user's current ruleset pp (PR right now uses user's default ruleset's pp)

This PR changes CarouselGroupEagerSelect's selection logic to allow its subclasses to overwrite it. Then there's new logic in CarouselBeatmapSet that selects it's children based on the recommended difficulty. Recommendation algorithm was taken from [osu-web](https://github.com/ppy/osu-web/blob/e6e2825516449e3d0f3f5e1852c6bdd3428c3437/app/Models/User.php#L1505).

Fixes #8612


bdach:
This should have been opened as a draft (or not at all, until the other missing pieces are in place).

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -31,6 +33,8 @@ public class BeatmapCarousel : CompositeDrawable, IKeyBindingHandler<GlobalActio
         private const float bleed_top = FilterControl.HEIGHT;
         private const float bleed_bottom = Footer.HEIGHT;
 
+        private readonly Bindable<User> localUser = new Bindable<User>();
```
bdach: I have mostly negative feelings about the carousel being API- and `User`-dependent with this change. I'd say suggestion logic should be hidden somewhere so that the carousel logic (which is already convoluted in places) doesn't now become entirely unreadable and untestable offline due to online coupling.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -4,30 +4,56 @@
...
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Game.Beatmaps;
 using osu.Game.Screens.Select.Filter;
+using osu.Game.Users;
 
 namespace osu.Game.Screens.Select.Carousel
 {
     public class CarouselBeatmapSet : CarouselGroupEagerSelect
     {
+        private readonly Bindable<User> localUser;
```
bdach: As above - this ideally should be avoided / encapsulated somewhere IMO.

Also I fail to see the point of injecting this manually via ctor from `BeatmapCarousel` when you could use DI directly here & resolve.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -4,30 +4,56 @@
...
+            this.localUser = localUser;
         }
 
         protected override DrawableCarouselItem CreateDrawableRepresentation() => new DrawableCarouselBeatmapSet(this);
 
+        protected override CarouselItem GetNextToSelect()
+        {
+            if (LastSelected == null)
+            {
+                decimal? pp = localUser.Value?.Statistics?.PP ?? 60; // TODO: This needs to get ruleset specific statistics
```
bdach: I don't think it's sane to assume a start-up PP level of 60 for a potential new player.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -4,30 +4,56 @@
...
+
+                var recommendedDifficulty = Math.Pow((double)pp, 0.4) * 0.195;
```
bdach: Not a fan of this being duplicated lazer- and web-side. Candidate for inclusion in API, perhaps?

bdach (commented):
Needs tests, as you said. Whatever solution you come up with in the end also *has* to somehow work offline. I'd imagine it should be disabled entirely in that case, which wouldn't work in the current state with your coercion of user PP to 60.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
LittleEndu: pp of 60 would recommend difficulties close to 1.003 stars

bdach: ...which is entirely dependent on diffcalc logic. What if it changes? I'd say it's vastly simpler to just select easiest diff.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
LittleEndu: Not sure how DI works for classes that aren't drawables.

bdach: It doesn't, which I suppose explains it. It's still really strange either way.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
peppy: I think it's fine to have this in both places, with a comment linking them. We don't really want to re-retrieve recommended difficulties every time we update this algorithm.

LittleEndu:
Should I close this in favour of making #8485 cover this change too. The logic there has changed a lot to cover highlighting design.

peppy:
I'd prefer to see this rebased after #8485 is merged, if anything.

LittleEndu:
Did you mean some other PR? #8485 is dependent on this PR so you can't merge it before it

peppy:
Oh, that feels a bit back to front. Will wait to review this to make a call then.

LittleEndu:
I made the selection code before the highlighting one. You could probably go either way, but that's how I've done it. After what was commented on they highlighting PR, I'm not so sure if we even want that behaviour.

peppy:
I don't think `CarouselGroupEagerSelect` or `CarouselBeatmapSet` should expose any more than a `Func<IEnumerable<Beatmap>,Beatmap> GetRecommdedSelection`. All the API logic should be moved out to `BeatmapCarousel` at very least.

LittleEndu:
So move API logic to SongSelect?

peppy:
I think `SongSelect` is probably best, yes.

peppy:
Also is there a reason you need to do an API call when the local user is already present in `OsuGame`?

LittleEndu:
Local user does not have current ruleset, only user's default

peppy:
Fair enough. If I was doing this I'd probably make it all in its own component (`DifficultyRecommender : Component` or something) and then provide a function from that component to `BeatmapCarousel`.

LittleEndu:
You can only get current ruleset via DI, or is there a component to read it from?

Just a thought I had, you could just iterate over all rulesets, make a API call for each, and keep the recommendation the same for the entire run.

peppy:
Just using DI is fine.

LittleEndu:
Had to remove tests. Unsure how this behaviour could be tested now that API logic is outside BeatmapCarousel.

peppy:
Make some changes (@LittleEndu was watching). Looks good from my end. @smoogipoo requesting review since I made significant LoC changes.

peppy (dismissed):

smoogipoo (approved):
lgtm

Issue #8085, opened 2020-03-02
LegacySkinDecoder does not support [CatchTheBeat] sections


smoogipoo:
For reference: https://osu.ppy.sh/help/wiki/Skinning/skin.ini#[catchthebeat]

PR #8446, opened 2020-03-26
Add support for custom hyper-dash legacy fruit colouring

frenzibyte:
Also adds `CatchSkinConfiguration` and sets up a test scene for upcoming custom hyper-dash catcher colouring and others.

Partially (resolves) #8085


osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,183 @@
...
+                Child = setupSkinHierarchy(() =>
+                    drawableFruit = new DrawableFruit(fruit)
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Scale = new Vector2(4f),
+                    }, false, false);
+            });
+
+            AddAssert("fruit colour default-hyperdash", () => checkFruitHyperDashColour(drawableFruit, Catcher.DefaultHyperDashColour, legacyFruit));
```
bdach: oh, and these test steps should be named better. they're borderline illegible now.

osu.Game.Rulesets.Catch/Objects/Drawables/FruitPiece.cs
```diff
@@ -60,14 +63,19 @@ private void load(DrawableHitObject drawableObject)
                 },
             });
 
+            var hyperDashColour =
+                skin.GetConfig<CatchSkinConfiguration, Color4>(CatchSkinConfiguration.HyperDashFruit)?.Value ??
+                skin.GetConfig<CatchSkinConfiguration, Color4>(CatchSkinConfiguration.HyperDash)?.Value ??
+                Catcher.DefaultHyperDashColour;
```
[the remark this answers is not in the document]
iiSaLMaN: I don't think there's a way to fix that, thinking about introducing an attribute that contains the fallback setting / default value here though.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,183 @@
...
+                if (componentName == "fruit-pear")
+                {
+                    // convince CatchLegacySkinTransformer to use the LegacyFruitPiece for pear fruit.
+                    var texture = new Texture(Texture.WhitePixel.TextureGL)
+                    {
+                        Width = 1,
+                        Height = 1,
+                        ScaleAdjust = 1 / 96f
+                    };
+                    return texture;
```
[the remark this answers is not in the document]
iiSaLMaN: why do you think that, way better than looking up for a texture file or something.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,173 @@
...
+            }
+
+            public Drawable GetDrawableComponent(ISkinComponent component) => null;
+
+            public Texture GetTexture(string componentName)
+            {
+                if (componentName == "fruit-pear")
+                {
+                    // convince CatchLegacySkinTransformer to use the LegacyFruitPiece for pear fruit.
+                    return new Texture(Texture.WhitePixel.TextureGL)
```
peppy: why not just return `Texture.WhitePixel` itself?

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,173 @@
...
+            var testSkinProvider = new SkinProvidingContainer(new TestLegacySkin(customHyperDashCatcherColour, customHyperDashFruitColour, customHyperDashAfterColour));
+
+            var legacySkinTransformer = new SkinProvidingContainer(new CatchLegacySkinTransformer(testSkinProvider));
+
+            return testSkinProvider
+                .WithChild(legacySkinTransformer
+                    .WithChild(getChild.Invoke()));
+        }
+
+        private bool checkFruitHyperDashColour(DrawableFruit fruit, Color4 expectedColour, bool isLegacyFruit) =>
```
peppy: is this exists it should be two functions

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        [TestCase(false)]
+        [TestCase(true)]
+        public void TestCustomHyperDashFruitColourFallback(bool legacyFruit)
+        {
+            DrawableFruit drawableFruit = null;
+
+            AddStep("setup fruit", () =>
+            {
+                var fruit = new Fruit { IndexInBeatmap = legacyFruit ? 0 : 1, HyperDashTarget = new Banana() };
```
peppy: i don't get this use of `IndexInBeatmap`

osu.Game.Rulesets.Catch/Skinning/CatchSkinConfiguration.cs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Rulesets.Catch.Skinning
+{
+    public enum CatchSkinConfiguration
```
peppy: shouldn't this be `CatchSkinColour`?

peppy (changes requested):

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
iiSaLMaN: then the fruit will be displayed in 1px

peppy: then use a real texture or something. i thought this was just to "convince.."

osu.Game.Rulesets.Catch/Skinning/CatchSkinConfiguration.cs
iiSaLMaN: I thought the `XXXSkinColour` enum concept is for doing a lookup on the custom colours of the skin configuration, but no harm in renaming here.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
iiSaLMaN: Added a default legacy skin provider instead to make things straightforward.

osu.Game.Rulesets.Catch/Skinning/CatchSkinConfiguration.cs
peppy: Isn't that what you're doing? I'm confused.

iiSaLMaN: as in `SkinConfiguration.CustomColours` which is retrieved from the `Colours` section.

bdach (commented):
Looks somewhat acceptable except for some nits.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,177 @@
...
+                    }, true, false, false, legacyFruit);
+            });
+
+            AddAssert("catcher custom colour", () =>
+                legacyFruit
+                    ? checkLegacyFruitHyperDashColour(drawableFruit, TestSkin.CustomHyperDashColour)
+                    : checkFruitHyperDashColour(drawableFruit, TestSkin.CustomHyperDashColour));
+        }
+
+        private Drawable setupSkinHierarchy(Drawable child, bool customCatcherColour = false, bool customAfterColour = false, bool customFruitColour = false, bool legacySkin = true)
```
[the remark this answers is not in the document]
iiSaLMaN: almost most of the test cases here do not set some of these 'custom' bools, but sure.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,177 @@
...
+
+                Child = setupSkinHierarchy(drawableFruit = new DrawableFruit(fruit)
+                {
+                    Anchor = Anchor.Centre,
+                    Origin = Anchor.Centre,
+                    Scale = new Vector2(4f),
+                }, false, false, false, legacyFruit);
+            });
+
+            AddAssert("default colour", () =>
```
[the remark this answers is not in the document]
iiSaLMaN: oh, I thought you meant short test step names.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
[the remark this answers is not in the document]
bdach: I'm pretty sure I triple-checked all call sites before adding this comment.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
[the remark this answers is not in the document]
bdach: Way I see it asserts say *what* is being tested and the step name *why* if it's not immediately obvious in context.

osu.Game.Rulesets.Catch/Objects/Drawables/FruitPiece.cs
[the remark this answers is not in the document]
iiSaLMaN: Fixing this will come separately, resolving from here.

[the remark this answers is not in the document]
iiSaLMaN: Re-resolved in an easier way here.

osu.Game.Rulesets.Catch/Skinning/CatchSkinColour.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    public enum CatchSkinColour
+    {
+        /// <summary>
+        /// The colour to be used for the catcher while on hyper-dashing state.
+        /// </summary>
+        HyperDash,
+
+        /// <summary>
+        /// The colour to be used for hyper-dash fruits.
```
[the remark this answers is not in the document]
iiSaLMaN: That seems a bit incorrect? the colour will be used for the fruits that give the catcher hyper-dashing ability. maybe something like `fruits that grant the catcher the ability to hyper-dash.` or `fruits that make the catcher hyper-dash.`?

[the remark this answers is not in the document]
bdach: See I didn't actually realise that was what the setting was (not described on wiki either). If that's what it does then sure.

osu.Game/Beatmaps/Formats/LegacyDecoder.cs
```diff
@@ -73,6 +73,9 @@ protected virtual void ParseLine(T output, Section section, string line)
             switch (section)
             {
                 case Section.Colours:
+                // osu!catch section only has colour settings
+                // so no harm in handling the entire section
```
bdach, line 77: I'm not sure I agree with this comment... Extracting out a separate decoder like mania did might be overkill but I don't think this should be handled in the very base decoder, either. `LegacySkinDecoder` is probably a better place.

osu.Game.Rulesets.Catch/Skinning/LegacyFruitPiece.cs
```diff
@@ -56,7 +57,7 @@ private void load(DrawableHitObject drawableObject, ISkinSource skin)
                 var hyperDash = new Sprite
                 {
                     Texture = skin.GetTexture(lookupName),
-                    Colour = Color4.Red,
+                    Colour = skin.GetHyperDashFruitColour()?.Value ?? Catcher.DefaultHyperDashColour,
```
bdach, line 60: IMO the extension method should also handle the default.

iiSaLMaN: I wasn't sure at the beginning as I felt like it may not be good for readability (what value will this be fell back to if no setting provides a value), but sure.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -21,6 +21,8 @@ namespace osu.Game.Rulesets.Catch.UI
 {
     public class Catcher : Container, IKeyBindingHandler<CatchAction>
     {
+        public static Color4 DefaultHyperDashColour { get; } = Color4.Red;
```
smoogipoo, line 24: Make all `static Color4`'s in this PR into `static readonly Color4`'s.

smoogipoo (changes requested):
Small nit, otherwise looks fine to me.

osu.Game.Rulesets.Catch/UI/Catcher.cs
iiSaLMaN: that'll rename `DefaultHyperDashColour` to `DEFAULT_HYPER_DASH_COLOUR` is that fine? because that's why I used a static getter-only property just like how `SkinInfo` provides `Default`.

bdach: That's what the convention is, so that's what it will be.

bdach (dismissed):
I think this is ok now 👍

osu.Game.Rulesets.Catch/Skinning/CatchSkinExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+using osu.Game.Rulesets.Catch.UI;
+using osu.Game.Skinning;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Catch.Skinning
+{
+    internal static class CatchSkinExtensions
+    {
+        [NotNull]
+        public static IBindable<Color4> GetHyperDashFruitColour(this ISkin skin)
```
peppy, line 15: this feels over-engineered and i would just inline, but whatever i guess

iiSaLMaN: right actually, things have changed after I dropped changes to the `FruitPiece` component, so this extension may not be of much help now, will inline back.

bdach:
I pushed in a commit which attempts to address the issues with the test code pointed out in the dependent PR.

@peppy would appreciate a look-through to see if this is what you had in mind whenever you have the time 🙂

peppy (approved):
looks good

PR #8448, opened 2020-03-26
Add support for custom hyper-dash catcher colouring

### LINKED ISSUE #8085: LegacySkinDecoder does not support [CatchTheBeat] sections

frenzibyte:
- [x] Depends on #8446 (requires merge conflict resolution)

Closes #8085


osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -28,6 +28,140 @@ namespace osu.Game.Rulesets.Catch.Tests
...
+            });
+
+            AddStep("set hyper-dash", () =>
+            {
+                catcherArea.MovableCatcher.SetHyperDashState(2);
+                catcherArea.MovableCatcher.FinishTransforms();
+            });
+
+            AddAssert("catcher colour default-hyperdash", () => catcherArea.MovableCatcher.Colour == Color4.OrangeRed);
+            AddAssert("catcher trails colour default-hyperdash", () => catcherArea.OfType<Container<CatcherTrailSprite>>().Any(c => c.Colour == Catcher.DefaultHyperDashColour));
```
bdach: Same as the other PR, these step names are not really legible.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -35,7 +37,30 @@ public class Catcher : Container, IKeyBindingHandler<CatchAction>
 
         public Container ExplodingFruitTarget;
 
-        public Container AdditiveTarget;
+        private Container additiveTarget;
+        private Container<CatcherTrailSprite> dashTrails;
+        private Container<CatcherTrailSprite> hyperDashTrails;
+        private Container<CatcherTrailSprite> endGlowSprites;
```
bdach: Not sure I like these being loose and out in the open in the catcher like this. Might make sense to encapsulate in some sort of private subcontainer class to reduce noise in the catcher class.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -237,18 +259,37 @@ public void SetHyperDashState(double modifier = 1, float targetPosition = -1)
...
+                this.FadeColour(hyperDashColour == DefaultHyperDashColour ? Color4.OrangeRed : hyperDashColour, hyper_dash_transition_length, Easing.OutQuint);
+                this.FadeTo(0.2f, hyper_dash_transition_length, Easing.OutQuint);
+            }
+            else
+            {
+                this.FadeColour(Color4.White, hyper_dash_transition_length, Easing.OutQuint);
+                this.FadeTo(1f, hyper_dash_transition_length, Easing.OutQuint);
+            }
+
+            // update hyper-dash colour of the hyper-dashing catcher sprites containers.
```
bdach: Comment basically restates the code.

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -33,10 +33,10 @@ public CatcherArea(BeatmapDifficulty difficulty = null)
...
             RelativeSizeAxes = Axes.X;
             Height = CATCHER_SIZE;
-            Child = MovableCatcher = new Catcher(difficulty)
-            {
-                AdditiveTarget = this,
-            };
+            Child = MovableCatcher = new Catcher(difficulty);
+
+            // this property adds containers to 'this' so it must not be set in the object initializer.
+            MovableCatcher.AdditiveTarget = this;
```
bdach: If it's unsafe to call in the object initialiser it shouldn't be a property at all...? I'd say this needs to be rectified, as it looks like a foot-gun.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -237,18 +259,37 @@ public void SetHyperDashState(double modifier = 1, float targetPosition = -1)
...
+            if (HyperDashing)
+            {
+                this.FadeColour(hyperDashColour == DefaultHyperDashColour ? Color4.OrangeRed : hyperDashColour, hyper_dash_transition_length, Easing.OutQuint);
+                this.FadeTo(0.2f, hyper_dash_transition_length, Easing.OutQuint);
+            }
+            else
+            {
+                this.FadeColour(Color4.White, hyper_dash_transition_length, Easing.OutQuint);
+                this.FadeTo(1f, hyper_dash_transition_length, Easing.OutQuint);
+            }
```
bdach: Did you check what are the implications of no longer checking hyper-dash state on the previous frame and skipping fades if there has been no change, as was done previously? The previous way would avoid redundant colour fades/invalidations if consecutive `updateCatcherColour()` calls were made that would not change value of `HyperDashing`.

bdach (dismissed):

osu.Game.Rulesets.Catch/UI/Catcher.cs
iiSaLMaN: just to be clear, do you mean to put these 3 trails containers inside a private container with public accessibility to all containers inside and access them on all usages?

bdach:
Going to hold off on reviewing this until the prereq is dealt with as due to the merging in-between it's difficult to see what the actual change of this PR is and give feedback on it.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,306 @@
...
+                {
+                    RelativePositionAxes = Axes.None,
+                    Anchor = Anchor.Centre,
+                    Origin = Anchor.Centre,
+                    Scale = new Vector2(4f),
+                }, true, false, false);
+            });
+
+            AddStep("start hyper-dashing", () => catcherArea.MovableCatcher.SetHyperDashState(2));
+            AddAssert("end-glow catcher sprite colour falls back to catcher colour from skin", () => catcherArea.OfType<Container<CatcherTrailSprite>>().Any(c => c.Colour == TestSkin.CustomHyperDashColour));
```
peppy, line 160: these assert texts are too long. should be a few words

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,306 @@
...
+                    Scale = new Vector2(4f),
+                }, customHyperDashCatcherColour, true, false);
+            });
+
+            AddStep("start hyper-dashing", () => catcherArea.MovableCatcher.SetHyperDashState(2));
+            AddAssert("end-glow catcher sprite use its hyper-dash colour from skin", () => catcherArea.OfType<Container<CatcherTrailSprite>>().Any(c => c.Colour == TestSkin.CustomHyperDashAfterColour));
+        }
+
+        [Test]
+        public void TestCustomHyperDashCatcherEndGlowColourFallback()
```
peppy, line 144: these are too long. remove unnecessary words (the class already asserts most of this, and "custom" is redundant since you already say "fallback")

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,306 @@
...
+
+            public Texture GetTexture(string componentName) => null;
+
+            public SampleChannel GetSample(ISampleInfo sampleInfo) => null;
+
+            public IBindable<TValue> GetConfig<TLookup, TValue>(TLookup lookup)
+            {
+                if (lookup is CatchSkinColour config)
+                {
+                    if (config == CatchSkinColour.HyperDash && customCatcherColour)
```
peppy, line 292: why nt switch

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -0,0 +1,306 @@
...
+        {
+            public static Color4 CustomHyperDashColour { get; } = Color4.Goldenrod;
+            public static Color4 CustomHyperDashFruitColour { get; } = Color4.Cyan;
+            public static Color4 CustomHyperDashAfterColour { get; } = Color4.Lime;
+
+            private readonly bool customCatcherColour;
+            private readonly bool customAfterColour;
+            private readonly bool customFruitColour;
+
+            public TestSkin(bool customCatcherColour, bool customAfterColour, bool customFruitColour)
```
peppy, line 275: bool bool bool is so hard to read. why not just make the custom color4 properties nullable and settable and set them in the tests? at least the tests would be legible then.

peppy (changes requested):

peppy:
review applies to previous pr i guess. not going to repost there

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
iiSaLMaN: So the assert names used to be like: `end-glow sprite colour default-hyperdash`, but I was change-requested to make it more described as the assert name are mostly for describing the assertion on CI in case it fails, although I agree that it is overly-described and it is my mistake not pointing that out in the request conversation.

peppy: "colour is correct" is fine. if you really need to explain more that isn't obvious from the test method's name, add an inline comment.

bdach:
Tried to see if I can merge master back in here but there's been so many changes it's nigh unmergeable. It would likely be easier to just rewrite this PR off master at this point.

iiSaLMaN:
> Tried to see if I can merge master back in here but there's been so many changes it's nigh unmergeable. It would likely be easier to just rewrite this PR off master at this point.

It was actually just a conflicting of the new written `TestSceneHyperDashColouring` so I fixed it, should be good now.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -13,16 +13,26 @@
...
 
 namespace osu.Game.Rulesets.Catch.UI
 {
-    public class Catcher : Container, IKeyBindingHandler<CatchAction>
+    public class Catcher : SkinReloadableDrawable, IKeyBindingHandler<CatchAction>
     {
         public static readonly Color4 DEFAULT_HYPER_DASH_COLOUR = Color4.Red;
 
+        /// <summary>
+        /// The default colour used directly for this <see cref="Catcher"/>'s <see cref="Drawable.Colour"/>.
```
peppy, line 29: this doesn't read well. also please document the above colour if you're going to do so here.

osu.Game.Rulesets.Catch/UI/Catcher.cs
peppy: I think this is fine.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +155,26 @@ private void load()
             updateCatcher();
         }
 
+        /// <summary>
+        /// Sets container target to provide catcher additive trails content in.
+        /// </summary>
+        /// <param name="target">The container to add catcher trails in.</param>
+        public void SetAdditiveTarget(Container target)
```
peppy, line 162: just pass the target into the `ctor` and remove all of this.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +155,26 @@ private void load()
...
+        {
+            if (additiveTarget == target)
+                return;
+
+            additiveTarget?.RemoveRange(new[] { dashTrails, hyperDashTrails, endGlowSprites });
+
+            additiveTarget = target;
+            additiveTarget?.AddRange(new[]
+            {
+                dashTrails ??= new Container<CatcherTrailSprite> { RelativeSizeAxes = Axes.Both, Colour = Color4.White },
```
peppy, line 172:
```suggestion
                dashTrails ??= new Container<CatcherTrailSprite> { RelativeSizeAxes = Axes.Both },
```

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +155,26 @@ private void load()
...
+                hyperDashTrails ??= new Container<CatcherTrailSprite> { RelativeSizeAxes = Axes.Both, Colour = hyperDashColour },
+                endGlowSprites ??= new Container<CatcherTrailSprite> { RelativeSizeAxes = Axes.Both, Colour = hyperDashEndGlowColour },
```
peppy, line 174: initialise all of this in the ctor or bdl

peppy (changes requested):

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -68,6 +110,40 @@ public void TestFruitColourFallback()
...
+                catcherArea.MovableCatcher.FinishTransforms();
+            });
+
+            AddAssert("catcher colour is correct", () =>
+                expectedCatcherColour == Catcher.DEFAULT_HYPER_DASH_COLOUR
+                    ? catcherArea.MovableCatcher.Colour == Catcher.DEFAULT_CATCHER_HYPER_DASH_COLOUR
+                    : catcherArea.MovableCatcher.Colour == expectedCatcherColour);
+
+            AddAssert("catcher trails colours are correct", () => catcherArea.OfType<Container<CatcherTrailSprite>>().Any(c => c.Colour == expectedCatcherColour));
+            AddAssert("catcher end-glow colours are correct", () => catcherArea.OfType<Container<CatcherTrailSprite>>().Any(c => c.Colour == (expectedEndGlowColour ?? expectedCatcherColour)));
```
peppy, line 136: this is completely wrong

osu.Game.Rulesets.Catch/UI/Catcher.cs
bdach: Not sure if this is me being dense again but I still kind of don't get what this particular xmldoc is trying to say... In its current state it could be interpreted in a way that makes it misleading (as in, that this colour applies to the whole catcher, which is not *technically* the case - it applies only in hyper-dash state, even if we use `Colour` for that due to additive blending).

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
iiSaLMaN: I'm not quite sure how to test this correctly without directly exposing the containers...

I would say applying what @bdach suggested regarding grouping all trail containers in one `CatcherTrailsContainer` may help this by doing:
```csharp
catcherArea.OfType<CatcherTrailsContainer>().Single().HyperDashTrails == expectedColour
```

peppy: Try it and see how it feels. Could work.

iiSaLMaN: After creating a public `CatcherTrailsDisplay` grouping all trails, I've decided to also move the trail-related logic into that component now that there's a chance to, for simplifying `Catcher`.

osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
```diff
@@ -23,7 +24,7 @@ public class TestSceneCatcher : CatchSkinnableTestScene
         [BackgroundDependencyLoader]
         private void load()
         {
-            SetContents(() => new Catcher
+            SetContents(() => new Catcher(new Container())
```
peppy, line 27: as mentioned in discord, this container should be displayed in the test scene. you've regressed the test as it stands.

iiSaLMaN: This test used to provide `null` for `additiveTarget` so I thought there won't be a problem with providing a dummy `Container`? 

Moreover, providing a real `additiveTarget` here sounds exactly like what `CatcherArea` provides, which has an own test scene (`TestSceneCatcherArea`), why not just remove this test scene if it's worthless.

peppy: i guess it's fine for this case, as input is also not handled here.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -68,6 +110,42 @@ public void TestFruitColourFallback()
...
+                Child = setupSkinHierarchy(catcherArea = new CatcherArea
+                {
+                    Anchor = Anchor.Centre,
+                    Origin = Anchor.Centre,
+                    Scale = new Vector2(4f),
+                }, skin);
+
+                trails = catcherArea.OfType<CatcherTrailDisplay>().Single();
+                catcherArea.MovableCatcher.SetHyperDashState(2);
+                catcherArea.MovableCatcher.FinishTransforms();
```
peppy, line 129: can this be replaced with an until step?

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -68,6 +110,42 @@ public void TestFruitColourFallback()
...
+            });
+
+            AddAssert("catcher colour is correct", () =>
+                expectedCatcherColour == Catcher.DEFAULT_HYPER_DASH_COLOUR
```
peppy, line 133: i don't understand this ternary conditional.

peppy (changes requested):

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -3,26 +3,45 @@
...
     {
+        /// <summary>
+        /// The default colour used for all hyper-dashing components. (catcher drawables and fruit)
+        /// </summary>
         public static readonly Color4 DEFAULT_HYPER_DASH_COLOUR = Color4.Red;
 
+        /// <summary>
+        /// The default hyper-dash colour used directly for this
+        /// <see cref="Catcher"/>'s <see cref="Drawable.Colour"/>.
+        /// </summary>
```
bdach, line 34: I still think this xmldoc is misleading (if not unreadable - I still don't quite understand what it's trying to say), and more importantly, it directly contradicts the xmldoc of `DEFAULT_HYPER_DASH_COLOUR` immediately preceding it. I've mentioned this once and I'm pretty sure I unresolved the relevant conversation, but the conversation is resolved again and I see no actual changes to the content of the xmldoc, save for insertion of one new line.

Please don't resolve conversations without addressing them in some shape or form.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -264,20 +276,37 @@ public void SetHyperDashState(double modifier = 1, float targetPosition = -1)
...
 
+        private void updateCatcherColour(bool hyperDashing)
+        {
+            if (hyperDashing)
+            {
+                // special behaviour for catcher colour if no skin overrides.
+                var catcherColour =
+                    hyperDashColour == DEFAULT_HYPER_DASH_COLOUR
+                        ? DEFAULT_CATCHER_HYPER_DASH_COLOUR
+                        : hyperDashColour;
```
bdach, line 295: I find this behaviour of silently overriding the colour from `Red` to `OrangeRed` really confusing. It complicates everything (including test code) and leads to unexpected behaviour (the user setting a red colour is now indistinguishable from the default - in both cases this colour will get silently converted to `OrangeRed`). I propose we just get rid of it for simplicity.

iiSaLMaN: I agree, I've tried comparing them with each other, the `Red` colour made the sprite more red-er than `OrangeRed`, but since the catcher is small I don't think it'll be that notable.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
```diff
@@ -130,9 +130,16 @@ private void checkHyperDashCatcherColour(ISkin skin, Color4 expectedCatcherColou
...
-                expectedCatcherColour == Catcher.DEFAULT_HYPER_DASH_COLOUR
-                    ? catcherArea.MovableCatcher.Colour == Catcher.DEFAULT_CATCHER_HYPER_DASH_COLOUR
-                    : catcherArea.MovableCatcher.Colour == expectedCatcherColour);
+            {
+                var expected = expectedCatcherColour;
+
+                if (expected == Catcher.DEFAULT_HYPER_DASH_COLOUR)
+                    // The expected colour for Catcher.Colour is another colour
+                    // for the default skin, assert with that instead.
+                    expected = Catcher.DEFAULT_CATCHER_HYPER_DASH_COLOUR;
```
peppy, line 138: this still does not make any sense to me, sorry. at this point you should be splitting out tests rather than doing this weird logic.

- what is "another colour" and why?
- why not just set it correctly to `expectedCatcherColour`?

peppy (changes requested):

osu.Game.Rulesets.Catch/UI/Catcher.cs
peppy: i'm fine with just removing it if you can't figure out how to write tests. your call.

iiSaLMaN: The only 2 sane ways I can provide to implement this would be:
 1. Do some type of calculation to get a colour that can blend with the trails similar to how `Color4.Red` blends with `Color4.OrangeRed(alpha: 0.2f)`, sounds nearly impossible or just way too complicated to even be done correctly, *not even sure to call it a sane way*.
2. Rename `hyperDashColour` to `customHyperDashColour` and let it fallback to `DEFAULT_CATCHER_HYPER_DASH_COLOUR` on null here, that's a bit more readable and will solve the issue @bdach stated above, but the effect will only come here. as for the tests, I'll replay in the below conversation about it.

If nothing here sounds sane to you then I honestly don't know, may be that I can't really be more helpful in this specific case or that this is just way too complicated to be simplified.

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
iiSaLMaN: What do you mean by "splitting out tests", might be what I'm gonna suggest here:

I guess the best this can become would be by splitting `expectedCatcherColour` to `expectedCatcherColour, expectedTrailsColour`, and fill in accordingly.

osu.Game.Rulesets.Catch/UI/Catcher.cs
iiSaLMaN:
```suggestion
        /// <summary>
        /// The default color used for the direct <see cref="Drawable.Colour"/> of the <see cref="Catcher"/> while in hyper-dashing state.
        /// </summary>
        /// <remarks>
        /// This colour is used specially in here on the default skin only.
        /// </remarks>
```

What do you think?

osu.Game.Rulesets.Catch.Tests/TestSceneHyperDashColouring.cs
bdach: Basically if your test assertions have conditionals in them it means it's time to split the test into two, each of which tests each branch of the conditional assertion - so you'd test the `DEFAULT_HYPER_DASH_COLOUR` case separately and the case of a custom non-red colour separately.

That said I'd still be in favour of just getting rid of that special case.

osu.Game.Rulesets.Catch/UI/Catcher.cs
bdach: No, I still don't understand what that xmldoc is trying to say. What is a "*direct colour*"? You're still not describing what the colours are actually *colouring*.

If I had to have a go at documenting these constants somewhat accurately (from what I can tell from a quick look), I'd probably write something along the lines of

```csharp
/// <summary>
/// The default colour used to tint hyper-dash fruit, the catcher's trail and its after-image during a hyper-dash.
/// </summary>
public static readonly Color4 DEFAULT_HYPER_DASH_COLOUR = Color4.Red;

/// <summary>
/// The default colour used to tint the moving catcher sprite, along with the fruits on the catcher plate, during a hyper-dash.
/// </summary>
public static readonly Color4 DEFAULT_CATCHER_HYPER_DASH_COLOUR = Color4.OrangeRed;
```

which to me shows quite well how arbitrary and weird this distinction has become (unless I'm wrong about this description, but then it shows that this colour rigamarole is way too complicated).

bdach:
I went ahead and removed the `DEFAULT_CATCHER_HYPER_DASH_COLOUR` thing entirely myself. It feels like an okay thing to do as it simplifies matters significantly while having barely noticeable (at least to me) visual implications.

The rest of this seems relatively sane as far as I can tell.

osu.Game.Rulesets.Catch/UI/CatcherTrailDisplay.cs
```diff
@@ -0,0 +1,138 @@
...
+
+            endGlow.MoveToOffset(new Vector2(0, -10), 1200, Easing.In);
+            endGlow.ScaleTo(endGlow.Scale * 0.95f).ScaleTo(endGlow.Scale * 1.2f, 1200, Easing.In);
+            endGlow.FadeOut(1200);
+            endGlow.Expire(true);
+        }
+
+        private void displayTrail()
+        {
+            if (!catcher.Dashing && !catcher.HyperDashing)
```
peppy, line 106: I find it super weird that `Catcher` is responsible for starting the trail and this new class is responsible for stopping it...

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -44,33 +58,23 @@ public class Catcher : Container, IKeyBindingHandler<CatchAction>
...
-            get => trail;
-            set
-            {
-                if (value == trail || AdditiveTarget == null) return;
-
-                trail = value;
-
-                if (Trail)
-                    beginTrail();
+                trails.DisplayTrail = value || HyperDashing;
```
peppy, line 77: this is horrible logic. you can't arbitrarily or in a value in a set property. the chances of this breaking are 99.9% (it is not going to be applied unless `Dashing` is set, but there's a precondition equality check potentially causing the set to be a no-op).

bdach: Would you mind having a look at the commits I just pushed in here and seeing if it improves things? I'm not 100% sure it does but after looking at this code for the better part of an hour I decided to avoid any bigger changes for now.

peppy (approved):
seems sane enough now

PR #8450, opened 2020-03-26
Reduce spread of stacked fruit

peppy:
Just eyeballing this one.

image


bdach (approved):
Nothing to discuss here code-wise, personally think it's an improvement in visibility but not a catch player so my opinion has little value. 👍

smoogipoo (approved):

PR #8451, opened 2020-03-26
Only play slider end sounds if tracking

### LINKED ISSUE #8437: Hit sounds are always played on slider end
**Describe the bug:** 
Hit sounds are always played on slider end even without clicking circles or following sliders.
And in the editor, Only a hit sound of slider ends is played. Others are not played.

Reuploaded due to recording mistake. Sorry! (Forgot to disable my mic)
https://streamable.com/wgrka

**osu!lazer version:** 
1587d4b26fbad691242544a62dbf017a78705ae3
Local Build with dotnet 3.1.102

**Logs:**
[network.log](https://github.com/ppy/osu/files/4382185/network.log)
[performance.log](https://github.com/ppy/osu/files/4382186/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4382187/runtime.log)

peppy:
Regressed with recent slider changes.

- Closes https://github.com/ppy/osu/issues/8437#issuecomment-604293595

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -11,6 +11,7 @@ using osu.Framework.Bindables;
 using osu.Framework.Graphics.Containers;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Osu.Skinning;
+using osu.Game.Rulesets.Scoring;
 using osuTK.Graphics;
 using osu.Game.Skinning;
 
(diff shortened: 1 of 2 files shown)


bdach (commented):
Not the ideal solution but I think I can see why it's necessary & have no better alternative to suggest, so I'd say this is fine for now. Definitely should try to get rid of it at some point though if possible.

Wondering if tests for this would be overkill? I suppose `PlaySamples` *is* now virtual, so you technically could...

peppy:
I don't think it's worth testing until we finalise how we want sliders to work. This is just a step (hopefully) in the right direction, alongside https://github.com/ppy/osu/pull/8452, but definitely not the last changes.

We've already discussed that we're missing a lot of testing on gameplay scenarios, so will probably take some time to update/add those after done.

smoogipoo (approved):
not super sure about this one

PR #8454, opened 2020-03-26
Update overlay header elements to match osu-web

### LINKED ISSUE #8423: ScreenTitle class needs to be rethought to implement the new design
Right now the`ScreenTitle` is used almost exclusively in `FullscreenOverlays` mimicking their web counterparts, the exception being the multiplayer screen - it's the only place where `ScreenTitle` is used for purposes other than matching the web design. 

As of a while ago the web design has been changed to a much simpler one, which can't really be shared with the multi screen as a whole anymore.

I'm working on implementing the new design and the question is: what should I do with the `MultiHeaderTitle`? It looks rather outdated and I'm sure it's due to be replaced by something newer. The new overlay title design can still share specific parts with it, though, so the question is, what would be better:

- removing `ScreenTitle`, completely moving the old design as-is to `MultiHeaderTitle` and making the new one into something like `OverlayScreenTitle` (in the `Overlays` namespace, where it really belongs) - this would cause some code duplication (for the icon + title text logic, the multi header would add its separator + section name in its own class), but I think it'd be acceptable since it'd allow for proper separation

or: 

- implementing the new design in the base `ScreenTitle`, leaving it where it is right now, only having the icon and title there (as needed by the new design) and adding the old separator + section name to the `MultiHeaderTitle` deriving from it. This would slightly change the old design (mostly spacing) and is a bit less clean IMO, since it still ties two kinda unrelated components together, but doesn't cause as much code duplication.

thewildtree:
### Summary

Updates `OverlayHeader` and its components to match the new osu-web design. 

Supersedes #7993, closes #8423, moves old `ScreenTitle` away to `MultiHeaderTitle` and add new `OverlayTitle` for use in overlays as discussed.

### Screenshots:
osu-web vs. this PR:

image
image

It will visually look the same once the font sizing (and weighting) fix eventually comes. After temporarily changing font sizes and weights to visually match and overlaying them in a graphics editor, the header parts overlapped 1:1. 

Colours differ for reasons described below.

### Remarks:

- old `ScreenTitle` icon creating mechanism got simplified a bit: `OverlayTitle` only accepts the icon texture name to display (since there don't seem to be any other ways needed) while `MultiHeaderTitle` simply uses a `SpriteIcon`.

- `OverlayTitle` has had its `Section` removed since the new design doesn't use it anymore.

- this doesn't touch `OverlayTabControl`'s colours, as that's been discussed in previous PR and needs to be refactored before changing anything.

- `OverlayTabControl`'s `BarHeight` had to be somehow coupled with its tabs' internal `ExpandingBar` margin for it to be properly aligned - decided to just use a `const` for that, I also tried a `Bindable` variant but it turned out to be a bit too complicated for such a simple change.

- `OverlayHeaderBreadcrumbControl` accent colour propagation is actually broken - you can see the `OsuTabItem` default blue popping up for newly created tabs, it's just covered up by the active tab colour fade. This would be instantly noticeable if tabs didn't have their colour changed to white upon activation. There is a band-aid fix for that in other overlay tab controls, but I decided against adding it here, since the whole `AccentColour` thing is due to be reworked in the near future. It does look a bit different from web (active tabs dont change their colour there) but I think it's okay for now.


peppy (approved):
I think this is good for now.

PR #8455, opened 2020-03-26
Fix NullReferenceException when starting the no-video version of a beatmap with video

Game4all:
Fixes potential NRE when playing the downloaded without video version of a beatmap with a video file.

Can be reproduced with any beatmap that has a video by downloading the no video version of the beatmap, importing it and start it.

### DIFF
--- a/osu.Game.Tests/Resources/storyboard_no_video.osu
+++ b/osu.Game.Tests/Resources/storyboard_no_video.osu
@@ -1,0 +1,31 @@
+osu file format v14
+
+[Events]
+//Background and Video events
+0,0,"BG.jpg",0,0
+Video,0,"video.avi"
+//Break Periods
+//Storyboard Layer 0 (Background)
+//Storyboard Layer 1 (Fail)
+//Storyboard Layer 2 (Pass)
+//Storyboard Layer 3 (Foreground)
+//Storyboard Layer 4 (Overlay)
+//Storyboard Sound Samples
+
+[TimingPoints]
+1674,333.333333333333,4,2,1,70,1,0
+1674,-100,4,2,1,70,0,0
+3340,-100,4,2,1,70,0,0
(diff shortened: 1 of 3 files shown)


bdach (commented):
Fix looks fine. It's a hard crash so this is pretty high priority.

Wondering however if we couldn't have an actual test for this. I don't think it should be too difficult to write (just try to display a storyboard from an `.osr` with a video event linking to a file that doesn't exist).

As an aside no video maps also have another regression in that the background image seems to no longer show (it's all black). That's another PR entirely, though.

peppy:
Really do need testing on this. Should have been there in the last PR to avoid this happening in the first place 🤕

Game4all:
Will come back with tests for this to ensure no further regresses

peppy:
@Game4all got an ETA on this? would like to get this fix out in a build today or tomorrow.

Game4all:
Here is my shot at writing a test case for this. It feels like it doesn't test for anything as it just ensures the storyboard throws no NRE and feels a bit hacky

osu.Game.Tests/Resources/storyboard_no_video.osu
```diff
@@ -0,0 +1,487 @@
...
+CircleSize:4
+OverallDifficulty:9
+ApproachRate:9
+SliderMultiplier:1.7
+SliderTickRate:1
+
+[Events]
+//Background and Video events
+0,0,"BG.jpg",0,0
+Video,0,"nevo cute.avi"
```
smoogipoo: We try to not copy full beatmaps for copyright reasons. Most of the contents of this .osu can be culled (just keep two hitobjects or something + the video). Likewise, remove the things above (name/creator/artist/etc).

smoogipoo (changes requested):

peppy (approved):

PR #8462, opened 2020-03-27
Fix animated follow points not (re)animating after rewind

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3407
- [x] Depends on https://github.com/ppy/osu-framework/pull/3405
- Reverts optimisations in https://github.com/ppy/osu/pull/8346 for simplicity. This only seemed to affect very specific maps (like centipede) which can wait until we fix the issue properly.

This also fixes the animations potentially not starting at the correct point in time. It's designed to scale with skinning and be usable by both legacy and non-legacy components.


smoogipoo (approved):
Seems fine for now, let's see where this takes us.

PR #8470, opened 2020-03-27
Make beatmap carousel use queue for its selection of next beatmap

### LINKED ISSUE #3051: choosing beatmap with lef/right arrows is broken

LittleEndu:
Resolves #3051

This PR splits ``BeatmapCarousel``'s ``SelectNext`` function into public and private ones. Public function will enqueue the request to be called later in ``Update``. Test is included.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -87,8 +87,9 @@ namespace osu.Game.Tests.Visual.SongSelect
         public void TestTraversalHold()
         {
             var sets = new List<BeatmapSetInfo>();
+            const int create_this_many_sets = 200;
 
-            for (int i = 0; i < 20; i++)
+            for (int i = 0; i < create_this_many_sets; i++)
             {
                 var set = createTestBeatmapSet(i);
                 sets.Add(set);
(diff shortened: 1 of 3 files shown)


LittleEndu:
```cs
            if (selectedBeatmap != null && (drawable = selectedBeatmap.Drawables.FirstOrDefault()) == null)
                // if the selected beatmap isn't present yet, we can't correctly change selection.
                // we can fix this by changing this method to not reference drawables / Items in the first place.
                return;
```

This could now just re-enqueue the request instead of ignoring it. Not sure if this is even needed as the code is from #1787 which is 2 years old now.

peppy:
This seems like a very convoluted solution and fails to describe how the issue was occurring in the first place. Could you please provide some reasoning behind the change and explain what the perceived issue was so we can potentially give direction for a better solution?

LittleEndu:
I have no idea what causes this error. But I just rewrote the selection logic to not use drawables, something that the comment I mentioned says we should do. Will PR that later. Closing this PR as this is indeed kinda convoluted.

PR #8471, opened 2020-03-27
Implement OverlayScrollContainer component

EVAST9919:
Split from https://github.com/ppy/osu/pull/7972 as suggested [here](https://github.com/ppy/osu/pull/7972#pullrequestreview-378930592)


smoogipoo:
The tooltip doesn't display in the test?

EVAST9919:
> The tooltip doesn't display in the test?

For some reason it's not visible in `OsuManualInputManagerTestScene`. It's visible in game and in `OsuTestScene`.

peppy:
Added a failing test

EVAST9919:
Fixed

peppy:
Why not just block the action itself, or disable the button? Also you called it "AreaState" even though there's nothing in your whole class called "Area"

I'm finding it really hard to follow since you have a variable called Button which is a button that has a button inside of it, where the inner button binds to the outer button's state in order to manage input.

EVAST9919:
Removed nested classes and other overcomplicated stuff.

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            currentTarget = Target;
+            Button.Current.Value = Current > button_scroll_position ? Visibility.Visible : Visibility.Hidden;
+        }
+
+        public class ScrollToTopButton : OsuHoverContainer, IHasCurrentValue<Visibility>
+        {
+            private const int fade_duration = 500;
+
+            private readonly BindableWithCurrent<Visibility> current = new BindableWithCurrent<Visibility>();
```
peppy, line 72: what's the point of this?

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,160 @@
...
+
+            public Bindable<Visibility> Current
```
peppy, line 74: remove both bindable and just use a local backed property

```csharp
        private Visibility state;

        public Visibility State
        {
            get => state;
            set
            {
                if (value == state)
                    return;

                state = value;
                // do thing here          
            }
        }
```

peppy (changes requested):

osu.Game/Overlays/OverlayScrollContainer.cs
EVAST9919: Done

peppy:
image

EVAST9919:
So I guess you want me to totally disable dragging in this area?

peppy:
The button isn't displaying.

peppy:
Please add test coverage for the failing case.

EVAST9919:
Added

osu.Game.Tests/Visual/UserInterface/TestSceneOverlayScrollContainer.cs
```diff
@@ -63,7 +63,7 @@ public void TestButtonVisibility()
             AddStep("scroll to start", () => scroll.ScrollToStart(false));
             AddAssert("button is hidden", () => scroll.Button.State == Visibility.Hidden);
 
-            AddStep("scroll to 250", () => scroll.ScrollTo(500));
+            AddStep("scroll to 500", () => scroll.ScrollTo(500));
             AddUntilStep("scrolled back to start", () => Precision.AlmostEquals(scroll.Current, 500, 0.1f));
```
peppy, line 67: while you're fixing typos, this is also wrong.

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,156 @@
...
+    /// <see cref="OsuScrollContainer"/> which provides <see cref="ScrollToTopButton"/>. Mostly used in <see cref="FullscreenOverlay"/>.
+    /// </summary>
+    public class OverlayScrollContainer : OsuScrollContainer
+    {
+        /// <summary>
+        /// Scroll position at which the <see cref="ScrollToTopButton"/> will be shown.
+        /// </summary>
+        private const int button_scroll_position = 200;
+
+        public ScrollToTopButton Button { get; }
```
smoogipoo, line 29: This should not be exposed publicly. Protected at most, and derived for testing purposes.

Also make `readonly` rather than `{ get; }`.

osu.Game/Overlays/OverlayScrollContainer.cs
```diff
@@ -0,0 +1,156 @@
...
+        {
+            base.UpdateAfterChildren();
+
+            if (ScrollContent.DrawHeight + button_scroll_position < DrawHeight)
+            {
+                Button.State = Visibility.Hidden;
+                return;
+            }
+
+            if (Target == currentTarget)
```
smoogipoo, line 59: What's the reason for this `currentTarget` variable? As far as I can see it has no real purpose other than to maybe reduce amount of times the property is set? Prefer removal if that's the case.

smoogipoo (changes requested):
Seems ok otherwise.

osu.Game/Overlays/OverlayScrollContainer.cs
EVAST9919: Clicking on button should cause it's immediately disappearance. Removing that will cause it only if we are close enugh to the start.

EVAST9919: And if we will leave everything as-is - setting button to Hidden state will be reverted on the next frame

peppy: at very least that should be commented, since it is not immediately obvious.

vLixX:
Is osu! lazer available on ios atm?

osu.Game/Overlays/OverlayScrollContainer.cs
smoogipoo: If that's true, it would already be the case. Because what you're saying implies `Target` is changing, which if it is will pass through that condition and cause the further comparison.

But that's not how `Target` works - it doesn't change all the time.

Am I misunderstanding?

peppy: Tests pass with the early exit present or not. @EVAST9919 If it is required, please add comments and test coverage. I think you will require it's not and is left-over from a previous solution.

peppy (approved):

smoogipoo (approved):
I think this looks fine now

PR #8447, opened 2020-03-26
Fix break overlay scaling with gameplay

peppy:
- Closes https://github.com/ppy/osu/issues/7702.

PR #8474, opened 2020-03-27
Fix break overlay not displaying progress information

### LINKED ISSUE #8469: No progress information displayed during breaks
**Describe the bug:** 
Self-explanatory. Looks like today's patch broke beatmap pause scores, and now they do not display at all (everything was normal yesterday). Restarting the game did not fix the issue.

Screenshot from 2020-03-27 17-23-57

**osu!lazer version:** 
2020.327.0

bdach:
Resolves #8469.

# Summary

Progress information display in break overlay has regressed in #8447 due to what seems to be a simple oversight of not passing `ScoreProcessor` to the break overlay.

# Remarks

Added a verification test step in `TestSceneAutoplay` to try and make sure this doesn't happen again. It kind of tests that autoplay works fine too, so two birds with one stone I guess?

As an aside this kind of bug is why I sort of dislike default argument values. Easy to forget that the argument is there as things just compile.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneAutoplay.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneAutoplay.cs
@@ -6,6 +6,7 @@ using System.Linq;
 using osu.Framework.Testing;
 using osu.Game.Rulesets;
 using osu.Game.Screens.Play;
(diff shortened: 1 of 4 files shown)


peppy:
we should probably make this a non-optional argument

peppy (approved):

PR #8475, opened 2020-03-27
Fix break overlay displaying in front of all other player overlays

### LINKED ISSUE #8473: Break overlay shows in front of pause menu
**Describe the bug:** 
self explanatory

Screen Shot 2020-03-27 at 3 25 28 PM

**osu!lazer version:** 
2020.327.0

bdach:
Resolves #8473.

- [x] Depends on #8474 (for mergeability).

# Summary

Another regression from #8447. Moving the break overlay from `DrawableRuleset`'s overlays to `Player` itself has affected z-order of all overlays, as the break overlay (which was added as last in that PR) then ended up the topmost overlay. This means the change also affected i.e. the cursor container, as can be seen on the following screenshots:

| `master` | this PR |
| :-: | :-: |
| osu_2020-03-27_22-26-01 | osu_2020-03-27_22-28-57 |

It's subtle, but on the first screenshot the break overlay is in front of the gameplay cursor.

To resolve, make `BreakOverlay` be the first drawable in `addOverlayComponents`. Since before the overlay was the topmost component in `DrawableRuleset`, this mostly restores behaviour, with the exception of `ComboEffects` (which used to be above, and is now below the break overlay).

# Remarks

No tests here, as I couldn't really figure out a good way to test this that wouldn't be too brittle in regards to potential future changes. Suggestions on best way to test however welcome as I'd like to include one here if possible.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -251,6 +251,12 @@ namespace osu.Game.Screens.Play
         {
             target.AddRange(new[]
             {
+                BreakOverlay = new BreakOverlay(working.Beatmap.BeatmapInfo.LetterboxInBreaks, ScoreProcessor)
+                {
+                    Clock = DrawableRuleset.FrameStableClock,
+                    ProcessCustomClock = false,
+                    Breaks = working.Beatmap.Breaks
+                },
                 // display the cursor above some HUD elements.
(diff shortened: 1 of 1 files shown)


peppy:
Huge oversight in my PR - I did have it in the place you put it, but went through four iterations of how to implement the change and ended up with it in the wrong place.

peppy (approved):

PR #8477, opened 2020-03-28
Add support for legacy skin sliderstartcircle / sliderstartcircleoverlay

peppy:


smoogipoo:
Want to add support for `sliderendcircle` too, while you're here?

smoogipoo (approved):

peppy:
`sliderendcircle` requires a bit more implementation. it's non-trivial as the drawable doesn't currently track correctly, nor display/animate.

PR #8481, opened 2020-03-28
Rewrite beatmap carousel's select next logic to not use drawables

### LINKED ISSUE #3051: choosing beatmap with lef/right arrows is broken

LittleEndu:
Resolves #3051

Before this PR, carousel's select next logic relied on drawables being present and being in a good state (unsure what that state was, but if select next was called more than once in an update frame, carousel's selection would reset).

This PR makes select next use ``CarouselItem`` instead of ``DrawableCarouselItem``. Also removes unnecessary while loop logic by splitting select next into multiple functions.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -85,66 +85,47 @@ namespace osu.Game.Tests.Visual.SongSelect
 
         [TestCase(true)]
         [TestCase(false)]
-        public void TestTraversalHold(bool forwards)
+        public void TestTraversalBeyondVisible(bool forwards)
         {
             var sets = new List<BeatmapSetInfo>();
-            const int create_this_many_sets = 200;
 
-            for (int i = 0; i < create_this_many_sets; i++)
-            {
-                var set = createTestBeatmapSet(i + 1);
-                sets.Add(set);
-            }
+            const int total_set_count = 200;
+
+            for (int i = 0; i < total_set_count; i++)
+                sets.Add(createTestBeatmapSet(i + 1));
 
             loadBeatmaps(sets);
 
+            for (int i = 1; i < total_set_count; i += i)
+                selectNextAndAssert(i);
+
             void selectNextAndAssert(int amount)
             {
-                setSelected(forwards ? 1 : create_this_many_sets, 1);
-                string text = forwards ? "Next" : "Previous";
-                AddStep($"{text} beatmap {amount} times", () =>
+                setSelected(forwards ? 1 : total_set_count, 1);
+
+                AddStep($"{(forwards ? "Next" : "Previous")} beatmap {amount} times", () =>
                 {
                     for (int i = 0; i < amount; i++)
                     {
                         carousel.SelectNext(forwards ? 1 : -1);
                     }
                 });
-                waitForSelection(forwards ? amount + 1 : create_this_many_sets - amount);
-            }
 
-            for (int i = 1; i < create_this_many_sets; i += i)
-            {
-                selectNextAndAssert(i);
+                waitForSelection(forwards ? amount + 1 : total_set_count - amount);
             }
         }
 
         [Test]
-        public void TestTraversalHoldDifficulties()
+        public void TestTraversalBeyondVisibleDifficulties()
         {
             var sets = new List<BeatmapSetInfo>();
 
-            for (int i = 0; i < 20; i++)
-            {
-                var set = createTestBeatmapSet(i + 1);
-                sets.Add(set);
-            }
+            const int total_set_count = 200;
+
+            for (int i = 0; i < total_set_count; i++)
+                sets.Add(createTestBeatmapSet(i + 1));
 
             loadBeatmaps(sets);
-
-            void selectNextAndAssert(int amount, bool forwards, int expectedSet, int expectedDiff)
-            {
-                // Select very first or very last difficulty
-                setSelected(forwards ? 1 : 20, forwards ? 1 : 3);
-                string text = forwards ? "Next" : "Previous";
-                AddStep($"{text} difficulty {amount} times", () =>
-                {
-                    for (int i = 0; i < amount; i++)
-                    {
-                        carousel.SelectNext(forwards ? 1 : -1, false);
-                    }
-                });
-                waitForSelection(expectedSet, expectedDiff);
-            }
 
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -253,46 +253,35 @@ public bool SelectBeatmap(BeatmapInfo beatmap, bool bypassFilters = true)
...
-
-            if (!visibleItems.Any())
-                return;
-
-            DrawableCarouselItem drawable = null;
-
-            if (selectedBeatmap != null && (drawable = selectedBeatmap.Drawables.FirstOrDefault()) == null)
-                // if the selected beatmap isn't present yet, we can't correctly change selection.
-                // we can fix this by changing this method to not reference drawables / Items in the first place.
+            if (!beatmapSets.Where(s => !s.Filtered.Value).ToList().Any())
```
```suggestion
            if (beatmapSets.All(s => !s.Filtered.Value))
```
The `ToList()` is gratuitous and adds an overhead of an additional list allocation.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -83,6 +83,41 @@ public void TestTraversal()
...
+                        carousel.SelectNext(forwards ? 1 : -1);
+                    }
+                });
+                waitForSelection(forwards ? amount + 1 : create_this_many_sets - amount);
+            }
+
+            for (int i = 1; i < create_this_many_sets; i += i)
+            {
+                selectNextAndAssert(i);
+            }
```
bdach: Not sure why this way was chosen to test this but I suppose it works...?

bdach (changes requested):
Looks mostly correct, save for some slight nitpicks. However, tests should also cover the `skipDifficulties = false` path IMO.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
LittleEndu: #8470 (PR before this PR) had an optimization that would fail over different amounts of select next called. And old behaviour that relied on drawables also would fail over different amounts (and for some reason would always fail after you had simultaneously selected at least 7 times). Not sure if it will catch any regressions with the new logic tho.

bdach: No, I get _what_ was tested, just not sure why the exponentially-increasing number of changes each step. It's just mostly my remark though, nothing worth changing really.

LittleEndu: The old optimization that didn't work (and thus wasn't commited) ramped up exponentially, that's why the test was exponential.

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: You forgot to invert the logic :/

bdach: that I did. resharper fixed it up for me automatically in IDE but I missed it when copying over here

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -118,6 +118,50 @@ void selectNextAndAssert(int amount)
...
+                    }
+                });
+                waitForSelection(expectedSet, expectedDiff);
+            }
+
+            // Selects next set once, difficulty index doesn't change
+            selectNextAndAssert(3, true, 2, 1);
+            // Selects next set 16 times (50 // 3 == 16), difficulty index changes twice (50 % 3 == 2)
+            selectNextAndAssert(50, true, 17, 3);
+            // Travels around the carousel thrice (200/60 == 3)
```
LittleEndu: This should say 200 // 60

peppy: where does "//" come from? i've never seen this used before

LittleEndu: It's floor division in programming languages that don't use // to start a comment

peppy: seems like maybe `\` is more universal in the math world. or just use `/` since it's quite obvious here.

peppy (dismissed):

bdach (dismissed):

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -83,6 +83,82 @@ public void TestTraversal()
...
+                waitForSelection(forwards ? amount + 1 : total_set_count - amount);
+            }
+        }
+
+        [Test]
+        public void TestTraversalBeyondVisibleDifficulties()
+        {
+            var sets = new List<BeatmapSetInfo>();
+
+            const int total_set_count = 200;
```
LittleEndu: Copy paste oversight, that's why tests are failing

bdach: to be more specific
```suggestion
            const int total_set_count = 20;
```

LittleEndu:
```suggestion
            const int total_set_count = 20;
```

bdach (approved):

PR #8485, opened 2020-03-28
Make beatmap carousel highlight recommended difficulties

### LINKED ISSUE #5172: Highlight the recommended difficulty
**Describe the new feature:** The feature is simple: highlight the recommended difficulty of a player on the carousel.

The feature is already implemented on the _Beatmaps Listing_ web page with a filter, and I think that can be very cool to see it in-game too:
image

- The idea is to highlight the recommended difficulty, based on my idea, with a golden colour and a star on the left side (but that's optional):
Recommended Difficulty

- I added that star for maybe a future proposal for other "warnings" (like missing maps with a `(!)` red icon), and the player can hover with the mouse to see what does it represent:
Recommended Difficulty pt 2

- And we can add another text saying something like "_This is the recommended difficulty_" on the info wedge (or even inside the "Details" tab):
Recommended Difficulty pt 3
Recommended Difficulty pt 4

**Pros:** A gold colour makes the recommended difficulty easy to see. If the player doesn't like that, he can disable the feature in the settings.
**Cons:** Many things are already coloured yellow/gold.. obviously I don't want to make everything yellow, so you can choose what's the best colour.

LittleEndu:
* Resolves #5172
* [x] Depends on #8444

This PR takes the recommended star difficulty of the user and finds out if a beatmap set has a difficulty around it (PR uses 1.5 either direction which I think matches osu-web). If such difficulty is found, that difficulty is highlighted (currently stars are colored yellow but it should be easy to change that design)


peppy:
Requires design consideration before implementation. @arflyte is this something we want to highlight and if so how should we be?

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -154,6 +162,37 @@ private void load(OsuConfigManager config)
...
+        }
+
+        internal bool IsRecommended(BeatmapInfo beatmap)
+        {
+            if (RecommendedStarDifficulty.Value == 0)
+                return false;
+
+            BeatmapSetInfo set = beatmap.BeatmapSet;
+            BeatmapInfo recommended = set.Beatmaps.OrderBy(b => SongSelect.DifferenceToRecommended(b, RecommendedStarDifficulty.Value)).FirstOrDefault();
+            return beatmap == recommended && Math.Abs(beatmap.StarDifficulty - RecommendedStarDifficulty.Value) < 1.5;
```
LittleEndu: This should actually be ``< 0.5``

Found in [osu-web](https://github.com/ppy/osu-web/blob/4687a1e8a5c132af3ccfc63a2b90d239ec395d71/app/Libraries/Search/BeatmapsetSearch.php#L174-L176)

arflyte:
Hmm, I'm not so sure. Making the stars yellow with the card slightly brighter might work, or a "recommended" tab.

peppy:
Want to provide a screenshot or something?

LittleEndu:
This is what I have right now not committed yet:
osu_2020-03-30_22-51-58

peppy:
i was hoping flyte would.. that looks pretty wrong as is

Gabixel:
maybe a temporary solution could be to put something near the star rating (without having to change other colors in the carousel). Here are some examples:

1. A text with a design similar to the beatmap's rank status one:
design 1

2. A text separated from the star rating by a vertical line, using a light color similar to the one used in the "empty" stars (I tried making it white but it looks bad in my opinion, maybe yellow/gold still works here). This one is a bit difficult to see on the screen, but it does its job:
design 2

But yeah, I guess flyte's design will be better anyway. These are only suggestions.

LittleEndu:
Yeah there are so many things that make the panel lighter (selecting it, hovering over it) that adding recommended highlight to it just makes it weird.

arflyte:
oh, sorry, didn't know you were referring to me. Not that simple; this design is pretty old so might take time to dig it up and update it to how it is now lol. I'll try to make a quick mockup.

peppy:
Something very temporary is fine, or not so temporary if it can be used in the old and new panels.

arflyte:
How about something like this like. Light orange glow instead of blue if the map has a recommended difficulty. The difficulty itself will have the same glow when selected. When not selected, it just have yellow star and the proposed recommended tag next to it.

Photoshop_200401-140926

peppy:
Every beatmap set has one recommended difficulty (it's a recommendation of which difficulty selection is best for the local player) so it shouldn't highlight the header panel.

arflyte:
Hmm, alright. Assumed that it wouldn't be the case for some maps with insane difficulties only lol. So let's keep it simple and make the star yellow with the recommended tag.

peppy:
I feel like the recommended bubble thing is going to be too big / distracting but we'll see.

LittleEndu:
"Recommended" should be a tooltip imo

LittleEndu:
> Every beatmap set has one recommended difficulty

In terms of selecting, yes, but I don't think we should highlight a 2* map for a top 100 player just because it's the closest to their overall recommended star difficulty.

> Light orange glow instead of blue if the map has a recommended difficulty

I think this looks busy (see screenshot), but I do like a way to know that a set has a recommended difficulty for me.

osu_2020-04-01_20-19-04

peppy:
As I mentioned, the set header should not be glowing at all.

Gabixel:
My last idea is to simply show like a golden-version of the ruleset icon for the recommended difficulty in the beatmap set (but I don't know if looks weird in the list below the "header": in this example, I used flyte's "bubble" idea for that). I guess it's simple and easy to understand. You can also add the "recommended" text inside its tooltip (based on @LittleEndu's idea):

design

Joehuu:
I would not color the stars as it will overlap https://github.com/ppy/osu/pull/3272 (if it ever gets merged).

bdach:
As a *personal* opinion, I think there should be no indication that a difficulty is recommended. It can be silently selected for the user, but I don't think it should be visible.

As for the reasoning: it reminds me of those pop-up windows from other games, that would suggest lowering the difficulty if you've failed something a few times. Those just make me double down and continue playing - I vastly prefer being able to select the diffculty myself and not be told what is "recommended" for me or not. So I'd dislike this being visible and would probably want to disable it, and this should probably not be a setting, so I'm against the recommended badge altogether myself.

smoogipoo:
I'm also veering on the edge of that the recommended difficulty shouldn't be displayed, agreeing with the reasonings by @bdach and @Joehuu.
Also it looks really weird as it is where some sets have a recommended difficulty and others don't.

LittleEndu:
Some sets don't highlight anything because the recommendation isn't close to the recommended star difficulty.

smoogipoo:
Yeah I got that, I just think the highest difficulty should be recommended if your rating is above it. As for if your rating is below it, well that goes into the other issues mentioned.

peppy:
For now let's leave this one. Once the panels are redesigned we can revisit this and potentially add it as a very subtle feature on the panel (I'm thinking just a small yellow dot somewhere if anything).

LittleEndu:
Writing this down to discuss and also so I wouldn't forget it. Doubt this will get used for a while tho considering how many redesigns are said to be in the working right now.

About why some difficulties were getting highlighted and some weren't. I thought it weird to highlight 2* maps for top 100 player and to highlight 7* maps for brand new players. But if the expectation is that something should be highlighted most of the time then this approach should be the best:

 * Find the recommended beatmap.
 * If it's not the lowest or the highest difficulty, highlight it
 * If it's the lowest, but our recommended star difficulty falls below some other difficulties, highlight it. (This should work for edge cases where difficulty spread isn't quite even. So if somebody's recommended star difficulty is 4* but the set only has 2* and 7* difficulties, the 2* one would be highlighted)
 * Vice-versa for highest.
 * If it's the only difficulty, or our recommended star difficulty falls outside the beatmaps difficulty spread (would always be the case for a single difficulty), see if the star rating falls into (-0.5, +0.5) range

This should avoid highlighting very easy stuff for best players and really hard stuff for new players, but still highlights something if there is a good choice to be made.

PR #8486, opened 2020-03-28
Make reverse arrows not follow snaking when they are already hit

### LINKED ISSUE #8283: Reverse arrow follows snaking out sliders
**Describe the bug:** 
When a repeating slider snakes out, the reverse arrow follows it instead of fading out on the spot. This looks weird on skins that have the "arrow" placed outside the slider.

Noticed on this vid from my last issue: https://www.youtube.com/watch?v=n03vo6jni9Y

**osu!lazer version:** 
2020.312.0

LittleEndu:
Fixes #8283

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
@@ -87,8 +87,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 
         public void UpdateSnakingPosition(Vector2 start, Vector2 end)
         {
-            // When the repeat is hit, the arrow should fade out on spot,
-            // it should no longer follow snaking
+            // When the repeat is hit, the arrow should fade out on spot rather than following the slider
             if (IsHit) return;
 
             bool isRepeatAtEnd = sliderRepeat.RepeatIndex % 2 == 0;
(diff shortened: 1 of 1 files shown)


peppy:
At very least this should have a comment explaining why it's done. Optimally, it would also have a test testing that the repeat is still following on miss. Also, this should probably be applied to `DrawableSliderTail` too, for future-proofing.

LittleEndu:
> Also, this should probably be applied to ``DrawableSliderTail`` too, for future-proofing.

``DrawableSliderTail`` does not implement ``ITrackSnaking``. Unsure how this would apply.

peppy:
It will need to once it displays `sliderendcircle` (see todo in the class).

LittleEndu:
> Optimally, it would also have a test testing that the repeat is still following on miss.

Sliders don't follow snaking in the test. Would test both ways if there is a way to force snaking.

peppy:
You could adjust the config setting in LocalConfig, I believe (check how other tests do it).

LittleEndu:
I was going to add a comment to ``DrawableSliderTail`` but InspectCode wasn't happy about "Redundant control flow jump statement" and I wouldn't know how to implement the rest of the function.

peppy:
For future reference, the test should be part of the PR, not dependent on it (makes for very annoying review). I'll let it fly this one time due to this being a relatively small change.

peppy (dismissed):

PR #8488, opened 2020-03-29
Fix auto mod results screen not displaying correctly

### LINKED ISSUE #8436: Auto play don't show score at result screen
When the mod AutoPlay is used, at the end of the song no score is shown at the result screen
Auto play score bug

**osu!lazer version:**  2020.325.0

[database.log](https://github.com/ppy/osu/files/4381802/database.log)
[network.log](https://github.com/ppy/osu/files/4381803/network.log)
[performance.log](https://github.com/ppy/osu/files/4381804/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4381805/runtime.log)

peppy:
- Closes https://github.com/ppy/osu/issues/8436.

Tidies up results screen handling, allowing ReplayPlayer to have its own logic.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneAllRulesetPlayers.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneAllRulesetPlayers.cs
@@ -4,7 +4,6 @@
 using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Allocation;
-using osu.Framework.Screens;
 using osu.Game.Configuration;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Catch;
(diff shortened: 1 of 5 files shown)


frenzibyte (changes requested):
~~I don't think this should close the linked issue, as we're supposed to fix the double-counting keys on the replay from the reported issue which I can still reproduce on this branch.~~

iiSaLMaN:
perhaps you meant to close https://github.com/ppy/osu/issues/8436 instead?

bdach:
Yeah, I think this works. `GotoRanking` existing is something that [was already discussed before but it turned out not to be necessary in the end](https://github.com/ppy/osu/pull/8084#issuecomment-596196608), so no objections there. Perhaps a smidge confusing that the normal player still has the replay path due to autoplay implementation details but I can't see a better way out right now.

The break waits could possibly go away when replay playback gets faster.

smoogipoo (approved):

PR #8489, opened 2020-03-29
Write a test for slider snaking

LittleEndu:
* [x] Depends on #8486

This PR writes a test for slider snaking. 

Behaviour that is tested both when snaking is enabled and disabled:
* Sliders snake in
* Sliders snake out
* Repeat sliders don't snake out unless all repeats are over
  * Tested for a slider with one repeat
  * Tested for a slider with two repeats 

Also includes tests for reverse arrow behaviour when snaking is enabled.


osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+            {
+                new Slider
+                {
+                    StartTime = 3000,
+                    Position = new Vector2(100, 100),
+                    Path = new SliderPath(PathType.PerfectCurve, new[]
+                    {
+                        Vector2.Zero,
+                        new Vector2(300, 200)
+                    }),
```
bdach: Seeing that we don't really care about position & path in these sliders I'd maybe extract out a helper method and take in `StartTime` and `RepeatCount` to make this a little leaner?

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+            testDoubleRepeating(true);
+            testDoubleRepeating(false);
+        }
+
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestArrowStays(bool isHit)
+        {
+            var isSame = isHit ? "is same" : "decreased";
+            var enable = isHit ? "enable" : "disable";
```
bdach: These are named super confusingly (having `isSame` not have a value of `"is same"` will trip people up).

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+        [Resolved]
+        private AudioManager audioManager { get; set; }
+
+        private TrackVirtualManual track;
+
+        protected override bool Autoplay => autoplay;
+        private bool autoplay;
+
+        private readonly Bindable<bool> snakingIn = new Bindable<bool>();
+        private readonly Bindable<bool> snakingOut = new Bindable<bool>();
```
bdach: Not a major gripe but `BindableBool` exists.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+            track = (TrackVirtualManual)working.Track;
+            return working;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(RulesetConfigCache configCache)
+        {
+            var config = (OsuRulesetConfigManager)configCache.GetConfigFor(Ruleset.Value.CreateInstance());
+            config.BindWith(OsuRulesetSetting.SnakingInSliders, snakingIn);
+            config.BindWith(OsuRulesetSetting.SnakingOutSliders, snakingOut);
```
bdach: We might actually want to restore these values to their initial ones at the end of each test. They are persisted to the database, which is shared with the game instance, so not restoring will actually end up interfering with other tests in the test browser or even overwriting devs' actual game settings.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+
+            addSeekStep(13500);
+            AddStep("retrieve 2nd slider repeat", () =>
+            {
+                var drawable = Player.DrawableRuleset.Playfield.AllHitObjects.Skip(1).First();
+                repeat = drawable.ChildrenOfType<Container<DrawableSliderRepeat>>().First().Children.First();
+            });
+            AddStep("Save repeat vector", () => vector = repeat.Position);
+            addSeekStep(13700);
+            AddAssert($"Repeat vector {isSame}", () => isHit ? Precision.AlmostEquals(vector.X, repeat.X, 1) && Precision.AlmostEquals(vector.Y, repeat.Y, 1) : repeat.X < vector.X && repeat.Y < vector.Y);
```
bdach: This is super hard to parse, I'd say it needs to be split into a method.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+            {
+                var body = (PlaySliderBody)slider.Body.Drawable;
+                vector = body.CurrentCurve.Last();
+            });
+            addSeekStep(1900);
+            AddAssert($"End vector {increased}", () =>
+            {
+                var body = (PlaySliderBody)slider.Body.Drawable;
+                var last = body.CurrentCurve.Last();
+                return snaking ? last.X > vector.X && last.Y > vector.Y : last == vector;
```
bdach: Why this is using strict equality and not `Precision.AlmostEquals` as above?

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,304 @@
...
+        {
+            var increased = snaking ? "increased" : "is same";
+            var decreased = snaking ? "decreased" : "is same";
+
+            setSnaking(snaking);
+            addSeekStep(8800);
+            AddStep("Save end vector", () =>
+            {
+                var body = (PlaySliderBody)slider.Body.Drawable;
+                vector = body.CurrentCurve.Last();
```
bdach: This (and the `First()` pattern) repeats a lot. I'd split a method or helper property here.

bdach (dismissed):
Test code could use some polish, it's a little repetitive even for tests. (I don't generally advocate for hard-enforcing DRY in test code but there are a few utility parts that could be factored out.)

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
LittleEndu: body path vectors don't update at all if snaking is turned off. repeat arrow position has a chance to update depending on when in frame hit is registered.

bdach: I'd make that a comment as it's not immediately obvious.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
LittleEndu: I didn't know how to include strings in formatted strings (or what ever they are called in c#) so I took them outside. These are only needed for step descriptions and you could just put them in the string to begin with.

bdach: You can nest ternaries in interpolated strings but I'm not necessarily advocating for that, just better naming that will encompass both values of the strings (like `repeatVectorState` and `snakingSettingState` or something).

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
peppy: for now this isn't a huge concern. should be fixed at a higher level.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
LittleEndu: I went with nested ternaries anyway. Shouldn't be that big of a deal now that the code is pretty DRY.

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
LittleEndu: Is there an issue opened for this?

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
LittleEndu: Should the creation of beatmap also be automatic?

LittleEndu:
Ouch. I moved the objects in beatmap and didn't think to check the test I didn't change....

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
peppy: i don't believe so

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -0,0 +1,225 @@
...
+
+        private DrawableSlider slider;
+        private DrawableSliderRepeat repeat;
+        private Vector2 savedVector;
+
+        [SetUpSteps]
+        public override void SetUpSteps() { }
+
+        [Test]
+        public void TestSnaking()
```
peppy, line 68: Putting all these tests in one `[Test]` is really hard to understand.

Split this out into `TestSnakingEnabledWithRepeats` and `TestSnakingDisabledWithRepeats`, each with [TestCase(0..3)] attributes.

peppy (changes requested):

bdach:
I don't know if it's just me, but I started reading through this and now the test code complexity is through the roof. There's so much conditional control flow and bool-checking that it makes everything really opaque.

Currently have the code checked out and am trying to make it a little more readable.

peppy:
Definitely much more can be done to make this legible. My review was just an initial pass and not a final foal.

bdach:
After a good few hours thinking of how to decrease complexity here I ended up with [this version](https://github.com/bdach/osu/blob/7135c997466f0ed207390a8f96604b28fcf0d464/osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs). It's a bit out there in terms of implementation (`checkPositionChange` could be considered... *weird*), so I purposefully did not push it in to gather feedback on whether this version is actually legible to anyone else but me first.

peppy:
That version already feels 10x better to read.

bdach:
Alright, that version (with some slight retouches) is now in here too. Abstaining from further review for obvious reasons, if there's any comments to the current state of affairs drop a comment and I'll fix them up.

peppy (approved):
I like this

PR #8490, opened 2020-03-29
Hide "retry" button on results screen after watching a replay

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/8488


TheWildTree:
Just a suggestion - I kind of feel that while quick-retry during a replay feels weird, the retry (or here it should say... replay?) button wouldn't be so bad to stay in this case - having a quick way to rewatch a replay after it's ended feels like a much better UX than having to exit, go to the leaderboards, sometimes scroll and find the same replay and only then be able to watch it again.

peppy:
There's a button to watch the replay again. It's the "watch replay" button.

TheWildTree:
Oops, I confused those two. My bad.

osu.Game/Screens/Ranking/ResultsScreen.cs
```diff
@@ -87,15 +91,20 @@ private void load()
 
             if (player != null)
             {
-                AddInternal(new HotkeyRetryOverlay
+                if (allowRetry)
```
bdach: Maybe collapse these conditionals into one or factor out a method for adding the retry stuff? It's a bit nested right now.

smoogipoo (approved):

PR #8494, opened 2020-03-30
Fix imports with no matching beatmap IDs still retaining a potentially invalid set ID

### LINKED ISSUE #8480: Not all maps have a "details" button.
If I have a local map that is not on the site, then some of these maps do not have a "details" button, while others have the same ones. Pressing the "details" button leads to a blank page of the map, so it’s better to remove it, and write in the information "local map".
The first screenshot shows a map where one difficulty does not have this button, and the second one has.
osu_2020-03-28_21-13-15
osu_2020-03-28_21-13-22
osu_2020-03-28_21-13-32

osu!lazer version:
2020.327.0

peppy:
- Closes https://github.com/ppy/osu/issues/8480.


peppy:
In an interest of not spending a huge amount of time rewriting tests and rethinking logic I've limited the scope of this change to only apply to cases where online retrieval is occurring. This still resolves the issue mentioned in the linked issue but is obviously not a final solution.

smoogipoo (approved):

PR #8502, opened 2020-03-30
Add support for HitCircleOverlayAboveNumber legacy skin property

EVAST9919:
Resolves #8502.

### DIFF
--- a/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
+++ b/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
@@ -62,6 +62,11 @@ namespace osu.Game.Rulesets.Osu.Skinning
                 }
             };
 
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -80,6 +80,11 @@ Texture getTextureWithFallback(string name)
 
                 return tex ?? skin.GetTexture($"hitcircle{name}");
             }
+
+            bool overlayAboveNumber = skin.GetConfig<OsuSkinConfiguration, bool>(OsuSkinConfiguration.HitCircleOverlayAboveNumber)?.Value ?? true;
```
peppy: should be above the nested function, probably right beneath the children definition so it's less likely it is missed when other drawables are (potentially) added, causing breaking behaviour.

peppy (changes requested):

peppy (approved):

PR #8509, opened 2020-03-30
Support widescreen per-layer storyboard masking

### LINKED ISSUE #8459: Storyboards and videos are forced into 4:3 ratio
**Describe the bug:** 
Most beatmaps I have have widescreen support, but after the new March 25th update, the video or storyboard is always in 4:3 ratio.
osu_2020-03-27_00-00-35
**osu!lazer version:** 
2020.325.0
**Logs:**

[performance.log](https://github.com/ppy/osu/files/4390710/performance.log)
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/4390711/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/4390712/performance-gamethread.draw.log)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/4390713/performance-gamethread.input.log)
[performance-gamethread.update.log](https://github.com/ppy/osu/files/4390714/performance-gamethread.update.log)
[runtime.log](https://github.com/ppy/osu/files/4390715/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4390716/updater.log)
[database.log](https://github.com/ppy/osu/files/4390717/database.log)
[network.log](https://github.com/ppy/osu/files/4390718/network.log)

bdach:
Resolves #8459.

# Summary

[As described here](https://github.com/ppy/osu/issues/8459#issuecomment-605698881), #8426 caused a regression in widescreen storyboard display due to adding per-layer masking, which did not work as intended, as all individual layers had a hard-coded 640x480 size for the sake of proper storyboard element placement.

This PR implements the proposal made in the comment above by constraining widescreen storyboards to a 16:9 aspect ratio. This is done by individual layers sizing to the parent, which is the entire `DrawableStoryboard`, whose size is properly set here:

https://github.com/ppy/osu/blob/f414b7fbb60d9e47a054f1d61493d846696125bc/osu.Game/Storyboards/Storyboard.cs#L57-L62

and then internally having a nested, centred `LayerElementContainer` that is hard-sized to 640x480 to ensure proper placement.

The overall behaviour is now as follows:
* the video layer is never masked,
* normal layers are masked to 4:3 if `WidescreenStoryboard` is not set, and masked to 16:9 otherwise.

This is a departure from stable, which as far as I can see will mask to 4:3 if `WidescreenStoryboard` is not set, but will not mask at all if it is set. Masking to 16:9 feels more correct to me as an owner of an ultra-wide monitor (many storyboards look janky without it since things are shown that are not supposed to be).

# Remarks

As far as I can see, this does not have a major effect on performance compared to current `master`. Here's some not-too-scientific evidence from one run of [`world.execute(me)`](https://osu.ppy.sh/beatmapsets/470977#osu/1012836) (times are the sum of update times for storyboard sprites & layers):

| timestamp |current `master` | | this PR | |
| -: | :-: | -: | :-: | -: |
| ~0:09 | osu_2020-03-30_21-57-21 | 4.8ms | osu_2020-03-30_22-00-34 | 5.4ms |
| ~0:33 | osu_2020-03-30_21-57-34 | 9.2ms | osu_2020-03-30_22-00-58 | 9.6ms |
| ~0:51 | osu_2020-03-30_21-57-42 | 7.0ms | osu_2020-03-30_22-01-16 | 7.2ms |
| ~1:31 | osu_2020-03-30_21-57-56 | 10.2ms | osu_2020-03-30_22-01-56 | 11.1ms |
| ~2:32 | osu_2020-03-30_21-58-12 | 21.2ms | osu_2020-03-30_22-02-28 | 24.2ms |
| ~2:53 | osu_2020-03-30_21-58-34 | 69.6ms | osu_2020-03-30_22-02-49 | 69.5ms |

Results are slightly worse, but I don't think that much worse as to disqualify this outright, and I think the correctness is worth it. It was one run, too, so might have been a fluke as well. I strongly encourage attempting reproduction.

(As a side note: I wrote a now-deleted comment in the issue thread earlier, thinking this way was inefficient due to the multiple maskings, but then noticed I had mistakenly used a `CompositeDrawable` instead of a `LifetimeManagementContainer` when going through all possible variants of fixes I could come up with. I *did* use a lifetime container initially but then wondered why'd performance tank so much the second time I tried this solution and forgot to use it...)

Not sure how testable this is. I can include something if you think it's worth it.

### DIFF
--- a/osu.Game/Storyboards/Drawables/DrawableStoryboard.cs
+++ b/osu.Game/Storyboards/Drawables/DrawableStoryboard.cs
@@ -50,7 +50,7 @@ namespace osu.Game.Storyboards.Drawables
 
             AddInternal(Content = new Container<DrawableStoryboardLayer>
             {
-                Size = new Vector2(640, 480),
+                RelativeSizeAxes = Axes.Both,
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
             });
--- a/osu.Game/Storyboards/Drawables/DrawableStoryboardLayer.cs
+++ b/osu.Game/Storyboards/Drawables/DrawableStoryboardLayer.cs
@@ -8,7 +8,7 @@ using osu.Framework.Graphics.Containers;
 
 namespace osu.Game.Storyboards.Drawables
 {
-    public class DrawableStoryboardLayer : LifetimeManagementContainer
+    public class DrawableStoryboardLayer : CompositeDrawable
     {
         public StoryboardLayer Layer { get; }
         public bool Enabled;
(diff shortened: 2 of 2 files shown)


peppy:
Not 100% sure about arbitrary 16:9 masking, as storyboards may have been made to support higher than that correctly (its not ever written in the spec as to what aspect ratio that setting means, right?)

bdach:
No, I couldn't see any specific details on what the wide-screen storyboard setting did on the wiki, much less a mention of that aspect ratio. There are a few community resources and forum threads that mention this ratio, but I expect it has more to do with 16:9 becoming the *de facto* standard for screen size rather than with how the setting actually works.

As to whether there are beatmaps with storyboards designed for wider screens, I couldn't really say.

peppy:
Can you include some reference storyboards which display *worse* without this? Just a few links would be useful in making a decision here.

peppy:
Hmm, never mind. I didn't realise it was already implemented this way but just not masked correctly. Let's leave it as it is for now.

peppy (approved):

PR #8519, opened 2020-03-31
Refactor mania column type check

mcendu:


bdach:
Code seems fine, but I feel like I've seen this before. (And now it has conflicts.)

smoogipoo:
Closing as this is included in https://github.com/ppy/osu/pull/8527 and will get merged there.

PR #8521, opened 2020-03-31
Remove ScaleDownToFit as it was implemented without enough safety

peppy:
After animation changes, animation does not get an initial DrawSize until the first `Update` (or `LoadComplete`). By this point, `SkinnableDrawable` has already decided it doesn't need to be scaled down and will never check again.

I attempted to implement this in a safer way, but based on there being zero required usages of it, I feel it is safer to just remove for the time being. We can consider adding it back and implementing more correctly if required in the future.


peppy:
Going to merge this for now as it's a pretty low impact change but required to keep skinned catchers working correctly. Can discuss a path forward at a later point in time.

PR #8524, opened 2020-03-31
Add mania hold note skinning

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8523

Stretches the texture for now, since we don't have wrapping support in o!f yet.


mcendu:
One footnote: A skin that does not provide mania skin elements would cause the mania playfield to seemingly render almost nothing.

peppy:
Yeah, looks like an oversight. Will render nothing.

Also the hit explosion area is off. It seems like explosions are now happening above the correct location. Will hold off merging this, although the two issues mentioned are already in master unfortunately.

peppy (approved):

peppy:
Going to merge this as it doesn't worsen the current state of master and looks correct itself. Will be easier to work towards fixing the remaining issues on a clean slate.

PR #8526, opened 2020-03-31
Fix osu!catch catcher hit area being too large

### LINKED ISSUE #8285: osu!catch catcher hitbox is too big
**Describe the bug:** For example with cs4 catcher's hitbox is 24 unit bigger than in stable (12 from each side).

image
image

Map to test: https://osu.ppy.sh/beatmapsets/1125773#fruits/2352219

I placed the first fruit at the platter's border in stable. All next fruits I moved 1 unit to the left and right. So in stable you only catch the two first fruit. In lazer you catch up to 12 unit moved to left & right.

**osu!lazer version:** 
2020.312.0

peppy:
Turns out we weren't correctly applying a sizeable "margin" adjustment present in stable. Tested using beatmap provided in https://github.com/ppy/osu/issues/8285 (and compared to be floating-point accurate against osu-stable for same circle size).

- Closes https://github.com/ppy/osu/issues/8285

### DIFF
--- a/osu.Game.Rulesets.Catch/Difficulty/CatchDifficultyCalculator.cs
+++ b/osu.Game.Rulesets.Catch/Difficulty/CatchDifficultyCalculator.cs
@@ -72,10 +72,7 @@ namespace osu.Game.Rulesets.Catch.Difficulty
         protected override Skill[] CreateSkills(IBeatmap beatmap)
         {
             using (var catcher = new Catcher(beatmap.BeatmapInfo.BaseDifficulty))
-            {
                 halfCatcherWidth = catcher.CatchWidth * 0.5f;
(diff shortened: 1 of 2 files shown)


peppy:
@smoogipoo can I get you to check on the failing test? as far as i can tell the margin should be considered for diffcalc.

smoogipoo:
There's a secondary 0.8 applied in the diffcalc, which you should maybe remove?

https://github.com/ppy/osu/blob/d4222f7d924a5df9dc7490aa4ed4d82d73c947b2/osu.Game.Rulesets.Catch/Difficulty/CatchDifficultyCalculator.cs#L77

peppy:
I guess the comment attached to that line is... incorrect reasoning?

smoogipoo:
Yeah, it was trying to simulate _gameplay_, in-fact :P

smoogipoo (approved):

PR #8535, opened 2020-04-01
Add check to detect whether mania is skinned

### LINKED ISSUE #8532: osu!mania is not visible when a skin is used without osu!mania components
image

mcendu:
Fixes #8532

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
@@ -35,10 +35,10 @@ namespace osu.Game.Rulesets.Mania.Skinning
         private void sourceChanged()
         {
             isLegacySkin = new Lazy<bool>(() => source.GetConfig<LegacySkinConfiguration.LegacySetting, decimal>(LegacySkinConfiguration.LegacySetting.Version) != null);
-            hasKeyTexture = new Lazy<bool>(() => source.GetTexture(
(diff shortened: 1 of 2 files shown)


smoogipoo:
I'm not sure about this. It's quite fragile:
1. The user may have skinned 4K but not 7K.
2. It also has no consideration for animations - `mania-note1-0` may be the only texture for example.
3. It's possible that the notes aren't skinned, but other components are (say, hold note body texture replaced with a flat texture because it's too distracting).

I see two possible solutions:
1. Fallback to the classic skin. We don't have the ability for this yet, not sure how hard it'd be to implement. This should be the eventual goal.
2. Perform the texture check per-component. I think this is probably the easiest solution, but will require giving `ManiaSkinComponent` probably the same amount of logic as in `ManiaSkinConfigurationLookup` and `ManiaLegacyElement`/`ManiaLegacyColumnElement`. This would only be a temporary measure really, since the fallbacks would be to lazer's skin, which we likely don't want for legacy skins (see point 3).

smoogipoo:
After reconsideration, I think this would be fine as a super temporary solution for the case that the user doesn't have _any_ mania skin, under the condition that this checks via `this.GetAnimation()`.

It'll have to be changed at some point to cover the other two cases I mentioned, though.

smoogipoo:
This also needs to not return legacy mania config values in the same cases.

smoogipoo (dismissed):
Yeah I think this is fine for now

smoogipoo (approved):

Issue #8530, opened 2020-03-31
Skinned mania notes don't match judgement area's size


peppy:
Stable:

![image](https://user-images.githubusercontent.com/191335/78042735-663c8480-73ad-11ea-981c-0c934d546909.png)

master:

![image](https://user-images.githubusercontent.com/191335/78042913-aa2f8980-73ad-11ea-89bc-6635bb27aaa5.png)

Doesn't match the judgement area and visually looks stretched.

peppy:
![image](https://user-images.githubusercontent.com/191335/78043214-0692a900-73ae-11ea-9fb9-0763ec6ac5d8.png)

Also don't look to match each other

smoogipoo:
Not sure what the OP in this issue is detailing. Are you talking about the 1px gap to the right? Stable has the same issue afaik.

peppy:
No, talking about the height of the note being different to the judgement area's height.second screenshot shows where they differ from each other, too.

peppy:
For better comparison:

![image](https://user-images.githubusercontent.com/191335/78100065-bdc70880-741e-11ea-8ac9-844fab4b7f41.png)

All three of these should match.

smoogipoo:
I think it's close enough after the referencing PR:
![image](https://user-images.githubusercontent.com/1329837/78106445-5dd85e00-742e-11ea-832e-caac887441b4.png)

The height of the blue portion is ~18px (26 / 2 * 0.9 * 1.6026), the height of a note given a column width of 48 is ~15px (measured from the texture).

Stable has a similar issue, which is about the same size:
![image](https://user-images.githubusercontent.com/1329837/78106855-46e63b80-742f-11ea-95db-bcaf377ddccb.png)

peppy:
The stable skin was meant such that the judgement should occur directly between the two (yellow) lines. Unfortunately that looks to have regressed at some point (and is currently wrong on stable) but the sizing does look correct for that.

Based on the screenshot in my OP it does match pixel perfect on stable:

![image](https://user-images.githubusercontent.com/191335/78107884-3e8f0000-7431-11ea-9f8b-3e6ee55246f8.png)

Note height fitting perfectly between the two yellow lines.

peppy:
Latest master looks in line now:

![image](https://user-images.githubusercontent.com/191335/78108023-87df4f80-7431-11ea-87d0-32b15c6c172e.png)

mcendu:
[...]

PR #8536, opened 2020-04-01
Implement column width/spacing

smoogipoo:
Resolves the part of https://github.com/ppy/osu/issues/8530 where the notes have different sizes.

### DIFF
--- a/osu.Game.Rulesets.Mania/UI/ManiaStage.cs
+++ b/osu.Game.Rulesets.Mania/UI/ManiaStage.cs
@@ -179,6 +179,7 @@ namespace osu.Game.Rulesets.Mania.UI
                                           ?.Value;
 
                 if (width == null)
+                    // only used by default skin (legacy skins get defaults set in LegacyManiaSkinConfiguration)
                     col.Width = col.IsSpecial ? Column.SPECIAL_COLUMN_WIDTH : Column.COLUMN_WIDTH;
                 else
                     col.Width = width.Value;
--- a/osu.Game.Rulesets.Mania/UI/ManiaStage.cs
+++ b/osu.Game.Rulesets.Mania/UI/ManiaStage.cs
@@ -14,6 +14,7 @@ using osu.Game.Rulesets.Mania.Objects.Drawables;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.UI;
 using osu.Game.Rulesets.UI.Scrolling;
+using osu.Game.Skinning;
 using osuTK;
 using osuTK.Graphics;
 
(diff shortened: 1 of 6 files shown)


osu.Game.Rulesets.Mania/UI/ManiaStage.cs
```diff
@@ -150,6 +150,41 @@ public ManiaStage(int firstColumnIndex, StageDefinition definition, ref ManiaAct
...
+
+                    col.Margin = new MarginPadding { Left = spacing };
+                }
+
+                float? width = currentSkin.GetConfig<LegacyManiaSkinConfigurationLookup, float>(
+                                              new LegacyManiaSkinConfigurationLookup(Columns.Count, LegacyManiaSkinConfigurationLookups.ColumnWidth, col.Index))
+                                          ?.Value;
+
+                if (width == null)
+                    col.Width = col.IsSpecial ? Column.SPECIAL_COLUMN_WIDTH : Column.COLUMN_WIDTH;
```
mcendu, line 182: Most skins assume that the column width is 30 instead of the lazer values. Might be another PR imo.

smoogipoo: Check the line below this one, where I'm using the skin width.

osu.Game.Rulesets.Mania/UI/ManiaStage.cs
```diff
@@ -150,6 +150,41 @@ public ManiaStage(int firstColumnIndex, StageDefinition definition, ref ManiaAct
...
+            skin.SourceChanged += onSkinChanged;
+
+            onSkinChanged();
+        }
+
+        private void onSkinChanged()
+        {
+            foreach (var col in columnFlow)
+            {
+                if (col.Index > 0)
```
peppy, line 168: what does index 0 mean here?

smoogipoo: The first column. It doesn't have spacing.

peppy: aha, missed that

osu.Game.Rulesets.Mania/UI/ManiaStage.cs
peppy: this is only ever hit by the default skin, correct?

i think we may want to re-think this lookup process. at very least the configuration lookup type should be renamed from "legacy" if we are using it in a mainstream class (`ManiaStage`)

smoogipoo: Yeah there's some configs that are 100% legacy, and others that probably aren't going to be, such as column width and spacing.

I didn't split it out because I'm not sure I want to implement 3 ways of retrieving mania skin stuff though: `ManiaSkinComponent`, `LegacyManiaSkinConfigurationLookup`, `ManiaSkinConfigurationLookup`, all before we have a system for dealing with drawable configs.

peppy (approved):

PR #8537, opened 2020-04-01
Fix column lights positioned incorrectly

smoogipoo:


peppy:
This still looks too high on the default skin:

image

smoogipoo:
That's the explosion, not the lighting, which hasn't been done yet.

mcendu:
> 
> 
> That's the explosion, not the lighting, which hasn't been done yet.

Would be an easy fix

smoogipoo:
Ah hang that's a regression on the default skin, different issue though.

peppy:
This is good to go, right?

peppy (approved):

smoogipoo:
Yep

PR #8542, opened 2020-04-01
Fix HP draining in non-play sections

### LINKED ISSUE #7607: HP drains in non-play sections
**Describe the bug:** In osu!lazer, HP drain occurs in non-play sections, like right after a break or after skipping the beginning of the song but before the first note was hit.

This may be an issue that occurs in sections like before a break but after the last note, but I have not tested that yet.

**Screenshots or videos showing encountered issue:** https://youtu.be/lzn6P9MOdG4
Notice that osu!lazer immediately drains HP after the break, but osu!stable only drains upon hitting the first note.

**osu!lazer version:** master branch 52b66fdc167059ca0336dde474e2ee84073d04b1

kornkaobat:
Fixes #7607.

### DIFF
--- a/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
@@ -47,7 +47,7 @@ namespace osu.Game.Rulesets.Scoring
         private double targetMinimumHealth;
         private double drainRate = 1;
 
-        private bool breakend;
+        private bool applyDrain;
 
         /// <summary>
         /// Creates a new <see cref="DrainingHealthProcessor"/>.
@@ -62,21 +62,23 @@ namespace osu.Game.Rulesets.Scoring
         {
             base.Update();
 
+            // can check this first, as drain is re-enabled externally by ApplyResultInternal
+            if (!applyDrain)
+                return;
+
+            // drain is on - check if a break is starting; if yes, disable and early-return
             if (IsBreakTime.Value)
             {
-                breakend = false;
+                applyDrain = false;
                 return;
             }
 
-            // Health should only be drained after the first post-Break Hit is judged
-            if (!IsBreakTime.Value && breakend)
-            {
-                // When jumping in and out of gameplay time within a single frame, health should only be drained for the period within the gameplay time
-                double lastGameplayTime = Math.Clamp(Time.Current - Time.Elapsed, drainStartTime, gameplayEndTime);
-                double currentGameplayTime = Math.Clamp(Time.Current, drainStartTime, gameplayEndTime);
+            // actually apply drain
+            // When jumping in and out of gameplay time within a single frame, health should only be drained for the period within the gameplay time
+            double lastGameplayTime = Math.Clamp(Time.Current - Time.Elapsed, drainStartTime, gameplayEndTime);
+            double currentGameplayTime = Math.Clamp(Time.Current, drainStartTime, gameplayEndTime);
 
-                Health.Value -= drainRate * (currentGameplayTime - lastGameplayTime);
-            }
+            Health.Value -= drainRate * (currentGameplayTime - lastGameplayTime);
         }
 
         public override void ApplyBeatmap(IBeatmap beatmap)
@@ -95,7 +97,7 @@ namespace osu.Game.Rulesets.Scoring
         {
             base.ApplyResultInternal(result);
             healthIncreases.Add((result.HitObject.GetEndTime() + result.TimeOffset, GetHealthIncreaseFor(result)));
-            breakend = true;
+            applyDrain = true;
         }
 
         protected override void Reset(bool storeResults)
--- a/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
@@ -47,7 +47,10 @@ namespace osu.Game.Rulesets.Scoring
         private double targetMinimumHealth;
         private double drainRate = 1;
 
+        private double lastHitTime;
+
         private bool applyDrain;
+        private bool firstObjectIsHit;
 
         /// <summary>
         /// Creates a new <see cref="DrainingHealthProcessor"/>.
@@ -68,6 +71,14 @@ namespace osu.Game.Rulesets.Scoring
 
             // drain is on - check if a break is starting; if yes, disable and early-return
             if (IsBreakTime.Value)
+            {
+                applyDrain = false;
+                return;
+            }
+
+            // Bandaid Fix for pre-break HP drain
+            // Checks if the first object is Hit. After that, checks if the difference between Current time and lastHitTime exceeds 600ms
+            if (firstObjectIsHit && (Time.Current - lastHitTime) > 600)
             {
                 applyDrain = false;
                 return;
@@ -97,6 +108,8 @@ namespace osu.Game.Rulesets.Scoring
         {
             base.ApplyResultInternal(result);
             healthIncreases.Add((result.HitObject.GetEndTime() + result.TimeOffset, GetHealthIncreaseFor(result)));
+            lastHitTime = result.HitObject.GetEndTime();
+            firstObjectIsHit = true;
             applyDrain = true;
         }
 
--- a/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
@@ -47,10 +47,10 @@ namespace osu.Game.Rulesets.Scoring
         private double targetMinimumHealth;
         private double drainRate = 1;
 
-        private double lastHitTime;
+        private int hitObjectCounter = 0;
+        private int breakCounter = 0;
 
         private bool applyDrain;
-        private bool firstObjectIsHit;
 
         /// <summary>
         /// Creates a new <see cref="DrainingHealthProcessor"/>.
@@ -71,14 +71,6 @@ namespace osu.Game.Rulesets.Scoring
 
             // drain is on - check if a break is starting; if yes, disable and early-return
             if (IsBreakTime.Value)
-            {
-                applyDrain = false;
-                return;
-            }
-
-            // Bandaid Fix for pre-break HP drain
-            // Checks if the first object is Hit. After that, checks if the difference between Current time and lastHitTime exceeds 600ms
-            if (firstObjectIsHit && (Time.Current - lastHitTime) > 600)
             {
                 applyDrain = false;
                 return;
@@ -108,8 +100,16 @@ namespace osu.Game.Rulesets.Scoring
         {
             base.ApplyResultInternal(result);
             healthIncreases.Add((result.HitObject.GetEndTime() + result.TimeOffset, GetHealthIncreaseFor(result)));
-            lastHitTime = result.HitObject.GetEndTime();
-            firstObjectIsHit = true;
+
+            if (beatmap.HitObjects[hitObjectCounter + 1].StartTime >= beatmap.Breaks[breakCounter].EndTime)
+            {
+                applyDrain = false;
+                breakCounter++;
+                hitObjectCounter++;
+                return;
+            }
+
+            hitObjectCounter++;
             applyDrain = true;
         }
 
--- a/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
@@ -47,6 +47,8 @@ namespace osu.Game.Rulesets.Scoring
         private double targetMinimumHealth;
         private double drainRate = 1;
 
+        private int breakendcounter;
+
         /// <summary>
         /// Creates a new <see cref="DrainingHealthProcessor"/>.
         /// </summary>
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -47,6 +47,8 @@ public class DrainingHealthProcessor : HealthProcessor
         private double targetMinimumHealth;
         private double drainRate = 1;
 
+        private int breakendcounter;
```
LittleEndu, line 50: Why is this an integer? It's only ever incremented and checked if above zero. Could be bool instead.

kornkaobat: It is so that I can keep track of Hits using `breakendcounter++;`

iiSaLMaN, line 50:
```suggestion
        private bool judgedAfterBreak;
```
is probably what you want

LittleEndu: That's what I said tho. You only write to this value by resetting it or incrementing it. And you only read from this value by seeing if it has been incremented. idk how is it not clear that you can make this a boolean instead.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -60,7 +62,12 @@ protected override void Update()
         {
             base.Update();
 
-            if (!IsBreakTime.Value)
+            if (IsBreakTime.Value)
+            {
+                breakendcounter = 0;
+            }
+
+            if (!IsBreakTime.Value && breakendcounter > 0)
```
iiSaLMaN, line 70:
```csharp
if (IsBreakTime.Value || !judgedAfterBreak)
{
    judgedAfterBreak = false;
    return;
}

// draining logic
...
```

iiSaLMaN: also provide a comment as to why draining will only be applied when a hit object is judged post-break.

frenzibyte (changes requested):

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
kornkaobat: Fixed in 61be2f5.

kornkaobat: Fixed ( naming is different ) in 61be2f5.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
kornkaobat: !judgedAfterBreak and judgedAfterBreak = false is the same thing or am I wrong here?

LittleEndu: ! is faster than ==

kornkaobat: I think I would only add `return;` because I can't seem to grasp the` ...|| !judgedAfterBreak and judgedAfterBreak = false` part.

LittleEndu: You can get to the assignment via the other variable in the or operator

TheWildTree: He means that there's a simple oversight in the suggestion, making it essentially

```csharp
if (!judgedAfterBreak)
    judgedAfterBreak = false;
```

Should've been without the `!` here.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
iiSaLMaN: `breakend` doesn't make much sense when talking about draining health after first hit object being hit (we're at the beginning, not after a break end).

Also please follow our naming style for consistency and better code readability (each word from second onwards starts with capital letter): 
`Judged after break` field for example becomes `judgedAfterBreak` not `judgedafterbreak`

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
iiSaLMaN: @kornkaobat I've suggested it for better code readability:

current code: 
1. if on break time, clear `judgedAfterBreak`. 
2. if not on break time and is `judgedAfterBreak` then do draining logic.

suggested code: 
1. if on break time or not `judgedAfterBreak` then clear `judgedAfterBreak` and don't do draining logic (aka `return;`). 
2. do draining logic

also clearing `judgedAfterBreak` inside the check is intended because we're featuring `IsBreakTime` on the check and don't really like breaking it to 2 separate `if`s.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -60,7 +62,14 @@ protected override void Update()
...
 
-            if (!IsBreakTime.Value)
+            if (IsBreakTime.Value)
+            {
+                breakend = false;
+                return;
+            }
+
+            // Health should only be drained after the first post-Break Hit is judged
+            if (!IsBreakTime.Value && breakend)
```
bdach, line 72: The `!IsBreakTime.Value` check is now redundant, as the method early-returns if it's true (so it can't *not* be false at this point).

After eliminating it you could as well negate the other check, and I *think* the following variant is equivalent:
```csharp
// can check this first, as drain is re-enabled externally by ApplyResultInternal
if (!applyDrain)
    return;

// drain is on - check if a break is starting; if yes, disable and early-return
if (IsBreakTime.Value)
{
    applyDrain = false;
    return;
}

// actually apply drain
```

bdach (changes requested):
This should have tests. `TestSceneDrainingHealthProcessor` already exists, so it shouldn't be very hard to add a test for this behaviour.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -60,7 +62,14 @@ protected override void Update()
         {
             base.Update();
 
-            if (!IsBreakTime.Value)
+            if (IsBreakTime.Value)
+            {
+                breakend = false;
```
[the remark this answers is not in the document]
kornkaobat: Fixed in e09786e.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
kornkaobat: Fixed in e09786e.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
kornkaobat: Fixed via bdach's suggestion e09786e.

mcendu: Why aren't you using the "resolve conversation" button?

kornkaobat: Already done that.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -86,6 +97,7 @@ protected override void ApplyResultInternal(JudgementResult result)
         {
             base.ApplyResultInternal(result);
             healthIncreases.Add((result.HitObject.GetEndTime() + result.TimeOffset, GetHealthIncreaseFor(result)));
+            applyDrain = true;
```
smoogipoo, line 100: A similar change needs to be done to `computeDrainRate()` below (change probably needs to happen to line 144), which simulates this whole `Update()` + `ApplyResult` cycle.

smoogipoo (changes requested):

smoogipoo:
As I mentioned in https://github.com/ppy/osu/issues/8585#issuecomment-608208468, this also needs to account for the section _before_ the break, which also shouldn't drain.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -73,6 +76,14 @@ protected override void Update()
                 return;
             }
 
+            // Bandaid Fix for pre-break HP drain
+            // Checks if the first object is Hit. After that, checks if the difference between Current time and lastHitTime exceeds 600ms
+            if (firstObjectIsHit && (Time.Current - lastHitTime) > 600)
```
smoogipoo, line 81: No.

smoogipoo (changes requested):
I believe this could be done by computing all the drain sections inside `ApplyBeatmap()` and checking if the current time lies within any of those sections. It's a similar implementation to how `IsBreakTime` is computed.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
kornkaobat: Can you explain the input-output of this method ApplyBeatmap(IBeatmap beatmap)?

smoogipoo: It receives in a beatmap, and it extracts whatever it needs out of it. It does whatever it wants to do.

mcendu: Maybe calculate something like `List<System.Tuple<int,int>> NonDrainSections`?

kornkaobat: /// <summary>
        /// The minimum duration required for a break to have any effect.
        /// </summary>
        public const double MIN_BREAK_DURATION = 650;

kornkaobat: I think we should approach this problem by changing broken beatmap Break Starttimes because, we can't predict into the future and assure if the Current Hitobject that we're hitting is the closest Hitobject to the upcoming Breaktime.

iiSaLMaN: @mcendu's idea is probably what you need to solve this (although the tuple must be of `double,double` because that's the type we use to store time values and also must be a `private` local). 
where you want to calculate this would be inside the `ApplyBeatmap(beatmap)` method, now think on how to calculate it, search off what `Beatmap` has. (should be fairly simple)

bdach:
Please, please, *please* do not push code that you know to be crashing.

kornkaobat:
Why is that though, does the CI waste bandwidth or something else? It won't be merged yet so what's the harm.

LittleEndu:
Bandwidth isn't the only resource CI is wasting when it has to test commits that one knows to fail.

kornkaobat:
Ok then.

bdach:
It does waste CI bandwidth (other PR builds won't run), and in general development by trial-and-error demonstrates the lack of necessary knowledge to actually fix the problem, and also a lack of knowledge of the *process* of git workflow (PRs are supposed to be functional code always, unless opened as draft, waiting for a prerequisite merge, or just plain mistake - those happen).

kornkaobat:
How to catch what specific error when a Unhandled Exception happens while osu! is running and is about to crash.

bdach:
Check log files for the stack trace, or just use an IDE and run in debug mode (most do automatic breakpoints on exceptions)...?

With all due respect, this sort of question you should be looking to find out the answer to on your own. The discussion on this PR has been inordinately long for what should be a relatively simple change. We cannot possibly provide this level of counsel if you choose to continue submitting code.

iiSaLMaN:
@kornkaobat Please understand what @smoogipoo's review tells you https://github.com/ppy/osu/pull/8542#pullrequestreview-387197443, otherwise you can't proceed with incorrect code like this, If you don't really know what he means then search up, understand, learn how to do it and attempt until you reach to the correct solution. 

Otherwise if you still can't figure it out then it would be better to just close this PR and let us solve it in the correct way.

kornkaobat:
Ok I'll leave it to you guys and girls to solve it yourselves then.

kornkaobat:
I've solved this problem by doing some hardcore programming. Enjoy.

bdach:
The latest commit is so illegible and wrong that instead of spending even more time on explaining basic programming practices I am going to go ahead and close this again. Please do not reopen this unless you are sure what you are doing.

kornkaobat:
It does work so what's wrong with it?

bdach:
If you insist:

* Ten added fields. I *really* don't think that many is necessary.
* Checking hit object type by getting it as string and *comparing to a magic string*.
* Duplicating the same 12-line snippet in two places.
* And the change in general is just unmaintainable. I don't understand what the code is doing with any *hashes* (which, as far as I can see, are not hashes) and I don't think I should have to attempt to.

I would strongly recommend starting with the basics of C# and object-oriented programming design before attempting to submit code again.

kornkaobat:
Ok then live with your inferior programs. I want to see if you guys and girls can do any better. Trust me you'll run into problems when you try to code this, and I'll not be there to help. This game release cycle will suffer because this is a core gameplay problem and you'll be remembering me but, I'm far ahead in life I wouldn't came back and look at this repo again. Farewell.

PR #8556, opened 2020-04-02
Implement mania normal-note hit-explosion skinning

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3418
- [x] https://github.com/ppy/osu/pull/8555
- [x] https://github.com/ppy/osu-resources/pull/95


osu.Game.Rulesets.Mania/Skinning/LegacyHitExplosion.cs
```diff
@@ -0,0 +1,69 @@
...
+            if (explosion != null)
+                explosion.Anchor = direction.NewValue == ScrollingDirection.Up ? Anchor.TopCentre : Anchor.BottomCentre;
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            explosion?.FadeInFromZero(80)
+                     .Then().FadeOut(120);
```
mcendu, line 66:
```suggestion
            explosion?.FadeInFromZero(80)
                     .Then().FadeOut(120);

            Expire(true);
```

osu.Game.Rulesets.Mania/UI/Components/ColumnHitObjectArea.cs
```diff
@@ -12,7 +11,7 @@ namespace osu.Game.Rulesets.Mania.UI.Components
 {
     public class ColumnHitObjectArea : HitObjectArea
     {
-        public readonly Container<HitExplosion> Explosions;
+        public readonly Container<Drawable> Explosions;
```
mcendu, line 14:
```suggestion
        public readonly Container Explosions;
```

Same for other occurences.

mcendu (changes requested):
A few nits

osu.Game.Rulesets.Mania/Skinning/LegacyHitExplosion.cs
smoogipoo: Good catch.

osu.Game.Rulesets.Mania/Skinning/LegacyHitExplosion.cs
```diff
@@ -0,0 +1,72 @@
...
+
+                d.Origin = Anchor.Centre;
+                d.Blending = BlendingParameters.Additive;
+                d.Scale = new Vector2(explosionScale);
+
+                if (!(d is TextureAnimation texAnimation))
+                    return;
+
+                if (texAnimation.FrameCount > 0)
+                    texAnimation.DefaultFrameLength = Math.Max(texAnimation.DefaultFrameLength, 170.0 / texAnimation.FrameCount);
```
peppy, line 48: isn't this too late? `DefaultFrameLength` only applies to frames which are added from this point forward.

smoogipoo: Ah, didn't realise. Would you be against adding another optional parameter to `GetAnimation()` for the frame length, alongside the current `applyConfigFrameRate`?

peppy: a similar thing seems to be required for sliderball too. i mean we already have 8 odd parameters so why not?

smoogipoo: Made the change, I don't think it's _too_ bad...

peppy (approved):

Issue #8019, opened 2020-02-27
Legacy skins should fallback to default classic skin elements when not available from skin


Koto1:
**Describe the bug:** 
My mix skin is not working well on laser for some reason, I know it used to work just fine around the time skins were implemented to laser and at some point, it broke, all of my other skins work great tho. I tried deleting all skins and importing back, and also importing by throwing the .osk on the laser and it still didn't work
**Screenshots or videos showing encountered issue:**
 
[skin.osk](https://drive.google.com/file/d/15WsHFoBSemZ0z0MkLQa49CrRkai2FbUN/view?usp=sharing)

[the skin on stable](http://prntscr.com/r87qjh)

[the skin on lazer](http://prntscr.com/r87rap)

**osu!lazer version:** 
2020.226.0
**Logs:**
*please attach logs here, which are located at:*
- `%AppData%/osu/logs` *(on Windows),*
- `~/.local/share/osu/logs` *(on Linux & macOS).*
[logs.zip](https://github.com/ppy/osu/files/4258817/logs.zip)

smoogipoo:
Can see what's going on here. This skin is expecting the default skin's `hitcircleoverlay.png` to be used, but it isn't contained within the skin.

xTerraUWU:
Not all elements from osu!stable work on osu!lazer if I remember. Could be hitcircleoverlay.png is not present when on the lazer client, but I could be wrong.

enoslayd:
Just a thought, what if you want a legacy skin with a mix of lazer default skin?

peppy:
You can't with legacy skinning – that's out of scope. That will be possibly with the future skinning system when we make that.

cl8n:
issue's title is a bit misleading--- the problem in OP's screenshots comes from hitcircleoverlay being forced to load from whichever skin has hitcircle, so if there's no hitcircleoverlay there it just won't show anything. correct behavior would have hitcircleoverlay load from a fallback skin

coincidentally the title correctly reports a different bug, that legacy skins are falling back to the "osu!lazer" skin instead of "osu!classic" for missing elements

peppy:
That's actually what the title is saying I think? The fallback to "default classic skin" is exactly what you're implying needs to be done.

cl8n:
[...]

PR #8562, opened 2020-04-02
Allow hold note tail to fallback to normal note image

### LINKED ISSUE #8560: mania missing elements doesnt fallback to default skin
Describe the bug: #8532 maybe related but it was closed. i have a skin where the long note is not skinned, only the head (which also skins the tail). it should fallback to the lazer (or classic???) skin but instead it shows invisible

osu_2020-04-02_14-12-25

osu!lazer version:2020.402.0

**Logs:**
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/4419185/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/4419186/performance-gamethread.draw.log)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/4419187/performance-gamethread.input.log)
[performance-gamethread.update.log](https://github.com/ppy/osu/files/4419189/performance-gamethread.update.log)
[runtime.log](https://github.com/ppy/osu/files/4419190/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4419191/updater.log)
[database.log](https://github.com/ppy/osu/files/4419192/database.log)
[performance.log](https://github.com/ppy/osu/files/4419195/performance.log)
[network.log](https://github.com/ppy/osu/files/4419196/network.log)

mcendu:
Fixes #8560

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyHoldNoteHeadPiece.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyHoldNoteHeadPiece.cs
@@ -10,6 +10,7 @@ namespace osu.Game.Rulesets.Mania.Skinning
     {
         protected override Texture GetTexture(ISkinSource skin)
         {
+            // TODO: Should fallback to the head from default legacy skin instead of note.
             return GetTextureFromLookup(skin, LegacyManiaSkinConfigurationLookups.HoldNoteHeadImage)
                    ?? GetTextureFromLookup(skin, LegacyManiaSkinConfigurationLookups.NoteImage);
         }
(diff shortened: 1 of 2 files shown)


smoogipoo:
Hmm as far as I can tell from stable, the head should never fall back to the note, and because of that neither should the tail.

That is, the implementation of the legacy head is also incorrect, but it exists because the head and the note are usually both identical. The fallback to the note type shouldn't exist, and both the head and tail should fallback to the head from the classic skin (https://github.com/ppy/osu/issues/8019).

I'd like a todo to be added here so we don't forget to remove it in the future.

smoogipoo (approved):

PR #8566, opened 2020-04-02
Fix crash caused by user json order changing

peppy:
See https://github.com/ppy/osu-web/pull/5863 for paper trail. JSON deserialization should not be order dependent.

### DIFF
--- a/osu.Game/Users/User.cs
+++ b/osu.Game/Users/User.cs
@@ -173,8 +173,27 @@ namespace osu.Game.Users
             public int Available;
         }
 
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

bdach:
Ouch, that's is not great. While this will work for now as a hotfix, I'm thinking this should maybe be refactored at some point so that the lazer-side model actually corresponds with what web returns one-to-one...?

peppy:
Yeah, that's the plan.

PR #8568, opened 2020-04-02
Fix weird slider ball sizing

### LINKED ISSUE #8550: Animated slider balls have zero size
**Describe the bug:** Animations of slider balls (if provided by skin) are not visible; the slider ball altogether is clipped in size.

**Screenshots or videos showing encountered issue:** [See video](https://streamable.com/1zvbhy) - probably the best way to describe this.

**osu!lazer version:** 2020.402.0, in particular seems to be anything since #8462 and its associated framework bump

**Logs:** Not relevant here.

<hr>

As for the investigation: it seems that this has regressed in a non-obvious way after the framework-side sprite animation changes with inclusion of clocks. In short:

* When the slider ball is being constructed in the skin transformer here:

  https://github.com/ppy/osu/blob/e183b2b14d163c8af3d5dd08669958dbb4b8ebd4/osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs#L58-L59

  the returned animation has zero size, contrary to the size of the first frame it used to have before this (or, at least in 2020.327.0).
* A few lines down, the transformer attempts to use first frame size to size the entire ball, but fails since the animated content has zero size:

  https://github.com/ppy/osu/blob/e183b2b14d163c8af3d5dd08669958dbb4b8ebd4/osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs#L64-L75

  Having set zero, the transformer then sets `RelativeSizeAxes = Axes.Both` on the animated content.
* Framework-side, `Animation` normally sets its size when the animation progresses along with the clock, but due to `RelativeSizeAxes` set to `Both` as above [does nothing](https://github.com/ppy/osu-framework/blob/7ce2d278163e4eeac15a57f33404ab0fe2272747/osu.Framework/Graphics/Animations/Animation.cs#L273-L282) and retains zero size.

Ran out of time to try to figure out and submit a fix myself today. FWIW, setting `AutoSize = Both` on the slider ball seemed to help but I'm not convinced of the correctness of that solution.

peppy:
Closes https://github.com/ppy/osu/issues/8550.

Tested against around 20 skins and this does not look to regress behaviour at all. Not sure why we have such complicated logic previously..


smoogipoo (approved):
It was probably to scale to fit? Looks fine though.

PR #8576, opened 2020-04-02
Attempt to fix multiple Spinner Spm overlaps

### LINKED ISSUE #7023: Multiple spin rate counters visible for consecutive spinners
(This isn't really a BUG, just a visual that looks off.)

**Describe the bug:** 
When a beatmap has multiple spinners next to each other, the SPM (Spins Per Minute) counter on the next spinner starts to appear.

osu_2019-11-30_14-48-32

**osu!lazer version:** 
2019.1128.0

**Logs:** 
As this is not technically a bug, no logs can be found.

kornkaobat:
Attempt to fix #7023. Tests and improvements will be evaluated on Tewi Spinners map.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
@@ -36,8 +36,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 
         private readonly SpriteIcon symbol;
 
-        private readonly Color4 baseColour = OsuColour.FromHex(@"002c3c");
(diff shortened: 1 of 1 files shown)


peppy:
i don''''''''t think this will work. did you actually test this?

kornkaobat:
I did. It does work for some cases but others not. The code needs some tweaking but I think the general idea is in the right direction.

smoogipoo:
Sorry but it's fairly obvious you haven't even done the most simple of tests for this. Please only PR when your code is actually ready for review.

The code makes no sense - you're checking a variable that can't possibly be false, for false, and you're setting a variable to true in all cases. This PR changes nothing functionally.

PR #8577, opened 2020-04-02
Add support for falling back on skin configuration settings

frenzibyte:
This adds a `SkinSettingFallback` attribute to assign to enum members for providing a fallback value in case the actual setting does not have a value assigned to it.

This will pretty much help retrieving a skin configuration setting without having to take care of what to do if the setting does not have a value **in every single retrieval**. with this new attribute, you can set the fallback value / fallback setting directly to the enum member:
## Current way:
```csharp
// in some file
var value = skin.GetConfig<Configuration, string>(Configuration.TestSetting)?.Value ??
            skin.GetConfig<Configuration, string>(Configuration.FallbackSetting)?.Value ??
            "fallback value";
...

// in another file
var value = skin.GetConfig<Configuration, string>(Configuration.TestSetting)?.Value ??
            skin.GetConfig<Configuration, string>(Configuration.FallbackSetting)?.Value ??
            "fallback value";
...

public enum Configuration
{
    TestSetting,
    FallbackSetting,
}
```
(existing case for osu!catch skin configuration value retrieval: `HyperDashFruit` -> `HyperDash` -> `#FF00000`)
## With the `SkinSettingFallback` attribute:
```csharp
// in some file
var value = skin.GetConfigOrDefault<Configuration, string>(Configuration.TestSetting).Value;
...

// in another file
var value = skin.GetConfigOrDefault<Configuration, string>(Configuration.TestSetting).Value;
...

public enum Configuration
{
    [SkinSettingFallback(SkinSettingFallbackType.Setting, FallbackSetting)]
    TestSetting,

    [SkinSettingFallback(SkinSettingFallbackType.Value, "fallback value")]
    FallbackSetting,
}
```

## Remarks
- Usage will come in a separate PR just to avoid complexity of reviewing in case it does.
- I was unsure of the name `GetConfigOrDefault` but didn't want to waste my time looking for better naming so raising it up for the reviewers and the team to decide.
- Hopefully the `TestConfiguration` & `FallbackConfiguration` enum member names in `SkinSettingFallbackTest` isn't that bad as i've tried to let it make sense but not make it stupidly long.

### DIFF
--- a/osu.Game.Tests/Skins/SkinSettingFallbackTest.cs
+++ b/osu.Game.Tests/Skins/SkinSettingFallbackTest.cs
@@ -1,0 +1,139 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Game.Skinning;
+
+namespace osu.Game.Tests.Skins
+{
+    [TestFixture]
+    public class SkinSettingFallbackTest
+    {
+        private readonly ISkin skin = new TestSkin();
+
+        /// <summary>
+        /// Tests behaviour of looking for value of a setting (<see cref="TestConfiguration.ValueFBSetting"/>) that has a fallback "default" value.
+        /// </summary>
+        [Test]
+        public void TestSettingWithDefaultValue()
+        {
+            var configSetting = skin.GetConfigOrDefault<TestConfiguration, string>(TestConfiguration.ValueFBSetting);
+
+            Assert.AreEqual("default value", configSetting.Value);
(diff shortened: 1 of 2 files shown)


peppy:
already said this when you discussed implementing it, but i *really* dislike this in terms of readability.

bdach:
Agree, it feels overengineered to what this aims to really do and also really narrow in terms of what this *can* do. I think we should be doing values with fallbacks as plain, well-named and well-defined extension methods.

smoogipoo:
I think it may be best to wait until we have attribute-based configs in the first place, since that’s going to change the structure of this a bit more.

iiSaLMaN:
Closing for the above reasons

PR #8584, opened 2020-04-03
Fix GetDecoder getting fallback decoder too often

ThereGoesMySanity:
Changes:
 * Fixed issue where `GetDecoder` would return the fallback decoder if the correct decoder wasn't first in the Dictionary.
 * Added test to make sure the JsonBeatmapDecoder is returned when calling GetDecoder<Beatmap> with a JSON beatmap file.

### DIFF
--- a/osu.Game.Tests/Beatmaps/Formats/OsuJsonDecoderTest.cs
+++ b/osu.Game.Tests/Beatmaps/Formats/OsuJsonDecoderTest.cs
@@ -127,6 +127,31 @@ namespace osu.Game.Tests.Beatmaps.Formats
                 .Assert();
         }
 
+        [Test]
+        public void TestGetJsonDecoder()
+        {
+            Decoder<Beatmap> decoder;
+
+            using (var stream = TestResources.OpenResource(normal))
+            using (var sr = new LineBufferedReader(stream))
+            {
+                var legacyDecoded = new LegacyBeatmapDecoder { ApplyOffsets = false }.Decode(sr);
+
+                using (var ms = new MemoryStream())
+                using (var sw = new StreamWriter(ms))
+                using (var sr2 = new LineBufferedReader(ms))
+                {
+                    sw.Write(legacyDecoded.Serialize());
+                    sw.Flush();
+
+                    ms.Position = 0;
+                    decoder = Decoder.GetDecoder<Beatmap>(sr2);
+                }
+            }
+
+            Assert.IsInstanceOf(typeof(JsonBeatmapDecoder), decoder);
+        }
+
         /// <summary>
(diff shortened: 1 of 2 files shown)


ThereGoesMySanity:
I think the failed check is wrong - InspectCode thinks that I'm calling `(T?).Value` even though I'm actually calling `(KeyValuePair<>).Value`, maybe?

Otherwise, I don't understand what the message is supposed to mean.

osu.Game.Tests/Beatmaps/Formats/OsuJsonDecoderTest.cs
```diff
@@ -127,6 +127,31 @@ public void TestParity(string beatmap)
...
+            Decoder<Beatmap> decoder;
+
+            using (var stream = TestResources.OpenResource(normal))
+            using (var sr = new LineBufferedReader(stream))
+            {
+                var legacyDecoded = new LegacyBeatmapDecoder { ApplyOffsets = false }.Decode(sr);
+
+                using (var ms = new MemoryStream())
+                using (var sw = new StreamWriter(ms))
+                using (var sr2 = new LineBufferedReader(ms))
```
mcendu, line 142: The variable names are cryptic.

osu.Game/Beatmaps/Formats/Decoder.cs
```diff
@@ -63,7 +63,7 @@ public static Decoder<T> GetDecoder<T>(LineBufferedReader stream)
             if (line == null)
                 throw new IOException("Unknown file format (null)");
 
-            var decoder = typedDecoders.Select(d => line.StartsWith(d.Key, StringComparison.InvariantCulture) ? d.Value : null).FirstOrDefault();
+            var decoder = typedDecoders.Where(d => line.StartsWith(d.Key, StringComparison.InvariantCulture)).FirstOrDefault().Value;
```
mcendu: Why is `.Value` there?

mcendu (changes requested):

osu.Game/Beatmaps/Formats/Decoder.cs
ThereGoesMySanity: We're selecting all of the KeyValuePairs where the key matches the file we're parsing and then getting the value of the first pair that matches.

ThereGoesMySanity: The latest change should make that a bit clearer (and hopefully get rid of the weird InspectCode message)

ThereGoesMySanity:
There's some sort of memory leak that happens when JSON beatmaps are loaded now, but I think that's an unrelated issue

smoogipoo (approved):

PR #8598, opened 2020-04-03
Add osu!taiko drum skinning support

peppy:
The first piece of taiko skinning support.


smoogipoo (approved):
Although the playing of the sounds should not really be a part of the skinnable components (i.e. in when we have non-legacy skins), let's go with it for now.

PR #8602, opened 2020-04-03
Support HitCircleOverlayAboveNumer typo for old legacy skins

LittleEndu:


peppy:
can you do this i the transformer instead? and comment a link referencing where this is documented (on the wiki or otherwise)

mcendu:
[Per wiki:](https://osu.ppy.sh/help/wiki/Skinning/skin.ini#%5Bgeneral%5D)

> Old command: `HitCircleOverlayAboveNumer` (with typo) still works for legacy support

peppy:
Yep, that link should be inline in the code as a reference!

LittleEndu:
> can you do this in the transformer instead?

You mean replace the typo when the skin is first imported?

mcendu:
> 
> 
> > can you do this in the transformer instead?
> 
> You mean replace the typo when the skin is first imported?

imo in ```ManiaLegacySkinTransformer.GetConfig`2``` when the entry without typo isn't found, we should try to get the entry with typo instead of returning null. This can make code that needs to check this property look cleaner.

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -132,6 +132,12 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                                 return SkinUtils.As<TValue>(new BindableFloat(LEGACY_CIRCLE_RADIUS));
 
                             break;
+
+                        case OsuSkinConfiguration.HitCircleOverlayAboveNumber:
+                            // Quote from https://osu.ppy.sh/help/wiki/Skinning/skin.ini#%5Bgeneral%5D
+                            // Old command: HitCircleOverlayAboveNumer (with typo) still works for legacy support
+                            var rv = source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumber);
```
mcendu, line 139: Could be inlined.

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -132,6 +132,12 @@ public Drawable GetDrawableComponent(ISkinComponent component)
...
+                            return rv ?? source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumer);
```
LittleEndu, line 140:
```suggestion
                            return source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumber) ?? source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumer);
                            
```

@mcendu You mean like this?

bdach: I'd say that's better (also gets rid of the very cryptic `rv` variable name).

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -136,7 +136,7 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                         case OsuSkinConfiguration.HitCircleOverlayAboveNumber:
                             // Quote from https://osu.ppy.sh/help/wiki/Skinning/skin.ini#%5Bgeneral%5D
                             // Old command: HitCircleOverlayAboveNumer (with typo) still works for legacy support
-                            return source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumber) ?? source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumer);                       
+                            return source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumber) ?? source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumer);
```
mcendu, line 139: That’s a _long_ line. Better put a new line before the `??`.

peppy (approved):

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -132,6 +132,12 @@ public Drawable GetDrawableComponent(ISkinComponent component)
                                 return SkinUtils.As<TValue>(new BindableFloat(LEGACY_CIRCLE_RADIUS));
 
                             break;
+
+                        case OsuSkinConfiguration.HitCircleOverlayAboveNumber:
+                            // See https://osu.ppy.sh/help/wiki/Skinning/skin.ini#%5Bgeneral%5D
+                            // HitCircleOverlayAboveNumer (with typo) should still be supported for now.
+                            return source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumber) ??
+                                   source.GetConfig<OsuSkinConfiguration, TValue>(OsuSkinConfiguration.HitCircleOverlayAboveNumer);
```
smoogipoo, line 140: Not a big deal but in cases like this I would rather send down a string in the first place, i.e. `source.GetConfig<string, TValue>("HitCircleOverlayAboveNumer");`, as it's a hack that shouldn't exist in the config enum.

PR #8607, opened 2020-04-03
Load user rulesets from the game data directory

### LINKED ISSUE #8545: Add the ability to transfer custom rullsets when updating the application
When updating, they are not saved and are in the folder with the old patch

I think that in the settings you can add a transfer button. Or enable automatic transfer

Game4all:
# Summary

This allows storing the user rulesets under the rulesets folder in the game data directory (as an interim fix in the wait for rulesets to get databased) and get them picked up by the `RulesetStore` (`loadUserRulesets()`).

The runtime won't be able to load dependencies for assemblies located outside of the executable directory, thus manual referencing to the game core assembly using .NET Standard AppDomain resolution mechanism has been added to `RulesetStore`
This assumes the only dependency custom rulesets have is on the game core, other transitive assembly references (framework) are resolved automatically by the runtime from the resolved game core assembly.

# Remarks

While messing with AppDomain assembly resolution in `RulesetStore`, I stumbled upon this line which I fail to understand its effect.

```csharp
private Assembly resolveRulesetAssembly(object sender, ResolveEventArgs args) => loadedAssemblies.Keys.FirstOrDefault(a => a.FullName == args.Name);
```

After testing without this line, I can confirm everything still works fine. So I suppose it is fine to remove that line as I did (unless I missed something on the reason of existence for this line?)

No automated tests for this as it is quite hard to automatically test this.
I tested the behaviour on my machine with the tau ruleset and can confirm it works.

Closes https://github.com/ppy/osu/issues/8545


osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -48,7 +53,17 @@ public RulesetStore(IDatabaseContextFactory factory)
         /// </summary>
         public IEnumerable<RulesetInfo> AvailableRulesets { get; private set; }
 
-        private Assembly resolveRulesetAssembly(object sender, ResolveEventArgs args) => loadedAssemblies.Keys.FirstOrDefault(a => a.FullName == args.Name);
```
peppy, line 51: it looks like this was here to cache assembly lookups. have you checked what happens (or even if it happens) when a ruleset assembly is resolved the second time?

Game4all: Ah, This makes sense now. 
However, after testing on master, this line doesn't have any effect effect because the event handler is attached after actual the assembly lookup is done (the event is never raised)
As of having a ruleset assembly being resolved twice; I don't have any scenario in mind where this could happen (apart from loading the same ruleset assembly but with different filenames, 
which is handled by the ruleset store)? Looks like I'm missing something.

peppy: I'm not sure myself, but the first thing i'd do is git blame and see if there are any hints.

Game4all: After looking at the class commit log, that was added back in f2a9e95d.
Gonna add the loaded assembly lookup back as it seems to allow resolving assemblies without having full assembly info matching.

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -120,6 +135,21 @@ private void loadFromAppDomain()
             }
         }
 
+        private void loadUserRulesets()
+        {
+            try
```
smoogipoo, line 140: Probably should be done over each individual ruleset instead, and the error should contain the ruleset's name. No reason to fail loading all other rulesets if one's broken.

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -17,16 +18,20 @@ public class RulesetStore : DatabaseBackedStore, IDisposable
...
         private readonly Dictionary<Assembly, Type> loadedAssemblies = new Dictionary<Assembly, Type>();
 
-        public RulesetStore(IDatabaseContextFactory factory)
+        private readonly Storage rulesetStorage;
+
+        public RulesetStore(IDatabaseContextFactory factory, Storage storage = null)
             : base(factory)
         {
+            rulesetStorage = storage?.GetStorageForDirectory("rulesets");
+            AppDomain.CurrentDomain.AssemblyResolve += resolveRulesetDependencyAssembly;
```
smoogipoo, line 27: Is there a reason why you moved this to the top?

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -48,7 +53,17 @@ public RulesetStore(IDatabaseContextFactory factory)
         /// </summary>
         public IEnumerable<RulesetInfo> AvailableRulesets { get; private set; }
 
-        private Assembly resolveRulesetAssembly(object sender, ResolveEventArgs args) => loadedAssemblies.Keys.FirstOrDefault(a => a.FullName == args.Name);
+        private Assembly resolveRulesetDependencyAssembly(object sender, ResolveEventArgs args)
+        {
+            var asm = new AssemblyName(args.Name);
+
+            // this assumes the only explicit dependency of the ruleset is the game core assembly.
+            // the ruleset dependency on the game core assembly requires manual resolving, transient dependencies should be resolved automatically
```
smoogipoo, line 61: So this is because the rulesets are in a subfolder, right? Might as well mention that.

smoogipoo (changes requested):
Seems ok otherwise.

osu.Game/Rulesets/RulesetStore.cs
Game4all: This may not be needed actually, as the exception is catched internally in `loadRulesetFromFile()`

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -17,16 +18,20 @@ public class RulesetStore : DatabaseBackedStore, IDisposable
...
+        public RulesetStore(IDatabaseContextFactory factory, Storage storage = null)
             : base(factory)
         {
+            rulesetStorage = storage?.GetStorageForDirectory("rulesets");
+
             // On android in release configuration assemblies are loaded from the apk directly into memory.
             // We cannot read assemblies from cwd, so should check loaded assemblies instead.
             loadFromAppDomain();
             loadFromDisk();
+            AppDomain.CurrentDomain.AssemblyResolve += resolveRulesetDependencyAssembly;
```
smoogipoo, line 32: As I commented before, is there a reason why this is done in the specific place it's being done here? Rather than at the bottom where it was initially?

Game4all: The event handler contains code for resolving dependency on the game assembly for Rulesets located outside the base game directory. 

It needs to be attached  to the lookup event before the actual call to `loadUserRulesets()` else Rulesets located out the base game directory will fail to load as unable to locate the game assemblies.

smoogipoo: Gotcha, thanks.

bdach: Maybe could have a comment here why this needs to be before the user ruleset load?

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -48,7 +53,18 @@ public RulesetStore(IDatabaseContextFactory factory)
...
         public IEnumerable<RulesetInfo> AvailableRulesets { get; private set; }
 
-        private Assembly resolveRulesetAssembly(object sender, ResolveEventArgs args) => loadedAssemblies.Keys.FirstOrDefault(a => a.FullName == args.Name);
+        private Assembly resolveRulesetDependencyAssembly(object sender, ResolveEventArgs args)
+        {
+            var asm = new AssemblyName(args.Name);
+
+            // the requesting assembly may be located out of the executable's base directory, thus requiring manual resolving of its dependencies.
+            // this assumes the only explicit dependency of the ruleset is the game core assembly.
+            // the ruleset dependency on the game core assembly requires manual resolving, transient dependencies should be resolved automatically
```
bdach, line 62: I think this means to say
```suggestion
            // the ruleset dependency on the game core assembly requires manual resolving, transitive dependencies should be resolved automatically
```

bdach (commented):
I tested this today. It does seem to work with pre-built user rulesets but *only* after master is merged here (otherwise loading fails due to framework transitive dependency version mismatch).

Some comments but not supremely relevant, just mostly cosmetics.

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -48,7 +57,18 @@ public RulesetStore(IDatabaseContextFactory factory)
...
 
-        private Assembly resolveRulesetAssembly(object sender, ResolveEventArgs args) => loadedAssemblies.Keys.FirstOrDefault(a => a.FullName == args.Name);
+        private Assembly resolveRulesetDependencyAssembly(object sender, ResolveEventArgs args)
+        {
+            var asm = new AssemblyName(args.Name);
+
+            // the requesting assembly may be located out of the executable's base directory, thus requiring manual resolving of its dependencies.
+            // this assumes the only explicit dependency of the ruleset is the game core assembly.
+            // the ruleset dependency on the game core assembly requires manual resolving, transitive dependencies should be resolved automatically
+            if (asm.Name.Equals(typeof(OsuGame).Assembly.GetName().Name, StringComparison.Ordinal))
```
smoogipoo, line 67: Loading the tau ruleset (https://github.com/Altenhh/tau) via the `rulesets/` directory fails on this branch. Loading via placing it alongside all other game executables works.
Seems like the reason is it's expecting an exact match for transitive dependencies, which will almost never be the case (frequent o!f updates).

Perhaps we don't want to have this automatic transitive dependency matching, and instead look for all local assemblies as an override to ruleset-provided ones, via:

```csharp
foreach (var curAsm in AppDomain.CurrentDomain.GetAssemblies())
{
    if (asm.Name.Equals(curAsm.GetName().Name, StringComparison.Ordinal))
        return curAsm;
}
```

smoogipoo (changes requested):

osu.Game/Rulesets/RulesetStore.cs
smoogipoo: That's exactly it, yes. But it'll fail the other way around too - with tau referencing an earlier version of o!f and osu! referencing a more up-to-date one.

bdach: The weird thing is that I also tested with another pre-built ruleset release ([maimai](https://github.com/LumpBloom7/osu-maimai) 2020.413.0) and it didn't fail, even though [it technically references an older version of framework than the game itself did](https://github.com/LumpBloom7/osu-maimai/blob/a5b221ba4867d9261b2a403686647a56879a8bb5/osu.Game.Rulesets.Maimai/osu.Game.Rulesets.Maimai.csproj#L21)... Probably needs more investigation/thought either way.

Game4all: The code path was assuming that the only explicit dependency was on the game core and so the resolved framework assembly version was the same as the game core but it isn't actually the case at all (looks like I took a shortcut in my thinking) as you pointed out , as rulesets can depend on a older framework version than the game depends on; while your proposed code path works fine (tested this with recent releases of tau, bosu, maimai and mvis rulesets),  there should be version checks for game version / framework version  added on top to handle ABI breaking changes, else game will crash when using the rulesets as there is a ABI mismatch as shown here:

image

smoogipoo: Ouch. I think that's fine, but it should stop the load of gameplay and throw a notification instead of crashing the game though. Can be done in a further PR.

Exception handling should probably exist everywhere where a ruleset is used, I'd say.

Game4all: In that case , will open a PR soon to mitigate this.

smoogipoo (approved):
Looks good to me!

osu.Game/Rulesets/RulesetStore.cs
```diff
@@ -17,16 +18,24 @@ public class RulesetStore : DatabaseBackedStore, IDisposable
 
         private readonly Dictionary<Assembly, Type> loadedAssemblies = new Dictionary<Assembly, Type>();
 
-        public RulesetStore(IDatabaseContextFactory factory)
+        private readonly Storage rulesetStorage;
+
+        public RulesetStore(IDatabaseContextFactory factory, Storage storage = null)
             : base(factory)
         {
+            rulesetStorage = storage?.GetStorageForDirectory("rulesets");
```
peppy, line 26: does this create the directory automatically?

peppy: ah yes, it does, all good.

osu.Game/Rulesets/RulesetStore.cs
Game4all: Actually, this may not be as trivial as I thought. Fixing this may require a quite a bit of investment in addings guards where `DrawableRuleset`s are manipulated as well as a bit of thinking since some pieces of the `DrawableRuleset` aren't created directly in `Ruleset.CreateDrawableRulesetWith()` but lazily created (namely `Playfield`), where errors could pop up without the ability to catch them.

PR #8613, opened 2020-04-04
Make legacy skins use startAtCurrentTime by default

### LINKED ISSUE #8604: Animated judgements are missing
**Describe the bug:** 
All expected judgements (100s and 50s) are missing from my skin ([LittleEndu/osuSkin](https://github.com/LittleEndu/osuSkin/tree/master/u200B)). This probably has something to do with animations, but I wasn't able to recreate this with any other skin I have :/

**osu!lazer version:** 
2020.404.0

LittleEndu:
Fixes #8604

Unsure why this wasn't enabled by default while every other implementation has it enabled (SkinnableTextureAnimation and its base TextureAnimation).

### DIFF
--- a/osu.Game/Skinning/LegacySkinExtensions.cs
+++ b/osu.Game/Skinning/LegacySkinExtensions.cs
@@ -14,7 +14,7 @@ namespace osu.Game.Skinning
     public static class LegacySkinExtensions
     {
         public static Drawable GetAnimation(this ISkin source, string componentName, bool animatable, bool looping, bool applyConfigFrameRate = false, string animationSeparator = "-",
-                                            bool startAtCurrentTime = false, double? frameLength = null)
+                                            bool startAtCurrentTime = true, double? frameLength = null)
         {
             Texture texture;
 
--- a/osu.Game/Skinning/LegacySkinExtensions.cs
+++ b/osu.Game/Skinning/LegacySkinExtensions.cs
@@ -72,7 +72,7 @@ namespace osu.Game.Skinning
                 if (timeReference != null)
                 {
                     Clock = timeReference.Clock;
-                    PlaybackPosition = timeReference.AnimationStartTime - timeReference.Clock.CurrentTime;
+                    PlaybackPosition = timeReference.Clock.CurrentTime - timeReference.AnimationStartTime;
                 }
             }
         }
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyHitExplosion.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyHitExplosion.cs
@@ -39,7 +39,7 @@ namespace osu.Game.Rulesets.Mania.Skinning
             if (tmp is IFramedAnimation tmpAnimation && tmpAnimation.FrameCount > 0)
                 frameLength = Math.Max(1000 / 60.0, 170.0 / tmpAnimation.FrameCount);
 
-            explosion = skin.GetAnimation(imageName, true, false, startAtCurrentTime: true, frameLength: frameLength).With(d =>
+            explosion = skin.GetAnimation(imageName, true, false, frameLength: frameLength).With(d =>
             {
                 if (d == null)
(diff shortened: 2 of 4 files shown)


peppy:
Have you checked each usage of this to make sure there are not any regressions?

LittleEndu:
I haven't checked all use cases, but are there even any places where legacy animations shouldn't start from the beginning?

bdach:
Animations have been a little bit of a pain over the last few days when it comes to regressions. It would be prudent to check just in case to prevent further back-and-forth.

LittleEndu:
So checking would mean that one would use a skin where every animatable element is animated and check if that animation plays as expected (compared to stable I would guess)

bdach:
Pretty much, yes.

Let's put it this way: the way I see it, with every PR the burden of proof that the change is correct and has no regressions lies on the *author*. Of course, not everything can ever be checked, and things will slip by, but checking all call sites when making a default value change such as this is pretty much expected procedure.

peppy:
@LittleEndu for reference this broke follow points, at very least, which is why we need to check these changes properly.

LittleEndu:
I'm would probably still like to make an easy test scene that let's you look at all the animations even if this gets fixed. There's still the issue with lazer applying transforms to animations while stable doesn't which would also need someone to look into all the animations case by case.

peppy:
If being 100% honest, any gameplay element which supports rewinding is *incorrect* using `startAtCurrentTime`, but we don't have an easy fix for this. Let me try to explain this just for your own knowledge:

Let's use the example of hit explosions (judgements).

- Seek beyond a hitobject and start rewinding backwards until it comes into view.
- The `LoadComplete` will trigger on the judgement and it will read in the current time, as a time value far beyond when the judgement actually occurred.
- This value will be used as the "start" time of the animation which is actually incorrect.

Currently this is weakly "handled" by the fact that gameplay always runs through all frames (see `FrameStabilityContainer`), which is why the `FollowPoint` issue is the only one we've fixed so far (using `IAnimationTimeReference`) – follow points are visible in the editor and therefore need the special handling as the editor allows seeking to arbitrary points in time *without* playback of frames up to that point.

A proper solution may be for the `GetAnimation` to take an enum like this:

```csharp
public enum AnimationStartPoint
{
    JudgementTime
    PreemptTime,
    CurrentTime
}
```

Then in the implementation it could DI the `DrawableHitObject` and set the reference time as required.

That said, I'm kinda against adding this logic until we need to as the underlying components may change along the way.

@LittleEndu let me know if the above is understandable to you, since it's a bit of a complicated issue.

LittleEndu:
I'll have to get back to that after work. It looks like I understand but I doubt I would be to implement the fix.

peppy:
As I mentioned, I don't think we need a fix just yet. For the time being `FrameStabilityContainer` is handling it for us and it's best to just keep things simple.

peppy:
Checked remaining usages. This looks good for now.

osu.Game/Skinning/LegacySkinExtensions.cs
```diff
@@ -72,7 +72,7 @@ protected override void LoadComplete()
                 if (timeReference != null)
                 {
                     Clock = timeReference.Clock;
-                    PlaybackPosition = timeReference.AnimationStartTime - timeReference.Clock.CurrentTime;
+                    PlaybackPosition = timeReference.Clock.CurrentTime - timeReference.AnimationStartTime;
```
peppy, line 75: @smoogipoo requesting you sanity check this – it was back-to-front before

peppy (approved):

osu.Game/Skinning/LegacySkinExtensions.cs
smoogipoo: I think this is correct, if the clock is 100ms behind the animation start, this will be -100, and will take 100ms to elapse before the animation starts.

smoogipoo (approved):

PR #8620, opened 2020-04-04
Display control points on timeline

Alchyr:
For https://github.com/ppy/osu/projects/2#card-33198852

### Notes:
This is based almost entirely off of how control points are displayed in the summary timeline, with small changes to origin/anchoring for positioning.
Timing points are displayed in the bottom half of the timeline, and inherited points are displayed on the top half. 
Example

This does have the same issue as #7634 which causes slowdown during playback when there are many lines.


osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineControlPointDisplay.cs
```diff
@@ -0,0 +1,79 @@
...
+
+        public TimelineControlPointDisplay()
+        {
+            RelativeSizeAxes = Axes.Both;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            //Bind to changes to control points
```
bdach, line 44: Why's this comment here? Is this PR incomplete?

bdach (commented):
You're updating your fork incorrectly (see the three merge commits at the start, you should not be merging in upstream changes via opening PRs). See the [*Syncing a fork*](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) github article for instructions as to how to do this properly.

As for the first impressions on the design of the display on provided screenshot - I'm not a mapper but UI-wise it looks really unintuitive.

osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineControlPointDisplay.cs
```diff
@@ -0,0 +1,79 @@
...
+
+            cpi.TimingPoints.ForEach(addTimingPoint);
+
+            // Consider all non-timing points as the same type
+            cpi.SamplePoints.Select(c => (ControlPoint)c)
+               .Concat(cpi.EffectPoints)
+               .Concat(cpi.DifficultyPoints)
+               .Distinct()
+               // Non-timing points should not be added where there are timing points
+               .Where(c => cpi.TimingPointAt(c.Time).Time != c.Time)
```
bdach, line 29: This looks like a potentially expensive quadratic computation. Possibly worth to have a look if this can be optimised further.

Alchyr:
Having looked through the code (of lazer) in more detail, I've had to come to the conclusion that this pull request was poorly thought out. I likely shouldn't have gone for something that didn't have an issue created for it.

Thank you for taking the time to review this, and I apologize for wasting your time.

The card itself states that timing control points should be displayed, which would only be those that change the timing, so displaying other timing points was likely unnecessary.

Thank you for the instructions on properly syncing a fork. I'm lacking in any experience, so regardless of what I'm doing I tend to feel that my knowledge is lacking.

bdach:
No apologies required. The editor in general is an area in which mostly the core team operates at this point, even I'm not really clued into what the general plans going forward are. So while displaying control points definitely *needs* to happen, doing so through community contributions is susceptible to being out-of-sync with the planned ideas and general direction for the editor.

peppy:
Thanks for the contribution. For UI elements which don't have a design yet, please open an issue and ask for a design so we can guide you in the right direction. As already mentioned, copying the lower timeline feels a bit unintuitive as we already use the "vertical line with colour" for a completely different purpose. I believe we do already have a design direction for showing timing points.

PR #8624, opened 2020-04-05
Fix results star rating display not being centered when no mods are present

### LINKED ISSUE #8599: Star Rating is not centered in result screen when there are no mods
container
star rating

(Probably low priority)

Joehuu:
Fixes https://github.com/ppy/osu/issues/8599.

Actually this is also misaligned with mods (fix in first commit), but made title represent issue.

Test scene uses `Position` while hud overlay uses `Margin`, not sure what's correct in these top right "padding" cases.

### DIFF
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneModSelectOverlay.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneModSelectOverlay.cs
@@ -70,7 +70,7 @@ namespace osu.Game.Tests.Visual.UserInterface
                     Anchor = Anchor.TopRight,
                     Origin = Anchor.TopRight,
                     AutoSizeAxes = Axes.Both,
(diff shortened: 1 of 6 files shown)


bdach (dismissed):
Looks fine as far as I can tell. I looked through all usages of `ModDisplay` and none of them looks broken. Would have been nice to have a test step for the misalignment in the test scene but alas, not a big issue (not something that's really assertable anyway).

As a framework consideration, maybe `FillFlowContainer` shouldn't be applying spacing to a zero-width drawable (that's why the mod display is added outside of the initialiser structure)? Tough to say. Either way seems like it has its benefits.

peppy:
I'd say that would be a good framework addition (`AddSpacingToZeroSizeElements = true` in `FillFlowContainer`)

smoogipoo (approved):
Looks good to me.

PR #8626, opened 2020-04-05
Don't allow new transformations for reverse arrow after it's hit

### LINKED ISSUE #8618: DrawableSliderRepeat and ReverseArrowPiece easings shouldn't be additive
**Describe the bug:** 
Depending on where in the beat a reverse arrow is hit, it's easing will look super wrong (sorta like waves that constructively interfere, both DrawableSliderRepeat and ReverseArrowPiece's easings will be added together)

Was unable to capture this. Easy to look on [Mata Ashita's insane difficulty](https://osu.ppy.sh/beatmapsets/172055#osu/443272) as the first repeating slider happens few seconds in and the beat syncs with when you would hit it. 

**osu!lazer version:** 
2020.404.0

LittleEndu:
Fixes #8618

Before, reverse arrow would quickly scale up while drawable repeat was also scaling up. This looked weird when drawable repeat's scaling synced up with reverse arrow's scaling, almost like as if wrong easing was used (one that ramps up quickly like ``InExpo`` instead of ``Out``). 

This PR ~~removes pending transformations on the arrow when the repeat is hit and~~ doesn't allow new transformations for the arrow to be applied.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
+using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Graphics;
 using osu.Framework.Utils;
 using osu.Game.Rulesets.Objects.Drawables;
@@ -21,7 +22,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 
         private double animDuration;
 
-        private readonly Drawable scaleContainer;
+        private readonly ReverseArrowPiece scaleContainer;
 
         public DrawableSliderRepeat(SliderRepeat sliderRepeat, DrawableSlider drawableSlider)
             : base(sliderRepeat)
@@ -79,6 +80,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 case ArmedState.Hit:
                     this.FadeOut(animDuration, Easing.Out)
                         .ScaleTo(Scale * 1.5f, animDuration, Easing.Out);
+                    scaleContainer.ShouldFollowBeats = false;
+                    scaleContainer.Transforms.ForEach(t => scaleContainer.RemoveTransform(t));
                     break;
             }
         }
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
@@ -22,7 +22,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 
         private double animDuration;
 
-        private readonly ReverseArrowPiece scaleContainer;
+        private readonly Drawable scaleContainer;
 
         public DrawableSliderRepeat(SliderRepeat sliderRepeat, DrawableSlider drawableSlider)
             : base(sliderRepeat)
@@ -80,7 +80,6 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 case ArmedState.Hit:
                     this.FadeOut(animDuration, Easing.Out)
                         .ScaleTo(Scale * 1.5f, animDuration, Easing.Out);
-                    scaleContainer.ShouldFollowBeats = false;
                     scaleContainer.Transforms.ForEach(t => scaleContainer.RemoveTransform(t));
                     break;
             }
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
@@ -5,7 +5,6 @@ using System;
 using System.Collections.Generic;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
-using osu.Framework.Extensions.IEnumerableExtensions;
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ReverseArrowPiece.cs
```diff
@@ -37,7 +39,10 @@ public ReverseArrowPiece()
             };
         }
 
-        protected override void OnNewBeat(int beatIndex, TimingControlPoint timingPoint, EffectControlPoint effectPoint, TrackAmplitudes amplitudes) =>
-            Child.ScaleTo(1.3f).ScaleTo(1f, timingPoint.BeatLength, Easing.Out);
+        protected override void OnNewBeat(int beatIndex, TimingControlPoint timingPoint, EffectControlPoint effectPoint, TrackAmplitudes amplitudes)
+        {
+            if (ShouldFollowBeats)
```
peppy, line 44: i think you may be able to resolve the `DrawableHitObject` and check its hit state rather than require this manual bool flow. Should feel a bit more logical.

LittleEndu: Oh, didn't know the hit object is injected. Will try.

peppy: yep, a bit hidden:

https://github.com/ppy/osu/blob/9477629a5686e33881bf0c0d28ad4f0c93bda4a4/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L24

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ReverseArrowPiece.cs
```diff
@@ -8,12 +8,15 @@
...
+using osu.Framework.Allocation;
+using osu.Game.Rulesets.Objects.Drawables;
 
 namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 {
     public class ReverseArrowPiece : BeatSyncedContainer
     {
-        public bool ShouldFollowBeats = true;
+        [Resolved]
+        private DrawableHitObject drawableSlider { get; set; }
```
LittleEndu, line 19: It isn't slider that's resolved. it's the repeat...

peppy: i would hope that's what you want?

LittleEndu: nah, it's the repeat that's hit. the first commit had the repeat set the bool.

peppy: yeah. so just the naming was wrong. that's fine.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderRepeat.cs
```diff
@@ -79,6 +80,7 @@ protected override void UpdateStateTransforms(ArmedState state)
                 case ArmedState.Hit:
                     this.FadeOut(animDuration, Easing.Out)
                         .ScaleTo(Scale * 1.5f, animDuration, Easing.Out);
+                    scaleContainer.Transforms.ForEach(t => scaleContainer.RemoveTransform(t));
```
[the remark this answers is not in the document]
LittleEndu: scaleContainer has a beat synced drawable, the reverse arrow, in it and gets scale transformations applied to it every beat.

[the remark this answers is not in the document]
smoogipoo: It's applied to a child of it as far as I can see?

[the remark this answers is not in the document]
smoogipoo: This would also probably crash if it did anything, with a collection enumeration exception. You have to use `ClearTransforms()` in cases like this.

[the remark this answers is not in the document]
LittleEndu: Oh, it is. Then removing transformations was never necessary. Makes sense, considering that the scaling up happens at the start of the beat and this PR disables that when the arrow is hit.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ReverseArrowPiece.cs
```diff
@@ -37,7 +42,10 @@ public ReverseArrowPiece()
             };
         }
 
-        protected override void OnNewBeat(int beatIndex, TimingControlPoint timingPoint, EffectControlPoint effectPoint, TrackAmplitudes amplitudes) =>
-            Child.ScaleTo(1.3f).ScaleTo(1f, timingPoint.BeatLength, Easing.Out);
+        protected override void OnNewBeat(int beatIndex, TimingControlPoint timingPoint, EffectControlPoint effectPoint, TrackAmplitudes amplitudes)
+        {
+            if (!drawableRepeat.IsHit)
```
smoogipoo, line 47: Should it still bounce if it's missed?/Does that even matter?

 Maybe this should be `Judged` instead if not?

peppy: unless it's doing a different animation on miss, it should definitely continue bouncing

smoogipoo (approved):
lgtm

PR #8630, opened 2020-04-05
Add a simple constructor for BreakPeriod

frenzibyte:
For simple construction of a list of break periods (e.g. filling a method with an array of break periods inside a test case)


bdach:
This PR on its own doesn't make sense, adding a constructor like this doesn't achieve anything. Just include this change in the PR this depends on.

iiSaLMaN:
It may look a bit weird as it is just something that wasn't added and is not related to the dependent PR + there are 2 files to be changed to use the constructor instead, so I thought it'll better to have an own PR.

Also this PR achieves something technically, it's for code quality purposes.

smoogipoo (approved):

PR #8632, opened 2020-04-05
Disallow HP draining in non-play sections

### LINKED ISSUE #7607: HP drains in non-play sections

frenzibyte:
- [x] Depends on #8630 
- [x] Depends on #8701

The solution provided by this PR goes by calculating all the *non-drain* sections (end time of the last hit object before the break to the start time of the first hit object after that break) and advancing through each in similar fashion to how `BreakTracker` works.

## Notes
 - `IsBreakTime.Value` now is not used in checking for non-play sections as the break periods provided by the beatmap are [indirectly accounted](https://github.com/iiSaLMaN/osu/blob/c902ba40860b2757ea960925322bf9731b4eeefc/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs#L112-L121) to *the list of non-drain sections*.

Closes #7607 
Alternative to #8542


osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -47,6 +49,34 @@ public class DrainingHealthProcessor : HealthProcessor
...
+                else
+                {
+                    while (time < nonDrainSections[currentNonDrainSection].startTime && currentNonDrainSection > 0)
+                        currentNonDrainSection--;
+                }
+
+                var closestSection = nonDrainSections[currentNonDrainSection];
+                return time >= closestSection.startTime && time <= closestSection.endTime;
+            }
+        }
```
bdach, line 78: * Having this much logic in a getter is a large stretch. Should be a method instead IMO.
* I'm thinking this might be a decent use case for a linked list. Then you can get rid of the index and just keep a reference to the current node.
* This is pretty much a copy of the core part of the `BreakTracker` implementation. I'd say this particular logic (determining if current time is within an interval, optimised for sequential access) needs to be shared. (Strongly encourage composition over inheritance if this happens.)

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -60,23 +90,36 @@ protected override void Update()
...
             if (beatmap.HitObjects.Count > 0)
                 gameplayEndTime = beatmap.HitObjects[^1].GetEndTime();
 
+            // Ranges between the end of last hit object before a break
+            // and the start of first hit object after a break should
+            // not allow HP draining. (with break periods in)
+            foreach (BreakPeriod b in beatmap.Breaks)
+            {
+                var startTime = beatmap.HitObjects.LastOrDefault(h => h.GetEndTime() < b.StartTime)?.GetEndTime() ?? double.MinValue;
+                var endTime = beatmap.HitObjects.FirstOrDefault(h => h.StartTime > b.EndTime)?.StartTime ?? double.MaxValue;
```
bdach, line 118: To avoid computing/accessing twice:
```suggestion
                var startTime = beatmap.HitObjects.Select(h => h.GetEndTime()).LastOrDefault(endTime => endTime < b.StartTime) ?? double.MinValue;
                var endTime = beatmap.HitObjects.Select(h => h.StartTime).FirstOrDefault(startTime => startTime > b.EndTime) ?? double.MaxValue;
```

bdach (commented):
Notes from an initial pass. Haven't ran.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
iiSaLMaN: Does making a `RangeList` implementation and a method inside to do this check sound good?

bdach: Not sure about the name (maybe `IntervalList`?) but yeah, sounds like a reasonable way forward.

iiSaLMaN: though keep in mind it's not gonna mention times in it since this can be applied to simply any value ranges.

bdach: *interval* is not a strictly time-specific term.

iiSaLMaN: @bdach `LinkedList` turns out to be limiting a bit of good behaviours on edge cases, for example: if the current node has been removed from the list, `Next` and `Previous` will also become `null` which means I would need to set the current node back to the first element in the list, while using indices in the other hand, the index will already point to the next element to the removed one, and I can clamp the index between `0` and the length of the list to ensure that it isn't already out-of-range.

bdach: I'm not sure why removal of nodes is relevant here as I envisioned this being a read-only structure, but fine I suppose.

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
iiSaLMaN: well, `StartTime` and `EndTime` are structures so default doesn't refer to `null` (`0.0d`), that's why I got forced into calculating twice...

iiSaLMaN: though tbh we may want to do something like a `FirstOrNull` & `LastOrNull` linq extensions (and even write it framework-side)

bdach: Ehhh I forgot struct types do that. Could throw in a `Cast<double?>` but it's ugly too.

Helpers might not be necessary as this might get rectified when nullable reference types come in? Not sure.

iiSaLMaN: *actually it may get worse as now all normal declarations now will be considered non-nullables?*

bdach: Actually it seems this can be written as:
```suggestion
                var startTime = beatmap.HitObjects.Select(h => h.GetEndTime()).Where(endTime => endTime < b.StartTime).DefaultIfEmpty(double.MinValue).Last();
                var endTime = beatmap.HitObjects.Select(h => h.StartTime).Where(startTime => startTime > b.EndTime).DefaultIfEmpty(double.MaxValue).First();
```
but that's a bit verbose too...

Luxiono:
#8701 is now merged as of around 1:00 today.

iiSaLMaN:
> #8701 is now merged as of around 1:00 today.

Yep, I'll get to it as soon as possible.

bdach:
I've gone ahead and integrated changes from the dependent PRs myself. It passes the tests added previously (at least locally, at the time of writing) and seems pretty sane to me.

The period calculation is maybe a little inefficient but I profiled it and I don't think it's a huge problem right now...

image

(entire method as well as the lambda delegate highlighted)

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -77,6 +81,19 @@ public override void ApplyBeatmap(IBeatmap beatmap)
...
 
+            noDrainPeriodTracker = new PeriodTracker(beatmap.Breaks.Select(breakPeriod => new Period(
+                beatmap.HitObjects
+                       .Select(hitObject => hitObject.GetEndTime())
+                       .Where(endTime => endTime < breakPeriod.StartTime)
+                       .DefaultIfEmpty(double.MinValue)
+                       .Last(),
+                beatmap.HitObjects
+                       .Select(hitObject => hitObject.StartTime)
+                       .Where(startTime => startTime > breakPeriod.EndTime)
```
smoogipoo, line 92: What happens if the hitobject's start time is equal to the break's end time? E.g.

```
[H] - - [B - - - [HB] - [H]
  ^^^^^^^^^^^^^^^^^^^^^^^
  ^^^^^^^^^^^^^^^^   
```
Won't the drain region be the first one marked above, rather than the second one? That is to say, these checks should probably be `>=` and `<=` above.

This also maybe shows a little bit of an interesting scenario scenario:

```
[H] - - [B - [H] - B] - [H]
  ^^^^^^^^^^^^ ^^^^^^^^^^
  ^^^^^^^^^^^^^^^^^^^^^^^
```
Where if a hitobject intersects a break, I'd expect the no-drain region to be the first one marked above but it seems like this code will do the second one? It's a bit of an edge-case scenario and I only know of aspire maps that abuse this, so maybe not worth consideration.

smoogipoo (changes requested):

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
bdach: Good points. I'll add these as test cases and see if changing the comparison suffices.

bdach:
I pushed in a fix for the first scenario. I also have *a* fix for the other one (drain inside breaks), but... it's not pretty - [take a look if you wish](https://github.com/bdach/osu/compare/848a3fb6d74ec3b443029cc048934c1df4ea728f...bdach:drain-inside-breaks). It's a sort of a poor man's state machine in an awkward loop. It passes the test I wrote for it though, for whatever that's worth, but I'm not sure if that code's legible to anyone but me.

Problem is that the logic of calculation gets fairly complex when you allow drain to happen inside breaks, and even then I think the above version could end up being slightly wrong if you start allowing overlapping objects.

kornkaobat:
I have a question, why shouldn’t we fix the break period itself and not the draining mechanism which is part of the break. Like why do Beatmap breaks have gaps between them and the nearest hitobject in the first place?

bdach:
Because the legacy beatmap format explicitly encodes breaks as a type of event with its own start and end time. 

The fact that the stable editor automatically inserts breaks doesn't change the fact that there are maps that purposefully change them via manual `.osu` editing (see aspire maps mentioned above), so it's a means of artistic license for mappers that shouldn't be retroactively removed without very good reason.

peppy:
Continuing on the above discussion, we may decide to drop the custom break logic going forward (I could get behind this).

osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
peppy: I think this second case is fine to disregard for now.

peppy:
This code reads well enough and plays as expected 👍 .

peppy (approved):

Issue #8546, opened 2020-04-01
Wide playfield osu!mania skins are not working.


Joppe27:
Mania skins with a wide playfield will not work. Those skins will play as if the default skin was selected. This breaks stepmania skins and others, a few examples that don't work: https://www42.zippyshare.com/v/Ey3wA9pO/file.html
Logs: https://www42.zippyshare.com/v/Q8QYaqfE/file.html
Edit: I originally thought this issue occured because of high ColumnWidth values, this is not the case(?)

PR #8640, opened 2020-04-06
Fix legacy skins providing null version instead of 1.0

smoogipoo:
This isn't actually used anywhere except mania right now, but going forward I think we'll want to refer to skin versions via comparisons, rather than having an extra check for the `null` case.

Fixes one of the skins in https://github.com/ppy/osu/issues/8546


smoogipoo:
Turns out this is required, going to close because this PR is going completely in the wrong direction.

PR #8641, opened 2020-04-06
Fix performance when parsing mania skins

smoogipoo:
O(n^2) complexity :(


peppy:
This mania parsing code is a bit hard to follow, not surprised..

peppy (approved):

PR #8642, opened 2020-04-06
Implement mania note + key image configs

### LINKED ISSUE #8546: Wide playfield osu!mania skins are not working.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8546 (along with https://github.com/ppy/osu/pull/8640)


bdach:
Seems mostly fine to me, although I'm not a *huge* fan of those asserts - would have been better it was possible to design decoding in a way that wouldn't require them, but I feel like legacy skinning might be a rabbit hole the depths of I have very little idea about.

smoogipoo:
I do agree that the nullability is ugly. I plan to eventually split the lookups into stage and column enums, but I've merged them together for simplicity for now.

peppy (dismissed):

peppy (approved):
seems ok for now.

PR #8643, opened 2020-04-06
Make version-less skins fallback to version 1.0

smoogipoo:
Fixes some skins in https://github.com/ppy/osu/issues/8546


bdach (commented):
In retrospect I regret ever suggesting `decimal` as a type for versions (versions are *really* not decimals and the only reason it has a chance to work is because legacy skins never had a minor revision past 9) but hindsight is 20/20 I suppose.

smoogipoo:
Yeah agree. Probably should use `Version`, but that can be done in the future.

peppy (approved):

PR #8648, opened 2020-04-06
Fix storyboard videos being offset incorrectly

peppy:
Setting the PlaybackPosition will ignore the underlying clock with recent changes. `startAtCurrentTime = false` alone (plus a `+= offset` in `DrawableStoryboardVideo.LoadComplete`) isn't enough as we would need to delay the `PlaybackPosition` call to after the video runs its `LoadComplete`. Instead, let's just set the complete correct offset.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3432


bdach (approved):
Tested as working and this way actually seems most correct and least prone to error in the future. 👍

smoogipoo (approved):

smoogipoo:
I do agree that the nullability is ugly. I play to eventually split the lookups into stage and column enums, but I've merged them together for simplicity for now.

smoogipoo:
Gonna merge this as there'll probably be another framework hotfix today anyway.

peppy:
Yeah, this isn't actually dependent on the framework-side change with my final approach.

PR #8650, opened 2020-04-07
Fix SkinnableTestScene losing test resources on dynamic recompilation

peppy:


smoogipoo (approved):
I'm approving this very hesitantly, and feel like we should maybe be providing the original assembly to a dynamically compiled one instead.

PR #8653, opened 2020-04-07
Fix inconsistent scroll speeds in mania

### LINKED ISSUE #8651: Inonsistent mania scrolling speed
Describe the bug: i think peppy said that the scrolling in lazer is fixed, so maybe i should report this bug. idk what is the correct scroll speed between the two. as you can see in the vid, the scroll is set to 14 (435ms). both maps are mania specific map.

first map : https://osu.ppy.sh/beatmapsets/913062#mania/1913473 which is slower
second map: https://osu.ppy.sh/beatmapsets/63089#mania/193127 which is faster

Screenshots or videos showing encountered issue: just remove zip extension
[My Movie.mp4.zip](https://github.com/ppy/osu/files/4442337/My.Movie.mp4.zip)

osu!lazer version: master 5694898 (idk if thats the right one cause there were commits while im running lazer)

**Logs:**
maybe not needed?

im still searching for maps with the same problem

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8651

### DIFF
--- a/osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
+++ b/osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
@@ -64,6 +64,7 @@ namespace osu.Game.Rulesets.Mania.UI
             {
                 // Mania doesn't care about global velocity
                 p.Velocity = 1;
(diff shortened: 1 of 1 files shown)


smoogipoo:
Sorry about the force-push, accidentally added some skinning stuff.

PR #8656, opened 2020-04-07
Add per-ruleset skinnable test scenes to correct dynamic compilation cases

peppy:
Tested each and every test scene to work as expected. `TestSceneSliders` doesn't work but that needs some more intensive fixes.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3434


peppy:
Framework fix is required for the `Ruleset` fix in corresponding ruleset skinnabletests (else the `RulesetSkinnableTestScene` needs to be included as a required type in every test scene implementing it).

smoogipoo (approved):
Fine for now. The base testscene could probably take in a ruleset, but /shrug.

Mania's still going to need some work to be compilable (needs to include `ManiaSettingsSubsection`) but that can be done in a further PR.

PR #8669, opened 2020-04-07
Fix legacy skin texture fallback logic

### LINKED ISSUE #8665: Skin reverse arrow and follow points not importing
**Describe the bug:** 
Reverse arrow and follow point skill elements are not being imported from skins and are falling back to lazer default versions.
Regression from 2020.404.0

Skin used:
[https://archive.btmc.live/noel_v6/Komori%20-%20MC%20NOEL%20(VI).osk](url)

Screenshot from 2020.408.0
image

Screenshot from 2020.404.0
image

**osu!lazer version:** 
2020.408.0

bdach:
Resolves #8665.

# Summary

The recently-modified (in eff17c2) skin texture fallback logic was very subtly incorrect. If at the end of the first loop no texture was found, it would be checked for null to avoid setting scale adjust on a null texture, but then returned anyway, bypassing the fallback logic for subsequent possible paths entirely.

Invert the check and explicitly continue to the next fallback path if neither a 2x, nor 1x texture with the given name is found in the store.

# Remarks

Missed the bug during review... The fallback logic now has somewhat comprehensive test cases so hopefully should be rigorously checked with any subsequent changes.

### DIFF
--- a/osu.Game/Skinning/LegacySkin.cs
+++ b/osu.Game/Skinning/LegacySkin.cs
@@ -293,9 +293,10 @@ namespace osu.Game.Skinning
                     texture = Textures?.Get(name);
                 }
 
-                if (texture != null)
-                    texture.ScaleAdjust = ratio;
+                if (texture == null)
+                    continue;
 
+                texture.ScaleAdjust = ratio;
                 return texture;
             }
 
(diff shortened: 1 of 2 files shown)


osu.Game/Skinning/LegacySkin.cs
```diff
@@ -293,9 +293,10 @@ public override Texture GetTexture(string componentName)
                     texture = Textures?.Get(name);
                 }
 
-                if (texture != null)
-                    texture.ScaleAdjust = ratio;
+                if (texture == null)
+                    continue;
```
iiSaLMaN, line 297: Can be merged with the above texture null-check I believe.

smoogipoo (approved):

peppy (approved):
I was a bit confused by the logic in the previous PR, should've taken more time to check it.

PR #8672, opened 2020-04-08
Adjust TimingControlPoint equivalency

### LINKED ISSUE #8286: Editor tick colours are swapped on certain beatmaps
When compared to stable, ticks on the editor's timeline have different colours - in this particular case it seems like red ticks are swapped with white ones.

image

This happens because lazer doesn't respect the ticks order being reset after passing a new timing point. Here's how stable does it (notice how the tick where the (2) slider is placed changes its colour) :

Lazer doesn't do that, causing the disprepancy here:

Example beatmap to test: https://osu.ppy.sh/beatmapsets/1001507

Alchyr:
Closes #8286 

The equivalency test for TimingControlPoints, used to remove unnecessary points, results in the omission of points that shouldn't be omitted. Timing points cannot be considered to be equivalent if they have different times, since offset is an important factor in timing.

A similar issue could occur with EffectControlPoints, which omit barlines, in the case of multiple barlines in a row being omitted.
I didn't change this as returning false if OmitFirstBarline is true would incorrectly cause a false result in the Equals method of ControlPoint.


peppy (dismissed):
unfortunate we don't have test coverage, but i think it's fine for the time being..

smoogipoo:
What, no. Please add a test for this, `ControlPointInfoTest` is the class you want to use and there already is a method `TestAddRedundantTiming` that should already be failing with this PR.

smoogipoo (changes requested):
as requested

Alchyr:
Thank you, after checking more carefully it was indeed functioning incorrectly. When the new timing point is checked, it is not yet assigned to a group and so Time returns 0. This results in incorrect comparison.

To solve this, the time was also passed along with the comparison point in checkAlreadyExisting, but this resulted in EquivalentTo not being used as it is defined. As such, I chose to move the check into a separate method, IsRedundant. This is then also used for changing the test for EffectControlPoints without affecting the Equals method of ControlPoint.

osu.Game/Beatmaps/ControlPoints/ControlPoint.cs
```diff
@@ -25,6 +25,14 @@ public abstract class ControlPoint : IComparable<ControlPoint>, IEquatable<Contr
         /// <returns>Whether equivalent.</returns>
         public abstract bool EquivalentTo(ControlPoint other);
 
+        /// <summary>
+        /// Whether this control point results in a meaningful change when placed after another.
+        /// </summary>
+        /// <param name="other">Another control point to compare with.</param>
+        /// <param name="time">The time this timing point will be placed at.</param>
```
peppy, line 32: do you mean "other" when you say "this"? If so, reference the param.

Also timing point -> control point.

peppy (changes requested):
Check CI – there's a few style fixes waiting for you as well.

osu.Game/Beatmaps/ControlPoints/ControlPoint.cs
Alchyr: It's intended to be called from the new timing point, with other being the already existing timing point. I think I'll rename the parameter so that this is more clear. (existing instead of just other)
This is significant since only existing timing points will have a proper Time value to compare to.

Will change timing point to control point.

peppy (dismissed):
Seems correct.

peppy:
@smoogipoo requesting your re-check on this one.

smoogipoo:
Can we remove `EquivalentTo()`? I don't like have 3 comparison methods... `EqualTo()`, `EquivalentTo()`, `IsRedundant()`, with `IsRedundant()` just forwarding to `EquivalentTo()` in the one case it's implemented.

peppy:
i was also hoping to remove one of them, but wasn't sure which can go as they all have usages.

smoogipoo:
As far as I could see, `EquivalentTo` is only ever used from `IsRedundant`.

smoogipoo (approved):

smoogipoo:
@peppy give this one another check. I've removed both `EquivalentTo()` and `Equals()`, because they were both awkward:
1. `EquivalentTo()` was merged into `IsRedundant()`.
2. If `Equals()` used `IsRedundant()` it would actually fail for `TimingControlPoint`s, which are never redundant. I don't envision a case where we want to compare control points by their property-based equivalency.

peppy (approved):

PR #8676, opened 2020-04-08
Implement mania stage bottom/left/right images

### LINKED ISSUE #8558: osu!mania stage sprites are not rendered
not sure if i should post this but i did

Describe the bug: mania-stage-left and mania-stage-right (maybe even the bottom) is not applied. idk if it will be implemented

**Screenshots or videos showing encountered issue: 
image

osu!lazer version:2020.402.0

**Logs:**
idk if needed

smoogipoo:
Resolves https://github.com/ppy/osu/issues/8558

Added a metrics display for the bottom image because the classic skin doesn't have one. There aren't really any metrics in mania so...


peppy:
Do we not want metrics for left/right stage too?

smoogipoo:
Reason I didn't add is because the classic skin already has displays.

As I said though, mania doesn't really have/care about any metrics, it just places things wherever and sizes things to fit whatever they're placed in.

If you want I guess I can spend time doing a metrics version for all mania skin components, but I haven't done so I've excluded it.

peppy:
guess it's fine for now.

PR #8677, opened 2020-04-08
Fix hidden notes due to 0 minimum width

### LINKED ISSUE #8667: Some wide playfield osu!mania skin elements are invisible
Some of the elements of wide playfield skins are completely invisible. This is different for all skins and makes mania completely unplayable for some skins, whereas on other skins it's not that big of a deal. 
Skins: 1 https://www44.zippyshare.com/v/LH5OkEmn/file.html (all of the notes are invisible)
2 https://www44.zippyshare.com/v/xvt7MFHC/file.html (only the 4th hold notes don't appear correctly)
Video: https://youtu.be/rg2vMRj9Zwo
Version: 2020.408.0 (latest)
Logs: none

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8667

### DIFF
--- a/osu.Game/Skinning/LegacyManiaSkinConfiguration.cs
+++ b/osu.Game/Skinning/LegacyManiaSkinConfiguration.cs
@@ -54,7 +54,7 @@ namespace osu.Game.Skinning
         public float MinimumColumnWidth
         {
             get => minimumColumnWidth ?? ColumnWidth.Min();
-            set => minimumColumnWidth = value > 0 ? (float?)value : null;
+            set => minimumColumnWidth = value;
         }
     }
 }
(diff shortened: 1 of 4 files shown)


osu.Game/Skinning/LegacyManiaSkinConfiguration.cs
```diff
@@ -54,7 +54,7 @@ public LegacyManiaSkinConfiguration(int keys)
         public float MinimumColumnWidth
         {
             get => minimumColumnWidth ?? ColumnWidth.Min();
-            set => minimumColumnWidth = value;
```
peppy, line 57: sure we don't want to do this in the actual parsing logic (`LegacyManiaSkinDecoder`)? seems weird to be doing it here.

smoogipoo: Yeah I don't mind

PR #8679, opened 2020-04-08
Fix TestSceneColumn columns not getting a width

peppy:
Also fixes dynamic compilation not working.


smoogipoo:
Moved it to the `Column` ctor, because the editor placement test scenes were also broken.

smoogipoo (approved):

PR #8680, opened 2020-04-08
Fix slider ball and follow circle blending for legacy skins

peppy:
This regressed with the blending *fix* framework side (has been wrongly applied all along but only started to actually show due to the fix).


smoogipoo:
Want to fix repeat points too, and look through all other usages?

peppy:
Good catch, also fixed. Checked through other cases and seems good.

smoogipoo (approved):

PR #8681, opened 2020-04-08
Fix osu!taiko input drum alignment for old skin versions

### LINKED ISSUE #8609: taiko drum inner and outer is not placed properly
Describe the bug: taiko inner and outer doesnt match the placement of taiko left.png skin

osu_2020-04-04_07-40-13
osu_2020-04-04_07-40-13-1
osu_2020-04-04_07-40-16
osu_2020-04-04_07-40-23

Skin:
[taiko.zip](https://github.com/ppy/osu/files/4430402/taiko.zip)

osu!lazer version: 2020.404.0

Logs:
[runtime.log](https://github.com/ppy/osu/files/4430399/runtime.log)
[performance.log](https://github.com/ppy/osu/files/4430400/performance.log)

peppy:
- Closes https://github.com/ppy/osu/issues/8609.
- Bumps default skin version in line with what it probably should be.

image


osu.Game/Skinning/DefaultLegacySkin.cs
```diff
@@ -20,7 +20,7 @@ public DefaultLegacySkin(IResourceStore<byte[]> storage, AudioManager audioManag
                 new Color4(242, 24, 57, 255)
             );
 
-            Configuration.LegacyVersion = 2.0m;
+            Configuration.LegacyVersion = 2.7m;
```
iiSaLMaN, line 23: An inline comment can be added here pointing out why `LATEST` const isn't used instead.

peppy: i think it's pretty obvious. DefaultLegacySkin is basically constants from stable. if it requires a comment then we might as well just use LATEST and rename it to LAST_VERSION.

smoogipoo (approved):

PR #8683, opened 2020-04-08
Add support for LayeredHitSounds skin option

### LINKED ISSUE #8595: skin hitsounds overlap beatmap hitsounds
**Describe the bug: skin hitsounds overlap beatmap the video begins with me trying out this beatmap ( https://osu.ppy.sh/beatmapsets/63089#mania/193128 ) in stable, as you can see in the video, the skin hitsounds wasnt present, only the beatmap ones. then i moved to lazer and try the same map, at the beginning i pressed the keys then the problem appear. i tried another beatmap near the end of the vid and it doesnt have the problem

Video: (sorry for laggy video and loud audio cause my laptop is such a potato) just remove the zip extension
[2020-04-03 16-27-29.mkv.zip](https://github.com/ppy/osu/files/4425879/2020-04-03.16-27-29.mkv.zip)

stable settings:
screenshot052

lazer settings:
osu_2020-04-03_16-32-46

osu!lazer version: commit 4e061ce

Logs: 
[performance-gamethread.update.log](https://github.com/ppy/osu/files/4425919/performance-gamethread.update.log)
[runtime.log](https://github.com/ppy/osu/files/4425920/runtime.log)
[database.log](https://github.com/ppy/osu/files/4425922/database.log)
[performance.log](https://github.com/ppy/osu/files/4425923/performance.log)
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/4425924/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/4425925/performance-gamethread.draw.log)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/4425917/performance-gamethread.input.log)

Fire937:
Closes  #8595

The current behaviour of osu!lazer is to add the hitnormal sample on top of the other samples for every single skin/beatmap/ruleset/hitobject, even in osu!mania, that is what causes the issue. Moreover osu!stable uses a LayeredHitSounds skin property that tells osu! whether it should add the hitnormal sample artificially or not based on the skin used (defaults to yes). LayeredHitSounds is also ignored when in osu!mania mode: osu!stable never adds the hitnormal sample artificially in osu!mania mode.

Currently the `LayeredHitSounds` property is included in `LegacySkinConfiguration,` however i am not sure about whether it belongs here or not. Maybe it should be moved to `SkinInfo`.
A ruleset can override the value of `AllowsLayeredHitSounds` in its `ConvertHitObjectParser` to disable the `LayeredHitSounds` property.

- [x] Depends on #8649


frenzibyte (changes requested):

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -304,17 +304,28 @@ public override Texture GetTexture(string componentName)
 
         public override SampleChannel GetSample(ISampleInfo sampleInfo)
         {
-            foreach (var lookup in sampleInfo.LookupNames)
+            var hsi = sampleInfo as HitSampleInfo;
+
+            if (hsi != null && hsi.IsLayered && !Configuration.LayeredHitSounds)
```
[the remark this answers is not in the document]
Fire937: I agree. I changed it.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -124,6 +124,9 @@ protected override void Dispose(bool isDisposing)
                     {
                         case LegacySkinConfiguration.LegacySetting.Version:
                             return SkinUtils.As<TValue>(new Bindable<decimal>(Configuration.LegacyVersion ?? LegacySkinConfiguration.LATEST_VERSION));
+
+                        case LegacySkinConfiguration.LegacySetting.LayeredHitSounds:
+                            return SkinUtils.As<TValue>(new Bindable<bool>(Configuration.LayeredHitSounds ?? true));
```
iiSaLMaN, line 129: This is actually redundant, to be precise on what you should do, `LegacySkinDecoder` simply takes all lines that're not handled specially as keys and values added to `LegacySkinConfiguration.ConfigDictionary`, so you would only need the enum member `LayeredHitSounds` to retrieve the value from the dictionary with, i.e. remove the property from `LegacySkinConfiguration` and remove the case added in `LegacySkinDecoder`.

Fire937: I see, thank you for your explanation I definitely misunderstood how the `SkinConfiguration` class works. I removed the property from `LegacySkinConfiguration` as you suggested and now access the property by string:
`GetConfig<string, bool>("LayeredHitSounds")`
Is that the proper way to do it?

iiSaLMaN: Almost there! the lookup should've been `LegacySetting.LayeredHitSounds` as the name of the enum member `LayeredHitSounds` will then be converted to a string to use as the lookup key in the dictionary, hence why the lookup type is almost always an enum, that's because it makes life easier :P, we definitely don't want to search what was the name of a config value lookup everytime we want to retrieve one.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -304,17 +304,28 @@ public override Texture GetTexture(string componentName)
...
         public override SampleChannel GetSample(ISampleInfo sampleInfo)
         {
-            foreach (var lookup in sampleInfo.LookupNames)
+            var hsi = sampleInfo as HitSampleInfo;
+
+            if (hsi != null && hsi.IsLayered && !Configuration.LayeredHitSounds)
+                // The layered hitsample is ignored if the skin disabled LayeredHitSounds
+                return null;
+
+            if (hsi == null || hsi.IsCustom)
```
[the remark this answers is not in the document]
bdach: this should go in the PR this one depends on

osu.Game/Skinning/LegacySkin.cs
Fire937: That's what I thought at first but then in `GetConfig` there is no default case for `LegacySkinConfiguration.LegacySetting` meaning it only tests for `LegacySkinConfiguration.LegacySetting.Version` and breaks, thus returning null when trying to retrieve the `LayeredHitSounds` property.

Maybe the case in `GetConfig` should be kept or a default case that translates the enum to a string could be added. However that would be exactly what the default case does when it is none of the enums that have specific cases thus inducing code redundancy.

iiSaLMaN: sounds like the version case should be:
```csharp
case LegacySkinConfiguration.LegacySetting legacySetting when legacySetting == LegacySkinConfiguration.LegacySetting.Version:
    // retrieve version.
```

osu.Game/Audio/HitSampleInfo.cs
```diff
@@ -37,6 +37,16 @@ public class HitSampleInfo : ISampleInfo
         /// </summary>
         public int Volume { get; set; }
 
+        /// <summary>
+        /// Whether the sample was artificially added or if it was a beatmap specification.
```
bdach, line 41: This xmldoc feels like a mental shortcut (what connection does "layering" have with "artificially adding a sample"? I wouldn't understand what it means without the PR description) I'd say it needs to mention explicitly that `IsLayered` is a normal hitsample layered over the beatmap one.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -304,17 +304,28 @@ public override Texture GetTexture(string componentName)
 
         public override SampleChannel GetSample(ISampleInfo sampleInfo)
         {
-            foreach (var lookup in sampleInfo.LookupNames)
+            var hsi = sampleInfo as HitSampleInfo;
+
+            if (hsi?.IsLayered == true && GetConfig<string, bool>("LayeredHitSounds")?.Value == false)
+                // The layered hitsample is ignored if the skin disabled LayeredHitSounds
```
bdach, line 304: Kind of a redundant comment.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -32,6 +32,8 @@ public abstract class ConvertHitObjectParser : HitObjectParser
 
         protected bool FirstObject { get; private set; } = true;
 
+        protected bool AllowsLayeredHitSounds { get; set; } = true;
```
bdach, line 35: Usually we use infinitives for these sorts of flags (so would be `AllowLayeredHitSounds`)

osu.Game/Rulesets/Objects/Legacy/Mania/ConvertHitObjectParser.cs
```diff
@@ -15,6 +15,8 @@ public class ConvertHitObjectParser : Legacy.ConvertHitObjectParser
         public ConvertHitObjectParser(double offset, int formatVersion)
             : base(offset, formatVersion)
         {
+            // LayeredHitSounds is disabled in mania
```
bdach, line 18: Also sort of obvious.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -349,16 +351,19 @@ private List<HitSampleInfo> convertSoundType(LegacyHitSoundType type, SampleBank
...
+                soundTypes.Add(new LegacyHitSampleInfo
                 {
                     Bank = bankInfo.Normal,
                     Name = HitSampleInfo.HIT_NORMAL,
                     Volume = bankInfo.Volume,
-                    CustomSampleBank = bankInfo.CustomSampleBank
-                }
-            };
+                    CustomSampleBank = bankInfo.CustomSampleBank,
+                    IsLayered = !type.HasFlag(LegacyHitSoundType.Normal)
```
bdach, line 364: I feel like this is backwards. Shouldn't this just use `AllowsLayeredHitSounds`? If both the flag is set and `AllowsLayeredHitSounds` is `true`, won't this wrongly consider the sample non-layered?

bdach (changes requested):
I really dislike how complicated `LegacySkin.GetSample()` is starting to be. I'd say it needs unit tests at the least, and a refactoring on top to decrease complexity if possible.

osu.Game/Audio/HitSampleInfo.cs
Fire937: That's true, what about: Whether the sample is layered over the beatmap ones or not.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
Fire937: My bad, i am not extremely familiar yet with the conventions used in this repo. I will change it.

osu.Game/Audio/HitSampleInfo.cs
bdach: I'd go for the very explicit
```suggestion
        /// True if the sample is a <c>hitnormal</c> sample added artificially due to the skin having
        /// the <see cref="osu.Game.Skinning.LegacySkinConfiguration.LegacySetting.LayeredHitSounds"/> option set.
```

But then again adding *that* kind of shows that if possible, `IsLayered` shouldn't really be a flag at all as it is out of place in the general `HitSampleInfo` (it's something that has explicitly to do with legacy skins).

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
Fire937: The thing is that the beatmap could actually explicitly indicate that the sample to be played (among others) should be hitnormal. In this case, whether `AllowLayeredHitSounds` is `true` or `false` should not affect the fact that the sample is not layered. That means it should always be played even in osu!mania. That's why this problem doesn't appear in other beatmap as they use hitnormal as opposed to this map that uses hitfinish and the others but not hitnormal.
I'm actually concerned about the naming of this flag, maybe `IsForced` would make more sense?

osu.Game/Audio/HitSampleInfo.cs
Fire937: In that regard I was thinking about making the `LegacyHitSampleInfo` class public and including the `IsLayered` property in it. Would that work or is it a bad idea? 
Incidentally `IsCustom` could also be included in it.

osu.Game/Beatmaps/Formats/LegacyDecoder.cs
```diff
@@ -171,6 +171,9 @@ public override HitSampleInfo ApplyTo(HitSampleInfo hitSampleInfo)
                 if (string.IsNullOrEmpty(baseInfo.Suffix) && CustomSampleBank > 1)
                     baseInfo.Suffix = CustomSampleBank.ToString();
 
+                // 0 means osu!'s default hitsounds
+                baseInfo.IsCustom = CustomSampleBank != 0;
+
```
smoogipoo, line 176: Same here, can be removed

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -32,6 +32,8 @@ public abstract class ConvertHitObjectParser : HitObjectParser
 
         protected bool FirstObject { get; private set; } = true;
 
+        protected bool AllowLayeredHitSounds { get; set; } = true;
```
smoogipoo, line 35: This cannot exist in the parser - it should only exist within the legacy skin itself.

The mania override can be done inside `LegacyManiaSkinTransformer`.

That is to say, layered hitsounds should _always_ be added.

osu.Game/Audio/HitSampleInfo.cs
```diff
@@ -37,6 +37,17 @@ public class HitSampleInfo : ISampleInfo
         /// </summary>
         public int Volume { get; set; }
 
+        /// <summary>
+        /// True if the sample is a <c>hitnormal</c> sample added artificially due to the skin having
+        /// the <see cref="osu.Game.Skinning.LegacySkinConfiguration.LegacySetting.LayeredHitSounds"/> option set.
+        /// </summary>
+        public bool IsLayered { get; set; }
```
smoogipoo, line 44: This must only exist within `ConvertHitObjectParser.LegacyHitSampleInfo`.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
smoogipoo: With application of the above comments, the outer condition can once again be removed and this can remain here.

osu.Game/Rulesets/Objects/Legacy/Mania/ConvertHitObjectParser.cs
```diff
@@ -15,6 +15,7 @@ public class ConvertHitObjectParser : Legacy.ConvertHitObjectParser
         public ConvertHitObjectParser(double offset, int formatVersion)
             : base(offset, formatVersion)
         {
+            AllowLayeredHitSounds = false;
```
smoogipoo, line 18: See other comments.

osu.Game.Tests/Skins/TestSceneLegacySkin.cs
```diff
@@ -0,0 +1,105 @@
...
+using osu.Game.Tests.Visual;
+using osu.Framework.IO.Stores;
+using osu.Framework.Testing;
+using osu.Framework.Audio;
+using osu.Framework.Audio.Sample;
+
+namespace osu.Game.Tests.Skins
+{
+    [HeadlessTest]
+    public class TestSceneLegacySkin : OsuTestScene
```
smoogipoo, line 20: This testing is inadequate because it doesn't test the result of gameplay loading. You want to add a test to `TestSceneHitObjectSamples`.

I'm fairly certain the implementation in this PR (inside `LegacySkin`) will fail, since `SkinnableSound` will fall back to the default lazer sample store if `null` is returned.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -311,17 +305,27 @@ public override Texture GetTexture(string componentName)
 
         public override SampleChannel GetSample(ISampleInfo sampleInfo)
         {
-            foreach (var lookup in sampleInfo.LookupNames)
+            var hsi = sampleInfo as HitSampleInfo;
+
+            if (hsi?.IsLayered == true && GetConfig<LegacySkinConfiguration.LegacySetting, bool>(LegacySkinConfiguration.LegacySetting.LayeredHitSounds)?.Value == false)
```
smoogipoo, line 310: This isn't being read from the skin configuration. You need to add a value to the `LegacySetting` switch above.

smoogipoo (changes requested):

peppy:
@Fire937 just looping back – are you intending to apply the remaining review?

Fire937:
@peppy yes I will try and commit in not too long.

bdach:
Seeing that there stlll has been no changes here in a while, I would be looking to supersede this PR over the coming days. I don't usually do that but this one would take too much work to fix up - I would basically have to revert all of the changes here, and even then there are major issues left unaddressed (such as how to make the sample fallback logic stick to a silent sample if layered hit sounds are to be disabled).

PR #8687, opened 2020-04-08
Allow DrawableCarouselBeatmapSet to check over its difficulty panels' positions

### LINKED ISSUE #8390: Panels in beatmap carousel don't pop open correctly after scrolling past the end
**Describe the bug:** 
After you scroll past the end of the beatmap carousel, sets directly after sets with many difficulties (like 8+ difficulties it seems) start popping open with incorrect vertical positions for their set panels. 

Expected result: set panel should appear above all its difficulty panels 

Actual result: set panel appears inside or below its difficulty panels

https://youtu.be/Cikc8jMDllM

"[Okaerinasai](https://osu.ppy.sh/beatmapsets/395642)" and "[Sore wa Chiisana Hikari no Youna](https://osu.ppy.sh/beatmapsets/430934)" are the beatmaps effected here because "[MIIRO vs. Ai no Scenario](https://osu.ppy.sh/beatmapsets/325158)" and "[Sakebe](https://osu.ppy.sh/beatmapsets/550344)" both have many difficulties.

**osu!lazer version:**
2020.320.1

### LINKED ISSUE #8391: Panels in beatmap carousel don't pop open correctly after deleting some amount of beatmaps
Probably related to #8390

**Describe the bug:** 
After you delete many (idk how many) beatmaps, sets start popping open with incorrect vertical positions for their difficulty panels.

Expected result:  difficulty panels should appear right below their parent's set panel

Actual result: difficulty panels appear way below their parent set panel, sometimes inside the next set's set panel

Managed to trigger the bug at 1:06 or so https://youtu.be/jVbm1Tvgz0E?t=68

**osu!lazer version:** 
2020.320.1

LittleEndu:
Closes #8390
Closes #8391 

This PR makes it so that the validity of ``DrawableCarouselBeatmap``'s position is checked by their parent ``DrawableCarouselBeatmapSet``. Beatmap's position is considered valid if a) they aren't higher than their parent set panel and b) they aren't lower than their final position in relation to their parent set's panel would be.

Crude drawing of what the logic looks like:
https://i.imgur.com/NVhoLPg.jpg
``■`` is the beatmap set's panel, ``•`` is the difficulty panel, arrows are their expected movement. Blue line is the minimum allowed Y for difficulty panels, orange is the maximum. If any of these are exceeded, they are brought to the line.


osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -111,6 +111,30 @@ private void load(BeatmapManager manager, BeatmapSetOverlay beatmapOverlay)
...
+
+            var i = 0;
+
+            beatmapsToCheck.ForEach(beatmap =>
+            {
+                if (beatmap.Y > Y + DrawHeight + 6 + i * (beatmap.DrawHeight + 5))
+                    beatmap.MoveToY(Y + DrawHeight + 5 + i * (beatmap.DrawHeight + 5)).MoveToY(beatmap.DesiredY, 750, Easing.OutExpo);
+
+                if (beatmap.Y < Y + DrawHeight + 4)
+                    beatmap.MoveToY(Y + DrawHeight + 5).MoveToY(beatmap.DesiredY, 750, Easing.OutExpo);
```
bdach, line 132: Why `> Y + DrawHeight + 6` and `< Y + DrawHeight + 4`? Those look like off-by-ones (I would have expected 5 in both places? Keep in mind it's strict inequality)

osu.Game/Screens/Select/Carousel/DrawableCarouselItem.cs
```diff
@@ -25,6 +25,10 @@ public abstract class DrawableCarouselItem : Container
 
         public override bool IsPresent => base.IsPresent || Item.Visible;
 
+        public bool DrawableIsPresent => base.IsPresent;
```
bdach, line 28: Not a fan of this, it'll be intensely confusing to anyone that stumbles upon it. I'd just rather remove and revert the change that uses this.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -111,6 +111,30 @@ private void load(BeatmapManager manager, BeatmapSetOverlay beatmapOverlay)
             };
         }
 
+        private readonly List<DrawableCarouselBeatmap> beatmapsToCheck = new List<DrawableCarouselBeatmap>();
+
+        public void AssureBeatmapPositionValid(DrawableCarouselBeatmap beatmap) => beatmapsToCheck.Add(beatmap);
+
+        public void ClearAssurance() => beatmapsToCheck.Clear();
```
bdach, line 118: If it the beatmap set panels weren't entirely decoupled from the difficulty panels before the introduction of this list, this could have been a `osu.Framework.Caching.Cached` (*not* the DI attribute), but as things stand this list is needed in this current implementation.

bdach (commented):
I'm not sold on correcting the Y positions inside `DrawableCarouselBeatmapSet` looking at this, but I'm not sure doing this in the carousel would be better either.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
LittleEndu: I logged the positions of beatmap panels when they were triggering these checks. They seemed to be needlessly hitting them when these were set to 5 so the off-by-ones are intentionally here.

osu.Game/Screens/Select/Carousel/DrawableCarouselItem.cs
LittleEndu: Yeah, this can be reverted, but then the ``alpha == 0`` should be replaced with ``alpha <= 0.0001f`` as that is what ``IsPresent`` checks.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
bdach: Needlessly how? If it's due to floating point then I'd use something from `Precision` to have a check with tolerance. The way this is currently written looks buggy.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
LittleEndu: I was confused about this when I first read it. The panels actually need to be decoupled and to my knowledge always have been. The logic to remove offscreen panels from the scene graph relies on all the panels being decoupled. Otherwise removing one panel might remove stuff some others rely on. 
So yeah, this can't use caching.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
LittleEndu: I don't think it's due to floating point. I think the easing is just different enough (each panel starts from different position so their velocity will also differ) that the difference between Y positions will just grow larger until it will hit any ``acceptableDifference`` threshold. Debating removing the check altogether if the animation doesn't suffer from it.

LittleEndu:
Closing in favour of #10494

PR #8691, opened 2020-04-09
Fix HitCirclePrefix not working correctly when directory is specified

### LINKED ISSUE #8157: Skin Fonts do not load if directed to a subfolder
Same issue as #6058 though they still won't load from a subfolder inside the skin folder. Here the HitCirclePrefix `blank` is located inside the subfolder `hitcircle4`:
`[Fonts]`
`HitCirclePrefix: hitcircle4\blank`

Lazer 2020.306.0

peppy:
Closes https://github.com/ppy/osu/issues/8157.

### DIFF
--- a/osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
+++ b/osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
@@ -4,6 +4,7 @@
 using System;
 using osu.Framework.Audio.Sample;
 using osu.Framework.Bindables;
+using osu.Framework.Extensions;
 using osu.Framework.Graphics;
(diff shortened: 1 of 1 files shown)


smoogipoo:
Have you checked that this actually still is an issue on master? 

This was added in a recent mania PR, which should be handling this: https://github.com/ppy/osu/blob/e8f836a410c24c95cadec1f9595f35b57aafe084/osu.Game/Skinning/LegacySkinResourceStore.cs#L31

Is it somehow related to `LegacySpriteText`, and if so shouldn't we be fixing it there?

peppy:
Yep, did check to not be working on master (and fixed by the change). Will see why that line isn't working.

peppy:
Actually it looks like it *is* fixed on master and I was checking the wrong line.

PR #8696, opened 2020-04-09
Add undo/redo support to the Editor

### LINKED ISSUE #6143: Editor action history
The idea is to log events during mapping process like creating, moving, copying etc., linking properties and displacement of copied objects to their parents and allowing for fast changes across the map or preserving relations between objects without manually adjusting them.

This feature have great implementation in Autodesk Fusion 360 and provides great flexability in desing, which i miss while desinging maps.

smoogipoo:
Resolves https://github.com/ppy/osu/issues/6143

Only hitobject additions/removals are supported in this PR.

### LegacyEditorBeatmapDiffer

Uses the library `DiffPlex` to perform by-line diffing for now, which gives the lines changed as a set of hunks. The hitobjects which those lines reference are then added/removed from the `EditorBeatmap`.

This supports any arbitrary hitobject changes, and eventually I foresee us diffing other editor states like the selected hitobjects.

### EditorChangeHandler

Handles saving and restoring of editor changes. Binds to `EditorBeatmap` events to determine some general events, however also provides `BeginChange()` and `EndChange()` for more advanced use cases such as:

* Changing of slider paths or something inside a hitobject, for which there is no bound event in `EditorBeatmap`.
* Grouping of events such as mass deletions or movements.

Am up for suggestions regarding the naming here :/

### Future plans

The goal is to eventually integrate git, but with the way this has been laid it it should be simple to do - only needing to change the `LegacyEditorBeatmapDiffer`.


peppy:
Looks pretty good, but curious about if there's reasoning for storing the streams as you are, instead of `byte[]` (and making a `MemoryStream` where necesssary). Just find the seeking logic to be a bit weird, along with storing a `List<Stream>` not being something I've seen done before.

smoogipoo:
Sounds like it should work, and I'd be fine with that. Didn't think of it at the time.

peppy (approved):

PR #8699, opened 2020-04-09
Add stereo shifted hitsound playback support

### LINKED ISSUE #8387: Add stereo shifted hitsound playback support
**Describe the new feature:** 
If you listen carefully enough in stable you will notice some sounds are panned according where you hit stuff (e.g. taiko), I'd like to see this feature come back in osu!lazer.

Fire937:
Closes #8387

There is now a setting in the _Gameplay/general_ settings called _Positional
hitsounds_. If the setting is enabled, the hitsounds playback will be
shifted according to their position on the beatmap.

Settings panel


bdach:
I'm admittedly a little bit skeptical of this being done at the drawable hitobject level instead of in a more audio-specific place (like `SkinnableSound` maybe?) It feels like the wrong place to put this. Not sure whether it's just me, though.

Fire937:
I feel like the fact the position of the object is required in ordered to calculate the balance makes it hard to implement this in an audio-specific place. We would have to somehow pass the positional information to such an audio component, that has no reason to be aware of the object's position.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -101,10 +114,11 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
         protected DrawableHitObject([NotNull] HitObject hitObject)
         {
             HitObject = hitObject ?? throw new ArgumentNullException(nameof(hitObject));
+            positionalSoundAdjustment.Value = PositionalSound;
```
TheWildTree, line 117: Will this work properly in case of mods that move hitobjects around after they've been created? It looks like this only uses the initial position and doesn't consider any changes later on. 

I know it's only `Wiggle` for now - which won't really cause any noticeable differences here - but more extreme mods might come in the future where this might become a problem.

Fire937: You're right, and now that you mention it I see that I slightly deviated from peppy's expression of the PositionalSound. The value considered should indeed be the position of the drawable and not the one specified in the beatmap. I will try implementing this behaviour.

peppy:
@Fire937 SkinnableSound is actually a component in the draw hierarchy. Give it a shot in there - it should work.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -84,8 +91,14 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
         /// </summary>
         public JudgementResult Result { get; private set; }
 
+        /// <summary>
+        /// The stereo balance of the samples played if <i>Positional hitsounds</i> is set.
+        /// </summary>
+        protected virtual float PositionalSound => (Position.X / 512f - 0.5f) * 0.8f;
```
peppy, line 97: This is going to be wrong for (at last) mania. Also I'd call this `SamplePlaybackBalance` maybe.

peppy (changes requested):

Fire937:
Thank you, it does indeed work and seems better. However if we want the feature to be available in every ruleset, the ruleset needs to have a way to override `SamplePlaybackBalance`.  I feel like it would be easier to do in `DrawableHitObject`. Otherwise maybe we would need to somehow have `DrawableHitObject` use the ruleset's subclass of `SkinnableSound`.
What would be the preferred solution? IMO having the implementation in `SkinnableSound` is not worth the hassle that having the ruleset subclass `SkinnableSound` would be.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -113,6 +126,16 @@ private void load()
                 throw new InvalidOperationException($"{GetType().ReadableName()} must provide a {nameof(JudgementResult)} through {nameof(CreateResult)}.");
 
             loadSamples();
+
+            userPositionalHitSounds = config.GetBindable<bool>(OsuSetting.PositionalHitSounds);
+            userPositionalHitSounds.BindValueChanged(positional =>
+            {
+                if (positional.NewValue)
+                    Samples?.AddAdjustment(AdjustableProperty.Balance, positionalSoundAdjustment);
```
peppy, line 134: rather than adding and removing the adjustment, *always* bind the adjustment and then set the valule differently based on the enable/disable (ie. set to default balance if user does not want positional).

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
```diff
@@ -70,6 +70,8 @@ public abstract class DrawableCatchHitObject : DrawableHitObject<CatchHitObject>
 
         public float DisplayRadius => DrawSize.X / 2 * Scale.X * HitObject.Scale;
 
+        protected override float SamplePlaybackBalance => 0.8f * HitObject.X - 0.4f;
```
peppy, line 73: where does 0.8/0.4 come from here?

peppy: right, i see where you're coming from. going to change the 0.8f into a shared constant for legibility.

peppy:
I made a few changes:

- Removed the osu!mania traversal by using DI (we try our best to not traverse the tree).
- Changed the range provided by the DHO method to 0..1. This seems easier for consumers to implement than -1 to 1.
- Applied the `0.4f` dampening at the base level, since it's applied across all rulesets.

Also just some general cleanup around the place.

@smoogipoo requesting a quick check of this, mainly to make sure you agree with the DHO interface for this.

smoogipoo (approved):
Looks good to me

PR #8701, opened 2020-04-09
Implement period-tracking component

frenzibyte:
To be used in #8632 

`PeriodTracker` is a component used to allow tracking whether a time value falls within any of the added periods to the tracker, this has been written to allow using that logic in any component that need it, currently used in `BreakTracker` and also to-be-used for #8632.

## Notes
- The `BreakTracker.CurrentBreakIndex` has been removed due to no longer being useful, although it turns out to be used in one of the test cases in `TestSceneBreakTracker`, but it didn't really have a purpose there so I've removed it, if anything we could potentially add an `out` parameter in the checking method providing the index of the interval if a value is inside one.
- The `BreakTracker.Breaks` getter has been removed due to it causing redundancy to the logic after using `IntervalList<double>`, so I've removed it and changed its usages to take the breaks directly from the beatmap instead. Should likely remove the getter from any component that provides one for no reason at some point.


bdach:
why is this a full collection with add/remove instead of a simple readonly structure...?

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,128 @@
...
+using System.Collections.Generic;
+using osu.Framework.Lists;
+
+namespace osu.Game.Lists
+{
+    /// <summary>
+    /// Represents a list of intervals that can be used for whether a specific value falls into one of them.
+    /// </summary>
+    /// <typeparam name="T">The type of interval values.</typeparam>
+    public class IntervalList<T> : ICollection<Interval<T>>, IReadOnlyList<Interval<T>>
```
bdach, line 15: this doesn't make sense for things that are not numbers (there is no such thing as an interval of strings). if C# had traits I'd use that, but it doesn't so why not just use `float` or something?

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,128 @@
...
+    {
+        private static readonly IComparer<T> type_comparer = Comparer<T>.Default;
+
+        private readonly SortedList<Interval<T>> intervals = new SortedList<Interval<T>>((x, y) => type_comparer.Compare(x.Start, y.Start));
+
+        /// <summary>
+        /// The index of the nearest interval from last <see cref="IsInAnyInterval"/> call.
+        /// </summary>
+        protected int NearestIntervalIndex;
```
bdach, line 24: why `protected`?

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,128 @@
...
+
+        #region IReadOnlyList<Interval<T>>
+
+        public Interval<T> this[int index]
+        {
+            get => intervals[index];
+            set => intervals[index] = value;
+        }
+
+        #endregion
```
bdach, line 105: why? who's going to ever use these?

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,128 @@
...
+        public readonly T Start;
+
+        /// <summary>
+        /// The end value of this interval.
+        /// </summary>
+        public readonly T End;
+
+        public Interval(T start, T end)
+        {
+            bool startLessThanEnd = Comparer<T>.Default.Compare(start, end) < 0;
```
bdach, line 122: no, don't silently fix things up for users. throw an exception instead. this makes it possible to introduce subtle bugs.

bdach (changes requested):
way too overdone IMO

osu.Game/Lists/IntervalList.cs
iiSaLMaN: I don't think we should expose such an internal field as `NearestIntervalIndex`, it's used for the search inside the method only, changed to `protected` for testing what the index becomes to ensure that it doesn't break.

osu.Game/Lists/IntervalList.cs
iiSaLMaN: implementation of `ICollection<Interval<T>>` and `IReadOnlyList<Interval<T>>`, I guess what your review suggests is to make it so simple that the implementation becomes enough for its usages.

osu.Game/Lists/IntervalList.cs
bdach: the answer is not to test it. you're testing an implementation detail. test the external behaviour only instead (as in, treat the list as a black box that only has `IsInAnyInterval`).

exposing is wrong design-wise as it breaks encapsulation. this is internal state that should not be visible to inheritors.

iiSaLMaN:
> why is this a full collection with add/remove instead of a simple readonly structure...?

I've reduced it enough to satisfy all conversations, but I'm gonna keep the add/remove methods just to actually simplify the process of adding intervals.

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,111 @@
...
+        private static readonly IComparer<T> type_comparer = Comparer<T>.Default;
+
+        private readonly SortedList<Interval<T>> intervals = new SortedList<Interval<T>>((x, y) => type_comparer.Compare(x.Start, y.Start));
+        private int nearestIndex;
+
+        public Interval<T> this[int i]
+        {
+            get => intervals[i];
+            set => intervals[i] = value;
+        }
```
bdach, line 27: Not required and potentially buggy (consider what happens if you do
```csharp
intervalList[1] = new Interval<double>(15, 20);
```
after construction - will it break? will it be fine? I don't know, but it shouldn't be possible in the first place)

osu.Game/Lists/IntervalList.cs
```diff
@@ -0,0 +1,111 @@
...
+        public bool Remove(Interval<T> interval) => intervals.Remove(interval);
+
+        /// <summary>
+        /// Removes all intervals from the list.
+        /// </summary>
+        public void Clear() => intervals.Clear();
+
+        public IEnumerator<Interval<T>> GetEnumerator() => intervals.GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
```
bdach, line 86: not required IMO

bdach (dismissed):
I'm still of the opinion that `IntervalList` should be a non-generic, read-only, *immutable* data structure.

iiSaLMaN:
> I'm still of the opinion that `IntervalList` should be a non-generic, read-only, _immutable_ data structure.

I don't mind making it non-generic, but are you fine with creating another normal list to use for adding calculated intervals to store inside `IntervalList` (which I will rename to `IntervalTracker` fwiw)

bdach:
I'm not sure what you're asking. I think the list/tracker should take in a `IEnumerable<Interval>` or similar.

iiSaLMaN:
like how the hp no-draining sections are calculated:
```csharp
private readonly PeriodTracker nonDrainingSections = new PeriodTracker();

public override void ApplyBeatmap(IBeatmap beatmap)
{
    ...

   foreach (var b in beatmap.Breaks)
   {
       var start = beatmap.HitObjects.Select(h => h.EndTime).LastOrDefault(e => e < b.StartTime);
       var end = beatmap.HitObjects.Select(h => h.StartTime).FirstOrDefault(s => s > b.EndTime);

       nonDrainingSections.AddPeriod(start, end);
   }
}
```

with no `AddPeriod`:
```csharp
private readonly PeriodTracker nonDrainingSections = new PeriodTracker();

public override void ApplyBeatmap(IBeatmap beatmap)
{
    ...

    nonDrainingSections.Periods = beatmap.Breaks.Select(b =>
    {
        var start = beatmap.HitObjects.Select(h => h.EndTime).LastOrDefault(e => e < b.StartTime);
        var end = beatmap.HitObjects.Select(h => h.StartTIme).FirstOrDefault(s => s > b.EndTime);
        return new Period(start, end);
    }
}
```

*actually nvm, it appears to be not that bad*

smoogipoo:
You'll want to revert your last commit. Also I agree with the notion that this should be immutable.

iiSaLMaN:
> You'll want to revert your last commit. Also I agree with the notion that this should be immutable.

I would agree that the design for referenced commits are confusing here but what you're seeing is just a reference to this PR from 10badc09d53d5586d065ed49f7e2cae09cfb4e9c, unless I'm mistaking for what you mean by "last commit"?

also I have changes ready for this, just discussing one last thing about the component in the above PR. (where it should reside)

smoogipoo:
Ah I misread, my mistake.

bdach:
What's the status on this PR? You've mentioned that you had the changes ready but there's been no updates for two weeks.

iiSaLMaN:
> What's the status on this PR? You've mentioned that you had the changes ready but there's been no updates for two weeks.

Been busy with other PRs + the windows to macOS switch (which I accidentally wiped all local changes in there), I'll get it ready by this day or tomorrow, sorry for this massive delay.

iiSaLMaN:
@bdach Pushed. couldn't get it in time, but I've taken a good look on it and hopefully it is in its best shape to you.

bdach:
I've made changes here (mostly cosmetic, with one major removal of the very strange setter that was there before). I think this is tolerable now.

osu.Game/Screens/Play/BreakTracker.cs
```diff
@@ -2,40 +2,37 @@
...
             set
             {
-                breaks = value;
-
-                // reset index in case the new breaks list is smaller than last one
                 isBreakTime.Value = false;
-                CurrentBreakIndex = 0;
+
+                breaks = new PeriodTracker(value.Where(b => b.HasEffect)
+                                                .Select(b => new Period(b.StartTime, b.EndTime - BreakOverlay.BREAK_FADE_DURATION)));
```
smoogipoo, line 35: Is this supposed to have changed, by the inclusion of the fade duration?

iiSaLMaN: This hasn't been changed, the old logic uses `BreakPeriod.Contains(time)` which checks by `time >= Start && time <= (End - BREAK_FADE_DURATION)`. With the `PeriodTracker` component, I wrote it as you can see. (add periods that `b.HasEffect` with exclusion of the fade duration at the end)

bdach: I kind of mindlessly glanced over that, good spot. This has the potential to crash for "too short" breaks (because `Period` checks if `start < end`).

smoogipoo: In this case it's safe because of the `HasEffect` check before it.

smoogipoo (approved):
I think this is fine for the time being. Keeping in mind that `BreakTracker` will likely be merged into `GameplayBeatmap`, and that's probably where the draining "tracker" should also be situated going forward.

osu.Game/Screens/Play/BreakTracker.cs
iiSaLMaN: Yes I've thought about that the time I wrote this, `HasEffect` checks if the duration is longer than `650`, so no problem there. I was just hesitant a bit since this logic will actually fail if that check inside the property changes.

I guess it's fine for this exact case.

bdach: `BREAK_FADE_DURATION` is explicitly tied to `MIN_BREAK_DURATION` so it should be pretty obvious something is going on there if it's ever changed.

PR #8703, opened 2020-04-10
Fix EditorBeatmap potentially not updating hitobjects

smoogipoo:


osu.Game/Screens/Edit/EditorBeatmap.cs
```diff
@@ -74,15 +75,27 @@ public EditorBeatmap(IBeatmap playableBeatmap)
         private void updateHitObject([CanBeNull] HitObject hitObject, bool silent)
         {
             scheduledUpdate?.Cancel();
-            scheduledUpdate = Scheduler.AddDelayed(() =>
+
+            if (hitObject != null)
+                pendingUpdates.Add(hitObject);
+
+            scheduledUpdate = Schedule(() =>
             {
```
peppy, line 83: i guess using schedule is fine, although feels like `Cached` might be more appropriate.

peppy (approved):

PR #8709, opened 2020-04-11
Cache ruleset dependencies for test scenes that require it

frenzibyte:
- [x] Depends on #8778 

Resolves an issue found by @LumpBloom7 (testing ruleset-specific components that retrieve resources added directly to the ruleset assembly itself doesn't work as the resources are not retrieved successfully).

This resolves the issue above by moving ruleset dependencies caching logic into its own dependency container and passing it to the `OsuTestScene` dependencies if the test scene instance overrides `CreateRuleset()` to the ruleset corresponding to the component (as mentioned in its xmldoc).

## Remarks
- I've also chosen to cache the ruleset dependencies instead of adding the `DrawableRuleset` as it may be an overkill when attempting to test simple functionality of a component that requires ruleset resources.
- Since the ruleset-specific store have been moved to `DrawableRulesetDependencies`, I've made it an `IDisposable` such that we can dispose the stores from whoever owns the container.


osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -64,7 +68,12 @@ protected IAPIProvider API
 
         protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
         {
-            Dependencies = new OsuScreenDependencies(false, base.CreateChildDependencies(parent));
+            var baseDependencies = base.CreateChildDependencies(parent);
+
+            if (this is IRulesetTestScene rts)
```
smoogipoo, line 73: I think this is a little bit over-engineeed, and it is probably better that a `CreateRuleset()` method is directly exposed from `OsuTestScene`. Would immediately replace all instances of `CreateRuleset*()` in existing test scenes - there's quite a few already.

@peppy thoughts?

smoogipoo (changes requested):

osu.Game/Tests/Visual/OsuTestScene.cs
peppy: and always caching it (with `OsuRuleset` as a sane default)? i'm not necessarily against that.

was unsure about the method used here since it feels like probably too much interface magic for tests

smoogipoo: I'm not sure if the `OsuRuleset` default is required (thinking forward to splitting all rulesets out into their own repos), but sure.

peppy: @iiSaLMaN wanna give this a try? removing the interface logic and moving `CreateRuleset` into `OsuTestScene`.

iiSaLMaN: Yes sorry, was busy setting up my new MacBook for work yesterday, will get to this.

iiSaLMaN: @peppy Although the `OsuTestScene` is a component inside the `osu.Game` assembly, which I think it should never reference any ruleset? will push without the ruleset default.

iiSaLMaN: Actually *luckily*, since I've used `Ruleset.Value` as the value of the ruleset, and as it always has the first available ruleset (most likely `OsuRuleset`), the test scene is already technically provides a default ruleset.

smoogipoo (approved):
lgtm!

PR #8711, opened 2020-04-11
Add basic taiko "hit" skinning support

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/97

This does some general maintenance on the hierarchy of taiko hits / pieces to actually allow skinning to exist. Let me know if you'd prefer those changes were split out into another PR (already a separate commit in this PR).

image


smoogipoo (approved):
lgtm

PR #8714, opened 2020-04-11
Add support for testing arbitrary API requests/responses via Dummy API

peppy:


osu.Game/Online/API/APIRequest.cs
```diff
@@ -16,20 +17,27 @@ public abstract class APIRequest<T> : APIRequest where T : class
...
         /// <summary>
         /// Invoked on successful completion of an API request.
         /// This will be scheduled to the API's internal scheduler (run on update thread automatically).
         /// </summary>
         public new event APISuccessHandler<T> Success;
+
+        internal void TriggerSuccess(T result)
+        {
+            // disallow calling twice
+            Debug.Assert(Result == null);
```
smoogipoo, line 36: I think this should be an exception instead. It's `internal` sure, but it's still much too open imo.

osu.Game/Online/API/DummyAPIAccess.cs
```diff
@@ -55,6 +61,7 @@ public DummyAPIAccess()
 
         public virtual void Queue(APIRequest request)
         {
+            HandleRequest?.Invoke(request);
         }
 
         public void Perform(APIRequest request) { }
```
smoogipoo, line 67: `HandleRequest` should be added here and in `PerformAsync` too

smoogipoo (changes requested):

bdach (approved):
Good stuff, will definitely help with testing online 👍

smoogipoo (approved):

PR #8716, opened 2020-04-11
Implement "prefer no-video" option

### LINKED ISSUE #8705: Option to "Prefer no-video downloads"
**Describe the new feature:** 
On metered or slow connections having an option to download beatmaps with no video would be nice to have.

Everywhere where osu!direct would download a beatmap without the user explicitly saying what type (video or not) they want, option stated in this new setting should be used (instead of defaulting to video included version)

LittleEndu:
Closes #8705

This PR adds a simple option in the settings. The option is added into "Web" category. It's used when default behaviour of a download button isn't clear/specified.

``BeatmapManager.Download`` seems to be used by direct, multiplayer, and listing. This PR doesn't change listing, as this is where user could force the download type opposite to their setting. Direct panel checks if it has a default value, if not it uses the one set in the settings. Not sure if there is a better way to do that in C#. Also not sure if multiplayer works correctly.


osu.Game/Overlays/Direct/PanelDownloadButton.cs
```diff
@@ -14,12 +15,12 @@ public class PanelDownloadButton : BeatmapDownloadTrackingComposite
...
         protected bool DownloadEnabled => button.Enabled.Value;
 
-        private readonly bool noVideo;
+        private readonly bool? noVideo;
 
         private readonly ShakeContainer shakeContainer;
         private readonly DownloadButton button;
 
-        public PanelDownloadButton(BeatmapSetInfo beatmapSet, bool noVideo = false)
+        public PanelDownloadButton(BeatmapSetInfo beatmapSet, bool? noVideo = null)
```
bdach, line 23: I'm a little perplexed by this. As far as I can see, there isn't a single usage where `noVideo` is ever explicitly provided, it's all defaults or delegating ctors. Therefore I don't really know what the `null` value is supposed to convey - could you explain this choice? xmldoc would also be helpful for future travellers if this stays.

osu.Game/Overlays/Direct/PanelDownloadButton.cs
```diff
@@ -66,7 +67,8 @@ private void load(OsuGame game, BeatmapManager beatmaps)
                         break;
 
                     default:
-                        beatmaps.Download(BeatmapSet.Value, noVideo);
+                        var minimiseDownloadSize = noVideo ?? osuConfig.GetBindable<bool>(OsuSetting.PreferNoVideo).Value;
```
bdach, line 70: I'm not sure that the bindable should be fetched every time the button is clicked. Might be a little leaner by having a single local bound copy in the button itself and reading from that instead.

bdach (commented):
Verified as working. I would have liked to see a test for this, as this is the sort of functionality that could very silently regress on us until somebody notices months down the line. However this is online and online is usually awkward to test, so I guess I could live without one, albeit I think writing one would be *just barely* possible.

The way I'd see testing this is by subclassing `BeatmapManager`, hooking into `CreateDownloadRequest`, storing value of `minimiseDownloadSize` for assertions there, and then forcing the test manager to be injected into the button via a providing container:

```csharp
public class TestBeatmapManagerProvidingContainer : Container
{
    [Cached]
    public TestBeatmapManager BeatmapManager { get; set; } = new TestBeatmapManager();

    public TestBeatmapManagerProvidingContainer()
    {
        Child = new TestDownloadButton(beatmap); // etc etc
    }
}
```

osu.Game/Overlays/Direct/PanelDownloadButton.cs
LittleEndu: I would remove the param and have the setting be the only thing that matters. Wasn't sure if this was fine.

osu.Game/Overlays/Direct/PanelDownloadButton.cs
LittleEndu: Can you even press these buttons more than once? But sure, this should be changed, yes.

bdach: Possibly, if the download fails.

LittleEndu: Not sure if I did it correctly. It seems that if you change the setting while in direct, the setting doesn't apply to old panel buttons (or maybe even this direct session?)

osu.Game/Overlays/Direct/PanelDownloadButton.cs
```diff
@@ -52,6 +51,8 @@ private void load(OsuGame game, BeatmapManager beatmaps)
                 return;
             }
 
+            noVideoSetting.BindTo(osuConfig.GetBindable<bool>(OsuSetting.PreferNoVideo));
```
bdach, line 54: Yeah, this looks unsafe. `GetBindable` already returns a copy, and as far as I know bindables use weak references, so the inner bindable could go out of scope and get collected, which means this local copy won't get value updates.

To resolve, make `noVideoSetting` non-readonly and just assign:
```suggestion
            noVideoSetting = osuConfig.GetBindable<bool>(OsuSetting.PreferNoVideo);
```
(I wouldn't initialise `noVideoSetting` either, let it be `null` until `load()` is done)

bdach (changes requested):

peppy (approved):

PR #8728, opened 2020-04-12
Let beatmap carousel scroll to screen middle

LittleEndu:
There's a nice arrow formed by two shadows in song select. That arrow is drawn to the middle of the screen (both shadows are same size). But the carousel scrolls using its own draw height and doesn't consider the draw height of the container its in.

This PR makes carousel consider its parent's padding when choosing where to scroll.

CenterPR
carouselDrawLogic


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -632,7 +632,7 @@ private void updateItems()
                         case DrawableCarouselBeatmap beatmap:
                         {
                             if (beatmap.Item.State.Value == CarouselItemState.Selected)
-                                scrollTarget = currentY + beatmap.DrawHeight / 2 - DrawHeight / 2;
+                                scrollTarget = currentY + beatmap.DrawHeight / 2 - (Parent.DrawHeight / 2 - Parent.Padding.Top);
```
bdach, line 635: While this change seems correct it took me about half an hour to work out what this exactly did. The double negation especially threw me in for a loop. I'd say this should read
```suggestion
                                scrollTarget = currentY - Parent.DrawHeight / 2 + Parent.Padding.Top + beatmap.DrawHeight / 2;
```
and potentially have a comment explaining what the last part is (first subtracting half of height to end up with the element at the middle of the screen, and then adding back possible top padding from parent and then half of our own height to achieve vertical centering of the item itself).

LittleEndu: Yeah I drew a graph for myself with like 6 different variables and arrived at the correct solution via substitution into one of the equations. Should have included that in the PR description at the very least, maybe even in code as a comment.

LittleEndu: I made my crude drawing into following image, should this be included as a comment?
carouselDrawLogic

bdach: The image is nice but I'd probably leave it as part of this PR and have a textual comment in the source.

LittleEndu: Is having a link to PR in the comment fine?

bdach: I'd much prefer just plain text, no links. Links are prone to breakage. (You could say that if github goes away we have bigger problems but IMO source should be stand-alone as much as possible.)

LittleEndu: Could I ask you to write that comment, like edit your original suggestion? I can't actually figure out a nice way to comment that in plain text.

bdach: Maybe something like
```csharp
// scroll position at currentY makes the set panel appear at the very top of the carousel in screen space
// move down by half of parent height (which is the height of the carousel's visible extent, including semi-transparent areas)
// then reapply parent's padding from the top by adding it
// and finally add half of the panel's own height to achieve vertical centering of the panel itself
```

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -632,7 +632,11 @@ private void updateItems()
                         case DrawableCarouselBeatmap beatmap:
                         {
                             if (beatmap.Item.State.Value == CarouselItemState.Selected)
-                                scrollTarget = currentY + beatmap.DrawHeight / 2 - DrawHeight / 2;
+                                // scroll position at currentY makes the set panel appear at the very top of the carousel in screen space
+                                // move down by half of parent height (which is the height of the carousel's visible extent, including semi-transparent areas)
+                                // then reapply parent's padding from the top by adding it
+                                // and finally add half of the panel's own height to achieve vertical centering of the panel itself
+                                scrollTarget = currentY - Parent.DrawHeight / 2 + Parent.Padding.Top + beatmap.DrawHeight / 2;
```
smoogipoo, line 639: I'm not a fan of tying this component to its parent like this. It's not a component that is tightly coupled to song select.

I'd like to see a `ScrollOffset` property instead which `SongSelect` sets, and for this to only reference our local size.

smoogipoo (changes requested):

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: If song select were to set a scroll offset, this (carousel) couldn't reference its own size. Referencing its own size was what caused this bug to begin with. Song select would need to calculate the correct offset based on the size of the container carousel is put in. 

Also saying that carousel isn't tightly coupled to song select is kinda wrong.

bdach: I think what the suggestion meant is to make `SongSelect` pass in the `- Parent.DrawHeight / 2 + Parent.Padding.Top` part. That way there's no need to directly reference the parent anymore.

LittleEndu: That's how to do it but I didn't get that from smoogi's suggestion. It sounded more like revert back to original and have song select try to calculate some magic offset from that.

I would still argue that using parent is OK in this case tho.

LittleEndu: Either using Parent or use bleed_top and visible half height instead

bdach: Actually yeah, that would work.

smoogipoo: As in, without a parent setting `ScrollOffset`, it _should_ centre it perfectly, assuming the parent also doesn't set padding/etc. So yes your understanding that it should reference its own size is correct, and that the parent's offset should apply on top of that.

But given that the parent and child size are related in some fashion (i.e. by the padding), it shouldn't be all that "magic" of a value.

What if the parent gives a different position, or changes its size, etc, rather than adjusting the top padding? That's what I mean by I don't want this to be tightly coupled to song select, because its layout now strictly depends on how SS it laid out.

LittleEndu: I was saying that the offset would look like a magic value because it would probably be the result of substituting two different equations. As in the parent first needs to undo the offset the child does and then apply its own offset.

LittleEndu: Did some mental math, if carousel were to set scroll to `currentY - DrawHeight / 2 + beatmap.DrawHeight / 2 + offset` then the offset would be `padding.top / 2 - padding.bottom / 2`. Not sure how to comment that again.

LittleEndu: Here's an idea tho. Why I was even able to suggest using ``bleed_top`` and ``visibleHalfHeight`` instead is because ``bleed_top`` and ``Parent.Padding.Top`` just happen to be set to the same value: ``FilterControl.HEIGHT``. If the carousel were to ever appear independently, this value wouldn't make sense. If any value were to be provided by the parent drawable it should be `bleed_top`` and ``bleed_bottom``. Then the carousel would have all the information needed to scroll its content to the middle for any situation its drawn in.

LittleEndu: Would like to know what's the preferred way to go forward.

bdach: Myself I don't really mind either option, I just wouldn't want it to be half-and-half (so I'd dislike a solution where the offset is passed in but `bleed_{top,bottom}` remain). So I'd advocate for either using these constants directly for centering, or removing them and passing them through public properties explicitly.

LittleEndu: I guess passing them through public properties is good enough for now and if an offset is ever needed it could be added in later. But right now when bleed_top (or BleedTop if it's public) is used an offset would be unnecessary.

bdach:
Made some simple touch-ups here. Seems good to go to me.

smoogipoo (approved):
This is looking fine to me now, 👍

PR #8729, opened 2020-04-12
Fix potential nullref while testing beatmap carousel

LittleEndu:
There's a potential nullref when carousel doesn't have a selection but ``selectNextDifficulty`` is called. This can't happen in regular runtime because ``CarouselRoot``/``CarouselGroupEagerSelect`` would always have something selected, but it can happen in testing where the call to have eager selection seems to be missing. 

Currently when selected beatmap is null, and selectNextSet is called to select previous set, the selected set would be second to last. That's because calling ``IndexOf(null)`` results in -1 which makes the function think that the last set is selected. Don't know if this is something that should be changed just because of testing.


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -276,6 +276,13 @@ private void selectNextSet(int direction, bool skipDifficulties)
 
         private void selectNextDifficulty(int direction)
         {
+            if (selectedBeatmapSet == null || selectedBeatmap == null)
```
bdach, line 279: The beatmap check feels a little gratuitous. `CarouselBeatmapSet` should ensure that at least one child is selected, and trying to create a beatmapset without difficulties results in it just not being added to carousel.

bdach (commented):
Not overly sure about this direction. It's either this or making the carousel eagerly select everywhere, I'm not sure which is better. I do think however that if this approach stays, selecting previous difficulty with `null` current should select the last diff of the last set.

peppy:
Can you add a test covering this fail case?

LittleEndu:
> Can you add a test covering this fail case?

That would be easy, but do we have a agreed upon behaviour for this case? Should the carousel just select the very first or the very last difficulty when <kbd>↑</kbd> or <kbd>↓</kbd> is used?

peppy:
Yes, it should select the first when going down, or the last when going up.

bdach:
Think this is good to go now.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -272,12 +272,33 @@ public void SelectNext(int direction = 1, bool skipDifficulties = true)
...
+                return;
+            }
+
             if (skipDifficulties)
                 selectNextSet(direction, true);
             else
                 selectNextDifficulty(direction);
         }
 
+        private void performInitialSelection(int direction, bool skipDifficulties)
```
peppy, line 287: so this whole logic path is added but only ever executed in a test scene? that seems a bit suss to me. can we not just trigger selection on the root?

peppy:
@LittleEndu i've simplified this to just not crash but not select. i disagree in adding extra complexity as you did just for potential test failures.

LittleEndu:
Sounds fair but it seems most tests now need help with initial selection

peppy:
You're welcome to fix that if you beat me.

LittleEndu:
I won't be near computer for 6 hours :/

peppy:
I'm thinking of just closing this PR. It seems pointless to fix.

PR #8731, opened 2020-04-12
Make beatmap info overlay present selected difficulty

### LINKED ISSUE #8719: Already downloaded button in beatmap listing should present the selected difficulty
**Describe the bug:** 
When pressing the download button for an already downloaded map, that beatmap is presented. When in beatmap listing overlay, the selected difficulty should be presented.

https://youtu.be/OhXmJMbbzms
Highest difficulty is selected, seaming random difficulty is presented (maybe another bug, would expect easiest to be selected on current release).

**osu!lazer version:** 
2020.408.1

LittleEndu:
Fixes #8719

This PR adds a predicate to ``PresentBeatmap`` that allows to specify what difficulty we want to present. ``PanelDownloadButton`` then uses that predicate to say that the selected difficulty (from ``BeatmapPicker``) should be the difficulty ``PresentBeatmap`` tries to select.


osu.Game/OsuGame.cs
```diff
@@ -315,8 +315,12 @@ private void load()
         /// The user should have already requested this interactively.
         /// </summary>
         /// <param name="beatmap">The beatmap to select.</param>
-        public void PresentBeatmap(BeatmapSetInfo beatmap)
+        /// <param name="findPredicate">Predicate used to find a difficulty to select</param>
```
bdach, line 318: I'd describe this properly and remove the inline comment:
```suggestion
        /// <param name="findPredicate">
        /// Optional predicate used to try and find a difficulty to select.
        /// If omitted, this will try to present the first beatmap from the current ruleset.
        /// In case of failure the first difficulty of the set will be presented, ignoring the predicate.
        /// </param>
```
Also I'd maybe rename to something like `difficultyCriteria`.

osu.Game/OsuGame.cs
```diff
@@ -315,8 +315,12 @@ private void load()
...
+        public void PresentBeatmap(BeatmapSetInfo beatmap, Predicate<BeatmapInfo> findPredicate = null)
         {
+            // Use this predicate if non was provided. This will try to find some difficulty from current ruleset so we wouldn't have to change rulesets
```
bdach, line 321: if this stays:
```suggestion
            // Use this predicate if none was provided. This will try to find some difficulty from current ruleset so we wouldn't have to change rulesets
```

osu.Game/Overlays/Direct/PanelDownloadButton.cs
```diff
@@ -16,6 +18,8 @@ public class PanelDownloadButton : BeatmapDownloadTrackingComposite
 
         private readonly bool noVideo;
 
+        public Bindable<BeatmapInfo> CurrentBeatmap = new Bindable<BeatmapInfo>();
```
bdach, line 21: This is usually done as
```suggestion
        public IBindable<BeatmapInfo> CurrentBeatmap { get; } = new Bindable<BeatmapInfo>();
```
The interface prevents external sets to `.Value`, and the lack of setter makes it impossible to replace the bindable (but does still allow binding via `.BindTo()`/`BindTarget`). Less foot-shooty if there are `BindValueChanged` subscriptions (I imagine plain overwrite could then result in a memory leak).

bdach (dismissed):
Needs tests; `TestScenePresentBeatmap` exists.

osu.Game/Overlays/Direct/PanelDownloadButton.cs
LittleEndu: So ``BintTo()`` is same as setting the value to ``GetBoundCopy()``?

LittleEndu:
About tests tho. ``TestScenePresentBeatmap`` tests the raw functionality. It could be made to also test the ``findPredicate``/``difficultyCriteria`` and if that's enough then all is fine, but I don't think it could be made to test that beatmap info overlay presents the selected difficulty.

osu.Game/Overlays/Direct/PanelDownloadButton.cs
bdach: Strictly the same - no. But as can be seen from [the implementation](https://github.com/ppy/osu-framework/blob/42202ea1ab1d55a930a9ab3a3638561b62ba1d4a/osu.Framework/Bindables/Bindable.cs#L389-L400) both can work in this scenario and the flow of passing `BindTarget` is something I've seen more.

Generally from what I've seen the usage pattern with bindables is to hold a single private (readonly even, if permissible) bindable reference in a field and keep using it throughout the component's lifetime instead of reallocating/reassigning as that is easy to mess up.

bdach:
Just testing the new presenting behaviour is fine, testing integration with the beatmap overlay is finicky and probably not worth it.

bdach:
I touched up one thing (check the `BindTarget` usage in the last commit to save a local). Looks OK otherwise, but not approving since I pushed here.

peppy (dismissed):

peppy (approved):

PR #8735, opened 2020-04-13
Move "tracking" components to their own namespace

frenzibyte:
## Summary

After implementation of `PeriodTracker` from #8701, I've realized that component didn't really have a namespace to reside in, so I've looked up all components that share similarity with it and I've found out that `BreakTracker` and `IdleTracker` are best cases, so I've moved them all up to their own namespace `osu.Game.Tracking` and will add `PeriodTracker` to `osu.Game.Tracking.Components` as a sub-component useful for tracking components and other usages.

## Notes
- I've reorganized the using statements of `OsuGame` (which uses `IdleTrakcer`) as it was a bit hurting to my eyes (at least for me).


peppy:
This makes little sense IMO. You've moved components which are doing completely different things in different areas of the game to a common namespace just because they have "track" in their name?

iiSaLMaN:
> This makes little sense IMO. You've moved components which are doing completely different things in different areas of the game to a common namespace just because they have "track" in their name?

I was a bit hesitant about this actually, I should've asked in the server instead..., anyways what do you think is the best namespace of the `PeriodTracker` component? it's a component that takes a list of periods and has a method with a value parameter that returns whether the provided value falls within any of the provided periods, a component that is used by `BreakTracker` and to be used somewhere else. (see #8632)

peppy:
Put it alongside the other two classes>

iiSaLMaN:
> Put it alongside the other two classes

What do you mean? the two classes are of completely different areas, do you mean to make the component inside `osu.Game.Tracking` in itself?

peppy:
Where are these two other components going to sit?

iiSaLMaN:
to where they were originally:
- `BreakTracker`: `osu.Game.Screens.Play`
- `IdleTracker`: `osu.Game.Input`

PR #8736, opened 2020-04-13
Rework slider control point placement to improve path progression

### LINKED ISSUE #8726: Slider placement incorrectly places red control points
In `SliderPlacementBlueprint.OnDoubleClick`, it's changing the last control point to a red control point by giving it a non-null type.

However, the last control point is actually the cursor, which means that the red control point is always the _next_ placed control point.

The fix is to reference the second-last control point (`[^2]`), and add tests.

smoogipoo:
Resolves https://github.com/ppy/osu/issues/8726


peppy:
@smoogipoo please double-check my commit, just changing the colouring of points to better know that you actually made a new segment during placement.

peppy (approved):

smoogipoo:
looks good

PR #8750, opened 2020-04-14
Fix scoring in classic mode not awarding exact numerical value for judgement

### LINKED ISSUE #8747: Base scores for notes not giving the correct score\
**Describe the bug:** 
Base score for 300’s, 100’s and 50’s on lazer are different from the base scores from osu stable. Lazer scores: 300=12 , 100=4 , 50=2. Stable scores: 300=300 , 100=100 , 50=50
876F1EAB-1044-4EF2-B4BF-9739E219DB81
DA711930-A6C5-4294-8862-9C916FAB56ED
5AE28B5A-F32E-47AA-92EB-2BFF8B045343
E0220FF2-43DB-41A4-9106-BC130128B3B1
5FD7E63D-67E7-448F-B659-8A86747C0E7C
AD31CDE0-8AE4-43B2-AA79-31B0FE9527D7

**osu!lazer version:** 
2020.413.0

bdach:
*Technically* resolves #8747.

# Summary

Upon closer inspection the classic score formula was subtly wrong. The version given in the wiki is:

	Score = Hit Value + (Hit Value * ((Combo multiplier * Difficulty multiplier * Mod multiplier) / 25))

The code previously used:

	bonusScore + baseScore * ((1 + Math.Max(0, HighestCombo.Value - 1) * scoreMultiplier) / 25);

which is not equivalent to the version on the wiki. The error is in the 1 factor, as in the above version it is being divided by 25, while it should be outside the division to keep parity with the previous formula.

Fix by hoisting the 1 outside of the inner parentheses.

The tests attached in the first commit demonstrate that this change causes a single hit without combo to increase total score by its exact numeric value.

# Remarks

I wanted to add more coverage for cases with more than 1 object, but looking at the current implementation compared to the scorev1 spec from wiki I arrived at the conclusion that the current implementation is different enough that in its current state it might not make sense to add tests for it.

The way I understand the wiki description, the combo modifier should be recalculated after and applied to each judgement, as in:
```
(combo start) 1x, 1x, 1.04x, 1.08x, 1.12x, (miss) 1x, 1x, 1.04x, ...
```
but the way it's currently done it looks as if the highest combo over the entire beatmap is taken to calculate a total multiplier of the overall score (so it's effectively applied *after* summing up all judgements). I'm not sure whether this was a conscious choice to simplify implementation, or an oversight, hence I abstained from adding such a test case for now.

If it's deemed something that needs to be fixed I can have a go either here or in a follow-up PR.


peppy:
The idea behind "classic" scoring is it should *always* be in the same order as other scoring methods. It is not for it to match 1:1 (it would make displaying leaderboards impossible without each user seeing a different order).

Can't comment on the change in this PR – would need testing to see how the scores match (the previous one was ballpark on visual comparison between 5 or so maps at various points throughout the map vs the stable score displayed at the same points with perfect combo).

bdach:
I see. This particular "issue" of 300s not being 300s might have been unnoticeable with that methodology, as once combo gets above 50 or so the base factor of the 1 in the new formula will start getting "crowded out" by the combo bonus and they should be comparable with high combos in the end.

This change would mostly boost scores with really short combos. Whether that's a valid change I don't really know or care too much myself, mostly submitted it since it seemingly makes users confused and seemed easy enough to fix.

peppy:
Yeah, seems like it won't make a huge difference and only bring it closer to stable scoring.

bdach:
Can't get to it right now but I can provide some before/after comparisons within, say, the next 24 hours or so.

peppy (approved):
This definitely feels more correct at initial combos, let's go with it for now!

PR #8755, opened 2020-04-14
Fix osu!mania replays actuating incorrect keys when multiple stages are involved

### LINKED ISSUE #8723: 10k+ mania maps fail to load results screen without auto mod
mania maps with 10 or more keys plays fine but the results screen doesn't load, the escape button also stops working. the retry button can be used to get out of this situation as it still works

issue doesn't seem to occur with the auto mod on

can be replicated by finishing a mania map with 10 keys or more

peppy:
- Closes https://github.com/ppy/osu/issues/8723.

Added tests and also manually tested on a 10k (5+5k) beatmap to be working as expected.

### DIFF
--- a/osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
+++ b/osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
@@ -1,8 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
-using System.Linq;
 using osu.Game.Beatmaps;
 using osu.Game.Replays.Legacy;
 using osu.Game.Rulesets.Mania.Beatmaps;
@@ -37,7 +37,7 @@ namespace osu.Game.Rulesets.Mania.Replays
 
             while (activeColumns > 0)
             {
-                var isSpecial = maniaBeatmap.Stages.First().IsSpecialColumn(counter);
+                bool isSpecial = isColumnAtIndexSpecial(maniaBeatmap, counter);
 
                 if ((activeColumns & 1) > 0)
                     Actions.Add(isSpecial ? specialAction : normalAction);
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -58,33 +58,90 @@ public LegacyReplayFrame ToLegacy(IBeatmap beatmap)
...
+
+            throw new InvalidOperationException("Special key index too high");
+        }
+
+        /// <summary>
+        /// Check whether the column at an overall index (across all stages) is a special column.
+        /// </summary>
+        /// <param name="beatmap">The beatmap.</param>
+        /// <param name="index">The overall index to check.</param>
+        /// <returns></returns>
```
smoogipoo, line 130:
```suggestion
```

osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -58,33 +58,90 @@ public LegacyReplayFrame ToLegacy(IBeatmap beatmap)
...
+        {
+            foreach (var stage in beatmap.Stages)
+            {
+                for (int stageIndex = 0; stageIndex < stage.Columns; stageIndex++)
+                {
+                    if (index == 0)
+                        return stage.IsSpecialColumn(stageIndex);
+
+                    index--;
+                }
```
smoogipoo, line 141: Unroll the inner loop.

```suggestion
                if (index >= stage.Columns)
                {
                    index -= stage.Columns;
                    continue;
                }

                return stage.IsSpecialColumn(index);
```

osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -58,33 +58,90 @@ public LegacyReplayFrame ToLegacy(IBeatmap beatmap)
...
+            }
+
+            return false;
```
smoogipoo, line 144:
```suggestion
            throw new ArgumentException("Column index is too high.", nameof(index));
```

osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -58,33 +58,90 @@ public LegacyReplayFrame ToLegacy(IBeatmap beatmap)
...
-                        keys |= 1 << specialColumns[1];
+                        keys |= 1 << getSpecialColumnIndex(maniaBeatmap, 1);
                         break;
 
                     default:
-                        keys |= 1 << (action - ManiaAction.Key1);
+                        // the index in lazer, which doesn't include special keys.
+                        int nonSpecialKeyIndex = action - ManiaAction.Key1;
+
+                        int overallIndex = 0;
```
smoogipoo, line 77:
```suggestion
                        // the index inclusive of special keys.
                        int overallIndex = 0;
```

osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
```diff
@@ -58,33 +58,90 @@ public LegacyReplayFrame ToLegacy(IBeatmap beatmap)
...
+                                // found a non-special column we could use.
+                                if (nonSpecialKeyIndex == 0)
+                                    break;
+
+                                // found a non-special column but not ours.
+                                nonSpecialKeyIndex--;
+                            }
+
+                            overallIndex++;
+                        }
```
smoogipoo, line 93: Personally found this while loop very confusing due to the inverted condition and then the non-inverted following it.

Prefer to be rewritten as:
```suggestion
                        // iterate to find the index including special keys.
                        for (; overallIndex < maniaBeatmap.TotalColumns; overallIndex++)
                        {
                            // skip over special columns.
                            if (isColumnAtIndexSpecial(maniaBeatmap, overallIndex))
                                continue;

                            // found a non-special column to use.
                            if (nonSpecialKeyIndex == 0)
                                break;

                            // found a non-special column but not ours.
                            nonSpecialKeyIndex--;
                        }
```

smoogipoo (changes requested):

osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs
peppy: nice one

smoogipoo (approved):

PR #8756, opened 2020-04-14
Fix 0 custom sample bank not falling back to user skin

### LINKED ISSUE #8528: Hitsounds only in the left speaker specific map

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8528
Supersedes / closes https://github.com/ppy/osu/pull/8649

The superseded PR headed into the wrong direction, which was hidden by the fact that there was 0 testing of gameplay samples. This is now tested.

master:
image

superseded PR:
image


smoogipoo:
Can't replicate the test failures after running with and without debugger attached for ~1hr :(

peppy (approved):
seems same enough..

PR #8759, opened 2020-04-14
Make presented beatmap a recommended one

### LINKED ISSUE #8732: PresentBeatmap should present recommended difficulty
After #8731 and #8444 are merged, I think we should change ``PresentBeatmap`` to present the recommended difficulty.

### LINKED ISSUE #11062: Very first beatmap selected isn't recommended one
recommender

**osu!lazer version:** 
2020.1128.0

LittleEndu:
Closes #8732
Closes #11062 (I assume it's fixed by DifficultyRecommender getting online faster)

There's a lot of small things this PR does:
 * Move ``DifficultyRecommender`` from ``SongSelect`` to ``OsuGameBase``
   * ``SongSelect`` makes use of it via injection
   * This *does not* move it from its current namespace
 * Rewrite ``DifficultyRecommender`` to recommend from all rulesets
   * This used to only recommend from current ruleset
   * Now orders all rulesets, based on which recommend higher star ratings, and tries to recommend something from them in order, while current ruleset is always recommended first
   * This shouldn't change ``CarouselBeatmapSet`` behaviour which already filtered out beatmaps before asking for recommendation 
   * Ordering of rulesets is set to happen only once every ruleset change. Probably could be optimized better :/ (probably want to order them once and just change the order of current ruleset in function scope only. Could make use of OrderedDictionary then)
 * Make ``PresentBeatmap()`` use new ``DifficultyRecommender``
   * First ``difficultyPredicate`` isn't used to find a single beatmap anymore. Instead it's used to filter all beatmaps from the ``BeatmapSetInfo``
   * Those filtered beatmaps are then fed to ``DifficultyRecommender`` and the recommended beatmap is presented
   * If ``DifficultyRecommender`` fails, first try to see if the predicate matched anything and present that, second try to use default predicate (current ruleset) and present that, if all fails just present first beatmap

Even before this PR, ``PresentBeatmap`` by-passed any user selection (``CarouselGroupEagerSelect`` tries to keep track of ``lastSelectedIndex``) and would have always presented the easiest difficulty. This is now swapped with presenting the recommended difficulty.


osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -72,6 +85,26 @@ private void calculateRecommendedDifficulties()
...
 
+        private IEnumerable<RulesetInfo> bestRulesetOrder;
+
+        private IEnumerable<RulesetInfo> getBestRulesetOrder()
+        {
+            if (bestRulesetOrder != null)
+                return bestRulesetOrder;
+
+            var otherRulesets = recommendedStarDifficulty.ToList()
+                                                         .Where(pair => !pair.Key.Equals(ruleset.Value))
```
bdach, line 96: To achieve the effect I assume is desired here, you could actually go a little weird to make this shorter and leverage `OrderBy`:
```csharp
> new[] { false, false, true, false }.OrderByDescending(x => x)
OrderedEnumerable<bool, bool> { true, false, false, false }
```
and then chain with `ThenByDescending` to get the secondary ordering.

osu.Game/OsuGame.cs
```diff
@@ -346,11 +344,24 @@ public void PresentBeatmap(BeatmapSetInfo beatmap, Predicate<BeatmapInfo> diffic
...
+
+                var selection = DifficultyRecommender.GetRecommendedBeatmap(beatmaps);
+
+                // fallback if a difficulty can't be recommended, maybe we are offline
+                if (selection == null)
+                {
+                    if (difficultyCriteria != null)
+                        selection = beatmaps.First();
+                    else
+                        selection = databasedSet.Beatmaps.Find(b => b.Ruleset.Equals(Ruleset.Value)) ?? databasedSet.Beatmaps.First();
```
bdach, line 360: Not sure this discarding `difficultyCriteria` to only check ruleset is correct. Sounds like it might break some previous changes in this area.

bdach (changes requested):
At this level of complexity of the logic in here tests are a must for me.

osu.Game/OsuGame.cs
LittleEndu: the first path of the if statement would use the beatmaps filtered by ``difficultyCriteria``

LittleEndu:
> tests are a must for me.
Everything in here is already tested by previously written tests. New functionality is tied to online as recommendations are based on performance points.

You might be able to overwrite the cached ``DifficultyRecommender`` in tests and make it not rely on online and then test it. Not sure if this is how DI works.

bdach:
We actually have mocking for online now since #8714 which might come in handy. Ideally I'd like to see the ruleset sorting thing tested.

LittleEndu:
I think I can write a test for it then. Seems simple enough, just need ``DummyAPIAccess``'s ``HandleRequest`` to return a dummy user with dummy statistics when ``DifficultyRecommender`` fires a ``GetUserRequest`` request. 

I'll look into it tomorrow.

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -44,16 +46,27 @@ private void load()
...
+            if (!recommendedStarDifficulty.Any())
+                return null;
+
+            BeatmapInfo beatmap = null;
+
+            foreach (var r in getBestRulesetOrder())
             {
-                return beatmaps.OrderBy(b =>
+                if (!recommendedStarDifficulty.TryGetValue(ruleset.Value, out var stars))
+                    break;
```
LittleEndu, line 57: This should probably be ``continue;`` instead or shouldn't be here at all.

osu.Game/OsuGame.cs
```diff
@@ -346,11 +344,24 @@ public void PresentBeatmap(BeatmapSetInfo beatmap, Predicate<BeatmapInfo> diffic
                     return;
                 }
 
-                // Find first beatmap that matches our predicate.
-                var first = databasedSet.Beatmaps.Find(difficultyCriteria) ?? databasedSet.Beatmaps.First();
+                // Find beatmaps that match our predicate.
+                var beatmaps = databasedSet.Beatmaps.Where(b => difficultyCriteria?.Invoke(b) ?? true);
+                if (!beatmaps.Any())
+                    beatmaps = databasedSet.Beatmaps;
```
LittleEndu, line 350: This should write null to ``difficultyCriteria`` as it wasn't able to provide any difficulties and shouldn't be used again. Would fix the comment below.

LittleEndu: made it use difficulty criteria only once so this shouldn't be needed.

LittleEndu:
Needs tests for when current ruleset doesn't have difficulties that the next ruleset recommended is the one player has higher recommended star difficulty in

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -90,19 +88,22 @@ private void calculateRecommendedDifficulties()
...
-                                                         .Select(pair => pair.Key)
-                                                         .Reverse();
+            bestRulesetOrder = recommendedStarDifficulty.ToList()
+                                                        .OrderBy(pair => pair.Value)
+                                                        .Select(pair => pair.Key)
+                                                        .Reverse();
 
-            var rulesetList = new List<RulesetInfo>(new[] { ruleset.Value });
-            rulesetList.AddRange(otherRulesets);
+            return moveCurrentRulesetToFirst();
```
LittleEndu, line 98: Both return statements return the same thing, can make if statement into ``bestRulesetOrder == null``

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -32,8 +32,6 @@ public class DifficultyRecommender : Component, IOnlineComponent
         private void load()
         {
             api.Register(this);
-
-            ruleset.ValueChanged += _ => bestRulesetOrder = null;
```
LittleEndu: ``bestRulesetOrder`` should instead be reset when new values are written to ``recommendedStarDifficulty`` when the request succeeds.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+            var taikoRuleset = rulesets.AvailableRulesets.First(r => r.ID == 1);
+            var catchRuleset = rulesets.AvailableRulesets.First(r => r.ID == 2);
+
+            var osuImport = importBeatmap(0, new List<RulesetInfo> { osuRuleset });
+            var mixedImport = importBeatmap(1, new List<RulesetInfo> { taikoRuleset, catchRuleset, taikoRuleset });
+
+            // Make sure we are on standard ruleset
+            presentAndConfirm(osuImport, 0, 1);
+
+            // Present mixed difficulty set, expect ruleset with highest star difficulty
```
LittleEndu, line 113: Should say second highest as the highest is missing in this test

osu.Game/OsuGameBase.cs
```diff
@@ -99,6 +100,9 @@ public class OsuGameBase : Framework.Game, ICanAcceptFiles
 
         public bool IsDeployedBuild => AssemblyVersion.Major > 0;
 
+        [Cached]
+        protected readonly DifficultyRecommender DifficultyRecommender = new DifficultyRecommender();
```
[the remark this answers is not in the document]
LittleEndu: I put the recommender in base game so that only one instance is active. That's to lower API calls, I don't know how expensive calling ``GetUserRequest`` 4 times is but I can't imagine it being cheap.

bdach:
Is this at a reviewable state right now? Hard to tell with the changes coming in.

LittleEndu:
It should be. Can't think of what else would need testing. Would like to figure out why testing in visual browser doesn't work, but I have no idea what broke it.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -104,14 +102,13 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
         private MusicController music { get; set; }
 
         [BackgroundDependencyLoader(true)]
-        private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, SkinManager skins, ScoreManager scores)
+        private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, SkinManager skins, ScoreManager scores, DifficultyRecommender recommender)
```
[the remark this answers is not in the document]
smoogipoo: Should be fine so long as a reference isn't being held out-in (i.e. object with longer lifetime holding reference to object with a shorter lifetime).

As far as I can see that's not the case here.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+                        case GetUserRequest userRequest:
+
+                            decimal pp = userRequest.Ruleset.ID switch
+                            {
+                                0 => 336, // Expected recommended star difficulty 2*
+                                1 => 928, // Expected recommended star difficulty 3*
+                                2 => 1905, // Expected recommended star difficulty 4*
+                                3 => 3329, // Expected recommended star difficulty 5*
+                                _ => 0
+                            };
```
bdach, line 47: We don't generally use switch expressions anywhere, prefer standard ones.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+                            {
+                                Username = @"Dummy",
+                                Id = 1001,
+                                Statistics = new UserStatistics
+                                {
+                                    PP = pp
+                                }
+                            });
+                            break;
+                    }
```
bdach, line 59: Way way *way* nested. Could factor out a method or two here.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+        }
+
+        [Test]
+        public void TestPresentedBeatmapIsRecommended()
+        {
+            var importFunctions = new List<Func<BeatmapSetInfo>>();
+
+            for (int i = 0; i < 5; i++)
+            {
+                importFunctions.Add(importBeatmap(i, new List<RulesetInfo> { null, null, null, null, null }));
```
bdach, line 73: `Enumerable.Repeat` exists, and even then, this is quite strange what with passing `null` five times for it to only fallback to the first available ruleset at `importBeatmap`.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+            for (int i = 0; i < 5; i++)
+            {
+                presentAndConfirm(importFunctions[i], i, 2);
+            }
+        }
+
+        [Test]
+        public void TestBestRulesetIsRecommended()
+        {
+            var osuRuleset = rulesets.AvailableRulesets.First(r => r.ID == 0);
```
bdach, line 85: Generally these should be `Single`s, but you can actually simply do
```suggestion
            var osuRuleset = rulesets.GetRuleset(0);
```
(not going to point out every single instance).

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+                    Beatmaps = beatmaps,
+                }).Result;
+            });
+
+            AddAssert($"import {importID} succeeded", () => imported != null);
+
+            return () => imported;
+        }
+
+        private void presentAndConfirm(Func<BeatmapSetInfo> getImport, int importedID, int expextedDiff)
```
bdach, line 160:
```suggestion
        private void presentAndConfirm(Func<BeatmapSetInfo> getImport, int importedID, int expectedDiff)
```
(+ usages)

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
```diff
@@ -0,0 +1,168 @@
...
+            var mixedImport = importBeatmap(1, new List<RulesetInfo> { taikoRuleset, catchRuleset, taikoRuleset });
+
+            // Make sure we are on standard ruleset
+            presentAndConfirm(osuImport, 0, 1);
+
+            // Present mixed difficulty set, expect ruleset with highest star difficulty
+            presentAndConfirm(mixedImport, 1, 2);
+        }
+
+        private Func<BeatmapSetInfo> importBeatmap(int importID, List<RulesetInfo> rulesets)
```
bdach, line 117: Not sure I like the way this is written too much. A little complicated to parse but I'm not sure if it's just me.

Would probably pre-construct the beatmap info in the first step of each test, store them to a variable/field and then operate on the created info through `Func`s in the asserts (so pass the info instead of passing magic indices to `presentAndConfirm`).

bdach (changes requested):
What do you mean when you say that tests don't work in the visual browser? The tests are headless, so they're not actually even visible in the browser, but I'm not sure if it's the cause or the bypass fix.

Skimmed through the code mostly as there are significant style issues to address first.

LittleEndu:
I marked the test as headless somewhere in the middle of this PR as I thought it was working but something changed and it no longer works in visual browser. So yeah, it being marked as headless is a fix not the cause.

bdach:
So I spent an hour or two on this code to see if I can bring it to a mergeable state. I did fix up the test code, which was very haphazard in many places and I'm actually honestly surprised it ever passed CI, as in my runs through rider they would reliably fail every time due to prematurely accessing rulesets. In addition to that the registration of the API handling function needed to happen *before* instantiation of `OsuGame`.

Another issue I found is that the current code of `DifficultyRecommender` would attempt to fire off requests to fetch user profile data for custom rulesets, which I also fixed up. I also simplified one sort, but then I found an issue major enough as to warrant stopping further work on this.

The current code does not work well with the decoupled ruleset mechanism. Here's a simple reproduction scenario:

* Switch ruleset to taiko
* Open direct
* Present an osu! map
* The osu! map will be presented, but the difficulty selected will have star rating as if it were a taiko map

Because the decoupled ruleset gets transferred *after* the present happens, the recommender will try to select a difficulty too early.

As this is a major sticking point (to the point of threatening this PR's validity) I will not be continuing to work on this code further until this is resolved.

bdach (changes requested):
As above.

LittleEndu:
I was confused when you said "the recommender will try to select a difficulty too early" as recommender doesn't even deal with making a beatmap selection and shouldn't care about what ruleset is current. Turned out there was what I assume to be a copy paste oversight and the recommender was indeed using current ruleset. 

I assume you meant "selecting from its choices" and not "marking the beatmap as selected" as that's what song select and carousel do.

osu.Game/OsuGame.cs
```diff
@@ -316,14 +316,12 @@ private void load()
         /// </summary>
         /// <param name="beatmap">The beatmap to select.</param>
         /// <param name="difficultyCriteria">
-        /// Optional predicate used to try and find a difficulty to select.
-        /// If omitted, this will try to present the first beatmap from the current ruleset.
+        /// Optional predicate used to filter which difficulties to select.
+        /// If omitted, this will try to present a recommended beatmap from the current ruleset.
         /// In case of failure the first difficulty of the set will be presented, ignoring the predicate.
         /// </param>
```
bdach, line 322: I'd reword this to something like
```csharp
        /// <param name="difficultyCriteria">Optional predicate used to narrow the set of difficulties to select from when presenting.</param>
        /// <remarks>
        /// Among items satisfying the predicate, the order of preference is:
        /// <list type="bullet">
        /// <item>beatmap with recommended difficulty, as provided by <see cref="DifficultyRecommender"/>,</item>
        /// <item>first beatmap from the current ruleset,</item>
        /// <item>first beatmap from any ruleset.</item>
        /// </list>
        /// </remarks>
```

osu.Game/OsuGame.cs
```diff
@@ -341,16 +339,28 @@ public void PresentBeatmap(BeatmapSetInfo beatmap, Predicate<BeatmapInfo> diffic
...
+                if (!beatmaps.Any())
+                    beatmaps = databasedSet.Beatmaps;
+
+                // Try to select recommended beatmap
+                // This should give us a beatmap from current ruleset if there are any in our matched beatmaps
+                var selection = DifficultyRecommender.GetRecommendedBeatmap(beatmaps) ?? (
+                    // Fallback if a difficulty can't be recommended, maybe we are offline
+                    // First try to find a beatmap in current ruleset, otherwise use first beatmap
+                    beatmaps.FirstOrDefault(b => b.Ruleset.Equals(Ruleset.Value)) ?? beatmaps.First()
+                );
```
bdach, line 360: This reads super awkwardly. I'd split into three lines using `??=`.

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -72,6 +84,36 @@ private void calculateRecommendedDifficulties()
             });
         }
 
+        private IEnumerable<RulesetInfo> bestRulesetOrder;
+
+        private IEnumerable<RulesetInfo> getBestRulesetOrder()
+        {
+            bestRulesetOrder ??= recommendedStarDifficulty.OrderByDescending(pair => pair.Value)
+                                                          .Select(pair => pair.Key)
+                                                          .ToList();
```
bdach, line 93: I'm not sure this aggressive caching is necessary, it's sorting a collection of at most 4 items. This isn't going to kill performance, and just ordering every time would get rid of this field.

osu.Game/OsuGame.cs
LittleEndu: This used to be split into multiple lines but that triggered CI errors. I'll see if using ``??=`` fixes that.

osu.Game/OsuGameBase.cs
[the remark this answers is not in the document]
LittleEndu: I'm thinking about again working on this PR over the weekend. I get all the other comments but this one I still would like clarified.

> resolve at necessary call sites 

I think in this PR the test, ``SongSelect``, and ``OsuGame`` are the only call sites. But the fact that ``OsuGame.PresentBeatmap`` needs access to recommender to present recommended beatmaps means that ``DifficultyRecommender`` needs to be accessed at base level. To me it made sense to cache it at base level too and you get reduction to API calls as a bonus.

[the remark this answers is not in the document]
bdach: The long and short of it is that `OsuGame` and `OsuGameBase` should be as slim as they could possibly hope to be. If this can't be placed anywhere else without significant overhead, so be it, but those classes *need* to remain manageable.

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapRecommendations.cs
LittleEndu: Seems like you resolved this yourself in 2913a8183538f823858aace50834f9e38a74b726 but I think that code was copied from [TestScenePresentBeatmap](https://github.com/ppy/osu/blob/93269606cd78c1193d561ba649f254afa1255c27/osu.Game.Tests/Visual/Navigation/TestScenePresentBeatmap.cs#L88-L129). Maybe something over there needs to be changed too

bdach: separate PR even if so.

osu.Game/OsuGameBase.cs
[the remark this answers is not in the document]
LittleEndu: This also helps ``DifficultyRecommender`` get online faster to start making recommendations from the start, fix for #11062

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -47,21 +47,35 @@ private void load()
...
 
-        private void calculateRecommendedDifficulties()
+        private void calculateRecommendedDifficulties(bool onlyIfNoPreviousValues = false)
         {
-            rulesets.AvailableRulesets.ForEach(rulesetInfo =>
+            if (recommendedStarDifficulty.Any() && onlyIfNoPreviousValues)
+                return;
+
+            // only query API for built-in rulesets
+            rulesets.AvailableRulesets.Where(ruleset => ruleset.ID <= 3).ForEach(rulesetInfo =>
```
peppy, line 78: Add this as a `const` inside of `ILegacyRuleset`

```csharp
        const int MAX_LEGACY_RULESET_ID = 3;
```

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -75,12 +89,35 @@ private void calculateRecommendedDifficulties()
             });
         }
 
+        private IEnumerable<RulesetInfo> getBestRulesetOrder()
```
peppy, line 92: what does "best" mean here?

osu.Game/Screens/Select/DifficultyRecommender.cs
```diff
@@ -75,12 +89,35 @@ private void calculateRecommendedDifficulties()
...
+            return orderedRulesets;
+        }
+
         private void onlineStateChanged(ValueChangedEvent<APIState> state) => Schedule(() =>
         {
             switch (state.NewValue)
             {
                 case APIState.Online:
-                    calculateRecommendedDifficulties();
+                    calculateRecommendedDifficulties(true);
```
peppy, line 120: probably want to clear any values on going offline

peppy (changes requested):

osu.Game/Screens/Select/DifficultyRecommender.cs
LittleEndu: Not sure there are any to clear. After we have been online once we never need to change anything again.

peppy: What if the user logs in as a different user?

osu.Game/Screens/Select/DifficultyRecommender.cs
LittleEndu: it means "best guess at what user wants to play". I'll add xmldoc to the function, but I can't come up with a better name.

osu.Game/Screens/Select/DifficultyRecommender.cs
LittleEndu: User IDs are now checked, but still no action needs to be taken if the user logs out, unless we don't want to continue recommending the last logged in user's star rating. I think if somebody wants to play offline, they would still expect to get correct recommendations.

bdach: Honestly it's debatable that it shouldn't be cleared. I can see an argument for clearing, too (say, having someone new play). We obviously won't know for sure which one it is, though, so I'm not sure I care much either way. This entire feature is almost too subtle to notice the existence of anyway.

LittleEndu: Also wanted to note, lest I multi account, I don't think I can test this.

> too subtle to notice the existence of anyway

In song select it is very noticeable. in presentBeatmap's case, I can agree.

bdach: I've never consciously been irked by (or even noted) the extra two or three arrow presses that it takes to select a difficulty in song select I want. Different strokes for different folks I suppose.

osu.Game/OsuGame.cs
```diff
@@ -361,16 +363,29 @@ public void PresentBeatmap(BeatmapSetInfo beatmap, Predicate<BeatmapInfo> diffic
...
-                if (Beatmap.Value.BeatmapSetInfo.Hash == databasedSet.Hash && difficultyCriteria(Beatmap.Value.BeatmapInfo))
+                if (Beatmap.Value.BeatmapSetInfo.Hash == databasedSet.Hash && (difficultyCriteria?.Invoke(Beatmap.Value.BeatmapInfo) ?? true))
                 {
                     return;
                 }
 
-                // Find first beatmap that matches our predicate.
-                var first = databasedSet.Beatmaps.Find(difficultyCriteria) ?? databasedSet.Beatmaps.First();
+                // Find beatmaps that match our predicate.
+                var beatmaps = databasedSet.Beatmaps.Where(b => difficultyCriteria?.Invoke(b) ?? true);
```
peppy, line 372: multiple enumeration

peppy (dismissed):

peppy (approved):

peppy:
@bdach would appreciate a quick sanity pass on this. i've made quite a few changes but am pretty confident.

bdach (approved):
seems reasonable

PR #8762, opened 2020-04-15
Add automatic pagination to BeatmapListingOverlay

aemino:
This PR adds automatic pagination to BeatmapListingOverlay and supersedes #7407. The diffs are mostly changes that have been ported from the original DirectOverlay version. However, there are some differences, including the fact that this version attempts to use the `cursor` mechanism rather than a page number for pagination.

This is somewhat of a work-in-progress; please see below for some comments on things I had questions/concerns about.

- [x] Depends on ppy/osu-framework#3302

### DIFF
--- a/osu.Game/Online/API/Requests/ResponseWithCursor.cs
+++ b/osu.Game/Online/API/Requests/ResponseWithCursor.cs
@@ -11,6 +11,15 @@ namespace osu.Game.Online.API.Requests
         /// A collection of parameters which should be passed to the search endpoint to fetch the next page.
         /// </summary>
         [JsonProperty("cursor")]
-        public object CursorJson;
+        public dynamic Cursor;
+    }
+
+    public abstract class ResponseWithCursor<T> : ResponseWithCursor where T : class
+    {
+        /// <summary>
+        /// A collection of parameters which should be passed to the search endpoint to fetch the next page.
+        /// </summary>
+        [JsonProperty("cursor")]
+        public new T Cursor;
     }
 }
--- a/osu.Game/Online/API/Requests/ResponseWithCursor.cs
+++ b/osu.Game/Online/API/Requests/ResponseWithCursor.cs
@@ -11,6 +11,6 @@ namespace osu.Game.Online.API.Requests
         /// A collection of parameters which should be passed to the search endpoint to fetch the next page.
         /// </summary>
         [JsonProperty("cursor")]
-        public dynamic CursorJson;
+        public object CursorJson;
     }
 }
--- a/osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
+++ b/osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.ComponentModel;
+using Newtonsoft.Json;
 using osu.Framework.IO.Network;
 using osu.Game.Overlays;
 using osu.Game.Overlays.Direct;
@@ -11,15 +12,24 @@ namespace osu.Game.Online.API.Requests
 {
     public class SearchBeatmapSetsRequest : APIRequest<SearchBeatmapSetsResponse>
     {
+        public class Cursor
+        {
+            [JsonProperty("_id")]
+            public string Id;
+
+            [JsonProperty("approved_date")]
+            public string ApprovedDate;
+        }
+
         private readonly string query;
         private readonly RulesetInfo ruleset;
-        private readonly object cursor;
+        private readonly Cursor cursor;
         private readonly BeatmapSearchCategory searchCategory;
         private readonly DirectSortCriteria sortCriteria;
         private readonly SortDirection direction;
         private string directionString => direction == SortDirection.Descending ? @"desc" : @"asc";
 
-        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, object cursor = null, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
+        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, Cursor cursor = null, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
         {
             this.query = string.IsNullOrEmpty(query) ? string.Empty : System.Uri.EscapeDataString(query);
             this.ruleset = ruleset;
@@ -38,7 +48,10 @@ namespace osu.Game.Online.API.Requests
                 req.AddParameter("m", ruleset.ID.Value.ToString());
 
             if (cursor != null)
-                req.AddParameter("cursor", cursor.ToString());
+            {
+                req.AddParameter("cursor[_id]", cursor.Id);
+                req.AddParameter("cursor[approved_date]", cursor.ApprovedDate);
+            }
 
             req.AddParameter("s", searchCategory.ToString().ToLowerInvariant());
             req.AddParameter("sort", $"{sortCriteria.ToString().ToLowerInvariant()}_{directionString}");
--- a/osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
+++ b/osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
@@ -13,15 +13,17 @@ namespace osu.Game.Online.API.Requests
     {
         private readonly string query;
         private readonly RulesetInfo ruleset;
+        private readonly object cursor;
         private readonly BeatmapSearchCategory searchCategory;
         private readonly DirectSortCriteria sortCriteria;
         private readonly SortDirection direction;
         private string directionString => direction == SortDirection.Descending ? @"desc" : @"asc";
 
-        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
+        public SearchBeatmapSetsRequest(string query, RulesetInfo ruleset, object cursor = null, BeatmapSearchCategory searchCategory = BeatmapSearchCategory.Any, DirectSortCriteria sortCriteria = DirectSortCriteria.Ranked, SortDirection direction = SortDirection.Descending)
         {
             this.query = string.IsNullOrEmpty(query) ? string.Empty : System.Uri.EscapeDataString(query);
             this.ruleset = ruleset;
+            this.cursor = cursor;
             this.searchCategory = searchCategory;
             this.sortCriteria = sortCriteria;
             this.direction = direction;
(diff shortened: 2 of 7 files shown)


osu.Game/Online/API/Requests/ResponseWithCursor.cs
```diff
@@ -11,6 +11,6 @@ public abstract class ResponseWithCursor
         /// A collection of parameters which should be passed to the search endpoint to fetch the next page.
         /// </summary>
         [JsonProperty("cursor")]
-        public dynamic CursorJson;
+        public object CursorJson;
```
aemino, line 14: I had to change this from `dynamic` -> `object` to get this to compile; working with `dynamic` directly in the codebase caused issues. However, this doesn't seem like a great solution to me; maybe `ResponseWithCursor` should be generic w.r.t. the cursor type?

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -35,6 +37,9 @@ protected override WebRequest CreateWebRequest()
             if (ruleset.ID.HasValue)
                 req.AddParameter("m", ruleset.ID.Value.ToString());
 
+            if (cursor != null)
+                req.AddParameter("cursor", cursor.ToString());
```
aemino, line 41: I doubt that this is the right way to actually add the cursor to the request. The result is that the pager just receives the same result every time. What's the proper way to do this?

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -260,10 +314,29 @@ private void addContentToPlaceholder(Drawable content)
             currentContent.FadeIn(200, Easing.OutQuint);
         }
 
+        private void addPanelsToContent(IEnumerable<DirectPanel> panels)
+        {
+            // TODO: Fade in?
```
aemino, line 319: As I wrote here, perhaps each panel should have a fade-in animation applied here?

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -260,10 +314,29 @@ private void addContentToPlaceholder(Drawable content)
...
+
+            if (currentContent == null)
+                return;
+
+            ((FillFlowContainer<DirectPanel>)currentContent).AddRange(panels);
```
aemino, line 324: I don't really like how this requires casting. Is there a reason that the container is referred to as a `Drawable` rather than specifically as a `FillFlowContainer`? This seemed like some kind of future-proofing thing to me.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
bdach: The problem here is that web actually expects _two_ parameters, ie. in the form of
```
GET /search?cursor[approved_date]=1586609019000&cursor[_id]=1112051
```
which requires lazer to be able to reinterpret that JSON structure somehow to split out the fields and send them back to web in the format required.

osu.Game/Online/API/Requests/ResponseWithCursor.cs
bdach: I would agree with this due to reasons outlined in the other comment. Maybe worth considering making the generic explicitly opt-in to avoid having to update previous usages (as in having both `ResponseWithCursor` and `ResponseWithCursor<T>`).  I'd also probably argue that cursor classes should be inner classes of the requests they correspond to themselves, unless deemed reusable (in the structural *and* semantic sense)

Ideally we wouldn't have to interpret this at all but alas...

osu.Game/Overlays/BeatmapListingOverlay.cs
bdach: The reason is this line, which makes it possible for `currentContent` to be a `NotFoundOverlay` (displayed when there are no search results):

https://github.com/ppy/osu/blob/9da76e3f3228aa6897d3b3aea31765ecce2724d5/osu.Game/Overlays/BeatmapListingOverlay.cs#L270

I wouldn't exactly call that ideal, though. They should probably not be clumped up in one field like that.

bdach (commented):
Inspections in CI need attention.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
aemino: Are `approved_date` and `_id` always the only two cursor parameters, or does it vary based on the sort...?

aemino:
As far as I'm aware, the CI build is currently failing due to using an outdated version of osu!framework.

peppy:
Yep, that's fine. We'll do a build today or tomorrow.

osu.Game/Overlays/BeatmapListingOverlay.cs
aemino: Do you have a recommendation for separating these? I don't have quite enough experience with the codebase to know what an idiomatic way to handle this would be.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
bdach: [Seems like they aren't](https://github.com/ppy/osu-web/blob/4687a1e8a5c132af3ccfc63a2b90d239ec395d71/app/Libraries/Search/BeatmapsetSearchRequestParams.php#L153-L212), which makes me think a more robust solution is needed (maybe a way to automatically unwrap all fields of `cursor` into the naming convention web expects?)

aemino: Perhaps the cursor can be represented using a `Dictionary<string, string>`?

bdach: I mean, that seems like it would *work*, but it assumes there are no further nested objects (which I can't guarantee right now) and also isn't the prettiest solution. In my mind the absolute cleanest way to handle this would be to not handle it and leave it up to web.

In general I'd actually review all existing requests that use cursors - maybe they've been broken this whole time but nobody actually noticed, since it seems like a long shot this popped up only now.

bdach: Oh, and if this re-mapping required currently does stay lazer-side, ideally there should be a method like
```csharp
req.AddCursor(cursor);
```
that does the mapping automagically internally.

aemino: I think that regardless of how it's done, some level of parsing will be required due to the way the parameters have to be added. Do you have any suggestions for how you would like to see this handled? On the topic of other requests that use cursors, I think that pagination via cursors simply isn't implemented anywhere else, so nothing is broken, just unimplemented.

bdach: If web accepted the same JSON object it sent out, it would require no interpretation on the client's side, which seems better to me (there's no reason for clients to parse the state of the cursor, it's just returned by API only to prevent the necessity of API having to keep track of client state for paging), which is why I'm partially reluctant of suggesting any particular way of handling this client-side.

Then again, putting a JSON blob in a query string is not exactly pretty, either; can't put it in a `GET` body either because that breaks HTTP spec.

I'd actually like the team's opinion on this as I see no one best path forward.

osu.Game/Overlays/BeatmapListingOverlay.cs
bdach: I'd probably go for something like

```csharp
class BeatmapListingResultDisplay
{
    private NotFoundDrawable noResults; // probably needs renaming
    private FillFlowContainer<DirectPanel> resultPanels;

    public BindableList<BeatmapSetInfo> SearchResults { get; private set; }

    // subscribe to SearchResults.CollectionChanged
    // if eventArgs.Action == Add, async load panels and add to fill flow
    // if eventArgs.Action == Reset, display the no results drawable
    // only mutate SearchResults when a request has actually arrived
    // (to avoid mistakenly showing the no results drawable)
}
```

Has the upside of actually being testable without offline.

aemino:
All right, I've *attempted* to whip up a component for the results display. I wasn't able to use BindableList like you suggested because I felt like the events that I could subscribe to wouldn't work well with the component's state. That said, I again feel that my attempt at this component isn't idiomatic, so feedback is appreciated. There is currently a bug where after displaying the "no results found" graphic, it will be unable to properly transition back to the DirectPanel result display. Not sure what's causing that; may very well be something silly that I missed.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -11,17 +12,28 @@ namespace osu.Game.Online.API.Requests
...
     public class SearchBeatmapSetsRequest : APIRequest<SearchBeatmapSetsResponse>
     {
+        public class Cursor
+        {
+            [JsonProperty("_id")]
+            public string Id;
+
+            [JsonProperty("approved_date")]
+            public string ApprovedDate;
+        }
```
bdach, line 22: As we were discussing earlier this won't work if sort criteria are changed - web will return different fields.

osu.Game/Overlays/BeatmapListing/BeatmapListingResultsDisplay.cs
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        public void ReplaceBeatmaps(List<BeatmapSetInfo> beatmaps)
+        {
+            clearBeatmapPanels();
+
+            if (beatmaps.Count > 0)
+            {
+                if (noResultsDrawable.IsAlive)
+                    noResultsDrawable.FadeOut(100, Easing.OutQuint).Then().Schedule(() => RemoveInternal(noResultsDrawable));
```
bdach, line 50: I see no reason to keep re-instantiating the no results drawable as it's static content. Can be just faded in and out whenever the need is (would probably be less likely to bug out, too).

osu.Game/Overlays/BeatmapListing/BeatmapListingResultsDisplay.cs
```diff
@@ -0,0 +1,141 @@
...
+
+                if (!resultPanels.IsAlive)
+                    AddInternal(resultPanels);
```
bdach, line 53: The fill flow also probably doesn't need to be removed & reinstantiated every time, just clearing would probably do fine.

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -195,119 +215,58 @@ private void updateSearch()
...
-                Margin = new MarginPadding { Vertical = 15 },
-                ChildrenEnumerable = beatmaps.Select<BeatmapSetInfo, DirectPanel>(b => new DirectGridPanel(b)
+                if (beatmapSetPager.IsPastFirstPage)
                 {
-                    Anchor = Anchor.TopCentre,
-                    Origin = Anchor.TopCentre,
-                })
-            };
+                    resultsDisplay.AddBeatmaps(beatmaps);
+                }
```
bdach, line 245: Kind of dislike this existing. As I said, if the pager took in a bindable list or was in control of the set of beatmaps, it could clear it itself and with proper event subscriptions this logic becomes cleaner and there's no need to expose `IsPastFirstPage` anymore. I haven't tried it though, if you have and it didn't work then fair enough I suppose.

osu.Game/Overlays/BeatmapListing/BeatmapSetPager.cs
```diff
@@ -0,0 +1,97 @@
...
+        private readonly SortDirection sortDirection;
+
+        private SearchBeatmapSetsRequest getSetsRequest;
+
+        private SearchBeatmapSetsResponse lastResponse;
+
+        public int? TotalSets => lastResponse?.Total;
+        public bool IsPastFirstPage { get; private set; } = false;
+        public bool IsLastPageFetched { get; private set; } = false;
+        public bool IsFetching => getSetsRequest != null;
```
bdach, line 35: shouldn't be `public`

bdach (changes requested):
Results of a cursory read-through.

bdach:
Thanks for the contribution - unfortunately seeing that the PR referenced above that aimed to supersede this one has now been merged, I'm going to go ahead and close this.

aemino:
Sounds good! Kudos to @namaenonaimumei for picking up and rewriting this implementation.

PR #8764, opened 2020-04-15
Disable undo/redo menu items when they can't be performed

smoogipoo:


osu.Game/Graphics/UserInterface/OsuMenuItem.cs
```diff
@@ -19,6 +22,9 @@ public OsuMenuItem(string text, MenuItemType type, Action action)
             : base(text, action)
         {
             Type = type;
+
+            Enabled.BindValueChanged(enabled => Action.Disabled = !enabled.NewValue);
+            Action.BindDisabledChanged(disabled => Enabled.Value = !disabled);
```
bdach, line 27: This gave me pause for a second. It's *correct*, but the fact of having both `Disabled` and `Enabled` made me re-read and make sure it's mapped correct. Would have been nice to have consistency (so `Disabled` everywhere or `Enabled` everywhere) but from what I see `ClickableContainer` *et al.* already have `Enabled` so that ship has probably already sailed.

osu.Game.Tests/Visual/UserInterface/TestSceneOsuMenu.cs
```diff
@@ -0,0 +1,91 @@
...
+    public class TestSceneOsuMenu : OsuManualInputManagerTestScene
+    {
+        public override IReadOnlyList<Type> RequiredTypes => new[]
+        {
+            typeof(OsuMenu),
+            typeof(DrawableOsuMenuItem)
+        };
+
+        private OsuMenu menu;
+        private bool actionPeformed;
```
bdach, line 24: insanely pedantic but just in case
```suggestion
        private bool actionPerformed;
```
(+ usages)

bdach (commented):
For whatever my opinion's worth this looks good to me save for a few annoyances.

osu.Game/Graphics/UserInterface/OsuMenuItem.cs
peppy: I find the outwards propagation a bit weird here. The action changing enabled state could potentially update the EditorChangeHandler's bindable. Maybe this should be a one-way thing instead? ie. the second line shouldn't exist (or at very least shouldn't propagate outwards).

smoogipoo: I added the bindable for the sole purpose of making it simpler to bind the event. I agree that it lead to unsafe scenarios so I've removed it since `Action.Disabled` can be used to promote one-way disabling.

peppy (approved):

Issue #8543, opened 2020-04-01
osu!mania holds / osu!taiko drum rolls do not stretch textures correctly


enoslayd:
Describe the bug:  long hold note appear dimmed near the tail note. beat map is https://osu.ppy.sh/beatmapsets/999735#mania/2091019 appears on 0:01 to 0:04

Screenshots or videos showing encountered issue:
![osu_2020-04-01_20-02-12](https://user-images.githubusercontent.com/41866328/78135322-490ec100-7454-11ea-899c-c5916fb35890.png)

osu!lazer version:  master commit 6d88ac2

Logs: not sure if needed but here (as im writing his issue, osu is still running)
[performance-gamethread.input.log](https://github.com/ppy/osu/files/4414955/performance-gamethread.input.log)
[performance-gamethread.update.log](https://github.com/ppy/osu/files/4414957/performance-gamethread.update.log)

[updater.log](https://github.com/ppy/osu/files/4414959/updater.log)
[database.log](https://github.com/ppy/osu/files/4414960/database.log)
[network.log](https://github.com/ppy/osu/files/4414961/network.log)
[performance.log](https://github.com/ppy/osu/files/4414962/performance.log)
[performance-gamethread.audio.log](https://github.com/ppy/osu/files/4414963/performance-gamethread.audio.log)
[performance-gamethread.draw.log](https://github.com/ppy/osu/files/4414964/performance-gamethread.draw.log)
[runtime.log](https://github.com/ppy/osu/files/4414969/runtime.log)

smoogipoo:
Please provide your skin.

enoslayd:
both are the same skin just have different colors

[ok3.zip](https://github.com/ppy/osu/files/4414999/ok3.zip)
[ok.zip](https://github.com/ppy/osu/files/4415012/ok.zip)

also dimming appears on start of the hold note on classic skin
beatmap when converted to mania https://osu.ppy.sh/beatmapsets/140662#osu/351190
time stamp: 3:57
![osu_2020-04-01_20-25-10](https://user-images.githubusercontent.com/41866328/78136867-e79c2180-7456-11ea-9f78-a4c79d014362.png)

enoslayd:
I also noticed this problem in taiko drumrolls

peppy:
Yes, this is known. It's an issue with the blending being applied to stretched textures (at an atlas level). Will likely be fixed with shader changes in the future.

enoslayd:
Closing since already resolved

PR #8766, opened 2020-04-15
Add taiko drumroll skinning support

peppy:
Still needs metric elements added, but mostly works.

- Adds remaining skinnable hitobjects (without implementations) to allow for using `SkinnableDrawable` across `TaikoHitObject` classes.
- Tidies up sizing logic (aka removes a heap that we don't need)
- Stretched piece gets gradiented to transparent at each end. Will fix this in a follow-up PR (also happens in mania, see https://github.com/ppy/osu/issues/8543)
- [x] Figure why LegacyCirclePiece is not visible after rewind (related to proxying?)

### DIFF
--- a/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRollTick.cs
+++ b/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRollTick.cs
@@ -21,7 +21,10 @@ namespace osu.Game.Rulesets.Taiko.Objects.Drawables
         public override bool DisplayResult => false;
 
         protected override SkinnableDrawable CreateMainPiece() => new SkinnableDrawable(new TaikoSkinComponent(TaikoSkinComponents.DrumRollTick),
-            _ => new TickPiece());
+            _ => new TickPiece
+            {
+                Filled = HitObject.FirstTick
+            });
 
         protected override void CheckForResult(bool userTriggered, double timeOffset)
         {
--- a/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRollTick.cs
+++ b/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRollTick.cs
@@ -3,10 +3,10 @@
 
 using System;
 using osu.Framework.Graphics;
-using osu.Framework.Graphics.Containers;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Rulesets.Taiko.Objects.Drawables.Pieces;
+using osu.Game.Skinning;
 
 namespace osu.Game.Rulesets.Taiko.Objects.Drawables
 {
@@ -20,10 +20,8 @@ namespace osu.Game.Rulesets.Taiko.Objects.Drawables
 
         public override bool DisplayResult => false;
 
-        protected override CompositeDrawable CreateMainPiece() => new TickPiece
-        {
-            Filled = HitObject.FirstTick
-        };
+        protected override SkinnableDrawable CreateMainPiece() => new SkinnableDrawable(new TaikoSkinComponent(TaikoSkinComponents.DrumRollTick),
+            _ => new TickPiece());
 
         protected override void CheckForResult(bool userTriggered, double timeOffset)
         {
--- a/osu.Game.Rulesets.Taiko/Objects/Drawables/Pieces/ElongatedCirclePiece.cs
+++ b/osu.Game.Rulesets.Taiko/Objects/Drawables/Pieces/ElongatedCirclePiece.cs
@@ -1,7 +1,9 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
+using osu.Game.Graphics;
 
 namespace osu.Game.Rulesets.Taiko.Objects.Drawables.Pieces
 {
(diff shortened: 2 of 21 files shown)


osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRollTick.cs
```diff
@@ -20,10 +20,8 @@ public DrawableDrumRollTick(DrumRollTick tick)
 
         public override bool DisplayResult => false;
 
-        protected override CompositeDrawable CreateMainPiece() => new TickPiece
-        {
-            Filled = HitObject.FirstTick
```
smoogipoo, line 25: Removed?

smoogipoo (changes requested):

bdach:
On a quick-and-dirty check the circle piece is indeed visible after rewind if it's proxied upon hit.

On another note, from what I can see the current implementation has `LegacyCirclePiece` take the width of 1, which brings back #6889 in the case of legacy skins, as the "fix" I applied once (calculating offset of drawable origin from its visible edge - which broke once already in catch) also stops working on the drum roll because of that ([video](https://streamable.com/2p8w9b)).

peppy:
Yeah, I figured there was probably a reason for that being the way it was. Guess I'll have to somehow hack it into the draw space.

peppy:
@bdach in your video the drumrolls are also offset incorrectly temporally. i can't reproduce this, were you on the latest commit of this branch?

bdach:
I'm somewhat sure I did - my head of this branch is at bfc0d41, but I messed with the code a little bit so I can't recall now if the tree was clean or not. Will try to repro and provide details later today.

osu.Game.Rulesets.Taiko/Objects/Drawables/Pieces/ElongatedCirclePiece.cs
```diff
@@ -12,18 +14,15 @@ public ElongatedCirclePiece()
...
         {
             base.Update();
-
-            var padding = Content.DrawHeight * Content.Width / 2;
-
-            Content.Padding = new MarginPadding
-            {
-                Left = padding,
-                Right = padding,
-            };
```
bdach, line 25: Can confirm reproduction of the misalignment; it shows up even on default lazer skin in this PR (or, to be more precise, on f36477e3 with a clean tree), doesn't show up on `master`. Will be visible in `TestSceneAutoplay`. Screenshots:

| `master` | this PR |
| :-: | :-: |
| image | image |

Seems to be related to removing the padding as highlighted. Bringing it back fixes the alignment of `ElongatedCirclePiece` itself but ticks remain broken (too stretched out).

peppy: The padding needs a local comment explaining exactly what it's doing until we fix this in a better way. I thought I was able to remove it due to the refactors I made and there was nothing noticeably failing as a result.

peppy: @bdach can you explain how the offset is occurring? in your screenshots you show it changing at the `DrawableDrumRoll` level, but I can't seem to find any changes I've made which should affect that.

bdach: Not currently in a position to investigate in detail but from what I remember the actual drumroll's position hasn't changed from `master`, it's the `ElongatedCirclePiece` lower down that starts protruding out of it, breaking alignment.

peppy: I think I've fixed this by applying at a more central level. Can you check on the solution?

bdach: Give me a few hours and I'll get back to you on that.

bdach: Took me a little longer than I promised, but as far as I can see the alignment is fixed, although this set:

https://github.com/ppy/osu/blob/58a1c6e17186084345feedbd1c5b0d44c2ce8695/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableDrumRoll.cs#L145

doesn't seem to have any effect as the getter is hard-overridden in `DrawableTaikoHitObject` itself:

https://github.com/ppy/osu/blob/8b83c3daabf83075ce0be0c47948a312529c9e1d/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableTaikoHitObject.cs#L113

so the value gets set but never actually read (draw visualiser shows the one from base).

bdach:
If you haven't figured it out yet, I looked into this some more and as for the reason that proxying helps for visibility of the circle piece at the start, I found out that once it scrolls off-screen it becomes masked away, which xmldocs indicate proxying would help with:

https://github.com/ppy/osu/blob/8b83c3daabf83075ce0be0c47948a312529c9e1d/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableTaikoHitObject.cs#L38-L41

One piece of the puzzle is that `DrawableHitObject` itself prevents updates of masking before all nested objects are judged:

https://github.com/ppy/osu/blob/9dfca4a374a20ac0aea6ca8d62fd6d41818e72b3/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L401

which prevents the masking from being reprocessed after rewind (removing the `AllJudged` check also makes the circle piece appear).

As to why only the circle piece at the start gets this treatment, I'm almost sure this is due to the positioning of the elements themselves. The circle piece is the only element of the drum roll that has a chance to be masked away before the entire drum roll is, so it gets set `false` when it goes off-screen. When the *entire* drum roll goes off-screen, however, it gets masked away *itself*, therefore not propagating masking downwards (`UpdateSubTreeMasking` early-returns and never sets `IsMaskedAway = true` on children).

smoogipoo:
I only allowed masking to continue when `AllJudged = true` just to enable optimisations. It wasn't fixing an issue or anything.

Given that we've tightened up lifetime quite a lot, I'd be fine with removal of the `AllJudged` component (`UpdateSubTreeMasking() => false`).

peppy:
Looks good with that change.

smoogipoo (approved):

PR #8768, opened 2020-04-16
Hide edit context menu item in multiplayer song select

Joehuu:
Fixes the UX concern in https://github.com/ppy/osu/pull/7667:
> We probably want to hide the context menu, but that's not really simple to do based on the current structure of things.


peppy:
I made some changes to this to simplify the checks (and standardise across other menu items).

@smoogipoo request a quick check on this one.

peppy (approved):

smoogipoo (approved):
lgtm

PR #8769, opened 2020-04-16
Fix spinners being considered the "first object" for increased visibility in hidden

### LINKED ISSUE #8063: For "Increase visibility of first object when visual impairment mods are enabled," ignore Spinners and work after breaks.

### LINKED ISSUE #8153: First hitcircle not vanishing with HD

peppy:
Closes #8063
Closes #8153

Alternative to #8227

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
+++ b/osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
@@ -1,0 +1,106 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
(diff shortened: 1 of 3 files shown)


bdach (approved):
looks good to me 👍

smoogipoo (approved):

PR #8772, opened 2020-04-16
Fix carousel not correctly updating when selection changes to a new beatmap from a child screen

### LINKED ISSUE #8269: Switching to a beatmap, downloaded while in SoloResults, can cause the game to show the previous one in the carousel
**Describe the bug:** If you try to switch to a new beatmap using direct (this also probably works with the "download completed" notification) from the `SoloResults` screen causes the game to be able to desync the carousel beatmap from the working one (like in my video example, I went back to `SoloResults` screen and there the issue appeared). Also, if you try to play the "wrong" beatmap, it's actually the new one (but I forgot to show it in the video, so it's hard to explain).

**Screenshots or videos showing encountered issue:** https://streamable.com/6904z

**osu!lazer version:** 2020.312.0

**Logs:** (same of the video)
[database.log](https://github.com/ppy/osu/files/4329602/database.log)
[runtime.log](https://github.com/ppy/osu/files/4329603/runtime.log)

peppy:
- Closes https://github.com/ppy/osu/issues/8269.

Basically, the carousel was performing its pending schedule task:

https://github.com/ppy/osu/blob/master/osu.Game/Screens/Select/BeatmapCarousel.cs#L157

after SongSelect:

https://github.com/ppy/osu/blob/master/osu.Game/Screens/Select/SongSelect.cs#L377

You could ask "why not schedule the workingBeatmapChanged?" but scheduling logic in SongSelect was recently cleaned up as it was deemed to be very fragile, so I'm against adding more schedules. I think it would require two, anyway.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -217,6 +217,9 @@ namespace osu.Game.Screens.Select
         /// <returns>True if a selection was made, False if it wasn't.</returns>
         public bool SelectBeatmap(BeatmapInfo beatmap, bool bypassFilters = true)
         {
+            // ensure that any pending events from BeatmapManager have been run before attempting a selection.
+            Scheduler.Update();
+
             if (beatmap?.Hidden != false)
                 return false;
 


bdach:
Checked this out, repro'd on master, confirmed that this PR fixes the issue and the explanation makes sense to me, but I'm a little skittish as to approve this outright as overall I'm not that comfortable around and with carousel code. Maybe I should just bite the bullet and dive into it one time.

smoogipoo (approved):
Seems ok

PR #8778, opened 2020-04-17
Add method of creating a ruleset for testing with in OsuTestScene

frenzibyte:
As proposed by @smoogipoo.

This does not only add a `CreateRuleset()` method, but centralize all types of test scenes that require a ruleset for testing into using this new global method.

## Remarks
 - This got way too big upon removing the `PlayerTestScene` constructor as you can see, so I've chose (*with discussion*) to split it from its dependent PR.
 - ~~I wasn't sure about constructing a whole ruleset by `CreateRuleset()` just to get its name to be used for a test step name inside [`PlayerTestScene`](https://github.com/iiSaLMaN/osu/blob/7f791dcdf04d3434ea7275de20fd8d369060b062/osu.Game/Tests/Visual/PlayerTestScene.cs#L49) (as the bindable is set while the test scene is in loading stages), but it's better than leaving the step with an anonymous name (`load ruleset`) or something.~~
 - ~~I've chosen to throw on null `CreateRuleset()`s for test scene implementations based on whether that would cause things to work *silently* improperly, for example on `SkinnableTestScene`, forgetting to override `CreateRuleset()` means the incorrect legacy skin transformer will be added and cause the content to behave improperly.~~
 - ~~Although the `Ruleset.Value` always has a default ruleset (most likely `OsuRuleset`), and `PlayerTestScene` taking that value as the ruleset of the beatmap & player, you can potentially use `PlayerTestScene` without overriding `CreateRuleset()`, but was unsure as now the bindable will not be disabled, so left that up to the reviewers to decide...~~


osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModPerfect.cs
```diff
@@ -50,5 +50,7 @@ public void TestJuiceStream(bool shouldMiss)
         // We only care about testing misses, hits are tested via JuiceStream
         [TestCase(true)]
         public void TestTinyDroplet(bool shouldMiss) => CreateHitObjectTest(new HitObjectTestData(new TinyDroplet { StartTime = 1000 }), shouldMiss);
+
+        protected override Ruleset CreateRuleset() => new CatchRuleset();
```
bdach, line 54: Not sure about the placement of these, I'd probably want them to be closer to the top (before ctor?)

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -70,7 +70,22 @@ protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnl
...
+            {
+                // re-enable the bindable in case it was disabled.
+                // happens when restarting current test scene.
+                Ruleset.Disabled = false;
+
+                // Set global ruleset bindable to the ruleset defined
+                // for this test scene and disallow changing it.
+                Ruleset.Value = definedRuleset;
+                Ruleset.Disabled = true;
+            }
```
bdach, line 86: What's this logic for? Is there a particular scenario that needs this?

If we do decide to have something like this then maybe `OsuTestScene` should take a lease of this?

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -124,6 +139,14 @@ public virtual void RecycleLocalStorage()
         [Resolved]
         protected AudioManager Audio { get; private set; }
 
+        /// <summary>
+        /// Creates the ruleset to be used for this test scene.
+        /// </summary>
+        /// <remarks>
+        /// When testing against ruleset-specific components, this method must be overriden to their ruleset.
+        /// </remarks>
+        protected virtual Ruleset CreateRuleset() => null;
```
bdach, line 148: Maybe mark as
```suggestion
        [CanBeNull]
        protected virtual Ruleset CreateRuleset() => null;
```
which reveals a potentially unsafe access at `PlayerTestScene`, which in turn I would resolve with
```csharp
        [NotNull]
        protected abstract Ruleset CreatePlayerRuleset();

        protected sealed override Ruleset CreateRuleset() => CreatePlayerRuleset();
```

osu.Game/Tests/Visual/SkinnableTestScene.cs
```diff
@@ -30,11 +29,15 @@ public abstract class SkinnableTestScene : OsuGridTestScene
         protected SkinnableTestScene()
             : base(2, 3)
         {
+            // avoid running silently incorrectly.
+            if (CreateRuleset() == null)
+            {
+                throw new InvalidOperationException(
+                    $"No ruleset provided, override {nameof(CreateRuleset)} to the ruleset belonging to the skinnable content."
+                    + "This is required to add the legacy skin transformer for the content to behave as expected.");
+            }
```
bdach, line 38: would do the same trick as above with a new abstract property that `CreateRuleset()` would delegate to, to avoid failing at runtime and move the failure to compile-time

iiSaLMaN: properties shouldn't return new instances afaik, will make it `CreateRulesetForSkinProvider()` instead.

osu.Game/Tests/Visual/OsuTestScene.cs
iiSaLMaN: After rethinking this for a bit, seems like going with your suggestion is the best way to have a good access to `Ruleset.Value`.

osu.Game/Tests/Visual/SkinnableTestScene.cs
```diff
@@ -27,14 +28,19 @@ public abstract class SkinnableTestScene : OsuGridTestScene
         private Skin specialSkin;
         private Skin oldSkin;
 
+        /// <summary>
+        /// Creates the ruleset for adding the ruleset-specific skin transforming component.
+        /// </summary>
+        [NotNull]
+        protected abstract Ruleset CreateRulesetForSkinProvider();
```
smoogipoo, line 35: Nits: Methods should not be able ctors (done quite a few times in this PR).

osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
```diff
@@ -15,10 +15,7 @@ namespace osu.Game.Rulesets.Osu.Tests.Mods
 {
     public class TestSceneOsuModHidden : ModTestScene
     {
-        public TestSceneOsuModHidden()
-            : base(new OsuRuleset())
-        {
-        }
+        protected override Ruleset CreatePlayerRuleset() => new OsuRuleset();
```
smoogipoo, line 18: At the point where you're making an `TestSceneOsuPlayer` and deriving from it for other tests, you might also make an abstract `OsuModTestScene` and derive that for all the mod test scenes.

It'd probably be a good path forward to provide ready-made abstractions for all ruleset-requiring tests anyway in the template projects.

osu.Game/Tests/Visual/OsuTestScene.cs
smoogipoo: Why was this resolved without comment? I'd also like to know the reason for this.

smoogipoo (changes requested):

osu.Game/Tests/Visual/OsuTestScene.cs
iiSaLMaN: I've let the treatment of this bindable be the same as `Beatmap`'s, and since it's only used in `PlayerTestScene` due to that, I don't think I really need to disable it.

osu.Game/Tests/Visual/SkinnableTestScene.cs
iiSaLMaN: I really don't want to repeat changing the order due to how many files changed for it, where should it reside? at the bottom of the ctor, or same as it was before?

bdach: My thought process with asking for it to be higher was to treat it in a similar manner to `RequiredTypes` is - as setup/test declaration code and not an actual method. Unless you mean only this abstract one in the base...?

osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
iiSaLMaN: > At the point where you're making an `TestSceneOsuPlayer` and deriving from it for other tests, you might also make an abstract `OsuModTestScene` and derive that for all the mod test scenes.
> 
> It'd probably be a good path forward to provide ready-made abstractions for all ruleset-requiring tests anyway in the template projects.

I was aware initially of making that for the `ModTestScene` in particular as that may potentially start containing mod-specific test scenes like the existing `ModPerfectTestScene`.

But actually that may help bringing up the ruleset used for the test scene to become more of a test-declaration code, should I enforce these ready-made abstraction classes for *all* usages? (excluding mod-specific test scenes)

osu.Game/Tests/Visual/OsuTestScene.cs
bdach: If nothing relies on this logic and it was added "just because" then I'd be in favour of removing it.

peppy:
@frenzibyte want to conflict resolve and make sure this is still in a state you're happy with?

frenzibyte:
@peppy Confliction resolved, along with some fixes and code rewrites that satisfy above conversations.

Seems to be in a good state, and works well with the dependent PR.

osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
frenzibyte: Applied only to rulesets that have more than one sub-`ModTestScene` (osu! ruleset currently).

osu.Game/Tests/Visual/SkinnableTestScene.cs
frenzibyte: Right, for declaring classes this should reside where protected methods usually reside, applied.

osu.Game/Tests/Visual/OsuTestScene.cs
frenzibyte: This logic was actually supposed to be removed at the time of my above comment, now I made sure I actually removed it, resolving.

osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -136,7 +146,8 @@ public virtual void RecycleLocalStorage()
         [BackgroundDependencyLoader]
         private void load(RulesetStore rulesets)
         {
-            Ruleset.Value = rulesets.AvailableRulesets.First();
+            if (!Ruleset.Disabled)
+                Ruleset.Value = rulesets.AvailableRulesets.First();
```
bdach, line 150: When can this be disabled? Not sure this should just silently keep going whenever it is, and the old code didn't care either...

Think it might be a vestige of the old now-removed disabling logic as it was alongside it in 67bd7bfa3905aa96f21f2225a517e2e369e80540.

osu.Game/Tests/Visual/PlayerTestScene.cs
```diff
@@ -53,7 +56,7 @@ protected void CreateTest(Action action)
 
             action?.Invoke();
 
-            AddStep(ruleset.RulesetInfo.Name, LoadPlayer);
+            AddStep(ruleset.Description, LoadPlayer);
```
bdach, line 59: Kind of a drive-by change, but sure.

osu.Game/Tests/Visual/PlayerTestScene.cs
```diff
@@ -19,20 +20,22 @@ public abstract class PlayerTestScene : RateAdjustedBeatmapTestScene
...
-            this.ruleset = ruleset;
+            ruleset = CreatePlayerRuleset();
         }
 
-        protected OsuConfigManager LocalConfig;
-
         [BackgroundDependencyLoader]
         private void load()
         {
+            Ruleset.Value = ruleset.RulesetInfo;
```
bdach, line 37: I would add your note in the commit message:

> There are test scenes using current value of ruleset bindable on their BDL (example in TestSceneSliderSnaking's BDL)

as an inline comment here.

bdach (commented):
Think this is mostly OK now. I have a few nits that I would have applied myself but there's still one outstanding change (the `Disabled` check) I want to clarify first.

osu.Game/Tests/Visual/OsuTestScene.cs
frenzibyte: oh right, I actually thought it stayed like this before this PR, definitely should be removed.

osu.Game/Tests/Visual/PlayerTestScene.cs
```diff
@@ -19,20 +20,24 @@ public abstract class PlayerTestScene : RateAdjustedBeatmapTestScene
...
         }
 
-        protected OsuConfigManager LocalConfig;
-
         [BackgroundDependencyLoader]
         private void load()
         {
+            // There are test scenes using current value of the ruleset bindable
+            // on their BDLs (example in TestSceneSliderSnaking's BDL)
+            Ruleset.Value = ruleset.RulesetInfo;
```
smoogipoo, line 39: I'm a little unusure of this being done in every place. Seems like something that could be missed?

Is there a reason this is done here (`PlayerTestScene`, `EditorTestScene`, `SkinnableTestScene`) rather than inside `OsuTestScene`'s BDL itself? I'm fine with the re-exposed abstracts, it's just this part that seems kinda weird.

smoogipoo (dismissed):
I think this looks fine? @peppy want your review too.

osu.Game/Tests/Visual/SkinnableTestScene.cs
```diff
@@ -120,6 +118,14 @@ private Drawable createProvider(Skin skin, Func<Drawable> creationFunction, IBea
             };
         }
 
+        /// <summary>
+        /// Creates the ruleset for adding the corresponding skin transforming component.
+        /// </summary>
+        [NotNull]
+        protected abstract Ruleset CreateRulesetForSkinProvider();
```
peppy, line 125: doesn't match other methods. should be `CreateSinProviderRuleset` or similar.

osu.Game.Tests/Visual/Gameplay/TestPlayerTestScene.cs
```diff
@@ -0,0 +1,16 @@
...
+
+using osu.Game.Rulesets;
+using osu.Game.Rulesets.Osu;
+
+namespace osu.Game.Tests.Visual.Gameplay
+{
+    /// <summary>
+    /// A <see cref="PlayerTestScene"/> with an arbitrary ruleset value to test with.
+    /// </summary>
+    public abstract class TestPlayerTestScene : PlayerTestScene
```
peppy, line 12: abstract test classes aren't tests and shouldn't start with `Test`. maybe `OsuPlayerTestScene`

peppy (changes requested):

osu.Game/Tests/Visual/SkinnableTestScene.cs
frenzibyte: was following up with its neighbour method (`CreateBeatmapForSkinProvider`)

peppy: fair enough

peppy (approved):

smoogipoo (approved):

smoogipoo:
Let's see how this goes...

PR #8779, opened 2020-04-17
Fix presenting new ruleset and beatmap at once causing wedge display desync

### LINKED ISSUE #8344: Switching to a beatmap of another ruleset can cause desync if the "Show converted beatmaps" option is enabled
**Describe the bug:** Switching to a beatmap (using the `UserProfileOverlay`) with a ruleset that accepts the conversion to the previous one (e.g. _osu!taiko_ → _osu!_), with the `Show converted beatmaps` option **enabled**, causes the `BeatmapInfoWedge` to desync.

**Screenshots or videos showing encountered issue:** https://streamable.com/g18k8
(The first example shows the normal behaviour with the `Show converted beatmaps` option disabled, the second one shows the issue)

**osu!lazer version:** 2020.317.0

peppy:
- Closes https://github.com/ppy/osu/issues/8344

Previously, if `OsuGame.PresentBeatmap` was called in a scenario where it would change the ruleset and beatmap while song select was already visible, the following flow would happen:

- Ruleset is changed, triggering `transferRulesetValue` and returning, under the presumption that this would end up calling `performUpdateSelected` again.
- This would not necessarily happen if the current beatmap is not filtered by the ruleset change (ie. convert displayed and selected).
- Due to the early exit, `updateComponentFromBeatmap` would not get hit.

This fix ensures there is a re-run of the `performUpdateSelected` method regardless. It's a very simple fix and should be easy to comprehend why this failed from a review perspective, but tracking down what was going wrong took several hours.

If song select logic regresses again it may be time to rethink this pattern altogether.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -448,8 +448,10 @@ namespace osu.Game.Screens.Select
                 {
                     Mods.Value = Array.Empty<Mod>();
 
-                    // required to return once in order to have the carousel in a good state.
-                    // if the ruleset changed, the rest of the selection update will happen via updateSelectedRuleset.
+                    // the ruleset transfer may cause a deselection of the current beatmap (due to incompatibility).
+                    // this can happen via Carousel.FlushPendingFilterOperations().
+                    // to ensure a good state, re-transfer no-debounce values.
+                    performUpdateSelected();
                     return;
                 }
 
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -426,7 +426,7 @@ namespace osu.Game.Screens.Select
         }
 
         /// <summary>
-        /// selection has been changed as the result of a user interaction.
+        /// Selection has been changed as the result of a user interaction.
         /// </summary>
         private void performUpdateSelected()
         {
@@ -435,7 +435,7 @@ namespace osu.Game.Screens.Select
 
             selectionChangedDebounce?.Cancel();
 
-            if (beatmap == null)
+            if (beatmapNoDebounce == null)
                 run();
             else
                 selectionChangedDebounce = Scheduler.AddDelayed(run, 200);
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -448,8 +448,10 @@ void run()
                 {
                     Mods.Value = Array.Empty<Mod>();
 
-                    // required to return once in order to have the carousel in a good state.
-                    // if the ruleset changed, the rest of the selection update will happen via updateSelectedRuleset.
+                    // the ruleset transfer may cause a deselection of the current beatmap (due to incompatibility).
+                    // this can happen via Carousel.FlushPendingFilterOperations().
+                    // to ensure a good state, re-transfer no-debounce values.
```
bdach, line 453: I *think* I understand the fix (took me a solid half an hour of staring at the control flow, but that's probably just my brain leaking), but this comment reads a bit backwards even in light of the PR description? Everything in it is *correct*, the transfer may cause deselection, but in this particular failure case it's the *lack* of deselection which causes the `run` delegate to not fire the second time, right? Again, might just be me, but this reads a little like it was the other way around.

peppy: I agree this reads pretty badly still. Will refactor accordingly...

peppy: @bdach I refactored this with a slightly different approach that should hopefully be more self-explanatory.

A next step forward may be reversing the flow such that `workingBeatmapChanged` just calls `performUpdateSelected` and the carousel select logic is moved to `performUpdateSelected`, but I'm not ready to take that on just yet..

smoogipoo (approved):
This seems to work okay and I can kind of gather my head around the change. But I don't trust this code path at all, the amount of mental hoops to jump through is too much.

In the future I'd hope we can remove this entire code path and handle ruleset/beatmap updates separately.

bdach:
Way too late for my opinion to be relevant but fwiw I did take a look and I can understand this variant of the fix as well. It's possibly a little more understandable but my level of confidence in saying that is not too high and I agree with @smoogipoo's assessment.

PR #8788, opened 2020-04-18
Fix hard crash in editor on legacy modes without encoder implementation

### LINKED ISSUE #8786: Unhandled error when trying to edit a non-std beatmap
**Describe the bug:** 
This will also make the editor unable to quit via ESC key, only toolbar button "Home" and [alt]+[f4] works

screenshot

**osu!lazer version:** 
commit 149efec

**Logs:**
[logs.zip](https://github.com/ppy/osu/files/4495984/logs.zip)

peppy:
Closes https://github.com/ppy/osu/issues/8786

### DIFF
--- a/osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
@@ -187,14 +187,13 @@ namespace osu.Game.Beatmaps.Formats
 
             writer.WriteLine("[HitObjects]");
 
+            // TODO: implement other legacy rulesets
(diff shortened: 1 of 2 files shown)


smoogipoo:
I think better is to remove the exceptions from being thrown within the encoder itself, since it'll still crash using the save menu item.

bdach:
Tested this, and it works fine as a stop-gap. 👍 

As work on custom rulesets continues we will eventually probably need some way to signal whether a ruleset has an editor/encoder implementation and disallow editing/saving it if one is missing, but that's not a high-priority thing for now.

smoogipoo (approved):

PR #8791, opened 2020-04-18
Let carousel fill most of song select

### LINKED ISSUE #2238: Small area between rankings and beatmap carousel that controls volume when scrollwheel is used
That small area doesn't exist in stable client and is really annoying. The entire area should scroll the carousel instead.

LittleEndu:
Fixes #2238

This PR moves ``BeatmapCarousel`` from its old grid it shared with ``WedgeBackground`` in ``SongSelect`` to it's very own grid. This grid is has the same width distribution as ``BeatmapInfoWedge``'s grid has, but carousel just makes use of the right side. 
Then this PR makes the ``scrollableContent`` in carousel have fixed width of 850, which is the old size of the carousel, and places ``scrollableContent``'s anchor & origin to top right.


LittleEndu:
The way I merged master moves the loading circle a bit to the left:

This PR:
osu_2020-05-01_10-56-28

Current master:
osu_2020-05-01_10-58-37

peppy:
How does this look with content displayed?

LittleEndu:
> How does this look with content displayed?

Exactly the same as it does on master. I was aiming for keeping it same but doing it for the loading circle would require having it as its own field and replacing both when carousel load completes.

peppy:
I really dislike having this second `GridContainer` just for this purpose...

LittleEndu:
> I really dislike having this second `GridContainer` just for this purpose...

There's three things that need to happen:
 * Carousel content needs to be 850 units wide from the right side
 * The shadows forming the arrow need to reach the content from the left side
 * Carousel should accept input starting from details area

This means that there's overlap between the shadows and the carousel. Unsure if this could be achieved in a different way and still have everything resize automatically.

bdach:
As a user of a 21:9 display I feel like the UX with this change is worse than it is on `master`. Here's a [video](https://streamable.com/wd2s7u), and to explain as to why:

* Being able to scroll the carousel when hovering over the semi-transparent wedge, far away from the actual carousel, feels wrong.
* In addition there is still a small gap between the metadata displays and the carousel "scrolling area" that you can still control volume on by scrolling, which feels unintended judging from the [proposal screenshot](https://user-images.githubusercontent.com/11320483/45510972-6f063f00-b7a4-11e8-8671-f5d540457c3e.png) in the original issue.

bdach (changes requested):
As above.

LittleEndu:
Do you also find the same scrolling behaviour strange in stable? Because that's the only reason I find it strange that most of song select does not currently control the carousel, I'm just so used to keeping my cursor in the middle and scrolling from there.

bdach:
I'd say so, yeah. Myself I usually keep the cursor somewhat close to carousel whenever I'm scrolling on stable. I think the wedge being darkened makes this feel worse in lazer, though, because the way it's done right now it *looks* as if that area is explicitly *not* the carousel as it's clearly delineated from it (on stable there's no such thing, it's just the beatmap background, so it's a bit less jarring).

LittleEndu:
I'm going to close this PR and wait for song select redesign. If the redesign doesn't fix #2238, a new design that does would need to be implemented anyway.

PR #8799, opened 2020-04-19
Replace AllJudged event logic with HasCompleted bindable

frenzibyte:
Dependency for #8800

Now with the dependent PR, the `Player` has to know if a judgement has been reverted after `AllJudged` has been raised, so I've decided to remove the event altogether and replace it with a `Bindable` instead.

## Remarks

- I'm not quite sure about whether to name it `HasCompleted` or `AllJudged` as they mostly serve the same meaning..., went with `HasCompleted`.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -197,7 +197,7 @@ namespace osu.Game.Screens.Play
             };
 
             // Bind the judgement processors to ourselves
-            ScoreProcessor.AllJudged += onCompletion;
+            ScoreProcessor.HasCompleted.ValueChanged += updateCompletionState;
             HealthProcessor.Failed += onFail;
 
             foreach (var mod in Mods.Value.OfType<IApplicableToScoreProcessor>())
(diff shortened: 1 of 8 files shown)


peppy:
What is the dependent PR?

iiSaLMaN:
@peppy Here it is: #8800

osu.Game/Screens/Play/Player.cs
```diff
@@ -412,7 +412,7 @@ public void Restart()
 
         private ScheduledDelegate completionProgressDelegate;
 
-        private void onCompletion()
+        private void updateCompletionState(ValueChangedEvent<bool> completionState)
```
peppy, line 415: this is going to trigger multiple times. you aren't cancelling or checking the value is true even.

iiSaLMaN: right, I'll close this. you can review #8800 directly instead.

iiSaLMaN:
Closing, #8800 can be reviewed as one PR.

PR #8800, opened 2020-04-19
Fix results screen pushed after rewinding in-between push delay

### LINKED ISSUE #8793: rewinding replay as soon as it ends may result in incorrect result screen score
When the replay ends, if you try to rewind the replay it shows a result screen with incorrect score

The final score should be 1000000 but it's not.
https://streamable.com/zawmoz

osu!lazer 2020.418.0

frenzibyte:
Closes #8793 

Although the issue is talking about the results screen pushed with current score after rewinding, as I've proposed in the thread, it's kind of annoying to push the results screen after rewinding from completion (by accident for example), so I've decided to solve this by cancelling the push delegate instead which technically solves this.

## Remarks
 - I wasn't initially sure about this solution but given that the `Player` push the results screen when all hit objects have been judged, this will do for now.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneCompletionCancellation.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneCompletionCancellation.cs
@@ -46,46 +46,54 @@ namespace osu.Game.Tests.Visual.Gameplay
         [Test]
         public void TestCancelCompletionOnRewind()
         {
-            cancelCompletionSteps();
+            complete();
+            cancel();
 
-            AddAssert("no attempt to push ranking", () => !((FakeRankingPushPlayer)Player).GotoRankingInvoked);
+            checkNoRanking();
         }
 
         [Test]
         public void TestReCompleteAfterCancellation()
         {
-            cancelCompletionSteps();
-
-            // Attempt completing again.
-            AddStep("seek to completion again", () => track.Seek(5000));
-            AddUntilStep("completion set by processor", () => Player.ScoreProcessor.HasCompleted.Value);
+            complete();
+            cancel();
+            complete();
 
             AddUntilStep("attempted to push ranking", () => ((FakeRankingPushPlayer)Player).GotoRankingInvoked);
         }
 
         /// <summary>
-        /// Tests whether can still pause after cancelling completion
-        /// by reverting <see cref="IScreen.ValidForResume"/> back to true.
+        /// Tests whether can still pause after cancelling completion by reverting <see cref="IScreen.ValidForResume"/> back to true.
         /// </summary>
         [Test]
         public void TestCanPauseAfterCancellation()
         {
-            cancelCompletionSteps();
+            complete();
+            cancel();
 
             AddStep("pause", () => Player.Pause());
             AddAssert("paused successfully", () => Player.GameplayClockContainer.IsPaused.Value);
+
+            checkNoRanking();
         }
 
-        private void cancelCompletionSteps()
+        private void complete()
         {
             AddStep("seek to completion", () => track.Seek(5000));
             AddUntilStep("completion set by processor", () => Player.ScoreProcessor.HasCompleted.Value);
+        }
 
+        private void cancel()
+        {
             AddStep("rewind to cancel", () => track.Seek(4000));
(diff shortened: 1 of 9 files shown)


bdach (commented):
As a side note I got the replay to pause after rewinding upon completion a few times (seems to be timing-dependent). Haven't looked in detail as to why, it could be a separate thing.

osu.Game/Screens/Play/Player.cs
```diff
@@ -412,12 +412,22 @@ public void Restart()
...
-        private void onCompletion()
+        private void updateCompletionState(ValueChangedEvent<bool> completionState)
         {
             // screen may be in the exiting transition phase.
             if (!this.IsCurrentScreen())
                 return;
 
+            // cancel push delegate in case judges reverted
+            // after delegate may have already been scheduled.
+            if (!completionState.NewValue)
```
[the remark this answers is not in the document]
iiSaLMaN: using `OldValue` in bool events is redundant here as I haven't used `BindValueChanged(..., true)` which makes `NewValue always the opposite of `OldValue`, therefore not adding anything.

btw the operations done inside this block is required to make the cancellation correct (hopefully), as they all reset everything that was set when a completion occurs:
 - `completionProgressDelegate?.Cancel()` – cancel the completion delegate to prevent from pushing. the delegate may already be null, so add a null-conditional, due to `allowResults` being false, thus not creating the delegate.
 - `completionProgressDelegate = null` – reset the delegate back to null to assure the `Player` that there isn't a push delegate scheduled.
- `ValidForResume = true;` – Reset back `ValidForResume` from being false to allow pausing back (though just remembered that rewinding comes only on replays which you can't pause, but to ensure that there are no traces of the player going through the completion process.)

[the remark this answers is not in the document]
bdach: Fair point about the old value being redundant, although there was no need to explain the rest as it was pretty obvious even to me what it did.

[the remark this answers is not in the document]
iiSaLMaN: ok great, *I was literally about to start commenting on each line in case something wasn't obvious*

osu.Game.Tests/Visual/Gameplay/TestSceneCompletionCancellation.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        private void cancelCompletionSteps()
+        {
+            AddStep("seek to completion", () => track.Seek(5000));
+            AddUntilStep("completion set by processor", () => Player.ScoreProcessor.HasCompleted.Value);
+
+            AddStep("rewind to cancel", () => track.Seek(4000));
+            AddUntilStep("completion cleared by processor", () => !Player.ScoreProcessor.HasCompleted.Value);
+
+            AddWaitStep("wait", 5);
```
peppy, line 86: this seems prone to potential failure. can you make it an until step instead?

iiSaLMaN: I don't think I can, this is a step that waits a couple of seconds before asserting to ensure that there was no attempt to push ranking screen, I will describe that in the step for clarity.

peppy: In that case please make a const in player (`RESULTS_DISPLAY_DELAY`) and base the wait step duration on `RESULTS_DISPLAY_DELAY / TimePerAction * 2` or similar (`* 2` is to add some degree of safety), and add a comment.

iiSaLMaN: Although with current values `(1000 / 200 * 2)`, `* 2` actually buffed up the wait count to `10`, I think ceiling is more appropriate here such that it doesn't fall one wait count behind.

peppy: you may be able to reproduce an fail under visual tests due to the way scheduler works. make sure to test it thoroughly if you want to cut it that close.

peppy (approved):

peppy (approved):

PR #8801, opened 2020-04-19
Add /np (now playing) command support in chat

peppy:


osu.Game.Tests/Visual/Online/TestNowPlayingCommand.cs
```diff
@@ -0,0 +1,85 @@
...
+using osu.Game.Online.Chat;
+using osu.Game.Rulesets;
+using osu.Game.Users;
+
+namespace osu.Game.Tests.Visual.Online
+{
+    [HeadlessTest]
+    public class TestNowPlayingCommand : OsuTestScene
+    {
+        [Cached(typeof(IChannelPostTarget))]
```
peppy, line 18: unsure why this is required even though it's specified on the interface (not required for the game usage, just for nunit testing).

iiSaLMaN: I have debugged a bit through this and it seems that this is an edge case where having a property / field that is of a type that has `Cached` attribute will never be cached..., will look into fixing this framework-side if you haven't gone already.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Online/TestNowPlayingCommand.cs
peppy: what you are saying makes no sense

iiSaLMaN: So the actual usage in the game works fine because the commands are children to the `ChannelManager` class with `IChannelPostTarget` interface assigned, while here in this test, the post target here is a property instead with type `PostTarget` that has `IChannelPostTarget` interface assigned to it.

There could be two solutions this:
 1. To solve this here by making the `PostTarget` class a `CompositeDrawable` and adding commands inside it.
 2. To solve this framework-side by making `CachedAttribute.CreateActivator(...)` aware of properties and fields that has a type with `CachedAttribute` assigned to it or the type's interfaces.

I'm a bit hesitant about the second solution as it may introduce in bugs with currently existing usages.

peppy: the problem is the cached is not working. stating arbitrary workarounds is unnecessary.

smoogipoo: I'd say this is intended, fwiw. Caching on class/interface definitions respects the top-down model of the hierarchy, where the cached value is available for all its children.

Having it work as expected here means there would be a lateral component to it, where a child now caches itself in its parent for all other children of the parent to access. Which sounds wrong.

The usage here is correct in my books.

peppy: I see. Might be worthwhile adding a mention of this to `[Cached]` (if it tripped me up, it's likely going to trip others up).

hbnrmx:
should add `np` to available commands in `/help`

bdach:
Implementation seems ok, even though the switch is getting a little long (as codefactor reports) and the fact commands need to be manually added to `/help` is a bit annoying; maybe we could make that implementation using `Component`s a pattern that every command could use? (not here, obviously, but as a future step)

smoogipoo (approved):

Issue #8413, opened 2020-03-23
Highlight percentage and combo counter when they are perfect


smileyhead:
**Describe the new feature:** Sometimes I glance over to the percentage or combo counters during gameplay to see if they are still perfect and that causes me to break the combo or screw up the timing. Highlighting these counters would make it easy to see them in your peripheral vision and not look away from what you're doing. As soon as the percentage dips below 100% or you lose your full combo, the respective counter would go back to normal.

**Proposal designs of the feature:** 

![image](https://user-images.githubusercontent.com/15932372/77346272-474b4c00-6d36-11ea-9def-761834793a00.png)

peppy:
This will 100% be a thing (was discussing last week with designer) but I don't think the design will look as you have it.

PR #8802, opened 2020-04-19
Allow user to always play first combo break sound regardless of lost combo count

LittleEndu:
This has been a feature many people have requested on [feature request forum](https://osu.ppy.sh/community/forums/topics/560354).

This PR adds a setting that allows user to always play first combo break sound. That setting is used in ``ComboEffects``. For some reason ``runOnceImmediately`` was set for combo bindable even tho our function needs both old and new values which ``runOnceImmediately`` doesn't give (it gives only the current value as both old and new). This actually broke the new functionality so I just removed it.


peppy:
If this is to be implemented at a non-skin level, it should be "always play combo break". Not just the first.

LittleEndu:
Always play combo break would sound ear shattering if you miss a stream or a buzz slider. This has been requested as "always play first" to prevent reverse chokes.

bdach:
To repeat the mantra: the new logic is complex, so I'd like to see tests here.

LittleEndu:
Yes, of course. Is there a way to see that a sound sample is playing?

peppy:
It would only play if you break, not on every miss. We are not implementing this feature as you have it. If we chose to do so, it would not be a setting.

peppy:
@LittleEndu are you planning to resolve this? either follow the logic i mention above, or make the setting a slider "Play combo break sound after combo of length: 1 ~ 100"

LittleEndu:
I can't get rider eap to work for free anymore but I guess I can work on my remaining PRs with visual studio.

Is having the setting as a slider fine? You previously said there won't be a setting.

huoyaoyuan:
The logic can be reformed as **always play break sound, unless another break happened recently (within an adjustable range)**.

EVAST9919:
Also would be nice to make this value (20 combo) adjustable per ruleset in the future

peppy:
@EVAST9919 reasoning? if that setting is required, it sounds like a ruleset is potentially overusing combo?

EVAST9919:
Bosu is a good example. There are thousands of objects and each avoided one is +combo. So if you miss one - you will easily gain 20 combo again within a few seconds. 
I'm not saying that changing this value per ruleset is the best approach, just a proposal

peppy:
Have you considered using the tick counter for smaller objects as to not have the combo increase so fast? The idea in lazer is that all rulesets should behave roughly the same, which is to say, not make the combo counter go up at insane speeds.

peppy:
@smoogipoo thinking this should default to on, what do you say?

smoogipoo:
I agree, have had it happen to me too.

peppy (approved):

LittleEndu:
Shouldn't there be a special case for missing the very first object (when you have 0x combo)?

frenzibyte:
> Shouldn't there be a special case for missing the very first object (when you have 0x combo)?

Hmm, yeah it doesn't make sense to play a "combo break" sample when there is no combo gained yet.

LittleEndu:
> Hmm, yeah it doesn't make sense to play a "combo break" sample when there is no combo gained yet.

I meant that there's no sound played at all and the first real combo break will still play the sound. So if for first 3 objects you miss, hit, miss them, the second miss would play combo-break even thought you already haven't *full comboed* the song

frenzibyte:
Would've been better if you were clear from the beginning as your comment says there should be a "special" case for missing while you haven't combo, which I would parse as a "behaviour unique than in other scenarios", which in this case, it would mean to me as not play the sample in this exact case.

Either way, I disagree with this for a couple of reasons:
 1. There is no combo break if you haven't really gained combo, so the sample name is misleading on its own.
 2. It would be quite clear to the user that an object has been missed as a `Miss` judgement would be shown right on their screen.
 3. Gameplay would be immediately paused when the first object is near if the user unfocused from the game to look at something else, otherwise their focus should probably be on the game.
 4. Even if (2) and (3) are not strong enough to convince against such behaviour, the user will be able to tell if his current score perfect enough to FC the beatmap when there is support for [combo highlighting](https://github.com/ppy/osu/issues/8413) or otherwise. (very easy to implement I believe, just missing design)

PR #8818, opened 2020-04-21
Fix beatmap encoder by enabling untested test cases

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8815

It turns out that the only beatmap that was tested would pass the encoder test.

Several fixes in this PR, as a result of enabling tests for all the beatmaps that were intended to be tested:
1. Some General/Metadata sections shouldn't be written if they're null, such as audio file, unicode titles, tags.
2. Spinners shouldn't have an implicit new combo.
3. Custom sample bank was still using the suffix, which would always be empty. Recently we changed CSS to be applied only to `LegacyHitSampleInfo` and not go through the suffix.
4. Overlapping control points (two timing points at the same time) need to write two lines, one for the timing point and one for the difficulty point.
5. Changes to `TypedListConverter`.
    1. Uses `IList` to fix serialisation of converted hitobjects (would lead to cast exceptions).
    2. Changed all custom properties to be prefixed by `$`, which is forbidden for C# variables to start their names with. This would become a problem later on with taiko, which contains a "type" of its own and would conflict with the "type" provided here. I'm open to suggestions here, but the JSON decoder is not currently used for anything outside testing right now.
6. Making the testing procedure apply the beatmap converter and processor. We want to test encoding from an already-converted beatmap, but we have to apply the processor to both beatmaps otherwise there's differences in combos/etc.


osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -103,15 +105,15 @@ private void handleMetadata(TextWriter writer)
             writer.WriteLine("[Metadata]");
 
             writer.WriteLine(FormattableString.Invariant($"Title: {beatmap.Metadata.Title}"));
-            writer.WriteLine(FormattableString.Invariant($"TitleUnicode: {beatmap.Metadata.TitleUnicode}"));
+            if (beatmap.Metadata.TitleUnicode != null) writer.WriteLine(FormattableString.Invariant($"TitleUnicode: {beatmap.Metadata.TitleUnicode}"));
```
bdach, line 108: Kind of dislike how this reads with the long lines and `if` lines interspersed with lines without them but I don't think splitting the body to a new line looks any better either so ¯\\\_(ツ)_/¯

bdach (commented):
Can't comment on the changes too much besides the technical part, I mostly trust the tests for correctness of the actual logic. Looks mostly sane.

History looks broken tho. I assume this was due to an interactive rebase gone wrong. Rebasing over someone else's commit will make it so the original author is still the author but you're the "committer" which breaks everything as that info goes into commit objects, so hashes are now all different.

osu.Game.Tests/Beatmaps/Formats/LegacyBeatmapEncoderTest.cs
```diff
@@ -1,54 +1,115 @@
...
         }
 
-        private Beatmap decode(string filename, out Beatmap encoded)
+        private void sort(IBeatmap beatmap)
+        {
+            // Sort control points to ensure a sane ordering, as they may be parsed in different orders. This works because each group contains only uniquely-typed control points.
+            foreach (var g in beatmap.ControlPointInfo.Groups)
+            {
+                ArrayList.Adapter((IList)g.ControlPoints).Sort(
+                    Comparer<ControlPoint>.Create((c1, c2) => string.Compare(c1.GetType().ToString(), c2.GetType().ToString(), StringComparison.Ordinal)));
```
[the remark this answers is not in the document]
smoogipoo: I didn't realise this was a thing either. I'd like to see BindableList expose a `Sort()` method, because this is super unsafe and should absolutely never be done. It's in a test though so /shrug ?

osu.Game/IO/Serialization/Converters/TypedListConverter.cs
```diff
@@ -61,7 +62,7 @@ public override object ReadJson(JsonReader reader, Type objectType, object exist
 
         public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
         {
-            var list = (List<T>)value;
+            var list = (IList)value;
```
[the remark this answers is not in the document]
smoogipoo: Class types can't be covariant, only interfaces. So either `IEnumerable<T>` or `IList<T>` would work here.

I've used `IEnumerable<T>` as you've suggested 👍

smoogipoo:
Will rebase + FP to clean up git history, since review has been addressed.

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -58,7 +60,7 @@ private void handleGeneral(TextWriter writer)
         {
             writer.WriteLine("[General]");
 
-            writer.WriteLine(FormattableString.Invariant($"AudioFilename: {Path.GetFileName(beatmap.Metadata.AudioFile)}"));
+            if (beatmap.Metadata.AudioFile != null) writer.WriteLine(FormattableString.Invariant($"AudioFilename: {Path.GetFileName(beatmap.Metadata.AudioFile)}"));
```
peppy, line 63: should this not be `virtual` instead?

smoogipoo: Can do, would need to also make it by default inside `BeatmapMetadata` too. Sound fine?

peppy: hmm... not sure about that (do we want to use it going forward?)

smoogipoo: Seems like stable isn't setting virtual anymore, and it behaves fine without it.

peppy (approved):

PR #8820, opened 2020-04-21
Fix BeatmapListingOverlay not taking focus

peppy:


smoogipoo (approved):
Feels a bit messy passing this through 3 levels, but let's go with it for now.

Issue #2506, opened 2018-05-09
Mania longnote releases trigger hitsounds


LastExceed:
In stable, this only happens in auto-converts, not in dedicated mania beatmaps. The reason for that is probably that osu!standard sliderends have hitsounds which are kept during the conversion while native mania longnote releases dont. To me this feels inconsistent might thus also be worth looking at in stable.

Anyway, in lazer even native LN releases cause hitsounds, which is different to stable. but before changing anything we should discuss whether we want to keep this the way it is in stable right now:

I personally would suggest keeping the release hitsounds, but make them skinnable - separately to normal hitsounds (maybe the same way combo and score digits can be skinned separately if the user wishes to do so).

Release sounds are something I always missed in mania, as they

- help with timing

- complete the rythm (imagine osu!standard without release sounds on sliderends)

Making them skinnable as proposed above brings 2 more advantages:

- People who don't like them, or just want things to stay the way they're used to can turn them off

- It allows to use a different sound for releases in contrast to normal hitsound, avoiding the confusion that may come up during overlapping longnotes.

AlFasGD:
There are the cases where the LNs' end does not represent a sound, especially when a song has ended and the LN only signifies the remaining sound. Which brings me to another suggestion: giving the mapper the option to use a hit sound for the beginning, the body and the ending of the LN. The body's hitsound could be customizable in various ways (which could be a backlog feature).

revelsix:
Long note releases ruin keysounded maps in lazer, [this is just one example](https://osu.ppy.sh/beatmapsets/930970#mania/1944029) and also ruin any drums that were designed only to trigger on the start of the LN (drums where there are none, drums hitting twice at the same time, etc.). Both players and mappers have gotten used to silent ends.

LastExceed:
@revelsix 
> Long note releases ruin keysounded maps

No they don't. Beatmaps would obviously be able to set release sounds just like they can set hit sounds

> drums that were designed only to trigger on the start of the LN

same as above (also why would you map a pointy sound like drums with LN in the first place?)

> Both players and mappers have gotten used to silent ends

That is a very conservative view. Opposing change just because of habit is short sighted, we should consider that in a few years from now the majority of players won't even know the old client and how it used to work. Also if you really can't make friends with release sounds you can still disable them via skinning

abraker95:
If you can skin them out then it should be fine.

revelsix:
[...]

PR #8823, opened 2020-04-21
Implement legacy beatmap encoding for all rulesets

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/8818

Biggest changes are those to mania since hitobject samples may have changed for converts. It needs to be looked into further anyway, since I'm fairly certain it's not handled correctly anyway:

https://github.com/ppy/osu/issues/8595
https://github.com/ppy/osu/issues/2506


bdach:
I read through this. A few places were a little weird (drumrolls with `IHasCurve`, although I do like that it was an explicit implementation, or the encoder not being able to reference `LEGACY_VELOCITY_MULTIPLIER` due to library separation) but after some thinking I figured this is legacy stuff anyway and doing this "properly" would probably be a headache and a half, so probably best to leave it as-is and deal with it *if* it ever becomes a problem (which it probably won't).

As to actual correctness of the changes I mostly defer to the tests themselves as I'm not very familiar with the osu! file format but it all looked correct as far as I could tell. History looks bad here too though, same as in the other PR.

peppy (approved):
can't see any issues.

PR #8830, opened 2020-04-22
Add static internal catcher calculation methods for CatchDifficultyCalculator

frenzibyte:
Better and more simpler way IMO to calculate the width of the catching area without having to instantiate a whole `Catcher`.


osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +141,26 @@ private void load()
             updateCatcher();
         }
 
+        /// <summary>
+        /// Calculates the scale of the catcher based off the provided beatmap difficulty.
+        /// </summary>
+        internal static Vector2 CalculateScale(BeatmapDifficulty difficulty)
```
peppy, line 147: private?

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +141,26 @@ private void load()
...
+        /// <summary>
+        /// Calculates the width of the area used for attempting catches in gameplay.
+        /// </summary>
+        internal static float CalculateCatchWidth(Vector2 scale)
+            => CatcherArea.CATCHER_SIZE * Math.Abs(scale.X) * allowed_catch_range;
+
+        /// <summary>
+        /// Calculates the width of the area used for attempting catches in gameplay.
+        /// </summary>
+        /// <param name="difficulty"></param>
```
peppy, line 159: missing xmldoc

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +141,26 @@ private void load()
...
+        /// <summary>
+        /// Calculates the scale of the catcher based off the provided beatmap difficulty.
+        /// </summary>
+        internal static Vector2 CalculateScale(BeatmapDifficulty difficulty)
+            => new Vector2(1.0f - 0.7f * (difficulty.CircleSize - 5) / 5);
+
+        /// <summary>
+        /// Calculates the width of the area used for attempting catches in gameplay.
+        /// </summary>
+        internal static float CalculateCatchWidth(Vector2 scale)
```
peppy, line 153: missing param xmldoc, no idea what scale is.

peppy (dismissed):

osu.Game.Rulesets.Catch/UI/Catcher.cs
iiSaLMaN: it's a method for calculating same as the others so wasn't sure about making it private, will do.

peppy: if a method is not used externally (and doesn't carry any foreseeable purpose) then it should be private.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -139,6 +141,26 @@ private void load()
...
+        /// Calculates the scale of the catcher based off the provided beatmap difficulty.
+        /// </summary>
+        private static Vector2 calculateScale(BeatmapDifficulty difficulty)
+            => new Vector2(1.0f - 0.7f * (difficulty.CircleSize - 5) / 5);
+
+        /// <summary>
+        /// Calculates the width of the area used for attempting catches in gameplay.
+        /// </summary>
+        /// <param name="scale">The scale of the catcher.</param>
+        internal static float CalculateCatchWidth(Vector2 scale)
```
smoogipoo, line 154: Overload is probably not required but ¯\\_(ツ)_/¯

smoogipoo (approved):

osu.Game.Rulesets.Catch/UI/Catcher.cs
iiSaLMaN: I think you meant the other one: `CalculateCatchWidth(BeatmapDifficulty)` (which is there for simplifying `Catcher.CalculateCatchWidth(Catcher.CalculateScale(difficulty))`). 

this one is required to be as-is for working with calculating the catch-width without a `BeatmapDifficulty`:
https://github.com/ppy/osu/blob/883788dd5aec1b228328812fb4ea14043c299118/osu.Game.Rulesets.Catch/UI/Catcher.cs#L111

smoogipoo: Nope, I mean't this one. As in I would've been fine just recomputing the scale a second time in all cases.

PR #8834, opened 2020-04-22
Fix TestBeatmap not setting appropriate ruleset ID

smoogipoo:
It sets the ruleset but not the ruleset ID :(


peppy:
is 0 actually what we use as a fallback?

smoogipoo:
It's 0 by default so...

Hold off on this PR for a bit anyway, I want to make sure the editor actually loads so tests fail next time.

peppy:
Need some conflict resolution here.

peppy (approved):

PR #8842, opened 2020-04-23
Make osu! ruleset placements happen on mouse down

smoogipoo:
Definitely feels better. For hitcircles, the position keeps updating while the mouse button hasn't been released, but the placement doesn't show - only the blueprint changes.


peppy:
This feels weird since it doesn't appear on the playfield immediately. I think it'd be better to just do the full placement on down?

smoogipoo:
Done. I think it feels better with full placement on mouse down?

peppy:
Feels fine. Can you apply the same change in https://github.com/ppy/osu/pull/8843?

PR #8846, opened 2020-04-23
Refactor osu!taiko playfield and add background skinning support

### LINKED ISSUE #8844: Taiko icons overlap
The drum icons overlap but only using this skin for an odd reason. https://mega.nz/file/jogFCSjb#U5bJXz3ZLUC30IBqbl133HfobPtY0bccdpgqiVA6Kyc

peppy:
This PR contains a few changes that are too intertwined to easily split up (sorry).

- Refactors `TaikoPlayfield` to be easier to understand.
- Adds custom scale adjust logic to legacy skinnables which need to be rendered at specific sizes (bypassing the otherwise relative context). A bit ugly that we need this in update methods, but local to legacy components and commented as such.
- Removes a weird size update line in `DrawableHit`. This can be split out into another PR if desired. I've tested and this was never actually making a valid change (value is constant).
- Adds `PlayfieldBackgroundRight` and `PlayfieldBackgroundLeft` for skinning

Closes https://github.com/ppy/osu/issues/8844.


osu.Game.Rulesets.Taiko/UI/PlayfieldBackgroundRight.cs
```diff
@@ -0,0 +1,61 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Effects;
+using osu.Framework.Graphics.Shapes;
+using osu.Game.Graphics;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public class PlayfieldBackgroundRight : CompositeDrawable
```
iiSaLMaN, line 15: sounds like this should also be `internal`?

peppy: really doesn't matter much. internal is probably more appropriate though.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -18,169 +15,113 @@
...
+                                {
+                                    RelativeSizeAxes = Axes.Both,
+                                },
+                                new Container
+                                {
+                                    Name = "Hit objects",
+                                    RelativeSizeAxes = Axes.Both,
+                                    Child = HitObjectContainer
+                                },
+                                kiaiExplosionContainer = new Container<KiaiHitExplosion>
```
smoogipoo, line 94: This should not be masked. The kiai glows are supposed to exceed the playfield:

image

image

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -18,169 +15,113 @@
...
-                        },
-                        judgementContainer = new JudgementContainer<DrawableTaikoJudgement>
-                        {
-                            Name = "Judgements",
-                            RelativeSizeAxes = Axes.Y,
-                            Margin = new MarginPadding { Left = HIT_TARGET_OFFSET },
-                            Blending = BlendingParameters.Additive
+                            Children = new Drawable[]
+                            {
+                                barlineContainer = new ProxyContainer
```
smoogipoo, line 84: These also shouldn't be masked as major barlines have triangles that exceed the bounds of the playfield.

smoogipoo (changes requested):

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
peppy: That explains why I couldn't see the triangles..

peppy: Removing masking altogether seems to be fine. Do you know any reason masking was previously required?

smoogipoo: I don't remember, no :/

smoogipoo (approved):

PR #8849, opened 2020-04-23
Rename editor tests namespace from "Editor" to "Editing"

frenzibyte:
Shouldn't be used as it makes using the `Editor` class requires writing the full namespace of it (`Screens.Edit.Editor`)

I was initially going to pick up `Edit` to line up with the namespace used in game `osu.Game.Screens.Edit` but it was a bit short for a visual test, so chose `Editing`.


peppy (approved):
seems in line with "gameplay"

Issue #5685, opened 2019-08-11
taiko gameplay doesnt have the correct scaling for other resolutions


mouzedrift:
**Describe the bug:** 
scaling for taiko seems off and not just a little bit i would say. my native resolution is 1280x1024 and i would never be able to play with such a small playfield. i dont use any extra options to scale the gameplay. (it doesnt really make it better in this case)

**Comparison below:**
osu!stable:
![screenshot276](https://user-images.githubusercontent.com/43358824/62840619-e9ce2780-bc9d-11e9-8b4d-1a2e90574056.png)

osu!lazer:
![osu_2019-08-12_00-12-14](https://user-images.githubusercontent.com/43358824/62840651-8264a780-bc9e-11e9-82ec-175cfc9868a2.png)
i hope i forgot nothing

**osu!lazer version:**
2019.809.0

**Logs:** 
not needed i suppose

peppy:
This is intentional currently to keep the playfield in a constant scale across all aspect ratios. If it was to change, you would have less visibility of hitobjects at square aspect ratios – is that what you want?

mouzedrift:
yes, at least for legacy skins.

ayyEve:
There are many people who play 4:3 because of the different play-field. I believe it would be unfair to them to force this change. I would keep the field consistent between resolutions but change based on aspect ratio (at least between 4:3 and 16:9).

PR #8857, opened 2020-04-24
Add taiko mascot display for legacy skins

Craftplacer:
What is broken / unfinished:
- [x] Layout of `DrawableTaikoMascot` in `TaikoPlayfield`
  - May require consideration about sizing (because state textures can have variable size)
- [x] Testing of Kiai state
- [x] Trigger for clear animation (with the exception of triggers at end of gameplay)
  - [x] Transition from and into clear state

Requires https://github.com/ppy/osu-resources/pull/99 for release


osu.Game.Rulesets.Taiko.Tests/Skinning/TestSceneDrawableTaikoMascot.cs
```diff
@@ -0,0 +1,155 @@
...
+
+            AddStep("Fail state", () => setState(TaikoMascotAnimationState.Fail));
+        }
+
+        private void setState(TaikoMascotAnimationState state)
+        {
+            foreach (var mascot in mascots)
+            {
+                if (mascot == null)
+                    continue;
```
bdach, line 84: Is this even possible? I'm pretty sure `null` should fail the `is DrawableTaikoMascot mascot` check. ([see docs](https://docs.microsoft.com/pl-pl/dotnet/csharp/language-reference/keywords/is#pattern-matching-with-is))

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,105 @@
...
+using osu.Framework.Audio.Track;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Textures;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Graphics.Containers;
+
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public sealed class DrawableTaikoMascot : BeatSyncedContainer
```
bdach, line 14: Any reason why this is `sealed`? Not sealing could actually help in tests here (to which I'll get in another comment).

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,105 @@
...
+        private static TaikoMascotTextureAnimation idleDrawable, clearDrawable, kiaiDrawable, failDrawable;
+        private EffectControlPoint lastEffectControlPoint;
+        private TaikoMascotAnimationState state;
+
+        public Bindable<TaikoMascotAnimationState> PlayfieldState;
+
+        /// <summary>
+        /// Determines if there should be no "state logic", intended for testing.
+        /// </summary>
+        public bool Dumb { get; set; }
```
bdach, line 25: This ideally shouldn't be a thing, much less be named this way and be `public`. I'd make it `protected` at the least, or better yet, make `getFinalAnimationState` `protected virtual`, subclass the mascot in the test scene and override that method to make it a no-op.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,105 @@
...
+        private TaikoMascotTextureAnimation getStateDrawable(TaikoMascotAnimationState state)
+        {
+            return state switch
+            {
+                TaikoMascotAnimationState.Idle => idleDrawable,
+                TaikoMascotAnimationState.Clear => clearDrawable,
+                TaikoMascotAnimationState.Kiai => kiaiDrawable,
+                TaikoMascotAnimationState.Fail => failDrawable,
+                _ => null
+            };
```
bdach, line 65: * We don't use `switch` expressions. Please use a normal `switch` instead.
* Not 100% sold on returning `null` instead of throwing, but I suppose in this case it's defensible.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,105 @@
...
+            InternalChildren = new[]
+            {
+                idleDrawable = new TaikoMascotTextureAnimation(TaikoMascotAnimationState.Idle),
+                clearDrawable = new TaikoMascotTextureAnimation(TaikoMascotAnimationState.Clear),
+                kiaiDrawable = new TaikoMascotTextureAnimation(TaikoMascotAnimationState.Kiai),
+                failDrawable = new TaikoMascotTextureAnimation(TaikoMascotAnimationState.Fail),
+            };
+
+            // making sure we have the correct sprite set
+            State = state;
```
bdach, line 88: This is very weirdly written. I'd say invoking a property's setter with its own backing field as the value, solely to trigger the setter's side effects, is bad practice. The way forward here for me would be:

* to extract the sprite visibility toggling code to a method,
* to remove it from the setter entirely,
* and finally to move the `BindValueChanged` subscription from the ctor to here, and call it with the extracted method and `runOnceImmediately: true` to get the effect you intended to get by doing this hack.

I *think* it's equivalent but I haven't actually gone to check so buyer beware.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,94 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Animations;
+using osu.Framework.Graphics.Textures;
+using osu.Game.Skinning;
+
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public sealed class TaikoMascotTextureAnimation : TextureAnimation
+    {
+        private const float clear_animation_speed = 1000 / 10F;
```
bdach, line 14: nit
```suggestion
        private const float clear_animation_speed = 1000 / 10f;
```

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,94 @@
...
+
+                    if (texture == null)
+                        break;
+
+                    AddFrame(texture);
+                }
+            }
+        }
+
+        /// <summary>Advances the current frame by one.</summary>
```
bdach, line 66: generally `<summary>` xmldocs are not one-liners
```suggestion
        /// <summary>
        /// Advances the current frame by one.
        /// </summary>
```

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,94 @@
...
+        public void Move()
+        {
+            if (FrameCount == 0) // Frames are apparently broken
```
bdach, line 69: mind expounding on that comment? broken in what way? how's that comment meant to help anyone reading this source even now, let alone in a few months?

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,94 @@
...
+                currentFrame = 0;
+
+            GotoFrame(currentFrame);
+
+            currentFrame += 1;
+        }
+
+        private string _getStateTextureName(int i) => $"pippidon{_getStateString(State)}{i}";
+
+        private string _getStateString(TaikoMascotAnimationState state)
```
bdach, line 82: By convention in this repository, private method names don't start with underscores.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -255,16 +267,30 @@ internal void OnNewResult(DrawableHitObject judgedObject, JudgementResult result
...
 
-        private class ProxyContainer : LifetimeManagementContainer
-        {
-            public new MarginPadding Padding
+            if (characterDrawable.Drawable is DrawableTaikoMascot mascot)
             {
-                set => base.Padding = value;
+                var isFailing = result.Type == HitResult.Miss;
+
+                // Only take combo in consideration when it's not a strong hit (it's always false)
```
bdach, line 275: what's always false? I don't understand what this interjection is trying to say.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -255,16 +267,30 @@ internal void OnNewResult(DrawableHitObject judgedObject, JudgementResult result
...
+                        isFailing = result.Judgement.AffectsCombo;
+                }
+
+                mascot.PlayfieldState.Value = isFailing ? TaikoMascotAnimationState.Fail : TaikoMascotAnimationState.Idle;
             }
+        }
+    }
 
-            public void Add(Drawable proxy) => AddInternal(proxy);
+    internal class ProxyContainer : LifetimeManagementContainer
```
bdach, line 287: why did this become `internal`? don't see any other new references to this class in this PR.

bdach (changes requested):
Haven't ran the source, but there are quite a few stylistic issues with the code alone.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
Craftplacer: peppy explained to me that strong hits have their AffectsCombo return `false` because they need multiple hits (IIRC), that's why I'm exempting them to the next check.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
Craftplacer: My analyzer script was outdated or something, I'll revert this line back.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
Craftplacer: Code would crash if the base `TextureAnimation` doesn't contain any frames, this is to prevent the code to fail and instead to just show nothing. (I encountered this while writing tests)

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
Craftplacer: Copy pasta code taken from peppy 😄

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
bdach: If this has to do with `AffectsCombo`, why not check that directly? It's available through `JudgementResult.Judgement`, and avoids the need to keep track whether that's true or false for particular hitobject types.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
bdach: The question here is whether the frame count being 0 constitutes a failure scenario (in which case it should die and it's evidence of test shortcomings) or it can sometimes happen in reasonably normal scenarios, in which case this is a sane guard and the comment can be safely removed (or replaced by a comment that explains precisely *when* an animation can have no frames).

Whether it's one or the other it's a question I or other reviewers will need to have an answer to before being able to decide with confidence to merge this code.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
Craftplacer: > and finally to move the BindValueChanged subscription from the ctor to here, and call it with the extracted method and runOnceImmediately: true to get the effect you intended to get by doing this hack.

The last part seems to me you're mixing up intentions of my code (which I can't blame you for obviously).
The PlayfieldState bindable and State property have different functions.
PlayfieldState is used by the taiko playfield to set current information on whether the input was judged good or bad.
State itself was for setting the appearance of DrawableTaikoMascot, but even that was meant to be set by the `DrawableTaikoMascot` itself (see `OnNewBeat`) (reason why it is public is that it's accessible for testing).

osu.Game.Rulesets.Taiko.Tests/Skinning/TestSceneDrawableTaikoMascot.cs
Craftplacer: To be honest, I'm surprised by this myself. I have no idea how Drawables can turn themselves to null.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
bdach: While I did miss that distinction the same conclusion sort of applies - ideally you'd need just one of these and the way forward I proposed earlier sounds like a possible way out.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
Craftplacer: Aren't I'm already using this? Thinking about it I might just remove it but I also know there must be a reason why I included it in the first place.

bdach:
What's the status on this? I see resolved conversations, but no changes that would address the issues mentioned.

Craftplacer:
> What's the status on this? I see resolved conversations, but no changes that would address the issues mentioned.

I fixed those issues on my local changes but I'm still battling with TextureAnimation to work without SkinnableDrawable.

I went sleeping after I decided I wasn't awake enough to debug the night through.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,105 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Textures;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Graphics.Containers;
+
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public sealed class DrawableTaikoMascot : BeatSyncedContainer
+    {
+        private static TaikoMascotTextureAnimation idleDrawable, clearDrawable, kiaiDrawable, failDrawable;
```
iiSaLMaN, line 16: why is this static? now you can't create multiple instances of this component with maintaining all of them together. (there is already an obvious existing case for this)

frenzibyte (changes requested):

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
Craftplacer: I figured out why I added it: strong hits that are only played with only one key have their JudgementResult type set to Miss.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,107 @@
...
+                    return "fail";
+
+                case TaikoMascotAnimationState.Idle:
+                    return "idle";
+
+                case TaikoMascotAnimationState.Kiai:
+                    return "kiai";
+
+                default:
+                    throw new ArgumentException($"There's no case for animation state ${state} available", nameof(state));
```
bdach, line 103:
```suggestion
                    throw new ArgumentOutOfRangeException(nameof(state), $"There's no case for animation state {state} available");
```

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -40,6 +42,10 @@ public class TaikoPlayfield : ScrollingPlayfield
 
         private Container hitTargetOffsetContent;
 
+        private SkinnableDrawable mascotDrawable;
+
+        private Bindable<double> frameTime = new Bindable<double>(100);
```
bdach, line 47: Unused field.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,107 @@
...
+            {
+                foreach (var textureIndex in clear_animation_sequence)
+                {
+                    var textureName = getStateTextureName(textureIndex);
+                    Texture texture = skin.GetTexture(textureName);
+
+                    if (texture == null)
+                        break;
+
+                    AddFrame(texture);
```
bdach, line 49: This snippet repeats twice in both branches and could be a method.

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
```diff
@@ -0,0 +1,107 @@
...
+        /// <summary>
+        /// Advances the current frame by one.
+        /// </summary>
+        public void Move()
+        {
+            // Check whether there are frames before causing a crash.
+            if (FrameCount == 0)
+                return;
+
+            if (FrameCount <= currentFrame)
```
bdach, line 76: This check reads a bit backwards - I'd expect the thing being bound-checked to be on the left side of the expression (now I have to mentally invert to make sure it's correct). You could also remove this guard entirely and incorporate modulo arithmetic when incrementing `currentFrame` at the end.

bdach (dismissed):

osu.Game.Rulesets.Taiko/UI/TaikoMascotTextureAnimation.cs
Craftplacer: Would this change be fine then?
```diff
- if (FrameCount <= currentFrame)
+ if (currentFrame >= FrameCount)
```

bdach: That'd work too I suppose, yes.

osu.Game.Rulesets.Taiko/UI/DrawableTaikoMascot.cs
```diff
@@ -0,0 +1,106 @@
...
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public class DrawableTaikoMascot : BeatSyncedContainer
+    {
+        private TaikoMascotTextureAnimation idleDrawable, clearDrawable, kiaiDrawable, failDrawable;
+        private EffectControlPoint lastEffectControlPoint;
+
+        public Bindable<TaikoMascotAnimationState> PlayfieldState;
+
+        public TaikoMascotAnimationState State { get; private set; }
```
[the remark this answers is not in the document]
Craftplacer: I turned `PlayfieldState` into a private field now, while making another method called `SetPlayfieldState(state)` (therefore making this variable set only).
I also annotated the method (with a <remarks> section, mentioning the `ShowState` method) to *really* make sure, that this isn't for reading or setting the end product state.

I also read through the C# docs on `protected`, when applying your suggestion on this I noticed I couldn't use this one, because I'm accessing **`DrawableTaikoMascot`**`.State` on the parts where I have no control over whether it's the test version (`TestDrawableTaikoMascot`) or not.

I hope that should be enough for anybody else reading this code. Do you have any other thoughts on this?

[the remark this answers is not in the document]
bdach: The solution you're describing kind of sounds like a Java-style method setter instead of an actual C# property setter. I can't comment on anything else without seeing the code.

I'm also not sure why you think `protected` is unviable. You need it for reading in the test scenes, so you might as well scope down types there from `DrawableTaikoMascot` to the test-specific one, or if *really* necessary, cast (but I think it shouldn't be necessary).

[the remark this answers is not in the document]
Craftplacer: I'll write what I'm observing and thinking here:
If you take a look at [here](https://github.com/ppy/osu/pull/8857/files#diff-e27de7f82e0e1bbfbecdaf785487aaacR102), you see that I have instances of the drawable from the `mascots` list (where they are being added [here](https://github.com/ppy/osu/pull/8857/files#diff-e27de7f82e0e1bbfbecdaf785487aaacR196)).

Now the problem is that the ~~Test~~DrawableTaikoMascot (in this case) is being provided from [TaikoLegacySkinTransformer](https://github.com/ppy/osu/pull/8857/files#diff-64da4d88b2512d381da640cd4b621b6cR88) (then given over to SkinnableDrawable then over to TaikoPlayfield and then finally to my tests and osu!taiko).

So in order to get a testing variant of the drawable with the easiest way possible (that I can come up with right now) I would have to add another value to [TaikoSkinComponents](https://github.com/ppy/osu/pull/8857/files#diff-052d44fe92731567f89e1c39414d6d53R18), and handle that also in TaikoLegacySkinTransformer.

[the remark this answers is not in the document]
bdach: The `.ChildrenOfType<T>()` extension method should come in useful here. Check out other usages of that method in test code.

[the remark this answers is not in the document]
Craftplacer: In what way? I'm already using it in `collectMascots()` and `collectPlayfields()`.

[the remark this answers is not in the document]
bdach:
```diff
-                var mascot = playfield.ChildrenOfType<DrawableTaikoMascot>()
-                                      .SingleOrDefault();
+                var mascot = playfield.ChildrenOfType<TestDrawableTaikoMascot>()
+                                      .SingleOrDefault();
```
no? (Leaving aside the fact that you could probably entirely skip iterating over playfields.)

But fine, if you insist, I can do a pass over this myself in the coming days.

bdach:
I have pushed in some commits here, but they do not constitute a full pass over this PR and I don't consider this to be in mergeable state yet. I only managed to find time to address the quality of test code and merge in changes from `master`, resolving conflicts on the way. The interface surface of both `DrawableTaikoMascot` and `TaikoMascotTextureAnimation` both need work still.

@peppy Unless you'd like to take over straightening this out, I'll be continuing work on this tomorrow. I'd appreciate some pointers on what the positioning of the mascot in stable is if you could provide some - I can leave the final touchups to you though, if you'd prefer.

peppy:
Huge thanks. I have some other pieces of taiko lined up so feel free to continue.

The stable code is [here](https://gist.github.com/peppy/620687ff7adb9c61c3f90f1e5867ed05) for reference. The positioning is 25 "stable" pixels above the taiko bar. But you're probably best just ballparking it using screenshots.

bdach:
@Craftplacer In your latest commits here [you've added yet another unrelated rider configuration file](https://github.com/Craftplacer/osu/blob/c61584bc2f9c64a1b4f29efe04838c97c79a3284/.idea/.idea.osu.Desktop/.idea/discord.xml).

Please check the list of changed files before committing/pushing. I will revert this change myself once I continue work on this PR.

bdach:
Another batch of commits is in, but this is still not done (I shot myself in the foot in a stupid way and spent like an hour on debugging how). I'd say it's *close*, as the only things left to do are to actually trigger going into the `clear` state and figure out layouting of the mascot, so I don't foresee changes to the overall structure myself unless you guys have better ideas on how this should look.

I ended up opting to split the logic of state transition itself from the logic of animation playback as much as possible (unfortunately the `clear` case requires tracking whether that animation's finished). Overall `DrawableTaikoMascot` is now a pseudo-state-machine, and the actual animations were wrapped in `BeatSyncedContainer`s to keep that on-beat animation logic hidden from state transitions. Looks maybe a little awkward having multiple `BeatSyncedContainer`s side-by-side but I kind of like how it looks now. It cleaned up test code quite a lot too. The animation code de-branched quite a lot too which is always nice, but the `IsPlaying` usage might be considered a bit hacky.

@peppy One more thing I'd like to confirm in regards to stable - as far as I can see the `clear` animation is triggered:

* on every 50 combo gained,
* upon successful completion of a swell,
* sometimes at the end of the map (?)

Could you confirm whether that's the case? I was mostly guessing and I couldn't find docs on that (just a vague mention that it has to do with combo, upon which I figured out the "every 50" part myself empirically).

Craftplacer:
amazing work you did there (I may hope I can pick this up for later PRs).
regarding the clear trigger, I don't think it's supposed to be implemented yet, as I think there should be generalized way for rulesets when a combo milestone has been hit, but I don't know how peppy sees this though.

by playing some taiko for a short while, it seems like that the mascot "clears" when you're passing the map.

osu.Game.Rulesets.Taiko/UI/TaikoMascotAnimation.cs
```diff
@@ -0,0 +1,127 @@
...
+        {
+            InternalChild = textureAnimation = createTextureAnimation(state).With(animation =>
+            {
+                animation.Origin = animation.Anchor = Anchor.BottomLeft;
+                RelativeSizeAxes = Axes.Both;
+            });
+
+            RelativeSizeAxes = Axes.Both;
+            Origin = Anchor = Anchor.BottomLeft;
+            AlwaysPresent = true;
```
bdach, line 32: just so I don't forget, this is leftover from debugging attempts and should be removable safely

bdach: I was super wrong - this is necessary for now, as `AnimationClockComposite` will consume time spent when not present otherwise, making `PlaybackPosition` jump forward in time.

bdach:
Day 3, and this is now done...ish? I feel that it's at a reviewable state, and I feel *okay* with how the code turned out in the end. The latest commits add triggers for the clear animation from combo and from hitting a swell, but do *not* add triggers for fail/clear at the end of the map, as I don't see a pre-existing clean way to access the `HealthProcessor` from ruleset drawables, and I feel making that possible is somewhat not fit for this PR and this one has run its course already.

The visual positioning has also been slightly improved. Looks like this right now:



Not perfect, but feels close enough. The reference points chosen to align are highlighted with the vertical & horizontal lines.

Would appreciate a review, especially when it comes to the gameplay logic parts, as I was mostly reverse engineering by deduction.

smoogipoo:
Perhaps this can now be updated to make use of the new `GameplayBeatmap` members added in https://github.com/ppy/osu/pull/8920 ?

bdach:
Yep, that was the plan. I'll get on it the next opportunity I get (hopefully today).

bdach:
Done - cleaned up code slightly due to being able to eliminate the field in `TaikoPlayfield`.

peppy:
@Craftplacer where were the `old-skin` resources sourced from? Generally those are supposed to come from the 2007 skin for testing very old versions, but I can't confirm that is the case here (they look to just be tinted versions of the standard pippidon?)

Craftplacer:
> @Craftplacer where were the `old-skin` resources sourced from? Generally those are supposed to come from the 2007 skin for testing very old versions, but I can't confirm that is the case here (they look to just be tinted versions of the standard pippidon?)

Yes it's a tinted version, if you want you can replace them.

To clarify, I didn't exactly know what skin folder was for what.

peppy:
Scaling seems a bit off. Not sure if it should/can be fixed easily in this pr...

bdach:
Yeah the positioning isn't very easy to get right. The biggest problem is the frames potentially having different sizes and being anchored bottom-right. The sprite animation also has a 0x0 size which doesn't help.

Last thing I tried was calculating max frame size and hacking around with `Scale` but it didn't work very well. I forget why exactly but I could try again today.

peppy:
Taiko scaling as a whole is a bit weird, and may be changing via #5685. For now, transferring the playfield's scale to the mascot may be the best direction.

bdach:
Not sure if it is what you intended to suggest but I have it working [pretty alright](https://streamable.com/o29e9u) with the latest commit, so thanks for the suggestion! Went for a `LayoutValue` too since free optimisations.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -40,9 +43,14 @@ public class TaikoPlayfield : ScrollingPlayfield
 
         private Container hitTargetOffsetContent;
 
+        private readonly LayoutValue<float> playfieldScaleLayout = new LayoutValue<float>(Invalidation.DrawSize);
+        private float playfieldScale => playfieldScaleLayout.IsValid ? playfieldScaleLayout.Value : playfieldScaleLayout.Value = DrawHeight / DEFAULT_HEIGHT;
```
peppy, line 47: is this really necessary for just saving on one divide? seems like way too much unnecessary complexity unless i'm missing something.

bdach: Maybe was a little overeager to use that.

peppy:
Fixed an issue with rewinding while the "clear" animation was playing, removed test resources that didn't make sense. Unfortunately they are stuck in git history because rebasing this mess of a branch is not worth the effort.

peppy (approved):

Issue #8854, opened 2020-04-23
Add a background video for the Showcase Scene on the lazer tournament client.


MiraiSubject:
**Describe the new feature:** 
This feature would add a background video to the lazer tournament client in the same way it does currently to gameplay, schedule etc. 

**Proposal designs of the feature:** 

Right now I have an implementation in my own fork of osu! lazer that creates an Array of drawables with the TournamentLogo and TourneyVideo.
This is the before situation:
![osu!_2020-04-23_20-23-12](https://user-images.githubusercontent.com/803255/80135254-46f5d900-85a0-11ea-954d-540179d48175.png)

This is what it looks like with the feature implemented: 
![osu!_2020-04-23_20-11-33](https://user-images.githubusercontent.com/803255/80134210-b23eab80-859e-11ea-9d7d-ba7647846dde.png)

With chroma key implemented: 
![image](https://user-images.githubusercontent.com/803255/80139576-f46beb00-85a6-11ea-9996-a1a273fde6d0.png)

In this design the file is handled the same as all the TourneyVideos that are placed in "tournament/videos". The file in the directory will be called showcase.m4v. 

This is what it will looks like in a streaming program (stable runs here at a custom resolution of 1920x980 to completely fill the space):
![obs64_2020-04-23_20-53-43](https://user-images.githubusercontent.com/803255/80138084-8e7e6400-85a4-11ea-95d0-bf9d5b1807bb.jpg)

A possible side effect of this is that the user could get forced to offline mode when using this.

I've made a basic WIP implementation over here on a separate branch: https://github.com/MiraiSubject/osu/tree/showcase-video

peppy:
If you already have a fork, why not PR the change following the way other screens do it?

PR #8859, opened 2020-04-24
Add background video for the Showcase Scene to the tournament client

MiraiSubject:
From https://github.com/ppy/osu/issues/8854.

This adds a background video and chromakeying to the Showcase Scene in the Lazer tournament client. 

I've considered adding height and width adjustments to the chromakeying in this scene, in the event that organizers want to utilise the showcase video area for other elements through their streaming software than just a background. But I'd like to hear your opinion first about this. 

Here are two screenshots of the end result:


bdach:
@MiraiSubject Thanks for the contribution! I've applied a few touch-ups, mostly to avoid using the hard-coded screen dimensions by using the `SongBar`'s height as padding to make sure the song bar has enough space to be shown properly. This way if the height ever changes the chroma key area won't have to be adjusted.

peppy:
Because we don't use this in official capacity, the IProvideVideo interface shouldn't be added for now (see `ScheduleScreen` for another example like this). This would cause the "generic" video to be hidden.

Eventually we'll want better logic for this (to detect whether a local video is defined, else use generic) but for the time being we just stack them on top of each other.

I'll make this change. Looks good otherwise 👍 .

MiraiSubject:
Thanks for the feedback! I'll be using this information in the future if I find something else to contribute to!

PR #8860, opened 2020-04-25
Fix score multiplier being cut off in mod select at higher ui scales

Joehuu:
Notable changes:
- ~Score multiplier number is now fixed width.~
	- ~Won't unnecessarily flow to next row when enabling/disabling mods. Also needed to have more width than the unranked label for layout.~
- ~Unranked label is now under the score multiplier number with a transition when appearing/disappearing.~
	- ~" unranked mods.~
	8kiB1syLWP
- ~Score multiplier number and unranked label text size have changed, 25 and 15 respectively, to add up to the button size of 40.~
	- ~Or it'll autosize the footer when the unranked label is visible.~

~640x480 at 1.60x (max) scale~
image

### DIFF
--- a/osu.Game/Overlays/Mods/ModSelectOverlay.cs
+++ b/osu.Game/Overlays/Mods/ModSelectOverlay.cs
@@ -272,7 +272,7 @@ namespace osu.Game.Overlays.Mods
                                                         Font = OsuFont.GetFont(size: 30, weight: FontWeight.Bold),
                                                         Origin = Anchor.CentreLeft,
                                                         Anchor = Anchor.CentreLeft,
-                                                        Width = 70, // to avoid footer from flowing when clicking mods
+                                                        Width = 70, // to prevent footer from flowing when clicking mods
                                                     },
                                                 },
                                             },
--- a/osu.Game/Overlays/Mods/ModSelectOverlay.cs
+++ b/osu.Game/Overlays/Mods/ModSelectOverlay.cs
@@ -219,6 +219,8 @@ namespace osu.Game.Overlays.Mods
                                         RelativeSizeAxes = Axes.X,
                                         Width = content_width,
                                         Spacing = new Vector2(footer_button_spacing, footer_button_spacing / 2),
+                                        LayoutDuration = 100,
+                                        LayoutEasing = Easing.OutQuint,
                                         Padding = new MarginPadding
                                         {
                                             Vertical = 15,
--- a/osu.Game/Overlays/Mods/ModSelectOverlay.cs
+++ b/osu.Game/Overlays/Mods/ModSelectOverlay.cs
@@ -268,18 +268,30 @@ namespace osu.Game.Overlays.Mods
                                                         Origin = Anchor.CentreLeft,
                                                         Anchor = Anchor.CentreLeft,
                                                     },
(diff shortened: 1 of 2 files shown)


peppy:
- fixed with looks bad (too spaced).
- unranked being moved feels weird. please don't make design changes in this PR.

Joehuu:
Those changes are intentional to avoid width changing and flowing to next row when changing mods.

*Trying to pick cinema (edge case where width is just about to flow to next row). Removed multiplier's fixed width here, but applies to old unranked label design as well.*

CfnSA3prhr

Unless there is a way to fix this ux, I propose we move the multiplier stuff somewhere else.

peppy:
Ah, I see. Let's just remove the "unranked" display completely for now. It's not really relevant to lazer anyways.

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -228,60 +232,50 @@ public ModSelectOverlay()
...
+                                                        Font = OsuFont.GetFont(size: 30),
+                                                        Origin = Anchor.CentreLeft,
+                                                        Anchor = Anchor.CentreLeft,
+                                                    },
+                                                    MultiplierLabel = new OsuSpriteText
+                                                    {
+                                                        Font = OsuFont.GetFont(size: 30, weight: FontWeight.Bold),
+                                                        Origin = Anchor.CentreLeft,
+                                                        Anchor = Anchor.CentreLeft,
+                                                        Width = 70, // to avoid footer from flowing when clicking mods
```
bdach, line 275: for slightly better English:
```suggestion
                                                        Width = 70, // to prevent footer from flowing when clicking mods
```

bdach (commented):
Visually looks good even on nearly square windows - I could not find an aspect ratio that would break this 👍

bdach (dismissed):

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -228,60 +232,50 @@ public ModSelectOverlay()
...
+                                                        Font = OsuFont.GetFont(size: 30),
+                                                        Origin = Anchor.CentreLeft,
+                                                        Anchor = Anchor.CentreLeft,
+                                                    },
+                                                    MultiplierLabel = new OsuSpriteText
+                                                    {
+                                                        Font = OsuFont.GetFont(size: 30, weight: FontWeight.Bold),
+                                                        Origin = Anchor.CentreLeft,
+                                                        Anchor = Anchor.CentreLeft,
+                                                        Width = 70, // to prevent footer from flowing when clicking mods
```
peppy, line 275: can you explain the scenario where this happens? i can't repro.

Joehuu: - Change UI scale just before the footer flows.
- Click on a multiplier decrease mod.
vKCNf3ACmb

This is because the multiplier number can increase/decrease width based on the numbers.

peppy: Aha, I see. Will reword the comment slightly.

peppy (approved):

PR #8862, opened 2020-04-25
Fix null reference while hiding toolbar on tests

frenzibyte:
Found while fixing something else relevant to here

### DIFF
--- a/osu.Game/Overlays/Toolbar/Toolbar.cs
+++ b/osu.Game/Overlays/Toolbar/Toolbar.cs
@@ -148,7 +148,7 @@ namespace osu.Game.Overlays.Toolbar
 
         protected override void PopOut()
         {
-            userButton?.StateContainer.Hide();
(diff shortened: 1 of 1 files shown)


peppy:
Can you explain how this happens?

bdach:
Also a test would be nice, if it's a scenario reproducible under tests.

iiSaLMaN:
> Can you explain how this happens?

Hide the toolbar in test scenes.
 - `userButton` is never null on `PopOut`as it is only invoked on `LoadComplete` are after so I removed the null conditional for it, 
 - `userButton.StateContainer` is null as the button resolves the `LoginOverlay` which may or may not be null

> Also a test would be nice, if it's a scenario reproducible under tests.

There will be a test case I'm preparing for another PR that hides the toolbar for testing something.

peppy:
A better title would be "Fix null reference while hiding toolbar in tests".

PR #8863, opened 2020-04-25
Fix ruleset selector not receiving key input on toolbar absence

### LINKED ISSUE #8861: Can't change ruleset with shortcuts while toolbar is hidden
**Describe the bug:** Doesn't work when pressing <kbd>CTRL</kbd> <kbd>1-9</kbd>.

**osu!lazer version:** 2020.421.0

frenzibyte:
Closes #8861 

- [x] Depends on #8862 (for null ref)

Fixes toolbar ruleset selector not receiving key input while the toolbar is not present / hidden off-screen.

### DIFF
--- a/osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
+++ b/osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
@@ -5,13 +5,17 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
+using osu.Framework.Allocation;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Testing;
 using osu.Game.Overlays.Toolbar;
+using osu.Game.Rulesets;
+using osuTK.Input;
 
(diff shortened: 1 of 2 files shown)


osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
```diff
@@ -21,24 +25,62 @@ public class TestSceneToolbar : OsuTestScene
...
+        public void TestRulesetSwitchingShortcut(bool toolbarHidden)
+        {
+            ToolbarRulesetSelector rulesetSelector = null;
+
+            if (toolbarHidden)
+                AddStep("hide toolbar", () => toolbar.Hide());
+
+            AddStep("retrieve ruleset selector", () => rulesetSelector = toolbar.ChildrenOfType<ToolbarRulesetSelector>().Single());
+
+            for (int i = 0; i < 4; i++)
```
iiSaLMaN, line 68: it would've been better to check on the count of available rulesets but I don't think I can

PR #8864, opened 2020-04-25
Fix custom rulesets not showing correct icons in key binding settings

Game4all:
# Summary
This PR allows setting an actual drawable constructed in `SettingsSection.CreateIcon()` as an icon for the Settings Sidebar button; which allows in-turn displaying the custom's ruleset icon in the related keybindings section:
image

# Remarks
No tests for this as I don't think they are required for such a small feature.

### DIFF
--- a/osu.Game/Overlays/KeyBinding/GlobalKeyBindingsSection.cs
+++ b/osu.Game/Overlays/KeyBinding/GlobalKeyBindingsSection.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Graphics;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Input.Bindings;
 using osu.Game.Overlays.Settings;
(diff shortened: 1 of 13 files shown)


osu.Game/Overlays/KeyBinding/GlobalKeyBindingsSection.cs
```diff
@@ -9,7 +10,12 @@ namespace osu.Game.Overlays.KeyBinding
 {
     public class GlobalKeyBindingsSection : SettingsSection
     {
-        public override IconUsage Icon => FontAwesome.Solid.Globe;
+        public override Drawable CreateIcon() => new SpriteIcon
+        {
+            RelativeSizeAxes = Axes.Both,
```
bdach, line 15: I kind of dislike how the `RelativeSizeAxes` specification is repeated in every usage migrated from `SpriteIcon`. Removing it entirely (due to usage of `ConstrainedIconContainer`) or directly setting relative axes on the actual call in `SidebarButton` both work fine, so maybe these should just be removed in favour of either?

bdach: You seem to have left a couple of in, but I think I got them all now.

bdach (approved):
Looks good 👍

peppy (approved):

PR #8872, opened 2020-04-26
Hide red tint when "Show health display even when you can't fail" option is on

### LINKED ISSUE #8789: red tint layer displays even though no fail is enabled
Describe the bug: the red tint layer displays when failing even though no fail is enabled. doesnt happen to stable though. feel free to close if intentional

stable w/ no fail
screenshot059

stable w/o no fail
screenshot060

lazer w/ no fail (std)
osu_2020-04-18_15-34-09

Lazer w/ no fail (mania)
osu_2020-04-18_15-37-42

osu!lazer version: 2020.418.0

Logs: not needed???

vruss:
Resolves #8789

Implemented based on the "Show health display even when you can't fail" option as suggested by @bdach

### DIFF
--- a/osu.Game/Screens/Play/HUDOverlay.cs
+++ b/osu.Game/Screens/Play/HUDOverlay.cs
@@ -168,11 +168,13 @@ namespace osu.Game.Screens.Play
                 if (healthBar.NewValue)
                 {
                     HealthDisplay.FadeIn(fade_duration, fade_easing);
+                    FailingLayer.FadeIn(fade_duration, fade_easing);
                     topScoreContainer.MoveToY(30, fade_duration, fade_easing);
                 }
                 else
                 {
                     HealthDisplay.FadeOut(fade_duration, fade_easing);
+                    FailingLayer.FadeOut(fade_duration, fade_easing);
                     topScoreContainer.MoveToY(0, fade_duration, fade_easing);
                 }
             }, true);


bdach:
Unfortunately this implementation is incorrect. The visibility of the red tint layer is already affected by another setting - "Fade playfield to red when health is low", which is read and interpreted in `FailingLayer` itself.

If you start a map with No Fail and both options disabled, and then just enable "Show health display even when you can't fail", the red tint layer will become visible, when it should still remain invisible as it is still affected by the other setting ([video](https://streamable.com/h9hgmb)). This is due to the fact that both settings now function separately instead of in conjunction.

If you decide to go ahead and fix this problem, please also include tests for this behaviour.

osu.Game/Screens/Play/HUDOverlay.cs
```diff
@@ -168,11 +168,13 @@ protected override void LoadComplete()
                 if (healthBar.NewValue)
                 {
                     HealthDisplay.FadeIn(fade_duration, fade_easing);
+                    FailingLayer.FadeIn(fade_duration, fade_easing);
```
bdach, line 171: This should be inside of `FailingLayer` itself to have all the visibility/setting logic in one place.

bdach (changes requested):
As above.

vruss:
I'm closing this PR as I'm not sure about the expected behaviour and I don't feel confident I'll be able to implement it. Thanks for the feedback!

PR #8877, opened 2020-04-27
Add taiko hit explosion skinning support

peppy:
Open for suggestion to clean up the enum weirdness, but it's quite local for now. I made the decision to do it this way so because we use the non-taiko ones to display just above the playfield (for now, not sure how permanent).

image

### DIFF
--- a/osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
+++ b/osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using osu.Framework.Audio.Sample;
 using osu.Framework.Bindables;
@@ -86,7 +87,7 @@ namespace osu.Game.Rulesets.Taiko.Skinning
                 case TaikoSkinComponents.TaikoExplosionGreat:
                 case TaikoSkinComponents.TaikoExplosionMiss:
 
-                    var sprite = this.GetAnimation(getHitname(taikoComponent.Component), true, false);
+                    var sprite = this.GetAnimation(getHitName(taikoComponent.Component), true, false);
                     if (sprite != null)
                         return new LegacyHitExplosion(sprite);
 
@@ -96,7 +97,7 @@ namespace osu.Game.Rulesets.Taiko.Skinning
             return source.GetDrawableComponent(component);
         }
 
-        private string getHitname(TaikoSkinComponents component)
+        private string getHitName(TaikoSkinComponents component)
         {
             switch (component)
             {
@@ -110,7 +111,7 @@ namespace osu.Game.Rulesets.Taiko.Skinning
                     return "taiko-hit300";
             }
 
-            return string.Empty;
+            throw new ArgumentOutOfRangeException(nameof(component), "Invalid result type");
         }
 
         public Texture GetTexture(string componentName) => source.GetTexture(componentName);
--- a/osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
+++ b/osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
@@ -81,9 +81,36 @@ namespace osu.Game.Rulesets.Taiko.Skinning
                         return new LegacyBarLine();
 
                     return null;
+
+                case TaikoSkinComponents.TaikoExplosionGood:
+                case TaikoSkinComponents.TaikoExplosionGreat:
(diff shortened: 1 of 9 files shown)


bdach:
The changes here inadvertently broke `TestSceneHits` a little bit due to the fact that that scene was playing a little fast-and-loose with its `JudgementResult`s (the hitobjects for the hit steps were never added to the drawable hierarchy, therefore they never ran `load()` to instantiate their hit results).

For restoring full visual feedback passing down the actual hit result to `DrawableTestHit` is also required, so that it can be passed back in `CreateResult()`.

osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
```diff
@@ -81,11 +81,38 @@ public Drawable GetDrawableComponent(ISkinComponent component)
...
+                    if (sprite != null)
+                        return new LegacyHitExplosion(sprite);
+
+                    return null;
             }
 
             return source.GetDrawableComponent(component);
         }
 
+        private string getHitname(TaikoSkinComponents component)
```
bdach, line 99: nit, as usual
```suggestion
        private string getHitName(TaikoSkinComponents component)
```
(+ usages)

osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
```diff
@@ -81,11 +81,38 @@ public Drawable GetDrawableComponent(ISkinComponent component)
...
+                    return "taiko-hit0";
+
+                case TaikoSkinComponents.TaikoExplosionGood:
+                    return "taiko-hit100";
+
+                case TaikoSkinComponents.TaikoExplosionGreat:
+                    return "taiko-hit300";
+            }
+
+            return string.Empty;
```
bdach, line 113: Out of curiosity, is there a reason this doesn't throw the same way `HitExplosion` does?

osu.Game.Rulesets.Taiko/UI/HitExplosion.cs
```diff
@@ -37,35 +34,36 @@ public HitExplosion(DrawableHitObject judgedObject, HitType type)
...
+                    return TaikoSkinComponents.TaikoExplosionMiss;
+
+                case HitResult.Good:
+                    return TaikoSkinComponents.TaikoExplosionGood;
+
+                case HitResult.Great:
+                    return TaikoSkinComponents.TaikoExplosionGreat;
+            }
+
+            throw new ArgumentException("Invalid result type", nameof(resultType));
```
bdach, line 59:
```suggestion
            throw new ArgumentOutOfRangeException(nameof(resultType), "Invalid result type");
```
fits a little better IMO.

bdach (commented):
Seems pretty good otherwise 👍

osu.Game.Rulesets.Taiko/Skinning/TaikoLegacySkinTransformer.cs
peppy: actually meant to change it here, but still not 100% on this whole solution in the first place.

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
```diff
@@ -185,7 +185,6 @@ internal void OnNewResult(DrawableHitObject judgedObject, JudgementResult result
                     var drawableTick = (DrawableDrumRollTick)judgedObject;
 
                     addDrumRollHit(drawableTick);
-                    addExplosion(drawableTick, drawableTick.JudgementType);
```
smoogipoo, line 188: I think we want this back, drum rolls feels empty otherwise (no explosion, no kiai, just bouncing objects...).

smoogipoo (changes requested):

osu.Game.Rulesets.Taiko/UI/TaikoPlayfield.cs
peppy: that's how they are on stable, and how i'd expect them to work.

if we added it back, what would you want? an exception for the default skin?

smoogipoo: Yeah well stage doesn't have the kiai flashes either, are we going to remove those too?

That is to say, yes, I expect the default skin to be an exception. You could have a skinnable "DrumHitExplosion" which the legacy skin would just return an empty drawable for (or null with allowFallback = false).

peppy: hmm, i disagree and felt the previous flashes were an oversight. they are far too big for every tick (this isn't something we do in any other game mode). can we reimplement in a better way, ie. tick explosions, at a later point?

smoogipoo: I guess...

smoogipoo (approved):

PR #8879, opened 2020-04-27
Create test specific select next beatmap function for TestSceneBeatmapCarousel

LittleEndu:
Alternative to and closes #8729

Not to copy the description from that PR... There's a nullref that will crash visual browser when you use ``TestSceneBeatmapCarousel``. When you initially create the carousel, it doesn't have a selection, when you then use <kbd>Up</kbd> or <kbd>Down</kbd>, current selection would be used to select next difficulty.

This PR disables that behaviour while testing. From what I can see, the fact that carousel doesn't have initial selection is actually relied on by some tests (namely those that want to choose what type of initial selection is needed, like testing recommendation function would mean that function needs to be set before initial selection) so you couldn't change actual behaviour to always have initial selection.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -112,6 +112,17 @@ namespace osu.Game.Tests.Visual.SongSelect
             waitForSelection(set_count, 3);
         }
 
+        [Test]
+        public void TestNoTraversalWithoutInitialSelection()
+        {
+            // This is testing behaviour when carousel root doesn't do eager selection. This is only happens in testing.
+
+            loadBeatmaps();
+            checkNoSelection();
+            advanceSelection(direction: 1, diff: true);
+            checkNoSelection();
+        }
+
         [TestCase(true)]
         [TestCase(false)]
         public void TestTraversalBeyondVisible(bool forwards)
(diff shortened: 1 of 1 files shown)


osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -112,6 +112,17 @@ public void TestTraversal()
...
+        [Test]
+        public void TestNoTraversalWithoutInitialSelection()
+        {
+            // This is testing behaviour when carousel root doesn't do eager selection. This is only happens in testing.
+
+            loadBeatmaps();
+            checkNoSelection();
+            advanceSelection(direction: 1, diff: true);
+            checkNoSelection();
+        }
```
bdach, line 124: I'm not sure it makes sense to test behaviour that only happens in the test-specific carousel class. Nobody should rely on this behaviour. I'd just probably leave an inline comment at the new version of `SelectNext()` and call it a day.

LittleEndu: I only left it here because if this test were to fail there would probably be others that would also fail which might make it easier to understand that having initial selection in tests is actually unwanted. But all what I said here could also be a comment somewhere.

bdach: In that case, the tests that require null selection at the start to work should assert that before proceeding with the actual test, therefore making sure this assumption doesn't break.

peppy: I agree with this direction. Rather than adding the new function can asserts just be added at the start of relevant tests?

peppy:
Going to close this for now. I think it makes more sense to add this when we actually run into it.

PR #8883, opened 2020-04-28
Avoid using internal EF methods in tests

peppy:
Is obsoleted in 3.x anyways.


smoogipoo:
Is the entire class obsoleted? Maybe add to `BannedSymbols.txt`: `T:Microsoft.EntityFrameworkCore.Internal.EnumerableExtensions;Don't use internal extension methods.`

bdach (approved):
Looks good 👍 

FWIW I can't see the `BannedSymbols` specs having any effect at all (rider is not showing me squiggles on those extension method calls). Good to have them there for documentation at least I guess, though. Also maybe would have been nice to ban the entire namespace outright but it seems unsupported by `BannedApiAnalyzers`.

peppy:
BannedSymbols only applies at build time (won't show in rider unless you build the project in question).

PR #8887, opened 2020-04-28
Fix speed adjustment mods affecting mania scroll speed

### LINKED ISSUE #8851: Mania doesn't factor DT/HT into scroll speed
Right now when using Double Time / Half Time or any other speed modifier, scroll speed scales with it, making playing maps with double time impossible and with half time rather painful unless you adjust your scroll speed at the start of every map whenever you decide to play with ht/dt which is very inconvenient.

Played a map with nomod
Tried the same map with DT > having to spend time adjusting the scroll speed for it to be playable
Decided to play a different map with nomod/ht > having to spend time adjusting scroll speed back again for it to be playable.

People hop between maps they play with DT/Nomod/HT all the time, so having to adjust the scroll speed every single time is inconvenient and annoying. 

Would be nice having a "Scroll osu!mania scroll speed with BPM" option like in stable, or just not having DT/HT alter the scroll speed.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8851

### DIFF
--- a/osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
+++ b/osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
@@ -2,7 +2,9 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Allocation;
+using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Graphics;
@@ -48,6 +50,10 @@ namespace osu.Game.Rulesets.Mania.UI
         protected new ManiaRulesetConfigManager Config => (ManiaRulesetConfigManager)base.Config;
 
         private readonly Bindable<ManiaScrollingDirection> configDirection = new Bindable<ManiaScrollingDirection>();
+        private readonly Bindable<double> configTimeRange = new Bindable<double>();
+
+        // Stores the current speed adjustment active in gameplay.
+        private readonly Track speedAdjustmentTrack = new TrackVirtual(1000);
 
         public DrawableManiaRuleset(Ruleset ruleset, IBeatmap beatmap, IReadOnlyList<Mod> mods = null)
             : base(ruleset, beatmap, mods)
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
```diff
@@ -48,6 +50,10 @@ public class DrawableManiaRuleset : DrawableScrollingRuleset<ManiaHitObject>
         protected new ManiaRulesetConfigManager Config => (ManiaRulesetConfigManager)base.Config;
 
         private readonly Bindable<ManiaScrollingDirection> configDirection = new Bindable<ManiaScrollingDirection>();
+        private readonly Bindable<double> configTimeRange = new Bindable<double>();
```
bdach, line 53:
```suggestion
        private readonly Bindable<double> configTimeRange = new BindableDouble();
```
The non-generic is a tad unsafe as the underlying setting has a [`Precision` of 1](https://github.com/ppy/osu/blob/e012af661658b6569297435293af1658d381ea29/osu.Game.Rulesets.Mania/Configuration/ManiaRulesetConfigManager.cs#L23).

osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
```diff
@@ -76,7 +85,8 @@ private void load()
             Config.BindWith(ManiaRulesetSetting.ScrollDirection, configDirection);
             configDirection.BindValueChanged(direction => Direction.Value = (ScrollingDirection)direction.NewValue, true);
 
-            Config.BindWith(ManiaRulesetSetting.ScrollTime, TimeRange);
+            Config.BindWith(ManiaRulesetSetting.ScrollTime, configTimeRange);
+            configTimeRange.BindValueChanged(_ => updateTimeRange());
```
bdach, line 89: Gave me a little pause - maybe a bit gratuitous if `updateTimeRange` is being called inside `Update()` anyways...? It won't have any effect as the underlying `BindableDouble` won't fire `ValueChanged` if the value is the same twice in one frame but still reads a little weird.

smoogipoo: Sure. Only did it for consistency with the above, so I'm fine with removing it.

osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
```diff
@@ -48,6 +50,10 @@ public class DrawableManiaRuleset : DrawableScrollingRuleset<ManiaHitObject>
         protected new ManiaRulesetConfigManager Config => (ManiaRulesetConfigManager)base.Config;
 
         private readonly Bindable<ManiaScrollingDirection> configDirection = new Bindable<ManiaScrollingDirection>();
+        private readonly Bindable<double> configTimeRange = new BindableDouble();
+
+        // Stores the current speed adjustment active in gameplay.
+        private readonly Track speedAdjustmentTrack = new TrackVirtual(1000);
```
peppy, line 56: is there a specific reason for the 1000 duration here? i think i'd feel better with 0 otherwise

peppy (approved):

Issue #8885, opened 2020-04-28
taiko hitbursts issues


enoslayd:
**Describe the bug:**  there are two problems, one is that taiko hitburst are much more brighter than in stable (in which sometimes, you cant see the face in the hitbursts) similar to #8853. the other one is that sorry, i cant really explain the problem. it has to do with the animation. just check the video to compare (just remove .zip extension)

**Screenshots or videos showing encountered issue:** 
[stable.mkv.zip](https://github.com/ppy/osu/files/4544724/stable.mkv.zip)
[lazer.mkv.zip](https://github.com/ppy/osu/files/4544760/lazer.mkv.zip)

**osu!lazer version:** commit 0ffd9da

**Logs:**not needed?

peppy:
The animation is intentionally different for now. It should probably be shortened though.

enoslayd:
GDrive???

Yeah, its not related but they have similar issue (brightness???)

enoslayd:
Not to be disrespectful, but the issue regarding the brightness?? (Or alpha??, I dont know the term) of hitbursts is still there, although the animation has already been solved

bdach:
I've already explained why the brightness looks off in [this comment](https://github.com/ppy/osu/issues/8891#issuecomment-620774990). It's due to additive blending of the hit explosions.

peppy:
Good find. That should be moved to the default explosion.

PR #8901, opened 2020-04-30
Move taiko additive blending locally to avoid applying to legacy skins

peppy:
Fixes added "brightness" mentioned in https://github.com/ppy/osu/issues/8885.


osu.Game.Rulesets.Taiko/Skinning/LegacyHitExplosion.cs
```diff
@@ -16,6 +16,8 @@ public LegacyHitExplosion(Drawable sprite)
             Origin = Anchor.Centre;
 
             AutoSizeAxes = Axes.Both;
+
+            Blending = BlendingParameters.Additive;
```
smoogipoo, line 20: Think you meant to put this in the non legacy class.

smoogipoo (changes requested):

smoogipoo (approved):

PR #8903, opened 2020-04-30
Show loading spinner when carousel is not ready to be displayed

### LINKED ISSUE #8865: Screens not loading until music controller is finished loading (which can take a long time)
when i launch lazer it puts me in the game and everything works fine, but i cant open settings or chat and when i go into solo nothing happens just a blank screen, can still go back to the menu once i select solo. if i go into editor it just stays at blank screen with the background but i cant back out unless i try and exit the game then it takes me back to the main menu.

https://streamable.com/q5n6bz

version 2020.418.0 and 2020.421.0

[runtime.log](https://github.com/ppy/osu/files/4533850/runtime.log)
[performance.log](https://github.com/ppy/osu/files/4533874/performance.log)
[debug.log](https://github.com/ppy/osu/files/4533875/debug.log)

peppy:
Allows song select to display and only the carousel to continue loading. Much better UX.

Going to say this closes #8865, as the performance is already quite a lot better. Test using database in #8865 as per usual.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -34,6 +34,7 @@ using System.Linq;
 using System.Threading.Tasks;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Input.Bindings;
+using osu.Game.Graphics.UserInterface;
 using osu.Game.Scoring;
 
 namespace osu.Game.Screens.Select
@@ -92,6 +93,8 @@ namespace osu.Game.Screens.Select
         private SampleChannel sampleChangeDifficulty;
         private SampleChannel sampleChangeBeatmap;
 
+        private Container carouselContainer;
+
         protected BeatmapDetailArea BeatmapDetails { get; private set; }
 
         private readonly Bindable<RulesetInfo> decoupledRuleset = new Bindable<RulesetInfo>();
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -105,9 +108,22 @@ private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, S
...
+            {
+                AllowSelection = false, // delay any selection until our bindables are ready to make a good choice.
+                Anchor = Anchor.CentreRight,
+                Origin = Anchor.CentreRight,
+                RelativeSizeAxes = Axes.Both,
+                BleedTop = FilterControl.HEIGHT,
+                BleedBottom = Footer.HEIGHT,
+                SelectionChanged = updateSelectedBeatmap,
+                BeatmapSetsChanged = carouselBeatmapsLoaded,
+                GetRecommendedBeatmap = (recommender = new DifficultyRecommender()).GetRecommendedBeatmap,
```
smoogipoo, line 121: Bit of a nit: How about moving this construction to the field, and making the field `readonly`? Quite easy to miss that's it's being done in a nested invocation, and then added later on.

smoogipoo (changes requested):

osu.Game/Screens/Select/SongSelect.cs
peppy: sounds good

smoogipoo (approved):

PR #8917, opened 2020-04-30
Add offline status handling to OverlayView

Game4all:
# Summary
This adds offline handling behaviour to OverlayView (as well as tests) in a very similar fashion to #8044, making it an actual replacement of what was proposed in that PR.

I have been tempted to include in this PR changes to DashboardOverlay and existing views in order to leverage the newly introduced behaviour (as for now the overlay will instant trash out the views on user log out, leaving no chance to display any hint) but decided to keep it for a follow up as it may require further discussion; unless it is fine adding it there.

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
@@ -3,7 +3,6 @@
 
 using System.Collections.Generic;
 using System.Linq;
-using System.Threading.Tasks;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
--- a/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
@@ -64,15 +64,6 @@ namespace osu.Game.Tests.Visual.Online
             AddUntilStep("spinner is hidden", () => !view.InternalContents.OfType<LoadingSpinner>().First().IsPresent);
         }
 
-        [Test]
-        public void TestPerformFetch()
-        {
-            AddStep("setup API request handler", () => ((DummyAPIAccess)API).HandleRequest = req => Task.Delay(2000).ContinueWith(t => req.TriggerSuccess()));
-            AddStep("perform fetch request", () => view.PerformFetch());
-            AddUntilStep("spinner is visible", () => view.InternalContents.OfType<LoadingSpinner>().First().IsPresent);
-            AddUntilStep("spinner hid on request completion", () => !view.InternalContents.OfType<LoadingSpinner>().First().IsPresent);
-        }
-
         private class TestOverlayView : OverlayView<DummyData>
         {
             public IEnumerable<Drawable> InternalContents => InternalChildren;
--- a/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
@@ -1,0 +1,123 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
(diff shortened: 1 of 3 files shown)


osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+            }
+
+            private class DummyAPIRequest : APIRequest<DummyData>
+            {
+                protected override string Target => "";
+            }
+        }
+
+        //ReSharper disable once ClassNeverInstantiated.Local
```
bdach, line 105:
```suggestion
        [UsedImplicitly]
```
would also work, although I'm not 100% sure this class is worth having (might as well an existing request if we don't care what it is).

osu.Game/Overlays/OverlayView.cs
```diff
@@ -15,18 +21,52 @@ namespace osu.Game.Overlays
...
+
+        protected override Container<Drawable> Content { get; } = new Container
+        {
+            RelativeSizeAxes = Axes.X,
+            AutoSizeAxes = Axes.Y,
+        };
+
+        private readonly Placeholder currentPlaceholder;
+
+        private readonly BlockingBox blockingBox;
```
bdach, line 42: All these members read *really* messy in this order. Please group them by access modifier and type as per [this ordering](https://github.com/ppy/osu-framework/issues/2278).

osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
Game4all: a dummy class is needed there as `OverlayView<T>` expects a data class to be passed as a generic parameter. I was thinking of creating a tests-wide `DummyAPIData` as well as its corresponding `APIRequest` for cases like this.

osu.Game/Overlays/OverlayView.cs
```diff
@@ -75,5 +132,26 @@ protected override void Dispose(bool isDisposing)
...
+            {
+                switch (e)
+                {
+                    case ScrollEvent _:
+                        return false;
+                }
+
+                return true;
+            }
+        }
```
bdach, line 155: `LoadingLayer` is extremely similar and should be used in favour of this.

osu.Game.Tests/Visual/Online/TestSceneOverlayView.cs
bdach: I know what the thought process was behind adding this. I just think it's unnecessary and you might as well use one of the existing request types instead of creating dummies that do nothing.

bdach (changes requested):
So `OverlayView` already has a user in the dashboard overlay, and I wanted to see how that integrates with the changes here. The answer is - not great, as the state transitions aren't even visible right now (because the overlay rolls its own `LoadingLayer`), and when even that is removed, this doesn't work great at all visually, as can be seen from the screenshot below:

image

which leads me to think that this direction might end up being incorrect in the long run. The big worry I have is the sizing of the dimmed layer - it needs to fill up the overlay space vertically, in its parent's draw space, but the content inside should expand as much as it needs to be and be scrollable from the outside.

Game4all:
Looks like you're suggesting that the logic for this is moved higher in the draw hierarchy to get the layer to dim the overlay content (excluding the header) completely as well as display the placeholder in a centered fashion, which is basically ... what #8044 was supposed to implement.

So I might fallback on that previous closed PR instead but I would like some more input on this before falling back.

bdach:
Well, yes, either that or somehow adjust the layout of `OverlayView` so that it fills the overlay vertically, but is also scrollable (and keep in mind that the "Scroll to top" button exists), but I'm reasonably certain that's not really possible. It definitely shouldn't look as it does in the attached screenshot, because it looks very wrong right now.

Game4all:
I will reopen the aforementionned PR once I'm done with refactoring some things and will probably include in it direct usages in overlays as well. As such, I'm closing this.

Game4all:
While experimenting with the other branch, I thought of giving a default size to the view (about enough to fill the overlay vertically, about 768 pixels) and decided to give it a shot and here's how it looks: 
image


asking @bdach as well as @peppy for thoughts on this before trashing the branch definitely.

bdach:
I'd say having a static height of about 768 is slightly hacky as it's relying on the `DrawSizePreservingFillContainer`'s height... The placeholder seems off-center vertically too.

How does this behave if the content overflows the visible height of the view?

Game4all:
Actually the placeholder centers in the middle of the view content; so if the placeholder is off-centered on the screen that means the content is already overflowing.

bdach:
Yes, which is a problem to begin with as it looks wrong (if we're centering the placeholder vertically it should be centered in its parent's space reserved for overlay content), but what I was actually interested in when I was asking that question is whether the view content scrolls properly if it overflows.

Issue #3504, opened 2018-09-28
Out of bounds text in song select


eenpersoon64:
This is a minor issue, since I don't think many songs have long song names, but once again, text can get out of bounds and cut off, this time in the song select.

![osu_2018-09-28_18-25-36](https://user-images.githubusercontent.com/37511342/46221024-30e45000-c34c-11e8-9472-68895d528865.jpg)

As you can see, the "I" from the word "FUBUKI" got cut off. In my opinion it would look nice if you would add word wrapping of some sorts.

Lolihunter1337:
also applies to the music controller
![song](https://user-images.githubusercontent.com/17437302/46241061-94ed2f80-c3b1-11e8-80dd-85e559f8197c.PNG)

ToIGLAK:
I think text should slide from right to left if it's too long.

arflyte:
Long titles are supposed to scroll, like ye olde Winamp player.

TPGPL:
should unassign?

Joehuu:
Duplicate of #2322.

peppy:
I'm going to keep this one open as it has more details.

Joehuu:
You've closed this accidentally.

Joehuu:
Broadening the scope (what needs to be done):
- [X] Beatmap info wedge
  - [X] Beatmap title/source
  - [X] Beatmap artist
  - [X] Beatmap version
- [ ] Direct panels
  - [ ] Grid panels
  - [ ] List panels (may be hard (text on both sides))
- [ ] Beatmap set overlay
  - [ ] Beatmap title
  - [ ] Beatmap artist

Joehuu:
For the beatmap set overlay, I need to move the `ExternalLinkButton` out of the title as it will be relative size.

The placement should be in a place where it doesn't move as much (title varies in width, so it should be moved anyway). It should be in an obvious place too at low and high scales.

Proposal placement:
At 1.0x scale:
![](https://puu.sh/ECYeE/7c7f0b0754.png)

At max:
![](https://puu.sh/ECYeo/b460e1be14.png)

bdach:
I'd actually consider maybe making an actual button out of it and putting it somewhere next to the favourite/download buttons. There's been users already who didn't know that the icon can be used to open the beatmap in their browser, and after moving it next to the other ones that aren't interactable I'd say it's even less obvious now.

redline2466:
Clipping in results screen:
[...]

PR #8919, opened 2020-05-01
Truncate beatmap title and artist on score panel

### LINKED ISSUE #8315: Song name extends outside of ranking result borders
**Describe the bug:** some long names such as the name of the artist or a song can extend outside of these borders

**Screenshots or videos showing encountered issue:** [Bliitzit - Team Magma & Aqua Leader Battle Theme (Unofficial)](https://osu.ppy.sh/beatmapsets/1002271#osu/2097898):
изображение

**osu!lazer version:** 2020.317.0 (branch https://github.com/ppy/osu/commit/156a3db57d7841c7eb26a2d2ae76ca0a754b281a)

Joehuu:
Fixes https://github.com/ppy/osu/issues/8315 and resolves part of https://github.com/ppy/osu/issues/3504.

Since `SpriteText`'s text can't be aligned when using `RelativeSizeAxes` right now, I am using its `MaxWidth` so truncation can occur even when auto sizing.

image

### DIFF
--- a/osu.Game.Tests/Visual/Ranking/TestSceneExpandedPanelMiddleContent.cs
+++ b/osu.Game.Tests/Visual/Ranking/TestSceneExpandedPanelMiddleContent.cs
@@ -20,6 +20,7 @@ using osu.Game.Rulesets.Osu;
 using osu.Game.Rulesets.Osu.Mods;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Scoring;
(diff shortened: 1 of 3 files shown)


peppy:
Eventually we'll want this to be displayed in a nicer way, but this works for now.

peppy:
Note that this doesn't cover mapper overflow. We'll probably want a more universal solution to this going forward, but let's get this in as a first step.

PR #8920, opened 2020-05-01
Add taiko scroller

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/100
- Adds passing tracking to `GameplayBeatmap` (quite a simple implementation but should do for now).

This was a pretty interesting one to implement due to the unique sizing requirements. There's still an issue which can be reproduced by resizing the window's height rapidly. I'm not sure how to solve this due to the dependency chain used for sizing here.

No default implementation for lazer just yet.


osu.Game/Screens/Play/GameplayBeatmap.cs
```diff
@@ -38,5 +41,14 @@ public BeatmapInfo BeatmapInfo
...
         public IBeatmap Clone() => PlayableBeatmap.Clone();
+
+        public IBindable<bool> Passing => passing;
+
+        private readonly BindableBool passing = new BindableBool(true);
+
+        public void OnNewResult(JudgementResult result)
+        {
+            passing.Value = result.Type > HitResult.Miss;
+        }
```
bdach, line 52: Perhaps a little monothematic of me, but in light of #8857 I'm wondering what direction this should go towards. The name of `Passing` could be a little misleading (could be misinterpreted for checking health), but the bigger issue I'm seeing is that the taiko pippidon mascot appears to need a *lot* more information to reimplement stable logic, namely:

* if the last result was a hit or not
* whether the last judgement affects combo (only combo-affecting judgements change state)
* what was the total combo at the judgement (special animation every 50 combo)
* what the judgement type was (special animation on every swell passed)
* once the map is done, whether the play is a pass or a fail (currently don't see an easy way to access health info from a skinnable drawable)

To get most of that info currently I'm jamming the entire `JudgementResult` into the mascot, but I'm a little worried that we might soon have two ways of doing essentially the same thing. Jamming all of the above data into `GameplayBeatmap` seems like an... inelegant out, but storing the last result in the beatmap might also not be great.

peppy: exposing a bindable `LastJudgement` in `GameplayBeatmap` might be enough to cover all of that?

bdach: Health is the possible one exception (taiko is weird in that `FailedAtJudgement` doesn't really work, since the fail doesn't happen until everything is actually judged) but as for everything else, seems like a possible way forward, sure.

peppy: I've updated to this. I guess the only thing remaining is combo and health.

bdach: Cool 👍 Once this is in I'll merge in these changes to that mascot PR and reuse. I'd say the health stuff can come as a separate PR as the changes I have in there are extensive enough as-is.

osu.Game.Rulesets.Taiko/Skinning/LegacyTaikoScroller.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            LastResult.BindValueChanged(result =>
+            {
+                foreach (var sprite in InternalChildren.OfType<ScrollerSprite>())
+                    sprite.Passing = result.NewValue == null || result.NewValue.Type > HitResult.Miss;
```
bdach, line 39: I think this should also be checking `AffectsCombo` to prevent missing drumrolls, single-hitting strong hits, etc. from transitioning the scroller to the miss state.

osu.Game.Rulesets.Taiko/Skinning/LegacyTaikoScroller.cs
```diff
@@ -0,0 +1,128 @@
...
+                        sprite.Expire();
+                }
+
+                var last = InternalChildren.LastOrDefault();
+
+                // only break from this loop once we have saturated horizontal space completely.
+                if (last != null && last.ScreenSpaceDrawQuad.TopRight.X >= ScreenSpaceDrawQuad.TopRight.X)
+                    break;
+
+                AddInternal(new ScrollerSprite());
```
bdach, line 70: If you miss a judgement while sprite 1 is currently in view, but then there are no further judgements until sprite 2 needs to start showing, sprite 2 will mistakenly start with `passing = true` - this needs to receive the current passing state when it is spawned.

Can be repro'd in the test scene by toggling to fail state and waiting for the current sprite to scroll out ([video](https://streamable.com/mjqdkv)).

bdach (changes requested):

osu.Game.Rulesets.Taiko/Skinning/LegacyTaikoScroller.cs
peppy: nicely spotted

osu.Game.Rulesets.Taiko/Skinning/LegacyTaikoScroller.cs
```diff
@@ -0,0 +1,137 @@
...
+
+                var last = InternalChildren.LastOrDefault();
+
+                // only break from this loop once we have saturated horizontal space completely.
+                if (last != null && last.ScreenSpaceDrawQuad.TopRight.X >= ScreenSpaceDrawQuad.TopRight.X)
+                    break;
+
+                AddInternal(new ScrollerSprite
+                {
+                    Passing = passing
```
bdach, line 78: This is set too early - when I did the repro scenario described in the earlier comment in the test scene, this crashed with a nullref:

```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at osu.Framework.Graphics.TransformableExtensions.PopulateTransform[TValue,TEasing,TThis](TThis t, Transform`3 transform, TValue newValue, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.MakeTransform[TThis,TEasing,TValue](TThis t, String propertyOrFieldName, TValue newValue, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.TransformTo[TThis,TValue,TEasing](TThis t, String propertyOrFieldName, TValue newValue, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.FadeTo[T,TEasing](T drawable, Single newAlpha, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.FadeIn[T,TEasing](T drawable, Double duration, TEasing& easing)
   at osu.Framework.Graphics.TransformableExtensions.FadeIn[T](T drawable, Double duration, Easing easing)
   at osu.Game.Rulesets.Taiko.Skinning.LegacyTaikoScroller.ScrollerSprite.set_Passing(Boolean value) in D:\Open Source\osu\osu.Game.Rulesets.Taiko\Skinning\LegacyTaikoScroller.cs:line 107
```

Should probably be done on `LoadComplete()`.

Even then the fade duration in the setter is a problem. It should probably fade instantly the first time and then normally after load complete, but even that can look slightly wrong if you miss close to the seam between two sprites (the left one will transition over 200ms, while the right one will spawn already transitioned).

I'm not sure how viable it is but I'd actually probably consider waiting for the texture wrapping framework-side changes and using that instead - it should be a relatively fool-proof method and would get rid of the infinite loop relayouting, too, if it doesn't end up being too expensive performance-wise.

bdach (changes requested):

osu.Game.Rulesets.Taiko/Skinning/LegacyTaikoScroller.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            LastResult.BindValueChanged(result =>
+            {
+                var r = result.NewValue;
+
+                passing = r == null || (r.Judgement.AffectsCombo && r.Type > HitResult.Miss);
```
bdach, line 42: This is also slightly wrong - if the judgement doesn't affect combo it shouldn't affect the value of `passing` in any way (the set should not happen at all).

bdach (approved):
Looks good with the latest changes. CI failure is unrelated (#8942), so I guess we can get this in for now and maybe iterate on the texture wrapping in the future 👍

smoogipoo (approved):
There's a few minor things I could point out, like the `IsLoaded` check of the scrollers should probably be `LoadState >= LoadState.Ready` instead, but this is fine for now.

PR #8924, opened 2020-05-02
Add local beatmap lookup cache to improve performance of imports

### LINKED ISSUE #8265: Reduce web requests in import process
Via static snapshots

peppy:
- Closes https://github.com/ppy/osu/issues/8265

Importing from stable (194 beatmaps):

Offline (master): 30.2s
Online (master): 38.2s
With local cache: 30.6s

Note that these results are with very good internet (<50ms responses to lookups). Performance will be *magnitudes* better on lesser internet connections. This also reduces a large burden on the server infrastructure during imports, where previously over 500,000 request could be fired for a normal user's beatmap collection.

Update support would need to be added at some point, but I think this is good to go as an initial implementation (it's not supposed to have 100% coverage and will gracefully fallback when not available locally). May add tests covering download / lookup if required (potentially moving the `BeatmapUpdateQueue` out of nesting – left it there for simplicity as it uses a couple of methods in `ArchiveModelManager`).

Notes about the database itself:
- 11mb after bz2 (35mb extracted)
- Contains all ranked / approved / loved beatmaps
- Contains more metadata than is being used (can be further pruned if we deem that to be useless)

### DIFF
--- a/osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
+++ b/osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
@@ -65,7 +65,7 @@ namespace osu.Game.Beatmaps
                     {
                         using (var db = new SqliteConnection(storage.GetDatabaseConnectionString("online")))
                         {
-                            var found = db.QueryFirstOrDefault<CachedOnlineBeatmapLookup>(
+                            var found = db.QuerySingleOrDefault<CachedOnlineBeatmapLookup>(
                                 "SELECT * FROM osu_beatmaps WHERE checksum = @MD5Hash OR beatmap_id = @OnlineBeatmapID OR filename = @Path", beatmap);
 
                             if (found != null)
--- a/osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
+++ b/osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
@@ -1,0 +1,180 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Dapper;
+using Microsoft.Data.Sqlite;
+using osu.Framework.IO.Network;
+using osu.Framework.Logging;
+using osu.Framework.Platform;
+using osu.Framework.Threading;
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests;
+using SharpCompress.Compressors;
+using SharpCompress.Compressors.BZip2;
+
+namespace osu.Game.Beatmaps
+{
+    public partial class BeatmapManager
+    {
+        private class BeatmapUpdateQueue
+        {
+            private readonly IAPIProvider api;
+            private readonly Storage storage;
+
+            private const int update_queue_request_concurrency = 4;
+
+            private readonly ThreadedTaskScheduler updateScheduler = new ThreadedTaskScheduler(update_queue_request_concurrency, nameof(BeatmapUpdateQueue));
+
+            private FileWebRequest cacheDownloadRequest;
+
+            private const string cache_database_name = "online.db";
+
+            public BeatmapUpdateQueue(IAPIProvider api, Storage storage)
+            {
+                this.api = api;
+                this.storage = storage;
+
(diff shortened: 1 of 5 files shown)


osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
```diff
@@ -0,0 +1,180 @@
...
+                return Task.WhenAll(beatmapSet.Beatmaps.Select(b => UpdateAsync(beatmapSet, b, cancellationToken)).ToArray());
+            }
+
+            // todo: expose this when we need to do individual difficulty lookups.
+            protected Task UpdateAsync(BeatmapSetInfo beatmapSet, BeatmapInfo beatmap, CancellationToken cancellationToken)
+                => Task.Factory.StartNew(() => update(beatmapSet, beatmap), cancellationToken, TaskCreationOptions.HideScheduler, updateScheduler);
+
+            private void update(BeatmapSetInfo set, BeatmapInfo beatmap)
+            {
+                if (cacheDownloadRequest == null && storage.Exists(cache_database_name))
```
bdach, line 62: I'd split the body of this `if` to a method, reads a little nested right now.

osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
```diff
@@ -0,0 +1,180 @@
...
+                {
+                    try
+                    {
+                        using (var db = new SqliteConnection(storage.GetDatabaseConnectionString("online")))
+                        {
+                            var found = db.QueryFirstOrDefault<CachedOnlineBeatmapLookup>(
```
bdach, line 68: Seeing as the query clauses for fetching the row all seem pretty uniquely identifying I'd probably switch to

```suggestion
                            var found = db.QuerySingleOrDefault<CachedOnlineBeatmapLookup>(
```

to avoid any mishaps with matching multiple records and assigning a wrong online ID in the process.

osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
```diff
@@ -0,0 +1,180 @@
...
+                    LogForModel(set, $"Online retrieval failed for {beatmap} ({e.Message})");
+                }
+            }
+
+            private void prepareLocalCache()
+            {
+                string cacheFilePath = storage.GetFullPath(cache_database_name);
+                string compressedCacheFilePath = $"{cacheFilePath}.bz2";
+
+                cacheDownloadRequest = new FileWebRequest(compressedCacheFilePath, $"https://assets.ppy.sh/client-resources/{cache_database_name}.bz2");
```
bdach, line 132: So as a general question, is this a one-and-done sort of thing for now, or is this going to be periodically updated? Because if this file will get updated, it'd be nice to automatically refetch it every now and again to improve cache hit rate for new beatmaps.

osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
```diff
@@ -0,0 +1,180 @@
...
+                        cacheDownloadRequest = null;
+                    }
+                    catch (Exception ex)
+                    {
+                        Logger.Log($"{nameof(BeatmapUpdateQueue)}'s online cache extraction failed: {ex}", LoggingTarget.Database);
+                    }
+                    finally
+                    {
+                        File.Delete(compressedCacheFilePath);
+                        File.Delete(cacheFilePath);
```
bdach, line 161: Are you sure? I think you might be missing an extraction failure check on this particular deletion, because as it stands when I run this code it downloads the archive fine, then unzips it and promptly nukes both files right after that, rendering the cache lookup code a little pointless...

bdach (changes requested):

osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
peppy: As per the PR description, that will be added at a future point. The main goal is to reduce queries on initial import, and even after months or years it will still suffice from that perspective. Still not 100% on how this should be distributed. nuget package in consumer projects seems like the best path forward in my mind (having it bundled with updates and install) as it will benefit from binary diffing for updates and not require its own download/update/patch logic.

osu.Game/Beatmaps/BeatmapManager_UpdateQueue.cs
peppy: guaranteed by the unique index but sure

bdach (approved):
seems good 👍

smoogipoo (approved):
Seems pretty sane

PR #8933, opened 2020-05-04
Fix hard crash when changing ruleset while presenting a score

### LINKED ISSUE #8692: Crash song select by changing ruleset while entering results screen
image

Repro:
- Go to song select
- Change ruleset using Ctrl-<number>
- Quickly click a leaderboard score

peppy:
- Closes https://github.com/ppy/osu/issues/8692

Hopefully this should cover all song select `Push` scenarios going forward (which would also use `FinaliseSelection` with new paramter).


smoogipoo (approved):
Seems fine

Issue #960, opened 2017-06-21
Move beatmaps and configurations to another location


taroxd:
I want to ask how to store beatmaps and configurations **not** in %Appdata%. That's because my C Drive is a small SSD, and beatmaps really take up a lot of space.
Also, writing these files to %Appdata% will make osu!lazer not portable. The current stable osu! release is portable so that I can put it in a USB drive.

I do not care about potential privacy or safety problem about these files. Even if I do, I would change ACL by myself.

Currently my workaround is to create a symlink with `mklink`, but I want an official solution to it.
What about checking whether `<installDir/Data>` exists? If it does, launch osu!lazer in portable mode, read and write all the beatmaps and configurations to  `<installDir/Data>` .
Or, what about implementing a command-line switch to specify where these files are, like `--user-data-dir="path/to/user data"` in Google Chrome?

smoogipooo:
This will come when we have the ability to read osu!stable-formatted beatmaps from disk (or import them), and there will be an option to change the song directory.

Assigning as low priority.

peppy:
That's not entirely correct (we will support importing, but there will be no "song" directory).

I agree with checking for a config file in the current working directory and changing storage's `BasePath` as a temporary measure, if you wish to implement this. Otherwise, it will come with the more advanced storage system which we will be migrating to before release.

Kontorted:
I don't believe the program is set to force you to install in a certain directory. Since beatmaps seem to be located in its own root, I would assume it would have portability? Correct me if I am in the wrong!

alexpado:
As a temporary mesure, you can change the `LocateBasePath()` value to `./`.
osu!lazer will store all beatmap etc... in the exe folder, making it runnable from a usb key (yeeeeay)

File to edit : `osu.Framework.Platform.Windows.WindowsStorage.cs:15`

alexpado:
Okay I'm back with some more info.

peppy won't be supporting osu!lazer being "installed" outside the default folder. You should keep you osu!lazer installation in your AppData and just change the `LocateBasePath()` to the desired location (ex: `D:\`. "osu" will be appended by default.).

Kazeragi:
hello, i dont quite understand where i can find the file on disk C:\ to change that configuration

Gringo1337:
sooo, how to change locatoin?

insomnyawolf:
I guess there are a lot of people that could not use lazer without this because they don't have a lot of  free space on their main disk.
[...]

PR #8935, opened 2020-05-04
Add basic custom data directory support

peppy:
A first step towards adding user configurable data directory location. Migration support / UI will come as a future PR.

To make use of this as-is, move your data to a new location and create a new file `storage.ini` with:

`FullPath = /new/location`

- Closes (kinda?) https://github.com/ppy/osu/issues/960
- [x] Depends on https://github.com/ppy/osu-framework/pull/3522


peppy:
Hold off on this one for now. `CustomStorage : NativeStorage` may not work well with other host storage implementations and there's no obvious solution to this, so may need some rethinking.

bdach:
This change breaks custom ruleset loading. Logs show a doubly-appended `rulesets` directory prefix:

```
2020-05-09 17:33:52 [error]: Failed to load ruleset osu.Game.Rulesets.Tau
2020-05-09 17:33:52 [error]: System.IO.FileNotFoundException: Could not load file or assembly 'C:\Users\dachb\AppData\Roaming\osu\rulesets\rulesets\osu.Game.Rulesets.Tau.dll'. The system cannot find the path specified.
2020-05-09 17:33:52 [error]: File name: 'C:\Users\dachb\AppData\Roaming\osu\rulesets\rulesets\osu.Game.Rulesets.Tau.dll'
```

The reason is located here:

https://github.com/ppy/osu/blob/2d492102509ae8e1f07970c7f705495379455fc7/osu.Game/Rulesets/RulesetStore.cs#L146-L149

`rulesetStorage.GetFiles()` returns paths relative to the *root* storage, while `rulesetStorage.GetFullPath()` assumes the path is relative to the *sub*storage, therefore appending the `rulesets/` folder a second time.

bdach (changes requested):
As above.

peppy:
Nice catch. Should be fixed with more path manipulation. Hopefully we don't nest these directory requests too much, since each layer will add quite a few extra string operations (though this seems to be a relative rare call).

bdach (approved):
Yep, looks good here now. Tested it a little bit and seems to be working fine. 👍

smoogipoo (approved):
Looks fine to me for now. Only (minor) issue is `OpenInNativeExplorer()` not really working for subpaths since it just forwards to the underlying storage.
Where this comes up in in the editor - `Export Package` will redirect to `/osu/` rather than `/osu/exports/`.

But I imagine fixing this will require an o!f change that exposes `OpenInNativeExplorer(string path)`.

PR #8945, opened 2020-05-05
Add ability for a ruleset to disable gameplay overlays from displaying

Game4all:
# Summary
- This exposes publicly the `HUDOverlay` in `Player` to allow disabling programmatically the HUD, which is useful when making a visualizer ruleset as it allows to hide every HUD piece with 
image
 instead of having to create a dummy Mod inheriting `IApplicableToHUD` and `IApplicableToPlayer` in order to hide the HUD as well as the BreakOverlay.


peppy:
I think this is okay, but keep in mind that at some point in the future rulesets likely won't have access to `Player` like this. I don't see this as a final solution.

@smoogipoo check you're okay with this for the time being.

peppy (dismissed):

Game4all:
To be honest, I have been hesitating to PR this as it is but ended up doing so because the two other ways I thought of turned out to be even uglier / not work at all:

- The `[Cached]` way on `HUDOverlay` but since the `DrawableRuleset` is DI'ed before the overlay is actually constructed and DI'ed. (would throw `DependencyNotRegistered`), so not working at all.

- Creating a dummy unselectable mod inheriting `IApplicableToPlayer` as well as `IApplicableToHUD` which gets passed to the DrawableRuleset` mod list constructor inside `Ruleset.CreateDrawableRulesetWith()` and having the mods being sourced from a `DrawableRuleset.Mods` property instead of `OsuScreen.Mods` (to include the unselectable mod.). While it may work, this really feels like an ugly way of doing it

peppy:
@Game4all let's go about this another way since this seems to be driving people in the wrong direction:

Please add a new virtual bool in `DrawableRuleset`:

```csharp
protected virtual bool HUDVisibleByDefault => true;
```

and use this instead.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -487,6 +487,15 @@ public HitWindows FirstAvailableHitWindows
...
         protected virtual ResumeOverlay CreateResumeOverlay() => null;
 
+        /// <summary>
+        /// Whether to display the HUD with this ruleset.
+        /// Override to false to completely disable the display of the HUD with this ruleset.
+        /// </summary>
+        /// <remarks>
+        /// HUD refers here to <see cref="HUDOverlay"/> in player as well as <see cref="BreakOverlay"/>.
+        /// </remarks>
+        public virtual bool DisplayHud => true;
```
peppy, line 497: any reason you didn't use the suggested naming? you also got the case incorrect for HUD. it's an acronym. and you can't just arbitrarily add other hidden elements and call it HUD because that acronym is already reserved.

Rename this to `AllowGameplayOverlays`

peppy (changes requested):

bdach:
It seems that the way `AllowGameplayOverlays` is currently implemented will work but I can't help but notice that `ModCinema` does the same things and more, namely

https://github.com/ppy/osu/blob/f414b7fbb60d9e47a054f1d61493d846696125bc/osu.Game/Rulesets/Mods/ModCinema.cs#L32-L45

I wonder if the dim disables wouldn't be relevant for this use case too, and whether we should revert to a settable property so that we can reuse the logic...

Game4all:
Making it a settable property to allow setting gameplay overlays visibility from mods sounds fair IMO; but I think the dim disable should be kept separately from the property as it is specific to Cinema Mod .

bdach:
If you're not interested in disabling the dim in both places then I'd stay with the virtual property as-is, as the logic is no longer entirely common, so I don't consider that code duplication myself.

Game4all:
The initial intent for this PR was to expose a way allowing to disable the HUD from the `DrawableRuleset` without having to use a dummy `Mod` inheriting `IApplicableToHUD` as well as `IApplicableToPlayer`; so I'm going to keep on that stance.

peppy:
Just to confirm: your usage of this in a ruleset does not intend to undo dim?

Game4all:
Yes; the only intent behind this is to completely disable any gameplay overlay in a visualizer ruleset without having to use a `Mod` for that.

peppy:
Kinda hoped for a test for this but seems non-straightforward to add. Manual testing seems to show this working as expected.

Issue #8272, opened 2020-03-13
Sliderborder width is higher than in stable


oSumAtrIX:
**Describe the bug:** 
Might be intended, but the slider border with in lazer is higher than on stable. 

**Screenshots or videos showing encountered issue:** 
![2020-03-13 23_46_35-Window](https://user-images.githubusercontent.com/13122796/76664727-a0bcb980-6585-11ea-8d33-b656af92ce76.png)

**osu!lazer version:** 
2020.312.0

PR #8948, opened 2020-05-06
Adjust slider border width to be closer to stable

oSumAtrIX:
Related issue: https://github.com/ppy/osu/issues/8272


bdach:
While this PR does seem to bring the width closer to stable when legacy skins are used:

image

it makes the default lazer skin look really bad (note the discrepancy between hit circle & slider border widths):

image

In general my level of confidence in this change is extremely low, seeing that you've committed these changes through the github web interface and submitted this PR off `master`. Please don't do that, and make & test your changes locally if you want to continue submitting code.

peppy:
There's already a path for skin lookups of this value. This should be fixed in the legacy transformer rather than at the global location you changed, due to reasons mentioned above by @bdach.

Please open a new pull request if you decide to go about this in a proper way. Please do not use the web editor to make changes as it suggests you have not tested this locally.

PR #8951, opened 2020-05-06
isn't it LICENSE?

ghost:


bdach:
Both versions are technically correct, but by convention British English word spellings are used in this repository.

Closing.

mrquantumoff:
Ok, thanks for info!

PR #8953, opened 2020-05-06
Make HUDOverlay overridable

andy840119:
In compare to make `HUDOverlay` public in #8945 
I think it's better to make `HUDOverlay` overridable, and do the logic(like hiding or change to customize component) in `MyCoolRulesetHUDOverlay`.

### DIFF
--- a/osu.Game/Rulesets/UI/DrawableRuleset.cs
+++ b/osu.Game/Rulesets/UI/DrawableRuleset.cs
@@ -487,6 +487,9 @@ namespace osu.Game.Rulesets.UI
 
         protected virtual ResumeOverlay CreateResumeOverlay() => null;
 
+        public virtual HUDOverlay CreateHUDOverlay(ScoreProcessor scoreProcessor, HealthProcessor healthProcessor, DrawableRuleset drawableRuleset, IReadOnlyList<Mod> mods)
+            => new HUDOverlay(scoreProcessor, healthProcessor, drawableRuleset, mods);
+
         /// <summary>
         /// Sets a replay to be used, overriding local input.
         /// </summary>
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -265,23 +265,20 @@ namespace osu.Game.Screens.Play
                 // display the cursor above some HUD elements.
                 DrawableRuleset.Cursor?.CreateProxy() ?? new Container(),
                 DrawableRuleset.ResumeOverlay?.CreateProxy() ?? new Container(),
-                HUDOverlay = new HUDOverlay(ScoreProcessor, HealthProcessor, DrawableRuleset, Mods.Value)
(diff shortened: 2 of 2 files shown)


osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -487,6 +487,9 @@ public HitWindows FirstAvailableHitWindows
 
         protected virtual ResumeOverlay CreateResumeOverlay() => null;
 
+        public virtual HUDOverlay CreateHUDOverlay(ScoreProcessor scoreProcessor, HealthProcessor healthProcessor, DrawableRuleset drawableRuleset, IReadOnlyList<Mod> mods)
```
andy840119, line 490: Not sure it's better to use `public` or `internal`.

Game4all, line 490: May be cool to add XMLDoc for this.

andy840119: Have no idea what to add 🤔

iiSaLMaN:
Since I've been reading and investigating through the `HUDOverlay` component and thinking about the per-ruleset HUD creation yesterday, I'll provide my own proposal here:

The `HUDOverlay` component shouldn't be entirely visible to ruleset developers as it contains methods and properties that can mess up the `Player` screen or even the entire game with, the exposure should be stricted enough to what rulesets should be allowed to do instead.

## My proposal & visioned implementation

I would say, the HUD components we're fine to let be modified (like health bar, combo counter, key counter, etc...) exposed in a `DefaultHUDComponents` class deriving from the base class `HUDComponents`.

Leaving components like *hold-to-quit overlay*, *player-settings overlay* and *selected-mods overlay* staying still and unmodifiable.

I've decided on the `HUDComponents` being a class that inherits `IEnumerable<Drawable>` such that it can be easily constructed using the collection initializer:
```csharp
new HUDComponents
{
    new SpecialBarDisplay(),
    ...
}
```

**while also** allowing to add to the existing default HUD components by:
```csharp
new DefaultHUDComponents
{
    new SpecialBarDisplay(),
    ...
}
```

The `DefaultHUDComponents` class would expose all of its *default* components as public get-only properties to allow for manipulating their position & size (in relative axes for simplicity):

```csharp
new DefaultHUDComponents
{
    HealthBar = { Scale = new Vector2(1f, 0.6f) },
    ComboCounter = 
    {
        RelativePositionAxes = Axes.Both,
        Position = new Vector2(0f, 0.9f),
    },
}
```

---

That's just my proposal, open for suggestions to get this more better if possible!

bdach:
I agree with not overriding if it can be avoided, but would like to hear first what *concrete* potential use cases someone might have for overriding the HUD (aside from "why not").

Having visual continuity across rulesets when it comes to the UI is a good and desirable thing. If extensibility is desired then it would be important to consider *why*, as if we expose the HUD now it's not going to be something we can take back easily without really breaking rulesets that could use that.

peppy:
The HUD is not available for ruleset modification. This is a design decision. The only valid operation here is to hide it.

PR #8958, opened 2020-05-07
Fix potential nullref in song select filtering

### LINKED ISSUE #8957: Enabling converted beatmaps with no beatmaps visible crashes song select
Enabling `Show converted beatmaps` which filtered crashes song select.

Steps to reproduce:
1. Disable `Show converted beatmaps`.
2. Enter song select.
3. Type a search filter that displays no beatmaps.
4. Enable `Show converted beatmaps`.
5. Crash.

Log:
```
Unhandled exception. System.NullReferenceException: Object reference not set to an instance of an object.
   at osu.Game.Screens.Select.Carousel.CarouselBeatmap.<>c__DisplayClass3_0.<Filter>b__0(String term) in /Users/smoogipoo/Repos/osu/osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs:line 64
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at osu.Game.Screens.Select.Carousel.CarouselBeatmap.Filter(FilterCriteria criteria) in /Users/smoogipoo/Repos/osu/osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs:line 64
   at osu.Game.Screens.Select.Carousel.CarouselGroup.<>c__DisplayClass10_0.<Filter>b__0(CarouselItem c) in /Users/smoogipoo/Repos/osu/osu.Game/Screens/Select/Carousel/CarouselGroup.cs:line 85
   at System.Collections.Generic.List`1.ForEach(Action`1 action)
   at osu.Game.Screens.Select.Carousel.CarouselGroup.Filter(FilterCriteria criteria) in /Users/smoogipoo/Repos/osu/osu.Game/Screens/Select/Carousel/CarouselGroup.cs:line 85

```

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8957

Looks like this was just me having some broken beatmaps, but should probably safeguard anyway.

### DIFF
--- a/osu.Game/Beatmaps/BeatmapInfo.cs
+++ b/osu.Game/Beatmaps/BeatmapInfo.cs
@@ -149,6 +149,11 @@ namespace osu.Game.Beatmaps
             }
         }
 
+        public string[] SearchableTerms => new[]
+        {
+            Version
+        }.Concat(Metadata?.SearchableTerms ?? Enumerable.Empty<string>()).Where(s => !string.IsNullOrEmpty(s)).ToArray();
+
         public override string ToString()
         {
             string version = string.IsNullOrEmpty(Version) ? string.Empty : $"[{Version}]";
--- a/osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs
+++ b/osu.Game/Screens/Select/Carousel/CarouselBeatmap.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Generic;
 using System.Linq;
 using osu.Game.Beatmaps;
 using osu.Game.Screens.Select.Filter;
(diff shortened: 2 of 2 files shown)


peppy:
Have you checked how a term can be added that is null in the first place? Sounds like we probably want to stop that from occurring rather than a local null check.

smoogipoo:
In this case it's a beatmap with a null version, which isn't a part of `SearchableTerms` and thus doesn't have a `!string.IsNullOrEmpty()` check around it.

An alternative solution is to include it into `SearchableTerms`, where the only other usage is the music controller playlist (which I think is fine to also have the version in there).

smoogipoo:
Ah, this likely wasn't done already since the searchable terms are part of the metadata, so the above proposal would have to add another property to `BeatmapInfo` which concats the metadata's terms.

Let me know if that is the more desired solution.

peppy:
Yeah, let's do what's needed to keep the search terms from including null/empty strings. Makes it easier for components to work with.

peppy (approved):

PR #8960, opened 2020-05-07
Fix date tooltip not showing in 24-hour format

Joehuu:


peppy:
I think 24h is a better default, but we probably want to switch to the user's OS format at some point.

peppy (approved):

PR #8962, opened 2020-05-07
Storage directory migration support

peppy:
Note that this does not have UI implementation yet.

- [x] Add edge case testing (for scenarios that cause aborts).
- [x] Depends on #8935.


smoogipoo:
Am assuming this PR is not ready yet? (Also needs conflict resolution)

peppy:
Yeah, is failing on tests when run on appveyor so at very least I'll fix that first. The concept is there but there are cases of database contexts still holding file handles open.

osu.Game/IO/OsuStorage.cs
```diff
@@ -9,17 +13,111 @@ namespace osu.Game.IO
...
+
+        public void Migrate(string newLocation)
+        {
+            var source = new DirectoryInfo(GetFullPath("."));
+            var destination = new DirectoryInfo(newLocation);
+
+            // ensure the new location has no files present, else hard abort
+            if (destination.Exists)
+            {
+                if (destination.GetFiles().Length > 0)
```
smoogipoo, line 54: Maybe also `GetDirectories()`?

osu.Game/IO/OsuStorage.cs
```diff
@@ -9,17 +13,111 @@ namespace osu.Game.IO
...
+                    throw new InvalidOperationException("Migration destination already has files present");
+
+                deleteRecursive(destination);
```
smoogipoo, line 57: Is this hoping to delete the destination directory itself for the `copyRecursive` call below? Because this doesn't seem like it would do that (only delete all files/directories within the destination).

osu.Game/IO/OsuStorage.cs
```diff
@@ -9,17 +13,111 @@ namespace osu.Game.IO
...
+
+                while (tries-- > 0)
+                {
+                    try
+                    {
+                        fi.CopyTo(Path.Combine(destination.FullName, fi.Name), true);
+                        break;
+                    }
+                    catch (Exception)
+                    {
```
smoogipoo, line 109: This should probably have some logging and perhaps completely abort the process if it fails.

osu.Game/IO/OsuStorage.cs
```diff
@@ -9,17 +13,111 @@ namespace osu.Game.IO
...
+            foreach (DirectoryInfo dir in target.GetDirectories())
+            {
+                if (IGNORE_DIRECTORIES.Contains(dir.Name))
+                    continue;
+
+                dir.Delete(true);
+            }
+        }
+
+        private static void copyRecursive(DirectoryInfo source, DirectoryInfo destination, bool topLevelExcludes = true)
```
smoogipoo, line 89: `topLevelExcludes` doesn't seem used? Both here and in `deleteRecursive`.

smoogipoo (changes requested):

osu.Game/OsuGameBase.cs
```diff
@@ -121,7 +121,7 @@ public OsuGameBase()
         protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent) =>
             dependencies = new DependencyContainer(base.CreateChildDependencies(parent));
 
-        private DatabaseContextFactory contextFactory;
+        protected DatabaseContextFactory ContextFactory;
```
smoogipoo, line 124: Did this need to be exposed? Seems like it can remain `private`.

osu.Game/IO/OsuStorage.cs
peppy: the directory itself is fine to continue existing in this case, i think.

smoogipoo (approved):

PR #8965, opened 2020-05-07
Add "Check for updates" button

### LINKED ISSUE #8947: Add manual "Check for updates" button in settings
**Describe the new feature:** 
-------------------------------------
Some times the internet connection are unstable or too slow that take too long time to update, least I am.
The game are not trying to update everytime when start if you close it before downloading update finish

Please add a _**Check Update**_ button in settings below _**Open Osu! Folder**_ for manually start update or check update everytime the game starts

-------------------------------------

Add a _**Check update**_ button in settings below **_Open Osu! Folder_**
Check for update everytime the game starts

Craftplacer:
Closes #8947
- [x] Depends on #9002


osu.Game/Updater/UpdateManager.cs
```diff
@@ -44,6 +45,11 @@ protected override void LoadComplete()
             config.Set(OsuSetting.Version, version);
         }
 
+        public virtual void CheckForUpdate()
+        {
+            Logger.Log("CheckForUpdate was called on the base class (UpdateManager)", LoggingTarget.Information);
```
bdach, line 50: Why is this being logged at all?

osu.Game/OsuGame.cs
```diff
@@ -609,6 +609,9 @@ protected override void LoadComplete()
 
             loadComponentSingleFile(screenshotManager, Add);
 
+            // dependency on notification overlay
+            loadComponentSingleFile(CreateUpdateManager(), Add, true);
```
bdach, line 613: Is there any particular reason for this location of this load?

osu.Game/OsuGame.cs
```diff
@@ -765,11 +767,17 @@ private void forwardLoggedErrorsToNotifications()
...
+        /// </summary>
+        /// <param name="d">The component to load.</param>
+        /// <param name="add">The method to invoke for adding the component.</param>
+        /// <param name="cache">Whether to cache the component as type <typeparamref name="T"/> into the game dependencies before any scheduling.</param>
         private T loadComponentSingleFile<T>(T d, Action<T> add, bool cache = false)
             where T : Drawable
         {
             if (cache)
-                dependencies.Cache(d);
+                dependencies.CacheAs(d);
```
bdach, line 780: I find this a little dangerous. I assume you've changed the `Cache` to `CacheAs` so that components aren't cached with their most derived type (returned at runtime), but with their *declaration* type as written in the field/local. This is an extremely subtle change that can cause regressions if any of the types anywhere don't match - did you make sure that *none* of the single-file-loaded cached dependencies break?

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -21,6 +28,16 @@ private void load(Storage storage, OsuConfigManager config)
...
 
+            if (game != null && updateManager != null)
+            {
+                Add(new SettingsButton
+                {
+                    Text = "Check for updates",
+                    Action = updateManager.CheckForUpdate,
+                    Enabled = { Value = game.IsDeployedBuild }
+                });
+            }
```
bdach, line 39: You could just as well do

```suggestion
            Add(new SettingsButton
            {
                Text = "Check for updates",
                Action = updateManager?.CheckForUpdate,
                Enabled = { Value = game?.IsDeployedBuild ?? false }
            });
```

to eliminate the `if` null-checks.

osu.Game/OsuGame.cs
Craftplacer: UpdateManager depends on NotificationOverlay. SettingsOverlay has UpdateSettings which depends on UpdateManager, so it has to be "sandwiched".

osu.Game/Updater/UpdateManager.cs
Craftplacer: I felt this is the right thing to do. You don't want to silently do nothing, because you missed something or expected something else.

bdach: Doesn't seem very useful. There is no practical difference to the user between doing nothing and logging this (default log level is verbose, so it won't show up anywhere unless you manually check the log). I'd argue that in the cases that would log this, the button should be either disabled or not even visible to users.

I'm pretty sure the only platform that uses the base `UpdateManager` directly is iOS, on which the update button is probably pretty useless anyway in the long run due to the app store existing (I presume testflight also handles app updates, but I wouldn't know).

osu.Game/OsuGame.cs
bdach: The `UpdateManager -> SettingsOverlay` dependency should be annotated with an inline comment for posterity in that case, albeit I don't really like how interdependent everything is starting to become.

osu.Game/OsuGame.cs
Craftplacer: Yes, everything is working.

osu.Game/Updater/UpdateManager.cs
Craftplacer: I'm going to keep the log message for purely debugging purposes. But I'm going to hide the button if we're having the base class.

osu.Game/OsuGame.cs
bdach: If I may ask, how did you determine that? To be able to say that with certainty would require to inspect every BDL/`[Resolved]` member that uses the types cached here and make sure none of them use a more-derived type than the one declared in this class.

iiSaLMaN: That's the same as checking for every `loadComponentSingleFile<T>` call to look for ones that have the instantiated type more-derived than the `T` generic type, in which case I couldn't find one ([other than the already-resolved case in the branch](https://github.com/ppy/osu/pull/8965/commits/c025814f403dc5fcf4f07342791387986c526809#diff-4af0c24c8977f999b5986c46cc749ec3R93)), that would mean a confirmed yes.

Moreover, caching components as `T` type IMO is better than most-derived type for this case of loading game components, just like what's being done here (with `CreateUpdateManager()` instance cached as `UpdateManager`), there may be specific components with platform-specific subclasses that need to be cached as their one shared base class to allow resolving them throughout the game.

bdach: I don't necessarily disagree with the content of the change, I mostly wanted to make sure it was tested as safe.

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -11,24 +11,22 @@ namespace osu.Game.Overlays.Settings.Sections.General
...
         {
             Add(new SettingsEnumDropdown<ReleaseStream>
             {
                 LabelText = "Release stream",
                 Bindable = config.GetBindable<ReleaseStream>(OsuSetting.ReleaseStream),
             });
 
-            if (game != null && updateManager != null)
+            // We shouldn't display the button for the base UpdateManager (without updating logic)
+            if (updateManager != null && updateManager.GetType() != typeof(UpdateManager))
```
bdach, line 29: Not sure I like this relying on types this much. I'd probably add a `public virtual bool CanPerformUpdate { get; }` on the update manager (`=> false` in the base, `=> true` in the derived).

Craftplacer: I believe doing a virtual property is more flexible, so I'm going with that.
Also the default implementation of `CreateUpdateManager()` in `osu.Game.OsuGame`
 passes no arguments to the constructor of `UpdateManager`, and it'd be better to keep it at that.

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -25,8 +25,8 @@ private void load(Storage storage, OsuConfigManager config, OsuGameBase game)
                 Bindable = config.GetBindable<ReleaseStream>(OsuSetting.ReleaseStream),
             });
 
-            // We shouldn't display the button for the base UpdateManager (without updating logic)
-            if (updateManager != null && updateManager.GetType() != typeof(UpdateManager))
+            // We should only display the button for UpdateManagers that do update the client
+            if (updateManager != null && updateManager.CanPerformUpdate)
```
peppy, line 29:
```suggestion
            if (updateManager?.CanPerformUpdate == true)
```

osu.Game/Updater/UpdateManager.cs
```diff
@@ -16,6 +17,8 @@ namespace osu.Game.Updater
     /// </summary>
     public class UpdateManager : CompositeDrawable
     {
+        public virtual bool CanPerformUpdate => false;
```
peppy, line 20: no xmldoc. what does this even mean?

osu.Game/Updater/SimpleUpdateManager.cs
```diff
@@ -30,9 +32,11 @@ private void load(OsuGameBase game)
             version = game.Version;
 
             if (game.IsDeployedBuild)
-                Schedule(() => Task.Run(checkForUpdateAsync));
+                CheckForUpdate();
         }
 
+        public override void CheckForUpdate() => Schedule(() => Task.Run(checkForUpdateAsync));
```
peppy, line 38: public method that bypasses the `IsDeployedBuild` check? should probably be protected at most.

osu.Game/OsuGame.cs
peppy, line 780: this should be its own PR with reasoning provided

osu.Game/OsuGame.cs
```diff
@@ -609,6 +609,9 @@ protected override void LoadComplete()
 
             loadComponentSingleFile(screenshotManager, Add);
 
+            // dependency on notification overlay, dependent by settings overlay
+            loadComponentSingleFile(CreateUpdateManager(), Add, true);
```
peppy, line 613: are you sure this doesn't regress anything?

osu.Game/OsuGame.cs
```diff
@@ -90,7 +90,7 @@ public class OsuGame : OsuGameBase, IKeyBindingHandler<GlobalAction>
 
         protected BackButton BackButton;
 
-        protected SettingsPanel Settings;
+        protected SettingsOverlay Settings;
```
peppy, line 93: How is this even related to this PR? Should not be here.

osu.Game/Updater/UpdateManager.cs
peppy: why is this not including the IsDeployedBuild check?

peppy (changes requested):

osu.Game/OsuGame.cs
Craftplacer: image
osu!'s update check at startup still works, so this is a yes from me.

osu.Game/Updater/UpdateManager.cs
Craftplacer: After I committed it, I myself doubted the name of this property, a more appropriate name would be "CanCheckForUpdate".  That name would make more sense and is self-explanatory.

peppy: but why is IsDeployedBuild not included and how does that check differ from the one you added?

osu.Game/OsuGame.cs
bdach: It's a corollary of the `Cache` -> `CacheAs` change [as I outlined here](https://github.com/ppy/osu/pull/8965#discussion_r421827375). I too dislike the subtlety of this.

osu.Game/OsuGame.cs
peppy: @Craftplacer are you going to split this out?

Craftplacer:
The latest commit intentionally breaks the current code (as I transferred the changes over there, which makes merging easier), while #9002 is pending to be discussed and accepted.

osu.Game/Updater/UpdateManager.cs
```diff
@@ -44,6 +50,11 @@ protected override void LoadComplete()
             config.Set(OsuSetting.Version, version);
         }
 
+        public virtual void CheckForUpdate()
+        {
+            Logger.Log("CheckForUpdate was called on the base class (UpdateManager)", LoggingTarget.Information);
```
peppy, line 55: This is wrong

osu.Game/Updater/SimpleUpdateManager.cs
```diff
@@ -19,17 +19,27 @@ namespace osu.Game.Updater
     /// </summary>
     public class SimpleUpdateManager : UpdateManager
     {
+        public override bool CanCheckForUpdate => true;
```
peppy, line 22: why is this not IsDeployedBuild still?

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -5,22 +5,37 @@
...
             });
 
+            // We should only display the button for UpdateManagers that do check for updates
+            if (updateManager?.CanCheckForUpdate == true)
+            {
+                Add(new SettingsButton
+                {
+                    Text = "Check for updates",
+                    Action = updateManager.CheckForUpdate,
+                    Enabled = { Value = game.IsDeployedBuild }
```
peppy, line 35: what?

osu.Desktop/Updater/SquirrelUpdateManager.cs
```diff
@@ -22,8 +22,11 @@ namespace osu.Desktop.Updater
 {
     public class SquirrelUpdateManager : osu.Game.Updater.UpdateManager
     {
+        public override bool CanCheckForUpdate => true;
```
peppy, line 25: why is this not IsDeployedBuild still?

peppy (dismissed):
I think you missed the point.

peppy:
@Craftplacer are you planning on revisiting this?

Craftplacer:
> @Craftplacer are you planning on revisiting this?

This weekend most likely. I was waiting for contributions of another contributor that didn't came out to be.

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -17,8 +19,8 @@ public class UpdateSettings : SettingsSubsection
 
         protected override string Header => "Updates";
 
-        [BackgroundDependencyLoader]
-        private void load(Storage storage, OsuConfigManager config, OsuGameBase game)
+        [BackgroundDependencyLoader(true)]
+        private void load(Storage storage, OsuConfigManager config, OsuGame game)
```
frenzibyte: What is this for?

osu.Desktop/Updater/SquirrelUpdateManager.cs
```diff
@@ -22,33 +22,25 @@ namespace osu.Desktop.Updater
...
         [BackgroundDependencyLoader]
-        private void load(NotificationOverlay notification, OsuGameBase game)
+        private void load(NotificationOverlay notification)
         {
-            gameBase = game;
             notificationOverlay = notification;
 
             Splat.Locator.CurrentMutable.Register(() => new SquirrelLogger(), typeof(Splat.ILogger));
-            CheckForUpdate();
+            Schedule(() => Task.Run(CheckForUpdateAsync));
```
frenzibyte, line 38: The base already calls this, you obviously don't want to check twice?

osu.Game/Updater/UpdateManager.cs
```diff
@@ -18,9 +18,11 @@ namespace osu.Game.Updater
     public class UpdateManager : CompositeDrawable
     {
         /// <summary>
-        /// Whether this UpdateManager is capable of checking for updates.
+        /// Whether this UpdateManager should be or is capable of checking for updates.
         /// </summary>
-        public virtual bool CanCheckForUpdate => false;
+        public bool CanCheckForUpdate => game.IsDeployedBuild;
```
frenzibyte, line 23: This will fail on iOS, maybe do:
```csharp
public bool CanPerformUpdate => game.IsDeployedBuild && CanCheckForUpdate;

protected virtual bool CanCheckForUpdate => false;
```
but not sure about the naming though, may want a more saner naming here

osu.Game/Updater/UpdateManager.cs
```diff
@@ -35,24 +37,37 @@ protected override void LoadComplete()
...
-            // debug / local compilations will reset to a non-release string.
-            // can be useful to check when an install has transitioned between release and otherwise (see OsuConfigManager's migrations).
-            config.Set(OsuSetting.Version, version);
-        }
-
-        public virtual void CheckForUpdate()
-        {
-            Logger.Log("CheckForUpdate was called on the base class (UpdateManager)", LoggingTarget.Information);
+            // we aren't doing any async in this method, so we return a completed task instead.
+            return Task.CompletedTask;
```
frenzibyte: what is this doing here, should stay in `LoadComplete`?

osu.Game/Updater/UpdateManager.cs
```diff
@@ -35,24 +37,37 @@ protected override void LoadComplete()
...
+
+            // debug / local compilations will reset to a non-release string.
+            // can be useful to check when an install has transitioned between release and otherwise (see OsuConfigManager's migrations).
+            config.Set(OsuSetting.Version, game.Version);
+        }
+
+        public async Task CheckForUpdateAsync()
+        {
+            if (!CanCheckForUpdate)
+                return;
```
frenzibyte, line 50: No, don't silently return when you shouldn't be called, throw `NotSupportedException` here.

frenzibyte (changes requested):

osu.Game/Updater/UpdateManager.cs
peppy: reasoning?

frenzibyte: `CheckForUpdateAsync` shouldn't be called if the `UpdateManager` itself isn't capable of doing it, that should be identified as a bug.

peppy: I don't follow. It's a public method. Are you expecting every usage to first check the bool? That sounds like really bad API. Why don't we just make UpdateManager null and throw a nullref then? Isn't the point of making it abstract to simplify external implementation?

osu.Game/Updater/UpdateManager.cs
Craftplacer: You highlighted two methods, what do you mean by "this"?

osu.Game/Updater/UpdateManager.cs
Craftplacer: Why would this be failing, because IsDeployedBuild is false on iOS?

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
Craftplacer: I'm pretty sure that this is another PR that got merged (it doesn't get highlighted as my change on Files Changed).

osu.Game/Updater/UpdateManager.cs
bdach: Yeah, I think as long as iOS reads the assembly version properly this should work with the simple update manager?

frenzibyte: iOS is on a completely different version publishing system (by TestFlight), hence the reason if you open `OsuGameIOS` you would find `SimpleUpdateManager` isn't used there.

With this staying as-is, `Check for update` button will be enabled for iOS, which we're likely not expecting here given how it's published.

osu.Game/Updater/UpdateManager.cs
frenzibyte: Right, you have a point there. It's fine in this way then.

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
frenzibyte: Weird, how did this got displayed to me as if it was your change o_O

osu.Game/Updater/UpdateManager.cs
frenzibyte: Another weird thing by GItHub, anyways why is the version finalizing logic inside `InternalCheckForUpdateAsync`, it should stay where it is at `LoadComplete`?

osu.Game/Updater/UpdateManager.cs
bdach: Fair. I got my virtual implementations confused.

osu.Game/Updater/UpdateManager.cs
Craftplacer: > I don't follow. It's a public method. Are you expecting every usage to first check the bool? That sounds like really bad API. Why don't we just make UpdateManager null and throw a nullref then? Isn't the point of making it abstract to simplify external implementation?

I thought about doing that too, but UpdateManager seems to go beyond of just checking updates.

At this point I would suggest splitting UpdateManager into 2 classes for their own functionality, that being a class for checking for new updates, and another one for doing the startup work when a new update has been applied.

This way we can stick to abstracting the checking process and make it null on not deployed builds/mobile versions.

osu.Game/Updater/UpdateManager.cs
Craftplacer: I mixed up the "Check what's new" code as another checking code, I'll be reverting this the next push.

osu.Game/Updater/UpdateManager.cs
bdach: I can't see the gain of splitting the two things. Sounds like extra complexity for not much gain.

peppy:
I'm taking this PR over. There is a lack of effort or testing and reviewing further is not efficient.

peppy:
- Fixed naming
- Fixed button not disabling while check is in progress
- Fixed update being run twice every startup
- Fixed no check for concurrent update checks

osu.Game/Updater/UpdateManager.cs
peppy: Fix in a slightly different way.

osu.Game/Updater/UpdateManager.cs
```diff
@@ -29,9 +37,13 @@ protected override void LoadComplete()
         {
             base.LoadComplete();
 
-            var version = game.Version;
+            Schedule(() => Task.Run(CheckForUpdateAsync));
+
+            // Query last version only *once*, so the user can re-check for updates, in case they closed the notification or else.
```
peppy, line 42: what

peppy:
@smoogipoo requesting your thorough check of this (although there's not much of it left).

osu.Game/Overlays/Settings/Sections/General/UpdateSettings.cs
```diff
@@ -23,6 +30,17 @@ private void load(Storage storage, OsuConfigManager config, OsuGame game)
...
             });
 
+            // We should only display the button for UpdateManagers that do check for updates
+            Add(checkForUpdatesButton = new SettingsButton
+            {
+                Text = "Check for updates",
+                Action = () =>
+                {
+                    checkForUpdatesButton.Enabled.Value = false;
+                    Task.Run(updateManager.CheckForUpdateAsync).ContinueWith(t => Schedule(() => checkForUpdatesButton.Enabled.Value = true));
```
smoogipoo, line 40: Deserves a null check (`CanBeNull = true` on the attribute).

osu.Game/Updater/UpdateManager.cs
```diff
@@ -44,6 +55,26 @@ protected override void LoadComplete()
...
+
+        public async Task CheckForUpdateAsync()
+        {
+            if (!CanCheckForUpdate)
+                return;
+
+            lock (updateTaskLock)
+                updateCheckTask ??= PerformUpdateCheck();
+
+            await updateCheckTask;
```
smoogipoo, line 70: Safest would be to store the above to a local variable and await on the local variable here. T1 could be executing the statement above, and T2 could be blocked on the statement below, after which T2 will unblock and set null and T1 will try `await null;`

smoogipoo (changes requested):

smoogipoo (approved):

PR #8972, opened 2020-05-08
Fix loading a ruleset with a new version specification causing a crash

### LINKED ISSUE #8936: Ruleset fails to load when version is changed
**Describe the crash:** 
A crash will occur when an ruleset gets updated with a different assembly version (File version, product version). Lazer treats such assemblies as completely new rulesets separate from the older version, and demands a unique ruleset shortname.

I have tested changing some properties in isolation and found that changing the `File Description`, `Product Name`, `Copyright` information doesn't affect lazer.

**Screenshots** 
Left (before, working), Right (after, broken)
image

**osu!lazer version:** 
2020.429.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/4573662/runtime.log)

peppy:
Closes https://github.com/ppy/osu/issues/8936

### DIFF
--- a/osu.Game/Rulesets/RulesetInfo.cs
+++ b/osu.Game/Rulesets/RulesetInfo.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Diagnostics.CodeAnalysis;
+using System.Linq;
 using Newtonsoft.Json;
 
 namespace osu.Game.Rulesets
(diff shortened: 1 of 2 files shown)


smoogipoo:
Crashes on first startup for me:

```
[database] 2020-05-08 10:27:56 [error]: Failed executing DbCommand (3ms) [Parameters=[@p0='?', @p1='?' (Size = 55), @p2='?' (Size = 3), @p3='?' (Size = 3)], CommandType='Text', CommandTimeout='10']
[database] 2020-05-08 10:27:56 [error]: INSERT INTO "RulesetInfo" ("Available", "InstantiationInfo", "Name", "ShortName")
[database] 2020-05-08 10:27:56 [error]: VALUES (@p0, @p1, @p2, @p3);
[database] 2020-05-08 10:27:56 [error]: SELECT "ID"
[database] 2020-05-08 10:27:56 [error]: FROM "RulesetInfo"
[database] 2020-05-08 10:27:56 [error]: WHERE changes() = 1 AND "ID" = last_insert_rowid();
[database] 2020-05-08 10:27:56 [error]: An exception occurred in the database while saving changes for context type 'osu.Game.Database.OsuDbContext'.
```

[client.db.zip](https://github.com/ppy/osu/files/4598443/client.db.zip)

peppy:
Just your database isn't enough to reproduce this. I am guessing one of the dlls you have loaded is causing the issue. Can you look into this one further?

image

Breakpoint this line and see what is hitting this point (this will be the ruleset causing issues) then see why the equality check failed.

smoogipoo:
image

Seems like maybe the check is the wrong way around? Maybe it should be context's instantiation info starting with the given instantiation info?

peppy:
yep, that's the issue. weird i didn't run into this in my testing.

smoogipoo (approved):
Seems to work fine.

PR #8973, opened 2020-05-08
Fix hold note selection pieces disappearing on movement

smoogipoo:
The biggest change of importance is to `ScrollingHitObjectContainer.cs`. It has 3 invalidations:

1. An invalidation based on layout (e.g. screen size changing).
    * This invalidates both (2) and (3).
2. An invalidation based on any hitobject changing (addition, removal, and now `ApplyDefaults`).
    * If this is invalidated, it can be assumed that _at least_ one of (3) is also invalidated.
3. A per-hitobject state invalidation.

The use of (2) is simply as a performance measure, otherwise the SHOC would have to iterate through all hitobjects every frame to see if it needs to recompute the state of any.

### DIFF
--- a/osu.Game/Screens/Edit/EditorBeatmap.cs
+++ b/osu.Game/Screens/Edit/EditorBeatmap.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 using JetBrains.Annotations;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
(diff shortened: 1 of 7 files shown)


osu.Game/Screens/Edit/EditorBeatmap.cs
```diff
@@ -201,6 +202,25 @@ public void RemoveAt(int index)
...
+            mutableHitObjects.Clear();
+
+            foreach (var b in startTimeBindables)
+                b.Value.UnbindAll();
+            startTimeBindables.Clear();
+
+            foreach (var h in removed)
+                HitObjectRemoved?.Invoke(h);
+
+            updateHitObject(null, true);
```
bdach, line 221: So as a general note, not necessarily relevant to this PR in itself - I haven't read a lot of editor code, but when reading this clear implementation it feels like the code could use some restructuring - a lot of things are done here. I was especially confused as to why `SelectedHitObjects` isn't being cleared here but found out that it does get cleared - and it gets cleared because the hit objects remove *themselves* from that list via `HitObjectRemoved` which is not something I would have expected.

smoogipoo: The editor's probably going to go through a dozen more refactors before anything really starts to feel good, I think. It's a complex system of interactions, and this was the result of a refactoring of selections not too long ago.

peppy (approved):

PR #8980, opened 2020-05-09
Check local availability of beatmap before disabling download buttons

### LINKED ISSUE #8974: Deleted songs cant be "jump to" in osu!direct
**Describe the bug:** 
Currently when youre in osu!!direct and you stumble across a deleted song which is in your database there isnt a way to jump said song

image

The button should be accessable when you have the song.
its gone on the beatmap page fully as well. 

image

**osu!lazer version:** 
2d49210

**Logs:**
Shouldnt be needed?

LittleEndu:
Fixes #8974

Download buttons are also used by the game to present the beatmap in song select. If you already have downloaded a beatmap that others no longer can, you should be able to use the download button to present said beatmap.

This PR adds a simple check of local availability for osu!direct and beatmap info overlay.


peppy:
I made a change to cover the case where an import occurs while a button is already displayed. Your code would have left the button disabled.

peppy (approved):

PR #8981, opened 2020-05-09
Use strongly-typed JsonConverter

huoyaoyuan:
Same as ppy/osu-framework#3532


osu.Game/IO/Serialization/Converters/TypedListConverter.cs
```diff
@@ -14,7 +14,7 @@ namespace osu.Game.IO.Serialization.Converters
     /// reconstruct the objects with their original types.
     /// </summary>
     /// <typeparam name="T">The type of objects contained in the <see cref="List{T}"/> this attribute is attached to.</typeparam>
-    public class TypedListConverter<T> : JsonConverter
+    public class TypedListConverter<T> : JsonConverter<List<T>>
```
smoogipoo, line 17: `IList<T>` will fix the CI errors here (covariance).

huoyaoyuan: In fact it doesn't fix the error.
I'm just obeying the type check in `CanConvert`. It seems to be violated somewhere.

huoyaoyuan: Actually `IEnumerable<T>` fixes it. Inspired by the `(IEnumerable<T>)value` cast in `WriteJson`.

osu.Game/IO/Serialization/Converters/TypedListConverter.cs
```diff
@@ -9,12 +9,12 @@
...
     /// <summary>
-    /// A type of <see cref="JsonConverter"/> that serializes a <see cref="List{T}"/> alongside
+    /// A type of <see cref="JsonConverter"/> that serializes an <see cref="IEnumerable{T}"/> alongside
     /// a lookup table for the types contained. The lookup table is used in deserialization to
     /// reconstruct the objects with their original types.
     /// </summary>
-    /// <typeparam name="T">The type of objects contained in the <see cref="List{T}"/> this attribute is attached to.</typeparam>
-    public class TypedListConverter<T> : JsonConverter
+    /// <typeparam name="T">The type of objects contained in the <see cref="IEnumerable{T}"/> this attribute is attached to.</typeparam>
+    public class TypedListConverter<T> : JsonConverter<IEnumerable<T>>
```
bdach, line 17: Honestly I don't really like this entire serialiser changing to `IEnumerable<T>`. To begin with, if we're doing that, it should probably change its name, but what I'm more concerned about is that this API is potentially unsafe to use.

There is no guarantee that an `IEnumerable` is finite - you could pass in an infinite yield iterator here and it would start trying to serialise it and at some point run out of memory. If `ICollection` were covariant then I'd advocate for that as it'd at least make sense semantically, but it is invariant, so that's a no-go. Due to that we can't even use `CanConvert` here, so I'd probably just advocate for reverting to the old code.

huoyaoyuan: > You could probably use `CanConvert` and check against the non-generic `IList/ICollection`, though.

The provided generic base class banned customizing `CanConvert`.

The real reason is that, the converter **is already used on an `IEnumerable<T>` member** somewhere, which caused unit test to fail. It's more sanitized to change the type of that member.

smoogipoo: It isn't used on an `IEnumerable<T>` member anywhere. That would've actually violated the previous `CanConvert` condition. As far as I remember it's only actually used inside `Beatmap<T>`.

I highly agree that this can't be `IEnumerable<T>` for the reasons previously mentioned. `IReadOnlyList<T>` seems like the best alternative (I must've mixed that up when I suggested `ILIst<T>` previously).

huoyaoyuan: Sure. `IList<T>` cannot pass the test, but `IReadonlyList<T>` can.

bdach (approved):
seems good 👍

peppy (approved):

PR #8982, opened 2020-05-09
Fix null reference causing hard freeze if game is forcefully closed during disclaimer

peppy:
As usual, missed this due to debug vs distribution differences :(.


bdach (approved):
can repro on master w/ release config, can't repro with this change, so seems good 👍

smoogipoo (approved):

PR #8999, opened 2020-05-11
BeatmapListingOverlay auto-pagination reimplementation

namaenonaimumei:
This is heavily a reimplementation of #8762, that seemed to be inactive for a bit now and differentiated with current master's overall `BeatmapListing` structure

* Follows current master `BeatmapListingOverlay` being split to `BeatmapListingFilterControl` - which also seemed to me making `BeatmapListingResultsDisplay` as its own class unnecessary
* Follows logic and naming scheme of current master's implementation (only adds `BeatmapListingPager` class)
* Tested everything I could think of and followed latest code review recommendations from #8762 for most part

I didn't know whether it's fine to create a new pull request, but considering file difference didn't really make much sense when added to existing PR commits, decided to create a new one.
If that is not fine, let me now, since this is my first pull request here and I've made quite a few presumptions

Most comments are likely unnecessary - I've left them temporary, in case it helps anyone wishing to review


osu.Game/Online/API/Requests/ResponseWithCursor.cs
```diff
@@ -13,4 +13,13 @@ public abstract class ResponseWithCursor
         [JsonProperty("cursor")]
         public dynamic CursorJson;
     }
+
+    public abstract class ResponseWithCursor<T> : ResponseWithCursor where T : class
+    {
+        /// <summary>
+        /// Cursor deserialized into T class type (cannot implicitly convert type to object using raw Cursor)
+        /// </summary>
```
bdach, line 21: This xmldoc is mostly unnecessary.

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -5,11 +5,21 @@
...
 {
     public class SearchBeatmapSetsRequest : APIRequest<SearchBeatmapSetsResponse>
     {
+        public class Cursor
+        {
+            [JsonProperty("approved_date")]
+            public string ApprovedDate;
+
+            [JsonProperty("_id")]
+            public string Id;
```
bdach, line 20: As I said in the discussion of the original PR this is not enough. If you change the sort criteria the cursor will contain different fields.

osu.Game/Overlays/BeatmapListing/BeatmapListingFilterControl.cs
```diff
@@ -24,6 +22,8 @@ public class BeatmapListingFilterControl : CompositeDrawable
     {
         public Action<List<BeatmapSetInfo>> SearchFinished;
         public Action SearchStarted;
+        /// <summary> List of currently displayed beatmap entries </summary>
```
bdach, line 25: Again, mostly unnecessary, but if you're going to add xmldoc, please do it consistently and don't single-line it like this.
```suggestion
        /// <summary>
        /// List of currently displayed beatmap entries.
        /// </summary>
```

osu.Game/Overlays/BeatmapListing/BeatmapListingPager.cs
```diff
@@ -0,0 +1,92 @@
...
+        private readonly SortCriteria sortCriteria;
+        private readonly SortDirection sortDirection;
+
+        public event PageFetchHandler PageFetched;
+        private SearchBeatmapSetsRequest getSetsRequest;
+        private SearchBeatmapSetsResponse lastResponse;
+
+        /// <summary> Reports end of results </summary>
+        private bool isLastPageFetched = false;
+        /// <summary> Job in process lock flag </summary>
```
bdach, line 29: This does not parse to me as an English sentence.

bdach (dismissed):
* Not all comments from the original PR have been addressed.
* Most of the xmldoc added does not read well.
* There are also code style issues that CI will probably point out for you.

Also as a general note there isn't really a need to squash commits into one, much less with the very weird commit message chosen here (as it does not describe the actual changes made here in any way).

osu.Game/Online/API/Requests/ResponseWithCursor.cs
namaenonaimumei: To removal

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
namaenonaimumei: Will be fixed

osu.Game/Overlays/BeatmapListing/BeatmapListingFilterControl.cs
namaenonaimumei: To removal

osu.Game/Overlays/BeatmapListing/BeatmapListingPager.cs
namaenonaimumei: It does to me - C/++ terminology took liberty here I think, sorry
To removal anyway so doesn't matter

namaenonaimumei:
* All these (added by me) xmldoc comments were temporary, so I wouldn't worry about `<summary>` single liners (didn't want to unnecessary increase file differences), will remove all next commit
* Code style issues are all misses on my end since I didn't use any additional formatting checkers etc.
* For cursor:
I did notice it in #8762 review aswell, but I misunderstood it as current APIv2 state limitations
I see what you meant now, will be fixed in a moment

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
namaenonaimumei: I think this can be consider fixed for now, as far as I can tell

bdach: I'm not sure how you can assert that. I see no new pushed changes that would introduce the other possible cursor fields.

namaenonaimumei: I forgot to public push, sorry - should be visible now

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -70,8 +91,24 @@ protected override WebRequest CreateWebRequest()
...
+                if (cursor.Ranked != null)
+                    req.AddParameter("cursor[approved_date]", cursor.Ranked);
+                if (cursor.Rating != null)
+                    req.AddParameter("cursor[rating]", cursor.Rating);
+                if (cursor.Plays != null)
+                    req.AddParameter("cursor[play_count]", cursor.Plays);
+                if (cursor.Favourites != null)
+                    req.AddParameter("cursor[favourite_count]", cursor.Favourites);
+                if (cursor.Relevance != null)
+                    req.AddParameter("cursor[_score]", cursor.Relevance);
```
bdach, line 109: I don't think this is a good way to resolve the issue. Consider the possibility of new fields being added to API at some point in time - without game-side changes sorting will be broken for those new fields.

I'm still of the opinion that the game should either not interpret the cursor at all and just send back what it received (which requires API-side changes), or if we decide to handle this game-side, it should be in a way that requires no manual intervention in the future in case new fields are introduced, as long as API still accepts the cursor back the way it currently does (i.e. the field naming convention is still the same).

namaenonaimumei:
* Sorting should be working now

Will look whether I can think of something smarter to remove this cluster of if statements
Not sure if current is fine by your standards

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
namaenonaimumei: I'm not really sure what I can do about it on my end

For now I can only think of one more way to do it:
* Only `"_id"` parameter would be parsed for all responses instead and second parameter would be taken no matter the property name
* But that presumes that there will be always only 2 cursor fields, which isn't guarantee aswell, so it doesn't sound any more robust

As I said, I'm not really sure as to what I'm expected to do about this one
I may have missed the point again, though

bdach: The way I'd go forward with this is to throw away the notion of this (or *any*) particular cursor and add some code path that would leverage JSON deserialisation to yield us the JSON object -> query string mapping. I'm not saying it should handle arbitrary JSON object graphs because that's overkill - I guess we can assume the object is a flat object (nothing nested) for now.

So in general I'd like to see something that can, for all requests, turn the cursor part of the response as follows:

```js
{
    // rest of response
    "cursor":
    {
        "foo": "xyz",
        "bar": "baz"
    }
}
```

into this query string:

```
?cursor[foo]=xyz&cursor[bar]=baz
```

*without* interpreting what `foo`, `bar`, `baz` and `xyz` are or mean.

The reason I'm pushing for this is that there will be more cursors for more API requests and if we can save code by not having specialised structures for each requests I'm all for it. In fact some other currently implemented cursors are probably broken but not used (haven't checked myself but I wouldn't be surprised).

namaenonaimumei: Would this be fine by you (if I understand correctly)

osu.Game/Online/API/Requests/SearchBeatmapSetsRequest.cs
```diff
@@ -89,27 +68,10 @@ protected override WebRequest CreateWebRequest()
...
-                    req.AddParameter("cursor[play_count]", cursor.Plays);
-                if (cursor.Favourites != null)
-                    req.AddParameter("cursor[favourite_count]", cursor.Favourites);
-                if (cursor.Relevance != null)
-                    req.AddParameter("cursor[_score]", cursor.Relevance);
-
-                req.AddParameter("cursor[_id]", cursor.Id);
-            }
+                req.AddParameter("cursor[" + x.Key + "]", x.Value?.ToString());
+            });
```
bdach: Definitely better in my opinion, but I'd make this an extension method of `WebRequest` so that it can be reused in other places:

```csharp
public static void AddCursor(this WebRequest webRequest, Cursor cursor)
{
    // this impl here
}
```

namaenonaimumei: Not sure whether I understand correctly, so double checking
I'm guessing you didn't mean as in adding extension method directly in osu-framework's `osu.Framework.IO.Network.WebRequest` class
Should it be still part of `SearchBeatmapSetsRequest.cs` as in something like this
```
public static class ExtensionClass
    {
        public static void AddCursor(this WebRequest webRequest, SearchBeatmapSetsRequest.Cursor cursor)
        {
            cursor?.Properties.ForEach(x =>
            {
                webRequest.AddParameter("cursor[" + x.Key + "]", x.Value.ToString());
            });
        }
    }
```
Or is there some namespace inside osu.Game for storing extensions like this already
Not sure where I'm expected to add it and whether to move Cursor class aswell

namaenonaimumei: Or just do something like creating `WebRequestExtensions.cs` in `osu.Game.Extensions` namespace

bdach: The latter. The cursor is osu!-specific and adding anything to do with it to framework does not make sense.

namaenonaimumei: I'm not sure whether this is a clean solution, at least structure-wise
But committed for now

namaenonaimumei:
* Fixed remaining code style issues
I believe all requested changes should've been addressed by now

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -112,27 +123,51 @@ private void onSearchStarted()
...
             if (!beatmaps.Any())
             {
-                LoadComponentAsync(new NotFoundDrawable(), addContentToPlaceholder, (cancellationToken = new CancellationTokenSource()).Token);
+                LoadComponentAsync(notFoundContent, addContentToPlaceholder, (cancellationToken = new CancellationTokenSource()).Token);
                 return;
             }
 
-            var newPanels = new FillFlowContainer<BeatmapPanel>
+            //New query case
+            if (!shouldAddNextPage)
```
peppy, line 134: flimsy as hell. try scrolling while you are fetching more. it'll replace existing panels.

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -112,27 +123,51 @@ private void onSearchStarted()
...
             {
-                RelativeSizeAxes = Axes.X,
-                AutoSizeAxes = Axes.Y,
-                Spacing = new Vector2(10),
-                Alpha = 0,
-                Margin = new MarginPadding { Vertical = 15 },
-                ChildrenEnumerable = beatmaps.Select<BeatmapSetInfo, BeatmapPanel>(b => new GridBeatmapPanel(b)
+                //Spawn new child
+                var newPanels = new FillFlowContainer<BeatmapPanel>
```
peppy, line 137: not actually "new" panels sinc eyou're appending to existing list. doesn't work.

peppy:
I'm just going to refactor this one myself, too many issues to warrant a productive review.

peppy:
Not even sure why there's a "pager" class

bdach:
> Not even sure why there's a "pager" class

I'll take the blame for that one. Originally in v1 of this PR by the original author the pagination logic was strewn all around the place in the root overlay and I thought it was really messy, so I suggested factoring it out.

If you do end up changing that then I'll have a look at the result and draw my conclusions from that for the future.

peppy:
It feels like too many levels of abstraction. I'm combining it into the filter control.

peppy:
@bdach want to give that a look over?

peppy:
is the cursor generic for a reason?

bdach:
> @bdach want to give that a look over?

On a quick look, yeah, looks better.

> is the cursor generic for a reason?

It no longer needs to be if you're fine with the `IDictionary<string, JToken>`. Would get rid of the `[UsedImplicitly]`, too - but I'd probably move namespace to something closer to API structures (and maybe the extension too, not a 100% on that)

peppy:
Should be good to go. @smoogipoo requesting a review on this.

smoogipoo (approved):
Seems fine to me for now

PR #9015, opened 2020-05-13
Profile overlay section display not always correct

### LINKED ISSUE #8704: Profile overlay section display not always correct
**Describe the bug:** 
Upper bar (navigation bar, i guess) in in-game user profile shows that i'm at "Kudosu!" section, while i'm still at "Beatmaps". I scrolled about 3 screens down from start of "Beatmaps" section, and it started to occur.
osu_2020-04-10_08-04-16
**osu!lazer version:** 
2020.408.1
**Logs:**
Don't think it needs any

ghost:
With this commit, the selected section changes as soon as the screen top border enters the new section, instead of choosing the closer section.

### DIFF
--- a/osu.Game.Rulesets.Taiko/UI/TaikoPlayfieldAdjustmentContainer.cs
+++ b/osu.Game.Rulesets.Taiko/UI/TaikoPlayfieldAdjustmentContainer.cs
@@ -13,18 +13,16 @@ namespace osu.Game.Rulesets.Taiko.UI
         private const float default_relative_height = TaikoPlayfield.DEFAULT_HEIGHT / 768;
         private const float default_aspect = 16f / 9f;
 
-        public TaikoPlayfieldAdjustmentContainer()
-        {
-            Anchor = Anchor.CentreLeft;
-            Origin = Anchor.CentreLeft;
-        }
-
         protected override void Update()
         {
             base.Update();
 
             float aspectAdjust = Math.Clamp(Parent.ChildSize.X / Parent.ChildSize.Y, 0.4f, 4) / default_aspect;
             Size = new Vector2(1, default_relative_height * aspectAdjust);
+
+            // Position the taiko playfield exactly one playfield from the top of the screen.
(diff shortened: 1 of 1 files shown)


bdach:
Please do not close and reopen pull requests after a CI failure, and also please do not edit code via the web interface. This makes me have very little (if not zero) confidence in the changes introduced here.

Fukashi13:
I'm sorry, this is my first contribution to a Git project. I pasted the changes from web because the folder was a bit in a mess and I wanted to get sure that only this file gets commited

bdach:
In that case I would recommend that you familiarise yourself with git and pull request workflow (staging changes, resetting files, committing, pushing) before contributing further. Otherwise this is going to be too much of a pain to manage.

Fukashi13:
Ok, won't happen again, thanks for the tipps

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -190,14 +190,13 @@ protected override void UpdateAfterChildren()
                 headerBackgroundContainer.Y = ExpandableHeader?.Y ?? 0;
 
                 T bestMatch = null;
-                float minDiff = float.MaxValue;
-                float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
```
peppy, line 194: why was this moved out of a constant?

bdach:
@Fukashi13 please do not force-push either until there's a very good reason to do so. Force-pushing does not work well with CI and should only be used in extraordinary situations.

peppy:
also don't PR from master.

Fukashi13:
> @Fukashi13 please do not force-push either until there's a very good reason to do so. Force-pushing does not work well with CI and should only be used in extraordinary situations.

I squashed the commits made on web

osu.Game/Graphics/Containers/SectionsContainer.cs
Fukashi13: Is it ok now?

PR #9018, opened 2020-05-13
Allow disabling SkipOverlay through AllowGameplayOverlays.

Game4all:
Forgot to include SkipOverlay in previous PR; thus opening this as a follow-up.

# Summary

- This allows to disable the display of `SkipOverlay` through `DrawableRuleset.AllowGameplayOverlays` property.

### DIFF
--- a/osu.Game/Screens/Play/SkipOverlay.cs
+++ b/osu.Game/Screens/Play/SkipOverlay.cs
@@ -24,13 +24,14 @@ using osu.Game.Input.Bindings;
 
 namespace osu.Game.Screens.Play
 {
-    public class SkipOverlay : VisibilityContainer, IKeyBindingHandler<GlobalAction>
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Play/SkipOverlay.cs
```diff
@@ -39,6 +39,8 @@ public class SkipOverlay : VisibilityContainer, IKeyBindingHandler<GlobalAction>
         [Resolved]
         private GameplayClock gameplayClock { get; set; }
 
+        public bool Disabled { get; set; }
```
peppy, line 42: This is not how this should be gone about. Internal use (inside of SkipOverlay) of alpha should be moved to a nested container.

peppy (dismissed):

smoogipoo (dismissed):
lgtm

smoogipoo (approved):

peppy (approved):

Issue #8420, opened 2020-03-23
Test runs leave behind files from headless host storage


bdach:
Running a test full test suite locally can result in having lots of leftover files in the execution directory, which can build up to as much as gigabytes - at least on Windows - if left unattended for long enough (personally racked up about ~25 GB).

Some of this is due to the game, some of it due to the framework. For instance, the framework does not dispose of font stores in `Game`, therefore i.e. holding references to the font cache when the headless host is cleaning up, and as such the cleanup partially fails due to the file being open by the same process, leaving behind a footprint of 1 MB per instantiated host. osu!-side, sqlite database files also stay behind after test runs, making up about ~200 KB per each test host.

Those are the two main cases of leftover storage I can see - there might be more.

I attempted a *very* simplistic fix by changing framework's disposal in `Game` to this:

```csharp
        protected override void Dispose(bool isDisposing)
        {
            Audio?.Dispose();
            Audio = null;

            Fonts?.Dispose();
            Fonts = null;

            localFonts?.Dispose();
            localFonts = null;

            Resources?.Dispose();
            Resources = null;

            Textures?.Dispose();
            Textures = null;

            base.Dispose(isDisposing);
        }
```

but it only fixes the framework side part of the problem and additionally causes test failures game-side that I haven't been able to investigate in detail yet, therefore I figured I'd open an issue even if I eventually get to resolving this myself.

The overall fingerprint of leftover headless host storages in one full game-side run is about 1.2 GB.

peppy:
This may be specific to the test runner / platform you're using. On macOS the only files left behind are the occasional logs (~200bytes per run). Are you using `dotnet test` or an nunit IDE plugin to run? Also which OS?

bdach:
I'm seeing this particular behaviour on Windows on unit runs, both by `dotnet test` and through rider.

peppy:
I think fixing the framework side would be a good start.

bdach:
[...]

PR #9020, opened 2020-05-14
Add ability to change data folder path

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/9027

### DIFF
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -87,6 +87,9 @@ namespace osu.Game.IO
 
                 dir.Delete(true);
             }
+
+            if (target.GetFiles().Length == 0 && target.GetDirectories().Length == 0)
+                target.Delete();
         }
 
         private static void copyRecursive(DirectoryInfo source, DirectoryInfo destination, bool topLevelExcludes = true)
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -51,6 +51,9 @@ namespace osu.Game.IO
             if (source.FullName == destination.FullName)
                 throw new ArgumentException("Destination provided is already the current location", nameof(newLocation));
 
+            if (destination.FullName.Contains(source.FullName))
+                throw new ArgumentException("Destination provided is inside the source", nameof(newLocation));
+
             // ensure the new location has no files present, else hard abort
             if (destination.Exists)
             {
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -48,10 +48,14 @@ namespace osu.Game.IO
             var source = new DirectoryInfo(GetFullPath("."));
             var destination = new DirectoryInfo(newLocation);
 
-            if (source.FullName == destination.FullName)
+            // using Uri is the easiest way to check equality and contains (https://stackoverflow.com/a/7710620)
+            var sourceUri = new Uri(source.FullName + Path.DirectorySeparatorChar);
+            var destinationUri = new Uri(destination.FullName + Path.DirectorySeparatorChar);
+
+            if (sourceUri == destinationUri)
                 throw new ArgumentException("Destination provided is already the current location", nameof(newLocation));
 
-            if (destination.FullName.Contains(source.FullName))
+            if (sourceUri.IsBaseOf(destinationUri))
                 throw new ArgumentException("Destination provided is inside the source", nameof(newLocation));
 
             // ensure the new location has no files present, else hard abort
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -48,11 +48,14 @@ namespace osu.Game.IO
             var source = new DirectoryInfo(GetFullPath("."));
             var destination = new DirectoryInfo(newLocation);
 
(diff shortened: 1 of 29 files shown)


MATRIX-feather:
In osu.Game/IO/OsuStorage.cs: line 19 :
``````
internal static readonly string[] IGNORE_DIRECTORIES = { "cache" }
``````
I think `osu-lazer` should be added to `IGNORE_DIRECTORIES` to avoid osu from getting stuck into a loop and finally make the disk runs out of space if the user selects the current data folder as a new location.

**Comment by peppy** [comment]
there's already a check for current directory. could you please provide reproduction for this?

**Comment by MATRIX-feather** [comment]
> there's already a check for current directory. could you please provide reproduction for this?

@peppy 
[simplescreenrecorder-2020-05-14_21.03.56.zip](https://github.com/ppy/osu/files/4628532/simplescreenrecorder-2020-05-14_21.03.56.zip)

**Comment by peppy** [comment]
thanks! that clears things up

File: osu.Game/IO/OsuStorage.cs [hunk 1]
```diff
@@ -48,11 +48,17 @@ public void Migrate(string newLocation)
             var source = new DirectoryInfo(GetFullPath("."));
             var destination = new DirectoryInfo(newLocation);
 
+            if (source.FullName == destination.FullName)
+                throw new ArgumentException("Destination provided is already the current location", nameof(newLocation));
+
+            if (destination.FullName.Contains(source.FullName))
```
bdach, line 54: This check is broken if there are two sibling directories
```
C:/somedir
C:/somedir2
```
as the test failure demonstrates (the paths returned by `FullName` don't have trailing slashes - at least on Windows, therefore there is a wrongly-matching common prefix).

osu.Game/Overlays/Settings/Sections/Maintenance/MigrationRunScreen.cs
```diff
@@ -0,0 +1,115 @@
...
+                },
+            };
+
+            Beatmap.Value = Beatmap.Default;
+
+            migrationTask = Task.Run(PerformMigration)
+                                .ContinueWith(t =>
+                                {
+                                    if (t.IsFaulted)
+                                        Logger.Log($"Error during migration: {t.Exception?.Message}", level: LogLevel.Error);
```
bdach, line 91: Arguably should be

```suggestion
                                        Logger.Error(t.Exception, "Error during migration: {t.Exception?.Message}");
```

as the current version completely discards stack traces, which means we won't be able to debug user reports in any of this.

bdach (dismissed):

bdach:
I've tried this a couple of times on Windows and it doesn't work too robustly. For example, you can start the migration process with imports running, leading to this occurring:

image

but I also think I got it to do that once after the import was already cancelled (I'm pretty sure). It might mean there's some kind of connection leak during the import process (possibly related to the leftover `client.db` files I've seen in #8420?)

I also got this once:

```
2020-05-14 20:51:41 [error]: System.AggregateException: One or more errors occurred. (The process cannot access the file '\\?\D:\lazertest\logs' because it is being used by another process.)
2020-05-14 20:51:41 [error]: ---> System.IO.IOException: The process cannot access the file '\\?\D:\lazertest\logs' because it is being used by another process.
2020-05-14 20:51:41 [error]: at System.IO.FileSystem.RemoveDirectoryInternal(String fullPath, Boolean topLevel, Boolean allowDirectoryNotEmpty)
2020-05-14 20:51:41 [error]: at System.IO.FileSystem.RemoveDirectoryRecursive(String fullPath, WIN32_FIND_DATA& findData, Boolean topLevel)
2020-05-14 20:51:41 [error]: at System.IO.FileSystem.RemoveDirectory(String fullPath, Boolean recursive)
2020-05-14 20:51:41 [error]: at System.IO.DirectoryInfo.Delete(Boolean recursive)
2020-05-14 20:51:41 [error]: at osu.Game.IO.OsuStorage.deleteRecursive(DirectoryInfo target, Boolean topLevelExcludes) in D:\Open Source\osu\osu.Game\IO\OsuStorage.cs:line 91
2020-05-14 20:51:41 [error]: at osu.Game.IO.OsuStorage.Migrate(String newLocation) in D:\Open Source\osu\osu.Game\IO\OsuStorage.cs:line 73
2020-05-14 20:51:41 [error]: at osu.Game.OsuGameBase.Migrate(String path) in D:\Open Source\osu\osu.Game\OsuGameBase.cs:line 364
```

Haven't debugged too much today as my time is running out for the day so I just wanted to see how well it works for now. Might have a deeper look at it on the weekend.

peppy:
yep, it’s not perfect and likely never will be. can probably add automatic retry logic in the future.

another plan is to have it lock the ability for database connections to be used during the process (after the connection flush)

osu.Game/Overlays/Settings/Sections/Maintenance/MigrationRunScreen.cs
peppy: Using `Logger.Error` actually fires the error to our server handler right now, so we avoid using it in user scenarios like this. Maybe not the best handling there.

osu.Game/IO/OsuStorage.cs
peppy: Fixed in #9027

smoogipoo (approved):

PR #9021, opened 2020-05-14
Implement beat snapping grid for mania's editor

smoogipoo:
Trying something a bit different with this one, since the model of osu!'s distance snapping grid doesn't really work (grid lines emanating from the last hitobject).

This does processing every update frame, but it's pretty fast. Should be fine for now?

- [x] Depends on https://github.com/ppy/osu/pull/9070

### DIFF
--- a/osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs
+++ b/osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs
@@ -22,10 +22,7 @@ namespace osu.Game.Rulesets.Mania.Edit
 {
     public class ManiaBeatSnapGrid : Component
     {
-        /// <summary>
-        /// The brightness of bar lines one beat around the time range from <see cref="SetRange"/>.
-        /// </summary>
-        private const float first_beat_brightness = 0.5f;
+        private const double visible_range = 1500;
 
         [Resolved]
         private IManiaHitObjectComposer composer { get; set; }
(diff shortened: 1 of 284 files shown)


bdach (commented):
Read through this, ran this and aside from tests failing it seems fine. The parts that gave me pause (`GetSnappedPosition` and the triple list alloc in `setRange`) I couldn't find any conclusively better alternative for so I think this could do for now?

smoogipoo:
It could be reduced to one list, not sure how important that is though - it's probably collectively <256B allocated per-frame.

peppy:
@smoogipoo I've adjusted this to only construct the required lines now. Performance is a bit weird when recreating, which is mostly due to bindable unbind logic (that will be fixed by pooling the hitobjects) so I see this being an easier future optimisation than what we were working with before.

smoogipoo:
Some change you've made has completely ruined the performance of this. It almost looks like it's creating new lines with every mouse movement and I can't maintain >8fps at 1/16th.

peppy:
Performance is lower on mouse movement, but I didn't see it go that low. Which beatmap are you testing on?

smoogipoo:
https://osu.ppy.sh/beatmapsets/174381#mania/421066

peppy:
Also please re-read the comment above yours which touches on performance.

smoogipoo:
Yeah which is exactly why I didn't want to recreate lines. We can't have that, it's completely unusable - not only can I not maintain 60fps at 1/4 (which that map is mapped in) but it's also causing huge amounts of GC stutters.

peppy:
Let me see if I can reduce the overhead (it should only be recreating when moving to a new tick – sounds like it might be doing more than that based on what you're seeing?), since I think this is the direction we will want code-wide going forward.

Also if you go back to the commit before this change on the same beatmap, then try to change the beat divisor you'll probably see it freezing the game for 20-60seconds, which is why I made this change in the first place.

peppy:
@smoogipoo i've added very basic pooling and it looks promising – initial frame rates are almost no drops. unfortunately something is causing it to slow down over time and I'm still tracking down the cause.

image

Unsure of the exact source yet, but just pushing/dropping this here in case you can pinpoint it.

osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            while (time < range.end + visible_range)
+            {
+                var nextTimingPoint = beatmap.ControlPointInfo.TimingPointAt(time);
+
+                // switch to the next timing point if we have reached it.
+                if (nextTimingPoint != timingPoint)
+                {
+                    beat = 0;
```
smoogipoo, line 129: I believe this should also set `time = nextTimingPoint.Time`?

osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs
```diff
@@ -0,0 +1,238 @@
...
+
+            if (!lineCache.IsValid)
+            {
+                lineCache.Validate();
+                createLines();
+            }
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
```
smoogipoo, line 77: Probably should be closer to the top

smoogipoo (commented):
Seems ok otherwise

osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs
peppy: good catch.

smoogipoo:
`TestSceneHitObjectComposer` is a real test. That test doesn't have any timing points so it seems the  new code is running into an infinite loop.

Rest of the code looks good.

ghost (commented):
Seems fine to me, maybe you can try and optimize it here and there (nothing comes to mind just a suggestion) but it works

peppy:
While I'm generally okay with the display component of this now, the actual snapping logic flow is impossible to follow. I'm going to retry this in a separate PR to see if I can come up with a better approach, and if so, will also PR the visual component of this separately.

I don't think they need to be attached at all (and will better benefit from the existing editor structure if they aren't).

peppy:
Ended up just pushing my changes to make this dependent on #9070, as I've made quite a few fixes along the way and am pretty confident in direction.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -257,7 +257,7 @@ private void updateState(ArmedState newState, bool force = false)
                 }
             }
 
-            if (state.Value != ArmedState.Idle && LifetimeEnd == double.MaxValue || HitObject.HitWindows == null)
+            if (LifetimeEnd == double.MaxValue && (state.Value != ArmedState.Idle || HitObject.HitWindows == null))
```
peppy, line 260: I'm pretty confident this change is more correct than what we had, but make sure to double-check you're OK with it.

smoogipoo (commented):
Seems pretty fine to me :+1:

PR #9023, opened 2020-05-14
Fix overlay section display highlighting wrong section

### LINKED ISSUE #8704: Profile overlay section display not always correct

ghost:
Closes #8704 The selected section changes as soon as the screen top border enters the section instead of choosing the nearer section. The problem was that if the section is big enough, the current scroll can be closer to the next section while the screen is still filled with the current section.

### DIFF
--- a/osu.Game/Graphics/Containers/SectionsContainer.cs
+++ b/osu.Game/Graphics/Containers/SectionsContainer.cs
@@ -189,23 +189,18 @@ namespace osu.Game.Graphics.Containers
                 headerBackgroundContainer.Height = (ExpandableHeader?.LayoutSize.Y ?? 0) + (FixedHeader?.LayoutSize.Y ?? 0);
                 headerBackgroundContainer.Y = ExpandableHeader?.Y ?? 0;
 
-                T bestMatch = Children.FirstOrDefault();
-                float minDiff = float.MinValue;
                 float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
+                Func<T, float> diff = section => scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
 
-                foreach (var section in Children)
+                if (scrollContainer.IsScrolledToEnd())
                 {
-                    float diff = scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
-
-                    if (minDiff < diff && diff < 0)
-                    {
-                        minDiff = diff;
-                        bestMatch = section;
-                    }
+                    SelectedSection.Value = Children.LastOrDefault();
                 }
-
-                if (bestMatch != null)
-                    SelectedSection.Value = bestMatch;
+                else
+                {
+                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 0).LastOrDefault()
+                        ?? Children.FirstOrDefault();
+                }
             }
         }
 
--- a/osu.Game/Graphics/Containers/SectionsContainer.cs
+++ b/osu.Game/Graphics/Containers/SectionsContainer.cs
@@ -189,15 +189,15 @@ namespace osu.Game.Graphics.Containers
                 headerBackgroundContainer.Height = (ExpandableHeader?.LayoutSize.Y ?? 0) + (FixedHeader?.LayoutSize.Y ?? 0);
                 headerBackgroundContainer.Y = ExpandableHeader?.Y ?? 0;
 
-                T bestMatch = null;
-                float minDiff = float.MaxValue;
+                T bestMatch = Children.FirstOrDefault();
+                float minDiff = float.MinValue;
                 float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
 
                 foreach (var section in Children)
                 {
-                    float diff = Math.Abs(scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset);
+                    float diff = scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
 
-                    if (diff < minDiff)
+                    if ((minDiff < diff) & (diff < 0))
                     {
                         minDiff = diff;
                         bestMatch = section;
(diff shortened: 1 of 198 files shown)


bdach:
Could you please elaborate why this was closed and opened anew a second time?

Fukashi13:
> Could you please elaborate why this was closed and opened anew a second time?

I tried squashing the commits and accidentally messed up the branches, and when I reset my fork the PR was automatically closed, so this time I (hopefully) did it right and made the commits to a separate branch

Fukashi13:
I'm sorry for the annoyances

bdach:
Again, I'm pretty sure you can't have done anything so serious as to outright require entirely reforking the project. Not to mention you shouldn't have been squashing in the first place, and even before that it seems you were pushing changes haphazardly.

The golden rule is not to push code until you're reasonably certain the change works and you've *tested* that it works.

Additionally as I said before, I really urge you to read up on git. It's going to be painful for both sides otherwise.

Fukashi13:
Ok

Fukashi13:
Btw the code works like this, you could only disagree with the design change / coding style

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -189,15 +189,15 @@ protected override void UpdateAfterChildren()
                 headerBackgroundContainer.Height = (ExpandableHeader?.LayoutSize.Y ?? 0) + (FixedHeader?.LayoutSize.Y ?? 0);
                 headerBackgroundContainer.Y = ExpandableHeader?.Y ?? 0;
 
-                T bestMatch = null;
-                float minDiff = float.MaxValue;
+                T bestMatch = Children.FirstOrDefault();
+                float minDiff = float.MinValue;
```
bdach, line 193: This no longer makes sense. This variable was named `minDiff` since it kept the minimum difference between the section position and current scroll position + offset. Now, as far as I can see, it stores the *biggest negative difference* between the two.

Not to mention, as far as I can see this *entire* loop below could be now entirely replaced with

```csharp
SelectedSection.Value = Children.TakeWhile(section => scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset < 0)
                                .LastOrDefault();

```

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -189,15 +189,15 @@ protected override void UpdateAfterChildren()
...
                 float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
 
                 foreach (var section in Children)
                 {
-                    float diff = Math.Abs(scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset);
+                    float diff = scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
 
-                    if (diff < minDiff)
+                    if ((minDiff < diff) & (diff < 0))
```
bdach, line 200: Don't use bitwise operations on booleans.

```suggestion
                    if (minDiff < diff && diff < 0)
```

bdach (dismissed):

bdach:
The only reason I haven't fixed this up myself is because I'm unconviced of the complete correctness of this fix UX-wise. Now, when you are scrolled to the end, the "Kudosu" tab will not highlight:

image

unless you forcefully drag the overlay so that the top of the section crosses the top of the scroll container:

image

I think at the least the last section should always be highlighted when the scroll container is at its end (or farther), but I'm not a 100% on that.

Fukashi13:
> The only reason I haven't fixed this up myself is because I'm unconviced of the complete correctness of this fix UX-wise. Now, when you are scrolled to the end, the "Kudosu" tab will not highlight:
> 
> image
> 
> unless you forcefully drag the overlay so that the top of the section crosses the top of the scroll container:
> 
> image
> 
> I think at the least the last section should always be highlighted when the scroll container is at its end (or farther), but I'm not a 100% on th

Fukashi13:
> The only reason I haven't fixed this up myself is because I'm unconviced of the complete correctness of this fix UX-wise. Now, when you are scrolled to the end, the "Kudosu" tab will not highlight:
> 
> image
> 
> unless you forcefully drag the overlay so that the top of the section crosses the top of the scroll container:
> 
> image
> 
> I think at the least the last section should always be highlighted when the scroll container is at its end (or farther), but I'm not a 100% on that.

This issue was there before, I think there should be an additional section for when you scroll to the botton, so the last section that cant he reached gets selected

TheWildTree:
That issue has definitely been there for a while now, see #7669.

osu.Game/Graphics/Containers/SectionsContainer.cs
Fukashi13: Yeah, I didn't know those methods existed...

Fukashi13:
If I remember it correctly, my change resembles the behaviour of osu web (eventually ignoring the unreachable section)

osu.Game/Graphics/Containers/SectionsContainer.cs
Fukashi13: Under the hood it does exactly the same like my loop but whit SelectedSection.Value directly instead of bestMatch

Fukashi13:
Closes #7669 with the last commit, the last section of the list gets selected when you scroll to the bottom, else it behaves the same

Fukashi13:
> If I remember it correctly, my change resembles the behaviour of osu web (eventually ignoring the unreachable section)

Actually, another part of issue #7669, that the first section is selected even though you are above, while in web none is selected by default, can be solved by removing line 202 in my commit 6416ace, tell me if I should do it

smoogipoo (approved):
lgtm

PR #9024, opened 2020-05-14
Use unique video in map pool scene of tournament client

MiraiSubject:
This gives the mappool scene its own background instead of relying on the gameplay file. 
This PR also includes a rename of the SeedingResultRow class.


bdach (approved):
Looks fairly obvious, so should be fine unless it interferes with usage of the tournament client in official capacity in some way or other.

peppy (approved):

PR #9031, opened 2020-05-15
Avoid disabling a host-level bindable from osu! code

### LINKED ISSUE #8753: Crash when running OsuGame test in rulesets
**Describe the crash:** 
A crash will occur when selecting `OsuGame` test in ruleset tests (ruleset initialized using osu-templates) after the same test have loaded before in the same debug instance.

Rulesets that exhibit this problem (Maybe others too):
https://github.com/Altenhh/tau
https://github.com/LumpBloom7/osu-maimai

Video: https://streamable.com/rttmzx
image

**osu!lazer version:** 
lazer: 2020.413.0

**Exception details**
[exception.txt](https://github.com/ppy/osu/files/4473608/exception.txt)

**Side Note**
Exceptions occasionally happen when escaping to close the test instance, I can't consistently recreate it though as I can't really find a pattern to it. Not sure if related to this...

peppy:
- Closes #8753.

I had a look at a few different approaches to fixing this problem, but this seems like the path of least resistance (least likely to break in the future).

Other alternatives I assessed:
- Exposing `OsuGameTestScene` for consumption by templates
- Reimplementing similar local storage logic in template projects

Both may fix this one instance but would require updates in all template based projects and the same issue may surface elsewhere if we aren't careful.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
@@ -19,7 +19,6 @@ namespace osu.Game.Overlays.Settings.Sections.Input
         private readonly BindableBool rawInputToggle = new BindableBool();
         private Bindable<double> sensitivityBindable = new BindableDouble();
         private Bindable<string> ignoredInputHandler;
-        private SensitivitySetting sensitivity;
 
         [BackgroundDependencyLoader]
         private void load(OsuConfigManager osuConfig, FrameworkConfigManager config)
@@ -38,7 +37,7 @@ namespace osu.Game.Overlays.Settings.Sections.Input
                     LabelText = "Raw input",
                     Bindable = rawInputToggle
                 },
-                sensitivity = new SensitivitySetting
+                new SensitivitySetting
                 {
                     LabelText = "Cursor sensitivity",
                     Bindable = sensitivityBindable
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
```diff
@@ -60,7 +68,7 @@ private void load(OsuConfigManager osuConfig, FrameworkConfigManager config)
             if (RuntimeInfo.OS != RuntimeInfo.Platform.Windows)
             {
                 rawInputToggle.Disabled = true;
-                sensitivity.Bindable.Disabled = true;
+                sensitivityBindable.Disabled = true;
```
bdach, line 71: I was confused as to why this was the only line changed when the original report came from Windows (and why I couldn't repro it), but then I pulled up the entire file and noticed this lower down:

https://github.com/ppy/osu/blob/aec2520ef41155ea14e7bdf0c68fce4a45c7e3e1/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs#L89

so I guess what I'd like to propose in a long-winded way is to either revert this change or change both places? It's the same reference so it doesn't really matter either way, but probably better to keep consistent.

peppy: Nice catch, completely missed this one.

bdach (approved):
Looks good to me, the test failure I can't repro and it seems unrelated so 👍

smoogipoo (approved):
One case where this could go wrong is if the bindables are changed by other parts of the game, in which case settings wouldn't update. But I think this is probably safe enough and we can probably figure out something better with bindables in the future.

PR #9034, opened 2020-05-15
Add support to disable user cancelling on progress notifications

frenzibyte:
Gives better UX for progress notifications tracking tasks that cannot be cancelled, instead of silently finishing the task regardless of the notification state (currently happening with `SquirrelUpdateManager`, to be fixed in a follow-up PR):

- Design is open for feedback, currently following what most disabled buttons do afaik.

### DIFF
--- a/osu.Game/Overlays/Notifications/Notification.cs
+++ b/osu.Game/Overlays/Notifications/Notification.cs
@@ -162,30 +162,26 @@ namespace osu.Game.Overlays.Notifications
 
         public class NotificationCloseButton : OsuClickableContainer
         {
-            private readonly ShakeContainer shakeContainer;
-
             private Color4 hoverColour;
 
             public NotificationCloseButton()
             {
                 Colour = OsuColour.Gray(0.2f);
 
-                Children = new[]
+                SpriteIcon icon;
+                Child = icon = new SpriteIcon
                 {
-                    shakeContainer = new ShakeContainer
-                    {
-                        RelativeSizeAxes = Axes.Both,
-                        Child = new SpriteIcon
-                        {
-                            Anchor = Anchor.Centre,
-                            Origin = Anchor.Centre,
-                            Icon = FontAwesome.Solid.TimesCircle,
-                            RelativeSizeAxes = Axes.Both,
-                        }
-                    }
+                    Anchor = Anchor.Centre,
+                    Origin = Anchor.Centre,
+                    Icon = FontAwesome.Solid.TimesCircle,
+                    RelativeSizeAxes = Axes.Both,
                 };
 
-                Enabled.ValueChanged += _ => updateState();
+                Enabled.ValueChanged += e =>
+                {
+                    // Hide the button internally to avoid conflicting with notification hovering logic.
+                    icon.FadeTo(e.NewValue ? 1 : 0, 120);
+                };
             }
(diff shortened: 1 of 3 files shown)


bdach:
Not entirely sure this is the best direction UX-wise. The fact that the cancel button fades in when you hover over a non-cancellable notification, only to shake at you when you actually try to use it, feels a little weird. It probably shouldn't show in the first place.

frenzibyte:
> Not entirely sure this is the best direction UX-wise. The fact that the cancel button fades in when you hover over a non-cancellable notification, only to shake at you when you actually try to use it, feels a little weird. It probably shouldn't show in the first place.

Sounds like a simple another way to go through this, thought about it initially but decided to follow up with what is currently done for disabled buttons as I've stated above. will see how it feels.

osu.Game/Overlays/Notifications/ProgressNotification.cs
```diff
@@ -35,6 +39,27 @@ public float Progress
...
+            set
+            {
+                if (cancellable == value)
+                    return;
+
+                if (state == ProgressNotificationState.Cancelled && value == false)
+                    throw new InvalidOperationException("Attempting to disable user cancelling while the notification is already cancelled.");
+
+                cancellable = value;
+                CloseButton.Enabled.Value = cancellable;
```
bdach, line 59: can use compound assignment

```suggestion
                cancellable = CloseButton.Enabled.Value = cancellable;
```

osu.Game/Overlays/Notifications/Notification.cs
```diff
@@ -161,24 +160,27 @@ public virtual void Close()
...
+                    Anchor = Anchor.Centre,
+                    Origin = Anchor.Centre,
+                    Icon = FontAwesome.Solid.TimesCircle,
+                    RelativeSizeAxes = Axes.Both,
+                };
+
+                Enabled.ValueChanged += e =>
+                {
+                    // Hide the button internally to avoid conflicting with notification hovering logic.
+                    icon.FadeTo(e.NewValue ? 1 : 0, 120);
```
bdach, line 183: This doesn't work. The button still has `HoverClickSounds` on it, so it's making sounds if you click where it should be while it's invisible.

osu.Game/Overlays/Notifications/ProgressNotification.cs
```diff
@@ -43,17 +68,18 @@ protected override void LoadComplete()
...
         {
             get => state;
             set
             {
                 if (state == value) return;
 
+                // Enable back user cancelling if notification transitioned to
+                // cancelled state by the component using it. likely due to task failure.
+                if (value == ProgressNotificationState.Cancelled)
+                    Cancellable = true;
```
bdach, line 81: I find this logic strange and I'd go as far as to say even a little wrong semantically. When you read this it makes no sense why you would allow cancelling an already cancelled notification. If anything I'd say transitioning to `Cancelled` state and setting `Cancellable = true` should *both* enable the button, but *independently* of each other.

bdach (changes requested):

frenzibyte:
Since the current direction is to fix all cases where tasks that are un-cancellable to become cancellable, this PR will not benefit anything in general. Closing as useless.

PR #9035, opened 2020-05-15
Disable user-cancelling in squirrel update notification

frenzibyte:
- [ ] Depends on #9034 

Can be disabled for now until cancellation is supported in squirrel updating methods.


frenzibyte:
Closing as incorrect direction.

PR #9036, opened 2020-05-15
Fix applied platform/user offsets being incorrect when rate adjust mods are active

### LINKED ISSUE #8774: Audio offset does not scale properly with HT or DT
**Describe the bug:** 
When half time or double time is active, the global audio offset does not scale properly and leads to a mismatch in timing. Due to an issue with my soundcard I have a very high audio offset (-155ms) which means this effect is greatly exaggerated on my system, so it may not have been noticed by users with smaller offsets which would (I assume) only result in a tiny error in timing.

Edit: Just tested this and HT requires lowering the offset while DT requires increasing it, both by the same ratio as the speed. i.e. HT at 75% speed matched an offset of about -110 ms (which is ~75% of 155)

https://i.imgur.com/Vf8WBfr.mp4 Normal speed, no mods (ignore the misses at the start, had to get used to the lag from running OBS)
https://i.imgur.com/JfM2LxY.mp4 Half time 75% speed
Don't know how to get sound in imgur uploads but you can see the timing bar on the left. I didn't bother taking a video of double time but it does the same thing in the opposite direction (i.e. all the marks are on the bottom of the timing bar)

**osu!lazer version:** 
2020.413.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/4490141/database.log)
[network.log](https://github.com/ppy/osu/files/4490142/network.log)
[performance.log](https://github.com/ppy/osu/files/4490143/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4490144/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4490145/updater.log)

peppy:
- Closes https://github.com/ppy/osu/issues/8774

Offsets and adjustments are hard.

Simplest way to understand this: UO and platform adjustments are made to account for constant delays in audio hardware or drivers. Because we are adjusting the playback rate relative to real time, we cannot apply these offsets in the same rate-adjusted timeline without accounting for this rate adjustment (making the offsets equivalent real-time values).

I believe this may also potentially be the issue mappers were having in stable when timing at 25%, reporting that the offset was off by roughly `WINDOWS_PLATFORM_OFFSET` amount.

### DIFF
--- a/osu.Game/Screens/Play/GameplayClockContainer.cs
+++ b/osu.Game/Screens/Play/GameplayClockContainer.cs
@@ -78,10 +78,10 @@ namespace osu.Game.Screens.Play
 
             // Lazer's audio timings in general doesn't match stable. This is the result of user testing, albeit limited.
             // This only seems to be required on windows. We need to eventually figure out why, with a bit of luck.
-            platformOffsetClock = new FramedOffsetClock(adjustableClock) { Offset = RuntimeInfo.OS == RuntimeInfo.Platform.Windows ? 15 : 0 };
+            platformOffsetClock = new HardwareCorrectionOffsetClock(adjustableClock) { Offset = RuntimeInfo.OS == RuntimeInfo.Platform.Windows ? 15 : 0 };
 
             // the final usable gameplay clock with user-set offsets applied.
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
Makes sense

PR #9041, opened 2020-05-16
Add a back button to sub screens in the Tournament Client

### LINKED ISSUE #9038: Add a back button for any submenus in the Tournament Client
**Describe the new feature:** 

Right now there is one submenu in the tournament client: Team Editor -> Edit Seeding Results 
Currently it's not very obvious on how to return back to the Team Editor from the Seeding Editor screen. So I have two proposals on making it obvious that it's a submenu for getting back: 

Both of these proposals are in the screenshot below: 

image

One is having a `TourneyButton` on the control panel that changes the screen with the scene manager back. And the other is using the back button from the game's user interface in the bottom left with the same functionality. 

The latter implementation would need some more work, but could potentially be used more universally across the tournament client, if more "subscreens" get added.

On the contrary the former one would need to be implemented on a case-by-case basis on each screen. 

The latter implementation would also require some kind overflow for the scrolling container so that bottom item does not overlap with the back button, as shown in this screenshot: 

image

These changes right now are on my local fork of this repository, but I wanted to get some feedback on what the proper direction (or if a back button even is desired) for this situation.

MiraiSubject:
Closes #9038 

This also handles the input of Esc so it goes back to the parent screen. 

Included are two screenshots: 

dotnet_2020-05-16_04-06-12

dotnet_2020-05-16_04-06-15

Note that there's a padding on the ScrollContainer to compensate for the back button, so that all the items stay accessible.

### DIFF
--- a/osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
+++ b/osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
@@ -33,13 +33,15 @@ namespace osu.Game.Tournament.Screens.Editors
 
         protected ControlPanel ControlPanel;
 
-        protected virtual bool IsSubScreen => false;
-
-        protected virtual System.Type ParentScreen { get; set; }
-
+        private readonly TournamentScreen parentScreen;
         private BackButton backButton;
 
-        private System.Action backAction => () => sceneManager?.SetScreen(ParentScreen);
+        private System.Action backAction => () => sceneManager?.SetScreen(parentScreen.GetType());
+
+        protected TournamentEditorScreen(TournamentScreen parentScreen = null)
+        {
+            this.parentScreen = parentScreen;
+        }
 
         [BackgroundDependencyLoader]
         private void load()
@@ -51,7 +53,7 @@ namespace osu.Game.Tournament.Screens.Editors
                 Origin = Anchor.BottomLeft,
                 Action = () =>
                 {
-                    if (IsSubScreen)
+                    if (parentScreen != null)
                         backAction.Invoke();
                 }
             };
@@ -98,7 +100,7 @@ namespace osu.Game.Tournament.Screens.Editors
                 }
             });
 
-            if (IsSubScreen)
+            if (parentScreen != null)
                 backButton.Show();
 
             Storage.CollectionChanged += (_, args) =>
--- a/osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
+++ b/osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
@@ -70,13 +70,13 @@ namespace osu.Game.Tournament.Screens.Editors
                     RelativeSizeAxes = Axes.Both,
                     Anchor = Anchor.TopCentre,
                     Origin = Anchor.TopCentre,
-                    Padding = new MarginPadding { Bottom = backButton.Height },
                     Child = flow = new FillFlowContainer<TDrawable>
                     {
                         Direction = FillDirection.Vertical,
                         RelativeSizeAxes = Axes.X,
                         AutoSizeAxes = Axes.Y,
-                        Spacing = new Vector2(20)
+                        Spacing = new Vector2(20),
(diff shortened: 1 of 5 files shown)


osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
```diff
@@ -17,19 +20,42 @@
...
         protected abstract BindableList<TModel> Storage { get; }
 
         private FillFlowContainer<TDrawable> flow;
 
+        [Resolved(canBeNull: true)]
+        private TournamentSceneManager sceneManager { get; set; }
+
         protected ControlPanel ControlPanel;
 
+        protected virtual bool IsSubScreen => false;
```
peppy, line 36: just use `ParentScreen` being null instead of a second bool

osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
```diff
@@ -42,6 +68,7 @@ private void load()
                     RelativeSizeAxes = Axes.Both,
                     Anchor = Anchor.TopCentre,
                     Origin = Anchor.TopCentre,
+                    Padding = new MarginPadding { Bottom = backButton.Height },
```
peppy, line 71: Don't think this is doing anything

osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
```diff
@@ -70,6 +98,9 @@ private void load()
                 }
             });
 
+            if (IsSubScreen)
```
peppy, line 101: rather than doing this, only construct and add the back button in this conditional.

osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
```diff
@@ -17,19 +20,42 @@
...
+        protected virtual System.Type ParentScreen { get; set; }
+
+        private BackButton backButton;
+
+        private System.Action backAction => () => sceneManager?.SetScreen(ParentScreen);
+
         [BackgroundDependencyLoader]
         private void load()
         {
+            BackButton.Receptor receptor = new BackButton.Receptor();
```
peppy, line 47: inline in the ctor below

osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
```diff
@@ -17,19 +20,42 @@
...
         private FillFlowContainer<TDrawable> flow;
 
+        [Resolved(canBeNull: true)]
+        private TournamentSceneManager sceneManager { get; set; }
+
         protected ControlPanel ControlPanel;
 
+        protected virtual bool IsSubScreen => false;
+
+        protected virtual System.Type ParentScreen { get; set; }
```
peppy, line 38: Rather than having this virtual, how about having the subscreen take a nullable parent in its constructor. so changing `TournamentEditorScreen` take an optional `OsuScreen` and storing that reference. That was we wouldn't have to resort to `System.Type` here (it can be an `OsuScreen`), although for the display call we will need to use `.GetType()` still.

peppy (changes requested):

osu.Game.Tournament/Screens/Editors/TournamentEditorScreen.cs
MiraiSubject: Without padding the rows will overlap with the back button, as demonstrated in [this video](https://streamable.com/0del3r). Adding the padding at the bottom makes sure the back button doesn't overlap with the rows. [Here](https://streamable.com/i56fmu) is another video to compare it against.

peppy: I see. I've reapplied this to a more appropriate location (on the content flow itself).

peppy (approved):

peppy:
@smoogipoo requesting a quick check of the changes i made to `BackButton` / `Receptor` logic.

smoogipoo (approved):
Looks good to me

PR #9048, opened 2020-05-17
Disallow catch-specific judgements in mania

### LINKED ISSUE #9047: idk whats this
**Describe the bug:** i really dont know how to explain this (though only happens to mania maps that i played today cause std and taiko maps maps i played today didnt have them)
beatmap: https://osu.ppy.sh/beatmapsets/318036#mania/795868

played before submitting this issue:
osu_2020-05-17_21-46-51

play hours before (different beatmap):
osu_2020-05-17_21-52-42

old play (weeks before):
osu_2020-05-17_21-53-50

**osu!lazer version:** 2020.515.0

**Logs:**
[logs.zip](https://github.com/ppy/osu/files/4640348/logs.zip)

bdach:
Resolves #9047.

# Summary

Regressed at #8934.

Due to `ManiaHitWindows` not directly overriding `IsHitResultAllowed` but instead [delegating to base](https://github.com/ppy/osu/blob/9003c069bb7d7214d05cfb12ce6f63efe4793f5f/osu.Game/Rulesets/Scoring/HitWindows.cs#L80), new hit result types introduced for the sake of new catch diffcalc started showing up in mania results.

Explicitly only allow the six hit results allowed for mania hitobjects.

# Remarks

No tests, as this shouldn't really regress again, but this exposes the fact that the current API is not safe in regard to extensibility. I thought a little bit about making `IsHitResultAllowed` abstract, but it's not an instant change, as it requires making the base `HitWindows` themselves abstract.

It wouldn't be breaking to any custom ruleset as far as I can see (all of the ones in the custom ruleset listing override `HitWindows` anyway... but not the template rulesets), but it does require direct instantiations of `HitWindows` in tests to be replaced (i.e. with `HitWindows.Empty`) and it also breaks beatmap serialisation due to the JSON deserialiser not being able to instantiate the abstract type, necessitating a new custom converter to resolve.

As such it's probably a separate PR and also one that I'd like to confirm is in line with the direction you want to take.

### DIFF
--- a/osu.Game.Rulesets.Mania/Scoring/ManiaHitWindows.cs
+++ b/osu.Game.Rulesets.Mania/Scoring/ManiaHitWindows.cs
@@ -7,5 +7,20 @@ namespace osu.Game.Rulesets.Mania.Scoring
 {
     public class ManiaHitWindows : HitWindows
     {
+        public override bool IsHitResultAllowed(HitResult result)
+        {
+            switch (result)
+            {
+                case HitResult.Perfect:
+                case HitResult.Great:
+                case HitResult.Good:
+                case HitResult.Ok:
+                case HitResult.Meh:
+                case HitResult.Miss:
+                    return true;
+            }
+
+            return false;
+        }
     }
 }


bdach:
As a side note the obvious solution to the serialisation problem (explicitly encoding the subtype as a string alongside the rest of the object) is very inefficient, as it'll get encoded for each individual instance.

It's a big annoyance - JSON serialisation is fighting very hard against that proposal here due to trying to (de)serialise nested polymorphic members, so most likely any optimised solution would require either rolling our own deserialisation entirely, or making weird adjustments like making base ruleset hitobjects define a `new` member with the proper subtype (haven't tested whether that helps for sure but it might?)

peppy (approved):

PR #9051, opened 2020-05-17
Add custom IPC directory selector to the tournament client

MiraiSubject:
This changes the Refresh button in the `SetupScreen`  of the tournament client to have the functionality of manually specifying a CE Stable location.

The auto detection option has been moved to the directory selector and error handling has also been added for invalid locations or failure for autodetecting. 

This feature could be used in scenarios where users have multiple stable CE installations to separate the different tournaments. 

When the Tournament Client first boots up, it does the ENV_VAR, registry and LocalAppData check as usual, but saves the ipc path now to `tournament/stable.json`.

When a user forces auto detection in the directory selector the stablePath gets set to an empty string so ENV_VAR, registry and LocalAppData check can take place again. 

**Screenshots:**
dotnet_2020-05-17_22-58-51

dotnet_2020-05-17_22-59-05

dotnet_2020-05-17_23-00-07

**GIF of the functionality:**


osu.Game.Tournament/TournamentGameBase.cs
```diff
@@ -33,6 +33,8 @@ public abstract class TournamentGameBase : OsuGameBase
     {
         private const string bracket_filename = "bracket.json";
 
+        private const string stable_config = "tournament/stable.json";
```
bdach, line 36: Shouldn't be defined four times (here, in `FileBasedIPC`, in `StablePathSelectScreen` and in `SetupScreen`).

osu.Game.Tournament/IPC/FileBasedIPC.cs
```diff
@@ -146,54 +156,143 @@ public Storage LocateStableStorage()
...
-                catch
-                {
+                    }
                 }
 
-                try
-                {
-                    using (RegistryKey key = Registry.ClassesRoot.OpenSubKey("osu"))
-                        stableInstallPath = key?.OpenSubKey(@"shell\open\command")?.GetValue(string.Empty).ToString().Split('"')[1].Replace("osu!.exe", "");
+                return stableInstallPath;
```
bdach, line 189: Nit, but I'd probably explicitly return null/empty string here for clarity (right now it looks a little weird, as if it somehow found a path after trying everything and failing).

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,203 @@
...
+                }
+            });
+        }
+
+        private void changePath(Storage storage)
+        {
+            var target = directorySelector.CurrentDirectory.Value.FullName;
+            Logger.Log($"Changing Stable CE location to {target}");
+
+            if (File.Exists(Path.Combine(target, "ipc.txt")))
```
bdach, line 147: I'd invert and deal with the negative part first (`ipc.txt` not found), and early-return if it didn't succeed.

This is also a piece of code (albeit a one-liner) that has duplicated filenames (see `FileBasedIPC.checkExists`). Would like to see at least the filename extracted to a common constant.

osu.Game.Tournament/IPC/FileBasedIPC.cs
```diff
@@ -146,54 +156,143 @@ public Storage LocateStableStorage()
...
+                    return stableInfo.StablePath.Value;
             }
-            finally
+            catch
             {
-                Logger.Log($"Stable path for tourney usage: {stableInstallPath}");
             }
+
+            return null;
+        }
```
bdach, line 247: This method has an actively misleading name (it doesn't read any JSON config or find anything in one, it just reads a previously-found path).

osu.Game.Tournament/IPC/FileBasedIPC.cs
```diff
@@ -146,54 +156,143 @@ public Storage LocateStableStorage()
...
+        private string findFromLocalAppData()
+        {
+            Logger.Log("Trying to find stable in %LOCALAPPDATA%");
+            string stableInstallPath = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.LocalApplicationData), @"osu!");
+
+            if (checkExists(stableInstallPath))
+            {
+                stableInfo.StablePath.Value = stableInstallPath;
+                saveStablePath();
+                return stableInstallPath;
```
bdach, line 258: This scheme of writing to the cached `StableInfo` and then returning the local is very strange. Why not just return `stableInstallPath`, and then assign to `StablePath.Value` and call `saveStablePath` higher up in `findStablePath`?

(Better yet, you could probably assign to the bindable inside of `saveStablePath`, too.)

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,203 @@
...
+                overlay.Push(new IPCErrorDialog("This is an invalid IPC Directory", "Select a directory that contains an osu! stable cutting edge installation and make sure it has an empty ipc.txt file in it."));
+                AddInternal(overlay);
+                Logger.Log("Folder is not an osu! stable CE directory");
+                // Return an error in the picker that the directory does not contain ipc.txt
+            }
+        }
+
+        private void autoDetect()
+        {
+            stableInfo.StablePath.Value = string.Empty; // This forces findStablePath() to look elsewhere.
```
bdach, line 186: `LocateStableStorage` is the only caller of `findStablePath` - so why not just remove the misleading `findFromJsonConfig`, which would render this unnecessary?

osu.Game.Tournament/TournamentGameBase.cs
```diff
@@ -141,6 +145,23 @@ private void load(Storage storage, FrameworkConfigManager frameworkConfig)
...
+                {
+                    stableInfo = JsonConvert.DeserializeObject<StableInfo>(sr.ReadToEnd());
+                }
+            }
+
+            if (stableInfo == null)
+                stableInfo = new StableInfo();
+
+            dependencies.Cache(stableInfo);
+        }
```
bdach, line 163: Not sure I like the fact that the json is being read here, but written in `FileBasedIPC`.

bdach (changes requested):

osu.Game.Tournament/TournamentGameBase.cs
```diff
@@ -141,6 +143,23 @@ private void load(Storage storage, FrameworkConfigManager frameworkConfig)
             });
         }
 
+        private void readStableConfig()
```
bdach, line 146: I still don't quite understand why this has to be here. Can this not be done inside of `FileBasedIPC`?

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,177 @@
...
+        private void load(Storage storage, OsuColour colours)
+        {
+            // begin selection in the parent directory of the current storage location
+            var initialPath = new DirectoryInfo(storage.GetFullPath(string.Empty)).Parent?.FullName;
+
+            if (!string.IsNullOrEmpty(stableInfo.StablePath.Value))
+            {
+                // If the original path info for osu! stable is not empty, set it to the parent directory of that location
+                initialPath = new DirectoryInfo(stableInfo.StablePath.Value).Parent?.FullName;
+            }
```
bdach, line 48: Can be simplified to

```csharp
var initialPath = new DirectoryInfo(storage.GetFullPath(stableInfo.StablePath.Value ?? string.Empty)).Parent?.FullName;
```

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,177 @@
...
+            Logger.Log($"Changing Stable CE location to {target}");
+
+            if (!FileBasedIPC.CheckExists(target))
+            {
+                overlay = new DialogOverlay();
+                overlay.Push(new IPCErrorDialog("This is an invalid IPC Directory", "Select a directory that contains an osu! stable cutting edge installation and make sure it has an empty ipc.txt file in it."));
+                AddInternal(overlay);
+                Logger.Log("Folder is not an osu! stable CE directory");
+                return;
+                // Return an error in the picker that the directory does not contain ipc.txt
```
bdach, line 151: This comment reads as if there is something missing here, which I don't think is the case (there's no error being returned). I'd just remove it.

osu.Game.Tournament.Tests/Screens/TestSceneStablePathSelectScreens.cs
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Game.Tournament.Screens;
+using osu.Framework.Platform;
+
+namespace osu.Game.Tournament.Tests.Screens
+{
+    public class TestSceneStablePathSelectScreens : TournamentTestScene
```
bdach, line 9:
```suggestion
    public class TestSceneStablePathSelectScreen : TournamentTestScene
```
(+ ctor)

osu.Game.Tournament/Models/StableInfo.cs
```diff
@@ -0,0 +1,21 @@
...
+    /// <summary>
+    /// Holds the path to locate the osu! stable cutting-edge installation.
+    /// </summary>
+    [Serializable]
+    public class StableInfo
+    {
+        public Bindable<string> StablePath = new Bindable<string>(string.Empty);
+
+        [JsonIgnore]
+        public const string STABLE_CONFIG = "tournament/stable.json";
```
bdach, line 19: I would prefer this to also reside in `FileBasedIPC` if the read can be moved there.

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,177 @@
...
+        }
+
+        protected virtual void AutoDetect()
+        {
+            var fileBasedIpc = ipc as FileBasedIPC;
+            fileBasedIpc?.LocateStableStorage();
+
+            if (fileBasedIpc?.IPCStorage == null)
+            {
+                // Could not auto detect
```
bdach, line 166: This one is also not really necesssary.

osu.Game.Tournament/TournamentGameBase.cs
MiraiSubject: The reason it's being read here is because it's being it's being put in the dependency container after it was read.
`dependencies.Cache(stableInfo);` 

 Moving the read to `FileBasedIPC` and losing DI on it means that `SetupScreen` can't use this binding any more to refresh the values on screen:

```
stableInfo.StablePath.BindValueChanged(_ =>
{
    Schedule(reload);
});
```

and `StablePathSelectScreen` then needs to access the value of `stableInfo` through `FileBasedIPC`

MiraiSubject:
I sloppily pushed some code with errors in it. VSCode show me these errors till just now. I'm right on it

MiraiSubject:
This PR is broken as of a recent update somewhere in the chain where a path gets read. I discovered this while moving the reads of `stableInfo` to `FileBasedIPC`. As a sanity check I stashed my changes and used my old implementation, but the same exception gets thrown now. 

This exception now gets thrown when calling `StablePathSelectScreen`:

```An unhandled exception of type 'System.ArgumentException' occurred in osu.Framework.dll: '"F:\SSDPrograms\osu! Tourney" traverses outside of "C:\Users\Subject\AppData\Roaming\osu" and is probably malformed'```

This exception initially points to this line in osu-framework: 
https://github.com/ppy/osu-framework/blob/8b00dbb68fb00b98db64644836dec8aadbec4eaa/osu.Framework/Platform/NativeStorage.cs#L65

But this check was already there when I initially created the PR, so I have no idea what else might have changed to trigger this exception. 

In the current state at least it's impossible to set a starting directory outside the original osu! lazer base path.

The specific line in my PR that breaks is: 

`var initialPath = new DirectoryInfo(storage.GetFullPath(stableInfo.StablePath.Value ?? string.Empty)).Parent?.FullName;`

Changing it to 
`var initialPath = new DirectoryInfo(storage.GetFullPath(string.Empty)).Parent?.FullName`

makes it work again, but the starting directory will then always be osu! lazer's base path, instead of the currently selected IPC location. 

I'll be implementing this change with the other requested changes regarding the read of `stableInfo` and await another review.

**Comment by peppy** [comment]
If you need to create a new storage outside of the previous bounds, you may want to look at using this new method available from `GameHost` (you can use DI to resolve this):

https://github.com/ppy/osu-framework/blob/6083659ade1b81deb9e0e44f1ddb6367c8af0c43/osu.Framework/Platform/GameHost.cs#L132

File: osu.Game.Tournament/Components/IPCErrorDialog.cs [hunk 14]
```diff
@@ -0,0 +1,26 @@
...
+        {
+            Icon = FontAwesome.Regular.SadTear;
+            HeaderText = headerText;
+            BodyText = bodyText;
+            Buttons = new PopupDialogButton[]
+            {
+                new PopupDialogOkButton
+                {
+                    Text = @"Alright.",
+                    Action = () => { Expire(); }
```
bdach, line 21: Braces are unnecessary.
```suggestion
                    Action = () => Expire()
```

osu.Game.Tournament/IPC/FileBasedIPC.cs
```diff
@@ -146,54 +155,135 @@ public Storage LocateStableStorage()
                 Logger.Error(e, "Stable installation could not be found; disabling file based IPC");
             }
 
-            return Storage;
+            return IPCStorage;
         }
 
+        public static bool CheckExists(string p) => File.Exists(Path.Combine(p, "ipc.txt"));
```
bdach, line 161: I'd rename to something like `IPCFileExistsInDirectory`. This method should ideally not be `public`, too.

osu.Game.Tournament/IPC/FileBasedIPC.cs
```diff
@@ -146,54 +155,135 @@ public Storage LocateStableStorage()
...
+
         private string findStablePath()
         {
-            static bool checkExists(string p) => File.Exists(Path.Combine(p, "ipc.txt"));
+            if (!string.IsNullOrEmpty(readStableConfig()))
+                return StableInfo.StablePath.Value;
```
bdach, line 166: I find the fact that this returns a path only in this one case, and `null` in all others, really strange. This API is unintuitive to use.

If I got a `null` path from a method called `findStablePath()`, I'd assume that a stable path wasn't found - which is not the case, it means that `FileBasedIPC` found the path on its own, which is really counterintuitive IMO.

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -0,0 +1,173 @@
...
+            {
+                overlay = new DialogOverlay();
+                overlay.Push(new IPCErrorDialog("This is an invalid IPC Directory", "Select a directory that contains an osu! stable cutting edge installation and make sure it has an empty ipc.txt file in it."));
+                AddInternal(overlay);
+                Logger.Log("Folder is not an osu! stable CE directory");
+                return;
+            }
+
+            fileBasedIpc?.SaveStableConfig(target);
+            fileBasedIpc?.LocateStableStorage();
```
bdach, line 152: This is calling three things in sequence on `fileBasedIpc`, which means that this snippet of code is a good candidate for encapsulating in a method. Better yet, if you exposed a method in `FileBasedIPC` like:

```csharp
public bool SetIPCLocation(string path);
```

then you could probably (a) get rid of the strange `string` return type and config lookup case in `getStablePath()`, *and* (b) you could probably reuse *that same* method inside of the stable path lookup code.

bdach (dismissed):
I was going to try to fix this up myself, but reading through I still have objections to how this code is structured, so I'll leave a review instead.

@MiraiSubject If you're tired of applying suggestions, then I can do a refactoring pass over this, but I'd need time to get to it.

osu.Game.Tournament/IPC/FileBasedIPC.cs
MiraiSubject: > If I got a null path from a method called findStablePath(), I'd assume that a stable path wasn't found - which is not the case, it means that FileBasedIPC found the path on its own, which is really counterintuitive IMO.

I was very surprised when I first read this, because it was intended to also return the path for all other methods.  I then noticed that I committed a null I used to test a failure case (namely CI failing). This was very sloppy on my part and I will revert that right away.

MiraiSubject:
I've applied your suggestions! All of them made sense and one of them was just a really sloppy error that I had slip through.

bdach:
I'm still not a fan of how `findStablePath` reads the json config. This makes it so that if you change the ipc location once manually, autodetect stops working as it'll just read its value over and over again.

I tried to make the change myself but it'd involve significant restructuring of `FileBasedIPC` so I'd rather hold off on that until #9210 is merged as that looks like a straight-forward fix and I'd like to avoid conflicts.

MiraiSubject:
Yeah I'm alright with that! I've been facing this problem myself but just haven't figured out a solution for it yet.

I had the idea of skipping the json file read when pressing auto detect so it tries to detect the other methods first, but that would indeed require significant restructuring. I believe I had this working in an older commit but it was very messy.

bdach:
I've made significant changes to this. I think it's tolerable now, but there might be more rough edges. Major changes include:

* factoring out `StableInfo` into a self-contained DI'd structure (which I think makes everything flow better),
* improving error-handling (return value of the old `locateStableStorage()` was being ignored),
* removing some unused arguments & members,
* privatising several members,
* adding xmldoc.

If you disagree with any of the changes let me know and I'll consider reverting.

osu.Game.Tournament/Screens/StablePathSelectScreen.cs
```diff
@@ -42,11 +38,6 @@ private void load(Storage storage, OsuColour colours)
         {
             var initialPath = new DirectoryInfo(storage.GetFullPath(string.Empty)).Parent?.FullName;
 
-            if (!string.IsNullOrEmpty(stableInfo.StablePath))
```
MiraiSubject: This is a little nitpick, but right now the DirectorySelector always defaults to the lazer data directory. In many cases users have their stable tournament install located next to their normal installation in the form of: 

``` 
<ParentDirectory>/osu!
<ParentDirectory>/osu! Tourney
```

Having it default to the parent directory of the default install (if detected) would make it more easier for the user to the get to the correct installation.

bdach: I was going to keep it but forgot about it along the way. The previous behaviour should be back now with my latest commit.

MiraiSubject (commented):
Was able to go through everything now and it looks good to me. API makes much more sense now.

I forget sometimes that it's just normal OOP and that there's nothing weird going on here, so I tend to restrict myself to what I already see.

peppy:
Made some minor changes to this, but feels pretty great otherwise 👍 .

peppy (approved):

PR #9054, opened 2020-05-18
Refactor performance calculators with local functions

huoyaoyuan:
Also add back changes of `float` to `double` which has regressed in conflict of #8934 .

### DIFF
--- a/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs
+++ b/osu.Game.Rulesets.Catch/Difficulty/CatchPerformanceCalculator.cs
@@ -17,14 +17,6 @@ namespace osu.Game.Rulesets.Catch.Difficulty
     {
         protected new CatchDifficultyAttributes Attributes => (CatchDifficultyAttributes)base.Attributes;
 
-        private Mod[] mods;
-
-        private int fruitsHit;
-        private int ticksHit;
-        private int tinyTicksHit;
-        private int tinyTicksMissed;
-        private int misses;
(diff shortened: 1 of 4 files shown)


bdach:
The type changes should have been split - either as a separate PR, or at least at commit boundary - from the refactors IMO. Now it's hard to tell them apart.

smoogipoo:
These are supposed to closely match implementations on osu-performance, so I don't want refactoring to happen just yet.

Also personally not a fan of turning everything into local functions conceptually. It doesn't separate code enough.

huoyaoyuan:
> These are supposed to closely match implementations on osu-performance

That repo uses `float`, but the state of this repo before #6974 is a mixture of `float` and `double`. You should tell me this in that PR.

> Also personally not a fan of turning everything into local functions conceptually. It doesn't separate code enough.

Personally, I think that putting intermediate calculation values into class members is a bad factor. Classes shouldn't be state-full if unnecessary.

smoogipoo:
Changing of floats to doubles is fine. Other changes should not be applied - line refactorings, removal of methods, local functions, etc.

Having class functions doesn't make a class stateful. In-fact these methods are already non-stateful. If you're really concerned about that then these should be `readonly struct`s with appropriate abstractions, but I see that as unnecessary at this point in time - keep it simple for something that isn't yet complete (it'll have to replace osu-performance before I consider anything like that).

huoyaoyuan:
Closing to reopen one for float-to-double unification right now.

Issue #8955, opened 2020-05-06
Every kat sounds like a finisher if the map is mapped with whistles instead of claps


Firmatorenio:
**Describe the bug:** Taiko: on Stable you can map kats by either putting a clap or a whistle on a circle, and in order to make a finisher kat you put either clap+finish or whistle+finish. Lazer doesn't read it well (or at least it doesn't with imported skins), and the ~15% of maps that players mapped with whistles become a hitsound mess.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=F6-0kmPuioU

**osu!lazer version:** 2020.429.0

Firmatorenio:
Still happening as of 2020.515.0

peppy:
Can you give more details? Has anything changed? In my testing it seemed to be resolved so any specific repro would be appreciated.

Firmatorenio:
I'm quite unsure what else to provide, the only thing I note changed regarding what the notes sound like is that the actual kat finishers in the 'unaffected' maps now actually play a whistle sound (before they'd play clap+finish). The exact described in the issue's title problem still happens in both legacy and the default skin: https://www.youtube.com/watch?v=md7D2-X-rzI

Here's a small list of maps that this issue occurs in:
- pretty much all of my own taiko maps (https://osu.ppy.sh/users/4935188)
- searching for maps by "yassu-" generally brings up messed up things:
- - https://osu.ppy.sh/beatmapsets/739116#taiko/1565331
- - https://osu.ppy.sh/beatmapsets/725826#taiko/1532456
- - https://osu.ppy.sh/beatmapsets/539888#taiko/1144904
- - https://osu.ppy.sh/beatmapsets/953713#taiko/1991296
- searching for maps by "[Zeth]" generally brings up at least partially messed up things:
- - https://osu.ppy.sh/beatmapsets/866694#taiko/1943064
- - https://osu.ppy.sh/beatmapsets/868624#taiko/1815528
- - short parts of https://osu.ppy.sh/beatmapsets/925304#taiko/1954240
- - https://osu.ppy.sh/beatmapsets/949737#taiko/1983172
- some other examples
- - https://osu.ppy.sh/beatmapsets/44964#taiko/140655
- - https://osu.ppy.sh/beatmapsets/160466#taiko/394235
- - https://osu.ppy.sh/beatmapsets/55191#taiko/167980
- - https://osu.ppy.sh/beatmapsets/316779#taiko/705830

peppy:
[...]

PR #9057, opened 2020-05-18
Fix taiko rim markers incorrectly playing as whistle samples

peppy:
Following up on https://github.com/ppy/osu/pull/8989, fixes one remaining case which was correctly documented in the PR but not implemented.

Tested against beatmap provided in https://github.com/ppy/osu/issues/8955#issuecomment-629807242.

### DIFF
--- a/osu.Game/Tests/Beatmaps/BeatmapConversionTest.cs
+++ b/osu.Game/Tests/Beatmaps/BeatmapConversionTest.cs
@@ -101,9 +101,6 @@ namespace osu.Game.Tests.Beatmaps
         {
             var beatmap = GetBeatmap(name);
 
-            var rulesetInstance = CreateRuleset();
-            beatmap.BeatmapInfo.Ruleset = beatmap.BeatmapInfo.RulesetID == rulesetInstance.RulesetInfo.ID ? rulesetInstance.RulesetInfo : new RulesetInfo();
-
             var converterResult = new Dictionary<HitObject, IEnumerable<HitObject>>();
 
             var working = new ConversionWorkingBeatmap(beatmap)
@@ -115,7 +112,7 @@ namespace osu.Game.Tests.Beatmaps
                 }
             };
 
-            working.GetPlayableBeatmap(rulesetInstance.RulesetInfo, mods);
+            working.GetPlayableBeatmap(CreateRuleset().RulesetInfo, mods);
 
             return new ConvertResult
             {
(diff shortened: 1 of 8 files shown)


smoogipoo:
So if I understand correctly, only strong rim hits can ever play whistles?

peppy:
Yep, which is what the inline comments and original PR were already stating. Just missed this odd case.

smoogipoo (dismissed):

Firmatorenio:
That sample name is very misleading. I hope it'll be different with lazer skinning.

peppy:
I've added a basic type comparison check, but adding a drawable samples test is a bit of a pain from a conversion.osu source. @smoogipoo you may be able to think of a test I can easily base this off?

- read .osu
- get DrawableRuleset (at the point of `LoadComplete`)
- run asserts against `drawableRuleset.Playfield.AllObjects` (they need to all be in a `LoadComplete` state)

smoogipoo:
Yeah there's not really such a test right now for you to base off of. Using `PlayerTestScene` is probably the easiest way I think, maybe with an override of `TestBeatmap` to read from the conversion .osu file or a separate ctor there.

peppy:
Seems to work well enough.

osu.Game/Tests/Beatmaps/BeatmapConversionTest.cs
```diff
@@ -143,14 +143,19 @@ private ConvertResult read(string name)
...
             using (var resStream = openResource($"{resource_namespace}.{name}.osu"))
             using (var stream = new LineBufferedReader(resStream))
             {
                 var decoder = Decoder.GetDecoder<Beatmap>(stream);
                 ((LegacyBeatmapDecoder)decoder).ApplyOffsets = false;
-                return decoder.Decode(stream);
+                var beatmap = decoder.Decode(stream);
+
+                // not sure but seems to be required.
+                beatmap.BeatmapInfo.Ruleset = CreateRuleset().RulesetInfo;
```
smoogipoo, line 156: We won't want this. This test is typically used for arbitrary ruleset-to-ruleset conversion, this should fail the mania test at the very least (mania's expecting an osu! beatmap for its tests).

I think ideally we want this to be populated with the correct ruleset for the beatmap. But considering going forward and splitting out the rulesets into their own repos, how about:
1. If `BeatmapInfo.RulesetId == CreateRuleset().RulesetInfo.Id`, use the ruleset as you are here.
2. Otherwise, use `null`. This is typically used in places like `TaikoBeatmapConverter` and `ManiaBeatmapConverter` to check if the beatmap is for the current ruleset or not. `null` can be used to indicate that it isn't for the current ruleset.

smoogipoo (changes requested):

osu.Game/Tests/Beatmaps/BeatmapConversionTest.cs
smoogipoo: Hmm looks like this passed since the test sets its own `Ruleset`...

https://github.com/ppy/osu/blob/eee8af28f6756d6a5f52d9a3cb528bd0378451a2/osu.Game/Tests/Beatmaps/BeatmapConversionTest.cs#L104-L105

smoogipoo (approved):

Issue #7218, opened 2019-12-15
Severe preformance issues in Windows 10


NexusXe:
**Describe the bug:** 
Severely degraded performance while playing a beatmap. All of the performance stats look normal, other than the "Draw" one which jumps from 800 to 40ish FPS, and seems to jump between just a few milliseconds to over 100.
**osu!lazer version:** 
2019.1213.0
**Logs:** 
due to multiple reasons logs could not have been uploaded through github, so here's a link to a zip with them instead, sorry: https://cdn.discordapp.com/attachments/559918571925405714/655586744238080020/aaaaa.zip

bdach:
Has this gotten worse with the last release? Does going to the previous one help?

TheUnkn0wnN00b:
I don't follow every release to a tee, but I've had this problem for a few months at least. I do know that at one point lazer ran fine, but for some reason it doesn't anymore. Oh, and reinstalling it doesn't fix the lagging issues (that I didn't exactly point out in the original post but that's what I was getting at)

peppy:
Could you try disabling any overlays or other graphically intensive apps you have running?

TheUnkn0wnN00b:
After killing all overlays and all unnecessary apps running in background, this issue still persists. For good measure I've updated all updatable things on my PC and this lagging is still there. Weirdly, this does not happen in osu!stable.

MitchvanWijngaarden:
Is your desktop background changing?

bdach:
@TheUnkn0wnN00b Have you checked recent releases to see if the problem persists? There have been performance improvements that could help in your case.

TheUnkn0wnN00b:
Sorry for the wait. This problem still persists but to a _much_ lower extent. Task manager shows that Core 0 is maxed and, upon disabling the core for the osu process, this problem disappears. I can only assume that this is a problem with game mode or resource allocation on Windows' part since this does not happen in stable or any other game.

peppy:
I'm wondering if it could be some kind of driver contention, since just having a core saturated will not cause that big of a drop.

Do you use any custom drivers (or have any installed application which may be running at a critical/realtime process priority)?

smoogipoo:
Does this happen from the moment you start the game? Wonder if this is related to https://github.com/ppy/osu/issues/9049 & https://github.com/ppy/osu/issues/8992 ...

bdach:
Closing due to inactivity and not enough information provided to begin to solve this problem. Will reopen if there are any new helpful details.

PR #9058, opened 2020-05-18
Fix memory leak due to undisposed drawables

### LINKED ISSUE #8992: Long freezes after playing for a period of time
**Describe the bug:** 
After about 10-15 minutes of playing the game will start to slowly freeze for a few frames then a few minutes later the game starts to freeze up every second for a few seconds. Game usually starts with about 800mb of memory used then after a few songs and scrolling through the song list memory usage gets higher and higher
System:
i5 4670
8GB of RAM
GTX 745

https://imgur.com/a/Hdyb3qI
https://gyazo.com/01bda4ca0961f9fe71ad41a6cddac3f8 (gif of the game freezing) 

**osu!lazer version:** 
2020.508.2
**Logs:**

### LINKED ISSUE #9049: Possible memory leak
**Describe the bug:** 
After about 2 hours of gameplay, the ram usage went up to over 7 GB(I have 16 GB). This also caused stuttering every few minutes for a few second. After a game restart, I tried replicating the behavior and found out, that every time a new Beatmap is selected in the menu, RAM usage goes up by up to 30 MB.

RAM and statistics at game start
ram after start
statistics after start

RAM and statistics after selecting random maps for 5 minutes
ram after 5 min
statistics after 5 min

Error messages
errors

**osu!lazer version:** 
2020.515.0

**Logs:**

[performance.log](https://github.com/ppy/osu/files/4640881/performance.log)
[performance-audio.log](https://github.com/ppy/osu/files/4640882/performance-audio.log)
[performance-draw.log](https://github.com/ppy/osu/files/4640883/performance-draw.log)
[performance-update.log](https://github.com/ppy/osu/files/4640884/performance-update.log)
[runtime.log](https://github.com/ppy/osu/files/4640885/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4640886/updater.log)
[database.log](https://github.com/ppy/osu/files/4640887/database.log)
[network.log](https://github.com/ppy/osu/files/4640888/network.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9049
Fixes https://github.com/ppy/osu/issues/8992
Possibly (fixes) https://github.com/ppy/osu/issues/7218

I hope this is a temporary fix. We will probably have a better way to do this in the future - whether via realm or otherwise perhaps by implementing some sort of pseudo-`BindableList` that can lead to the removal of the `ItemAdded`/`ItemRemoved` events altogether.

### Explanation

`TopLocalRank` binds to `ScoreManager.ItemAdded` and unbinds correctly in `Dispose()`. But this only works if the `Drawable` itself is disposed, which occurs via one of two processes:
1. Via the draw hierarchy, if the `DrawableCarouselBeatmap` is visible on the screen and hasn't been removed due to `BeatmapCarousel`'s optimisations: https://github.com/ppy/osu/blob/87e501d4ba2422f2c14851902ffd925bd6b8582a/osu.Game/Screens/Select/BeatmapCarousel.cs#L490
2. Via the eager disposal in `BeatmapCarousel.Dispose()`, if the `DrawableCarouselBeatmap` was added to the `Items` list: https://github.com/ppy/osu/blob/87e501d4ba2422f2c14851902ffd925bd6b8582a/osu.Game/Screens/Select/BeatmapCarousel.cs#L574-L575

Unfortunately, this is a case where neither of the above was true, because the `Items` list gets re-created quite a lot (e.g. when selecting a new item): https://github.com/ppy/osu/blob/87e501d4ba2422f2c14851902ffd925bd6b8582a/osu.Game/Screens/Select/BeatmapCarousel.cs#L627 and this list does _not_ contain items that are not present (e.g. alpha=0 due to being offscreen): https://github.com/ppy/osu/blob/87e501d4ba2422f2c14851902ffd925bd6b8582a/osu.Game/Screens/Select/Carousel/CarouselItem.cs#L28
If an item had previously had its `Drawable` representation created and previously loaded a `TopLocalRank` but wasn't added to the `Items` list on the next refresh, then it would exist as this sort of hidden object that would bypass both the above `Disposal()` pathways. Finalisation wouldn't occur because of external references to it. This would actually in-turn keep quite a large reference tree through dependency containers - even up to `SongSelect` itself since it's cached.

## Solution

At first I thought I could implement `IDisposable` on `CarouselItem` (the non-disposable model), however that isn't sufficient since beatmap sets can be removed altogether from `BeatmapCarousel` via the `BeatmapManager` events.

So instead I'm storing all `Drawable`s ever created into a `HashSet` and forcing a disposal on them, which seems like the intention of the original code anyway.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -129,6 +129,9 @@ namespace osu.Game.Screens.Select
 
         protected List<DrawableCarouselItem> Items = new List<DrawableCarouselItem>();
 
+        // Stores all the items ever created for disposal purposes.
+        private readonly HashSet<DrawableCarouselItem> allItems = new HashSet<DrawableCarouselItem>();
+
         private CarouselRoot root;
 
         public BeatmapCarousel()
@@ -571,7 +574,7 @@ namespace osu.Game.Screens.Select
             }
 
             // aggressively dispose "off-screen" items to reduce GC pressure.
-            foreach (var i in Items)
+            foreach (var i in allItems)
                 i.Dispose();
         }
 
@@ -625,6 +628,8 @@ namespace osu.Game.Screens.Select
         private void updateItems()
         {
             Items = root.Drawables.ToList();
+            foreach (var item in Items)
+                allItems.Add(item);
 
             yPositions.Clear();
 


peppy:
Isn't this potentially going to keep more references than intended in some scenarios?

The (probably) worst example I can think of: Import all beatmaps -> Delete all beatmaps -> repeat x 10

That list will be keeping reference to potentially a lot of carousel items, right?

smoogipoo:
It is, yes. Perhaps it's possible for `allItems` to be made into a dictionary, but I don't know if that sort of situation is a point of concern...

peppy:
Maybe just a comment mentioning this is the case for now, so next time we refactor the carousel it won't be forgotten.

smoogipoo:
Superseded by https://github.com/ppy/osu/pull/9062

PR #9062, opened 2020-05-19
Convert dangerous events to IBindables

peppy:
Proof of concept. Potentially alternative to #9058. 

Seems to work and pass tests. Unsure about code quality relative to before (not having dispose unbinds is a plus but..)


smoogipoo (approved):
Looks good!

PR #9064, opened 2020-05-19
Add support for storyboard overlay layer

### LINKED ISSUE #8017: Support for storyboard overlay layer
**Describe the bug:** 
After Starting a Map, The osu!lazer paused to show the notification bar and here is what came out.

**osu!lazer version:** 
image

**Logs:**
*please attach logs here, which are located at:*
- `%AppData%/osu/logs` *(on Windows),*
- `~/.local/share/osu/logs` *(on Linux & macOS).*
[network.log](https://github.com/ppy/osu/files/4258518/network.log)
[performance.log](https://github.com/ppy/osu/files/4258519/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4258520/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4258521/updater.log)
[database.log](https://github.com/ppy/osu/files/4258522/database.log)

bdach:
Resolves #8017.

# Summary

This PR allows for parsing and display of the storyboard overlay layer. It is pretty much identical to other storyboard layers, aside from the fact that it is always displayed front-most (and in front of the playfield, which includes hit objects).

The overlay layer will show regardless of whether the "Show storyboard" option is on at beatmap load, or it is enabled during gameplay - in which case it shows after the asynchronous load is complete - which takes a while in case of more complicated storyboards, but that's not a regression. The opacity of the layer will follow the user dim setting, as in the alpha of the overlay layer is the same as the rest of the storyboard, which as far as I can see is consistent with stable.

Here are a few screenshots of the feature in action:

| [Come And Get It (Razihel Remix) - Krewella (set by Plaudible, storyboard by Hokichi)](https://osu.ppy.sh/beatmapsets/803126) | [SPIDER NET WORK - CleanTears Remix - - Ayakura Mei (set and storyboard by -Tochi, 0% dim)](https://osu.ppy.sh/beatmapsets/941875) | [SPIDER NET WORK - CleanTears Remix - - Ayakura Mei (set and storyboard by -Tochi, ~90% dim)](https://osu.ppy.sh/beatmapsets/941875) |
| :-: | :-: | :-: |
| example1 | example2 | example3-dimmed |

Storyboard parsing tests have been extended to cover the new layer's attributes, but I haven't modified the `.osb` file, which had no objects in the overlay layer.

# Remarks

Hesitantly PRing this, as I'm not too sure of the content, but seeing that it's somewhat done and seems to work reasonably well... Due to the nature of this feature (it's an exception from a nice abstraction) it ends up having quite a few weird corners. I'll attempt to explain the reasoning but any of the kinks I'll yield on immediately if you have an idea that you'd consider better.

* I've opted to make the overlay layer have the minimum depth possible because I think it makes sense - it should always be the topmost layer even if it's not displayed inside of `Player` and proxied above playfield.
* When it comes to the exposition of the overlay layer, I didn't feel like passing a container down from `Player` was right as it goes against the flow of other things in there (`DrawableRuleset.Cursor`, `DrawableRuleset.ResumeOverlay`).
* Exposing just the overlay layer outwardly instead of providing a public by-name layer accessor was a conscious choice as to make the minimal change required and not overexpose things that we might regret exposing later. If you really wanted to, you could grab any layer anyway I guess, but having a public magic-string interface felt *off*, so this attempts to contain the scope of that.
* The double proxy is admittedly weird and done to accommodate the async storyboard load path. From a code perspective I prefer it from having `DimmableStoryboard` expose some sort of event and then `Player` having to subscribe to & handle it - more declarative that way? Reduces clutter in `Player`, for sure.

No visual tests included. I'm not sure the additions are very visually-testable (as in writing actual assertions that will test things isn't trivial, and I'm not sure if it's worth the effort).

### DIFF
--- a/osu.Game/Screens/Play/DimmableStoryboard.cs
+++ b/osu.Game/Screens/Play/DimmableStoryboard.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Allocation;
+using osu.Framework.Graphics.Containers;
 using osu.Game.Graphics.Containers;
 using osu.Game.Storyboards;
 using osu.Game.Storyboards.Drawables;
@@ -13,6 +14,8 @@ namespace osu.Game.Screens.Play
     /// </summary>
     public class DimmableStoryboard : UserDimContainer
     {
+        public Container OverlayLayerContainer;
+
         private readonly Storyboard storyboard;
         private DrawableStoryboard drawableStoryboard;
 
@@ -24,6 +27,8 @@ namespace osu.Game.Screens.Play
         [BackgroundDependencyLoader]
         private void load()
         {
+            Add(OverlayLayerContainer = new Container());
+
             initializeStoryboard(false);
         }
 
@@ -46,9 +51,15 @@ namespace osu.Game.Screens.Play
             drawableStoryboard = storyboard.CreateDrawable();
 
             if (async)
-                LoadComponentAsync(drawableStoryboard, Add);
+                LoadComponentAsync(drawableStoryboard, onStoryboardCreated);
             else
-                Add(drawableStoryboard);
+                onStoryboardCreated(drawableStoryboard);
+        }
+
+        private void onStoryboardCreated(DrawableStoryboard storyboard)
+        {
+            Add(storyboard);
+            OverlayLayerContainer.Add(storyboard.OverlayLayer.CreateProxy());
         }
     }
 }
--- a/osu.Game/Screens/Play/DimmableStoryboard.cs
+++ b/osu.Game/Screens/Play/DimmableStoryboard.cs
@@ -14,7 +14,7 @@ namespace osu.Game.Screens.Play
     /// </summary>
     public class DimmableStoryboard : UserDimContainer
     {
-        public Container OverlayLayerContainer;
+        public Container OverlayLayerContainer { get; private set; }
 
         private readonly Storyboard storyboard;
         private DrawableStoryboard drawableStoryboard;
--- a/osu.Game.Tests/Beatmaps/Formats/LegacyStoryboardDecoderTest.cs
+++ b/osu.Game.Tests/Beatmaps/Formats/LegacyStoryboardDecoderTest.cs
@@ -26,7 +26,7 @@ namespace osu.Game.Tests.Beatmaps.Formats
                 var storyboard = decoder.Decode(stream);
 
                 Assert.IsTrue(storyboard.HasDrawable);
-                Assert.AreEqual(5, storyboard.Layers.Count());
+                Assert.AreEqual(6, storyboard.Layers.Count());
 
                 StoryboardLayer background = storyboard.Layers.FirstOrDefault(l => l.Depth == 3);
                 Assert.IsNotNull(background);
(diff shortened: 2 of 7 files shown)


osu.Game/Screens/Play/DimmableStoryboard.cs
```diff
@@ -46,9 +51,15 @@ private void initializeStoryboard(bool async)
...
 
             if (async)
-                LoadComponentAsync(drawableStoryboard, Add);
+                LoadComponentAsync(drawableStoryboard, onStoryboardCreated);
             else
-                Add(drawableStoryboard);
+                onStoryboardCreated(drawableStoryboard);
+        }
+
+        private void onStoryboardCreated(DrawableStoryboard storyboard)
```
frenzibyte, line 59: small nit:
```csharp
        private void addStoryboard(DrawableStoryboard storyboard)
```

osu.Game/Screens/Play/DimmableStoryboard.cs
```diff
@@ -24,6 +27,8 @@ public DimmableStoryboard(Storyboard storyboard)
         [BackgroundDependencyLoader]
         private void load()
         {
+            Add(OverlayLayerContainer = new Container());
```
frenzibyte, line 30: I think you should auto-size this to the overlay layer?

osu.Game/Screens/Play/DimmableStoryboard.cs
```diff
@@ -13,6 +14,8 @@ namespace osu.Game.Screens.Play
     /// </summary>
     public class DimmableStoryboard : UserDimContainer
     {
+        public Container OverlayLayerContainer;
```
frenzibyte, line 17: nit: wouldn't it be better to expose a `CreateOverlayLayerProxy()` method instead? (with a xmldoc explaining that it has to be proxied at a `Player`-level above the gameplay and below the HUD)

frenzibyte (commented):
Just a couple of nits

osu.Game/Screens/Play/DimmableStoryboard.cs
bdach: This method doesn't only add the storyboard as a child, so I'd say that name would be misleading.

osu.Game/Screens/Play/DimmableStoryboard.cs
bdach: I'm not too convinced. Nothing else in `Player` does anything like this.

osu.Game/Screens/Play/DimmableStoryboard.cs
bdach: Checked this one - setting auto-size on this container has no impact on anything as it has 0x0 size regardless (draw visualiser shows that the only child of this container is a 0x0 `ProxyDrawable`). Sizing of the actually-drawn layer is dealt with by proxying, so I'm inclined to leave this as is.

osu.Game/Screens/Play/DimmableStoryboard.cs
frenzibyte: I mean, it adds the storyboard, and also adds the overlay layer loaded by the storyboard to the needed container, doesn't sound misleading to me?

smoogipoo: To me `onStoryboardCreated` sounds like a callback (which it is), whereas `addStoryboard` sounds like something that should be called by the user.

I think it makes sense as is. In the scope of this file it doesn't make much of a difference.

smoogipoo (approved):
Looks fine to me

PR #5317, opened 2019-07-10
Silence gameplay while seeking

peppy:
Requires performance/alloc regression testing before merge. I'll look at this myself; code is proof-of-concept.

PR #9068, opened 2020-05-20
Use DrawableSample in SkinnableSound class

peppy:
Split out from https://github.com/ppy/osu/pull/5317. I think we should get this change in as it *seems* like a better path forward. At very least it gives more attention to the drawable audio container implementations, which are currently not seeing much use.

### DIFF
--- a/osu.Game/Skinning/SkinnableSound.cs
+++ b/osu.Game/Skinning/SkinnableSound.cs
@@ -4,11 +4,11 @@
 using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
-using osu.Framework.Audio;
-using osu.Framework.Audio.Sample;
 using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.IEnumerableExtensions;
+using osu.Framework.Graphics.Audio;
+using osu.Framework.Graphics.Containers;
 using osu.Game.Audio;
 
 namespace osu.Game.Skinning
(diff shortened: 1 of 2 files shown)


osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -45,33 +52,23 @@ public bool Looping
...
-        public void Play() => channels?.ForEach(c => c.Play());
-
-        public void Stop() => channels?.ForEach(c => c.Stop());
-
-        public void AddAdjustment(AdjustableProperty type, BindableDouble adjustBindable)
+        public void Play() => samplesContainer.ForEach(c =>
         {
-            if (adjustments == null) adjustments = new List<(AdjustableProperty, BindableDouble)>();
+            if (c.AggregateVolume.Value > 0)
+                c.Play();
```
bdach, line 62: So this is a little weird to me - it's a tiny edge-case for sure, but if you did (for whatever reason)

```csharp
skinnableSound.Volume.Value = 0;
skinnableSound.Play();
skinnableSound.Volume.Value = 0.5; // assume this is called before the sample should be done
```

the sound wouldn't play at all. It's just something I'd personally consider unexpected behaviour if that is the case (I'd probably expect for it to be running "in the background", as it does without the guard).

bdach (commented):
Aside from one possibly irrelevant thing looks all good to me. 👍

osu.Game/Skinning/SkinnableSound.cs
frenzibyte: Same case for samples that set looping and adjust the volume after playing:
```csharp
skinnableSound.Looping = true;
skinnableSound.Volume.Value = 0;
skinnableSound.Play();
skinnableSound.TransformBindableTo(Volume, 0.75f, 250); // adjust volume.
```

which sounds like something I would expect from people to do in their rulesets

peppy: rulesets should not be modifying the internal volume of hitsounds. they are set by the mapper.

smoogipoo (approved):

PR #9071, opened 2020-05-20
Seek to start of placed object, not end

peppy:
Discussed IRL as a definite usability improvement. I've also made it only seek if the seek is in a forwards direction, which seems to feel better.

Still need some kind of motion to see the final UX of this, but definitely an improvement IMO.


bdach (approved):
Probably not the target audience to be giving UX feedback on this but I also like how this feels.

smoogipoo (approved):

PR #9078, opened 2020-05-21
Move distance snap grid to osu! HitObjectComposer

peppy:
- [x] Depends on #9070

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/OsuHitObjectComposer.cs
+++ b/osu.Game.Rulesets.Osu/Edit/OsuHitObjectComposer.cs
@@ -4,6 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using osu.Framework.Allocation;
+using osu.Framework.Caching;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Edit;
 using osu.Game.Rulesets.Edit.Tools;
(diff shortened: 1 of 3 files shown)


osu.Game.Rulesets.Osu/Edit/OsuHitObjectComposer.cs
```diff
@@ -32,9 +37,66 @@ protected override DrawableRuleset<OsuHitObject> CreateDrawableRuleset(Ruleset r
...
+        {
+            if (distanceSnapGrid == null)
+                return base.SnapScreenSpacePositionToValidTime(screenSpacePosition);
+
+            (Vector2 pos, double time) = distanceSnapGrid.GetSnappedPosition(distanceSnapGrid.ToLocalSpace(screenSpacePosition));
+
+            return new SnapResult(distanceSnapGrid.ToScreenSpace(pos), time);
+        }
+
+        private void updateDistanceSnapGrid()
```
smoogipoo, line 84: This also needs to be called when the select tool is selected. Without that the placement grid remains visible.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Edit/OsuHitObjectComposer.cs
peppy: Aha. I was only testing using 1 / 2 hotkeys, where it works as expected. Not 100% happy on the invalidation logic for this

peppy: Fix in a better feeling way – now the grid will disappear when exiting the placement area.

smoogipoo (approved):
Seems to work fine for me

PR #9080, opened 2020-05-21
Fix display start time in sequential scroll algorithm 

### LINKED ISSUE #8826: mania notes render at the center of stage on sudden slow scrolling speeds
Describe the bug: title says it, its just that this beatmap: https://osu.ppy.sh/beatmapsets/372942 has a very slow mid section and in lazer, that part generates notes at the center parts of the screen. *just change the title, i really dont know what title it must be*

Screenshots or videos showing encountered issue: (just remove .zip)
[My Movie.mp4.zip](https://github.com/ppy/osu/files/4510595/My.Movie.mp4.zip)

osu!lazer version: 2020.421.0

Logs:
[updater.log](https://github.com/ppy/osu/files/4510585/updater.log)
[database.log](https://github.com/ppy/osu/files/4510587/database.log)
[performance.log](https://github.com/ppy/osu/files/4510588/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4510589/runtime.log)

bdach:
Resolves #8826.

# Summary

As reported in the issue thread, some especially slow-scrolling sections of mania maps would look wrong due to hitobjects appearing in the middle of the scrolling container. This in turn was caused, [as mentioned in the issue](https://github.com/ppy/osu/issues/8826#issuecomment-617520244), by the sequential scroll algorithm making a quite arbitrary choice of the lifetime start time of hitobjects:

https://github.com/ppy/osu/blob/230435b5a365ed8011f80c02b6d89d3a8bb75104/osu.Game/Rulesets/UI/Scrolling/Algorithms/SequentialScrollAlgorithm.cs#L25-L26

Note that the subtractions to `adjustedTime` do not take into account multiplier points at all. If the actual time range at some point in the map is bigger than `timeRange + 1000` due to having a very low multiplier applied, then the lifetime start time would be too late.

The proposed resolution is quite simple, but took me some time to arrive at. Keeping in mind the invariant that objects at both ends of the scrolling container are separated by exactly `scrollLength` and the fact that there is no overlap, we can apply the same technique used to adjust for origin position, and leverage `TimeAt` to do the necessary calculation for us, by passing the negated scroll length to it.

Internally `TimeAt` then accounts for all of the multiplier control points and their effects on the time range, by calculating the absolute position of the hitobject, then subtracting the scroll length and finally reapplying control points one by one.

In other words, given that the start time of the hitobject is its time at the near edge of the scrolling container, we apply a *spatial* offset equal to the length of the scroll to make sure it starts being visible at its far edge, which then `TimeAt` automagically turns into the correct time value.

In addition to the fix, this PR contains unit tests that check the values returned by `GetDisplayStartTime` after this change with a moderately verbose explanation of the math, and also a visual test scene which demonstrates the incorrect behaviour at 24d898c and the correct one at 6f388b7.

Here are some videos for visual comparison purposes:

* the added visual test case ([before](https://streamable.com/rbj648), [after](https://streamable.com/eveo2j)),
* gameplay of map reported as affected ([Ninja Re Bang Bang - Kyary Pamyu Pamyu by ecafree2](https://osu.ppy.sh/beatmapsets/372942#mania/1728348) - [before](https://streamable.com/mkphau), [after](https://streamable.com/uk7h8r)).

# Remarks

The `ScrollingHitObjects` test scene has actually regressed with c06db5a due to the following check:

https://github.com/ppy/osu/blob/0997886df6b093c576252257e779fb116d912dbb/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L260-L261

As the drawable hit objects used for the test specified no hit windows (defaulting to `null`), they would expire immediately upon creation and therefore never show. I've added empty hit windows to them as I see it the easiest and most hassle-free resolution.

The unit tests for `GetDisplayStartTime` do not cover the offset path, as it just boils down to the same thing as the normal path.

I'm reasonably certain that this lifetime is as tight as as can be, but of course more pairs of eyes are definitely needed here as this technically relaxes the lifetime specifications.

### DIFF
--- a/osu.Game.Tests/ScrollAlgorithms/SequentialScrollTest.cs
+++ b/osu.Game.Tests/ScrollAlgorithms/SequentialScrollTest.cs
@@ -29,8 +29,22 @@ namespace osu.Game.Tests.ScrollAlgorithms
         [Test]
         public void TestDisplayStartTime()
         {
-            // Sequential scroll algorithm approximates the start time
-            // This should be fixed in the future
+            // easy cases - time range adjusted for velocity fits within control point duration
+            Assert.AreEqual(2500, algorithm.GetDisplayStartTime(5000, 0, 2500, 1)); // 5000 - (2500 / 1)
+            Assert.AreEqual(13750, algorithm.GetDisplayStartTime(15000, 0, 2500, 1)); // 15000 - (2500 / 2)
+            Assert.AreEqual(20000, algorithm.GetDisplayStartTime(25000, 0, 2500, 1)); // 25000 - (2500 / 0.5)
+
+            // hard case - time range adjusted for velocity exceeds control point duration
+
+            // 1st multiplier point takes 10000 / 2500 = 4 scroll lengths
+            // 2nd multiplier point takes 10000 / (2500 / 2) = 8 scroll lengths
+            // 3rd multiplier point takes 2500 / (2500 * 2) = 0.5 scroll lengths up to hitobject start
+
+            // absolute position of the hitobject = 1000 * (4 + 8 + 0.5) = 12500
+            // minus one scroll length allowance = 12500 - 1000 = 11500 = 11.5 [scroll lengths]
+            // therefore the start time lies within the second multiplier point (because 11.5 < 4 + 8)
+            // its exact time position is = 10000 + 7.5 * (2500 / 2) = 19375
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):
I think this makes sense :+1:

PR #9081, opened 2020-05-22
Replace Color4 with Colour4

swoolcock:
Required as part of osuTK removal.

- [ ] Depends on https://github.com/ppy/osu-framework/pull/3573


swoolcock:
Closing for now, will likely do this in small chunks when we are ready.

PR #9082, opened 2020-05-22
Avoid using CurrentDirectory to decide storage location in OsuGameTestScene

peppy:


osu.Game/Tests/Visual/OsuTestScene.cs
```diff
@@ -118,7 +121,8 @@ public virtual void RecycleLocalStorage()
                 }
             }
 
-            localStorage = new Lazy<Storage>(() => new NativeStorage($"{GetType().Name}-{Guid.NewGuid()}"));
+            localStorage = new Lazy<Storage>(() => new NativeStorage(
+                Path.Combine(Path.GetDirectoryName(host.FullPath), $"{GetType().Name}-{Guid.NewGuid()}")));
```
frenzibyte, line 125: although it's very unlikely to happen, but this will fail if used before the `GameHost` is resolved, what do you think about using `AppDomain.CurrentDomain.BaseDirectory` instead? (which should work exactly the same way)

peppy: please provide a case where that happens.

peppy:
There's multiple cases of similar access framework-side (see `TemporaryNativeStorage` usage) and this fix shouldn't be done locally. Hopefully we can resolve the underlying issue as it is something we don'e expect to have happen during runtime (although there's argument that we shouldn't be relying on `Environment.CurrentDirectory` anywhere post-startup, too).

Closing this for now.

PR #9083, opened 2020-05-22
Cache EditorClock as its own type

peppy:
For simplicity and to allow further customisations to the clock implementation.

- [x] Depends on https://github.com/ppy/osu/pull/9078
- [x] Depends on https://github.com/ppy/osu/pull/9070
- [x] Depends on https://github.com/ppy/osu/pull/9021


smoogipoo:
Wanna resolve conflicts on this one?

smoogipoo (approved):

PR #9084, opened 2020-05-22
Transform seeks in editor to allow for better visual context

peppy:
- [x] Depends on #9083 + chain


smoogipoo (approved):
I gotta say, it's a little bit scary that we can set a `Drawable`'s `Clock` to a component that is added elsewhere in the hierarchy. Not sure how I feel about that yet, instead of going down a pathway similar to `GameplayClockContainer` / `GameplayClock`.

Let's see how it goes...

PR #9088, opened 2020-05-22
Fix defaults not being applied correctly to blueprints after StartTime is changed

### LINKED ISSUE #9085: Slider velocity applied during placement may be incorrect
Reproduced by attempting to place a slider in https://osu.ppy.sh/beatmapsets/1135050#osu/2370560.

peppy:
- Closes #9085

Sliders now change velocity when they are moved to a new control point, for instance:

image


bdach (approved):
Looks good to me 👍

smoogipoo (approved):

PR #9089, opened 2020-05-22
Fix beatmap-online-lookup-queue potentially throwing on OsuGameTestScene

frenzibyte:
This is happening because of test storages being recycled per-test-case, so requests have to be immediately aborted before any further steps are performed resulting in exceptions thrown (`DirectoryNotFoundException` from `File.Delete()`, etc...).

Disposing `FileWebRequest`s while they're downloading seems to be the best way to fix this, as it deletes the compressed cache file (internally from `FileWebRequest.Complete(abortWebException)`) and not invoke both `Failed` and `Finished`.


peppy:
A better fix would be to stop it from actually performing the download in said tests (as is already being done when run under nunit). But having this disposal logic also makes sense.

peppy (approved):

PR #9092, opened 2020-05-22
Add custom resolution set option to tournament client

omkelderman:
The "set to 1080p" button is really nice, but if you have a 1080p monitor, a 1080p tourney size obviously wont fit. So I thought I can add another button with "set to 720p", but then someone might ask, hey, I wanna have a button for 900p. So then I thought, why not have an inputbox where you can specify it yourself.

So thats what this is: it replaces the "set to 1080p" button with a number input field and and a "set size" button, and everything else will be calculated.

I dont have a ton of experience with the framework yet, so I might be using certain elements in a way thats not intended, or maybe there is a better way of doing certain things, in that case lemmy know and I'll be happy to adjust.

Also if this PR does not align with potential future plans of the tourney client im not aware of, feel free to ignore it. It started as a fun idea of like "hey I wonder if I can do this" and in the end I was like, eh, might as well PR it, maybe it is useful to someone.


omkelderman:
I have no idea whats going on with that failing test on AppVeyor. If I test the `osu.Game.Tests` project (which is where the failing test seems to be in) in Visual Studio I get

osu.Game.Tests
  Tests in group: 1118
   Total Duration: 3,4 min

Outcomes
   1108 Passed
   10 Skipped

bdach:
While it's something that should be investigated (and probably not easily reproducible as it's probably a concurrency issue) it's not something that would be caused by changes here.

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -108,18 +108,22 @@ private void reload()
                     Items = rulesets.AvailableRulesets,
                     Current = LadderInfo.Ruleset,
                 },
-                resolution = new ActionableInfo
+                resolution = new ActionableInfoWithNumberBox
                 {
                     Label = "Stream area resolution",
-                    ButtonText = "Set to 1080p",
-                    Action = () =>
+                    ButtonText = "Set size",
```
bdach, line 114: Should probably say

```suggestion
                    ButtonText = "Set height",
```

because currently it's unclear which dimension will be set.

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -196,5 +200,56 @@ public bool Failing
...
+        {
+            public new Action<int> Action;
+
+            private OsuNumberBox numberBox;
+
+            public int NumberValue
+            {
+                get
+                {
+                    int.TryParse(numberBox.Text, out var val);
```
bdach, line 214: Could just use `int.Parse` directly. `OsuNumberBox` should ensure there are no invalid chars here.

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -196,5 +200,56 @@ public bool Failing
                 }
             };
         }
+
+        private class ActionableInfoWithNumberBox : ActionableInfo
```
bdach, line 204: This is not named too well. Seeing that this drawable is used in only one context for now I'd probably rename to something like `ResolutionSelector`.

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -196,5 +200,56 @@ public bool Failing
...
+                    },
+                    numberBox = new OsuNumberBox
+                    {
+                        Anchor = Anchor.CentreRight,
+                        Origin = Anchor.CentreRight,
+                        Width = 100,
+                        Margin = new MarginPadding
+                        {
+                            Right = 110
+                        }
```
bdach, line 239: This is not a reliable way to do this sort of layouting. You'd want to nest this and the button in a fill flow:

```
Container
    ValueText (TournamentSpriteText)
    FillFlowContainer
        numberBox (OsuNumberBox)
        Button (TriangleButton)
```

Better yet, you could:

* add the fill flow to `ActionableInfo` and make `Button` its only child,
* expose the new fill flow as protected,
* and then here you wouldn't have to duplicate the body of `CreateComponent()`, but you could just call base and then add the number box to the exposed fill flow by calling
  ```
  Insert(0, numberBox = new OsuNumberBox { ... }
  ```

bdach (dismissed):

bdach:
Also as a side note not being careful about the values that you input into the textbox makes it really easy to trigger ppy/osu-framework#3370 on Windows.

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: not if string is empty or if string contains a number too big to fit in an int, altho in that case there should probably some error displayed to the user, but no idea how to do that.

iirc I do check somewhere else for empty string and then just do nothing lol

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: I'll be the first to admit I'm terrible at naming things, but what I usually end up doing is try naming a thing after what it is, or what it contains, not what its being used for, since that might change in the future, but I dont really have a strong opinion on this either way so im fine with whatever

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: I knew I was doing some hacky shit when coding this haha, thanks for the clear explanation, will implement it in the "better yet" way! 👍

osu.Game.Tournament/Screens/SetupScreen.cs
bdach: I mean something like `LabelledNumberSetting` could maybe work a little better if we want a generic name here.

osu.Game.Tournament/Screens/SetupScreen.cs
bdach: Fair points. I'd probably leave this as it is, the tournament client isn't supposed to be the most user-friendly and polished thing at this point.

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: I already started coding again and ended up naming it `ResolutionSelector` and after some refactoring actually ended up having some resolution specific logic in it cuz that worked nicer so uh ye hahaha 👀

omkelderman:
refactored the thing, its now using a `FillFlowContainer` as suggested.

However after the `Insert(0, ...)` I noticed I had to manually also do a `SetLayoutPosition(Button, 1)`. I dont know if this intended behaviour or if this is supposed to be handled automatically (thats what would make sense to me). If I can nuke that line, then the button can again be private inside `ActionableInfo`.

As for parsing the number out of the inputbox: I now check first on empty string, and then do a TryParse and if that fails assume its a big number and clamp the value to max value.
Not sure what to do when the inputbox was empty tho, just do nothing? inform user? something else? I left that line blank for now. So either I should just remove that else or implement something there.

About that ppy/osu-framework#3370 bug, not sure how to take that into account tbh. Right now I am clamping the value between 480 (it seems that the window doesnt want to be set smaller anyway) and 2160 (4k) which is just some arbitrary max value I made up since I figured it was probably smart to put some sort of max on it. No idea if there is a better solution (eg detecting users monitor size and limiting based on that, but not sure how complicated that is)

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -186,15 +189,70 @@ public bool Failing
...
+            private OsuNumberBox numberBox;
+
+            protected override Drawable CreateComponent()
+            {
+                var drawable = base.CreateComponent();
+                FlowContainer.Insert(0, numberBox = new OsuNumberBox
+                {
+                    Width = 100
+                });
+                FlowContainer.SetLayoutPosition(Button, 1);
```
bdach, line 226: This is my oversight - the `0` doesn't work as the button already has layout position 0. `-1` will work (it's not a conventional list).

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -186,15 +189,70 @@ public bool Failing
...
+
+                        // in case number got clamped, reset number in numberBox
+                        numberBox.Text = number.ToString();
+
+                        Action?.Invoke(number);
+                    }
+                    else
+                    {
+                        // TODO: input box was empty, give user feedback? do nothing?
+                    }
```
bdach, line 252: You could disable the button if you really wanted in this branch (remembering to re-enable it once input is non-empty again). But that's about as far as I'd go.

bdach (commented):
I was mostly mentioning the framework issue as an off-hand thing to keep in mind. Not sure it's something that should be addressed here really.

Squashing commits is not necessary. We try to limit force pushes to an absolute minimum.

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: oooh right so its basically just arbitrary numbers and in the end the elements are ordered by whatever values they have?

osu.Game.Tournament/Screens/SetupScreen.cs
omkelderman: well this code runs when you click the button, so that wouldnt make a ton of sense.

guess I'll just keep it at just doing nothing at all. I guess should be obvious enough that if you want that button to do something you should enter a value. Altho while im typing this I get an idea: give the inputbox focus as to hint the user they should type something there. Is there a way to set focus to an inputbox in code? I tried to look for something but couldnt find anything 🤔

bdach: It would visually indicate that the button cannot be clicked due to lack of input in the text box. But again, this doesn't really matter this much at all. As far as I'm concerned it can just do nothing, I'd just remove the todo.

osu.Game.Tournament/Screens/SetupScreen.cs
bdach: Yes. `Insert` internally calls `SetLayoutPosition`.

bdach:
I've made some menial changes here, but other than that I think this is in a mergeable state.

peppy (approved):

PR #9095, opened 2020-05-23
Add universal back button to the toolbar

AstralPhnx:
Since the current system for going back one step is a bit obtuse (clicking outside the window isn't the first thing I'd think of for going back), I've implimented a universal back button onto the top toolbar. This takes you back to whatever previous sub screen you were on (I.E if you're browsing osu direct and click on a beatmap to read it, you can click the back button to return to where you were in osu direct).

I've attached some screenshots of the addition below:

image

image

(As a side note, the excessive commits were me struggling with github)


bdach:
Thanks for the contribution, but admittedly I'm not convinced this is the direction we'd like to be going UX-wise.

* I find it confusing that the added back button is always visible regardless of whether there is a screen to go back to.
* The way it is currently implemented, it does not only affect screens (like multi) and overlays, it also closes sidebars or the music controller, which I feel shouldn't be the case.
* Finally, the <kbd>Esc</kbd> key (back button on Android) already serves the exact same purpose.

EZ2ACTux:
> 
> 
> Thanks for the contribution, but admittedly I'm not convinced this is the direction we'd like to be going UX-wise.
> 
>     * I find it confusing that the added back button is always visible regardless of whether there is a screen to go back to.
> 
>     * The way it is currently implemented, it does not only affect screens (like multi) and overlays, it also closes sidebars or the music controller, which I feel shouldn't be the case.
> 
>     * Finally, the Esc key (back button on Android) already serves the exact same purpose.
> 
> 
> Also, as an aside, I'd appreciate it if you had a look at how git works if you decide to continue contributing, and read up on the PR workflow - it is clear that you are not entirely familiar with how it works (judging by the unnecessary reverts in history & submitting the PR off `master` and not a topic branch).

Fair enough. I just felt that a back button somewhere might be a good idea since it isn't entirely clear at first glance how to actually back out of certain screens.

And yeah, git has never quite been my strong point so I have a bit to work on. I'll close this off for now but I feel it might be worth opening an issue looking into maybe making a more obvious way to back out of nested screens, at least on PC since I know a few people including myself got a bit confused.

PR #9101, opened 2020-05-24
Add the ability to export skins

peppy:
Just filling up my [karma quota](https://twitter.com/RealHorizon/status/1264319044875890690). This is the third or forth time this has been requested from troubled users, so feels like it's worth adding (considering the low implementation overhead).

### DIFF
--- a/osu.Game/Database/ArchiveModelManager.cs
+++ b/osu.Game/Database/ArchiveModelManager.cs
@@ -375,7 +375,7 @@ namespace osu.Game.Database
         }, cancellationToken, TaskCreationOptions.HideScheduler, import_scheduler).Unwrap();
 
         /// <summary>
-        /// Exports an item to an legacy (.zip based) package.
+        /// Exports an item to a legacy (.zip based) package.
         /// </summary>
         /// <param name="item">The item to export.</param>
         public void Export(TModel item)
@@ -383,7 +383,7 @@ namespace osu.Game.Database
             var retrievedItem = ModelStore.ConsumableItems.FirstOrDefault(s => s.ID == item.ID);
 
             if (retrievedItem == null)
-                throw new ArgumentException("Specified model count not be found", nameof(item));
+                throw new ArgumentException("Specified model could not be found", nameof(item));
 
             using (var archive = ZipArchive.Create())
             {
--- a/osu.Game/Database/ArchiveModelManager.cs
+++ b/osu.Game/Database/ArchiveModelManager.cs
@@ -22,6 +22,7 @@ using osu.Game.IO.Archives;
 using osu.Game.IPC;
 using osu.Game.Overlays.Notifications;
 using osu.Game.Utils;
+using SharpCompress.Archives.Zip;
 using SharpCompress.Common;
 using FileInfo = osu.Game.IO.FileInfo;
 
@@ -82,6 +83,8 @@ namespace osu.Game.Database
         // ReSharper disable once NotAccessedField.Local (we should keep a reference to this so it is not finalised)
         private ArchiveImportIPCChannel ipc;
 
+        private readonly Storage exportStorage;
+
         protected ArchiveModelManager(Storage storage, IDatabaseContextFactory contextFactory, MutableDatabaseBackedStoreWithFileIncludes<TModel, TFileModel> modelStore, IIpcHost importHost = null)
         {
             ContextFactory = contextFactory;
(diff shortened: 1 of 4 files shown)


osu.Game/Overlays/Settings/Sections/SkinSection.cs
```diff
@@ -81,6 +102,9 @@ private void load(OsuConfigManager config)
             skinDropdown.Bindable = dropdownBindable;
             skinDropdown.Items = skins.GetAllUsableSkins().ToArray();
 
+            currentSkin = skins.CurrentSkin.GetBoundCopy();
+            currentSkin.BindValueChanged(skin => exportButton.Enabled.Value = skin.NewValue.SkinInfo.ID > 0);
```
frenzibyte, line 106: Should check immediately:
```suggestion
            currentSkin.BindValueChanged(skin => exportButton.Enabled.Value = skin.NewValue.SkinInfo.ID > 0, true);
```

osu.Game/Overlays/Settings/Sections/SkinSection.cs
```diff
@@ -81,6 +102,9 @@ private void load(OsuConfigManager config)
             skinDropdown.Bindable = dropdownBindable;
             skinDropdown.Items = skins.GetAllUsableSkins().ToArray();
 
+            currentSkin = skins.CurrentSkin.GetBoundCopy();
```
frenzibyte, line 105: Is there a reason this is not using the `dropdownBindable` instead? (saves for another bindable copy)

osu.Game/Overlays/Settings/Sections/SkinSection.cs
```diff
@@ -34,13 +35,33 @@ public class SkinSection : SettingsSection
...
+                    Action = () =>
+                    {
+                        try
+                        {
+                            skins.Export(skins.CurrentSkin.Value.SkinInfo);
+                        }
+                        catch (Exception)
+                        {
+                            Logger.Log("Could not export current skin", level: LogLevel.Error);
+                        }
```
frenzibyte, line 62: Should log the exception we have along the way?:
```suggestion
                        catch (Exception e)
                        {
                            Logger.Error(e, "Could not export selected skin!");
                        }
```

frenzibyte (changes requested):
Looks good otherwise, although I've noticed an incorrect behaviour with sub-`WrappedStorage`s (from `GetStorageForDirectory()`), when opening them in native explorer they lead to the top-level storage instead of their path, out of scope from this PR but I think it's worth mentioning it.

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -369,6 +374,29 @@ void rollback()
...
+        /// <summary>
+        /// Exports an item to an legacy (.zip based) package.
+        /// </summary>
+        /// <param name="item">The item to export.</param>
+        public void Export(TModel item)
+        {
+            var retrievedItem = ModelStore.ConsumableItems.FirstOrDefault(s => s.ID == item.ID);
+
+            if (retrievedItem == null)
+                throw new ArgumentException("Specified model count not be found", nameof(item));
```
bdach, line 386:
```suggestion
                throw new ArgumentException("Specified model could not be found", nameof(item));
```

osu.Game/Overlays/Settings/Sections/SkinSection.cs
peppy: Logger.Error is not meant for errors triggered by a user. It will report it as an error to us, which we do not want.

osu.Game/Overlays/Settings/Sections/SkinSection.cs
peppy: I prefer having it separated.

frenzibyte: Looks better now after moving implementation to own class.

osu.Game/Overlays/Settings/Sections/SkinSection.cs
frenzibyte: Though you're logging it on an error-level (exactly like `Logger.Error(...)`), should change that to `Verbose` instead.

bdach: No, `Verbose` is incorrect. It's an error, but one that shouldn't be logged to sentry.

If anything I'd be for logging the stack trace here for diagnostic purposes.

peppy:
> Looks good otherwise, although I've noticed an incorrect behaviour with sub-`WrappedStorage`s (from `GetStorageForDirectory()`), when opening them in native explorer they lead to the top-level storage instead of their path, out of scope from this PR but I think it's worth mentioning it.

Yeah, also saw this. Was thinking it was a platform level issue but I guess it's actually a regression. Will push a fix.

smoogipoo (approved):

PR #9108, opened 2020-05-25
Simplify scrolling ruleset editor implementation by adding base position support

peppy:
While continuing to work on the editor implementation workflow (making an osu!taiko editor) it became very apparent that a lot of implementation is shared with osu!mania, and will be share with osu!catch and many more custom scrolling rulesets in the future.

This PR aims to move as much of that logic to a base implementation that can be assumed as a sane default. Already tested/used against the WIP [taiko editor](https://github.com/ppy/osu/compare/master...peppy:editor-remove-size-propagation?expand=1).

In addition, this manages to remove the `ManiaSnapResult` class, which was mentioned as being a bit of a wierd case in the previously merged PR.


bdach (commented):
Yeah, I'm liking this direction 👍 Aside from the typical annoying nits I got nothing to this code.

I can see why you said central anchoring in mania would be simpler. The adjustments for note height do stick out a little bit, but I wouldn't call it bad in any way really.

smoogipoo (approved):

PR #9109, opened 2020-05-25
New osu standard sr and pp system

HeBuwei:
[Here](https://docs.google.com/document/d/1rymnSJ35avbWLwFe45twHnB5zM1BIQcj_mV6D4RE-H8/edit?usp=sharing) is a description of the new system. And you can view the results of the new system on [this website](https://newpp.stanr.info/).
The new system has been developed with feedback from the community.


osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyAttributes.cs
```diff
@@ -7,8 +7,26 @@ namespace osu.Game.Rulesets.Osu.Difficulty
...
+
+        public double AimSR;
+        public double AimDiff;
+        public double AimHiddenFactor;
+        public double[] ComboTPs;
+        public double[] MissTPs;
+        public double[] MissCounts;
+        public double CheeseNoteCount;
+        public double[] CheeseLevels;
+        public double[] CheeseFactors;
```
smoogipoo, line 27: This is a lot of attributes. Are all of these used in the PP processor/can any of them be removed?

How much data is expected in each one of these?

HeBuwei: All of them are used in the pp processor. CheeseLevels is the same for every map so it can be removed, but others are necessary.
ComboTPs has 50 elements, MissTPs has 20, MissCounts has 20, CheeseLevels has 11, CheeseFactors has 11.

smoogipoo:
Just so everyone knows ahead of time, this is likely still several months away from testing on our side at this point. The infrastructure isn't going to be able to handle the sheer amount of new attribute data.

A likely path forward here is to implement on-demand SR calculation in osu-server first, which is something we've discussed conceptually in the past but haven't designed/planned yet.

osu.Game.Rulesets.Osu/Difficulty/MathUtil/HitProbabilities.cs
```diff
@@ -0,0 +1,165 @@
...
+
+            return count;
+        }
+
+        public bool IsEmpty(int sectionCount)
+        {
+            bool isEmpty = false;
+            for (int i = 0; i <= sections.Count - sectionCount; i++)
+            {
+                isEmpty = isEmpty || Count(i, sectionCount) == 0;
```
Mechan0, line 42: This should just return false as soon as a non-empty section is found rather than iterating through no matter what

osu.Game.Rulesets.Osu/Difficulty/MathUtil/FittsLaw.cs
```diff
@@ -0,0 +1,76 @@
...
+            {
+                return 1 / Power2(-x);
+            }
+            if (x>60)
+            {
+                return double.PositiveInfinity;
+            }
+
+            int floor = (int)x;
+            double frac = x - floor;
```
Mechan0, line 61: This is a bit difficult to read. In theory a loop could make this much clearer while making it easier to add/remove coefficients without modifying this code:
```
int floor = (int)x;
double coeffSum= coeffs[0];
double frac = x - floor;
for(int i= 1; i < coeffs.Count(); i++) {
        coeffSum+= coeffs[i] * frac;
        frac *= frac;
}
return (1L << floor) * coeffSum;
```
or something along those lines

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
```diff
@@ -0,0 +1,571 @@
+﻿using System;
```
Mechan0, line 1: It would be nice if `ExtractMovement` could be subdivided into more functions because it spans a few hundred lines in it's current state

osu.Game.Rulesets.Osu/Difficulty/Skills/Tap.cs
```diff
@@ -0,0 +1,161 @@
...
+
+        /// <summary>
+        /// Calculates the strain values at each note and the maximum strain values
+        /// </summary>
+        private static (List<Vector<double>>, double) calculateTapStrain(List<OsuHitObject> hitObjects,
+                                                                                 double mashLevel,
+                                                                                 double clockRate)
+        {
+            var strainHistory = new List<Vector<double>> { decay_coeffs * 0, decay_coeffs * 0 };
+            var currStrain = decay_coeffs * 1;
```
Mechan0, line 47: isn't this just effectively `decay_coeffs`?

osu.Game.Rulesets.Osu/Difficulty/Skills/Tap.cs
```diff
@@ -0,0 +1,161 @@
...
+                                  Math.Pow(calculateMashNerfFactor(distance, mashLevel), 3) *
+                                  Math.Pow(1 + spacedBuff, 3);
+
+                    prevPrevTime = prevTime;
+                    prevTime = currTime;
+                }
+            }
+
+            // compute difficulty by aggregating strainHistory
+            var strainResult = decay_coeffs * 0;
```
Mechan0, line 82: isn't this just 0?

osu.Game.Rulesets.Osu/Difficulty/Skills/FingerControl.cs
```diff
@@ -0,0 +1,74 @@
...
+                if (Math.Abs(strainTime - prevStrainTime) > 0.004)
+                    repeatStrainCount = 1;
+                else
+                    repeatStrainCount++;
+
+                if (hitObjects[i] is Slider)
+                    strain /= 2.0;
+
+                if (repeatStrainCount % 2 == 0)
+                    strain = 0;
```
Mechan0, line 48: Curious why current strain is effectively only being added here for every other note. Seems like an odd way to seemingly curtail the growth than just applying a 0.5 modifier or something.

osu.Game.Rulesets.Osu/Difficulty/Skills/Tap.cs
HeBuwei: it should be `decay_coeffs * 0` because it is the initial strain value

osu.Game.Rulesets.Osu/Difficulty/Skills/Tap.cs
HeBuwei: it's a new vector of the same shape as `decay_coeffs`

tybug: might communicate the intent better by using `Vector<double>.Build(decay_coeffs.Count)`, or extract it out to a method like `NewOfShape(Vector v)` if it's something you do often. Unless there are performance impacts with building?

Mechan0: That is really confusing to say the least, vector multiplication to create an identically sized empty vector does not really express intent well as @tybug said

osu.Game/Scoring/ScoreInfo.cs
```diff
@@ -31,6 +31,10 @@ public class ScoreInfo : IHasFiles<ScoreFileInfo>, IHasPrimaryKey, ISoftDelete,
         [JsonProperty("accuracy")]
         [Column(TypeName = "DECIMAL(1,4)")]
         public double Accuracy { get; set; }
+        
+        [JsonProperty("modified_accuracy")]
+        [Column(TypeName = "DECIMAL(1,4)")]
+        public double ModifiedAccuracy { get; set; }
```
peppy, line 37: what is this here for? it's not used and there's also no migrtion for it.

osu.Game.Rulesets.Osu/Difficulty/Skills/Aim.cs
```diff
@@ -14,46 +24,248 @@ namespace osu.Game.Rulesets.Osu.Difficulty.Skills
     /// </summary>
     public class Aim : Skill
     {
-        private const double angle_bonus_begin = Math.PI / 3;
-        private const double timing_threshold = 107;
+        private const double probability_threshold = 0.02;
```
peppy, line 27: can you go through and add xmldoc for all constants that aren't immediately obvious?

peppy (changes requested):

HeBuwei:
> Just so everyone knows ahead of time, this is likely still several months away from testing on our side at this point. The infrastructure isn't going to be able to handle the sheer amount of new attribute data.
> 
> A likely path forward here is to implement on-demand SR calculation in osu-server first, which is something we've discussed conceptually in the past but haven't designed/planned yet.

@smoogipoo would it be better if we cut down the total number of elements from 100+ to 40+? If this is a very important issue then we can trade a bit of accuracy for a lower number of elements.

smoogipoo:
Even at 40 (minimum) elements that's 8x increase in data. I'll defer to @peppy on that one but I believe the consensus will be no on that.

peppy:
Will these be stored as a json string against one attribute?

smoogipoo:
That's what I was thinking, would there be any easier way to store arrays? You could probably make tables for each one of them since they're fixed-length, but I don't believe that's going to be any easier/better and requires significant changes to both osu-server and osu-performance anyway.

peppy:
The main issue is that the attribs table is currently numeric. We can make a new table or convert the data type to support json storage. Should be fine, will figure something out if that's what is required.

osu.Game/Scoring/ScoreInfo.cs
HeBuwei: I think vinxis added it for his bot. I'll remove it.

osu.Game.Rulesets.Osu/Difficulty/Skills/Aim.cs
HeBuwei: ok

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
HeBuwei: It's indeed a complex function. I'm not sure if dividing it would make it more readable though.

osu.Game.Rulesets.Osu/Difficulty/Skills/Tap.cs
HeBuwei: ok i'll change that

osu.Game/Scoring/ScoreInfo.cs
HeBuwei: removed

osu.Game.Rulesets.Osu/Difficulty/Skills/Aim.cs
HeBuwei: i've added xmldoc for constants and functions

peppy:
@HeBuwei please fix the remaining issues as pointed out by CI.

osu.Game.Rulesets.Osu/Difficulty/MathUtil/PoissonBinomial.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using MathNet.Numerics.Distributions;
+
+namespace osu.Game.Rulesets.Osu.Difficulty.MathUtil
+{
+    /// <summary>
+    /// approximate poisson binomial CDF defined by miss probabilities
+    /// see "Refined Normal Approximation (RNA)" from
+    /// https://www.researchgate.net/publication/257017356_On_computing_the_distribution_function_for_the_Poisson_binomial_distribution
```
bdach, line 14: I have reservations to sites like ResearchGate due to the possibility of them gating access to publications at some point in the future. If there was a reference available in a clearly open-access journal like arxiv, then I'd much prefer that.

osu.Game.Rulesets.Osu/Difficulty/MathUtil/PoissonBinomial.cs
```diff
@@ -0,0 +1,50 @@
...
+
+            foreach (double p in probabilities)
+            {
+                sigma += p * (1 - p);
+                gamma += p * (1 - p) * (1 - 2 * p);
+            }
+
+            sigma = Math.Sqrt(sigma);
+
+            v = gamma / (6 * Math.Pow(sigma, 3));
```
bdach, line 35: This paren grouping confused me when I was cross-referencing the paper, since `gamma` here isn't actually the gamma given by the formulas here, so I wasn't sure where the 6 comes from (it comes from the formula for *G*(*x*)).

osu.Game.Rulesets.Osu/Difficulty/MathUtil/PoissonBinomial.cs
```diff
@@ -0,0 +1,50 @@
...
+        public double Cdf(double count)
+        {
+            double k = (count + 0.5 - mu) / sigma;
+
+            double result = Normal.CDF(0, 1, k) + v * (1 - k * k) * Normal.PDF(0, 1, k);
+
+            if (result < 0) return 0;
+            if (result > 1) return 1;
+
+            return result;
```
bdach, line 47: Can use `Math.Clamp`.

osu.Game.Rulesets.Osu/Difficulty/MathUtil/Mean.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+
+namespace osu.Game.Rulesets.Osu.Difficulty.MathUtil
+{
+    public static class Mean
```
bdach, line 9: References to formulae used could be nice.

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs
```diff
@@ -13,28 +13,51 @@
...
 using osu.Game.Rulesets.Osu.Scoring;
 using osu.Game.Rulesets.Scoring;
 
 namespace osu.Game.Rulesets.Osu.Difficulty
 {
     public class OsuDifficultyCalculator : DifficultyCalculator
     {
-        private const double difficulty_multiplier = 0.0675;
+        private const double aim_multiplier = 0.641;
+        private const double tap_multiplier = 0.641;
```
bdach, line 25: Two distinct constants with the same value?

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs
```diff
@@ -13,28 +13,51 @@
...
         {
-            if (beatmap.HitObjects.Count == 0)
-                return new OsuDifficultyAttributes { Mods = mods, Skills = skills };
+            var hitObjects = beatmap.HitObjects as List<OsuHitObject>;
+
+            double mapLength = 0;
+            if (beatmap.HitObjects.Count > 0)
+                mapLength = (beatmap.HitObjects.Last().StartTime - beatmap.HitObjects.First().StartTime) / 1000 / clockRate;
+
+            double preemptNoClockRate = BeatmapDifficulty.DifficultyRange(beatmap.BeatmapInfo.BaseDifficulty.ApproachRate, 1800, 1200, 450);
```
bdach, line 43: I don't understand this variable name.

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs
```diff
@@ -13,28 +13,51 @@
...
 
-            double aimRating = Math.Sqrt(skills[0].DifficultyValue()) * difficulty_multiplier;
-            double speedRating = Math.Sqrt(skills[1].DifficultyValue()) * difficulty_multiplier;
-            double starRating = aimRating + speedRating + Math.Abs(aimRating - speedRating) / 2;
+            // Finger Control
+            double fingerControlDiff = FingerControl.CalculateFingerControlDiff(hitObjects, clockRate);
+
+            // Aim
+            var (aimDiff, aimHiddenFactor, comboTps, missTps, missCounts, cheeseNoteCount, cheeseLevels, cheeseFactors) =
+                Aim.CalculateAimAttributes(hitObjects, clockRate, strainHistory, noteDensities);
```
bdach, line 55: This tuple deconstruction is ridiculous. I'd strongly advocate for actually returning a proper struct with named members.

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs
```diff
@@ -49,36 +72,48 @@ protected override DifficultyAttributes CreateDifficultyAttributes(IBeatmap beat
...
         }
 
-        protected override Skill[] CreateSkills(IBeatmap beatmap) => new Skill[]
+        protected override DifficultyAttributes CreateDifficultyAttributes(IBeatmap beatmap, Mod[] mods, Skill[] skills, double clockRate)
         {
-            new Aim(),
-            new Speed()
-        };
+            throw new NotImplementedException();
+        }
```
bdach, line 116: What's the explanation for these losing their implementations?

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -12,14 +14,31 @@
...
+        /// </summary>
+        private const double total_value_exponent = 1.5;
+
+        /// <summary>
+        /// This exponent is used to convert throughput to aim pp and tap skill to tap pp.
+        /// </summary>
+        private const double skill_to_pp_exponent = 2.7;
+
+        /// <summary>
+        /// The first 0.5 miss doesn't count when we penalize misses
```
bdach, line 36: *Half* miss? What is a half miss?

Does this mean that the first miss carries half the normal penalty?

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -51,33 +75,52 @@ public override double Calculate(Dictionary<string, double> categoryRatings = nu
...
+                else
+                    comboBasedMissCount = 0;
+            }
+            else
+            {
+                double fullComboThreshold = beatmapMaxCombo - 0.1 * countSliders;
+                if (scoreMaxCombo < fullComboThreshold)
+                    comboBasedMissCount = fullComboThreshold / scoreMaxCombo;
+                else
+                    comboBasedMissCount = Math.Pow((beatmapMaxCombo - scoreMaxCombo) / (0.1 * countSliders), 3);
```
bdach, line 109: Is this approximation derived in some way, or was it just an experimentation-based "good enough" one?

bdach (commented):
Round 1; got to 20 comments, so posting first batch as not to kill github when trying to add more.

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -89,112 +132,162 @@ public override double Calculate(Dictionary<string, double> categoryRatings = nu
 
         private double computeAimValue()
         {
-            double rawAim = Attributes.AimStrain;
+            if (Beatmap.HitObjects.Count <= 1)
+                return 0;
 
-            if (mods.Any(m => m is OsuModTouchDevice))
-                rawAim = Math.Pow(rawAim, 0.8);
+            // Get player's throughput according to combo
```
bdach, line 138: "throughput" also requires explanation as I don't understand what it's supposed to mean in this context.

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
```diff
@@ -0,0 +1,631 @@
...
+using System.Collections.Generic;
+using MathNet.Numerics;
+using MathNet.Numerics.LinearAlgebra;
+using MathNet.Numerics.Interpolation;
+using osu.Game.Rulesets.Osu.Objects;
+using osu.Game.Rulesets.Osu.Difficulty.MathUtil;
+
+namespace osu.Game.Rulesets.Osu.Difficulty.Preprocessing
+{
+    public class OsuMovement
```
bdach, line 14: This class is mostly obfuscated magic. I don't understand what `coeffs` are trying to do or mean.

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
```diff
@@ -0,0 +1,631 @@
...
+
+            return movementWithNested;
+        }
+
+        /// <summary>
+        /// Calculates the movement time, effective distance and other details for the movement from obj1 to obj2.
+        /// </summary>
+        public static List<OsuMovement> ExtractMovement(OsuHitObject obj0, OsuHitObject obj1, OsuHitObject obj2, OsuHitObject obj3,
+                                                        Vector<double> tapStrain, double clockRate,
+                                                        bool hidden = false, double noteDensity = 0, OsuHitObject objMinus2 = null)
```
bdach, line 216: Is this operating on sequences of four consecutive hitobjects, sometimes also taking into account the one before last? What do the indices mean? It's not exactly clear.

bdach:
I was going to keep going, but I can't at this stage. I can't understand so much of this stuff that it's pointless to continue. It wasn't this bad at the start, but about at `OsuDifficultyCalculator` it's where it starts being too much.

At this point I don't know if xmldocs can ever be enough for this code. Maybe a full paper-style document is required to be able to even start understanding this stuff.

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
stanriders: Added xmldoc, but i think it needs some variable renaming as well but I'm not sure what will make it more readable considering all t01/12/etc vars

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyCalculator.cs
stanriders: New system isn't using hitobject strain based skills so it can't really fit into current skill implementation

bdach:
Just so we're on the same page here, as I see changes are being done to this PR: I am willing to help bring this to a mergeable state code-quality-wise the same way I did with the taiko rework, but I *really* need a proper exhaustive explanation of the methodology involved such that I'm not working blind. As I said before the current doc is not enough.

stanriders:
Problem is that some things were done by different people and the only one who knows the system completely is delta, but he disappeared like 2 months ago so writing a complete doc is a bit problematic

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
joseph-ireland: he attempts to get a lower bound on misses+slider breaks, but it's an approximation because he doesn't want to include missed slider ends, so the approximation varies continuously

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
joseph-ireland: if there were no slider ends, and if you had a combo of x out of max, then you must have at least ceil(max/x) slider breaks. It uses a continuous value instead of ceil because it's weird for PP to step discretely at certain combos, and it eases off towards the end because slider ends don't count.

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
joseph-ireland: "throughput" is a term from fitt's law, basically it's a measure of skill

osu.Game.Rulesets.Osu/Difficulty/Preprocessing/OsuMovement.cs
joseph-ireland: all the coeffs go into "corrections" which add or subtract from the "index of difficulty" basically just by tweaking them to make results look better (though there is reasoning for each one).

The coeffs stuff is pretty opaque, they are for "correction 0" and "correction 3", which apply to the position of note 0 and note 3. In these numberings, he defines note 2 as the one that we're actually trying to hit, note 0 is two before it and note 3 is one after it.

For each one of these there is a big list of "coeffs" which go into a big formula that defines how the position of 0 and 3 affect the difficulty of hitting note 2 - the position is first transformed to be in units where the prev note (i.e. note 1) is the origin and the current note (note 2) is somewhere on the positive x axis, then it's passed into this big formula.

I tried refactored the coeffs stuff into a different class and give all the parts better names, but it's still confusing, and pretty impossible to visualise in your head. There is external code to visualise all the coeffs stuff somewhere on wolfram cloud, but I haven't used it. 

Ultimately the coeffs stuff is just a way of defining a tweakable function from position of the note to a correction, 

I think it's probably best to replace this entirely with something simpler that gives the same results. Maybe just define some values on a grid of (angle, distance) and interpolate?

joseph-ireland:
I've read through most of the complicated bits and I think I understand most of it, I added some comments to your review there. I'm not sure how is best to proceed - it's such a big change that it's pretty much its own program. I don't think the pull request format is the best way to manage it

bdach:
Although this is a huge change it's probably not very splittable, so it would have been fine if the doc was in-depth enough. However the comments I left aren't a full review, as I stopped quite early, and frankly I'm not too interested in spending time on reverse engineering what the rest of this code does instead of actually helping finish the game.

stanriders:
While the complete doc probably isn't going to happen any time soon, the main "magic" part has been reworked so it should be much easier to work on in case you want to try again

bdach:
Is this deemed ready? I'd rather not approach this again until it is. A ~3k line diffstat is almost an entire weekend's worth of time to understand and try to make legible.

stanriders:
We have a couple of things to look at but it's mostly tweaks and not major things

bdach:
I guess I'll put this on cooldown for a week or something then and if there won't have been any significant I'll approach this again.

stanriders:
Sure, I'll ping you when we're done

MitsukaiDokuro:
While your proposal definitly looks like an improvement to the current system, it's too simple. There is more to a (possibly modded) map than just what is called tap, aim and accuracy in pdf. I have thought about how best to divide osu skills/difficulties into categories when I discovered http://osuskills.com. Since osu skill and map difficulty categories are pretty much two sides of a coin, I decided to call it aspect instead. Here's what I have so far, feedback is very welcome:

- **reaction**
-  - _desc_: time to react (basically of AR)
-  - _impl_: based on approach time derived from AR value
- **accuracy**
-  - _desc_: timing an object (basically OD)
-  - _impl_: based on hit window derived from OD value
- **precision**
-  - _desc_: hitting an object (basically CS)
-  - _impl_: based on circle area derived from CS value
- **tempo**
-  - _desc_: hit object density (somewhat correlating to BPM)
-  - _impl_: based on hit object count per interval
- **reading**
-  - _desc_: grasping of what to hit where in which order, related to reaction aspect
-  - _impl_: based on visible object count, visible object area resulting from overlapping
- **rhythm**
-  - _desc_: rhythmical patterns / deviation of object timing, related to accuracy aspect
-  - _impl_: based on object's beat snap differences
- **random**
-  - _desc_: spacial layout deviation / deviation of object placement, related to precision aspect
-  - _impl_: based on angles and time-distance differences between objects or within (sliders)
- **stamina**
-  - _desc_: physical strength/endurance, related to rhythm and tempo aspect
-  - _impl_: based on strain derived from rhythm and tempo aspects
- **memory**
-  - _desc_: memory recalling, related to reading aspect
-  - _impl_: based on visible object area of the next object, from the perspective of the current object (thinking of FL mod)
- **consistency**
-  - _desc_: maintaining performance-per-aspect over time
-  - _impl_: based on map length/object count (difficulty) or based on individual aspect's performance drops over time/object count

bdach:
With all due respect, @MitsukaiDokuro, vague proposals like yours don't help much. The current system definitely isn't "simple" in its implementation.

The devil is in the details. If you show particular metrics that correlate with your particular chosen "aspects", then we can talk.

MitsukaiDokuro:
@bdach https://github.com/MitsukaiDokuro/OsuAspects, and yes I know, having implemented the easiest aspects reaction, accuracy and precision isn't much of a complete rating system. It's more of an personal experiment than a sophisticated system.

stanriders:
Please direct any unrelated proposals to https://github.com/ppy/osu-performance/issues this pull request isnt the place

MitsukaiDokuro:
It's related thanks for the hint though

stanriders:
@bdach we're done for now, any additional changes will be done after refactoring so that it doesnt stall the progress any more

bdach:
Thanks for the heads up. I'll try to get to it this weekend at the latest.

osu.Game.Rulesets.Osu/Difficulty/Interp/HermiteSpline.cs
```diff
@@ -0,0 +1,41 @@
...
+            if (x > X1)
+                return (x - X1) * D1 + Val1;
+            if (x < X0)
+                return (x - X0) * C1 + C0;
+
+            double t = (x - X0);
+            double t2 = t * t;
+            double t3 = t2 * t;
+
+            //return C0 + t * (C1 + t * (C2 + t * C3));
```
tybug, line 36: remove this commented code?

osu.Game.Rulesets.Osu/Difficulty/MathUtil/FittsLaw.cs
```diff
@@ -0,0 +1,88 @@
...
+        public static double CalculateHitProb(double d, double mt, double tp)
+        {
+            if (d == 0)
+                return 1.0;
+
+            if (mt * tp > 50)
+                return 1.0;
+
+            if (mt <= 0.03)
+                mt = 0.03;
```
tybug, line 32:
```suggestion
mt = Math.Max(mt, 0.03);
```

osu.Game.Rulesets.Osu/Difficulty/MathUtil/Mean.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public static double PowerMean(double x, double y, double i) =>
+            Math.Pow((Math.Pow(x, i) + Math.Pow(y, i)) / 2, 1 / i);
+
+        public static double PowerMean(IEnumerable<double> values, double i)
```
tybug, line 14: I see that wikipedia uses `p` as the power variable instead of `i`, probably worth changing from `i` at least to avoid confusion with an index

Liam-DeVoe (changes requested):

osu.Game.Rulesets.Osu/Difficulty/MathUtil/Mean.cs
bdach: This should all be xmldoc'd with references to formulae anyway. I'll grab all of this sort of stuff once I do a pass over this.

bdach:
Just to be clear, I failed to get to this due to playing catch-up on notifications this weekend, but I *am* still planning to go over this. Next week will actually probably be convenient as I have the entire week off, so I should be able to put an hour or two into cleaning this up every day.

osu.Game.Rulesets.Osu/Difficulty/Interp/HermiteSpline.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            X0 = x0;
+            X1 = x1;
+            D1 = d1;
+            Val1 = val1;
+
+            C0 = val0;
+            C1 = d0;
+            C2 = 3 * (val1 - val0) * scale2 - (2 * d0 + d1) * scale;
+            C3 = (2 * (val0 - val1) * scale + d0 + d1) * scale2;
```
bdach, line 21: are you sure this is correct? I mainly mean the `scale` part - I assume you're adding it to rescale *x* values onto the unit interval, but if that's the case, shouldn't `C1` be scaled by `scale`, `C2` by `scale2` entirely, and `C3` by `scale3` (which doesn't exist yet) entirely?

osu.Game.Rulesets.Osu/Difficulty/Interp/CubicInterp.cs
```diff
@@ -0,0 +1,68 @@
...
+            return (valNext - val) / (xNext - x);
+        }
+
+        public static double ThreePointDerivative(double xPrev, double valPrev, double x, double val, double xNext, double valNext)
+        {
+            return (
+                (xNext - x) * TwoPointDerivative(xPrev, valPrev, x, val)
+                + (x - xPrev) * TwoPointDerivative(x, val, xNext, valNext)
+                ) / (xNext - xPrev);
+        }
```
bdach, line 25: I'm lost as to what is the purpose of this. Isn't this essentially tantamount to `TwoPointDerivative(xPrev, valPrev, xNext, valNext)`, but with more possibility of rounding errors?

joseph-ireland: It's not the same, that first line is `(xNext-x) * (val-valPrev)/(x-xPrev)`. It doesn't cancel out to anything. 

The idea was to weight each side of the derivative by how close it is. If you do `TwoPointDerivative(xPrev, valPrev, xNext, valNext)` then it's kind of the same but backwards, the furthest away point gets the most weight.

bdach: I see, I was assuming a uniform grid. Thanks for clarifying.

bdach:
I've made a first batch of changes over at https://github.com/ppy/osu/compare/master...bdach:delta-rework-wip - I merged `master`, resolved the conflicts (which consisted of adapting the changes done here to the new attribute-based model of pp calc instead of the old one that had access to the beatmap), and also cleaned up/documented the interpolation helpers. Obviously nowhere near done yet.

I plan on pushing the changes out in one big batch, but if the concern raised above: re the implementation of the Hermite spline segment is correct (which I'm pretty sure it is) I'm relatively certain even today's changes will already impact results.

osu.Game.Rulesets.Osu/Difficulty/Interp/HermiteSpline.cs
joseph-ireland: I'm fairly sure it's right. 

x values are translated from (x1,x2) to (0,x2-x1), but not scaled down to (0,1). I just cross referenced against scipy [here](https://github.com/scipy/scipy/blob/v1.5.4/scipy/interpolate/_cubic.py#L136), `1/dxr` is `scale`, `slope` is `(val1-val0)*scale`

joseph-ireland: nvm, it's much easier to cross reference it [here](https://github.com/mathnet/mathnet-numerics/blob/8cdd91d37f36cb4daecc7838898a08ef9c3aa1a4/src/Numerics/Interpolation/CubicSpline.cs#L77) there's less stuff factored out.

bdach: Okay, after staring at this for a long time I think I get it. It's not that `x` values are not scaled down, it's that derivatives are *scaled back up* by `1 / scale` again.

stanriders:
I've deployed your branch on https://testpp.stanr.info/ (gasha:singletap) so you can compare results to a clean version, though you'll probably need to ping me whenever you want to update it (StanR#3012)
So far things are very wild for some reason

bdach:
Interesting. As pointed out above the interpolation code did seem to have a pretty fundamental bug before my changes, so fixing it might have impacted results, but not sure if to this degree. I haven't so much as looked toward how that stuff factors into calculations, though.

osu.Game.Rulesets.Osu/Difficulty/Interp/HermiteSpline.cs
joseph-ireland: Just checked the refactor, you've changed `3*(...) * scaleSquared - 2*(...)*scale` to `(3*(...)* scale -2*(...))*scale`. Should still work. I personally would prefer not to nest brackets like that when it's not necessary, but it shouldn't change any results. 

If there are discrepancies in the refactor they are elsewhere.

bdach: i mean it's a bit strange that c2 doesn't have nested brackets but c3 does, but sure, i guess

bdach:
I've force-pushed changes to my branch, after which the code should behave more as expected. One was my bad (forgot to bring back one instance of the three-point derivative approximation), one was an off-by-one that has no effect here, but surface after I changed lists to arrays:

https://github.com/ppy/osu/blob/73cb12cae4b36e1a18cb751ed9e8f040c8bead32/osu.Game.Rulesets.Osu/Difficulty/Interp/CubicInterp.cs#L39-L44

`splines` reserved `Length` items but only ever added `Length - 1`, but since passing count to `List` just pre-allocates memory it had no observable effect. I've also updated `OsuDifficultyCalculatorTest` that was not updated in here to catch these sorts of issues easier without having to reprocess everything.

peppy:
Making such knee-jerk changes reduces the credibility of this PR heavily and I'd advise against it. Any change to the algorithm should be documented with more than four words.

stanriders:
Badeu's play made us look at the issue properly. There were a lot of overweighted CS7+ maps that were overweighted just cos of the high cs buff. Change wasn't made to "nerf one play", but to fix issue that became obvious after this play.
Properly documenting changes in a wip system is redundant and until refactoring is complete it'll stay wip, nobody locked it down.

bdach:
Just so it's out in the open, I'm still in the process of trying to get through and understand this code but recently I'm struggling to find time for it alongside normal lazer traffic, so I'm not going to demand that any changes wait until I'm done.

apollo-dw:
We're (pp committee) taking the decision to close this PR due to a lack of interest and activity on pretty much everyone's behalf: I really doubt that tackling this PR in "massive overhaul all at once" mode is feasible. Select changes and ideas have been making their way into master incrementally anyways, and I suspect that this will continue into the future (which would be better than reviewing *this whole thing*)...

bdach:
Having stared at this for a while, I guess I haven't left an update here either, so just for the record - I've pretty much given up on this diff as well. I've invested a significant amount of time into it last year (almost a week's worth) and would need to invest several times more to even approach a level of basic understanding of everything that this diff did. So as far as I'm concerned this was always going to get closed sooner or later.

PR #9112, opened 2020-05-26
Allow SelectionHandler to provide custom context menu items without local hover check

peppy:
Previous virtual method would require each implementation to do the validity checks. Luckily, wasn't being used yet.

As a future change, the context menu will likely need to be moved to the `DragBox` for visual reasons, but just going to fix this for now.


smoogipoo (approved):
Seems fairly sane

PR #9115, opened 2020-05-26
Split out IHasPath from IHasCurve to better define hitobjects

peppy:
This fixes the future osu!taiko editor attempting to adjust repeat count of `DrumRoll`s when there is no such concept (only implementing the interface for legacy encoding reasons).

Test is failing but I think the failure may actually just be the output requiring an update. I'm not 100% how to read the tests to @smoogipoo checking/advice appreciated.


smoogipoo:
Unfortunately the output isn't good for this test since it's using the default nunit assert.

If you alter the test class to `private static IEnumerable<string> allBeatmaps => TestResources.GetStore().GetAvailableResources().Where(res => res.EndsWith("sample-beatmap-taiko.osu"));` you'll get the failure beatmap.

A further breakpoint on the assertion and inspecting the values (maybe by just evaluating the `.Serialize()` expressions of both), will show that the "decoded" beatmap has drum rolls at T=3553,3906,4612,5318, and the encoded-then-decoded beatmap has no drumrolls there (looks completely skipped).

osu.Game/Beatmaps/Formats/LegacyBeatmapEncoder.cs
```diff
@@ -233,9 +233,9 @@ private void handleHitObject(TextWriter writer, HitObject hitObject)
             writer.Write(FormattableString.Invariant($"{(int)getObjectType(hitObject)},"));
             writer.Write(FormattableString.Invariant($"{(int)toLegacyHitSoundType(hitObject.Samples)},"));
 
-            if (hitObject is IHasCurve curveData)
+            if (hitObject is IHasPathWithRepeats curveData)
```
bdach, line 236: Tests seem to pass when this is changed to

```suggestion
            if (hitObject is IHasPath curveData)
```

As noted by @smoogipoo drumrolls weren't entering this code path. This seems like an accidental mistype more than anything, seeing that `addPathData` already accepts an `IHasPath` instance.

bdach (dismissed):

smoogipoo (dismissed):

smoogipoo:
So I tested this against user rulesets and had to make a few changes around the `IHasPathWithRepeats` interface.

First, user rulesets are already matching against `IHasCurve` (see https://github.com/EVAST9919/bosu/blob/master/osu.Game.Rulesets.Bosu/Extensions/CherriesExtensions.cs), so we need some way to convey that all `IHasPathWithRepeat` convert objects are `IHasCurve` objects.

You may be led to believe it's enough to invert the inheritance order, but since `Path` was removed from `IHasCurve`, 1: anything referencing the path directly will fail, 2: anything referencing one of the extension methods will also fail, because the extension method class itself was changed.

The way I resolved both of these was completely splitting out `IHasCurve` and implementing it for `ConvertSlider`. Since `ConvertSlider` already implements all elements, this is a super simple change.

smoogipoo (approved):

smoogipoo:
Go ahead and merge if you're fine with the changes I applied.

peppy:
Looks fine 👍

PR #9119, opened 2020-05-26
Allow GracefullyExit() to not presist on exiting

frenzibyte:
Useful when testing blocking exits.

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -934,14 +934,14 @@ namespace osu.Game
 
         /// <summary>
         /// Use to programatically exit the game as if the user was triggering via alt-f4.
-        /// Will keep persisting until an exit occurs (exit may be blocked multiple times).
         /// </summary>
-        public void GracefullyExit()
+        /// <param name="repeatExit">Whether to keep presisting until an exit occurs (exit may be blocked multiple times).</param>
+        public void GracefullyExit(bool repeatExit = false)
         {
             if (!OnExiting())
                 Exit();
-            else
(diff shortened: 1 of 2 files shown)


osu.Game/OsuGame.cs
```diff
@@ -934,14 +934,14 @@ protected override bool OnExiting()
 
         /// <summary>
         /// Use to programatically exit the game as if the user was triggering via alt-f4.
-        /// Will keep persisting until an exit occurs (exit may be blocked multiple times).
         /// </summary>
-        public void GracefullyExit()
+        /// <param name="repeatExit">Whether to keep presisting until an exit occurs (exit may be blocked multiple times).</param>
```
bdach, line 938: Did you mean:

```suggestion
        /// <param name="repeatExit">Whether to keep persisting until an exit occurs (exit may be blocked multiple times).</param>
```

osu.Game/OsuGame.cs
```diff
@@ -934,14 +934,14 @@ protected override bool OnExiting()
...
+        public void GracefullyExit(bool repeatExit = false)
```
bdach, line 939: Why is `false` the default?

bdach (commented):
I don't think I like this direction. To have this exposed just for testing feels too much. Why not just override `OsuGameTestScene.CreateGame()` and provide a subclass of `TestOsuGame` that does
```csharp
protected override bool OnExiting() => false;
```
so that the exit never actually happens? It seems like it'll be equivalent...?

frenzibyte:
Closing as useless.

Issue #9117, opened 2020-05-26
osu!mania hit lighting doesn't rewind correctly


mas-alone:
**Describe the bug:*If you rewind while playing beat map, the special effects sometimes make mistakes
* 

**Screenshots or videos showing encountered issue:*
![bug](https://user-images.githubusercontent.com/52590027/82907314-08857c80-9f99-11ea-9f19-d1906fb2f14f.jpg)
![bug2](https://user-images.githubusercontent.com/52590027/82907323-0cb19a00-9f99-11ea-93dd-92c2ee79ae6c.jpg)

* 

**osu!lazer version:*2020.523.0
* 

**
[logs.zip](https://github.com/ppy/osu/files/4682567/logs.zip)
*

bdach:
Potentially related to #8082.

PR #9120, opened 2020-05-26
Disallow exiting from the game if un-cancellable work is active

frenzibyte:
- [ ] Depends on #9034 
- [ ] Depends on #9117 (for testing)

Separated for reviewing ease.

### DIFF
--- a/osu.Game.Tests/Visual/Menus/TestSceneSafeExiting.cs
+++ b/osu.Game.Tests/Visual/Menus/TestSceneSafeExiting.cs
@@ -1,0 +1,96 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
+using osu.Framework.Extensions.TypeExtensions;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Testing;
+using osu.Game.Configuration;
+using osu.Game.Input.Bindings;
+using osu.Game.Overlays;
+using osu.Game.Overlays.Notifications;
+using osu.Game.Screens.Menu;
+using osu.Game.Screens.Menu.Exiting;
+using osu.Game.Tests.Visual.Navigation;
+
+namespace osu.Game.Tests.Visual.Menus
+{
+    [System.ComponentModel.Description("Ensurance of exiting from game safely and intentionally")]
+    public class TestSceneSafeExiting : OsuGameTestScene
+    {
+        private MainMenu mainMenu => (MainMenu)Game.ScreenStack.CurrentScreen;
+        private NotificationOverlay notificationOverlay => Game.ChildrenOfType<NotificationOverlay>().Single();
+
+        [Test]
+        public void TestConfirmExitOnZeroHoldToExit()
+        {
+            performHoldToExit(0f);
+            waitForDialogAndClose<ConfirmExitDialog>();
+        }
+
+        [Test]
+        public void TestConfirmExitOnWindowCloseRequest()
+        {
+            requestCloseWindow();
+            waitForDialogAndClose<ConfirmExitDialog>();
+        }
+
+        [Test]
+        public void TestCannotExitWithUnCancellableWork()
+        {
+            doUnCancellableWork();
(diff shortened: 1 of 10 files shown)


osu.Game.Tests/Visual/Menus/TestSceneSafeExiting.cs
```diff
@@ -0,0 +1,96 @@
...
+using osu.Game.Input.Bindings;
+using osu.Game.Overlays;
+using osu.Game.Overlays.Notifications;
+using osu.Game.Screens.Menu;
+using osu.Game.Screens.Menu.Exiting;
+using osu.Game.Tests.Visual.Navigation;
+
+namespace osu.Game.Tests.Visual.Menus
+{
+    [System.ComponentModel.Description("Ensurance of exiting from game safely and intentionally")]
```
bdach, line 19:
```suggestion
    [System.ComponentModel.Description("Ensures exiting from game safely and intentionally")]
```

osu.Game/Screens/Menu/Exiting/SafeGameExiter.cs
```diff
@@ -0,0 +1,103 @@
...
+        }
+
+        /// <summary>
+        /// Performs an exit from the game with safety and intention checks.
+        /// </summary>
+        /// <remarks>
+        /// If called and returned <c>true</c> at least once, all calls afterwards will early-return with <c>true</c>.
+        /// </remarks>
+        /// <param name="enforceConfirmation">Whether to enforce showing the confirmation dialog.</param>
+        /// <returns>Whether to continue exiting if it is.</returns>
```
bdach, line 57: if it is...? I don't understand this.

osu.Game/Screens/Menu/Exiting/SafeGameExiter.cs
```diff
@@ -0,0 +1,103 @@
...
+                onSafeExitPassed();
+                return true;
+            }
+
+            // Block exit when the notification overlay has un-cancellable progress notifications.
+            if (notificationOverlay?.HasUnCancellableWork == true)
+            {
+                notificationOverlay.Show();
+                if (!(dialogOverlay.CurrentDialog is CannotExitDialog))
+                    dialogOverlay.Push(new CannotExitDialog("There are un-cancellable work active currently.", onExitBlock));
```
bdach, line 74: incorrect grammar (work is singular), but I'd reword overall to something like
```suggestion
                    dialogOverlay.Push(new CannotExitDialog("There is ongoing work that can't be cancelled.", onExitBlock));
```

osu.Game/Screens/Menu/Exiting/SafeGameExiter.cs
```diff
@@ -0,0 +1,103 @@
...
+        public bool PerformSafeExit(bool enforceConfirmation)
+        {
+            if (safeExitPerformed)
+                return true;
+
+            if (dialogOverlay == null)
+            {
+                onSafeExitPassed();
+                return true;
+            }
```
bdach, line 67: why is this branch checked before `notificationOverlay?.HasUnCancellableWork`? Requires explanation IMO.

Overall having paths return `true` before the paths that return `false` (save for checking `safeExitPerformed`) looks buggy.

peppy:
Applications shouldn't block close ever, that seems very anti-user.

Where is this requirement coming from? I don't recall any discussion of issues mentioning this as an issue (or suggesting this as the correct direction). I think we would rather fix the cases where you think a process cannot be cancelled to be cancellable.

frenzibyte:
> Applications shouldn't block close ever, that seems very anti-user.
> 
> Where is this requirement coming from? I don't recall any discussion of issues mentioning this as an issue (or suggesting this as the correct direction). I think we would rather fix the cases where you think a process cannot be cancelled to be cancellable.

Closing the game while it is installing a squirrel update will cause the game to never run back. (or that's what I got once while trying it)

peppy:
Can you report that issue (including the state of the game files when this occurred) so we can fix that rather than this?

frenzibyte:
> Can you report that issue (including the state of the game files when this occurred) so we can fix that rather than this?

Sure, will do once I get access to Windows.

peppy:
Highly recommend asking about these kinds of things (at very least opening an issue) before spending time on implementation. To avoid wasted time.

peppy:
Closing as incorrect direction.

Blocking user actions like exiting the application are not something we want to do.

PR #9121, opened 2020-05-26
Display a play's performance points on the results screen

### LINKED ISSUE #8587: PP Counter In Completion Screen Not Working
**Describe the bug:** 
After completing a map, the OSU PP counter does not display.
image

**osu!lazer version:** 2020.403.0

I dig some digging in the source code to try to find a solution myself, but I do not have enough cs experience to actually submit anything. It'd be nice if this score always displayed, even if the score wasn't submitted to the servers or if the play was unranked.

Game4all:
# Summary

Closes #8587

- This adds performance point calculation to `Player` using `PerformanceCalculators` so `ScoreInfo`s now get their pp field populated, and the pp shows on the results screen. 

Only "drawback" is that it won't display for older scores set before this PR.

## Testing

No tests for this as quite hard to test as it is.
Manual testings shows it to work as expected.

---

I have been thinking about actually moving the pp calculation to Results screen to do the pp calculation "on-demand" in background and show it on the results pp counter rather than calculating once and persisting it to database (potentially requiring to recalculate the whole database in case of a change in the pp algorithm) but this may not be the direction to go with.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -472,6 +472,16 @@ namespace osu.Game.Screens.Play
             else
                 score.User = api.LocalUser.Value;
 
+            try
+            {
+                var performanceCalculator = ruleset?.CreatePerformanceCalculator(Beatmap.Value, score);
+                var totalPp = performanceCalculator?.Calculate(new Dictionary<string, double>());
+                score.PP = totalPp;
+            }
+            catch (Exception)
+            {
+            }
+
             ScoreProcessor.PopulateScore(score);
 
             return score;
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -474,12 +474,13 @@ namespace osu.Game.Screens.Play
 
             try
             {
-                var performanceCalculator = ruleset?.CreatePerformanceCalculator(Beatmap.Value, score);
-                var totalPp = performanceCalculator?.Calculate(new Dictionary<string, double>());
+                var performanceCalculator = ruleset.CreatePerformanceCalculator(Beatmap.Value, score);
+                var totalPp = performanceCalculator?.Calculate(null);
                 score.PP = totalPp;
             }
-            catch (Exception)
+            catch (Exception e)
             {
+                Logger.Log($"Failed to calculate performance points: {e}", LoggingTarget.Runtime, LogLevel.Debug);
             }
 
             ScoreProcessor.PopulateScore(score);


osu.Game/Screens/Play/Player.cs
```diff
@@ -472,6 +472,16 @@ protected virtual ScoreInfo CreateScore()
...
 
+            try
+            {
+                var performanceCalculator = ruleset?.CreatePerformanceCalculator(Beatmap.Value, score);
+                var totalPp = performanceCalculator?.Calculate(new Dictionary<string, double>());
+                score.PP = totalPp;
+            }
+            catch (Exception)
+            {
+            }
```
bdach, line 483: why is this swallowed? Does it really fail this often to warrant outright silencing any potential exceptions?

mcendu:
Might be viable to handle unranked maps in a special manner, e.g. gray out the pp display if map is not ranked/approved? would be a separate pr though

osu.Game/Screens/Play/Player.cs
Game4all: The try-catch block is only here to wrap calls to Ruleset code (which is user code), so I'm not 100% sure on logging the (very few) potential exceptions here

bdach: If we're really worried about exposing these exceptions to the end user (which I don't really think we need to be), then this can always be logged at debug level so that the entries don't show up in release builds.

osu.Game/Screens/Play/Player.cs
```diff
@@ -472,6 +472,16 @@ protected virtual ScoreInfo CreateScore()
             else
                 score.User = api.LocalUser.Value;
 
+            try
+            {
+                var performanceCalculator = ruleset?.CreatePerformanceCalculator(Beatmap.Value, score);
+                var totalPp = performanceCalculator?.Calculate(new Dictionary<string, double>());
```
frenzibyte, line 478: Unnecessary?
```suggestion
                var totalPp = performanceCalculator?.Calculate();
```

osu.Game/Screens/Play/Player.cs
```diff
@@ -472,6 +472,16 @@ protected virtual ScoreInfo CreateScore()
             else
                 score.User = api.LocalUser.Value;
 
+            try
+            {
+                var performanceCalculator = ruleset?.CreatePerformanceCalculator(Beatmap.Value, score);
```
frenzibyte, line 477:
```suggestion
                var performanceCalculator = ruleset.CreatePerformanceCalculator(Beatmap.Value, score);
```

osu.Game/Screens/Play/Player.cs
frenzibyte: IMO it's kind-of weird to actually guard against a calculator..., not to mention that a similar component to this (`DifficultyCalculator`) is used without being wrapped in a try-catch. 

if anything, maybe let the calculator log and early-return if they find something incorrect.

osu.Game/Screens/Play/Player.cs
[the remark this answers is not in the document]
frenzibyte: same as what I said above: https://github.com/ppy/osu/pull/9121#discussion_r430880543

osu.Game/Screens/Play/Player.cs
bdach: 99% confident the above suggestion [will crash for custom rulesets](https://github.com/ppy/osu/blob/ad10a7f0b281f92aed90b614192d62fb05c0c656/osu.Game/Rulesets/Ruleset.cs#L156).

osu.Game/Screens/Play/Player.cs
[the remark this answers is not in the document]
mcendu: forgot to submit the review

bdach: Shouldn't this be passing the score's mods instead?

osu.Game/Screens/Play/Player.cs
frenzibyte: Read again, this is about access to ruleset instance.

osu.Game/Screens/Play/Player.cs
frenzibyte: Nope this dictionary is to pass the score ratings from the calculator (aim, speed, etc...). the mods are passed to the score itself above.

osu.Game/Screens/Play/Player.cs
Game4all: yep this null check is useless. nice catch.

osu.Game/Screens/Play/Player.cs
Game4all: the reason for guarding here is that `Ruleset.CreatePerformanceCalculator` may throw `NotImplementedException`; (as it is usercode) we may end up doing this everywhere in the codebase where we call into ruleset code.

frenzibyte: Guarding against that sounds like caring for a very specific case of failing when it could be easily fixed by the user actually fixing the code by returning null instead. (same can happen with any overridable `CreateXXX` method, you would need to wrap all that in a try-catch as well).

Moreover, afaik throwing a `NotImplementedException` is most likely to remind the developer to implement later on. so it should still crash intentionally. I still can't see a reason this is in a try-catch.

Game4all: > Guarding against that sounds like caring for a very specific case of failing when it could be easily fixed by the user actually fixing the code by returning null instead. (same can happen with any overridable `CreateXXX` method, you would need to wrap all that in a try-catch as well).
>

Wrapping in a try-catch block all interactions with rulesets in the codebase will have to be done at some point in time as we are calling user code and have got little control over what it can do (returning null, throwing exceptions, etc). Let's say I'm doing it there in an anticipated manner.
 
> Moreover, afaik throwing a `NotImplementedException` is most likely to remind the developer to implement later on. so it should still crash intentionally. I still can't see a reason this is in a try-catch.

`NotImplementedException` is part of one of the scenarios of what could go wrong with usercode (throwing exceptions). 


This justifies (at least to me) the need of a try-catch block there

peppy:
I think we want this to be calculated dynamically at the ranking screen, as you mentioned. Performance algorithm can change over time and persisting it to the database will result in it requiring recalculation anyway.

Maybe this is fine as a first step though?

Game4all:
I'm fine with adding calculation of performance upon completion for the time being, as it would "fill" the pp counter for the freshly set scores.

Concerning making the performance calculation fully dynamic, I remember having a branch lying around. I'll try to check it and pushing as soon I can but I can't guarantee anything for now.

As to make sure everything is clear to me: should score infos which were returned by API (ie: leaderboard scores) be also be dynamically computed even though their pp property is populated?

smoogipoo:
> should score infos which were returned by API (ie: leaderboard scores) be also be dynamically computed

It's fine not to.

PR #9123, opened 2020-05-27
Move setter from EndTime to Duration

peppy:
`EndTime` is dependent on `StartTime`, making it weird to set when StartTime may also be changed.


osu.Game/Rulesets/Objects/Types/IHasEndTime.cs
```diff
@@ -1,24 +1,12 @@
...
-using Newtonsoft.Json;
+using System;
 
 namespace osu.Game.Rulesets.Objects.Types
 {
-    /// <summary>
-    /// A HitObject that ends at a different time than its start time.
-    /// </summary>
-    public interface IHasEndTime
+    [Obsolete("Use IHasDuration instead.")] // can be removed 20201126
```
smoogipoo, line 8: This is in the opposite direction and will break user rulesets that already use `IHasEndTime`, rendering this obsoletion   notice useless. E.g.:

https://github.com/EVAST9919/bosu/blob/33ea85fef20340ab24168aa4f4372a46fa5df34d/osu.Game.Rulesets.Bosu/Beatmaps/BosuBeatmapConverter.cs#L45

https://github.com/LumpBloom7/sentakki/blob/2a637af0127fa17d7d8b07c5b199301377d777d4/osu.Game.Rulesets.Sentakki/Beatmaps/SentakkiBeatmapConverter.cs#L47

Can also be seen with our own ruleset - change the catch converted back to use `IHasEndTime` and run `CatchBeatmapConversionTest`.

Would you be against making `IHasDuration : IHasEndTime` instead and using interface implementations to fix the setters?

```cs
/// <summary>
/// A HitObject that ends at a different time than its start time.
/// </summary>
#pragma warning disable 618
public interface IHasDuration : IHasEndTime
#pragma warning restore 618
{
    double IHasEndTime.EndTime
    {
        get => EndTime;
        set => Duration = (Duration - EndTime) + value;
    }

    double IHasEndTime.Duration => Duration;

    /// <summary>
    /// The time at which the HitObject ends.
    /// </summary>
    new double EndTime { get; }

    /// <summary>
    /// The duration of the HitObject.
    /// </summary>
    [JsonIgnore]
    new double Duration { get; set; }
}
```

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Slider.cs
```diff
@@ -19,14 +19,14 @@ namespace osu.Game.Rulesets.Osu.Objects
 {
     public class Slider : OsuHitObject, IHasCurve
     {
-        public double EndTime
+        public double EndTime => StartTime + this.SpanCount() * Path.Distance / Velocity;
+
+        public double Duration
         {
-            get => StartTime + this.SpanCount() * Path.Distance / Velocity;
+            get => EndTime - StartTime;
```
bdach, line 26: a little inconsistent maybe that pretty much everything else was adjusted (`Duration` is the primary thing calculated in other places and `EndTime` is its derivative, and here it's the other way around), but it's not very significant

osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableSwell.cs
```diff
@@ -237,7 +237,7 @@ protected override void UpdateStateTransforms(ArmedState state)
 
                 case ArmedState.Miss:
                 case ArmedState.Hit:
-                    using (BeginAbsoluteSequence(Time.Current, true))
+                    using (BeginDelayedSequence(HitObject.Duration, true))
```
bdach, line 240: I can't seem to see this causing any change in behaviour, but it kind of looks like it would - what am I missing?

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -321,18 +321,18 @@ private PathControlPoint[] convertControlPoints(Vector2[] vertices, PathType typ
...
 
         /// <summary>
         /// Creates a legacy Hold-type hit object.
         /// </summary>
         /// <param name="position">The position of the hit object.</param>
         /// <param name="newCombo">Whether the hit object creates a new combo.</param>
         /// <param name="comboOffset">When starting a new combo, the offset of the new combo relative to the current one.</param>
-        /// <param name="endTime">The hold end time.</param>
-        protected abstract HitObject CreateHold(Vector2 position, bool newCombo, int comboOffset, double endTime);
+        /// <param name="duration">The hold end time.</param>
```
bdach, line 334:
```suggestion
        /// <param name="duration">The hold duration.</param>
```

osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableSwell.cs
peppy: Didn't really mean to bundle this with this PR. The old code would actually work incorrectly if a large time jump occurred without the presence of a `FrameStabilityContainer`, since it would base the animations on the new time rather than the point in time that the hit should have occurred (`HitObject.StartTime + HitObject.Duration`).

osu.Game.Rulesets.Osu/Objects/Slider.cs
peppy: Yeah, I think I mentioned in discord that a better direction here would be to write out the velocity to a private variable whenever any of the components affecting it change. Is that the direction you're thinking?

It needs to be done in order to have an interface-level `DurationBindable`, supported in `DrawableHitObject` just like `StartTime` is, which is one of the future goals I had. Just haven't done that yet as we were also discussing whether having bindable property counterparts in `HitObject`s was the correct direction.

bdach: Right, agreed. For the bindable to happen, internal recomputation of the duration needs to happen any time anything changes that would impact it.

I think there's a good chance this can be relatively self-contained (imagining something like `updateDuration()` inside `Slider` that would recompute this and raise value changed events) but if it pops up/starts being ridiculous in other places then a per-HO invalidation might be the way.

smoogipoo (approved):

PR #9125, opened 2020-05-27
Fix saving in the editor causing the current beatmap to change

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3583


smoogipoo (approved):
Not really a fan that the rename happened from a review standpoint, but I think this is fine...

peppy:
Currently causing a regression on master (was merged without a framework bump).

I'll fast-track a framework update to fix this.

smoogipoo:
Ah whoops, I saw that and hit merge anyway >_<

PR #9130, opened 2020-05-27
Alter the update manager structure to better expose ability to update

frenzibyte:
Helps quality of #8947

Alters the `UpdateManager` structure by moving the updating-logic piece into its own component and abstracts similar code from existing updaters to a base `Updater` class.

While this only benefits better exposing of update-operations-capability for use in the aforementioned PR, this makes things a bit more organized and better for future updaters, so I decided to PR it and know your opinions on it.

## Note
- Updater namespace has changed from `osu.Game.Updater` to `osu.Game.Updating` due to it conflicting with the `Updater` class name and causing annoying syntax issues from it.


bdach:
My opinion is that I don't get it. This PR exchanges an abstraction for an alternative, slightly more convoluted abstraction that in my eyes doesn't yield any discernible benefit.

peppy:
I don't think this is a good direction. Especially since in-game updating is going to change going forward (squirrel is likely not a final / permanent solution).

PR #9139, opened 2020-05-28
Add window size dropdown

jubiman:
Needs fix to avoid freeze on borderless resolutions and maximized screens

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -125,6 +125,8 @@ namespace osu.Game.Screens.Play
 
         private GameplayBeatmap gameplayBeatmap;
 
+        private ScreenSuspensionHandler screenSuspension;
+
         private DependencyContainer dependencies;
 
         protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
@@ -179,6 +181,7 @@ namespace osu.Game.Screens.Play
             InternalChild = GameplayClockContainer = new GameplayClockContainer(Beatmap.Value, Mods.Value, DrawableRuleset.GameplayStartTime);
 
             AddInternal(gameplayBeatmap = new GameplayBeatmap(playableBeatmap));
+            AddInternal(screenSuspension = new ScreenSuspensionHandler(GameplayClockContainer));
 
             dependencies.CacheAs(gameplayBeatmap);
 
@@ -628,12 +631,16 @@ namespace osu.Game.Screens.Play
 
         public override void OnSuspending(IScreen next)
         {
+            screenSuspension?.Expire();
+
             fadeOut();
             base.OnSuspending(next);
         }
 
         public override bool OnExiting(IScreen next)
         {
+            screenSuspension?.Expire();
+
             if (completionProgressDelegate != null && !completionProgressDelegate.Cancelled && !completionProgressDelegate.Completed)
             {
                 // proceed to result screen if beatmap already finished playing
--- a/osu.Game/Screens/Play/ScreenSuspensionHandler.cs
+++ b/osu.Game/Screens/Play/ScreenSuspensionHandler.cs
@@ -1,0 +1,52 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Diagnostics;
+using JetBrains.Annotations;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Platform;
+
+namespace osu.Game.Screens.Play
+{
+    /// <summary>
+    /// Ensures screen is not suspended / dimmed while gameplay is active.
+    /// </summary>
+    public class ScreenSuspensionHandler : Component
+    {
+        private readonly GameplayClockContainer gameplayClockContainer;
+        private Bindable<bool> isPaused;
+
+        [Resolved]
+        private GameHost host { get; set; }
+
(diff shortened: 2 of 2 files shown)


bdach:
The freeze is likely to be related to https://github.com/ppy/osu-framework/issues/3370.

bdach:
I'm not honestly sure how useful this is. I'd personally never use it (all I care about when window is whether the window is maximized or not). Might be just me, though.

omkelderman:
Just to drop my own 2cents: I would very much care about the window size, or at least the aspect ratio, for recording/streaming purposes

osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
```diff
@@ -140,15 +149,33 @@ private void load(FrameworkConfigManager config, OsuConfigManager osuConfig, Gam
...
+                    {
+                        windowSizeDropdown.Show();
+                        resolutionDropdown.Hide();
+                        windowedSize.TriggerChange();
+                    }
                     else
+                    {
                         resolutionDropdown.Hide();
+                        windowSizeDropdown.Hide();
+                    }
```
bdach, line 178: I'm not a fan of these three branches having their independent `Show()/Hide()` calls; that's bug-prone. Instead I'd probably do something like:

```csharp
resolutionDropdown.FadeTo(mode.NewValue == WindowMode.Fullscreen ? 1 : 0);
windowSizeDropdown.FadeTo(mode.NewValue == WindowMode.Windowed ? 1 : 0);
```

Unfortunately there's no good way to do the same thing with the `TriggerChange()` calls but I feel that would bring much readability and significantly de-branch this code.

bdach (changes requested):
I guess I stand corrected on whether this change would be useful to people. One note to the code, however.

osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
jubiman: I'll get to that tomorrow. I'm guessing that you mean to fully remove the if else statement and change it to a conditional statement. Might it be possible to get a value from `resolutionDropdown.Visible` or something like get and put it in a `switch()` statement to trigger the `TriggerChange()` call?

bdach: I'd keep the `TriggerChange()` calls in the if-else as they are right now, just hoist the dropdown visibility stuff outside of it and use the branchless versions above. Using dropdown visibility to trigger these bindable changes reads worse in my opinion.

bdach:
This time I actually ran this. On Windows this PR exposes several problems, some of which require addressing framework-side, so I'm not sure this is mergeable until they are addressed as we'll just get bug reports that the feature doesn't work.

* The one issue that can be addressed here is the fact that the `Default` option is just not valid for the windowed screen size dropdown, as it internally corresponds to a window size of 9999x9999:

  image

  so it should not even be an option.
* Secondly, choosing the highest non-default resolution will reliably trigger the framework bug mentioned above.
* Finally, attempting to use the window resolution dropdown on a maximised window will "look" like it's working inside the game (the on-screen display will pop up showing the new window size), but the window size will not actually change.

I would say that this addition is blocked pending the resolution of all above issues, but I'm willing to defer if there are differing opinions.

bdach (changes requested):
As above.

jubiman:
I removed the Default resolution by changing the `Items` variable from `Items = resolutions` to `Items = resolutions.ToList().GetRange(1, resolutions.Count - 1)`
About the freezing issue: I am going to look into what exactly is causing it and how to possibly precome it. Same goes for maximized windows

bdach:
I'm pretty sure I scoped the issue down to osuTK, most likely - please see aforementioned framework issue for findings.

bdach:
Also, your way of fixing the issue is (a) reliant on the fact that `Default` is first in the list, which should be at least documented with a comment, and (b) creates two copies of the list. You should be able to use `.Skip(1)` at least.

jubiman:
Oops, looked over that function when looking...

I think the best way to solve the maximized problem is to write a small algorithm to check the resolution and whether or not the screen is maximized, and if it is, change it to normal state. The only problem is that I don't know how to reference the `WindowState` of osu's window.

bdach:
`WindowState` is readable and settable via `game.Window`. Not sure if setting it will actually work.

jubiman:
It doesn't seem to work. I've also tried with `host.Window.WindowState`, but that also doesn't work. This might be something that has to do with the configuration in the framework

bdach:
I figured that might be the case. For now the sanest thing to do might be to disable the dropdown if the screen is maximised, preferably with a tooltip explaining why it is disabled.

jubiman:
Something seems off. When I try to get the `WindowState` in the `windowModeDropdown.Bindable.BindValueChanged` function I only get Normal `WindowState` as if it only does it on start up, and not real-time

jubiman:
It seems to have to do something with Bordeless and Fullscreen. Windowed mode is seen as "Normal", Bordeless as "Maximized" and Fullscreen as "Fullscreen". Even if I hit the Maximize button in the top right the `WindowState` seems to return "Normal"

peppy:
A good example of why you shouldn't pull request from a master branch.. ouch

peppy:
@jubiman did you want to push your code to a new branch and re-open a new pull request?

something like this:

```
git checkout -B window-size-dropdown;
git reset --hard ebdcfd5;
git push origin window-size-dropdown;
```

jubiman:
Oh oops, I might've did something wrong then. I tried to reset my fork's master to ppy/osu/master and keep my changes in a local branch. My mad

peppy:
In the future please remember to use a non-master branch for pull requests

jubiman:
Yeah that is what I was trying to create on my local/forked rep. Quite new to git...

PR #9142, opened 2020-05-29
Tidy up and complete xmldoc for HitObjectComposer

peppy:
Part of an ongoing effort to tidy up and make implementing ruleset editors understandable. Many parts to follow – this is just to set precedent and ensure it's the direction I should be going.

PR #9143, opened 2020-05-29
Pass hitobjects as a parameter to CreateBlueprintContainer

peppy:
Every implementation was manually accessing a property just for this. As a result, the exposed `HitObjects` property is no longer used in a public manner, but leaving it around for now.

This also adds a default implementation to make starting on a new editor easier.

- [x] Depends on #9142 just for ease-of-merge


bdach (approved):
seems simple enough 👍

smoogipoo (approved):

PR #9146, opened 2020-05-29
Implement osu!taiko editor

peppy:
Initial pass. Will be following up with further refactors, mostly to reduce local implementation. As it stands, after all the external refactoring I did the whole taiko editor implementation is under 400 lines of code. There's room to bring this down to 1-200 by standardising blueprint drawables (needs to be done anyway for consistency).

- [x] Depends on https://github.com/ppy/osu/pull/9143
- [x] Depends on https://github.com/ppy/osu/pull/9123
- [x] Depends on https://github.com/ppy/osu/pull/9115

There's a few visual fixes to blueprint / selection display (drumroll specifically) which I will likely push to this branch. Not necessarily blocked as it doesn't affect functionality, but should come today or tomorrow regardless.


bdach:
I've ran this. Looks to be working quite well. There are a few mishaps with blueprints (drumroll in general) but as you've mentioned them already I assume you're aware, so I won't bother going into detail.

I have found one weird issue with snapping. I can't really attach a video of this, as it's super subtle, but I noticed that if I drag an already-snapped object on a pre-existing map just enough to initiate drag (but not enough for it to move to a neighbouring beat), it will *ever so slightly* move position.

Debugging in I can see something not quite right going on:

image

I don't think it's something that would be caused by any of the changes here, but then again I haven't read the diff yet as it's a bit hard to tell which changes I have or haven't read before due to the dependency chain, so I held off on that a while.

peppy:
This may just be added precision which wasn't visible previously (due to osu-stable using `int`s for time values).

osu.Game.Rulesets.Taiko/Edit/Blueprints/TaikoSpanPlacementBlueprint.cs
```diff
@@ -0,0 +1,104 @@
...
+        private readonly HitPiece headPiece;
+        private readonly HitPiece tailPiece;
+
+        private readonly LengthPiece lengthPiece;
+
+        private readonly IHasDuration spanPlacementObject;
+
+        public TaikoSpanPlacementBlueprint(HitObject hitObject)
+            : base(hitObject)
+
```
smoogipoo, line 28:
```suggestion
```

osu.Game.Rulesets.Taiko/Edit/Blueprints/TaikoSpanPlacementBlueprint.cs
```diff
@@ -0,0 +1,104 @@
...
+                    {
+                        HitObject.StartTime = endTime;
+                        spanPlacementObject.Duration = Math.Abs(endTime - originalStartTime);
+                        headPiece.Position = ToLocalSpace(result.ScreenSpacePosition);
+                        lengthPiece.X = headPiece.X;
+                        lengthPiece.Width = tailPiece.X - headPiece.X;
+                    }
+                    else
+                    {
+                        spanPlacementObject.Duration = Math.Abs(endTime - originalStartTime);
```
smoogipoo, line 89: This probably needs to also set `HitObject.StartTime = originalStartTime`?

Behaves kinda weirdly: https://drive.google.com/file/d/1RbHwwtKOxjwUOCGAI8KtXFK7mJjrt6E_/view

osu.Game.Rulesets.Taiko/Edit/Blueprints/TaikoSelectionBlueprint.cs
```diff
@@ -0,0 +1,40 @@
...
+        protected override void Update()
+        {
+            base.Update();
+
+            // Move the rectangle to cover the hitobjects
+            var topLeft = new Vector2(float.MaxValue, float.MaxValue);
+            var bottomRight = new Vector2(float.MinValue, float.MinValue);
+
+            topLeft = Vector2.ComponentMin(topLeft, Parent.ToLocalSpace(DrawableObject.ScreenSpaceDrawQuad.TopLeft));
+            bottomRight = Vector2.ComponentMax(bottomRight, Parent.ToLocalSpace(DrawableObject.ScreenSpaceDrawQuad.BottomRight));
```
smoogipoo, line 34: This is a weird way of going about it :P

smoogipoo (changes requested):

osu.Game.Rulesets.Taiko/Edit/Blueprints/TaikoSelectionBlueprint.cs
peppy: This is plagiarised from `SelectionHandler`. I'd hope it's not a permanent thing but I'm going to be redoing these blueprints either way.

smoogipoo (approved):

PR #9149, opened 2020-05-29
Fix searching by genre/language not working in beatmap listing overlay

EVAST9919:


bdach (approved):
seems reasonable 👍 

Still not really a fan of the number of ctor parameters on `SearchBeatmapSetsRequest` but it hasn't changed here so...

osu.Game/Overlays/BeatmapListing/BeatmapListingFilterControl.cs
```diff
@@ -186,6 +188,9 @@ private void performRequest()
                 if (sets.Count == 0)
                     noMoreResults = true;
 
+                if (CurrentPage == 0)
+                    searchControl.BeatmapSet = sets.FirstOrDefault();
```
peppy, line 192: This **really** should have been in its own PR. Please don't mix fixes together again.

peppy:
Parameter count kinda has changed here. Really should be doing something like [FilterCriteria](https://github.com/ppy/osu/blob/master/osu.Game/Screens/Select/FilterCriteria.cs).

peppy (approved):
Required changes can be made in a follow-up PR.

PR #9158, opened 2020-05-30
Replace any potential usage of Environment.CurrentDirectory

frenzibyte:
- [x] Depends on ppy/osu-framework#3590

### DIFF
--- a/osu.Game.Tests/Resources/TestResources.cs
+++ b/osu.Game.Tests/Resources/TestResources.cs
@@ -18,14 +18,14 @@ namespace osu.Game.Tests.Resources
 
         public static string GetTestBeatmapForImport(bool virtualTrack = false)
         {
-            var temp = Path.GetTempFileName() + ".osz";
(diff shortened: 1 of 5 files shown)


osu.Game.Tests/Resources/TestResources.cs
```diff
@@ -20,10 +21,10 @@ public static string GetTestBeatmapForImport(bool virtualTrack = false)
             var temp = Path.GetTempFileName() + ".osz";
 
             using (var stream = GetTestBeatmapStream(virtualTrack))
-            using (var newFile = File.Create(temp))
+            using (var newFile = RuntimeInfo.StartupStorage.GetStream(temp, FileAccess.Write))
```
smoogipoo, line 24: A few more usages of `StartupStorage` that should be removed.

frenzibyte: Whoops, will remove

peppy (approved):

PR #9160, opened 2020-05-31
Fix osu!mania stage-left and stage-right sprites incorrectly scaling horizontallly

mcendu:
This is to prevent skins using small sprites for these components ([e.g. this from a skin I made for my own use](https://user-images.githubusercontent.com/43639171/83343883-dc6c5180-a332-11ea-8203-3891cdc6af8d.png)) from displaying weirdly.


peppy:
For reference can you provide before/after screenshots?

mcendu:
Before:
Before

After:
After

bdach:
I can see why this would happen and the fix looks good, but it might be even better to include a similar sprite for the skinning test scenes just to make sure this doesn't reappear. Any sprite that is with a significantly smaller height than the stage height would do.

I *think* it could fit in `special-skin`? Not overly sure what the meanings of the test skin names are supposed to indicate, however.

peppy:
special-skin is for edge cases, precisely like this. would be great to add it in.

bdach (approved):

PR #9161, opened 2020-05-31
Allow horizontal scroll on results screen when not hovering expanded panel

peppy:
Still need more consideration of ui scale (re-centering at very least) but with this change I'm happy to push this out in a release, at least.


bdach (approved):
Feels like an acceptable interim solution to me 👍

smoogipoo (approved):
Yeah this handles fine to me

PR #9162, opened 2020-05-31
Implement basic support for osu!mania special style

### LINKED ISSUE #2203: Implement osu!mania special styles
osu!mania beatmaps have a `SpecialStyle` parameter that indicates whether they're 5K+1 or 7K+1 styled. This parameter should be used to determine whether the special column should be mirrored (see discussion in: https://github.com/ppy/osu/issues/2067).

The user is allowed the ability to customise where this location is (left/right of the stage), or whether it is displayed as a special column at all.

Most importantly, `Beatmap` should not house a `SpecialStyle` property. Perhaps the best way to do this is to store unhandled beatmap values inside a lookup table of sorts, and have the mania beatmap converter pull the property from that lookup table.

mcendu:
*Note: S refers to special column.*
- Closes #2203
- S column's position should be switchable
- Co-op maps should use S-keys keys-S layout, instead of S-keys S-keys
- Mirror should not suppress mirroring of S when special style flag is enabled on non-6K-or-8K maps

### DIFF
--- a/osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
+++ b/osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
@@ -17,11 +17,39 @@ namespace osu.Game.Rulesets.Mania.Beatmaps
         public int Columns;
 
         /// <summary>
+        /// Whether the SpecialStyle flag is set.
+        /// </summary>
+        public bool SpecialStyle;
+
+        /// <summary>
+        /// Whether this is a special style map.
+        /// </summary>
+        /// <remarks>
+        /// Aside from 6K and 8K, special style is almost never used in other
+        /// keymodes, so this function always gives false for non-6/8K modes.
+        /// This also matches stable behavior of only allowing configuration
+        /// of special style in 6/8K.
+        /// </remarks>
+        public bool UseSpecialStyle()
+        {
+            if (Columns != 6 && Columns != 8)
+                return false;
+
+            return SpecialStyle;
+        }
+
+        /// <summary>
         /// Whether the column index is a special column for this stage.
         /// </summary>
         /// <param name="column">The 0-based column index.</param>
         /// <returns>Whether the column is a special column.</returns>
-        public bool IsSpecialColumn(int column) => Columns % 2 == 1 && column == Columns / 2;
+        public bool IsSpecialColumn(int column)
+        {
+            if (UseSpecialStyle())
+                return column == 0;
+
+            return Columns % 2 == 1 && column == Columns / 2;
+        }
 
         /// <summary>
         /// Get the type of column given a column index.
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
```diff
@@ -16,12 +16,40 @@ public struct StageDefinition
...
+        /// <summary>
+        /// Whether this is a special style map.
+        /// </summary>
+        /// <remarks>
+        /// Aside from 6K and 8K, special style is almost never used in other
+        /// keymodes, so this function always gives false for non-6/8K modes.
+        /// This also matches stable behavior of only allowing configuration
+        /// of special style in 6/8K.
+        /// </remarks>
+        public bool UseSpecialStyle()
```
mcendu, line 33: Could make `SpecialStyle` into a property, and use this as its getter.

osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
```diff
@@ -16,12 +16,40 @@ public struct StageDefinition
...
+        public bool SpecialStyle;
+
+        /// <summary>
+        /// Whether this is a special style map.
+        /// </summary>
+        /// <remarks>
+        /// Aside from 6K and 8K, special style is almost never used in other
+        /// keymodes, so this function always gives false for non-6/8K modes.
+        /// This also matches stable behavior of only allowing configuration
+        /// of special style in 6/8K.
```
bdach, line 31: "almost never" is not "never" - what does this intend to say? If it *does* get used in other keymodes then it should be supported.

I also don't think this should be limited to 6K and 8K even if stable does ignore the config value for other modes. The less magic conditionals, the better, in my view.

osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
```diff
@@ -16,12 +16,40 @@ public struct StageDefinition
...
         /// <summary>
         /// Whether the column index is a special column for this stage.
         /// </summary>
         /// <param name="column">The 0-based column index.</param>
         /// <returns>Whether the column is a special column.</returns>
-        public bool IsSpecialColumn(int column) => Columns % 2 == 1 && column == Columns / 2;
+        public bool IsSpecialColumn(int column)
+        {
+            if (UseSpecialStyle())
+                return column == 0;
```
bdach, line 49: I'm very much not a fan of magic conditionals like this, relying on the numerical index of a column to determine if it's special. It's very flaky logic, and will get even more flaky once switching the special column position is reintroduced (which I understand isn't supported at this stage of the PR?)

I'm thinking `StageDefinition` needs an overhaul. The key count and special flags should be passed via ctor to sanely initialise the stage definition. The ctor would then assign types to the columns, and this would reduce to a simple read of that type.

bdach (changes requested):

osu.Game.Rulesets.Mania/Beatmaps/StageDefinition.cs
smoogipoo: I think the user skin setting needs to be implemented before I can imagine the future direction of this, since that will completely change this implementation.

smoogipoo (changes requested):

bdach:
I am going to go ahead and close this draft due to several outstanding core issues, which have not been addressed in over a year, and because it is generally not achieving very much. Given the rather small size of the diff, reimplementing from scratch shouldn't really take more than half an hour maybe, and even if need be the pull will still be around and linked to the issue as a jump-off point.

PR #9166, opened 2020-06-01
JIT and GC optimization

hez2010:
1. Disable QuickJIT for best quality code at first time
QuickJIT will generate degraded performance code at startup to decrease startup time, but it may lead to more stack space, more memory allocation and slower speed. And while re-JIT during runtime, the application may freeze for short-time which may lead to frame drop.

2. Use Background Server GC for lower GC latency
Background Server GC may increase memory consumption but it significantly reduces GC times and lead to lower GC latency, which is suitable for game scenario.

### DIFF
--- a/osu.Desktop/osu.Desktop.csproj
+++ b/osu.Desktop/osu.Desktop.csproj
@@ -11,6 +11,9 @@
     <ApplicationManifest>app.manifest</ApplicationManifest>
     <Version>0.0.0</Version>
     <FileVersion>0.0.0</FileVersion>
+    <TieredCompilationQuickJit>false</TieredCompilationQuickJit>
+    <ServerGarbageCollection>true</ServerGarbageCollection>
+    <ConcurrentGarbageCollection>true</ConcurrentGarbageCollection>
   </PropertyGroup>
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


peppy:
Please provide profiling / benchmarks for such changes.

hez2010:
@peppy I'm figuring out a way for quantitative measurement and benchmark.

For differences between Background Server GC and other GC modes, here's a post from Microsoft: https://devblogs.microsoft.com/premier-developer/understanding-different-gc-modes-with-concurrency-visualizer

peppy:
We are already aware of the documented settings. Without benchmarks we cannot act on these changes.

Also note that we already [set the GC mode](https://github.com/ppy/osu-framework/blob/815f03f6fd7143e7531f37062451dc4edfc802f9/osu.Framework/Platform/GameHost.cs#L476), which may conflict with what you are changing here.

Also we do care about startup times so if that is regression it is also a consideration to have in mind.

bdach:
As a side note, please avoid PRing from `master` - topic branches are easier for us to work with.

peppy:
Closing this for now. Please reopen if you plan on providing data/benchmarks.

PR #9173, opened 2020-06-02
Add classic "Welcome" intro

MiraiSubject:
This adds an intro screen to osu! lazer with the soundtrack from older osu! stable versions (also known as fallback), nekodex - welcome to osu!

- [x] Depends on https://github.com/ppy/osu-resources/pull/102
- [x] Depends on https://github.com/ppy/osu-resources/pull/103
- [x] Depends on https://github.com/ppy/osu/pull/9236
- [x] Depends on ppy/osu-framework#3648


osu.Game/Configuration/IntroSequence.cs
```diff
@@ -6,6 +6,7 @@ namespace osu.Game.Configuration
     public enum IntroSequence
     {
         Circles,
+        Fallback,
```
peppy, line 9: Let's call it "Welcome" for now.

peppy:
If you're going to implement this, please make the welcome/seeya sound also match. also needs the piano reverse-reverb. All resources attached.

[resources.zip](https://github.com/ppy/osu/files/4715337/resources.zip)

peppy (changes requested):
As specified

MiraiSubject:
I added the welcome and seeya sounds that match this sequence to the screens and an initial implementation of the legacy start up animation. 

Right now I'm having some trouble restoring the glow using `GlowingSpriteText` so as an alternative for now an `OsuSpriteText` is used. 

The visualizer, shown below in fallback, is also still absent. I'm still thinking of a way on how to properly approach the implementation of that. 

osu!_2020-06-02_13-18-24

peppy:
Which glow are you talking about? It should be baked into the sprite.

I'll check the stable implementation if you want the exact transforms.

MiraiSubject:
I can use the original png sprite. In that case the original transforms would definitely be desired and make implementing it easier.

I had the thought that using the built-in Drawable for `OsuSpriteText` was preferred.

peppy:
Definitely use the sprite. I don't think the correct font is even available.

MiraiSubject:
I mentioned on the discord that right now the piano reverb is a separate mp3, but the `LogoVisualisation` does not take arbitrary audio streams. 

So right now I've prepended the piano reverb to the audio file inside `welcome.osz` to achieve the desired start up effect. 

Because I'm basically editing the original beatmap, I'd like to know if this is the desired direction or if I should approach it differently.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,152 @@
...
+                                AutoSizeAxes = Axes.Both,
+                                Children = new Drawable[]
+                                {
+                                    visualizer = new LogoVisualisation
+                                    {
+                                        RelativeSizeAxes = Axes.Both,
+                                        Anchor = Anchor.Centre,
+                                        Origin = Anchor.Centre,
+                                        Alpha = 0.5f,
+                                        AccentColour = Color4.Blue,
```
MiraiSubject, line 98: This does nothing at the moment. The accent color gets overriden in `LogoVisualisation`

peppy:
You could just playback the piano reverb manually. As a sample, similar to the welcome voice line.

MiraiSubject:
Alright, I have reverted the sample back!

> I'll check the stable implementation if you want the exact transforms.

May I have the original transforms of the `Welcome` sprite ?

MiraiSubject:
I've managed to make a rough approximation of the transforms by recording a video of the fallback intro and then by playing it back at a slower rate I was able to observe the individual animations happening on the `Welcome` sprite.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,163 @@
...
+                    welcomeText.FadeIn(remainingTime());
+                    welcomeText.ScaleTo(welcomeText.Scale + new Vector2(0.1f), remainingTime(), Easing.Out).OnComplete(_ =>
+                    {
+                        elementContainer.Remove(visualizer);
+                        circleContainer.Remove(blackCircle);
+                        elementContainer.Remove(circleContainer);
+                        Remove(welcomeText);
+                        visualizer.Dispose();
+                        blackCircle.Dispose();
+                        welcomeText.Dispose();
```
bdach, line 157: I'm not overly familiar with drawable lifetime stuff, but isn't `Expire()` what you actually want here? If you expire the elements they should automatically get cleaned up from the containers. [There's even a framework-side test case for that.](https://github.com/ppy/osu-framework/blob/2637ba3af75e5ef0d3e05388309e153e457f489c/osu.Framework.Tests/Containers/TestSceneContainerState.cs#L222-L239)

osu.Game/Screens/Menu/LogoVisualisation.cs
```diff
@@ -70,6 +70,7 @@ public class LogoVisualisation : Drawable, IHasAccentColour
 
         private IShader shader;
         private readonly Texture texture;
+        public bool IsIntro = false;
```
bdach, line 73: Flags added for specific purposes like this are always asking for trouble. It's doing things backwards; to reuse this visualisation for the legacy intro we should look for ways to *remove* the user colour changing logic for that particular use case entirely, and to keep it in all the other places that need it right now, and not ways to override it.

I can't say I see a very good way of achieving this right now, but I think extracting out a simplified variant of `LogoVisualisation` that doesn't have `User` or `Skin`, and then inheriting from it to add that functionality back where it's needed right now could be a decent idea to start with. That might end up being a separate prerequisite PR, however, depending if any renames/significant code reshuffle happens.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,163 @@
...
+        protected override void LogoArriving(OsuLogo logo, bool resuming)
+        {
+            base.LogoArriving(logo, resuming);
+
+            if (!resuming)
+            {
+                welcome?.Play();
+                pianoReverb?.Play();
+
+                Scheduler.AddDelayed(delegate
```
bdach, line 47: `delegate` is a really old thing now. Lambdas are a bit more hip and modern nowadays.

```suggestion
                Scheduler.AddDelayed(() =>
```

bdach (dismissed):
Haven't ran this yet, just skimmed the source.

osu.Game/Screens/Menu/IntroWelcome.cs
MiraiSubject: Yeah you're right! That was just a blind copy-paste from `IntroCircles` really. I'll include this with the change for the drawable lifetimes.

osu.Game/Screens/Menu/LogoVisualisation.cs
MiraiSubject: I've implemented this in https://github.com/ppy/osu/pull/9236. The naming for the new base class for `LogoVisualisation` might be a bit questionable though.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,149 @@
...
+        protected override string BeatmapHash => "64e00d7022195959bfa3109d09c2e2276c8f12f486b91fcf6175583e973b48f2";
+        protected override string BeatmapFile => "welcome.osz";
+        private const double delay_step_two = 2142;
+        private SampleChannel welcome;
+        private SampleChannel pianoReverb;
+
+        [BackgroundDependencyLoader]
+        private void load(AudioManager audio)
+        {
+            Seeya = audio.Samples.Get(@"Intro/welcome/seeya");
```
peppy, line 29: unnecessarily loading two samples (the base is also loading). revert the `SampleChannel` to private and make a `virtual string SeeyaSampleName => "Intro/seeya"`

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,149 @@
...
+
+            if (MenuVoice.Value)
+            {
+                welcome = audio.Samples.Get(@"Intro/welcome/welcome");
+                pianoReverb = audio.Samples.Get(@"Intro/welcome/welcome_piano");
```
peppy, line 34: piano revert plays even when menuvoice is off

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,149 @@
...
+                    },
+                    welcomeText = new Sprite
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Scale = new Vector2(0.3f),
+                        Width = 750,
+                        Height = 78,
+                        Alpha = 0,
+                        Texture = textures.Get(@"Welcome/welcome_text@2x")
```
peppy, line 129: `@2x` shouldn't be here. everything in lazer is @2x.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,149 @@
...
+                                }
+                            }
+                        }
+                    },
+                    welcomeText = new Sprite
+                    {
+                        Anchor = Anchor.Centre,
+                        Origin = Anchor.Centre,
+                        Scale = new Vector2(0.3f),
+                        Width = 750,
```
peppy, line 126: width and height specification should not be required. it's automatic when you set a texture.

osu.Game/Screens/Menu/IntroWelcome.cs
```diff
@@ -0,0 +1,149 @@
...
+                        Texture = textures.Get(@"Welcome/welcome_text@2x")
+                    },
+                };
+            }
+
+            protected override void LoadComplete()
+            {
+                base.LoadComplete();
+
+                double remainingTime() => delay_step_two - TransformDelay;
```
peppy, line 138: shouldn't be a function. also not sure this is correct.

peppy (dismissed):

osu.Game/Screens/Menu/IntroWelcome.cs
MiraiSubject: The texture doesn't display at all when I remove `@2x`, I'd have assumed first as well that it wouldn't be needed, but it's actually not visible on my side. 

`@2x` removed, so `textures.Get(@"Welcome/welcome_text")`
dotnet_2020-06-13_16-28-06

`@2x` added
dotnet_2020-06-13_16-27-39

osu.Game/Screens/Menu/IntroWelcome.cs
MiraiSubject:
```csharp
Width = 750,
Height = 78,
```
This is hardcoded like this to prevent the game from proportionally scaling it. The original dimensions for the file are `750*156`

`welcomeText.ResizeHeightTo(welcomeText.Height * 2, 500, Easing.In);` resizes the height from `750*78` to `750*156`

peppy:
I've tidied things up. The main issue remaining is that the menu visualisation does not correctly animate due to not receiving the amplitudes of the piano reverb sample. I'll look into this further - a fix will either require some changes to the sample playback logic or combination of the sample into the track (may actually be the simplest way for now).

peppy:
I've applied the final touches to this, including reading the amplitudes from the reverb sample (requires a framework change).

smoogipoo:
The new amplitude source stuff crashes when changing rulesets in song select:

```
[runtime] 2020-06-23 07:21:44 [error]: System.NullReferenceException: Object reference not set to an instance of an object.
[runtime] 2020-06-23 07:21:44 [error]: at osu.Game.Screens.Menu.LogoVisualisation.addAmplitudesFromSource(IHasAmplitudes source) in /home/smgi/Repos/osu/osu.Game/Screens/Menu/LogoVisualisation.cs:line 161
[runtime] 2020-06-23 07:21:44 [error]: at osu.Game.Screens.Menu.LogoVisualisation.updateAmplitudes() in /home/smgi/Repos/osu/osu.Game/Screens/Menu/LogoVisualisation.cs:line 113
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal() in /home/smgi/Repos/osu-framework/osu.Framework/Threading/ScheduledDelegate.cs:line 88
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Threading.Scheduler.Update() in /home/smgi/Repos/osu-framework/osu.Framework/Threading/Scheduler.cs:line 102
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Drawable.cs:line 477
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.updateChild(Drawable c) in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 883
[runtime] 2020-06-23 07:21:44 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree() in /home/smgi/Repos/osu-framework/osu.Framework/Graphics/Containers/CompositeDrawable.cs:line 864
```

peppy:
ah yes, because they can still be null in some cases for whatever reason. shall we fix that framework-side by initialising the amplitude array to an empty array, or leave as nullable?

smoogipoo:
I'd probably prefer initialising to an empty array...

smoogipoo (approved):
Seems good

PR #9174, opened 2020-06-02
Update framework

peppy:
Includes https://github.com/ppy/osu/pull/9158 but seems to be broken, let's see what CI says.

### DIFF
--- a/osu.Game.Tests/Resources/TestResources.cs
+++ b/osu.Game.Tests/Resources/TestResources.cs
@@ -18,14 +18,14 @@ namespace osu.Game.Tests.Resources
 
         public static string GetTestBeatmapForImport(bool virtualTrack = false)
         {
-            var temp = Path.GetTempFileName() + ".osz";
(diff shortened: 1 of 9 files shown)


osu.Game.Tests/Resources/TestResources.cs
```diff
@@ -17,14 +18,14 @@ public static class TestResources
 
         public static string GetTestBeatmapForImport(bool virtualTrack = false)
         {
-            var temp = Path.GetTempFileName() + ".osz";
+            var tempPath = Path.Combine(RuntimeInfo.StartupDirectory, Path.GetTempFileName() + ".osz");
```
smoogipoo, line 21: This is documented as returning the full path to the temporary file (https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netcore-3.1). I don't think we want to insert the startup directory here.

Can you please check to make sure this isn't creating a local subfolder?

peppy: this is definitely wrong. should be reverted.

smoogipoo (approved):

PR #9175, opened 2020-06-02
Add timeshift beatmap checksum validation

### LINKED ISSUE #9132: Self-edited beatmap can be played in multiplayer
**Describe the bug:** 
It is possible to play the self-modified map in multiplayer

**Screenshots or videos showing encountered issue:** https://youtu.be/z-zDds7_M9U

**osu!lazer version:** 2020.530.0.

**Logs:** [runtime.log](https://gist.github.com/ilsubyeega/b3e04e3883c70022a8139796a734a575)

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9132

A few changes here:

1. I've added an all-encompassing `PreUpdate()` method that follows the `PreImport()` signature for this task. I imagine this method to be used in the future to queue up diffcalc re-computations, etc.
2. The download button does a second local query against the beatmap checksum. This is done this way because `DownloadTrackingComponent` only knows about the parenting `TModel` (beatmap set) and `ArchiveModelManager<TModel>`. I think maybe we'll want to extract this logic out, but doing so is quite a large refactoring which I'll save for a further PR.
3. `MatchSubScreen` now always re-queries for the working beatmap when a change happens on the `BeatmapManager`. `Save()` removes the working beatmap from the cache and expects consumers to re-query for the updated beatmap. If the current beatmap isn't changed, the cache will still be hot and the re-query should be pretty much instantaneous.
4. `TestBeatmap` now populates MD5 + SHA2 hashes for testability.

Tested:
- Entering room, downloading updated beatmap, playing the downloaded version.
- Creating a room, downloading updated beatmap (because we send a `beatmap_id` and receive back a `checksum`).

### DIFF
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -215,7 +215,7 @@ namespace osu.Game.Beatmaps
 
             foreach (var info in item.Beatmaps)
             {
-                var file = item.Files.FirstOrDefault(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase))?.FileInfo.StoragePath;
+                var file = item.Files.SingleOrDefault(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase))?.FileInfo.StoragePath;
 
                 using (var stream = Files.Store.GetStream(file))
                     info.MD5Hash = stream.ComputeMD5Hash();
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -201,7 +201,9 @@ namespace osu.Game.Beatmaps
                 using (var sw = new StreamWriter(stream, Encoding.UTF8, 1024, true))
                     new LegacyBeatmapEncoder(beatmapContent).Encode(sw);
 
-                stream.Seek(0, SeekOrigin.Begin);
+                var attachedInfo = setInfo.Beatmaps.Single(b => b.ID == info.ID);
+                var md5Hash = stream.ComputeMD5Hash();
+                attachedInfo.MD5Hash = md5Hash;
 
                 UpdateFile(setInfo, setInfo.Files.Single(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase)), stream);
             }
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -201,6 +201,8 @@ namespace osu.Game.Beatmaps
                 using (var sw = new StreamWriter(stream, Encoding.UTF8, 1024, true))
                     new LegacyBeatmapEncoder(beatmapContent).Encode(sw);
 
+                stream.Seek(0, SeekOrigin.Begin);
+
                 UpdateFile(setInfo, setInfo.Files.Single(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase)), stream);
             }
 
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -201,16 +201,25 @@ namespace osu.Game.Beatmaps
                 using (var sw = new StreamWriter(stream, Encoding.UTF8, 1024, true))
                     new LegacyBeatmapEncoder(beatmapContent).Encode(sw);
 
-                var attachedInfo = setInfo.Beatmaps.Single(b => b.ID == info.ID);
-                var md5Hash = stream.ComputeMD5Hash();
-                attachedInfo.MD5Hash = md5Hash;
-
                 UpdateFile(setInfo, setInfo.Files.Single(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase)), stream);
             }
 
             var working = workingCache.FirstOrDefault(w => w.BeatmapInfo?.ID == info.ID);
             if (working != null)
                 workingCache.Remove(working);
+        }
+
+        protected override void PreUpdate(BeatmapSetInfo item)
+        {
+            base.PreUpdate(item);
+
+            foreach (var info in item.Beatmaps)
+            {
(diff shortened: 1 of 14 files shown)


osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -211,6 +209,19 @@ public void Save(BeatmapInfo info, IBeatmap beatmapContent)
                 workingCache.Remove(working);
         }
 
+        protected override void PreUpdate(BeatmapSetInfo item)
+        {
+            base.PreUpdate(item);
+
+            foreach (var info in item.Beatmaps)
+            {
+                var file = item.Files.FirstOrDefault(f => string.Equals(f.Filename, info.Path, StringComparison.OrdinalIgnoreCase))?.FileInfo.StoragePath;
```
bdach, line 218: Could this be a `SingleOrDefault`?

osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
```diff
@@ -223,11 +229,26 @@ protected override void LoadComplete()
...
             }
 
             private void stateChanged(ValueChangedEvent<DownloadState> state)
             {
-                this.FadeTo(state.NewValue == DownloadState.LocallyAvailable ? 0 : 1, 500);
+                switch (state.NewValue)
+                {
+                    case DownloadState.LocallyAvailable:
+                        // Perform a local query of the beatmap by beatmap checksum, and reset the state if not matching.
+                        if (beatmapManager.QueryBeatmap(b => b.MD5Hash == playlistItem.Beatmap.Value.MD5Hash) == null)
```
bdach, line 241: As an aside - this is way out of scope here and also a separate consideration for web, but in the long *long* run we might want a more reliable hash algorithm here - md5 is broken and I'm told collisions are easily manufactured. (Whether a colliding file can pass as a valid `.osu` file - I don't know)

bdach (commented):
If the locally-modified beatmap has scores imported for it, the import process will fail trying to overwrite it with the downloaded version due to violating foreign key constraints when trying to delete it.

Repro steps are same as in test, but after making sure the map modified has at least one imported score.

Not sure it can be easily fixed in this PR. Other than that I can't find much fault with this code.

peppy:
For the scores, we may want to just delete the old ones on modification for the time being.

osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
peppy: Definitely. I believe this could already be done quite easily. Not sure when we'll get to doing this (probably quite low priority until we notice people doing it).

smoogipoo:
Given that the re-import thing is a completely separate issue, I don't want to fix it in this PR.

I also don't know the best way to fix it in the first place, since it seems the cause in this case is that the `ItemRemoved` event is being delayed and `OsuGameBase` which would otherwise delete scores in the exact case mentioned, isn't deleting the scores.

Would prefer @peppy to look into this one instead, since he probably knows best how to implement this with the current database structure and something that will if we still consider Realm going forward.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -211,6 +209,19 @@ public void Save(BeatmapInfo info, IBeatmap beatmapContent)
                 workingCache.Remove(working);
         }
 
+        protected override void PreUpdate(BeatmapSetInfo item)
+        {
+            base.PreUpdate(item);
+
+            foreach (var info in item.Beatmaps)
```
peppy, line 216: Is there a reason this couldn't just be done in the `Save()` method? Not saying that we necessarily should be, but just curious if there is another (non-save) case that requires it to be here.

peppy (changes requested):
a

osu.Game/Beatmaps/BeatmapManager.cs
smoogipoo: I was thinking it could be used for updating of beatmaps, but in thinking about it all of that is going to go through `Save()` anyway. I've removed it for now.

peppy (dismissed):

peppy:
Directly executing Save will fix the test, please confirm this is correct direction:

```csharp
        [Test]
        public async Task TestUpdateBeatmapFile()
        {
            using (HeadlessGameHost host = new CleanRunHeadlessGameHost(nameof(TestUpdateBeatmapFile)))
            {
                try
                {
                    var osu = loadOsu(host);
                    var manager = osu.Dependencies.Get<BeatmapManager>();

                    var temp = TestResources.GetTestBeatmapForImport();
                    await osu.Dependencies.Get<BeatmapManager>().Import(temp);

                    BeatmapSetInfo setToUpdate = manager.GetAllUsableBeatmapSets()[0];

                    var beatmapInfo = setToUpdate.Beatmaps.First(b => b.RulesetID == 0);
                    Beatmap beatmapToUpdate = (Beatmap)manager.GetWorkingBeatmap(setToUpdate.Beatmaps.First(b => b.RulesetID == 0)).Beatmap;
                    BeatmapSetFileInfo fileToUpdate = setToUpdate.Files.First(f => beatmapToUpdate.BeatmapInfo.Path.Contains(f.Filename));

                    string oldMd5Hash = beatmapToUpdate.BeatmapInfo.MD5Hash;

                    beatmapToUpdate.HitObjects.Clear();
                    beatmapToUpdate.HitObjects.Add(new HitCircle { StartTime = 5000 });

                    manager.Save(beatmapInfo, beatmapToUpdate);

                    // Check that the old file reference has been removed
                    Assert.That(manager.QueryBeatmapSet(s => s.ID == setToUpdate.ID).Files.All(f => f.ID != fileToUpdate.ID));

                    // Check that the new file is referenced correctly by attempting a retrieval
                    Beatmap updatedBeatmap = (Beatmap)manager.GetWorkingBeatmap(manager.QueryBeatmap(b => b.ID == beatmapToUpdate.BeatmapInfo.ID)).Beatmap;
                    Assert.That(updatedBeatmap.HitObjects.Count, Is.EqualTo(1));
                    Assert.That(updatedBeatmap.HitObjects[0].StartTime, Is.EqualTo(5000));
                    Assert.That(updatedBeatmap.BeatmapInfo.MD5Hash, Is.Not.EqualTo(oldMd5Hash));
                }
                finally
                {
                    host.Exit();
                }
            }
        }
```

peppy (approved):

PR #9182, opened 2020-06-02
Fix cursor trail not displaying

### LINKED ISSUE #9179: No cursor trail
**Describe the bug:** In the latest update, the cursor trail is missing, whether you enable it or even disable it. Happens with the standard lazer skin and with others

**Screenshots or videos showing encountered issue:** https://streamable.com/f1sjvo

**osu!lazer version:** 2020.602.0

**Logs:**
[network.log](https://github.com/ppy/osu/files/4717754/network.log)
[performance.log](https://github.com/ppy/osu/files/4717755/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4717756/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4717757/updater.log)
[database.log](https://github.com/ppy/osu/files/4717758/database.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9179

Really want to have auto-generated shaders someday :(

The actual setting of the texture rectangles isn't required, but /shrug. The fix is adding the attribute in the first place, which pads it enough for the `Time` attribute.


bdach (approved):
fixes it on my machine - double-checked to be sure 👍

PR #9186, opened 2020-06-02
Add cursor rotation on right mouse drag

### LINKED ISSUE #9144: Cursor does not rotate on right mouse button drag
When holding down right-click, the cursor does not rotate.

kloczewiak:
Closes #9144


bdach:
Thanks for the contribution, but for future reference:

* Please do not PR off `master` - create a topic branch instead.
* Additionally, please do not edit code using the github web interface. It causes reviewers to have very low confidence in the correctness of any changes made in the PR.

I have not looked properly at the code or ran it, but I can already tell it will fail CI due to incorrect whitespace, which you would have most likely caught if you were using an actual IDE.

power9maker:
> * Please do not PR off `master` - create a topic branch instead.

Ok, will do that next time.

> I have not looked properly at the code or ran it, but I can already tell it will fail CI due to incorrect whitespace, which you would have most likely caught if you were using an actual IDE.

Well, I use Visual Studio (not Code), is it supposed to highlight incorrect whitespaces by default?

bdach:
VS Code is the least helpful of the recommended IDEs when it comes to formatting, but you can use the `InspectCode.ps1` script in the root of this repository in its stead.

power9maker:
I meant I DON'T use VS Code. I use VS Community 2019. Is there somewhere a setting for highlighting incorrect whitespaces? It would have been better than running this script for 20+ minutes.

bdach:
Misread that, sorry.

InspectCode can sometimes take a while, but 20+ minutes sounds wrong. I don't think I've ever had it take more than 5 at most.

ReSharper would highlight everything for sure, but it is generally a paid option. I'm pretty sure all Roslyn analyser errors should be displayed by VS, but that's not everything we use for code style. There are some R#-specific inspections that leverage solution-wide analysis and won't be picked up by VS right now.

If you really would like to avoid running the script locally, you could look into adding appveyor to your fork.

jorolf:
There's also a whitespace plugin for visual studio

power9maker:
> InspectCode can sometimes take a while, but 20+ minutes sounds wrong. I don't think I've ever had it take more than 5 at most.

Oh, sorry then. CI took about 20 minutes so I thought this would be about the same.

peppy:
These conditionals read really badly to me, I can no longer understand what they are trying to convey. Why not create a private property called something like `shouldRotate` which takes in a `MouseEvent` and does the checking for pressed buttons. Or just drop the conditionals and make it rotate on any button.

power9maker:
```
private bool shouldRotate;

private MouseEvent ShouldRotate
{
        get => ;
        set
        {
            if (value.IsPressed(MouseButton.Left) || value.IsPressed(MouseButton.Right))
                shouldRotate = true;
            else
                shouldRotate = false;
        }
}
```
Good idea but I would like it to output a bool which it can't because it's a MouseEvent. And also `ShouldRotate = e` would have to be called every `OnMouseDown()` and `OnMouseUp()` which doesn't look clean.
I could just make a function that will have MouseEvent as an argument and it will return a bool, which will be saying whether or not cursor should be in rotating state. Is that okay?

bdach:
Yeah, I think a private (static) method was the intended suggestion here. Adding state like that is definitely not a good idea.

osu.Game/Graphics/Cursor/MenuCursor.cs
```diff
@@ -97,13 +98,14 @@ protected override bool OnMouseDown(MouseDownEvent e)
 
         protected override void OnMouseUp(MouseUpEvent e)
         {
-            if (!e.IsPressed(MouseButton.Left) && !e.IsPressed(MouseButton.Middle) && !e.IsPressed(MouseButton.Right))
+            // cursor should go back to original size when none of main buttons are pressed
+            if (!(e.IsPressed(MouseButton.Left) || e.IsPressed(MouseButton.Middle) || e.IsPressed(MouseButton.Right)))
```
bdach, line 102: I'd factor this out to a static method too (`anyMainButtonPressed(MouseEvent e)`). Can remove the comment at that point as the code will become self-documenting.

osu.Game/Graphics/Cursor/MenuCursor.cs
```diff
@@ -125,6 +127,14 @@ protected override void PopOut()
             activeCursor.ScaleTo(0.6f, 250, Easing.In);
         }
 
+        private static bool shouldRotate(MouseEvent e)
+        {
+            if (e.IsPressed(MouseButton.Left) || e.IsPressed(MouseButton.Right))
+                return true;
+            else
+                return false;
+        }
```
bdach, line 136: No need to inspect the bool to then return its exact value explicitly anyway:

```suggestion
        private static bool shouldRotate(MouseEvent e) => e.IsPressed(MouseButton.Left) || e.IsPressed(MouseButton.Right);
```

(maybe rename to `shouldRotateCursor` too for a tad more readability too?)

bdach (dismissed):

osu.Game/Graphics/Cursor/MenuCursor.cs
power9maker: I didn't know you could do that. Cool!

bdach:
@power9maker Please try to not push individual commits, as it incurs a significant load on CI and queues many builds.

power9maker:
> @power9maker Please try to not push individual commits, as it incurs a significant load on CI and queues many builds.

I was intending to do that but forgot these things, will be more careful next time.

power9maker:
This will be the last one I swear

bdach (commented):
I think this is fine...? Not too much of a fan of that nested conditional but other options (inlining as third condition, early-returning `base.OnMouseDown`, extracting the inner `if` to a method) seem just as bad to me...

power9maker:
I can move `cursorRotate.Value` check into `shouldRotateCursor()` method.
Just tested it and works as expected.

bdach:
Well, yes, but I'm not sure it reads much better than the current version as three conditions in an `if` is already a little much. Different strokes for different folks I suppose - I'll defer to the team's opinion on that one.

power9maker:
Hey, is this pr going to be merged or do I close it, because I have to reset my fork if I want work on something else.

osu.Game/Graphics/Cursor/MenuCursor.cs
```diff
@@ -83,24 +83,28 @@ protected override bool OnMouseDown(MouseDownEvent e)
...
         {
-            if (!e.IsPressed(MouseButton.Left) && !e.IsPressed(MouseButton.Right))
+            if (!anyMainButtonPressed(e))
             {
                 activeCursor.AdditiveLayer.FadeOutFromOne(500, Easing.OutQuint);
                 activeCursor.ScaleTo(1, 500, Easing.OutElastic);
             }
 
-            if (e.Button == MouseButton.Left)
+            if (!shouldRotateCursor(e))
```
peppy, line 107: isn't this back to front?

osu.Game/Graphics/Cursor/MenuCursor.cs
```diff
@@ -122,6 +126,10 @@ protected override void PopOut()
             activeCursor.ScaleTo(0.6f, 250, Easing.In);
         }
 
+        private static bool shouldRotateCursor(MouseEvent e) => e.IsPressed(MouseButton.Left) || e.IsPressed(MouseButton.Right);
```
peppy, line 129: `cursorRotate.Value check` should be moved here.

peppy (dismissed):

osu.Game/Graphics/Cursor/MenuCursor.cs
bdach: Wow, it is. Garbage reading on my part. I would have sworn I tested this but at this point I don't trust my brain on that.

power9maker: > isn't this back to front?

What does back to front mean?

bdach: The calling pattern does not match the expected logic. The code is first asking if it should rotate the cursor, but then rotates it if the method returns `false` (so if it *shouldn't* rotate it).

power9maker: Either I still don't understand what you mean or you misread the code:
shouldRotateCursor() checks whether or not cursor should be in its dragged state.
On line 107 it checks if it is not in its dragged state so it can do the spinning and rotate back to 0 degrees.

bdach: Okay, yeah, but it's confusing reading as it is anyway. Maybe rename the method to `shouldKeepRotating`?

peppy: even if this does work, it's actually opposite to the original logic, which means further checking is necessary.

bdach: I think the extra layer of confusion here is because the previous code used the mouse up event directly and the new code uses `IsPressed`.

power9maker: I used `IsPressed` here, because if I did it with `MouseUpEvent.Button` and you let go of one mouse button, it would stop rotating even though you have the other mouse button held down.

peppy:
Okay, so from a correctness angle, why does only middle mouse now behave differently? Originally it was left-only intentionally, but if adding right mouse to the equation, why not also middle?

power9maker:
Because that's what was requested in #9144, but yeah, sure, I'll add it to middle mouse button too.

peppy:
I've simplified this further by including all buttons and using the already available property of mouse events to check as such. @smoogipoo requesting a quick re-review.

smoogipoo (approved):
Yeah this seems fine to me

PR #9190, opened 2020-06-03
Fix results screen crashing for beatmaps with no online ID

swoolcock:
Likely regressed in #9086 

Logic taken from `ScoresRequest.getScores()`

### DIFF
--- a/osu.Game.Tests/Visual/Ranking/TestSceneResultsScreen.cs
+++ b/osu.Game.Tests/Visual/Ranking/TestSceneResultsScreen.cs
@@ -36,12 +36,14 @@ namespace osu.Game.Tests.Visual.Ranking
                 Beatmap.Value = beatmaps.GetWorkingBeatmap(beatmapInfo);
         }
 
-        private TestSoloResults createResultsScreen() => new TestSoloResults(new TestScoreInfo(new OsuRuleset().RulesetInfo));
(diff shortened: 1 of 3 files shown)


peppy:
What's the crash this is fixing specifically? Not sure this is the correct fix. Should probably be in `SoloResultsScreen` at very least.

Test coverage would be appreciated if possible.

swoolcock:
`GetScoresRequest` constructor:
`throw new InvalidOperationException($"Cannot lookup a beatmap's scores without having a populated {nameof(BeatmapInfo.OnlineBeatmapID)}.");`

Found while testing my ruleset on an unsubmitted map.

smoogipoo:
This should be fixed in `SoloResultsScreen` as mentioned.

swoolcock:
> This should be fixed in `SoloResultsScreen` as mentioned.

No problem, will do.  I was going to put it there originally but I figured there would be code duplication if future screens needed to implement `FetchScores` and do the same check.

peppy:
Consider a case such as local score display, where no online fetching or online id requirement is present.

smoogipoo (approved):
This is fine for now imo. We'll probably want to display the local leaderboard but that can be saved for a future PR, and fix this crash here for now.

peppy:
Should be fine to merge, CI failures will be fixed separately.

PR #9191, opened 2020-06-03
Attempt to create a hard link when importing beatmap from a non-archive source

peppy:
Not sure where to put the pinvoke function (it seems to be the first of its kind in the osu-side project... which is probably a bad precedent to set). Open to advice, just PRing for testing / feedback purposes.

We may want to make a helper function (framework? osu?) which also handles other platforms where hard links are supported.

Reduces IO (write) overhead during import from stable to near zero. Also disk space.

### DIFF
--- a/osu.Game/IO/FileStore.cs
+++ b/osu.Game/IO/FileStore.cs
@@ -6,7 +6,9 @@ using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Runtime.InteropServices;
 using Microsoft.EntityFrameworkCore;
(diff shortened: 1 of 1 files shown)


peppy:
Due to the nature of how hard links work when modifications are made (will affect all copies), let's not support this for the time being. It has the potential to leave lazer's file store in a bad state due to user changes in stable.

To have this work we would want to:

- Delete the old files once we had finished migrating (will be useful once we are ready to have users transfer to lazer permanently).
or
- Add checksum verification on file reads (at very least for cases where it matters, such as ini/osu files).

Will close for now.

PR #9193, opened 2020-06-03
Update inspectcode / CodeFileSanity versions used in CI

peppy:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3600

### DIFF
--- a/osu.Desktop/OsuGameDesktop.cs
+++ b/osu.Desktop/OsuGameDesktop.cs
@@ -59,7 +59,7 @@ namespace osu.Desktop
             try
             {
                 using (RegistryKey key = Registry.ClassesRoot.OpenSubKey("osu"))
-                    stableInstallPath = key?.OpenSubKey(@"shell\open\command")?.GetValue(string.Empty).ToString().Split('"')[1].Replace("osu!.exe", "");
+                    stableInstallPath = key?.OpenSubKey(@"shell\open\command")?.GetValue(string.Empty).ToString()?.Split('"')[1].Replace("osu!.exe", "");
 
                 if (checkExists(stableInstallPath))
                     return stableInstallPath;
--- a/osu.Game.Tests/Chat/MessageFormatterTests.cs
+++ b/osu.Game.Tests/Chat/MessageFormatterTests.cs
@@ -428,23 +428,28 @@ namespace osu.Game.Tests.Chat
             Assert.AreEqual(5, result.Links.Count);
 
             Link f = result.Links.Find(l => l.Url == "https://osu.ppy.sh/wiki/wiki links");
-            Assert.AreEqual(44, f!.Index);
+            Assert.That(f, Is.Not.Null);
+            Assert.AreEqual(44, f.Index);
             Assert.AreEqual(10, f.Length);
 
             f = result.Links.Find(l => l.Url == "http://www.simple-test.com");
-            Assert.AreEqual(10, f!.Index);
+            Assert.That(f, Is.Not.Null);
+            Assert.AreEqual(10, f.Index);
             Assert.AreEqual(11, f.Length);
 
             f = result.Links.Find(l => l.Url == "http://google.com");
-            Assert.AreEqual(97, f!.Index);
+            Assert.That(f, Is.Not.Null);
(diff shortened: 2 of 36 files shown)


smoogipoo:
Only moved a few suggestions to warnings, not all of them. Unsure on whether some of them should be warnings at all rather than hints/none - probably need to go through them exhaustively at some point.

osu.Desktop/OsuGameDesktop.cs
```diff
@@ -59,7 +59,7 @@ private string getStableInstallPath()
             try
             {
                 using (RegistryKey key = Registry.ClassesRoot.OpenSubKey("osu"))
-                    stableInstallPath = key?.OpenSubKey(@"shell\open\command")?.GetValue(string.Empty).ToString().Split('"')[1].Replace("osu!.exe", "");
+                    stableInstallPath = key?.OpenSubKey(@"shell\open\command")?.GetValue(string.Empty).ToString()?.Split('"')[1].Replace("osu!.exe", "");
```
huoyaoyuan, line 62: This nullable warning is quite unprecise.
`GetValue` declares to return nullable, but no warning (I don't really expect it because we haven't fully opt-in NRT).

Here, `object.ToString()` is declared nullable, but seldom nullable in override in practice. I reported an issue for them always treating `ToString()` non-nullable, but they look to aggressive now.

osu.Game.Tests/Chat/MessageFormatterTests.cs
```diff
@@ -428,23 +428,23 @@ public void TestLinkComplex()
             Assert.AreEqual(5, result.Links.Count);
 
             Link f = result.Links.Find(l => l.Url == "https://osu.ppy.sh/wiki/wiki links");
-            Assert.AreEqual(44, f.Index);
+            Assert.AreEqual(44, f!.Index);
```
huoyaoyuan, line 431: We haven't enable NRT, so the `!` means nothing for the compiler. Asserting f isn't null should be better approach.

osu.Game.Tests/Chat/MessageFormatterTests.cs
```diff
@@ -428,23 +428,23 @@ public void TestLinkComplex()
...
             Assert.AreEqual(10, f.Length);
 
             f = result.Links.Find(l => l.Url == "http://www.simple-test.com");
-            Assert.AreEqual(10, f.Index);
+            Assert.AreEqual(10, f!.Index);
```
huoyaoyuan, line 435: [sharplab](https://sharplab.io/#v2:D4YgdgrgNlCGBGUCmACJYHILAChcAEAmARgIGYUiUBhXAb1xScovwBYUBZACn2IAYA/CgDOASkbMGOZrJQBLMABcFKALyiAhADoAMugDmSgBYBuSXNUBqDSL2GT5mcwC+uF0A===)

The compiler and the language say that, a first `!` will imply the variable being non-null in the following.

huoyaoyuan (commented):
I've had a negative bias to ReSharper(Rider), because it often provide too aggressive suggestions that are even wrong.

osu.Game.Tests/Chat/MessageFormatterTests.cs
smoogipoo: The f is not the same.

osu.Game.Tests/Chat/MessageFormatterTests.cs
smoogipoo: Resharper analysis != NRT, so it means something here. It means resharper won't complain and will let an exception occur if it's null. This won't ever be the case but since it's a test it's effectively the same as an assertion imo.

osu.Desktop/OsuGameDesktop.cs
smoogipoo: /shrug

osu.Game.Tests/Chat/MessageFormatterTests.cs
peppy: What does this mean? Are there any alternatives to `!`? Feels quite foreign.

PR #9195, opened 2020-06-03
Fix chat history not being loaded for multiplayer matches

### LINKED ISSUE #9181: Multiplayer room chat history doesn't show

peppy:
The lazy load logic for `fetchInitialMessages` was relying on the channel becoming the `ChannelManager`'s `CurrentChannel`, which it doesn't for external chat displays.

Closes https://github.com/ppy/osu/issues/9181. Also tidies things up a bit (not sure why there was an extra pathway/state for `alreadyJoined`.. nothing seems to break with this change).

### DIFF
--- a/osu.Game.Tests/Visual/Online/TestSceneChatOverlay.cs
+++ b/osu.Game.Tests/Visual/Online/TestSceneChatOverlay.cs
@@ -246,7 +246,12 @@ namespace osu.Game.Tests.Visual.Online
             {
                 ((BindableList<Channel>)ChannelManager.AvailableChannels).AddRange(channels);
 
-                Child = ChatOverlay = new TestChatOverlay { RelativeSizeAxes = Axes.Both, };
+                InternalChildren = new Drawable[]
+                {
+                    ChannelManager,
+                    ChatOverlay = new TestChatOverlay { RelativeSizeAxes = Axes.Both, },
+                };
+
                 ChatOverlay.Show();
             }
         }
--- a/osu.Game/Online/Chat/ChannelManager.cs
+++ b/osu.Game/Online/Chat/ChannelManager.cs
@@ -358,6 +358,13 @@ namespace osu.Game.Online.Chat
             {
                 switch (channel.Type)
                 {
+                    case ChannelType.Multiplayer:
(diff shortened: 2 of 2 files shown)


smoogipoo:
This still doesn't seem to be working correctly: https://drive.google.com/file/d/1uAlbiNPULK6M_IUPo4ueEdH8yGc5EMsg/view?usp=sharing Several things:
1. Tabs appear -> then disappear -> then re-appear (in what I presume is the final "joined" state after receiving the response from the server).
2. Multiplayer chat still doesn't load messages - go into `sebastianpc`'s room, observe that chat remains empty until you open up the tab in the chat overlay. Even when you do this the multiplayer chat itself doesn't get filled, only the chat overlay does.

(2) behaves worse than current master, in which at least the messages appear in the multiplayer chat when opening the tab in the chat overlay...

peppy:
Have you looked into why this is occurring for you? I already heavily tested this so am a bit baffled.

peppy:
Are you opening chat manually? I was testing using the multiplayer chat interface, not the ChatOverlay.

smoogipoo:
If you mean manually by pressing F8, yes.

peppy:
Have you tried without doing so, and checking the actual multiplayer room display? Does that work?

smoogipoo:
In the failing cases I've had this appear in my logs:

```
[network] 2020-06-03 12:09:01 [verbose]: Request to https://osu.ppy.sh/api/v2/chat/channels/28096586/users/1040328 failed with System.Net.WebException: Forbidden.
[network] 2020-06-03 12:09:01 [verbose]: Response was: {"error":"You do not have access to that channel."}
[network] 2020-06-03 12:09:01 [verbose]: Failing request osu.Game.Online.API.Requests.JoinChannelRequest (osu.Game.Online.API.APIException: You do not have access to that channel.
[network] 2020-06-03 12:09:01 [verbose]: ---> System.Net.WebException: Forbidden
[network] 2020-06-03 12:09:01 [verbose]: --- End of inner exception stack trace ---)

```

peppy:
That sounds pretty weird. This is on your normal user account? Do you see that on master at all?

smoogipoo (approved):

PR #9197, opened 2020-06-03
Check all files match before using an existing archive model

peppy:
Fixes people trying to reimport skins or beatmaps after making changes and not succeeding.


peppy:
Blocking temporarily. One test is failing because it never actually worked correctly; another is failing due to the nature of how `FileStore` works (checking timestamp doesn't work in some import scenarios), and requires a bit more thinking.

peppy:
Modified times wasn't really usable due to the fact that we... import all files to the `FileStore` before checking for existing. Which is really inefficient and probably not what we want going forward, but it made implementing this new logic simple (just compare `FileInfo` `ID`s and `ModelFileInfo` `Filename`s for safety).

smoogipoo (approved):

PR #9210, opened 2020-06-04
Fix tournament displayed beatmap potentially being out of order on quick changes

### LINKED ISSUE #9184: Beatmap shown in tournament client desyncs with cuttingedge
**Describe the bug:** Hi, I meet a problem when I use the tournament end of lazer to make live broadcast of the game. However, there are only two ways to trigger this error.

As you can see, the lower right corner of the lazer client wrongly displays a message of the beatmap when I quickly switch beatmap in the video.

As for another triggering method, it is that the stable client end will automatically change to the next beatmap if the player has not started the game after a beatmap is played. Then, the lazer client end will synchronize the beatmap of the stable client end. At this time, the lazer client end will also display the wrong beatmap information if the player starts the game.

I hope that the tournament client end of lazer can synchronize the songs played in the stable client end again before starting the game, so as to correct the displayed beatmap information.

Finally, thank you very much for seeing what I write here.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=XwXZJ0IjTAs

**osu!lazer version:** 2020.523.0

**Logs:** N/A

peppy:
Should close https://github.com/ppy/osu/issues/9184.

### DIFF
--- a/osu.Game.Tournament/IPC/FileBasedIPC.cs
+++ b/osu.Game.Tournament/IPC/FileBasedIPC.cs
@@ -34,6 +34,7 @@ namespace osu.Game.Tournament.IPC
 
         private int lastBeatmapId;
         private ScheduledDelegate scheduled;
+        private GetBeatmapRequest beatmapLookupRequest;
 
         public Storage Storage { get; private set; }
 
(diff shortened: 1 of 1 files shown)


bdach (approved):
Yeah, I can definitely see that such a write race could cause the linked issue 👍 New version definitely more correct.

smoogipoo (approved):

PR #9211, opened 2020-06-04
Change format of date on score panel

### LINKED ISSUE #9202: Play date in result screen always in MM/DD/YYYY format
**Describe the bug:** self-explanatory, might want a setting to change format

image
This score was set on 11th January, 2018

**osu!lazer version:** 2020.602.0

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9202

Matches the osu!-web format.


peppy (approved):
we may want to consider using the local regional settings in the future but i guess this is an improvement

PR #9215, opened 2020-06-04
Fix incorrect sample choices for hold notes

### LINKED ISSUE #2506: Mania longnote releases trigger hitsounds

bdach:
Resolves #2506 (and another issue I found along the way).

# Summary

Overall handling of `NodeSamples` was generally not consistent with stable. There are two cases to consider:

* As reported in the aforementioned issue, on stable mania-specific beatmaps only play their samples at the start on the hold note and have silent ends.
  This has been resolved in the way I proposed in the issue.

* Along the way I also noticed that converts were also incorrect, in that they always used the unmodified `NodeSamples` list.
  That is actually not correct when tiled hold notes are generated, as they should receive a sliced node sample list, starting from the node sample that corresponds to the start time of the hold note, since the hold note itself will always choose the first node sample for its head:

  https://github.com/ppy/osu/blob/a8b137bb715db0c370148f04a548e7db6cc3cc9c/osu.Game.Rulesets.Mania/Objects/HoldNote.cs#L101-L106

  This is resolved by extracting out a method that slices the list, and then retouching `sampleInfoListAt` to use it.

The PR also includes conversion test cases. For demonstration and comparison of behaviour, see the following videos:
| | | | |
| :-: | :-: | :-: | :-: |
| `mania-samples` | [stable](https://streamable.com/v3dfl4) | [`master`](https://streamable.com/qhzfvf) | [this PR](https://streamable.com/ykgivp) |
| `convert-samples` | [stable](https://streamable.com/yziujg) | [`master`](https://streamable.com/giud7z) | [this PR](https://streamable.com/yu7juv) |

# Remarks

The fact that I didn't use `ManiaBeatmapConversionTest` itself might raise an eyebrow - the reasoning was that (a) I think it's a little better to not conflate the two and keep those test cases as isolated as possible, and (b) I was sort of put off by the `RngSnapshot` magic in there as I don't know how it works or how to generate the values of the four variables within. If it's an issue I can merge the new test class back into it.

Another potential issue is that the sliced node sample list doesn't have its end cut off, which could cause an analogous problem if the conversion algorithm could create a hold note that is "in the middle of a slider" (that is, that can end at one of the repeat points of the slider and not its actual end time). As far as I can see skimming through the various pattern types that seems impossible, so I refrained from slicing the end for simplicity (and I want to avoid creating list copies as much as possible), but I'm not 100% sure on that.

### DIFF
--- a/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
+++ b/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
@@ -478,7 +478,7 @@ namespace osu.Game.Rulesets.Mania.Beatmaps.Patterns.Legacy
         /// Retrieves the list of node samples that occur at time greater than or equal to <paramref name="time"/>.
         /// </summary>
         /// <param name="time">The time to retrieve node samples at.</param>
-        private IEnumerable<IList<HitSampleInfo>> nodeSamplesAt(double time)
+        private List<IList<HitSampleInfo>> nodeSamplesAt(double time)
         {
             if (!(HitObject is IHasPathWithRepeats curveData))
                 return null;
@@ -486,7 +486,9 @@ namespace osu.Game.Rulesets.Mania.Beatmaps.Patterns.Legacy
             double segmentTime = (EndTime - HitObject.StartTime) / spanCount;
 
             int index = (int)(segmentTime == 0 ? 0 : (time - HitObject.StartTime) / segmentTime);
-            return curveData.NodeSamples.Skip(index);
+
+            // avoid slicing the list & creating copies, if at all possible.
+            return index == 0 ? curveData.NodeSamples : curveData.NodeSamples.Skip(index).ToList();
         }
 
         /// <summary>
@@ -517,7 +519,7 @@ namespace osu.Game.Rulesets.Mania.Beatmaps.Patterns.Legacy
                     Duration = endTime - startTime,
                     Column = column,
                     Samples = HitObject.Samples,
-                    NodeSamples = nodeSamplesAt(startTime)?.ToList()
+                    NodeSamples = nodeSamplesAt(startTime)
                 };
             }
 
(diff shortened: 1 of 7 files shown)


peppy (approved):

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
```diff
@@ -511,7 +519,7 @@ private void addToPattern(Pattern pattern, int column, double startTime, double
                     Duration = endTime - startTime,
                     Column = column,
                     Samples = HitObject.Samples,
-                    NodeSamples = (HitObject as IHasRepeats)?.NodeSamples
+                    NodeSamples = nodeSamplesAt(startTime)
```
smoogipoo, line 522: Are you sure about this? The reason I didn't merge this PR myself is that I don't think this is actually the case for converted beatmaps.

peppy: I didn't check stable code, just a couple of hitsounded beatmaps (that looked to match). May be worth doing if you're unsure.

PR #9224, opened 2020-06-06
Add cinema mod selection shortcut to song select

### LINKED ISSUE #9221: Cinema mod shortcut not implemented
**Describe the feature:** Shortcut is <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Enter</kbd>. Right now it just enables autoplay (because of <kbd>Ctrl</kbd>+<kbd>Enter</kbd>) in song select.

Game4all:
# Summary

Closes #9221

- Adds the cinema mod selection shortcut to song select <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Enter</kbd>
- A notification similar to what was implemented in #9168 will be displayed if the user attempts to start a beatmap using a ruleset which hasn't an available cinema mod.

## Testing
~~Again no automatic testing as it is quite hard to automate.
However, manual testing shows it works as expected.~~

See `TestScenePlaySongSelect` test scene.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -31,6 +31,7 @@ using osu.Game.Screens.Select.Carousel;
 using osu.Game.Screens.Select.Filter;
 using osu.Game.Users;
 using osuTK.Input;
+using osu.Game.Overlays.Mods;
 
 namespace osu.Game.Tests.Visual.SongSelect
 {
@@ -589,6 +590,65 @@ namespace osu.Game.Tests.Visual.SongSelect
         }
 
         [Test]
+        public void TestCinemaViaCtrlShiftEnter()
+        {
+            addRulesetImportStep(0);
+
+            createSongSelect();
+
+            AddStep("press ctrl+shit+enter", () =>
+            {
+                InputManager.PressKey(Key.ControlLeft);
+                InputManager.PressKey(Key.ShiftLeft);
+                InputManager.PressKey(Key.Enter);
+
+                InputManager.ReleaseKey(Key.ControlLeft);
+                InputManager.ReleaseKey(Key.ShiftLeft);
+                InputManager.ReleaseKey(Key.Enter);
+            });
+
+            AddUntilStep("wait for player", () => Stack.CurrentScreen is PlayerLoader);
+
+            AddAssert("cinema enabled", () => songSelect.Mods.Value.FirstOrDefault() is ModCinema);
+
+            AddUntilStep("wait for return to ss", () => songSelect.IsCurrentScreen());
+
+            AddAssert("mod disabled", () => songSelect.Mods.Value.Count == 0);
+        }
+
+        [Test]
+        public void TestAutoplayWithCtrlShiftEnter()
+        {
+            addRulesetImportStep(0);
+
+            createSongSelect();
+
+            AddStep("select autoplay", () =>
+            {
+                songSelect.ModSelect.SelectTypes(new[] { typeof(ModAutoplay) });
+            });
+
+            AddStep("press ctrl+shit+enter", () =>
+            {
+                InputManager.PressKey(Key.ControlLeft);
+                InputManager.PressKey(Key.ShiftLeft);
+                InputManager.PressKey(Key.Enter);
+
+                InputManager.ReleaseKey(Key.ControlLeft);
+                InputManager.ReleaseKey(Key.ShiftLeft);
+                InputManager.ReleaseKey(Key.Enter);
+            });
+
+            AddUntilStep("wait for player", () => Stack.CurrentScreen is PlayerLoader);
+
+            AddAssert("only cinema enabled", () => songSelect.Mods.Value.FirstOrDefault() is ModCinema && songSelect.Mods.Value.Count == 1);
+
+            AddUntilStep("wait for return to ss", () => songSelect.IsCurrentScreen());
+
+            AddAssert("mod disabled", () => songSelect.Mods.Value.Count == 0);
+        }
+
+        [Test]
         public void TestHideSetSelectsCorrectBeatmap()
         {
             int? previousID = null;
@@ -929,6 +989,8 @@ namespace osu.Game.Tests.Visual.SongSelect
             public WorkingBeatmap CurrentBeatmap => Beatmap.Value;
             public WorkingBeatmap CurrentBeatmapDetailsBeatmap => BeatmapDetails.Beatmap;
             public new BeatmapCarousel Carousel => base.Carousel;
+
+            public new ModSelectOverlay ModSelect => base.ModSelect;
 
             public new void PresentScore(ScoreInfo score) => base.PresentScore(score);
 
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -82,28 +88,48 @@ protected override bool OnStart()
...
+                    modType = mod?.GetType();
+
+                    if (modType == null)
+                    {
+                        notifications?.Post(new SimpleNotification
+                        {
+                            Text = "The current ruleset doesn't have an autoplay mod avalaible!"
+                        });
+                        return false;
+                    }
```
bdach, line 126: This begs to be factored out somehow. Something like this maybe?
```csharp
private bool trySelectMod(Func<Ruleset> selectFunc, string notificationErrorMessage);
```

osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -82,28 +88,48 @@ protected override bool OnStart()
...
                 }
+
+                if (mods.All(m => m.GetType() != modType))
+                    Mods.Value = mods.Append(mod).ToArray();
+
+                removeAutoModOnResume = true;
```
bdach, line 132: This doesn't only remove the auto mod on resume any more. Variable rename is warranted here if this is to stay this way.

osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -55,10 +57,14 @@ public override void OnResuming(IScreen last)
             if (removeAutoModOnResume)
             {
                 var autoType = Ruleset.Value.CreateInstance().GetAutoplayMod()?.GetType();
+                var cinemaType = Ruleset.Value.CreateInstance().GetAllMods().OfType<ModCinema>().FirstOrDefault()?.GetType();
 
                 if (autoType != null)
                     ModSelect.DeselectTypes(new[] { autoType }, true);
 
+                if (cinemaType != null)
+                    ModSelect.DeselectTypes(new[] { cinemaType }, true);
```
bdach, line 66: I would be in favour of storing the selected mod instead of the `removeAutoModOnResume` flag, and then using that to remove it on resume.

bdach (changes requested):
This is not hard to test. See `TestScenePlaySongSelect.TestAutoplayViaCtrlEnter()` for precedent.

bdach:
This addition also exposes the following bug:

1. Manually select the autoplay mod via the mod overlay.
2. Press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Enter</kbd> to progress into gameplay.
3. Gameplay will start with *both* autoplay and cinema shown in the player loader, which should not be a valid combination in the first place.

I would like to see an automated test case for this.

peppy:
Also not sure we want this shortcut going forwards. Seems like a *very* rare use case with little benefit.

Game4all:
Well if usage is deemed too rare (compared to the auto shortcut) to be added back as a shortcut, then this can be closed.

osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
```diff
@@ -588,6 +589,65 @@ public void TestAutoplayViaCtrlEnter()
...
         }
 
+        [Test]
+        public void TestCinemaViaCtrlShiftEnter()
+        {
+            addRulesetImportStep(0);
+
+            createSongSelect();
+
+            AddStep("press ctrl+shit+enter", () =>
```
bdach, line 599: that's *quite* a typo

```suggestion
            AddStep("press ctrl+shift+enter", () =>
```

osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
```diff
@@ -588,6 +589,65 @@ public void TestAutoplayViaCtrlEnter()
...
+                InputManager.PressKey(Key.Enter);
+
+                InputManager.ReleaseKey(Key.ControlLeft);
+                InputManager.ReleaseKey(Key.ShiftLeft);
+                InputManager.ReleaseKey(Key.Enter);
+            });
+
+            AddUntilStep("wait for player", () => Stack.CurrentScreen is PlayerLoader);
+
+            AddAssert("only cinema enabled", () => songSelect.Mods.Value.FirstOrDefault() is ModCinema && songSelect.Mods.Value.Count == 1);
```
bdach, line 644: Could simplify by doing
```suggestion
            AddAssert("only cinema enabled", () => songSelect.Mods.Value.SingleOrDefault() is ModCinema);
```

osu.Game/Overlays/Mods/ModSection.cs
```diff
@@ -128,15 +128,16 @@ public void DeselectTypes(IEnumerable<Type> modTypes, bool immediate = false)
         /// Select one or more mods in this section and deselects all other ones.
         /// </summary>
         /// <param name="modTypes">The types of <see cref="Mod"/>s which should be selected.</param>
-        public void SelectTypes(IEnumerable<Type> modTypes)
+        /// <param name="deselect">Whether to deselect mods not present in <paramref name="modTypes"/></param>
+        public void SelectTypes(IEnumerable<Type> modTypes, bool deselect = true)
```
bdach, line 132: Having a `SelectTypes` method with a `deselect` parameter is quite dubious, especially given that its xmldoc shows the name is misleading. It should at worst be named `deselectOthers`, and at best not exist at all.

Have you looked at enforcing this via specifying `IncompatibleMods`?

osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
```diff
@@ -588,6 +589,65 @@ public void TestAutoplayViaCtrlEnter()
...
+        [Test]
+        public void TestAutoplayWithCtrlShiftEnter()
+        {
+            addRulesetImportStep(0);
+
+            createSongSelect();
+
+            AddStep("select autoplay", () =>
+            {
+                songSelect.ModSelect.SelectTypes(new[] { typeof(ModAutoplay) });
```
bdach, line 628: Exposition of `ModSelect` shouldn't be necessary - `SelectedMods` is exposed as protected in `OsuTestScene`.

bdach (changes requested):

osu.Game/Overlays/Mods/ModSection.cs
Game4all: Actually the method I came up with doesn't look fine anymore to me for the sole reason that I discovered it was working fine because it made the assumption that `ModAutoplay` and `ModCinema` were inside a MultiMod (that's the case for the 4 default rulesets, tested this by moving the `Autoplay` and `Cinema` mod implementations out of the MultiMod for the osu! ruleset). So i'll probably remove the whole method.

So I may now be more in favour of adding a check inside `trySelectMod` to ensure that the selected auto mod isn't incompatible with the already selected mods (by checking against `IncompatibleMods` property + a `IsSubclassOfCheck` (would love to hear your opinion on this).

I may give this a shot tomorrow but can't say for sure.

smoogipoo:
Given that the implementation required to make this work is quite chunky with little benefit, let's hold off on this for the time being.

PR #9230, opened 2020-06-08
Improve handling of missing intro beatmap / files

peppy:
Fixes [recurring reports](https://github.com/ppy/osu/issues?q=is%3Aissue+%22won%27t+start%22+is%3Aclosed) of lazer failing to start, usually due to user touching files they shouldn't.

- Fixes stall at intro screen if mp3 or beatmap files are missing (due to user deletion, usually)
- Fixes incorrect intro playback behaviour when user has disabled intro theme but also has no beatmaps
- Improves legibility of code, renamed "theme" specific lookups to separate from the case where the intro is using a user selection
- Fixes `WorkingBeatmap` cache not clearing on an underlying store change (would cause the implicit filestore reimport to not be used even on a second `GetWorkingBeatmap` call, because the `Track` was already resolved to a `TrackVirtual`)

Tests could be added for this with some degree of effort, but I also think this logic and process is going to change as we add more customisation going forward, so not sure how I feel about putting in the effort just yet. Manually tested all scenarios listed below:

- `rm client.db` -> imports from scratch
- `rm files/4/47/47b895484e7751f3ab429694ff6dbf21e774ab023e4f6c5b481476f04ff22f0f` (triangles mp3) -> uses existing but reimports missing files
- loading with menu theme disabled, with no beatmaps (both triangles and circles)
- loading with menu theme disabled, with beatmaps (both triangles and circles)

As an aside, the reason the `TrackVirtual` was stalling is that it has a length of 1000ms, which isn't enough for the intro to play out. Seems safest to avoid using it altogether rather than increase the length though.

### DIFF
--- a/osu.Game/Beatmaps/BeatmapManager.cs
+++ b/osu.Game/Beatmaps/BeatmapManager.cs
@@ -79,6 +79,8 @@ namespace osu.Game.Beatmaps
             beatmaps = (BeatmapStore)ModelStore;
             beatmaps.BeatmapHidden += b => beatmapHidden.Value = new WeakReference<BeatmapInfo>(b);
             beatmaps.BeatmapRestored += b => beatmapRestored.Value = new WeakReference<BeatmapInfo>(b);
+            beatmaps.ItemRemoved += removeWorkingCache;
+            beatmaps.ItemUpdated += removeWorkingCache;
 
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):
Seems sane

PR #9231, opened 2020-06-08
Fetch existing private message channels on re-joining

peppy:
- [x] Depends on https://github.com/ppy/osu-web/pull/6139.


smoogipoo:
I works but when re-joining a PM the initial messages are retrieved in the incorrect order:

image

Unsure whether this is an osu!web problem.

peppy:
Ah yep, that's osu-web side. @nanaya is looking into it.

smoogipoo (approved):

smoogipoo:
Have approved for merge, in-case there's no additional fixes required osu!side.

peppy:
https://github.com/ppy/osu-web/pull/6145

PR #9236, opened 2020-06-08
Move user and skin specific settings for LogoVisualisation to a subclass

MiraiSubject:
Applies this review comment: https://github.com/ppy/osu/pull/9173#discussion_r436869249

The naming of the newly made classes could probably be improved upon, because right now it's `BasicLogoVisualisation` for the main `LogoVisualisation` class. 

This is potentially a more cleaner approach to applying a custom accent color in https://github.com/ppy/osu/pull/9173 than what was originally included in the PR itself.


osu.Game/Screens/Menu/LogoVisualisation.cs
```diff
@@ -1,123 +1,31 @@
...
-using System;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Extensions.Color4Extensions;
-using osu.Framework.Utils;
 
 namespace osu.Game.Screens.Menu
 {
-    public class LogoVisualisation : Drawable, IHasAccentColour
+    public class LogoVisualisation : BasicLogoVisualisation
```
bdach, line 14: Yeah the naming isn't too fortunate. I'd probably rename:

* `BasicLogoVisualisation` -> `LogoVisualisation`
* `LogoVisualisation` -> `SkinnableLogoVisualisation`

bdach: Although now I think about it it might be a breaking change for rulesets... At least two or three custom rulesets use this class on their playfield.

peppy: The change you've proposed doesn't sound breaking because I don't think the skinned one should have been exposed in the first place (and the name still matches).

Also how about calling the skinned one `MenuLogoVisualisation` and making it internal while we're here?

Also please fully xmldoc the base class if possible.

MiraiSubject: Alright! Will do.

smoogipoo (approved):
Looks fine to me

PR #9241, opened 2020-06-09
Implement per-beatmap results screen for timeshift

smoogipoo:
Currently implemented as a button in the column that displays the playlist.


peppy (approved):

peppy:
There's one minor visual issue with the loading spinner but it's not unique to this new usage, so I'll PR that as a separate fix.

PR #9247, opened 2020-06-09
Move graphics from TournamentGameBase to TournamentGame

MiraiSubject:
Right now `TournamentGameBase` also adds all the user interface elements on startup. In this PR I have moved all the user interface related code inside `TournamentGameBase` to `TournamentGame` to match the relationship that `OsuGameBase` and `OsuGame` have.

Afaik `OsuGameBase` does not add any drawables to the screen, so this PR is supposed to achieve the same goal, but for the tournament side. 

This allows for headless tests to be created for the Tournament Client as well.

### DIFF
--- a/osu.Game.Tournament.Tests/TournamentTestBrowser.cs
+++ b/osu.Game.Tournament.Tests/TournamentTestBrowser.cs
@@ -19,9 +19,6 @@ namespace osu.Game.Tournament.Tests
                 Depth = 10
             }, AddInternal);
 
-            MenuCursorContainer.Cursor.AlwaysPresent = true;
-            MenuCursorContainer.Cursor.Alpha = 0;
-
             // Have to construct this here, rather than in the constructor, because
             // we depend on some dependencies to be loaded within OsuGameBase.load().
             Add(new TestBrowser());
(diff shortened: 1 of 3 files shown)


bdach:
Did you have a particular end goal when making this change? I'm not sure this is worth doing purely for the sake of it.

MiraiSubject:
I noticed that I wasn't able to headlessly test the tournament asset migration that I've been implementing in: https://github.com/MiraiSubject/osu/tree/tourney-asset-refactor

In order to prevent it from turning into a really big pull request that contains this change as well, I thought it would be better to split it up.

osu.Game.Tournament.Tests/TournamentTestBrowser.cs
```diff
@@ -19,6 +19,9 @@ protected override void LoadComplete()
                 Depth = 10
             }, AddInternal);
 
+            MenuCursorContainer.Cursor.AlwaysPresent = true;
```
peppy, line 22: would it not make sense to leave this in the base implementation if it's done in both places?

MiraiSubject: Right, I didn't consider that. I moved back to the base implementation

bdach (approved):
seems ok 👍

peppy (approved):

PR #9248, opened 2020-06-09
Override selected automation mod when starting song with autoplay

### LINKED ISSUE #9244: Ctrl+Enter doesn't launch AutoPlay when relax mod is turned on
**Describe the bug:** 
Usually, we press Ctrl+Enter to launch an AutoPlay on the selected map. But if the relax mod is turned on (in the mod selection), Ctrl+Enter doesn't launch AutoPlay anymore. Instead, it starts the map with relax mod (just like press Enter only).
(In stable, Ctrl+Enter always launch AutoPlay even if relax mod is on, and I guess lazer should match stable behaviour here)

**Screenshots or videos showing encountered issue:** not required I guess

**osu!lazer version:** 
2020.603.0

**Logs:** not required I guess

Game4all:
Fixes #9244, Bringing behaviour inline with stable.

### DIFF
--- a/osu.Game/Screens/Select/PlaySongSelect.cs
+++ b/osu.Game/Screens/Select/PlaySongSelect.cs
@@ -9,6 +9,7 @@ using osu.Framework.Screens;
 using osu.Game.Graphics;
 using osu.Game.Overlays;
 using osu.Game.Overlays.Notifications;
+using osu.Game.Rulesets.Mods;
 using osu.Game.Scoring;
 using osu.Game.Screens.Play;
 using osu.Game.Screens.Ranking;
@@ -101,7 +102,7 @@ namespace osu.Game.Screens.Select
 
                 if (mods.All(m => m.GetType() != autoType))
                 {
-                    Mods.Value = mods.Append(auto).ToArray();
+                    Mods.Value = mods.Where(m => !m.IncompatibleMods.Contains(typeof(ModAutoplay))).Append(auto).ToArray();
                     removeAutoModOnResume = true;
                 }
             }
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -589,6 +589,33 @@ namespace osu.Game.Tests.Visual.SongSelect
         }
 
         [Test]
+        public void TestAutoplayViaCtrlEnterOverridesSelectedAutomationMods()
+        {
+            addRulesetImportStep(0);
+
+            createSongSelect();
+
+            AddStep("select another automation mod", () => SelectedMods.Value = SelectedMods.Value.Append(new OsuModRelax()).ToArray());
+
+            AddStep("press ctrl+enter", () =>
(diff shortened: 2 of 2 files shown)


bdach:
I'm not sure about this fix. It seems like a band-aid. The `IncompatibleMods` should ideally just work; in this case it didn't, because the previous logic would want to start gameplay with relax + autoplay. This is where things go south:

https://github.com/ppy/osu/blob/3d13fe485d471ea83fc57a124218431fd3c91f83/osu.Game/Overlays/Mods/ModSection.cs#L127-L142

The list of mods is relax and then autoplay, in that order. The method loops over the *buttons*, in *their* order, which means that:

* it first tries to select `ModAutoplay` as it's first on the list; because it's incompatible with `ModRelax`, it selects autoplay and deselects relax,
* then it tries to select `ModRelax`; because it's incompatible with `ModAutoplay` it deselects it again and selects relax back.

This is why the fix will work, but also why I'm not 100% convinced it's a good fix. This stuff should ideally *just* work everywhere (i.e. if you ask to select autoplay with a given mod state from anywhere, it should be automatically ensured *for you* that autoplay will take priority as the last selected mod - just as if you went into the mod select overlay and selected it yourself).

Making it "just" work like that is going to take some more time and effort, however, so I'm not completely sure it's worth it.

osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -101,7 +102,7 @@ protected override bool OnStart()
 
                 if (mods.All(m => m.GetType() != autoType))
                 {
-                    Mods.Value = mods.Append(auto).ToArray();
+                    Mods.Value = mods.Where(m => !m.IncompatibleMods.Contains(typeof(ModAutoplay))).Append(auto).ToArray();
```
bdach, line 105: Oh, and also there's this - if a mod doesn't specify the base `ModAutoplay` but a deriving class in `SelectedMods`, then I'm pretty sure this won't work.

Game4all:
@bdach I'm not sure I get your point, especially the _looping over mods buttons_ part (Since all mods potentially incompatible with `ModAutoplay` are filtered out, `SelectTypes()` would get called with only autoplay mod type for the automation section); but I suppose you meant that the some of the logic bits should be added / moved higher up to `SelectTypes()` so given a mod list with mods incompatible with autoplay, autoplay would still take precedence over them; instead of making a local fix for this?

bdach:
What I described in my first post is an analysis of the original bug, and yeah, I'm wondering if a local fix is proper here. Ideally the code would go "select autoplay for me" and the mod select logic would know that it should deselect relax.

In fact you could probably use the old/new value in the value changed event to compensate for it.

osu.Game/Screens/Select/PlaySongSelect.cs
peppy: Why was this resolved?

peppy:
Keep in mind that adding more business logic to `ModSelectOverlay` is not the right direction. the incompatibility rules should be handled at somewhere more global. Possibly on `OsuGame.modsChanged`

PR #9249, opened 2020-06-09
Add support for decoding storyboard files that start with a [Variables] section

MasterLogick:
.osb file may start with Variables section located at the top of the file

### DIFF
--- a/osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
+++ b/osu.Game/Beatmaps/Formats/LegacyStoryboardDecoder.cs
@@ -33,6 +33,7 @@ namespace osu.Game.Beatmaps.Formats
             // note that this isn't completely correct
             AddDecoder<Storyboard>(@"osu file format v", m => new LegacyStoryboardDecoder());
             AddDecoder<Storyboard>(@"[Events]", m => new LegacyStoryboardDecoder());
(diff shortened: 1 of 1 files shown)


frenzibyte:
Please attach a link to the beatmap you're encountering this on.

bdach:
Thanks for the contribution! As mentioned above, it would be good to link a few examples of existing storyboard files that need this change to function.

Additionally, please refrain from making code changes using the github online interface in the future.

MasterLogick:
I just was looking through the osu!source code and osu!wiki trying to understand how it works and caught this bug. I don't know any beatmap that has variables section but I can create it for tests.

I am so lazy to open IDE for copy/paste one line of code.

smoogipoo:
We already had a test that happened to cover this scenario:
https://github.com/ppy/osu/blob/969d0b7228ad08ba3156de7fa8568bd36f832d7a/osu.Game.Tests/Beatmaps/Formats/LegacyStoryboardDecoderTest.cs#L98-L111
+ https://github.com/ppy/osu/blob/master/osu.Game.Tests/Resources/variable-with-suffix.osb

Unfortunately it creates a decoder rather than using `Decoder.GetDecoder<Storyboard>()`, but I've tested it that it works correctly without this change - the `SetFallbackDecoder()` case handles this.

I wouldn't be opposed to changing the tests to use `Decoder.GetDecoder<Storyboard>()`, but this change is unnecessary by itself.

PR #9253, opened 2020-06-11
Fix cursor not hiding with SDL2 backend

swoolcock:
`SetIconFromStream` is not currently implemented in the SDL2 backend and will need to be revisited here.

### DIFF
--- a/osu.Desktop/OsuGameDesktop.cs
+++ b/osu.Desktop/OsuGameDesktop.cs
@@ -122,14 +122,22 @@ namespace osu.Desktop
         {
             base.SetHost(host);
 
-            if (host.Window is DesktopGameWindow desktopWindow)
+            switch (host.Window)
             {
-                desktopWindow.CursorState |= CursorState.Hidden;
+                // Legacy osuTK DesktopGameWindow
+                case DesktopGameWindow desktopGameWindow:
+                    desktopGameWindow.CursorState |= CursorState.Hidden;
+                    desktopGameWindow.SetIconFromStream(Assembly.GetExecutingAssembly().GetManifestResourceStream(GetType(), "lazer.ico"));
+                    desktopGameWindow.Title = Name;
+                    desktopGameWindow.FileDrop += fileDrop;
+                    break;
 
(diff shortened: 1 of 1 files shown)


osu.Desktop/OsuGameDesktop.cs
```diff
@@ -122,14 +122,22 @@ public override void SetHost(GameHost host)
...
+                    desktopGameWindow.Title = Name;
+                    desktopGameWindow.FileDrop += fileDrop;
+                    break;
 
-                desktopWindow.FileDrop += fileDrop;
+                // SDL2 DesktopWindow
+                case DesktopWindow desktopWindow:
+                    desktopWindow.CursorState.Value |= CursorState.Hidden;
+                    desktopWindow.Title = Name;
+                    desktopWindow.FileDrop += fileDrop;
```
bdach, line 139: Kind of confused about this. File drop doesn't work as far as I can see on SDL, `FileDrop` itself [originates from osuTK](https://github.com/ppy/osuTK/blob/5d9cfdb1488b89d777afafa6f517bab83f9478ae/src/osuTK/INativeWindow.cs#L276-L279), and the SDL2 backend has a [`DragDrop` event](https://github.com/ppy/osu-framework/blob/c1a09d05e52c763c6776802eda2d289265b927b1/osu.Framework/Platform/Sdl/Sdl2WindowBackend.cs#L474-L485); what are the plans for this? If `FileDrop` is to be phased out then I think I'd be in favour of just removing this for now.

bdach (commented):
The cursor hiding part works on Windows. A little confused by one drive-by change here, though.

osu.Desktop/OsuGameDesktop.cs
swoolcock: Ah nice pickup, that was indeed meant to use the `DragDrop` event.  Will fix in a moment.

smoogipoo: Very weird that it's not explicitly implemented in order to hide it...Kinda weird that `FileDrop` isn't explicitly implemented so that it'd be hidden.

smoogipoo (approved):

Issue #9199, opened 2020-06-03
Add setup switching to the tournament client. 


MiraiSubject:
**Describe the new feature:** 

Right now the tournament client is not friendly towards having a setup for multiple tournaments. I've been moving around files multiple times to test out different videos, because I design graphics for tournaments, while using this client as well to run my own tournament. To make it easier for streamers and tournament hosts/graphic designers to manage files/setups across multiple tournaments a `TournamentProfileManager` for the tournament client could be implemented. 

This feature request also includes splitting up the current bracket.json into a `mappool.json` (or `rounds.json`), `teams.json` and `bracket.json`. This makes it easier for hosts that automate this data in their backend to publish this information to the streamers as this allows updates to just individual parts without having to re-download the whole bracket.json. This split up also makes it easier to manually make adjustments to teams in bulk using a text editor. 

**Proposal designs of the feature:** 

## Changing the directory structure

In the current state the assets gets supplied in this structure
```
(%APPDATA%/osu)
tournament
 ┣ flags
 ┣ mods
 ┣ videos
```

With this feature the directory structure will change to: 
```
(%APPDATA\osu)
tournaments
 ┣ mt-2020
 ┃ ┣ config
 ┃ ┃ ┣ bracket.json
 ┃ ┃ ┣ mappool.json
 ┃ ┃ ┗ teams.json
 ┃ ┣ flags
 ┃ ┣ mods
 ┃ ┣ videos
 ┃ ┣ header.png // A recognizable graphic of the tournament (UI Fluff)
 ┃ ┗ manifest.json
 ┗ owc-2021
 ┃ ┣ config
 ┃ ┃ ┣ bracket.json
 ┃ ┃ ┣ mappool.json
 ┃ ┃ ┗ teams.json
 ┃ ┣ flags
 ┃ ┣ mods
 ┃ ┣ videos
 ┃ ┗ manifest.json
 ┃...
```
## A tournament package

A tournament package consists of the configuration folder containing the essential information to run a match and it includes all the flags, videos and mod icons in their respective folders. A `manifest.json`  is provided to differentiate between the tournaments in the UI. 

The `manifest.json` contains metadata about the tournament that will be displayed to the user in a dropdown menu on the `SetupScreen` in the tournament client. This is the place where the user will be able to switch between profiles and create new tournaments. 

Example of what could be in a manifest.json

```json
{
    "name": "osu! World Cup 2021",
    "host": "osu! team",
    "acronym": "OWC",
[...]

PR #9256, opened 2020-06-11
Add basic setup switching for the tournament client

MiraiSubject:
- [x] Depends on https://github.com/ppy/osu/pull/9247 for tests to pass

From https://github.com/ppy/osu/issues/9199, this fulfils 
> Create a default tournament profile and the necessary migrations for the flags, mods, videos and other assets.

This PR changes the default path for tournaments from `%APPDATA%\osu\tournament`  to `%APPDATA%\osu\tournaments\default`.
Migrations for this are implemented so that this change does not break existing configurations and the necessary tests are also written for that. The migration also includes moving the `bracket.json` to the same directory. 

Upon migration a `tournament.ini` gets created with the following default value:
```
CurrentTournament = default
```
## Post-migration scenario
The user can now add directories in the following format:
```
(%APPDATA%/osu/tournaments)
tournament-name
 ┃ ┣ config
 ┃ ┃ ┗ drawings.ini
 ┃ ┣ flags
 ┃ ┣ mods
 ┃ ┣ videos
 ┃ ┣ bracket.json
 ┃ ┣ drawings.txt
 ┃ ┗ drawings_results.txt
```
and then change the `tournament.ini` to:
```
CurrentTournament = tournament-name
```
Because of this, switching the storage requires a reboot of the tournament client.


peppy:
Generally if something depends on another branch/PR, you should rebase it on top of that (so it will compile and work for testing).

MiraiSubject:
> Generally if something depends on another branch/PR, you should rebase it on top of that (so it will compile and work for testing).

I locally tested by merging this branch into https://github.com/ppy/osu/pull/9247, stashed any changes I wanted to keep and then applied it to this branch. It's a very roundabout way of doing things, but it was the only way I was familiar and comfortable with. 

I'll look into rebasing.

peppy:
It's fine this time around, just a heads-up for future contributions.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,153 @@
...
+{
+    public class TournamentStorage : WrappedStorage
+    {
+        private readonly GameHost host;
+        internal readonly TournamentVideoResourceStore VideoStore;
+        internal readonly Storage ConfigurationStorage;
+        private const string default_tournament = "default";
+        private const string config_directory = "config";
+
+        public TournamentStorage(GameHost host)
```
peppy, line 22: Why can't this just take in a storage, rather than the host itself?

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,153 @@
...
+            var destination = new DirectoryInfo(GetFullPath(default_tournament));
+            var cfgDestination = new DirectoryInfo(GetFullPath(default_tournament + Path.DirectorySeparatorChar + config_directory));
+
+            if (!destination.Exists)
+                destination.Create();
+
+            if (!cfgDestination.Exists)
+                destination.CreateSubdirectory(config_directory);
+
+            if (host.Storage.Exists("bracket.json"))
```
peppy, line 61: please learn how to avoid doing this by using functions.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,153 @@
...
+            {
+                Logger.Log("Migrating bracket to default tournament storage.");
+                var bracketFile = new System.IO.FileInfo(host.Storage.GetFullPath("bracket.json"));
+                moveFile(bracketFile, destination);
+            }
+
+            if (host.Storage.Exists("drawings.txt"))
```
peppy, line 68: why can't you use the existing `Migrate` method? this is mostly copy-paste of existing implementation.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            if (source.Exists)
+            {
+                Logger.Log("Migrating tournament assets to default tournament storage.");
+                copyRecursive(source, destination);
+                deleteRecursive(source);
+            }
+        }
+
+        private void copyRecursive(DirectoryInfo source, DirectoryInfo destination)
```
peppy, line 97: don't copy paste please.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,153 @@
...
+
+namespace osu.Game.Tournament.IO
+{
+    public class TournamentStorage : WrappedStorage
+    {
+        private readonly GameHost host;
+        internal readonly TournamentVideoResourceStore VideoStore;
+        internal readonly Storage ConfigurationStorage;
+        private const string default_tournament = "default";
+        private const string config_directory = "config";
```
peppy, line 20: why even bother? isn't there only one config file ever?

peppy (changes requested):

MiraiSubject:
There a couple of changes I made to this: 

- Introduce a new abstract class `MigratableStorage` that inherits from `WrappedStorage`
- Change the inheritance for `OsuStorage` and `TournamentStorage` to `MigratableStorage`
- `TournamentStorage` now takes in an existing storage as ctor parameter
- File moving in `TournamentStorage` is now called `moveFileIfExists` to prevent filling up `migrate()` up with if-statements.
- `drawings.ini` has been moved back from `tournaments/default/config` to `tournaments/default`

`MigratableStorage` now contains the Copy, Delete and Attempt operation methods as well as the constants for `ignore_files` and `ignore_directories`

I also made sure all the tests for `CustomDataDirectoryTest` still pass. 

I'm not sure if the `ignore_files` and `ignore_directories` should stay there, because imo every `MigratableStorage` should have their own array of ignored files and directories and now it's globally declared across every subclass. I just haven't found a solution to this yet that does not look bad. If anybody has an idea about this, let me know.

osu.Game/IO/MigratableStorage.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// </summary>
+    public abstract class MigratableStorage : WrappedStorage
+    {
+        internal static readonly string[] IGNORE_DIRECTORIES = { "cache" };
+
+        internal static readonly string[] IGNORE_FILES =
+        {
+            "framework.ini",
+            "storage.ini"
+        };
```
bdach, line 23: The obvious way to make this customisable by consumers is by making these into protected get-only virtual properties.

osu.Game.Tournament/Components/TourneyVideo.cs
```diff
@@ -29,7 +30,7 @@ public TourneyVideo(string filename, bool drawFallbackGradient = false)
         [BackgroundDependencyLoader]
         private void load(TournamentStorage storage)
         {
-            var stream = storage.GetStream($@"videos/{filename}");
+            var stream = storage.VideoStore.GetStream($@"{filename}");
```
bdach, line 33: This string interpolation is not required any more.
```suggestion
            var stream = storage.VideoStore.GetStream(filename);
```

osu.Game.Tournament/Configuration/TournamentStorageManager.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public TournamentStorageManager(Storage storage)
+            : base(storage)
+        {
+        }
+
+        protected override void InitialiseDefaults()
+        {
+            base.InitialiseDefaults();
+            Set(StorageConfig.CurrentTournament, string.Empty);
```
bdach, line 21: Is this even necessary if it's going to be overwritten upon first execution of `TournamentStorage` anyway?

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -0,0 +1,72 @@
...
+using osu.Game.IO;
+using System.IO;
+using osu.Game.Tournament.Configuration;
+
+namespace osu.Game.Tournament.IO
+{
+    public class TournamentStorage : MigratableStorage
+    {
+        private readonly Storage storage;
+        internal readonly TournamentVideoResourceStore VideoStore;
```
bdach, line 15: The `internal` looks really strange to me. I'd write this as

```suggestion
        public readonly TournamentVideoResourceStore VideoStore { get; }
```

osu.Game.Tournament/Screens/Drawings/Components/StorageBackedTeamList.cs
```diff
@@ -14,9 +14,9 @@ public class StorageBackedTeamList : ITeamList
     {
         private const string teams_filename = "drawings.txt";
 
-        private readonly Storage storage;
+        private readonly TournamentStorage storage;
 
-        public StorageBackedTeamList(Storage storage)
+        public StorageBackedTeamList(TournamentStorage storage)
```
bdach, line 19: Is this required? A `TournamentStorage` is a `Storage` still...

osu.Game.Tournament/Screens/Drawings/DrawingsScreen.cs
```diff
@@ -36,12 +36,12 @@ public class DrawingsScreen : TournamentScreen
...
         private Task writeOp;
 
-        private Storage storage;
+        private TournamentStorage storage;
 
         public ITeamList TeamList;
 
         [BackgroundDependencyLoader]
-        private void load(TextureStore textures, Storage storage)
+        private void load(TextureStore textures, TournamentStorage storage)
```
bdach, line 44: This one is confusing too, but I can understand why this would be required more (due to caching in `TournamentGameBase`, I presume). You can `CacheAs<Storage>`, fwiw.

osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
```diff
@@ -151,13 +152,13 @@ public void TestMigration()
...
 
-                    foreach (var file in OsuStorage.IGNORE_FILES)
+                    foreach (var file in osuStorage?.IgnoreFiles ?? Array.Empty<string>())
                     {
                         Assert.That(host.Storage.Exists(file), Is.True);
                         Assert.That(storage.Exists(file), Is.False);
                     }
 
-                    foreach (var dir in OsuStorage.IGNORE_DIRECTORIES)
+                    foreach (var dir in osuStorage?.IgnoreDirectories ?? Array.Empty<string>())
```
MiraiSubject, line 161: Not sure if this is the right way to prevent a Nullref exception and it also looks weird to loop over an empty array

peppy: You're correct in thinking this is not good code. just do a null check *before* iterating instead (you can use it to surround both the two `foreach` statements).

osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
```diff
@@ -151,16 +152,19 @@ public void TestMigration()
                     Assert.That(!host.Storage.ExistsDirectory(Path.Combine("test-nested", "cache")));
                     Assert.That(storage.ExistsDirectory(Path.Combine("test-nested", "cache")));
 
-                    foreach (var file in OsuStorage.IGNORE_FILES)
+                    if (osuStorage != null)
```
bdach, line 155: Shouldn't this be an actual assertion, as in the following?

```csharp
Assert.That(osuStorage, Is.Not.Null);
```

Imagine this regresses for some reason and `osuStorage` becomes null; this `if` will then mean that the test will silently skip over the nested asserts and stop testing if the ignore lists work at all.

(As an aside, I think it'd also make sense to change type of `osuStorage` to `MigratableStorage` as it's the most general type possible in this scenario)

osu.Game/IO/MigratableStorage.cs
```diff
@@ -0,0 +1,96 @@
...
+
+namespace osu.Game.IO
+{
+    /// <summary>
+    /// A <see cref="WrappedStorage"/> that is migratable to different locations.
+    /// </summary>
+    public abstract class MigratableStorage : WrappedStorage
+    {
+        internal virtual string[] IgnoreDirectories => Array.Empty<string>();
+        internal virtual string[] IgnoreFiles => Array.Empty<string>();
```
bdach, line 18: This is a weird modifier mix. Any consumers from a different assembly that might want to override this will not be able to. I think this calls for the [rarely-seen](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal)

```suggestion
        protected internal virtual string[] IgnoreDirectories => Array.Empty<string>();
        protected internal virtual string[] IgnoreFiles => Array.Empty<string>();
```

(Or heck, just make it public, I don't think that's that much of a deal.)

Needs to be fixed up in the override too if this happens.

osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
MiraiSubject: Good point! I'll add that and change the cast right away. I'll also include the property accessor change right away.

osu.Game/IO/MigratableStorage.cs
MiraiSubject: I've changed it to public for now, will await further feedback about this in case `protected` becomes the preferred way to go.

MiraiSubject:
I've been looking into making `Migrate` more sane.

Right now I moved the original `Migrate` from `OsuStorage` to `MigratableStorage` (+xmldoc) so that consumers can use this as a general purpose migrator.

I've also looked into making the `TournamentStorage`'s `Migrate` use `MigratableStorage`'s but that's not possible because the destination is nested inside the source. That goes along with the structure of what's being migrated is different from what it would be in a general case. 

Before:
```
(%APPDATA%/osu/)
... cache, files, etc.
 ┃ ┣ tournament/
 ┃ ┃ ┣  flags
 ┃ ┃ ┣  mods
 ┃ ┃ ┗  videos
 ┃ ┣  bracket.json
 ┃ ┣ drawings.txt.
 ┃ ┣ drawings_results.txt.
 ┃ ┗ drawings.ini
...  framework.ini, online.db, etc.
```

After: 
```
(%APPDATA%/osu/tournaments)
default
 ┃ ┣ flags
 ┃ ┣ mods
 ┃ ┣ videos
 ┃ ┣ bracket.json
 ┃ ┣ drawings.txt
 ┃ ┣ drawings.ini
 ┃ ┗ drawings_results.txt
```

bdach:
As a heads up, but this is probably going to have major merge conflicts with #9417 / #9418...

As for the issue mentioned I'd probably advise that the `Migrate()` method be overridable and that `TournamentStorage` hoist the files one level out of `tournament/` as a post-migration step.

MiraiSubject:
Alright thanks! I'll hold off from any more big changes until after those are merged

I made `Migrate` already overridable and its implementation in `TournamentStorage` is being completely overridden without ever calling `base` because this exception in `MigratableStorage` will get thrown:

```csharp
if (sourceUri.IsBaseOf(destinationUri))
    throw new ArgumentException("Destination provided is inside the source", destination.FullName);
```

I also believe at the same time that `TournamentStorage`'s `Migrate` is different enough to not need to call `base`.

I could just be talking about some non-issue and that overriding like this is fine, but I am not sure. Hopefully it will come to light once it can be reviewed again.

MiraiSubject:
I resolved the conflicts in an incoming commit and reran the tests locally again to make sure they didn't fail. 

I thought some more about this and my comment above can actually be disregarded. The intention of the migration method is that it is supposed to be a temporary method so that existing configurations get migrated after users update the client.

For future versions users would be instructed to create their new configurations in the format of `tournaments/<tournament acronym>` and then edit the `tournament.ini` appropriately to `CurrentTournament = <tournament acronym>`. 

The end goal is to put this functionality of adding and selecting tournaments in the UI itself.

peppy (approved):
Should be good to go now.

Issue #9242, opened 2020-06-09
Custom file paths for mania judging through skin.ini not implemented yet


Joppe27:
Instead of using the correct files with a custom path in skin.ini, osu!lazer uses the default ones. Lazer will use mania-hit0 even if there's a different file specified in skin.ini, for example combo_display/judge/0 in this skin: https://www41.zippyshare.com/v/Jib9a9n4/file.html .
**osu!lazer version:** 2020.603.0

mcendu:
After a bit of investigation, this is not the only configuration set that does not work. Custom file path for some other files (here `StageLeft` and `StageRight`) are also missing, [as demonstrated by this demo skin.](https://github.com/ppy/osu/files/4771589/special-skin.zip) I think it is because such data is not written into `ImageLookups` when generating a mania skin configuration.

smoogipoo:
Yep, full list of ones that are missing:
```
StageLeft
StageRight
StageBottom
StageHint
StageLight
```

Should be done in a separate PR though.

PR #9264, opened 2020-06-12
Add support for custom file paths for mania hit judgement

mcendu:
Partial fix to #9242 (https://github.com/ppy/osu/issues/9242#issuecomment-643328742 is not addressed)


smoogipoo (dismissed):

peppy:
Not sure if this affected mergeability, but I cannot for the life of me follow this code.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -257,6 +257,24 @@ private IBindable<TValue> lookupForMania<TValue>(LegacyManiaSkinConfigurationLoo
                 case LegacyManiaSkinConfigurationLookups.RightLineWidth:
                     Debug.Assert(maniaLookup.TargetColumn != null);
                     return SkinUtils.As<TValue>(new Bindable<float>(existing.ColumnLineWidth[maniaLookup.TargetColumn.Value + 1]));
+
+                case LegacyManiaSkinConfigurationLookups.Hit0:
+                    return SkinUtils.As<TValue>(getManiaImage(existing, "Hit0"));
```
peppy, line 262: why is this done this way, not in a similar way to existing `NoteImage` / `KeyImage`?

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
```diff
@@ -95,30 +126,13 @@ public Drawable GetDrawableComponent(ISkinComponent component)
...
-
-                case HitResult.Meh:
-                    return this.GetAnimation("mania-hit50", true, true);
-
-                case HitResult.Ok:
-                    return this.GetAnimation("mania-hit100", true, true);
-
-                case HitResult.Good:
-                    return this.GetAnimation("mania-hit200", true, true);
+            string image = GetConfig<ManiaSkinConfigurationLookup, string>(
```
peppy, line 131: i don't know how to read this line. maybe needs some commenting and splitting out.

also "image" is a bad variable name for this, it's more likely a "filename"?

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
```diff
@@ -47,15 +78,15 @@ private void sourceChanged()
 
         public Drawable GetDrawableComponent(ISkinComponent component)
         {
+            if (!isLegacySkin.Value || !hasKeyTexture.Value)
```
peppy, line 81: are you sure this hasn't changed behaviour? since it's now applying to other cases it wasn't before.

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
```diff
@@ -19,6 +20,36 @@ public class ManiaLegacySkinTransformer : ISkin
         private readonly ISkin source;
         private readonly ManiaBeatmap beatmap;
 
+        /// <summary>
+        /// Mapping of <see cref="HitResult"/> to ther corresponding
```
peppy, line 24:
```suggestion
        /// Mapping of <see cref="HitResult"/> to their corresponding
```

peppy (changes requested):

osu.Game.Rulesets.Mania.Tests/Skinning/TestSceneDrawableJudgement.cs
```diff
@@ -16,14 +17,17 @@ public class TestSceneDrawableJudgement : ManiaSkinnableTestScene
     {
         public TestSceneDrawableJudgement()
         {
+            var HitWindows = new ManiaHitWindows();
+
             foreach (HitResult result in Enum.GetValues(typeof(HitResult)).OfType<HitResult>().Skip(1))
```
peppy, line 22:
```suggestion
            foreach (HitResult result in new ManiaHitWindows(). GetAllAvailableWindows())
```

may work better (haven't tested, but looks to be what you want).

mcendu: Tested. Would end up causing the miss animation to be excluded

mcendu: Tested. Would end up causing the miss animation to be excluded

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
mcendu: https://github.com/ppy/osu/blob/ca2d0a6ea2de85313c20f5dc59dfcccd7c1aa15f/osu.Game.Rulesets.Mania/Skinning/LegacyManiaElement.cs#L21

Looking back at this piece of code, extension methods seems a lot better than what we have now -- We can then refactor that to use it. That would need another PR imo because it would touch some other unrelated files.

mcendu: Change variable `image` to `filename` anyway

osu.Game/Skinning/LegacySkin.cs
smoogipoo: What do you mean? It is done similar to those.

peppy: Yeah, realised they are all using column as the variable. Not sure it's worth making a custom lookup type just for hit results, but I guess I am hoping to not have separate implementations for each hit value (taiko's `LegacySkinTransformer` is probably the best example)

smoogipoo: This is a step above taiko's transformer, because taiko doesn't have the ability to skin individual images via skin.ini in the first place.
One potential code quality improvement is creating a different type to pass to the legacy skin, like `LegacyManiaSkinConfigurationHitLookup`, which could take a string and probably centralise things into the transformer. But it's legacy so /shrug from me.

peppy: Fair. Are you okay with the following?

```csharp

                case LegacyManiaSkinConfigurationLookups.Hit0:
                case LegacyManiaSkinConfigurationLookups.Hit50:
                case LegacyManiaSkinConfigurationLookups.Hit100:
                case LegacyManiaSkinConfigurationLookups.Hit200:
                case LegacyManiaSkinConfigurationLookups.Hit300:
                case LegacyManiaSkinConfigurationLookups.Hit300g:
                    return SkinUtils.As<TValue>(getManiaImage(existing, maniaLookup.Lookup.ToString()));
```

smoogipoo: Looks fine

PR #9270, opened 2020-06-13
Add new beatmap genres and languages

cl8n:
https://github.com/ppy/osu-web/issues/5852

- [x] Depends on https://github.com/ppy/osu-web/pull/6176


peppy:
Did you test this one? Changing the enum ordering looks like it will completely break the system (request uses int values of the enum).

cl8n:
I didn't, I thought `Order` there was just the display order

bdach:
Yeah, `[Order]` is just for UI. Perhaps the name is a smidge misleading (`DisplayOrder` maybe better?) This change looks safe when it comes to the request.

However seeing that the web-side PR has to get merged and go live before this can I've marked it as dependent on it.

peppy:
Ah, right. I thought some of the items were being added into the middle of the enum, but misread the diff. Should be fine as long as it fits in the UI correctly.

peppy:
As an aside, we'll probably want to fix this:
 
image

bdach:
Yeah that was a temporary solution at time of merging the original controls. It should wrap sanely at some point.

PR #9272, opened 2020-06-13
Fix crash on local score display

### LINKED ISSUE #9266: Crash when opening a local score in solo
**Describe the bug:**
- If you don't have any local score, download it from any online leaderboard (or play a beatmap)
- Click the score from the `Local` leaderboard
- [The game crashes]

**osu!lazer version:** 2020.613.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/4772323/runtime.log)

peppy:
Not sure it's worth the time to add test coverage for this (non-trivial). Can on request if it's deemed necessary.

Closes #9266.


bdach:
I can't decide whether this needs a test or not. I'm leaning "no", since I think this is unlikely to regress (either we stay with EF, in which case the includes will still be there/lazy loading will come, or we go to realm and lazy loading will be there too), but I'm not too sure.

smoogipoo (approved):
Let's get this in as is for now.

PR #9274, opened 2020-06-13
Add background loop at pause menu

### LINKED ISSUE #9170: Add back pause sound
**Describe the new feature:** 
On Stable when you pause the song you hear a "calming" sound which makes the pauses less empty. It would be really nice if this gets added back in

N/A ( https://streamable.com/4hmh25 ) to show what i mean

BananeVolante:
- Resolves #9170
- [x] Depends on `osu-resources` bump

The sound does not play without using the script given in the [osu-resources repository](https://github.com/ppy/osu-resources), but it doesn't crash.
This was done by looping a sample in the pause() function of player.cs , and stopping it when leaving the menu( resume and performUserRequestedExit() )


frenzibyte (changes requested):
Thanks for the contribution!

Looking at the code here, I think this should be more treated as a background music to the pause overlay instead of something played in `Player`, i.e. move all this logic to `PauseOverlay` instead (with overriding `PopIn` and `PopOut` for playing and stopping)

Also this is missing sample volume fading, maybe fade with a duration of `400` and easing of `Easing.InQuint`, from testing, that gives good effects and feels like osu!stable.

I would highly recommend checking on wrapping the channel inside a `DrawableSample` as that'll simplify the logic of fading the volume (`DrawableSample.VolumeTo(volume, duration, easing)`) and also for the sample to be automatically stopped and disposed on player disposal (instead of manually doing it on exit like you have now).

BananeVolante:
Thanks a lot for the recommandations.
I'm sorry to bother again, but do i have to do something specific about the checklist that was added in my precedent message?

Game4all:
For now you can only wait for the new version (containing the ambient pause sound) of the osu-resources package to be uploaded on NuGet (as game resources are distributed as a nuget package) and once it is you'll just have to update the package version in the game csproj file.

bdach:
As a clarification, most likely you won't have to bump the version of `osu-resources` yourself, either - when the next release is out, dependabot will automatically open a PR to bump the version, at which point checking off the check-box amounts to just merging `master`.

I've added the check-box so that everyone is aware that a resources bump needs to happen before this can be merged.

osu.Game/Screens/Play/Player.cs
```diff
@@ -583,6 +594,8 @@ void completeResume()
             {
                 GameplayClockContainer.Start();
                 IsResuming = false;
+
+                samplePause?.Stop();
```
smoogipoo, line 598: By doing this in the callback of `RequestResume`, you're only making it stop playing after the user completes the resume sequence (in osu! - clicking the orange cursor). This should likely be done outside.

osu.Game/Screens/Play/Player.cs
```diff
@@ -407,7 +413,10 @@ private void performUserRequestedExit()
             if (canPause)
                 Pause();
             else
+            {
+                samplePause?.Stop();
```
smoogipoo, line 417: Going into the pause screen and then pressing Alt+F4 to exit back to the menu will keep the sample playing in the background. You'd likely want this somewhere inside `OnExiting()` instead.

osu.Game/Screens/Play/Player.cs
```diff
@@ -564,6 +573,8 @@ public void Pause()
             GameplayClockContainer.Stop();
             PauseOverlay.Show();
             lastPauseActionTime = GameplayClockContainer.GameplayClock.CurrentTime;
+
+            samplePause?.Play();
```
smoogipoo, line 577: Related to issue I mentioned about resuming - this can play multiple times if you go into the resume sequence and then pause again, keeping one sample in a playing state in the background while the other is being stopped. Would also be fixed by moving the stop out of the callback.

smoogipoo (changes requested):

bdach:
There are usages of `DrawableSample.VolumeTo` [in framework tests](https://github.com/ppy/osu-framework/blob/3ed9af1937885fd73ea18017e39eb95c7f09f933/osu.Framework.Tests/Visual/Audio/TestSceneSamples.cs#L66-L67).

It's nigh impossible to meaningfully help you with the problems you're stuck on without being able to see your code.

BananeVolante:
Here's the code


```c#
[BackgroundDependencyLoader]
private void load(AudioManager audio, OsuConfigManager config)
{
     //.....
     // some unchanged code
     //.......
     sampleRestart = audio.Samples.Get(@"Gameplay/restart");

     // get the sample
     SampleChannel samplePause = audio.Samples.Get(@"Gameplay/pause-loop");
     // create a drawable sample using that sample
     drawableSamplePause = new DrawableSample(samplePause, true)
     {
         // make it loop
         Looping = true
     };

     mouseWheelDisabled = config.GetBindable<bool>(OsuSetting.MouseDisableWheel);
     //......
     //some unchanged code
     //........
```
```c#
public void Pause()
{
     //.....
     //some unchanged code
     //.....

     
     //volume at 1.0, sample isn't playing
     drawableSamplePause?.Play(); // play it
     //volume at 1.0, sample is playing
}
```

```c#
public void Resume()
{
     if (!canResume) return;

     IsResuming = true;
     PauseOverlay.Hide();
     
     //volume at 1, sample is playing
     drawableSamplePause?.VolumeTo(drawableSamplePause.Volume.Value -1.0f, 1000, Easing.OutQuint);
     //volume at 0, sample is playing
     
     //.....
     //some unchanged code
     //.....
}

```

In terms of sound, I can't hear any kind of fading, the sound stops when the continue button(calls the resume() function) is pressed

bdach:
Pasting code like that is inefficient and I don't want to spend time grafting it onto the existing tree and possibly do it wrong. It'd be way easier for everyone if you just pushed the code to a WIP branch or something that I (or anyone else) could easily check out and debug.

frenzibyte:
Definitely agreeing with @bdach's comment above, but looking closely at your snippets, you seem to just not have loaded and added the `DrawableSample` to the hierarchy (thus transforms will not be processed), but you also didn't check against the sample being potentially null, here's how your code should be:
```csharp
var sampleChannel = audio.Samples.Get(@"Gameplay/pause-loop");
if (sampleChannel != null)
{
    AddInternal(pauseLoop = new DrawableSample(sampleChannel)
    {
        Looping = true,
    }
}
```

But as I've also suggested, this logic should be moved to the `PauseOverlay`.

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -16,14 +19,40 @@ public class PauseOverlay : GameplayMenuOverlay
...
+            var sampleChannel = audio.Samples.Get(@"Gameplay/pause-loop");
+
+            if (sampleChannel != null)
+            {
+                pauseLoop = new DrawableSample(sampleChannel)
+                {
+                    Looping = true,
+                };
+                AddInternal(pauseLoop);
+                pauseLoop?.Play();
```
frenzibyte, line 42: Not quite sure about always playing this, is there a problem with playing and stopping it in the below methods?

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -16,14 +19,40 @@ public class PauseOverlay : GameplayMenuOverlay
...
+
+            var sampleChannel = audio.Samples.Get(@"Gameplay/pause-loop");
+
+            if (sampleChannel != null)
+            {
+                pauseLoop = new DrawableSample(sampleChannel)
+                {
+                    Looping = true,
+                };
+                AddInternal(pauseLoop);
```
frenzibyte, line 41: I would write like:
```suggestion
                AddInternal(pauseLoop = new DrawableSample(sampleChannel)
                {
                    Looping = true,
                });
```

frenzibyte (changes requested):
Otherwise seems to be good. (may request some input on the fading of the sample though, to me it seems good)

Not sure if adding a test scene is any worthy for this, since this is easily testable by hopping into `TestScenePause` and run through the basic `TestPauseResume` test case method.

osu.Game/Screens/Play/PauseOverlay.cs
BananeVolante: The only reason was that it felt more natural to me, since the volume change already "stops" the sound for the user.

frenzibyte: I guess that's fine.

bdach:
Seems to works fine and the source looks sane, but I'm wondering whether we'll be bringing legacy skinning of the pause sample back from stable. It could be a separate PR but I think there's a chance it could happen even in this one.

frenzibyte:
My bad, didn't know it was skinnable..., should be fairly simple to support skinning in this one, one liner change I believe.

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -24,13 +26,13 @@ public class PauseOverlay : GameplayMenuOverlay
...
         [BackgroundDependencyLoader]
-        private void load(OsuColour colours, AudioManager audio)
+        private void load(OsuColour colours, AudioManager audio, SkinManager skins)
         {
             AddButton("Continue", colours.Green, () => OnResume?.Invoke());
             AddButton("Retry", colours.YellowDark, () => OnRetry?.Invoke());
             AddButton("Quit", new Color4(170, 27, 39, 255), () => OnQuit?.Invoke());
 
-            var sampleChannel = audio.Samples.Get(@"Gameplay/pause-loop");
+            var sampleChannel = skins.GetSample(new SampleInfo("pause-loop")) ?? audio.Samples.Get(@"Gameplay/pause-loop");
```
frenzibyte, line 35: Changing skins will break this. If you check how other components have a skinnable sample playing (e.g. `NightcoreBeatContainer`), you'll find they're using the `SkinnableSound` component, put that in place of `DrawableSample`.

Though now you won't have access to the audio properties transforms, I would say write ```csharp
pauseLoop?.TransformBindableTo(pauseLoop.Volume, ...);
```
in replace to that for now.

**Review by frenzibyte — CHANGES_REQUESTED** [review]
As per above.

File: osu.Game/Screens/Play/PauseOverlay.cs [hunk 7]
```diff
@@ -24,6 +29,32 @@ private void load(OsuColour colours)
...
             AddButton("Retry", colours.YellowDark, () => OnRetry?.Invoke());
             AddButton("Quit", new Color4(170, 27, 39, 255), () => OnQuit?.Invoke());
+
+            AddInternal(pauseLoop = new SkinnableSound(new SampleInfo("pause-loop"))
+            {
+                Looping = true,
+            });
+            // PopIn is called before updating the skin, and when a sample is updated, its "playing" value is reset
+            // the sample must be played again(and if it plays when it shouldn't, the volume will be at 0)
+            pauseLoop.OnSkinChanged += () => pauseLoop.Play();
```
bdach, line 39: If the pause overlay isn't visible, this won't even fire (as the pause overlay is not present), so this seems to be even safer than written here.

bdach:
FWIW my question about whether we want to bring UI sample skinning back was genuine as I simply don't know if this is the direction we want to be taking.

BananeVolante:
It would be weird not to bring back an osu!stable feature because we can't do it in a perfectly clean way with the current classes.
It could be solved by turning SkinnableSound 's Play() into Play(bool playMutedSounds = false), but if this class was only meant for rulesets (that shouldn't change the volume), it would only encourage its (mis)use.
We could go around and make SkinnableSound keep the playing state when updating skins, which would mean that I could start the pause loop when loading the object and keep playing it at 0 volume when the overlay is hidden, but it sounds like a bad idea.
Would it mean that a new class is needed if we want to have a perfectly clean code AND a skinnable pause loop ?

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -24,6 +29,32 @@ private void load(OsuColour colours)
...
+        }
+
+        protected override void PopIn()
+        {
+            base.PopIn();
+
+            //SkinnableSound only plays a sound if its aggregate volume is > 0, so the volume must be turned up before playing it
+            pauseLoop?.TransformBindableTo(pauseLoop.Volume, 0.00001);
```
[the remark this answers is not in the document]
peppy: I think it's okay for now with the attached comment. What's not okay are the null checks. It should not be able to be null?

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -24,6 +29,32 @@ private void load(OsuColour colours)
...
+            pauseLoop?.TransformBindableTo(pauseLoop.Volume, 1.0f, 400, Easing.InQuint);
+            pauseLoop?.Play();
+        }
+
+        protected override void PopOut()
+        {
+            base.PopOut();
+
+            pauseLoop?.Stop();
```
peppy, line 56: null checks should not be required.

osu.Game/Screens/Play/PauseOverlay.cs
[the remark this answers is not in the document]
peppy: I've exposed better transform helpers, please use this post-merge https://github.com/ppy/osu/pull/9419

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -24,6 +29,32 @@ private void load(OsuColour colours)
...
+            pauseLoop?.TransformBindableTo(pauseLoop.Volume, 0.0f);
```
peppy, line 57: why is this required? what is it actually doing?

peppy (changes requested):

osu.Game/Screens/Play/PauseOverlay.cs
BananeVolante: This one is called when the user want to continue.
Since we exit the pause screen, we want to stop the loop, and its volume is muted to properly perform the volume fading in the next PopIn call.
I guess that since we have to change the volume before playing the loop, it's unneeded, and changing it to something different than 0 would only confuse peoples, so I will remove it.

peppy: it should probably be faded out and then stopped.

BananeVolante: It seems that the fading stops when the pause overlay is hidden, so i can't use the same fading as the one used when the overlay appears.
On osu!Stable, there is no fading out, but it's hidden by a loud sound when pressing continue.
For now an OutQuad easing of 190ms sounds okay.

BananeVolante:
Simply changing pauseLoop.TransformBindableTo(...) to pauseLoop.VolumeTo(...) does not work if a fading is used.
The only change is that the transformBindableTo call went from pauseLoop.TransformBindableTo to pauseLoop.samplesContainer.TransformBindableTo.
Did i miss something?

bdach:
The issue with the lack of playback is caused by the override of `IsPresent` inside of `SkinnableSound`:

https://github.com/ppy/osu/blob/640050977b80ee129583a2f958c094508518a3da/osu.Game/Skinning/SkinnableSound.cs#L97

which is `false` any time save for one frame after a skin change. Changing to:

```csharp
        public override bool IsPresent => Scheduler.HasPendingTasks || samplesContainer.Transforms.Any();
```

is *a* fix (but unsure if it's the right fix).

BananeVolante:
The problem effectively comes from IsPresent.
Wouldn't it be better to change to 
```
public override bool IsPresent =>Scheduler.HasPendingTasks || AlwaysPresent;
`` 
and set AlwaysPresent to true/false when entering/exiting the pause screen(or simply keeping it always set to true)?

**Comment by bdach** [comment]
there is not much reason as to why an audio sample should be always present. it does not need its `Update()` run, save for when it needs to queue playback or run transforms.

**Comment by peppy** [comment]
I've fixed remaining issues and added test coverage.

Some of the issues I found:
- Transitions didn't match the container
- the Finally statement wasn't working as expected (due to the `PauseOverlay` becoming not present) so the loop would never play a second time
- There were no tests

**Comment by peppy** [comment]
@smoogipoo requesting review on this.

File: osu.Game/Screens/Play/PauseOverlay.cs [hunk 11]
```diff
@@ -13,17 +16,50 @@ public class PauseOverlay : GameplayMenuOverlay
...
             AddButton("Quit", new Color4(170, 27, 39, 255), () => OnQuit?.Invoke());
+
+            AddInternal(pauseLoop = new SkinnableSound(new SampleInfo("pause-loop"))
+            {
+                Looping = true,
+            });
+
+            // PopIn is called before updating the skin, and when a sample is updated, its "playing" value is reset
+            // the sample must be played again
+            pauseLoop.OnSkinChanged += () => pauseLoop.Play();
```
bdach, line 44: FYI no longer necessary since #9645.

smoogipoo (approved):
Seems ok

osu.Game/Screens/Play/PauseOverlay.cs
```diff
@@ -13,17 +16,46 @@ public class PauseOverlay : GameplayMenuOverlay
     {
         public Action OnResume;
 
+        public override bool IsPresent => base.IsPresent || pauseLoop.IsPlaying;
```
peppy, line 19: I think this is causing input to be handled for longer and thus breaking tests.

bdach: I've ran the failing test a few times locally and it does fail sometimes, but not always. The autoplay tests which are failing seem to be very sensitive to update framerate (try running them in browser with window focus/no window focus - for me at least the first option is lengthy but mostly fits within the allotted timeframe while the other can fail fairly often on one of the seeks).

peppy (commented):
intentionally running CI a few times

PR #9276, opened 2020-06-13
Refactor lookup of mania skin configuration entries

mcendu:


osu.Game.Rulesets.Mania/Skinning/LegacyManiaColumnElement.cs
```diff
@@ -42,7 +42,7 @@ private void load()
             }
         }
 
-        protected override IBindable<T> GetManiaSkinConfig<T>(ISkin skin, LegacyManiaSkinConfigurationLookups lookup, int? index = null)
-            => base.GetManiaSkinConfig<T>(skin, lookup, index ?? Column.Index);
+        protected IBindable<T> GetManiaSkinConfig<T>(ISkin skin, LegacyManiaSkinConfigurationLookups lookup, int? index = null)
+            => skin.GetManiaSkinConfig<T>(lookup, index ?? Column.Index);
```
bdach, line 46: I find this staying in like this a little bit strange for two reasons.

* There's only one place where `index` is ever specified, and it's here:

  https://github.com/ppy/osu/blob/7662cdd3c6771f7f2458d9a248b7bc49449112ba/osu.Game.Rulesets.Mania/Skinning/LegacyColumnBackground.cs#L35-L36

  Maybe worth to replace just that one with the extension method and drop the `index` param?

* It's a little awkward how now it's half static extension methods, half protected methods. The extreme proposal would be to get rid of this one entirely and pass `Column.Index` (or just `Column`) everywhere whenever necessary but that's maybe a little too extreme?

bdach (commented):
I could get behind this in general but I feel this might not have gone far enough...?

bdach (approved):
I don't think I have any further improvements to suggest here :+1: 

Briefly considered hoisting `GetColumnSkinConfig` out as an extension method too (taking in something like an `IHasColumn`) but in the end it probably would have to take in `this` at usage sites so it'd probably be pretty ugly, and that's not what that interface is really for anyway.

smoogipoo (approved):
Seems ok

PR #9281, opened 2020-06-14
Fix gameplay clock not correctly reporting elapsed time

### LINKED ISSUE #8145: mania hitsounds do not work properly
mania maps with custom hitsounds, especially keymaps do not work correctly. a lot of the hitsounds are correct, but some are the wrong file and some don't play at all. sorry if this is a known issue, i couldn't find it anywhere
maps: https://osu.ppy.sh/beatmapsets/136986#mania/365278 https://osu.ppy.sh/beatmapsets/351167#mania/1700101 etc
video: https://youtu.be/Thxm-U6F1C0
**osu!lazer version:** latest
**Logs:** none

peppy:
Due to the way CurrentTime was implemented (ignoring the base value), elapsed time would not correctly propagate up the clock stack. This changes the approach to be able to use the base value while still applying the real-time adjust offset.

Fixes the part of #8145 where storyboard samples fail to play.

### DIFF
--- a/osu.Game/Screens/Play/GameplayClockContainer.cs
+++ b/osu.Game/Screens/Play/GameplayClockContainer.cs
@@ -253,7 +253,7 @@ namespace osu.Game.Screens.Play
         {
             // we always want to apply the same real-time offset, so it should be adjusted by the difference in playback rate (from realtime) to achieve this.
             // base implementation already adds offset at 1.0 rate, so we only add the difference from that here.
-            public override double CurrentTime => base.CurrentTime + Offset * (1 - Rate);
+            public override double CurrentTime => base.CurrentTime + Offset * (Rate - 1);
 
             public HardwareCorrectionOffsetClock(IClock source, bool processSource = true)
                 : base(source, processSource)
--- a/osu.Game/Screens/Play/GameplayClockContainer.cs
+++ b/osu.Game/Screens/Play/GameplayClockContainer.cs
@@ -251,8 +251,9 @@ namespace osu.Game.Screens.Play
 
         private class HardwareCorrectionOffsetClock : FramedOffsetClock
         {
-            // we always want to apply the same real-time offset, so it should be adjusted by the playback rate to achieve this.
-            public override double CurrentTime => SourceTime + Offset * Rate;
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -251,8 +251,9 @@ private void removeSourceClockAdjustments()
 
         private class HardwareCorrectionOffsetClock : FramedOffsetClock
         {
-            // we always want to apply the same real-time offset, so it should be adjusted by the playback rate to achieve this.
-            public override double CurrentTime => SourceTime + Offset * Rate;
+            // we always want to apply the same real-time offset, so it should be adjusted by the difference in playback rate (from realtime) to achieve this.
+            // base implementation already adds offset at 1.0 rate, so we only add the difference from that here.
+            public override double CurrentTime => base.CurrentTime + Offset * (1 - Rate);
```
bdach, line 256: Am I wrong in thinking this formula seems different to what it was before?

```
was: SourceTime + Offset * Rate

is:  FramedOffsetClock.CurrentTime + Offset * (1 - Rate) =
     FramedClock.CurrentTime + Offset + Offset * (1 - Rate) =
     FramedClock.CurrentTime + Offset * (2 - Rate) =
     SourceTime + Offset * (2 - Rate)
```

Think you might have meant `(Rate - 1)`?

bdach (commented):
I've checked that this does fix the zero `ElapsedTime`, but the linked issue shouldn't be closed yet as there's another issue with the sample's lifetime end which isn't yet resolved (see [last paragraph](https://github.com/ppy/osu/issues/8145#issuecomment-643686612) of my comment).

It might be hard to tell right now due to the layered hit samples which shouldn't be there, but due to that lifetime problem the kick drums that are supposed to play at about ~0:13 of the linked map aren't played right now even with this fix.

osu.Game/Screens/Play/GameplayClockContainer.cs
peppy: That was the intention, oops.

bdach (approved):

smoogipoo (approved):

PR #9285, opened 2020-06-14
Add shortcut to go home

Joehuu:
The shortcut (<kbd>Alt</kbd>+<kbd>Home</kbd>) is used in most browsers.

[Reference](https://en.wikipedia.org/wiki/Table_of_keyboard_shortcuts):

image

Not sure if the second commit is right, visually it looks the same though. `HoverBackground`'s alpha needs to be > 0 for `OnClick`'s `FlashColour` to be visible when not hovered, so I changed `Opacity` instead.

### DIFF
--- a/osu.Game/Overlays/Toolbar/ToolbarButton.cs
+++ b/osu.Game/Overlays/Toolbar/ToolbarButton.cs
@@ -62,6 +62,7 @@ namespace osu.Game.Overlays.Toolbar
         protected ConstrainedIconContainer IconContainer;
         protected SpriteText DrawableText;
         protected Box HoverBackground;
+        private readonly Box FlashBackground;
         private readonly FillFlowContainer tooltipContainer;
         private readonly SpriteText tooltip1;
         private readonly SpriteText tooltip2;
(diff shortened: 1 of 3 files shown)


bdach:
How are you testing this? I can't get the shortcut to work on any screen on WIndows. I thought it was textboxes and scroll containers eating up the input but even on screens that don't have either it doesn't work.

osu.Game/Overlays/Toolbar/ToolbarButton.cs
```diff
@@ -78,9 +78,8 @@ public ToolbarButton()
                 HoverBackground = new Box
                 {
                     RelativeSizeAxes = Axes.Both,
-                    Colour = OsuColour.Gray(80).Opacity(180),
+                    Colour = OsuColour.Gray(80).Opacity(0),
```
bdach, line 81: Yeah I'm not sure about this one, either. Splitting flash & hover entirely into separate boxes would definitely be cleaner.

peppy: Agree with this. Splitting out would be best.

peppy (approved):

PR #9299, opened 2020-06-16
Fix regression in android build parsing behaviour

### LINKED ISSUE #9183: Notification that an update is available no longer appears
**Describe the bug:** 

The notification that appears at the game start when an update is available don't work

Screenshot_20200602_173906_sh ppy osulazer

**osu!lazer version:** 

2020.523.0

**Logs:**
 Android

peppy:
The previous change made was not tested, and badly regressed the ability to read android build numbers. It has been a while since I've looked into why we did things the way we did, so I made a note of documenting inline.

The pain point is that we need to keep the obsoleted (in version 28) method call in order to support older android SDK versions (which are still recent enough that we do want to support, for instance my test device is running version 24 but still runs osu! 60fps with no issues).

Closes https://github.com/ppy/osu/issues/9183.

### DIFF
--- a/osu.Android/OsuGameAndroid.cs
+++ b/osu.Android/OsuGameAndroid.cs
@@ -38,7 +38,6 @@ namespace osu.Android
                     }
                     else
                     {
-
 #pragma warning disable CS0618 // Type or member is obsolete
                         // this is required else older SDKs will report missing method exception.
                         versionName = packageInfo.VersionCode.ToString();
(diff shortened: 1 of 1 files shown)


bdach:
This should fix #9183, right?

peppy:
Oh yep, forgot to add that.

smoogipoo (dismissed):
I've got nothing to test this with, so let's see if it works on deploys.

smoogipoo (approved):

PR #9301, opened 2020-06-16
Implement rate adjustment for storyboard samples

H2n9:
Implements IApplicableToSample to allow mods to adjust storyboard samples.
All rate adjust mods now implement this interface to match playback rates of the map and storyboard samples.

As SampleChannel does not support tempo adjustments, all adjustments are instead done as frequency adjustments, matching osu!stable's behaviour.


frenzibyte (approved):
While the code seems to be good and works fine for what it is aiming to fix, there is still the issue where modifying the playback rate from a replay / autoplay will not adjust for storyboard samples playback, but seems to be non-straightforward to fix, so I think the approach here is fine for now.

frenzibyte (changes requested):
Though this is missing a test scene, should be fairly simple to add (set `SelectedMods` and add the drawable storyboard sample inside).

osu.Game/Rulesets/Mods/ModRateAdjust.cs
```diff
@@ -2,12 +2,13 @@
...
 
 using osu.Framework.Audio;
+using osu.Framework.Audio.Sample;
 using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
 
 namespace osu.Game.Rulesets.Mods
 {
-    public abstract class ModRateAdjust : Mod, IApplicableToTrack
+    public abstract class ModRateAdjust : Mod, IApplicableToTrack, IApplicableToSample
```
bdach, line 11: This is a minor point, but I find the double interface specification a little iffy/fine-grained. It's not unreasonable that someone might find `IAppplicableToTrack` but not find/forget about `IApplicableToSample` (as you even did in 4138f61).

I don't think the two should be merged outright, but how about a grouping interface?

```csharp
public interface IApplicableToAudio : IApplicableToTrack, IApplicableToSample
{
}
```

osu.Game/Rulesets/Mods/IApplicableToAudio.cs
```diff
@@ -1,6 +1,5 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
```
peppy, line 1: please label commits better. this should be something like "Add missing licence header"

H2n9:
In regards to the test scene, I was trying to look for existing examples to work off from.
Is something like this what I should be doing -
```
[Test]
public void TestSamplePlaybackWithRateMods()
{
    GameplayClockContainer gameplayContainer = null;
    DrawableStoryboardSample sample = null;

    AddStep("create container", () =>
    {
        Add(gameplayContainer = new GameplayClockContainer(CreateWorkingBeatmap(new OsuRuleset().RulesetInfo), new[] { new OsuModDoubleTime() }, 0));

        gameplayContainer.Add(sample = new DrawableStoryboardSample(new StoryboardSampleInfo(string.Empty, 0, 1))
        {
            Clock = gameplayContainer.GameplayClock
        });
    });

    AddStep("start", () => gameplayContainer.Start());

    // Test here
}
```

If so, how can I confirm that the sample actually successfully received the rate adjustment?
Thanks.

bdach:
You would have to set `SelectedMods` before the storyboard sample is created, so that its value is DI'd in when it loads and the adjustment is applied. But even then I suppose you'd have to expose the `SampleChannel` in `DrawableStoryboardSample` as protected to assert anything.

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -70,6 +73,37 @@ public void TestSamplePlaybackAtZero()
             AddUntilStep("sample playback succeeded", () => sample.LifetimeEnd < double.MaxValue);
         }
 
+        [Test]
+        public void TestSamplePlaybackWithRateMods()
+        {
+            GameplayClockContainer gameplayContainer = null;
+            TestDrawableStoryboardSample sample = null;
+
+            OsuModDoubleTime doubleTimeMod = null;
```
frenzibyte, line 82: Why not test more than one mod per-case, along with mods that have custom `SpeedChange` values set. (using `[TestCase(...)]`s)

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -70,6 +73,37 @@ public void TestSamplePlaybackAtZero()
...
+
+            AddStep("create container", () =>
+            {
+                var beatmap = Beatmap.Value = CreateWorkingBeatmap(new OsuRuleset().RulesetInfo);
+
+                Add(gameplayContainer = new GameplayClockContainer(beatmap, new[] { doubleTimeMod = new OsuModDoubleTime() }, 0));
+
+                SelectedMods.Value = new[] { doubleTimeMod };
+                Beatmap.Value = new TestCustomSkinWorkingBeatmap(beatmap.Beatmap, gameplayContainer.GameplayClock, Audio);
```
frenzibyte, line 91: The ordering is *very* incorrect, the beatmap should be created first along the mods, then the container will be created with providing what has been set (i.e.):

**Though this depends on another suggestion in the construction parameters of `TestCustomSkinWorkingBeatmap`.**

```suggestion
                Beatmap.Value = new TestCustomSkinWorkingBeatmap(new OsuRuleset().RulesetInfo, Audio);
                SelectedMods.Value = new[] { doubleTimeMod };
                
                Add(gameplayContainer = new GameplayClockContainer(Beatmap.Value, SelectedMods.Value, Audio);
```

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -70,6 +73,37 @@ public void TestSamplePlaybackAtZero()
...
+            });
+
+            AddStep("create storyboard sample", () =>
+            {
+                gameplayContainer.Add(sample = new TestDrawableStoryboardSample(new StoryboardSampleInfo("test-sample", 1, 1))
+                {
+                    Clock = gameplayContainer.GameplayClock
+                });
```
frenzibyte, line 99: Why not merge with above and call that step `setup storyboard sample`, makes sense that way IMO (since we're not testing something with the clock, we just want to setup a working sample).

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -99,5 +133,28 @@ public void Dispose()
             {
             }
         }
+
+        private class TestCustomSkinWorkingBeatmap : ClockBackedTestWorkingBeatmap
+        {
+            private readonly AudioManager audio;
+
+            public TestCustomSkinWorkingBeatmap(IBeatmap beatmap, IFrameBasedClock referenceClock, AudioManager audio)
+                : base(beatmap, null, referenceClock, audio)
```
frenzibyte, line 142:
```suggestion
            public TestCustomSkinWorkingBeatmap(RulesetInfo ruleset, AudioManager audio)
                : base(ruleset, null, audio)
```

frenzibyte (changes requested):

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
H2n9: ModRateAdjust and ModTimeRamp both have SpeedChange properties but I would require some type checking in the test to be able to change their values as a test case.
Is there a reason why ModTimeRamp doesn't inherit ModRateAdjust?

bdach: It would be definitively out of scope, and I would tend to disagree with adding that inheritance.

bdach:
I've made some changes here. They're mostly menial; 1d5084c is by far the most interesting one.

You've exposed a bit of a flaw in the design of `ModTimeRamp`, namely that its `SpeedChange` bindable should not be manually settable (it's supposed to be computed based on `InitialRate` and `FinalRate`). This is why I was also opposed to the suggestion of unifying it with `ModRateAdjust`.

As it's out of scope I didn't do anything about it except change the usage in the test, but going forward we might want to restrict access to `SpeedChange`. A decent way forward for now would be to expose it as an `IBindableNumber<double>` (although that doesn't make it entirely immutable, either).

frenzibyte (dismissed):
Seems fine for now.

H2n9:
I added UserPlaybackRate adjustments now, fixing the replay issues.

osu.Game/Storyboards/Drawables/DrawableStoryboardSample.cs
```diff
@@ -28,12 +34,18 @@ public DrawableStoryboardSample(StoryboardSampleInfo sampleInfo)
...
+                return;
+
+            Channel.Volume.Value = sampleInfo.Volume / 100.0;
+
+            foreach (var mod in mods.Value.OfType<IApplicableToSample>())
+                mod.ApplyToSample(Channel);
 
-            if (channel != null)
-                channel.Volume.Value = sampleInfo.Volume / 100.0;
+            Channel.AddAdjustment(AdjustableProperty.Frequency, player.GameplayClockContainer.UserPlaybackRate);
```
bdach, line 48: Please clarify what this is fixing - I assume it's fixing the storyboard samples not adhering to the playback rate set in the replay settings?

I'm not a fan of this solution as the rate shouldn't be exposed like that. I would have preferred the current playback rate to be stored in `GameplayBeatmap` which could be DI'd in here without having to make it public in the gameplay clock container.

H2n9: My bad, It was intended to fix playback rate in replay settings. I'm reverting these changes and will rework those specific changes for a new PR. Thanks for the feedback.

smoogipoo (changes requested):
Seems pretty fine otherwise.

frenzibyte (approved):

smoogipoo (approved):

PR #9313, opened 2020-06-18
Block screen suspend while gameplay is active

peppy:
Adds very basic usage of framework-side implementation. Currently only supported on android and iOS. Wanted to get this in before we completely forget that the framework support exists for it.

- [x] Needs mobile device testing (will do this tomorrow)

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -125,6 +125,8 @@ namespace osu.Game.Screens.Play
 
         private GameplayBeatmap gameplayBeatmap;
 
+        private ScreenSuspensionHandler screenSuspension;
+
         private DependencyContainer dependencies;
 
         protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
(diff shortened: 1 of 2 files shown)


Game4all:
Tested this on my android device and it seems to work as intended (blocking screen suspension when the gameplay is paused).

However, that behaviour of only blocking screen suspension when gameplay is paused doesn't make much sense to me. Wouldn't it be better if the screen suspension was blocked during all the time the player is in gameplay? That means that the game could currently suspend on the user's device during gameplay if he weren't to have interaction with the device during a few seconds (you can reproduce this on this branch by enabling NoFail and not interacting with the device during a few seconds / or by hitting a very long break in a map).

bdach:
That just looks like a forgotten negation in the `BindValueChanged` callback. Judging by PR title and commit message of add1265 I can't imagine it was intended to block suspension only during pause.

Game4all:
I'm still of the opinion that the screen suspension should be blocked during gameplay even when it is paused (at least for now) given that the framework doesn't support resuming suspended activity on mobile devices.

peppy:
Not supporting resuming is not a concern here. The idea is it should only avoid suspension during gameplay, not when paused.

peppy:
@bdach no, the direction is actually correct here. if you pause the game your device should be allowed to go to sleep.

swoolcock:
> ...given that the framework doesn't support resuming suspended activity on mobile devices.

Works fine on iOS, not Android.

peppy:
Tested and confirmed working as intended.

smoogipoo (approved):
Looks ok

PR #9316, opened 2020-06-19
Add more detailed score statistics to the results screen

smoogipoo:
Click on the expanded score to show more statistics!
Preview: https://drive.google.com/file/d/1nCMhn3kUPBMnGLKJCpMAcNI5cJdR0J4a/view?usp=sharing

---

1. Doesn't contain any fancy transforms for now since the PR is already huge and I just want to get this in for now. It's coded in such a way that the transforms could be done with relative ease.
2. Currently only generates statistics via a gameplay run-through. The eventual goal is for this to happen in the background in the future, but that's too significant of an implementation to happen in this PR.
3. More statistics also coming in further PRs, this is only 10% implemented for osu! and other rulesets haven't been considered yet.

Implementation-wise there's a few important areas:

### `HitEvents`

A generic format that describes the relevant state around a `HitResult`, such as the last object, the current object, the time offset, and an optional positional offset.
The positional offset isn't explicitly defined so rulesets can use this to mean whatever they want.

The format may be expanded in the future, but I would like to avoid that if possible as it would also mean having an `OsuScoreInfo` etc.

This is emitted by the base `ScoreProcessor`, and rulesets can override `CreateHitEvent()` to 

### Changes to `ScorePanel` / `ScorePanelList`

I've had to introduce the `ScorePanelTrackingContainer`, which basically wraps a `ScorePanel`. This is done this way for two primary reasons:
1. I need to avoid 1-frame issues (flashing, not scrolling, etc) caused by ordering of events around the flow/scroll areas of `ScrolPanelList`. It's tightly controlled and putting everything inside the flow is the only way I can avoid all 1-frame issues.
2. I need a way to transfer the `ScorePanel` into another container, which is done by the `Attach()` and `Detach()` methods of `ScorePanelTrackingContainer`. The `ScorePanelTrackingContainer` will continue to remain in the flow and take on the size of the `ScorePanel`, but the `ScorePanel` itself is free to move.

It's a little ugly in terms of the `ResultsScreen` implementation, I agree, but I think it's quite well defined in terms of what it's doing.

### The creation of statistics

I've tried to make the API for creating statistics pretty flexible, for user rulesets to be able to do whatever they want.
You essentially return an array of `StatisticRow`s to `Ruleset.CreateStatisticsForScore()`, each containing an array of `StatisticItem` columns. As you'll notice, this is structurally similar to a `GridContainer`, except more expressive.


osu.Game.Rulesets.Osu/Judgements/OsuHitCircleJudgementResult.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        /// <summary>
+        /// The <see cref="HitCircle"/>.
+        /// </summary>
+        public HitCircle HitCircle => (HitCircle)HitObject;
+
+        /// <summary>
+        /// The position of the player's cursor when <see cref="HitCircle"/> was hit.
+        /// </summary>
+        public Vector2? CursorPositionAtHit;
```
frenzibyte, line 21: Shouldn't this be named `PositionOffset`?

smoogipoo: I didn't because it's not really an offset. It's a full playfield coordinate position.

frenzibyte: Ah right, then shouldn't `HitEvent.PositionOffset` be renamed to `HitEvent.HitPosition` or similar? I thought that property for a perfect aim would be `0f`.

smoogipoo: Hmm, I conceptualised that one as a general positional value - could be hit position, could be a miss position, could be offset, could be absolute position, could be null for a hit, etc.

However now I'm not 100% sure on the API here - in thinking about it a bit more, rulesets that require multiple simultaneous positions (touchscreen rulesets) couldn't really make effective use of just this one property.

frenzibyte: My opinion here would be to declare a specific `OsuHitEvent` that holds the positional offset value instead of adding it in the base class, other custom rulesets can declare their own hit events with their own properties usable by their own statistics.

smoogipoo: As I mentioned in the OP, that would require an `OsuScoreInfo`/`CatchScoreInfo`/etc to make use of, which I'm also not convinced is the best way to go. I think it'd have to be a dictionary if that's the most suitable path.

osu.Game/Screens/Play/Player.cs
```diff
@@ -460,7 +460,7 @@ protected virtual ScoreInfo CreateScore()
         {
             var score = new ScoreInfo
             {
-                Beatmap = Beatmap.Value.BeatmapInfo,
+                Beatmap = gameplayBeatmap.BeatmapInfo,
```
bdach, line 463: Causes crashes [as discussed on Discord](https://discord.com/channels/188630481301012481/188630652340404224/723926724244537424) as it looks the gameplay beatmap's instance is not necessarily attached.

osu.Game/Screens/Ranking/ResultsScreen.cs
```diff
@@ -169,32 +203,82 @@ public override void OnEntering(IScreen last)
...
-                ScrollbarVisible = false;
-            }
+                // Detach and move into the local container.
+                scorePanelList.Detach(expandedPanel);
+                detachedPanelContainer.Add(expandedPanel);
+
+                // Move into its original location in the local container.
+                var origLocation = detachedPanelContainer.ToLocalSpace(screenSpacePos);
+                expandedPanel.MoveTo(origLocation);
+                expandedPanel.MoveToX(origLocation.X);
```
bdach, line 242: Does the `MoveToX` need to exist? It seems superfluous, unless I'm missing something...

osu.Game/Screens/Ranking/ScorePanelList.cs
```diff
@@ -145,15 +162,92 @@ protected override void Update()
...
+        public ScorePanel GetPanelForScore(ScoreInfo score) => flow.Single(t => t.Panel.Score == score).Panel;
+
+        /// <summary>
+        /// Detaches a <see cref="ScorePanel"/> from its <see cref="ScorePanelTrackingContainer"/>, allowing the panel to be moved elsewhere in the hierarchy.
+        /// </summary>
+        /// <param name="panel">The <see cref="ScorePanel"/> to detach.</param>
+        /// <exception cref="InvalidOperationException">If <paramref name="panel"/> is not a part of this <see cref="ScorePanelList"/>.</exception>
+        public void Detach(ScorePanel panel)
+        {
+            var container = flow.FirstOrDefault(t => t.Panel == panel);
```
bdach, line 205: usual nit (places earlier even used `Single` too)

```suggestion
            var container = flow.SingleOrDefault(t => t.Panel == panel);
```

also goes for `Attach`

bdach (dismissed):
Cool stuff! :+1:

It took me a few hours but I went through this. Besides the DB crash things look alright to me although I admittedly didn't review the layout stuff that thoroughly. I might have another look at that with a fresher mind.

osu.Game/Screens/Ranking/ResultsScreen.cs
smoogipoo: It seems like the X transform that touches property `X` doesn't cancel the movement transform that touches `Position`. But I can improve this one - it previously wasn't extracted into the `origLocation` field.

osu.Game.Rulesets.Osu/Statistics/AccuracyHeatmap.cs
```diff
@@ -0,0 +1,226 @@
...
+
+            double angle1 = Math.Atan2(end.Y - hitPoint.Y, hitPoint.X - end.X); // Angle between the end point and the hit point.
+            double angle2 = Math.Atan2(end.Y - start.Y, start.X - end.X); // Angle between the end point and the start point.
+            double finalAngle = angle2 - angle1; // Angle between start, end, and hit points.
+            float normalisedDistance = Vector2.Distance(hitPoint, end) / radius;
+
+            // Convert the above into the local search space.
+            Vector2 localCentre = new Vector2(points_per_dimension) / 2;
+            float localRadius = localCentre.X * inner_portion * normalisedDistance; // The radius inside the inner portion which of the heatmap which the closest point lies.
+            double localAngle = finalAngle + Math.PI - MathUtils.DegreesToRadians(rotation); // The angle inside the heatmap on which the closest point lies.
```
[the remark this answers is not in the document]
smoogipoo: See how the comments I've added feel. I'm not really good at explaining this stuff :(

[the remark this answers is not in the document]
bdach: Hmm, I feel that the previous version was easier to grasp upon understanding the pi trick.

The way I would look at it is that the angles calculated go `end -> start` and `end -> actual hit point`. If we consider the line going through start and end:

* If the actual hit point was on the *near* side, the `end -> actual hit point` vector will have the same direction as `end -> start`, therefore the angle will be near zero, therefore it will end up on the right side of the coordinate origin (relative to the positive part of the X axis) - but we want it to be exactly opposite that.
* Conversely, if the hit is on the *far* side, it will create a 180 degree angle, ending up on the left, but we actually want it on the right. This is what the 180deg rotation (or pi in radians) is for.

Might just be me, though - interested in @peppy's take on this.

osu.Game.Rulesets.Osu/Judgements/OsuHitCircleJudgementResult.cs
smoogipoo: Are you fine with the naming for now too?

bdach: Ah, sort of glossed over that discussion. `Offset` sort of implies relative positioning for me too. Just `Position` would probably be ok for now if we're going for a general thing.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs
```diff
@@ -86,6 +91,13 @@ private void load()
             AccentColour.BindValueChanged(accent => ApproachCircle.Colour = accent.NewValue, true);
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            inputManager = GetContainingInputManager();
```
peppy, line 98: Quite a performance concern, but I guess it can be overlooked until we have pooling in?

smoogipoo: Indeed it is. If only the keybinding system provided the input state...

Maybe we can DI in the playfield instead, and store the current cursor position globally there?

osu.Game.Rulesets.Mania/ManiaRuleset.cs
```diff
@@ -309,6 +310,21 @@ private PlayfieldType getPlayfieldType(int variant)
...
+        public override StatisticRow[] CreateStatisticsForScore(ScoreInfo score, IBeatmap playableBeatmap) => new[]
+        {
+            new StatisticRow
+            {
+                Columns = new[]
+                {
+                    new StatisticItem("Timing Distribution", new HitEventTimingDistributionGraph(score.HitEvents)
+                    {
+                        RelativeSizeAxes = Axes.X,
+                        Height = 250
```
peppy, line 323: not sure we want rulesets to be specifying the sizing of these components..

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableHitCircle.cs
peppy: I think this is okay for now. LoadComplete is called on first display of each hitcircle so the overall load should be spread out enough that it doesn't matter.

osu.Game/Screens/Ranking/Statistics/HitEventTimingDistributionGraph.cs
```diff
@@ -0,0 +1,173 @@
...
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
+using osu.Game.Rulesets.Scoring;
+
+namespace osu.Game.Screens.Ranking.Statistics
+{
+    /// <summary>
+    /// A graph which displays the distribution of hit timing in a series of <see cref="HitEvent"/>s.
+    /// </summary>
+    public class HitEventTimingDistributionGraph : CompositeDrawable
```
peppy, line 21: would've been nice if each component was in its own PR.

peppy:
Any reason this PR wasn't split out at all? I'm on my fourth sitting and still having huge trouble reviewing and navigating the changes :(.

peppy:
Okay, finally got through all of this. There's a few remaining (UI/UX) issues I have but the code seems sane enough that we can move forward with it. Also the performance issue seems to be just the first display, so again, we can investigate further when we add proper animations to the elements.

PR #9330, opened 2020-06-21
Reduce mania's HP drain by 20%

smoogipoo:
Since stable didn't have any HP drain, mania mappers ramped up the HP drain rate to be harsh via misses.
So I've just uniformly reduced the HP drain by 20% at all levels for now, hoping it'll be enough. This changes the mappings of `(95%, 70%, 30%)` at `(OD0, OD5, OD10)` respectively to `(96%, 76%, 44%)`.

### DIFF
--- a/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
@@ -44,6 +44,7 @@ namespace osu.Game.Rulesets.Scoring
         private double gameplayEndTime;
 
         private readonly double drainStartTime;
+        private readonly double drainLenience;
 
         private readonly List<(double time, double health)> healthIncreases = new List<(double, double)>();
         private double targetMinimumHealth;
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/Scoring/DrainingHealthProcessor.cs
```diff
@@ -95,6 +101,8 @@ public override void ApplyBeatmap(IBeatmap beatmap)
             )));
 
             targetMinimumHealth = BeatmapDifficulty.DifficultyRange(beatmap.BeatmapInfo.BaseDifficulty.DrainRate, min_health_target, mid_health_target, max_health_target);
+            targetMinimumHealth += drainLenience * (1 - targetMinimumHealth);
+            targetMinimumHealth = Math.Min(1, targetMinimumHealth);
```
peppy, line 105: i don't quite follow this math.

set drainLenience = 1

targetMinimumHealth is calculated to be 0.2.

`targetMinimumHealth = 0.2 + 1 * (1 - 0.2)`

targetMinimumHealth = 1

wouldn't this make the drain higher?

smoogipoo: `targetMinimumHealth = 1` means the minimum health is 100%, so there's effectively no drain.

peppy: so can this effectively be written as:

```csharp
var maximumDrop = BeatmapDifficulty.DifficultyRange(beatmap.BeatmapInfo.BaseDifficulty.DrainRate, min_health_target, mid_health_target, max_health_target);

targetMinimumHealth = 1 - Math.Clamp(maximumDrop * (1 - drainLenience), 0, 1);
```

Because I would prefer that (feels much easier to read).

smoogipoo: So plug in 0.3 (`max_health_target`) with a lenience of 0 (default):

```
tMH = 1 - clamp(0.3 * (1 - 0), 0, 1)
        = 1 - 0.3
        = 0.7
```
Whereas the expected value is 0.3.

smoogipoo: Logically what it's doing is calculating the target minimum (`target`), calculating the amount of hp drained to reach that minimum (`1 - target`), and adding back a portion of that amount (`target = target + lenience * (1 - target)`).

I can add comments if you like.

peppy (approved):
comments work well enough

PR #9334, opened 2020-06-21
Extract base legacy skin transformer

bdach:
Step 3 towards addressing #8595.

- [x] Depends on #9333.

# Summary

Once I took the direction of handling layered hit sounds in legacy skin transformers, it became apparent that two rulesets (namely osu! and osu!catch) would be sharing the same non-trivial logic in `GetSample()` (which is, play non-layered samples always, and layered samples only if the skin option isn't explicitly turned off - otherwise return a virtual channel).

Up until now skin transformers had a very similar interface (by the virtue of adhering to `ISkin`), but there wasn't really much reason to unify them in any way; even though they had the same implementation of `GetTexture`, it was just proxying.

To avoid duplication when handling layered hitsounds, this PR extracts a base legacy skin transformer, to which the "standard" (2/4 rulesets) layered sample handling logic can be applied, and which any ruleset can override at its leisure (which taiko already does and mania will also do).

# Remarks

I am quite unsure of this change, code-wise. Aside from being a refactoring to a major visual part which makes me uneasy (although it *should* be fine and I proofread it several times, for however little that is worth), if I was writing this code on a personal project, I would have probably gone for composition over inheritance, but that seems non-idiomatic in the codebase here, so I'm mostly trying to follow suit.

Step 4 to add layered hit sounds looks very close to ready and likely to be the last - I've [pushed the branch with what I have right now](https://github.com/bdach/osu/compare/base-legacy-skin-transformer...bdach:layered-hit-sounds), but I'd like to give it one more evening of scrutiny before I PR it (possibly with more tests) as there is a surprising number of edge-cases to consider. I figure all three changes introduced thus far should stand on their own as to their usefulness.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/CatchSkinColourDecodingTest.cs
+++ b/osu.Game.Rulesets.Catch.Tests/CatchSkinColourDecodingTest.cs
@@ -17,7 +17,8 @@ namespace osu.Game.Rulesets.Catch.Tests
         {
             var store = new NamespacedResourceStore<byte[]>(new DllResourceStore(GetType().Assembly), "Resources/special-skin");
             var rawSkin = new TestLegacySkin(new SkinInfo { Name = "special-skin" }, store);
-            var skin = new CatchLegacySkinTransformer(rawSkin);
+            var skinSource = new SkinProvidingContainer(rawSkin);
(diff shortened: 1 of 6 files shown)


smoogipoo (approved):
I'm happy with this direction.

PR #3632, opened 2018-10-24
Fade MetadataSection back in if non-null Text is set

holly-hacker:
MetadataSections were not fading back in when assigning a normal value after it was null/blank.

- Closes #3579

---

Fixes Source and Tags not reappearing if they are once set to null.

PR #9338, opened 2020-06-22
Only fadeout beatmap metadata when it's visible and required

### LINKED ISSUE #9335: Beatmap tags section bouncing when there is no beatmap source in song select
a3IsLZhTml

**osu!lazer version:** 2020.620.0

Cyklan:
This fixes #9335.
When the title of Beatmap Details Metadata was empty, it would display the container and then fade it out, resulting in the bouncing animation.
Changing `this.FadeOut();` to `this.Hide();` was everything needed.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapDetails.cs
+++ b/osu.Game/Screens/Select/BeatmapDetails.cs
@@ -331,7 +331,7 @@ namespace osu.Game.Screens.Select
                 {
                     if (string.IsNullOrEmpty(value))
                     {
-                        this.FadeOut(transition_duration);
+                        this.Hide();
                         return;
                     }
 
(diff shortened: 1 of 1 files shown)


peppy:
This is changing behaviour when switching from a beatmap showing text to not showing text.

osu.Game/Screens/Select/BeatmapDetails.cs
```diff
@@ -331,7 +331,13 @@ public string Text
...
                     if (string.IsNullOrEmpty(value))
                     {
-                        this.FadeOut(transition_duration);
+                        if (Height > 0)
+                        {
+                            this.FadeOut(transition_duration);
+                            return;
+                        }
+
+                        Hide();
```
bdach, line 340: I don't think this is a proper fix, either. It's pretty hacky to say the least.

The core problem here is that the metadata section is initially visible while it should probably not be. I think the way this should be fixed is by setting `Alpha = 0` in the ctor of `MetadataSection` (but *not* moving it out of the nested `FillFlowContainer` - see discussion in #3632 for the reason why).

Keeping that above PR in mind (this is an ages-old regression, it seems), I'd be much more comfortable with this change in general if it had visual test cases attached.

bdach (changes requested):

PR #9340, opened 2020-06-22
Add support for layered hit sounds skin option

### LINKED ISSUE #8595: skin hitsounds overlap beatmap hitsounds

bdach:
- Closes #8595.
- Supersedes / closes #8683.
- [x] Depends on framework bump.

# Summary

This PR introduces handling of the `LayeredHitSounds` skin option.

A layered hit sound is added to each hit object that specifies a custom sample bank without having the `Normal` flag set. In osu! and osu!catch, those hit sounds play unless explicitly disabled via a skin, while in osu!mania they are never played. osu!taiko seems to have entirely different hit sound handling, so layered hit sounds seem to not apply.

As per the direction outlined in reviews of the aforementioned first attempt at making this change, the implementation leverages legacy skin transformers. The "baseline" behaviour of osu! and osu!catch is done in the base `LegacySkinTransformer`, while local changes in taiko and mania are done via overrides.

The following three videos demonstrate the behaviour after this change in all four rulesets in three scenarios:

* [on the default legacy skin](https://streamable.com/qliuub),
* [on a custom user skin with `LayeredHitSounds: 0`](https://streamable.com/sm6k40),
* [on a custom user skin with `LayeredHitSounds: 0`, but with a beatmap skin with `LayeredHitSounds: 1`](https://streamable.com/ammr66).

(The beatmap those videos were on are taken from [one of the test cases attached](https://github.com/bdach/osu/blob/8233f5fbc4e532cedc6a02b54d453ab106f5bf64/osu.Game.Tests/Resources/SampleLookups/hitobject-beatmap-custom-sample-bank.osu) - save for the last video, in which the beatmap is exactly the same as the one in the first two videos, but with the beatmap skin override added in.)

# Remarks

As far as I could see from legacy experimentation, the behaviour of being able to override the user skin's value of `LayeredHitSounds` is new to lazer. I did nothing to make it happen, it's just a byproduct of how `SkinProvidingContainer`s work; that's also why it isn't tested.

This feature is unfortunately very edge-case'y due to the nature of the functionality itself. I think I've tested the main edge cases I could imagine, but as I have zero knowledge of stable source it's not unlikely it has some surprises left in store. If you see something I could have missed please let me know.

In comparison to the PR that this aims to supersede, I moved the skin lookup enum value to `GlobalSkinConfiguration` to avoid adding a new field to `LegacySkinConfiguration` and to also leverage the existing bool-handling code path.


smoogipoo:
Stable doesn't take `LayeredHitSounds` option from the beatmap, as far as I can see. That goes for most/all skin settings.

bdach:
Right, is that something I should be looking to disable here then? Not sure what the desired behaviour is.

smoogipoo:
If we were to go by stable, everything except custom and combo colours should be blocked from falling to `base`, however since it was already unsupported in stable I think it's fine to leave it. Chances are that existing beatmaps won't contain those settings and we may want to support them going forward (haven't really discussed it at this point).

So let's leave it as is for this PR.

smoogipoo (approved):

smoogipoo:
Good stuff :+1:

PR #9349, opened 2020-06-23
Fix player not exiting immediately on Alt-F4

### LINKED ISSUE #2887: Alt F4 in-map brings up pause menu, then quit the game on second trigger

peppy:
Regressed in https://github.com/ppy/osu/pull/6992 (not quite sure what this PR was trying to achieve, maybe that will become apparently once this is merged, and we can apply a proper fix).

Closes https://github.com/ppy/osu/issues/2887.


smoogipoo (approved):
sure

PR #9350, opened 2020-06-23
Fix editor drag selection not continuing to select unless the mouse is moved

### LINKED ISSUE #9141: Selection in Editor does not update to include new objects when playing
Selection in the editor does not automatically update to include new objects when a song is playing if the cursor is standing still. The selection only updates when the cursor is moving. To match stable behaviour, the former would also have to be true. I've tested this in both the osu!standard and osu!mania editor.

Gif showing the issue: https://i.imgur.com/GQeUE9z.gifv

peppy:
Closes #9141.


smoogipoo:
1. If you pause before letting go of the mouse button, selected objects get cleared.
2. I'm not sure how correct it is that you can't deselect selected objects. Imo you should be able to move the mouse and have objects deselect - mania and taiko should behave more like the timeline where the drag box goes offscreen, but I understand that isn't a simple change.
3. If you select -> pause -> move to another point in the beatmap -> play -> select, the selection doesn't get cleared. Also don't know how correct this is...

peppy:
1. So you think it should enter a "select only" mode until release?
2. Yeah, that will come later. This is just a best-effort fix for now.
3. Will change it to clear selection on a fresh drag

smoogipoo:
1. Yeah I think that would feel better.

peppy:
Not quite simple to implement with the current structure (`BlueprintContainer` is unaware of the drag state and making it aware isn't trivial). Will need to think on this.

smoogipoo (approved):

PR #9361, opened 2020-06-24
Implement an import beatmap confirmation dialog

### LINKED ISSUE #9357: Make it clearer for the user that importing beatmaps from stable will create copies of the beatmaps
I realise this is probably something a lot of people take for granted but it would be a nice quality-of-life improvement. 

**Describe the new feature:** 
Make it clearer for the user that importing beatmaps from stable will create new copies of the beatmaps.

A popup explaining the effect and consequences of importing beatmaps from Stable and that it may take a long time depending on the number of beatmaps.

vruss:
Temporarily resolves #9357 until an interactive user-welcome screen in implemented.

The text and font might not be up to standard.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Maintenance/GeneralSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Maintenance/GeneralSettings.cs
@@ -35,8 +35,11 @@ namespace osu.Game.Overlays.Settings.Sections.Maintenance
                     Text = "Import beatmaps from stable",
                     Action = () =>
                     {
-                        importBeatmapsButton.Enabled.Value = false;
-                        beatmaps.ImportFromStableAsync().ContinueWith(t => Schedule(() => importBeatmapsButton.Enabled.Value = true));
+                        dialogOverlay?.Push(new ImportBeatmapsConfirmationDialog(() =>
+                        {
+                            importBeatmapsButton.Enabled.Value = false;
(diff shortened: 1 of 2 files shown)


peppy:
I'm *really* not sure this is required...

bdach:
I also think that the fact import creates copies is implied on its own, but I can understand people getting confused about that.

I mostly floated the idea of having a confirm dialog as a loose idea which wouldn't take much effort - I did not intend to suggest it as a definitive direction.

vruss:
No worries. I thought I would implement this as a quick temporary solution but if this isn't the right way to go then feel free to close this PR.

peppy:
Also I don't believe this covers the initial prompt for importing, which may be a place we can amicably add an addition about the operation being a copy (see `ImportFromStablePopup`)

vruss:
Would you be ok with the BodyText of `ImportFromStablePopup` is expanded to include:

```C#
BodyText = "An existing copy of osu! was found, though.\nWould you like to import your beatmaps, skins and scores?\nThis will create new copies of all your assets.";
```

Or do I close this PR?

peppy:
"This will create a second copy of all files on disk."

And I think remove the confirmation dialog from settings (people hitting that option likely already know the consequences. Might be best to open a new PR since the existing changes would be rolled back anyway.

vruss:
Ok I will do so!

PR #9362, opened 2020-06-24
Standardise all tournament tests to use placeholder data

peppy:
This stops some tests from relying on DI components provided by `TournamentGameBase`. Is a prerequisite for moving the bracket reading logic into `TournamentGame` to support a better UX flow when starting the client.


bdach (approved):
seems good to me :+1:

smoogipoo (approved):

PR #9363, opened 2020-06-24
Add text explaining a second copy will be made on importing beatmaps 

### LINKED ISSUE #9357: Make it clearer for the user that importing beatmaps from stable will create copies of the beatmaps

vruss:
Resolves #9357 

osu_2020-06-24_11-33-24

### DIFF
--- a/osu.Game/Screens/Select/ImportFromStablePopup.cs
+++ b/osu.Game/Screens/Select/ImportFromStablePopup.cs
@@ -12,7 +12,7 @@ namespace osu.Game.Screens.Select
         public ImportFromStablePopup(Action importFromStable)
         {
             HeaderText = @"You have no beatmaps!";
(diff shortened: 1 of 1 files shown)


power9maker:
Similar notification should probably also be added to "Import beatmaps" and "Import skins" buttons in settings
image

peppy:
As per the closed issue, it's intentionally *not* being added there. Import means copy and using the maintenance buttons implies a user knows what they are doing already.

peppy (approved):

PR #9364, opened 2020-06-24
Increase coordinate parsing limits

### LINKED ISSUE #9342: Slider length parsing limit is too low
for example [this map](https://osu.ppy.sh/beatmapsets/944317#fruits/1972149) uses a super long slider to map the chorus. you get an OverflowException trying to import it because its length is over 65536

not sure what limit stable has for this (if any), but that should probably be used instead.

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9342

Technically, we could forego these limits now that we have cancellable beatmap loads, but stable will need the same limit.

For @peppy , probably need to recompute SR of https://osu.ppy.sh/beatmapsets/944317#fruits/1972149 (from 5.89* to 6.04*).

### DIFF
--- a/osu.Game/Beatmaps/Formats/Parsing.cs
+++ b/osu.Game/Beatmaps/Formats/Parsing.cs
@@ -11,7 +11,7 @@ namespace osu.Game.Beatmaps.Formats
     /// </summary>
     public static class Parsing
     {
-        public const int MAX_COORDINATE_VALUE = 65536;
+        public const int MAX_COORDINATE_VALUE = 131072;
 
         public const double MAX_PARSE_VALUE = int.MaxValue;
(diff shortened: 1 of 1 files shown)


bdach (approved):
Slider length on affected map is ~76400 so this leaves plenty of headspace still.

I assume we're not really interested in fixing aspire in any way at this point, and this is an exception since the map is ranked?

peppy:
Reprocessing queued for 944317.

peppy:
Oh, I guess this also needs a diffcalc update. Will look into that.

smoogipoo:
@bdach Yeah that's correct, just an exception since it's a ranked map.

PR #9365, opened 2020-06-24
Add right click context menu to beatmap link

### LINKED ISSUE #8894: Add right click context menu to beatmap links to allow immediate presenting at song select
**Describe the new feature:** 

Currently clicking "now playing <map>" does open osu!direct. in stable clicking on this leads you directly to the map as long youre in song select. i feel like this should behave the same way in lazer.

Send us directly to the map if you are in song select instead of opening osu!direct

kloczewiak:
Closes #8894 
If you don't have the beatmap/beatmapset context menu will only show "Details" option.
It's quite messy, so I'd like some advice how to clean this up.

### DIFF
--- a/osu.Game/Graphics/Containers/LinkFlowContainer.cs
+++ b/osu.Game/Graphics/Containers/LinkFlowContainer.cs
@@ -67,8 +67,7 @@ namespace osu.Game.Graphics.Containers
 
         private void createLink(IEnumerable<Drawable> drawables, LinkDetails link, string tooltipText, Action action = null)
         {
-            DrawableLinkCompiler context;
-            AddInternal(context = new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList())
+            AddInternal(new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList())
             {
                 RelativeSizeAxes = Axes.Both,
                 TooltipText = tooltipText,
--- a/osu.Game/Graphics/Containers/LinkFlowContainer.cs
+++ b/osu.Game/Graphics/Containers/LinkFlowContainer.cs
@@ -67,11 +67,10 @@ namespace osu.Game.Graphics.Containers
 
         private void createLink(IEnumerable<Drawable> drawables, LinkDetails link, string tooltipText, Action action = null)
         {
-            AddInternal(new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList())
+            AddInternal(new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList(), link)
             {
                 RelativeSizeAxes = Axes.Both,
                 TooltipText = tooltipText,
-                Link = link,
                 Action = () =>
                 {
                     if (action != null)
--- a/osu.Game/Graphics/Containers/LinkFlowContainer.cs
+++ b/osu.Game/Graphics/Containers/LinkFlowContainer.cs
@@ -72,7 +72,6 @@ namespace osu.Game.Graphics.Containers
                 RelativeSizeAxes = Axes.Both,
                 TooltipText = tooltipText,
                 Link = link,
-                Game = game,
                 Action = () =>
                 {
                     if (action != null)
--- a/osu.Game/Graphics/Containers/LinkFlowContainer.cs
+++ b/osu.Game/Graphics/Containers/LinkFlowContainer.cs
@@ -67,10 +67,13 @@ namespace osu.Game.Graphics.Containers
 
         private void createLink(IEnumerable<Drawable> drawables, LinkDetails link, string tooltipText, Action action = null)
         {
-            AddInternal(new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList())
+            DrawableLinkCompiler context;
+            AddInternal(context = new DrawableLinkCompiler(drawables.OfType<SpriteText>().ToList())
             {
                 RelativeSizeAxes = Axes.Both,
                 TooltipText = tooltipText,
+                Link = link,
+                Game = game,
                 Action = () =>
                 {
                     if (action != null)
--- a/osu.Game/Online/Chat/DrawableLinkCompiler.cs
+++ b/osu.Game/Online/Chat/DrawableLinkCompiler.cs
@@ -26,9 +26,11 @@ namespace osu.Game.Online.Chat
         /// </summary>
         public List<Drawable> Parts;
 
-        private OsuGame game;
+        [Resolved(CanBeNull = true)]
+        private OsuGame game { get; set; }
 
-        private BeatmapManager beatmapManager;
+        [Resolved]
+        private BeatmapManager beatmapManager { get; set; }
 
         public LinkDetails Link;
 
@@ -41,7 +43,7 @@ namespace osu.Game.Online.Chat
             Parts = parts.ToList();
         }
 
-        [BackgroundDependencyLoader]
+        [BackgroundDependencyLoader(true)]
         private void load(OsuColour colours, OsuGame game, BeatmapManager beatmapManager)
         {
             IdleColour = colours.Blue;
--- a/osu.Game/Online/Chat/DrawableLinkCompiler.cs
+++ b/osu.Game/Online/Chat/DrawableLinkCompiler.cs
@@ -8,6 +8,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Cursor;
 using osu.Framework.Graphics.UserInterface;
+using osu.Game.Beatmaps;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.UserInterface;
@@ -27,6 +28,8 @@ namespace osu.Game.Online.Chat
 
         private OsuGame game;
 
+        private BeatmapManager beatmapManager;
+
         public LinkDetails Link;
 
         public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) => Parts.Any(d => d.ReceivePositionalInputAt(screenSpacePos));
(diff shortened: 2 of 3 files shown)


osu.Game/OsuGame.cs
```diff
@@ -313,6 +315,46 @@ private void load()
...
 
+        public void SelectBeatmap(int beatmapId)
+        {
+            var map = GetBeatmapFromId(beatmapId);
+            bool findPredicate(BeatmapInfo b) => b.OnlineBeatmapID == beatmapId;
+
+            PresentBeatmap(map.BeatmapSet, findPredicate);
+        }
+
+        public void SelectBeatmapSet(int setId)
```
Game4all, line 326: All the following methods you added seem to be used at a unique place, you may want to inline them instead as these are all fairly short

osu.Game/Graphics/Containers/LinkFlowContainer.cs
```diff
@@ -71,6 +71,8 @@ private void createLink(IEnumerable<Drawable> drawables, LinkDetails link, strin
             {
                 RelativeSizeAxes = Axes.Both,
                 TooltipText = tooltipText,
+                Link = link,
+                Game = game,
```
bdach, line 75: This should not be passed around like this.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -39,6 +46,64 @@ private void load(OsuColour colours)
...
+
+                            if (Game?.GetBeatmapFromId(beatmapId) != null)
+                                goToBeatmap = new OsuMenuItem("Go To Beatmap", MenuItemType.Highlighted, () => Game?.SelectBeatmap(beatmapId));
+                            else
+                            {
+                                goToBeatmap = new OsuMenuItem("Go To Beatmap", MenuItemType.Highlighted);
+                                goToBeatmap.Action.Disabled = true;
+                            }
+
+                            items.Add(goToBeatmap);
```
bdach, line 73: this is quite bad:

* `GetBeatmapIdFromLink` is called twice for no reason instead of storing to a local
* `GetBeatmapFromId` shouldn't be a thing (why not get the entire beatmap by id to begin with? and if it's null then it doesn't exist locally?)
* unnecessary branch instead of doing something like
```csharp
new OsuMenuItem("Go to beatmap", MenuItemType.Highlighted)
{
    Action =
    {
        Value = () => Game?.PresentBeatmap(beatmap),
        Disabled = beatmap == null
    }
};
```

exact same review applies to the beatmap set option

osu.Game/OsuGame.cs
```diff
@@ -313,6 +309,36 @@ private void load()
...
 
+        /// <summary>
+        /// Show a beatmap set as an overlay.
+        /// </summary>
+        /// <param name="setId">The set to display.</param>
+        public void ShowBeatmapSet(int setId) => waitForReady(() => beatmapSetOverlay, _ => beatmapSetOverlay.FetchAndShowBeatmapSet(setId));
+
+        public void SelectBeatmap(int beatmapId) => PresentBeatmap(GetBeatmapFromId(beatmapId).BeatmapSet, b => b.OnlineBeatmapID == beatmapId);
+
+        public void SelectBeatmapSet(int setId) => PresentBeatmap(GetBeatmapSetFromId(setId).BeatmapSet);
```
bdach, line 320: these should not exist at all

osu.Game/OsuGame.cs
```diff
@@ -313,6 +309,36 @@ private void load()
...
+
+        public BeatmapInfo GetBeatmapFromId(int beatmapId) => BeatmapManager.QueryBeatmap(b => b.OnlineBeatmapID == beatmapId);
+
+        public BeatmapInfo GetBeatmapSetFromId(int setId) => BeatmapManager.QueryBeatmap(b => b.BeatmapSet.OnlineBeatmapSetID == setId);
```
bdach, line 324: these should not exist here (if at all).

please check other usages of `BeatmapManager` for pointers on how this is supposed to be done.

osu.Game/OsuGame.cs
```diff
@@ -313,6 +309,36 @@ private void load()
...
+            return 0;
+        }
+
+        public int GetBeatmapSetIdFromLink(LinkDetails link)
+        {
+            if (int.TryParse(link.Argument, out int setId))
+                return setId;
+
+            return 0;
+        }
```
bdach, line 340: these also have no reason to be in `OsuGame`. it should be the responsibility of a link to provide this data, and I'd say we're nearing the point in which we might want to have an actual type hierarchy of `Link`s so that we can properly read these as plain old properties instead of parsing strings.

if that is the direction we are to take things then that is a separate preparatory refactoring PR.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,13 +18,17 @@ namespace osu.Game.Online.Chat
...
+    public class DrawableLinkCompiler : OsuHoverContainer, IHasContextMenu
     {
         /// <summary>
         /// Each word part of a chat link (split for word-wrap support).
         /// </summary>
         public List<Drawable> Parts;
 
+        public OsuGame Game;
+
+        public LinkDetails Link;
```
bdach, line 30: public fields are generally a no-go here, `Parts` is more an exception than a rule. `OsuGame` should be DI'd, and `LinkDetails` should be `{ get; private set; }` if exposition is deemed absolutely necessary

bdach (changes requested):
You seem to be unaware of either the concept of dependency injection itself, or [our implementation thereof in the framework](https://github.com/ppy/osu-framework/wiki/Dependency-Injection). Please familiarise yourself with the concept before proceeding further.

power9maker:
I still can't navigate git... Sorry for these pushes ^

power9maker:
I wanted to make this shorter but I can't seem to solve one thing, why doesn't it let me do something like that:
```
items.Add(new OsuMenuItem("Details", MenuItemType.Standard)
{
    Action =
    {
        Value = link.Action == LinkAction.OpenBeatmap ? () => game?.ShowBeatmap(getBeatmapIdFromLink(link).Id) : () => game?.ShowBeatmapSet(getBeatmapIdFromLink(link).Id)
    }
});
```
Am I doing something wrong or is what I'm trying to do impossible?

bdach:
the compiler can't figure out what delegate type is supposed to be returned from the ternary; if you cast one of the lambdas to `Action` it will compile. anyways, that is mostly moot as that isn't good code and would likely be solved by one of my earlier suggestions of cleaning up the link class hierarchy properly.

smoogipoo:
I'm not sure how you tested this but it doesn't seem to be working. In chat the menu item is disabled:
image
In user profiles the context menu doesn't pop up at all (e.g. recent section).

Etc...

smoogipoo:
Re: My first issue, it was just because I hadn't downloaded the beatmap. It probably shouldn't show at all.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,133 @@ namespace osu.Game.Online.Chat
...
+                BeatmapSet
+            }
+        }
+
+        public MenuItem[] ContextMenuItems
+        {
+            get
+            {
+                List<MenuItem> items = new List<MenuItem>();
+                var map = getBeatmapFromLink(link);
```
smoogipoo, line 99: The way this is done is super weird to me. Not all links are going to be beatmaps.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,133 @@ namespace osu.Game.Online.Chat
...
+        {
+            if (link.Action == LinkAction.OpenBeatmap && link.Argument != null && int.TryParse(link.Argument.Contains('?') ? link.Argument.Split('?')[0] : link.Argument, out int beatmapId))
+                return new StoreId(beatmapId, StoreId.IdType.Beatmap);
+            if (link.Action == LinkAction.OpenBeatmapSet && int.TryParse(link.Argument, out int setId))
+                return new StoreId(setId, StoreId.IdType.BeatmapSet);
+
+            return new StoreId(0, StoreId.IdType.Beatmap);
+        }
+
+        private class StoreId
```
smoogipoo, line 76: This class and the two methods above it are basically completely unnecessary. Why is there a "type" when you know the type of the link?

I'd put this all in the one switch below, or MAYBE lay this out as separate methods: `createBeatmapContextMenuItems()`, `createBeatmapSetContextMenuItems()`, `createProfileContextMenuItems()`, etc, though that may be over-engineered.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,133 @@ namespace osu.Game.Online.Chat
...
+            var id = getBeatmapIdFromLink(link);
+            if (id.Type == StoreId.IdType.Beatmap)
+                return beatmapManager.QueryBeatmap(b => b.OnlineBeatmapID == id.Id);
+            else
+                return beatmapManager.QueryBeatmap(b => b.BeatmapSet.OnlineBeatmapSetID == id.Id);
+        }
+
+        private StoreId getBeatmapIdFromLink(LinkDetails link)
+        {
+            if (link.Action == LinkAction.OpenBeatmap && link.Argument != null && int.TryParse(link.Argument.Contains('?') ? link.Argument.Split('?')[0] : link.Argument, out int beatmapId))
```
smoogipoo, line 68: This is really hard to parse. Please split the `int.TryParse()` part into a separate line.

smoogipoo (changes requested):

power9maker:
> In user profiles the context menu doesn't pop up at all (e.g. recent section).

Okay this is weird, I went through the code and it seems like at least recent section's links should be working fine, because they use `DrawableLinkCompiler` just like chat links. I even added a test context menu that would be added to EVERY link and it worked in chat but didn't in user profile's recent section. 
Recent section's profile link also uses `ToolipText` which should say "view profile", but it also doesn't appear.

smoogipoo:
Ah, we don't currently have a global context menu so it has to be added per-component and the profile doesn't have one. Fine to ignore that issue.

power9maker:
I wanted to cleanup this code, so on line 117 in `DrawableLinkCompiler` I replaced `game?.ShowBeatmap(getBeatmapIdFromLink(link).Id` with `game?.ShowBeatmap(map.OnlineBeatmapID ?? 0`, but InspectCode says that there's possible `System.NullReferenceException` even though I'm checking for null with `?? 0`.
I even manually set `map.OnlineBeatmapID` to `null` and it didn't throw an exception.
Am I doing something wrong again or is InspectCode wrong?

bdach:
it's intensely hard to tell without actually seeing the code you've wrote, but i'm guessing inspectcode is reporting that `map` *itself* might be null.

power9maker:
Ah, yes, sorry. I'm blind...

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,109 @@ namespace osu.Game.Online.Chat
...
+            return null;
+        }
+
+        private int getBeatmapIdFromLink(LinkDetails link)
+        {
+            if (int.TryParse(link.Argument.Contains('?') ? link.Argument.Split('?')[0] : link.Argument, out int beatmapId))
+                return beatmapId;
+
+            return 0;
+        }
```
bdach, line 81: i find this very weird. don't represent invalid data in-band like this by coercing to 0, use an `int?` instead.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,109 @@ namespace osu.Game.Online.Chat
...
+            int setId = getBeatmapSetIdFromLink(link);
+            if (setId != 0)
+                return beatmapManager.QueryBeatmap(b => b.BeatmapSet.OnlineBeatmapSetID == setId)?.BeatmapSet;
+
+            return null;
+        }
+
+        private int getBeatmapIdFromLink(LinkDetails link)
+        {
+            if (int.TryParse(link.Argument.Contains('?') ? link.Argument.Split('?')[0] : link.Argument, out int beatmapId))
```
bdach, line 77: So I'll elaborate on the problem I have with the way this is done. In my view it's unorganised - this is doing link parsing on-the-fly where presumably some places have already *done* such parsing and/or might want to do that parsing on their own again.

What I'd like to see here is introduction of a `BeatmapLink` or a `BeatmapSetLink` model (so *only* and *strictly* data, ideally immutable, too) that is already parsed *once*, at a *higher* level, which has the beatmap ID stored into a sane structure that can just be read here without doing this split & parse locally.

osu.Game/Online/Chat/DrawableLinkCompiler.cs
```diff
@@ -15,30 +19,109 @@ namespace osu.Game.Online.Chat
...
+
+                        items.Add(new OsuMenuItem("Details", MenuItemType.Standard, () => game?.ShowBeatmapSet(getBeatmapSetIdFromLink(link))));
+
+                        return items.ToArray();
+
+                    default:
+                        return Array.Empty<MenuItem>();
+                }
+            }
+        }
```
bdach, line 123: And this is problem number 2 in this for me. Going forward I presume we might want to handle news links, wiki links, etc. `DrawableLinkCompiler` is a very general class. I would be strongly opposed to all logic handling dispatch of links to game views landing in this huge getter in the future.

The way I would propose this to be resolved is by having analogues to the link types (so i.e. `DrawableBeatmapLink` and `DrawableBeatmapSetLink`). Each of these will have their own logic for context menu, their own options, etc., all scoped down and not everything-at-once. I don't see this as over-engineering as we know we might need this in the future already.

bdach (changes requested):

power9maker:
I just came back home 2 days ago and tried working on this, but I don't think I'm experienced enough to finish this.
I tried to at least work on [this](https://github.com/ppy/osu/pull/9365#discussion_r449182467), but for some reason even though `DrawableLinkCompiler`'s width is equal to whole chat line width, the context menu appears only when you right click the link. So when I added a `DrawableBeatmapLink` child (with all right click logic) to `DrawableLinkCompiler` with `RelativeSizeAxes = Axes.Both` the context menu would appear whenever I right click anywhere on the chat line, instead of when I right click the link.

bdach:
I am going to go ahead and close this pull request due to inactivity, generally incorrect direction, and the author's stated inability to resolve the issues. `DrawableLinkCompiler` is generally the wrong place for link parsing logic - it should solely be a presentation-focused class. The parsing should happen at a higher level.

Another way that may be worth exploring is having the dropdowns and their items associated with `Link`/`LinkDetails` directly somehow - although the current structure of those classes has been called into question before (https://github.com/ppy/osu/issues/11754), so cleaning up that area first may be a wise step.

PR #9369, opened 2020-06-25
Fix colours with 0 alpha being invisible in legacy skins

### LINKED ISSUE #9306: Hyperfruits are unreadable when using skins
**Describe the bug:** 
In lazer hyperfruits are different then in stable. Most notably the (red) glow on the outside has changed into a completely red fruit.
Since the hyperfruit is a different colour (=essentially a different object), it is harder to read, especially on dense hyper-sections (since this means red glowy circle-spam).
I do not know how far skin implementation is in lazer, but currently using ctb skins does not show hypers, presumably because the red glow isn't there anymore?
Although the red glow is actually there on the default stable skin, so this contradicts that statement and supports this being a bug.

Either way, myself and essentially all other players, use a variant of the square or circle skins shown below.
While not everyone uses full white fruits, red glows on hyperdashes are used everywhere (as far as I am aware).
Because of this, I see this as a bug, but I guess it is in some way a feature request to make it possible to only have the red-glowy outline on hyperfruits.

Also, while I can see some players using this new fruit as hyperfruits, more experienced players would prefer the red-glow instead of the red fruits, as this prevents them from reading hyperfruits as they did before and I/we are used to the red outline.

*Note: while it is possible to play without hyperfruits, it is significantly harder to do so.

Lazer standard skin:
hypernoskin
Lazer + Square skin:
hypersquareskin
Lazer + Circle skin:
hypercircleskin
Stable + Square skin:
hypernormal

**osu!lazer version:** 
2020.616.0

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9306

For some reason stable ignores the alpha of colours with `A <= 0` and `A >= 255`. ¯\\_(ツ)_/¯

### DIFF
--- a/osu.Game/Skinning/LegacySkin.cs
+++ b/osu.Game/Skinning/LegacySkin.cs
@@ -275,7 +275,7 @@ namespace osu.Game.Skinning
             if (!source.CustomColours.TryGetValue(lookup, out var col))
                 return null;
 
-            if (col.A == 0)
+            if (col.A <= 0 || col.A >= 255)
                 col.A = 1;
 
             return new Bindable<Color4>(col);
--- a/osu.Game/Skinning/LegacySkin.cs
+++ b/osu.Game/Skinning/LegacySkin.cs
@@ -275,7 +275,7 @@ namespace osu.Game.Skinning
             if (!source.CustomColours.TryGetValue(lookup, out var col))
                 return null;
 
-            if (col.A <= 0 || col.A >= 255)
+            if (col.A <= 0 || col.A >= 1)
                 col.A = 1;
 
             return new Bindable<Color4>(col);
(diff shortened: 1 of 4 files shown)


peppy:
Not sure about the "some reason" part, sounds quite logical? I guess you could argue it should be clamped instead, but if a user specifies -1 i'm not sure they meant 0..

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -271,7 +271,15 @@ private IBindable<TValue> lookupForMania<TValue>(LegacyManiaSkinConfigurationLoo
         }
 
         private IBindable<Color4> getCustomColour(IHasCustomColours source, string lookup)
-            => source.CustomColours.TryGetValue(lookup, out var col) ? new Bindable<Color4>(col) : null;
+        {
+            if (!source.CustomColours.TryGetValue(lookup, out var col))
+                return null;
+
+            if (col.A <= 0 || col.A >= 1)
```
peppy, line 278: can this be done at parsing time instead of lookup time?

PR #9370, opened 2020-06-25
Restore key repeat functionality to the song select carousel

### LINKED ISSUE #8379: Holding the "up" or "down" key will not scroll the map
But "right" and "left" work as usual. Why can’t you move around inside a map that has many difficulties?
https://youtu.be/oeSzcD7VGLo
Here you can add a suggestion for setting the scroll speed (map per second)

osu!lazer version:
2020.320.1

peppy:
Was lost when custom binding support was added.

Closes https://github.com/ppy/osu/issues/8379

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -279,9 +279,6 @@ namespace osu.Game.Screens.Select
         /// <param name="skipDifficulties">Whether to skip individual difficulties and only increment over full groups.</param>
         public void SelectNext(int direction = 1, bool skipDifficulties = true)
         {
-            if (selectedBeatmap == null)
-                return;
-
             if (beatmapSets.All(s => s.Filtered.Value))
                 return;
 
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -486,8 +503,46 @@ public bool OnPressed(GlobalAction action)
...
+        /// Begin repeating the specified selection action.
+        /// </summary>
+        /// <param name="action">The action to perform.</param>
+        /// <param name="source">The source of the action. Used in conjunction with <see cref="endRepeatSelection"/> to only cancel the correct action (most recently pressed key).</param>
+        private void beginRepeatSelection(Action action, object source)
+        {
+            endRepeatSelection();
+
+            lastRepeatSource = source;
+            Scheduler.Add(repeatDelegate = new ScheduledDelegate(action, Time.Current, repeat_interval));
```
smoogipoo, line 530: See: `DrawableExtensions.BeginKeyRepeat()`.

smoogipoo (changes requested):

osu.Game/Screens/Select/BeatmapCarousel.cs
peppy: wow, that exists. will have to check how the delays feel.

smoogipoo (dismissed):

smoogipoo:
Please look through my changes and see if you agree. Otherwise, lgtm.

PR #9374, opened 2020-06-26
Update UI of KudosuHistory items in UserOverlay

### LINKED ISSUE #7824: Recent activities should share it's drawable logic with kudosu history in UserOverlay
As seen in web these elements are very similar, however in lazer they are totally separate.

EVAST9919:
Closes https://github.com/ppy/osu/issues/7824

Before:
Screenshot_3
After:
Screenshot_1
Web:
Screenshot_2


peppy:
The new logic isn't going to work with localisation. Might want to check how osu-web is localising it and copy roughly the same formatting.

EVAST9919:
Should be better now

osu.Game/Overlays/Profile/Sections/DrawableHistoryItem.cs
```diff
@@ -0,0 +1,150 @@
...
+            }
+
+            var variableIndex = s.IndexOf(':');
+
+            AddText(s.Substring(0, variableIndex));
+
+            var trimmed = handleVariable(s.Substring(variableIndex + 1));
+
+            handleString(trimmed);
+        }
```
bdach, line 115: I don't think what the suggestion meant to convey is to literally copy the [localisation format from laravel](https://laravel.com/docs/7.x/localization)...

First off, *even if* this is the format that we want to be using, this code should not live here, it should be in a more general place. Secondly, I'd argue we should aim for a format which is more idiomatic to C# (something akin to `FormattableString`, but for drawables). Unfortunately I think rolling our own parsing would be unavoidable either way since `string.Format` seems inextensible to other types.

Seeing that the above needs a more general (maybe even framework-side) mechanism, I'd be for holding this change off entirely until that is dealt with.

bdach (changes requested):

EVAST9919:
I refactored the code to use curly brackets in the string and removed weird Actions logic

osu.Game/Overlays/Profile/Sections/DrawableHistoryItem.cs
```diff
@@ -99,33 +97,19 @@ private void handleString(string s)
...
-                if (s.StartsWith(name))
-                {
-                    action.Invoke();
-                    return s.Substring(name.Length);
-                }
-            }
-
-            return s;
+            HandleVariable(variableName);
+            handleString(s.Substring(variableEndIndex + 1));
```
bdach, line 112: Why recursion? This can be done iteratively. (And without creating string copies, no less.)

osu.Game/Overlays/Profile/Sections/Recent/DrawableRecentActivity.cs
```diff
@@ -76,64 +76,93 @@ protected override string GetString()
...
-            ("previousUsername", () => AddText($"{Item.User?.PreviousUsername}")),
-            ("rank", () => AddText($"{Item.Rank}")),
-            ("count", () => AddText($"{Item.Count}")),
-            ("mode", () => AddText($"{getRulesetName()}")),
-            ("approval", () => AddText($"{Item.Approval.ToString().ToLowerInvariant()}"))
-        };
+            switch (name)
+            {
+                case "achievement":
+                    AddText($"{Item.Achievement.Name}");
```
bdach, line 130: Why string-interpolate a single literal?

bdach (changes requested):
You still haven't addressed the fact that this formatting mechanism should ideally be more general (*if* we go in that direction) and not just a thing residing in user profile classes.

bdach:
I am going to go ahead and close this pull request, due to inactivity and generally incorrect direction. The user overlay should not contain local text formatting logic - they should reside in the framework (https://github.com/ppy/osu-framework/issues/4530).

Given that initial localisation support has now been added and localisation/formatting of complex texts like here will be a larger focus going forward, said framework amendments may happen sooner than later and are (in my opinion, anyway) the correct way to proceed with these sorts of changes.

PR #9375, opened 2020-06-26
Hide red tint when "Show health display even when you can't fail" option is off

### LINKED ISSUE #8789: red tint layer displays even though no fail is enabled

kloczewiak:
Closes #8789 

Kept in mind what @bdach said in #8872 
Fixed tests and added new one for this new behaviour


osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -74,7 +75,8 @@ private void load(OsuColour color, OsuConfigManager config)
             boxes.Colour = color.Red;
 
             configEnabled = config.GetBindable<bool>(OsuSetting.FadePlayfieldWhenHealthLow);
-            enabled.BindValueChanged(e => this.FadeTo(e.NewValue ? 1 : 0, fade_time, Easing.OutQuint), true);
+            enabled.BindValueChanged(e => TryToFade(fade_time, Easing.OutQuint, e.NewValue), true);
+            HUDEnabled.BindValueChanged(e => TryToFade(fade_time, Easing.OutQuint, e.NewValue), true);
```
frenzibyte, line 79: Why introduce another bindable, that only convoluted logic, shouldn't you only hide the layer in the `HUDOverlay.ShowHealthbar` bindable changes, and add an internal `Container` layer to hide the layer when `enabled` is false.

frenzibyte (changes requested):
As per above, really convoluted and adds unnecessary logic to it.

osu.Game/Screens/Play/HUD/FailingLayer.cs
frenzibyte: @bdach Are you sure about having the visibility logic change internally here, I think it shouldn't.

bdach: I'd rather have it all in one place myself, rather than half in `Player` and half here.

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -31,6 +31,7 @@ public class FailingLayer : HealthDisplay
         /// </summary>
         public double LowHealthThreshold = 0.20f;
 
+        public readonly Bindable<bool> HUDEnabled = new Bindable<bool>();
```
bdach, line 34: Unnecessary. Inject the setting bindable the same way `Player` does.

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -105,6 +107,28 @@ private void updateBindings()
...
+        /// <param name="easing">Type of easing</param>
+        /// <param name="fadeIn">True when you want to fade in, false when you want to fade out</param>
+        public void TryToFade(float fadeDuration, Easing easing, bool fadeIn)
+        {
+            if (HUDEnabled.Value)
+            {
+                if (fadeIn)
+                {
+                    if (enabled.Value)
+                        this.FadeIn(fadeDuration, easing);
```
bdach, line 123: You cannot nest three conditional statements. This is unreadable.

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -105,6 +107,28 @@ private void updateBindings()
...
+        /// <param name="fadeDuration">Duration of the fade</param>
+        /// <param name="easing">Type of easing</param>
+        /// <param name="fadeIn">True when you want to fade in, false when you want to fade out</param>
+        public void TryToFade(float fadeDuration, Easing easing, bool fadeIn)
+        {
+            if (HUDEnabled.Value)
+            {
+                if (fadeIn)
+                {
+                    if (enabled.Value)
```
bdach, line 122: `enabled` should be renamed for legibility to be able to tell what setting toggles what.

bdach (dismissed):

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -105,6 +107,28 @@ private void updateBindings()
                 enabled.Value = false;
         }
 
+        /// <summary>
+        /// Tries to fade based on "Fade playfield when health is low" setting
+        /// </summary>
+        /// <param name="fadeDuration">Duration of the fade</param>
+        /// <param name="easing">Type of easing</param>
+        /// <param name="fadeIn">True when you want to fade in, false when you want to fade out</param>
+        public void TryToFade(float fadeDuration, Easing easing, bool fadeIn)
```
bdach, line 116: oh, and the `fadeIn` argument is entirely redundant since you're checking value of both bindables inside this method anyway.

osu.Game/Screens/Play/HUD/FailingLayer.cs
power9maker: Totally missed the fact that all of these do the same thing on `else`

osu.Game/Screens/Play/HUD/FailingLayer.cs
power9maker: Huh? `HUDOverlay` [does it the same way](https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Mods/ModBlockFail.cs#L26-L29) I did here.
Or maybe I just didn't get something.

osu.Game/Screens/Play/HUD/FailingLayer.cs
power9maker: So do I keep it as it is or not?

frenzibyte: @bdach Fair enough. @power9maker No, remove it and just put a check inside `enabled.BindValueChanged(...)` to control the visibility based on "show health display ..." setting and new value of `enabled`.

osu.Game/Screens/Play/HUD/FailingLayer.cs
bdach: you can get the setting bindable by doing
```csharp
config.GetBindable<bool>(OsuSetting.ShowHealthDisplayWhenCantFail);
```
which is a few lines above the lines you linked. there is [an analogous usage is in this same file](https://github.com/ppy/osu/blob/798e8e7a8deea5d1ac665bc9491604c0f082e5ed/osu.Game/Screens/Play/HUD/FailingLayer.cs#L77).

power9maker: I don't think I can, because this bindable isn't just the settings bindable, it's also controlled based on whether or not you have `No Fail` mod ON.

osu.Game/Screens/Play/HUD/FailingLayer.cs
power9maker: Well yes, but I would have to add `TryToFade()` to `HUDOverlay`.
Is that what you want?

bdach:
I've fixed this up myself. While I was wrong on the setting part (there's another level of indirection here, as the "show health display even when you can't fail" is actually applied by mods, so reading the value of the setting directly does not help here), I think you'll find that the code is now more readable.

The test cases are weak, but doing a proper one is a bit more involved (would have to be in `HUDOverlay` with a health processor provided to be substantive). I can write one on request.

power9maker:
Oh, so I wasn't wrong [here](https://github.com/ppy/osu/pull/9375#discussion_r446272343)
I thought you were right and this bindable was bound directly to the setting and was controlled by mods somewhere in the middle.

osu.Game/Screens/Play/HUD/FailingLayer.cs
```diff
@@ -96,18 +102,24 @@ private void updateBindings()
...
-            enabled.UnbindBindings();
+            fadePlayfieldWhenHealthLow.UnbindBindings();
 
             // Don't display ever if the ruleset is not using a draining health display.
             if (healthProcessor is DrainingHealthProcessor)
-                enabled.BindTo(configEnabled);
+                fadePlayfieldWhenHealthLow.BindTo(fadePlayfieldWhenHealthLowSetting);
             else
-                enabled.Value = false;
+                fadePlayfieldWhenHealthLow.Value = false;
```
frenzibyte, line 111: Sounds like this is just adding unnecessary complexity to the visibility logic now, why not just add to the check inside `updateState` and call that method on `BindHealthProcessor` as well?

frenzibyte (changes requested):

osu.Game/Screens/Play/HUD/FailingLayer.cs
bdach: I fail to see the gain, honestly. It would amount to unnecessarily conflating ruleset-specific logic with visibility logic in my view.

frenzibyte: for readability, having both `fadePlayfieldWhenHealthLow` and `fadePlayfieldWhenHealthLowSetting` together really makes thing unnecessarily complex, you can get rid of that and get rid of the whole `updateBindings()` method, and then very simply add `healthProcessor is DrainingHealthProcessor` check to `updateState()` in replace.

bdach: I thought it would look worse than it actually does. Combining three `bool`s has me a little nervous but it passes tests and seems to work ok, so sure.

frenzibyte (approved):
Sounds good and straightforward now. 👍

smoogipoo (approved):
Seems fine to me

PR #9377, opened 2020-06-26
Fix crash due to unsafe mod deserialisation

smoogipoo:


peppy (dismissed):

smoogipoo:
As requested, I've added some global error handling for cases like these (where the room fails to update), with the caveat that it only handles the listing poll request, not the request that polls for the currently-joined room.

osu.Game/Screens/Multi/RoomManager.cs
```diff
@@ -142,6 +143,8 @@ public void PartRoom()
             joinedRoom = null;
         }
 
+        private readonly List<int> roomsFailedUpdate = new List<int>();
```
bdach, line 146: How about making this a

```suggestion
        private readonly HashSet<int> roomsFailedUpdate = new HashSet<int>();
```

and just skipping the `Contains()` check in the usage site entirely?

peppy: @bdach i've refactored this slightly to avoid throwing exceptions each poll (i think this is saner). please double-check.

peppy (approved):

bdach (approved):
Looks correct to me :+1:

PR #9381, opened 2020-06-27
Fix incorrect sample list slicing in mania conversion algorithm

### LINKED ISSUE #8880: Reverse Arrow Slider fails to translate for mania
**Describe the bug:** 
In mania, there appears to be a note translation failure resulting in an extra note set even though the reverse arrow section of the slider was silent in normal.

[running-in-90s.zip](https://github.com/ppy/osu/files/4541905/running.in.the.90s.with.an.extra.note.zip)
[tornado.zip](https://github.com/ppy/osu/files/4543706/video-2.zip)

**Confirmed Beatmaps** (The list of beatmaps I have encountered this bug on):
[Running In the 90's (Hardcore Rave Edit)~PHUCK...](https://osu.ppy.sh/beatmapsets/642158#osu/1361476) **The NoPHUCK difficulty I showed in the first video is the PHUCK difficulty, but lowered circle size.**
[Tornado~Insane](https://osu.ppy.sh/beatmapsets/236292#osu/569636)

**osu!lazer version:** 
2020.421.0

**Logs:**
[logs.zip](https://github.com/ppy/osu/files/4541911/logs.zip)

bdach:
Resolves #8880.

# Summary

As outlined in the issue thread, it was possible for hit objects with repeats to play wrong samples when converted into mania. This, in turn, was caused by yet another floating-point bug, namely one in which in an unfortunate enough scenario the index calculated here:

https://github.com/ppy/osu/blob/5f6c35a57767c3ff3c0a6aa91110293b2bbc8da1/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs#L488

could get miscalculated - the index before truncation would be slightly less than a whole number (i.e. 0.999...), therefore getting truncated down one index at the cast.

The proposed resolution is to discard the division altogether in favour of doing an iterative add. All of the callers of `getNodeSamplesAt()` already calculate the time points of each repeat by incrementing by `SegmentDuration`, and as floating-point math is deterministic this is a safer way to perform such a check that should not fail.

The following videos demonstrate the effect of this change:

| | | |
| :- | :-: | :-: |
| affected map (note the double beat at ~0:08) | [current `master`](https://streamable.com/045qsk) | [this PR](https://streamable.com/jphag5) |
| isolated test case | [current `master`](https://streamable.com/uj4tsm) | [this PR](https://streamable.com/aeijqm) |

The isolated test case is a slightly-modified fragment of the affected map and has been attached as an automated test case.

# Remarks

It was necessary to add the capability to test both `Samples` and `NodeSamples` in this PR, hence some unrelated changes to existing test cases. In particular I assigned the `normal` hit sample bank explicitly in the test cases modified here instead of specifying 0, as specifying 0 leaves a `null` sample bank until a `SampleControlPoint` overrides it (which doesn't happen in conversion test cases), and specifying sample names such as `-hitnormal` or `-hitclap` in expected test case results felt really off.

I considered just rounding but this felt like a safer solution.


osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
```diff
@@ -483,9 +483,12 @@ private List<IList<HitSampleInfo>> nodeSamplesAt(double time)
...
                 return null;
 
-            double segmentTime = (EndTime - HitObject.StartTime) / spanCount;
-
-            int index = (int)(segmentTime == 0 ? 0 : (time - HitObject.StartTime) / segmentTime);
+            // mathematically speaking this could be done by calculating (time - HitObject.StartTime) / SegmentDuration
+            // however, floating-point operations can introduce inaccuracies - therefore resort to iterated addition
+            // (all callers use this method to calculate repeat point times, so this way is consistent and deterministic)
+            int index = 0;
+            for (double nodeTime = HitObject.StartTime; nodeTime < time; nodeTime += SegmentDuration)
```
smoogipoo, line 490: How do you feel about `Math.Round`ing this? Given that `sliderDuration = n * segmentDuration`, `sliderDuration / segmentDuration` should give us a value that is approximately equal to `n` and `Math.Round` handles sanitisation.

Seems to pass the test cases too.

bdach: I was a little worried that the round could round things that it shouldn't but maybe it's a bit paranoid.

A bit of an awkward situation, this. Ideally I'd say `nodeSamplesAt` should just take in an integer index but the surrounding code is structured such that doing that is not straightforward

bdach: I've changed my mind on this and pushed in the rounding variant. This is a private function and one that shouldn't see much traffic going forward, so I now feel that being worried that someone will call this method with `time` somewhere between repeats is unreasonable.

smoogipoo (approved):
Should probably update stable to be able to generate the new conversions, but that can come later. :+1:

PR #9384, opened 2020-06-28
Temporarily disable masking of tournament song bar

### LINKED ISSUE #9200: Song bar doesn't show in tournament client at particular resolution
**Describe the bug:** 
Beatmap song bar wrong.
as you can see the pic.
if 1280x720 Beatmap song bar was wrong
if other resolution ,beatmap song bar was show

### ↓other resolution↓
新建文件夹12345

新建文件夹1234

### ↓1280*720↓
新建文件夹5678

新建文件夹56789

**osu!lazer version:** 
2020.602.0

bdach:
Temporarily resolves #9200 (hopefully).

# Summary

The previous attempt to resolve this used the following hack:

https://github.com/ppy/osu/blob/a4254802c1355bdafe7895aa7851eb36b657b66c/osu.Game.Tournament/Components/SongBar.cs#L80-L82

As far as I can tell it didn't ever work. The reason why is the `AutoSize` set directly after the hard-coded height specification; because it was set *after* `Height`, `CompositeDrawable` just [silently discards that height and falls back to auto size mode](https://github.com/ppy/osu-framework/blob/446ee955a5dc4239bbda1030defbf27e57c4d308/osu.Framework/Graphics/Containers/CompositeDrawable.cs#L1687-L1694), and the `SongBar` gets masked away anyway.

As an alternative, disable masking on `SongBar` entirely.

# Remarks

There were other temporary solutions here, such as removing the `AutoSize` or removing `LayoutDuration`, but this felt like the most direct way to address the problem without changing behaviour.

As an aside, this only becomes a problem because the song bar's initial `Y` is slightly bigger than 720 in this particular case.

### DIFF
--- a/osu.Game.Tournament/Components/SongBar.cs
+++ b/osu.Game.Tournament/Components/SongBar.cs
@@ -6,6 +6,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Beatmaps;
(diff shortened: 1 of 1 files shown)


MiraiSubject:
Locally checked this out and it fixes the issue. I didn't see anything regress when trying out 1080p or higher as well (not that I would expect it to happen considering the changes)

image

PR #9389, opened 2020-06-28
Implement UserPlaybackRate adjustments for StoryboardSamples

### LINKED ISSUE #13703: Storyboard samples rate doesn't follow Player's playback speed
**Describe the bug:** 
If music = beatmap sample, than you cant change speed that sample. In other words, hitsounds do not speed up and slow down as the map speed changes.
Map - https://osu.ppy.sh/beatmapsets/1237363#osu/2571858
https://youtu.be/v8Tk_JdRPqg (0:53)
**osu!lazer version:** 
2021.623.0

H2n9:
Makes user playback rate adjustments in replays(and autoplay) affect storyboard sample playback rates.

Continuation from feedback of #9301 

Closes #13703

### DIFF
--- a/osu.Game/Screens/Play/GameplayBeatmap.cs
+++ b/osu.Game/Screens/Play/GameplayBeatmap.cs
@@ -16,9 +16,14 @@ namespace osu.Game.Screens.Play
     {
         public readonly IBeatmap PlayableBeatmap;
 
-        public GameplayBeatmap(IBeatmap playableBeatmap)
+        private readonly GameplayClockContainer gameplayClockContainer;
+
+        public GameplayBeatmap(IBeatmap playableBeatmap, GameplayClockContainer gameplayClockContainer = null)
         {
             PlayableBeatmap = playableBeatmap;
+
+            if (gameplayClockContainer != null)
+                this.gameplayClockContainer = gameplayClockContainer;
         }
 
         public BeatmapInfo BeatmapInfo
@@ -44,6 +49,8 @@ namespace osu.Game.Screens.Play
         private readonly Bindable<JudgementResult> lastJudgementResult = new Bindable<JudgementResult>();
 
         public IBindable<JudgementResult> LastJudgementResult => lastJudgementResult;
+
+        public BindableNumber<double> UserPlaybackRate => gameplayClockContainer.UserPlaybackRate;
 
         public void ApplyResult(JudgementResult result) => lastJudgementResult.Value = result;
     }
--- a/osu.Game/Screens/Play/GameplayBeatmap.cs
+++ b/osu.Game/Screens/Play/GameplayBeatmap.cs
@@ -16,14 +16,9 @@ namespace osu.Game.Screens.Play
     {
         public readonly IBeatmap PlayableBeatmap;
 
-        private readonly GameplayClockContainer gameplayClockContainer;
-
-        public GameplayBeatmap(IBeatmap playableBeatmap, GameplayClockContainer gameplayClockContainer = null)
+        public GameplayBeatmap(IBeatmap playableBeatmap)
         {
             PlayableBeatmap = playableBeatmap;
-
-            if (gameplayClockContainer != null)
-                this.gameplayClockContainer = gameplayClockContainer;
         }
 
         public BeatmapInfo BeatmapInfo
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Play/GameplayBeatmap.cs
```diff
@@ -16,9 +16,14 @@ public class GameplayBeatmap : Component, IBeatmap
...
 
-        public GameplayBeatmap(IBeatmap playableBeatmap)
+        private readonly GameplayClockContainer gameplayClockContainer;
+
+        public GameplayBeatmap(IBeatmap playableBeatmap, GameplayClockContainer gameplayClockContainer = null)
         {
             PlayableBeatmap = playableBeatmap;
+
+            if (gameplayClockContainer != null)
+                this.gameplayClockContainer = gameplayClockContainer;
```
bdach, line 26: I am not sure `GameplayBeatmap` should be holding a reference to the entire `GameplayClockContainer`. I'd much prefer `GameplayBeatmap` only binding to `GameplayClockContainer.UserPlaybackRate` itself instead.

H2n9: Something like this
`public BindableNumber<double> UserPlaybackRate = new BindableNumber<double>();`
and bind that to `GameplayClockContainer.UserPlayback` instead?

bdach: Yeah, something like that, if at all possible. Then you could even move the actual binding to `Player` by using `BindTarget` instead of passing it down.

peppy:
I'm wondering if this is actually how we want to do this. Is it correct that a mod should be responsible for applying these adjustments, or would we expect it to "just work" at a gameplay level?

If the second is true, could we not just use the `GameplayClock`'s playback rate and reduce complexity of this implementation considerably? Maybe I'm overlooking something.

bdach:
That would imply that `Player` & co. should be somehow checking mods to ascertain if they should be adjusting playback rate of the gameplay clock? Not passing judgment yet or anything, just following the thread of logic.

It does sound slightly inconsistent with how mods seem to work generally but I'd really have to see it happen to say anything for sure.

peppy:
Would it not just be checking the single aggregate playback rate of the track?

bdach:
Right, yeah, but I'd probably want to have one central place where the speed adjustments are specified. Right now it's half the user setting, half the mod, and done separately for samples & tracks.

This is spitballing in my head so not sure if feasible or even making sense, but I'd want to separate the mod applying the rate change to the actual clock from the adjustment to the actual track (so that we can support HD/NC properly). So it'd be something like

- apply user playback numerical speed setting to one central bindable
- apply mod playback numerical speed change (by deferring to mod implementation)
- now we have one source of truth as to what the gameplay rate is
- apply gameplay rate to audio tracks and samples (the value is now explicitly given) - this is where HD/NC would differ

peppy:
I *think* `GameplayClock.Rate` is already the ground truth, only issue is it's not bindable?

bdach:
Hmm, I suppose that is correct and its value could be transferred out to a bindable every frame, but going bindable -> double -> bindable feels a little weird to me. It might be just me though.

peppy:
Yeah, I'd hope we can better encapsulate in bindable the whole way. Just wanted to make sure we were on the same page.

H2n9:
Just to make I sure understand correctly... Would it make sense to have a `PlaybackRate` bindable in the `GameplayBeatmap` which is then in some way binded to the `GameplayClock`'s rate. Then from there, samples can just fetch their rate from the `GameplayBeatmap` removing the need for the individual adjustments for mods and user playback rates here -

https://github.com/ppy/osu/pull/9389/files#diff-bff581ab9f71100af6532272955324daR45-R49

That would also then I assume make IApplicableToSample obsolete.

If that's the case then the only part I'm not sure how to approach is in regards to what was mentioned about exposing `GameplayClock`'s rate as a bindable. I attempted the simple way of just setting a bindable's value every frame just to confirm this implementation works (which it did) but I am unsure what the best way to expose it the whole way would be as was discussed.

peppy:
That sounds like the correct direction.

Honestly in interest of keeping this PR simple, I'd avoid making that change here and just do it as a follow-up (after the issue is first fixed).

I'm still not a fan of the `UserPlaybackRate` being in two places here though. Not sure I understand why it's not just accessed via the `GameplayClock`. Is it not available via DI?

bdach:
It was my intention to suggest that the playback rate be stored to `GameplayBeatmap` once and be DI'd in all read usage sites. Might not have conveyed that too well.

H2n9:
The `UserPlaybackRate` bindable is tracked within the `GameplayClockContainer` which is not available with DI; the `GameplayClock` itself however is. Can the `GameplayClockContainer` also be made available for DI so that its possible to fetch the `UserPlaybackRate` in `DrawableStoryboadSample`?

peppy:
can you just put the playback rate inside the `GameplayClock` in that case?

peppy:

Things have changed quite a bit since this PR was last updated so it's going to be easier to revisit this from the start.

PR #9391, opened 2020-06-28
Change Welcome intro track to loop

### LINKED ISSUE #9390: Welcome intro sequence music loop
The music should loop for the Welcome intro sequence, like it does in osu!fallback.

MiraiSubject:
Closes https://github.com/ppy/osu/issues/9390

~~The default setting for this right now is set to `true`.~~
~~This might not be desired for `Triangles` considering the `welcome to osu!` is baked into the beatmap.~~
~~This should be fine for `Circles` and `Welcome` though.~~

This makes the Welcome intro loop on the main menu until a user performs an action like changing the song or entering song select.

### DIFF
--- a/osu.Game.Tests/Visual/Menus/TestSceneIntroWelcome.cs
+++ b/osu.Game.Tests/Visual/Menus/TestSceneIntroWelcome.cs
@@ -11,5 +11,18 @@ namespace osu.Game.Tests.Visual.Menus
     public class TestSceneIntroWelcome : IntroTestScene
     {
         protected override IScreen CreateScreen() => new IntroWelcome();
+
+        public TestSceneIntroWelcome()
+        {
+            AddAssert("check if menu music loops", () =>
+            {
+                var menu = IntroStack?.CurrentScreen as MainMenu;
+
+                if (menu == null)
+                    return false;
+
+                return menu.Track.Looping;
+            });
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


frenzibyte (changes requested):
I really don't think all intros should have their track looped, the `Triangles` intro would not work well for it (and doesn't seem to be made to), same goes to `Circles`. Only the `Welcome` intro would work well as it is made to produce a continuation when looping. 

So I think the welcome intro should just loop its own track by itself, and nothing more (with tests providing assertions).

MiraiSubject:
Makes sense.

I made it an option because there could be users that would like the `Welcome` intro but not having it loop. 
Then again it would only make sense for `Welcome` to be looped in the first place. 

I'll update it so only `Welcome` gets looped

peppy:
Is that how it is back in stable-fallback?

Nyanekoy:
> Is that how it is back in stable-fallback?

Yea it does loop always unless you switch to another song then you can only listen to it again until you restart

osu.Game.Tests/Visual/Menus/TestSceneIntroWelcome.cs
```diff
@@ -11,5 +11,18 @@ namespace osu.Game.Tests.Visual.Menus
     public class TestSceneIntroWelcome : IntroTestScene
     {
         protected override IScreen CreateScreen() => new IntroWelcome();
+
+        public TestSceneIntroWelcome()
+        {
+            AddAssert("check if menu music loops", () =>
```
peppy, line 17: kinda pointless test but i guess it's fine.

peppy (approved):

PR #9393, opened 2020-06-29
Improve quality of song select beatmap wedge

### LINKED ISSUE #9305: Visual quality of sprites / buffered containers degrades at very high resolutions
**Describe the bug:**
Some images at different places have very low-quality rendering. I found this on both home background and beatmap preview.

This is platform-specific, can be reproduced stably on iPad here, but not on Windows or Android.

Device is iPad Pro 11 inch (2018).

378806B1-013C-4235-8050-69DB7DEE5AC8
E8675FD1-AC93-422C-8206-7118A48F31CE

**osu!lazer version:** 
2020.616.0 (TestFlight)

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9305

It doesn't look to be drawing too much anyway... I considered moving the buffered container to only the background, but the BC wrapping everything still needs to exist anyway otherwise there'd be weirdness during the fadeout transition when changing wedges (details fade out and background peeks through).

### DIFF
--- a/osu.Game/Screens/Select/BeatmapInfoWedge.cs
+++ b/osu.Game/Screens/Select/BeatmapInfoWedge.cs
@@ -155,7 +155,6 @@ namespace osu.Game.Screens.Select
                 var metadata = beatmapInfo.Metadata ?? beatmap.BeatmapSetInfo?.Metadata ?? new BeatmapMetadata();
 
                 CacheDrawnFrameBuffer = true;
-                RedrawOnScale = false;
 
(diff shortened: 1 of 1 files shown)


peppy:
Have you checked the case where aspect ratio changes due to toolbar hiding/showing? That was the reason for having this in place, IIRC.

smoogipoo:
I drop to 40FPS on my macbook on both master and this branch, so I don't think this is where the performance hit is coming from.

Issue #9395, opened 2020-06-29
Set the range of the Timing Distribution (in results screen) to be based on effective OD


Magnus-Cosmos:
Currently the range for the axis for the Timing Distribution is based the hit errors themselves. This makes comparing replays very confusing, e.g. 
![osu_2020-06-29_01-42-07](https://user-images.githubusercontent.com/39173355/85977422-95fa2900-b9aa-11ea-81ec-da524da60d69.jpg)
![osu_2020-06-29_01-37-20](https://user-images.githubusercontent.com/39173355/85977438-9d213700-b9aa-11ea-859a-942cc9f2e99e.jpg)
These two scores are done on the same map with the same mods. Both of them have most of their hit errors between -20ms and +20ms, but it looks like the bottom one is much worse.

I think it would make more sense to have the range based on the OD (more specifically effective OD, e.g. playing with HR on a map will decrease the range), this will give players a better idea of how accurate they are hitting, and makes comparing scores much easier.

smoogipoo:
I think this is probably fine to do. It was originally like that in the initial implementation, but it also only had 51 bars at the time (currently 101).

peppy:
Tracking in https://github.com/ppy/osu/issues/33952.

PR #9396, opened 2020-06-29
Exclude hit objects with an EmptyHitWindow from being considered in TimingDistributionGraph

LumpBloom7:
HitObjects with `EmptyHitWindow`s should not be considered in the statistics graph.

This also prevents hitobjects like Mania's Holds (which is not judged, but wraps two other judged objects), Slider repeats, ticks, and so on from artificially inflating the number of 0ms in the graph, making it inaccurate.

Before:
image
After:
image

### DIFF
--- a/osu.Game/Screens/Ranking/Statistics/HitEventTimingDistributionGraph.cs
+++ b/osu.Game/Screens/Ranking/Statistics/HitEventTimingDistributionGraph.cs
@@ -48,7 +48,7 @@ namespace osu.Game.Screens.Ranking.Statistics
         /// <param name="hitEvents">The <see cref="HitEvent"/>s to display the timing distribution of.</param>
         public HitEventTimingDistributionGraph(IReadOnlyList<HitEvent> hitEvents)
         {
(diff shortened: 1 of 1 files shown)


bdach:
0 still looks inflated on the "after" graph. I'd check whether hold note ticks are excluded, too.

LumpBloom7:
Nothing seems to be returning 0ms offset.

I think it's simply because it includes TimeOffsets roughly +-2ms from 0ms as well.

LumpBloom7:
One thing I noticed is that any decent playthrough (on mania at least) will result in the 0ms line being far longer. 
image
In this screenie the 0ms bar includes offsets up to around +-4ms.

It would probably be better if the graph is spread out a bit more, leaving less flat bits on the sides.

smoogipoo:
That also relates to https://github.com/ppy/osu/issues/9395, which suggests reducing the spread in favour of consistency.

peppy (approved):

smoogipoo (approved):

PR #9409, opened 2020-06-30
Improve performance of sequential scrolling algorithm

### LINKED ISSUE #9331: Mania maps with many control points have bad load times
This has to do with `SequentialScrollAlgorithm.GetDisplayStartTime()`, which goes through many control points for potentially many objects, leading to `O(n^2)` in the worst case.

Example: https://osu.ppy.sh/beatmapsets/376293#mania/823842

Also doesn't help that this stuff is done in the first `Update()`. It would _maybe_ be acceptable if it could be done in `Add()` (while we're on the async context), but that would probably require making this `scrollLength`-agnostic.
I say acceptable since a theoretical `O(n^2)` is very unlikely in practice.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9331

From:
Screenshot_2020-06-30_21-42-45

To:
Screenshot_2020-06-30_21-41-16


smoogipoo:
Also try changing the scroll speed during gameplay with the beatmap mentioned in the issue :)

peppy (approved):
nice one

PR #9417, opened 2020-07-01
Update with framework-side storage creation changes

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3671


peppy:
Looks like the migration test is failing with these changes locally.

image

Haven't looked into cause yet.

smoogipoo:
I've fixed the test up - the issue is that framework changes the host's storage, so this test now needs to store the original directory for assertions after the migration happens.

I think the framework-side logic makes sense? You shouldn't ever see the default storage directory if you've specified a different one.

bdach (approved):
seems ok to me :+1:

PR #9418, opened 2020-07-01
Fix possible startup error when using custom directory

### LINKED ISSUE #9379: Game crashes if storage location is not available
**Describe the crash:** 
Nothing. Launch it, the app won't start and stop trying after few seconds. No pop up error whatsoever.

**osu!lazer version:** 
Lazer 2020.623.1

**Logs:**
2020-06-27 07:08:14 [verbose]: ----------------------------------------------------------
2020-06-27 07:08:14 [verbose]: runtime Log for leehoemun (LogLevel: Verbose)
2020-06-27 07:08:14 [verbose]: osu 2020.623.1.0
2020-06-27 07:08:14 [verbose]: Running on Unix 19.5.0.0, 8 cores
2020-06-27 07:08:14 [verbose]: ----------------------------------------------------------
2020-06-27 07:08:14 [verbose]: Unhandled exception has been allowed with 0 more allowable exceptions 

**Computer Specifications:** 
macbook pro 13.3 2019, 256+8gb, i5 1.4Ghz

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9379

Prereqs:
- [x] https://github.com/ppy/osu/pull/9417

If an error happens, the default storage directory is used and a notification is shown after the game finishes starting up. The config isn't cleared, because the user could potentially mount the device and fix the issue.


bdach (dismissed):
Tested by doing a `$ chmod -x` on a custom data directory, seems to be doing the job fine :+1:

peppy:
This works fine to cover the hard crash, but before merging there's a few points which we should probably discuss / reach consensus on:

- What happens when the target location is writeable but no longer exists?

We may want to show the same message and make the user aware that they have potentially moved the folder and osu! is no longer aware of this. Right now, it will create a new folder structure in the specified location, which is probably not what the user wants under any circumstance.

A common scenario for this happening is if the user moves the custom folder to a new location themselves, and still wants lazer to use it.

We probably want to have the "change location" use an already existing folder rather than erroring out if files already exist at the target location (as it currently does).

- If the user ignores the error, they now have two data locations

For the average user, I'd hope this be a fullscreen dialog on startup forcing action, rather than an ugly exception message. It should give the user a chance to resolve the issue or reset their folder location to the default. ie. a dialog with [ Retry ] [ Reset to default ] [ Choose new location ]. The third button would obviously work best if the above points were first resolved.

If you agree, I think changing the PR even to address only point 2 may be a saner starting point to handling this scenario.

peppy (changes requested):
blocking pending discussion

smoogipoo:
I agree. I've updated the PR to add the dialog:

When empty:
Screenshot_2020-07-06_21-54-18

When not accessible:
Screenshot_2020-07-06_21-55-22

We could eventually add a folder selection dialog, but I didn't want to do that just yet as it would take quite some abstraction of the migration dialog (we don't want to migrate, and we also don't want to create the "osu-lazer" sub-directory, and we also probably want to prompt).

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+            private OsuGameBase osuGame { get; set; }
+
+            public StorageErrorDialog(OsuStorage storage, OsuStorageError error)
+            {
+                HeaderText = "osu! storage error";
+                Icon = FontAwesome.Solid.ExclamationTriangle;
+
+                var buttons = new List<PopupDialogButton>();
+
+                BodyText = $"osu! encountered an error when trying to use the custom storage path ('{storage.CustomStoragePath}').\n\n";
```
peppy, line 335:
```suggestion
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+
+                switch (error)
+                {
+                    case OsuStorageError.NotAccessible:
+                        BodyText += $"The default storage path ('{storage.DefaultStoragePath}') is currently being used because the custom storage path is not accessible.\n\n"
```
peppy, line 340:
```suggestion
                        BodyText += $"The specified osu! data location (\"{storage.CustomStoragePath}\") is not accessible. If it is on external storage, please reconnect the device and try again."
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+                                    + "Is it on a removable device that is not currently connected?";
```
peppy, line 341:
```suggestion
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+                                Text = "Try again",
+                                Action = () =>
+                                {
+                                    if (!storage.TryChangeToCustomStorage(out var nextError))
+                                        dialogOverlay.Push(new StorageErrorDialog(storage, nextError));
+                                }
+                            },
+                            new PopupDialogOkButton
+                            {
+                                Text = "Use the default path from now on",
```
peppy, line 356:
```suggestion
                                Text = "Reset to default location",
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+                                Action = storage.ResetCustomStoragePath
+                            },
+                            new PopupDialogCancelButton
+                            {
+                                Text = "Only use the default path for this session",
```
peppy, line 361:
```suggestion
                                Text = "Use default location for this session",
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+                            },
+                        });
+                        break;
+
+                    case OsuStorageError.AccessibleButEmpty:
+                        BodyText += "The custom storage path is currently being used but is empty.\n\n"
+                                    + "Have you moved the files elsewhere?";
```
peppy, line 368:
```suggestion
                        BodyText = $"The specified osu! data location (\"{storage.CustomStoragePath}\") is empty. If you have moved the files, please close osu! and move them back.";
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+
+                        // Todo: Provide the option to search for the files similar to migration.
+                        buttons.AddRange(new PopupDialogButton[]
+                        {
+                            new PopupDialogOkButton
+                            {
+                                Text = "Reset to default",
```
peppy, line 375:
```suggestion
                                Text = "Reset to default location",
```

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -308,5 +316,76 @@ public ConfirmExitDialog(Action confirm, Action cancel)
...
+                                Action = storage.ResetCustomStoragePath
+                            },
+                            new PopupDialogCancelButton
+                            {
+                                Text = "Keep using the custom path"
```
peppy, line 380:
```suggestion
                                Text = "Start fresh at specified location"
```

I'd also put this button above the "reset to default"

peppy (changes requested):

peppy (approved):

PR #9421, opened 2020-07-01
Standardise osu!catch coordinate system to 0..512

### LINKED ISSUE #9405: Eliminate usage of legacy coordinate system
To denote an x-position of a fruit, osu!stable uses an integer coordinate between 0 and `BASE_WIDTH`.
https://github.com/ppy/osu/blob/9815a0ab6626312849251710426699ef47c3a931/osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs#L19
For example, this constant is required to parse `.osu` file. Because the speed of a catcher is set to "1 unit per milliseconds" based on this coordinate, this constant is also used in logic.
https://github.com/ppy/osu/blob/4198f4d96e88f5d75def4cd21acad59c9e1193e7/osu.Game.Rulesets.Catch/UI/Catcher.cs#L212-L213
https://github.com/ppy/osu/blob/4198f4d96e88f5d75def4cd21acad59c9e1193e7/osu.Game.Rulesets.Catch/UI/CatcherArea.cs#L21

The new coordinate system denotes an x-position by a float between 0 and 1.

Both coordinate systems are used in different places, and it is confusing which coordinate system a variable is represented in.

I propose to refactor code to reduce the usage of the legacy coordinate system as much as possible.

ekrctb:
Resolves #9405. Use 0..512 consistently for all fruit/catcher position variables.
Also, renamed `CatchPlayfield.BASE_WIDTH` to `CatchPlayfield.WIDTH` because I thought it is clearer (and if some code depends on this constant then it will be a compiler error rather than causing a silent bug). Also added `CatchPlayfield.CENTER_X` for convenience. Maybe those constants shouldn't be in this `CatchPlayfield` class (because it is not only used for "UI", but also used for everything) but not sure where to put.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/CatchBeatmapConversionTest.cs
+++ b/osu.Game.Rulesets.Catch.Tests/CatchBeatmapConversionTest.cs
@@ -8,7 +8,6 @@ using NUnit.Framework;
 using osu.Framework.Utils;
 using osu.Game.Rulesets.Catch.Mods;
 using osu.Game.Rulesets.Catch.Objects;
-using osu.Game.Rulesets.Catch.UI;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Tests.Beatmaps;
 
@@ -83,7 +82,7 @@ namespace osu.Game.Rulesets.Catch.Tests
 
         public float Position
         {
-            get => HitObject?.X * CatchPlayfield.BASE_WIDTH ?? position;
+            get => HitObject?.X ?? position;
             set => position = value;
         }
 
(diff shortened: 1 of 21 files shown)


peppy (approved):
I think this looks better than what we had.

peppy:
@smoogipoo confirm you're okay with this direction (as per [discussion in issue](https://github.com/ppy/osu/issues/9405)).

smoogipoo (approved):
I'm fine with these changes. I've also had many bad experiences with normalised coordinates in catch.

PR #9426, opened 2020-07-02
Add welcome background to the main menu

MiraiSubject:
Depends on: 
- [x] https://github.com/ppy/osu-resources/pull/106
- [x] `osu-resources` bump

This is enabled when a user switches their `Intro Sequence` to `Welcome`. Skins and Beatmaps (the supporter only features) override this behaviour as usual. 

When the user picks the `Welcome` intro sequence their background will also change to the appropriate background. This also serves as an indicator to show which theme the user has enabled currently.


osu.Game/Screens/Backgrounds/BackgroundScreenDefault.cs
```diff
@@ -22,11 +22,12 @@ public class BackgroundScreenDefault : BackgroundScreen
         private int currentDisplay;
         private const int background_count = 7;
 
-        private string backgroundName => $@"Menu/menu-background-{currentDisplay % background_count + 1}";
+        private string backgroundName;
```
bdach, line 25: Does this need to be a class field? As far as I can tell it could as well be a local inside of `createBackground`.

MiraiSubject: Good point, resolved it in an incoming commit

peppy (dismissed):

peppy (approved):

PR #9433, opened 2020-07-04
Add option to dim overlay.

ondrejvajdak:


bdach:
Thanks for the contribution, however:

* In the future please do not PR off `master`, as mentioned [in the contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md).
* It would have been better if you had raised this as a proposal before, as I'm unsure if this is a direction we want to be taking in general.
* As it comes to the code itself, ideally the dim shouldn't be applied to all of the individual drawables separately, but to one encapsulating drawable.

Drison64:
Hello,

I am sorry, i did not know how to make a proposal, still not good at using Git/Github.
I have found the overlay too bright, so i thought i would try it, without reading those guidelines.

Should I continue here or make that proposal? (And sorry for the code, first try with C#)

bdach:
I am not comfortable making a decision whether we want a feature like this implemented myself, but I'm weakly against it seeing that the gameplay UI can be disabled entirely using <kbd>Shift</kbd>+<kbd>Tab</kbd>, and that I'm pretty sure gameplay UI customisation/skinning is in long-term plans. I would however wait for other opinions on this matter.

Drison64:
Oh nevermind, since there is already a way to disable overlay i will close this request.
Thanks for your time!

PR #9434, opened 2020-07-04
Replace further spinner transforms with manual lerp

### LINKED ISSUE #8367: Spinner length degrades update performance indefinitely
**Describe the bug:** 
Spinners even when not being spun lower the update performance, once the spinner finishes it causes a stutter, once that's done update promptly returns to normal. Tested on this map with a 2 minute long spinner https://osu.ppy.sh/beatmapsets/679738#osu/1437412

I'm also very unsure if the logs really picked it up in a clear way, so i'm sorry if that's the case 
Near the start
osu_2020-03-19_15-23-59
By the end
osu_2020-03-19_15-26-13
And not sure if this is helpful but I noticed CPU usage on SpriteIcon goes up with SpinnerDisc in this situation, albeit at a slower rate
osu_2020-03-19_16-43-12

**osu!lazer version:** 
2020.320.1
**Logs:**
[runtime.log](https://github.com/ppy/osu/files/4357165/runtime.log)
[performance.log](https://github.com/ppy/osu/files/4357166/performance.log)
[performance-draw.log](https://github.com/ppy/osu/files/4357167/performance-draw.log)
[performance-update.log](https://github.com/ppy/osu/files/4357164/performance-update.log)

bdach:
Resolves #8367 again (hopefully for the last time).

# Summary

This is the same fix as #8394 (replacing transforms with non-zero duration that would start queueing up and taking up more and more of update frame time with manual lerps) in the places that got overlooked.

Video demonstrations of the change:

* [before](https://streamable.com/obthsu) (note update time increases on `DrawableSpinner` and `SpriteIcon`),
* [after](https://streamable.com/ff5mx4).

Both videos taken on a debug build, with frame limiter off. Results might not be fully accurate as running with frame limiter off triggers thermal throttling on my box sometimes so can't guarantee they were taken with identical parameters in that sense, but the difference in behaviour exhibited is visible and should be reproducible.

# Remarks

Lerp value for the symbol stolen off #8394; value for the disc scale is proportionally scaled (200 is to 500 what 1/100 is to 1/40).

@Azureharp Would appreciate it if you could test this to make sure the issue doesn't have to be re-opened a third time.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
@@ -14,6 +14,7 @@ using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Utils;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Screens.Ranking;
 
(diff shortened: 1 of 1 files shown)


peppy:
interesting, i thought these were here for rewind support so didn't touch them. good to know we don't need them (assuming rewind behaviour hasn't changed).

bdach:
I did check out what happens on rewind and it seemed to look correct. Might be wrong on that though.

bdach:
Hmm, I sort of take that back... This is [how rewind looks with this change](https://streamable.com/lut07z). In some cases the sprite icon stops rotating which is probably wrong.

Not sure whether we want to take the transform performance hit or whether I should figure out how to make the lerp work backwards. It's getting very late here so I'll sleep on it and come back to it tomorrow if you don't mind.

peppy:
one thing that hav become apparently with my pooling efforts is that we can't really have DrawableHitObjects storing transforms for rewinds of user input (we would lose them when returned to the pool).

A more correct approach may be to store such information inside the judgement, and handle manually in the spinner's update method. Changing to simple lerp for the time being may be fine.

I'll check visually and make a call on this.

Azureharp:
Works really well solid ~4200 update throughout, and pretty much identical all the way up to 3 hour long spinners. Couldn't find much in the way of strange occurrences either.
Also thank you for the sheer amount of dev stuff you do and put up with

peppy:
Looks fine!

PR #9435, opened 2020-07-05
Implement support for custom file paths for remaining mania elements

### LINKED ISSUE #9242: Custom file paths for mania judging through skin.ini not implemented yet

mcendu:
Closes https://github.com/ppy/osu/issues/9242#issuecomment-643328742


peppy (approved):
Looks sane

Issue #3812, opened 2018-12-03
ModSelectOverlay causes the Home button to work incorrectly


Gabixel:
**Describe your problem:** The issue basically happens when the `ModSelectOverlay` is open: the `Home` button doesn't send you to the `MainMenu`, but it just closes the `ModSelectOverlay`.

Gabixel:
Seems like this issue happens again

Joehuu:
Regressed in #5149.

peppy:
This regression was intentional for now.

Joehuu:
Can reopen.

Gabixel:
Seems like this issue also happens when pressing the `X` button on Windows with the new `2019.919.0` version

Joehuu:
When the mod select is open, should clicking back button close it (current implementation) or exit to main menu?

![](https://puu.sh/EmZxe/34b913ddb9.png)

We have this close button and the escape key already.

smoogipoo:
I think we intended the back button to close mod select.

peppy:
Yep, for now this is intended.

Joehuu:
I'm trying to fix this, and wondering if there's something to check if a drawable (in this case a button) is clicked.

I'm planning to make `BackButton` public in `OsuGame`, DI `OsuGame` to `SongSelect`, and add a clicked check to the current logic inside `OnExiting`.

peppy:
A better way is likely going to be adding a new interface `ICustomBackButtonAction : IOsuScreen { Func<bool> OnBackButton() }`, then adding code in the `BackButton.Action` delegate in `OsuGame` to check whether the current screen implements this new interface, then give it a chance to handle and block (don't call `ScreenStack.Exit()`) if it returns `true`.

peppy:
In fact, you can probably get away with just adding it to `IOsuScreen` / `OsuScreen` as `protected virtual bool OnBackButton() => false;`

PR #9436, opened 2020-07-05
Fix quick retry hotkey not working from statistics display screen

### LINKED ISSUE #9430: Using quick-retry hotkey in results screen when stats are expanded results in black screen
**Describe the bug:** 
Using the quick-retry hotkey in the results screen when the play statistics are expanded results in a black screen, which go backs to the player loader when exited (as seen in the attached video).

Reproduction steps:
1. Play a beatmap and reach the results screen
2. Expand the statistics
3. Press your quick-retry hotkey

https://streamable.com/mit0rq
**osu!lazer version:** 
Happens on latest `master`

**Logs:**
Irrelevant here.

peppy:
- Closes #9430


Joehuu:
This makes the back button close the results screen entirely. Not sure if that's intentional.

Joehuu:
I have a better fix that also fixes the home button being blocked here too (bundled with https://github.com/ppy/osu/issues/3812 / other places).

peppy:
Will block this for now.

PR #9437, opened 2020-07-06
Implement NewsCard component for NewsOverlay

EVAST9919:
Screenshot_1


osu.Game/Online/API/Requests/Responses/APINewsPost.cs
```diff
@@ -0,0 +1,38 @@
...
+        public string EditUrl { get; set; }
+
+        [JsonProperty(@"first_image")]
+        public string FirstImage { get; set; }
+
+        [JsonProperty(@"published_at")]
+        public DateTime PublishedAt { get; set; }
+
+        [JsonProperty(@"updated_at")]
+        public DateTime UpdatedAt { get; set; }
```
bdach, line 27: Pretty sure these should be `DateTimeOffset`s.

osu.Game/Online/API/Requests/Responses/APINewsPost.cs
```diff
@@ -0,0 +1,38 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using Newtonsoft.Json;
+using System;
+
+namespace osu.Game.Online.API.Requests.Responses
+{
+    public class APINewsPost
+    {
+        [JsonProperty(@"id")]
```
bdach, line 11: Raw string prefixes seem unnecessary.

osu.Game/Overlays/News/NewsCard.cs
```diff
@@ -0,0 +1,196 @@
...
+                if (source.StartsWith('/'))
+                    return "https://osu.ppy.sh" + source;
+
+                return source;
+            }
+        }
+
+        private class DateContainer : CircularContainer, IHasTooltip
+        {
+            public string TooltipText => date.ToString("d MMMM yyyy hh:mm:ss UTCz");
```
bdach, line 161: This tooltip is not consistent with web design. This should be using whatever `DrawableDate` is, probably by hoisting out `DrawableDate.DateTooltip` of the inner class and making it public.

osu.Game/Online/API/Requests/Responses/APINewsPost.cs
```diff
@@ -0,0 +1,38 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using Newtonsoft.Json;
+using System;
+
+namespace osu.Game.Online.API.Requests.Responses
+{
+    public class APINewsPost
```
bdach, line 9: Where did you get this endpoint contract structure from? I looked at the news page on web and I can't find a straight API request that would just fetch this as JSON, [API docs are missing](https://osu.ppy.sh/docs/index.html#apiv2newsnews), and this PR does not have a web request attached that I can inspect.

bdach (changes requested):

osu.Game/Online/API/Requests/Responses/APINewsPost.cs
EVAST9919: I just sent a raw request and checked the response
Screenshot_1

bdach: I notice an `&amp;` HTML entity in the `author` field of post with ID 802 (I'm betting it's possible for `title` and `preview` to contain HTML entities, too). This should be handled at some point using `WebUtility.HtmlDecode` - probably best at deserialisation time.

EVAST9919: Nice catch will consider!
Also, as a side note: web can't handle this either, should probably open an issue
Аннотация 2020-07-06 232512

bdach: Opened ppy/osu-web#6274.

osu.Game/Graphics/DrawableDate.cs
```diff
@@ -82,7 +82,7 @@ private void updateTimeWithReschedule()
 
         public object TooltipContent => Date;
 
-        private class DateTooltip : VisibilityContainer, ITooltip
+        public class DateTooltip : VisibilityContainer, ITooltip
```
smoogipoo, line 85: Would like this to be moved to a separate file.

smoogipoo (changes requested):
Looks fine otherwise

bdach:
I've adjusted one font size to match web; otherwise than that looks good.

Going to make and exception and approve this despite having pushed to this PR as the change was menial.

bdach (approved):

smoogipoo (approved):

PR #9439, opened 2020-07-06
Add ability to disable the Windows keys while in gameplay

Game4all:
Opening this for potential review, as this is months away from getting merged as it only works with the SDL backend (Windows Keyboard hooks don't play nicely with osuTK `GameWindow`).

# Summary

- This adds the ability to disable the (left and right) windows key ~~(as well as the ALT+TAB, ALT+ESC and CTRL + ESC shortcuts)~~  through a settings option (enabled by default) while in gameplay, similarly to osu!stable on windows.

## Testing

No Automated Tests for this.
Manual testing on my two windows installations shows it to _work as expected_. (use the `--sdl` command line flag to self-test)

---
As a side note, if the game was to be added more platform-specific features toggleable through settings options, I guess it may be worth looking at adding a way to register platform-specific settings in the future, as polluting the game config manager with platform-specific settings feels weird.

### DIFF
--- a/osu.Desktop/Windows/GameplayWinKeyHandler.cs
+++ b/osu.Desktop/Windows/GameplayWinKeyHandler.cs
@@ -11,7 +11,7 @@ namespace osu.Desktop.Windows
 {
     public class GameplayWinKeyHandler : Component
     {
-        private Bindable<bool> winKeyEnabled;
+        private Bindable<bool> allowScreenSuspension;
         private Bindable<bool> disableWinKey;
 
         private GameHost host;
@@ -21,11 +21,11 @@ namespace osu.Desktop.Windows
         {
             this.host = host;
 
-            winKeyEnabled = host.AllowScreenSuspension.GetBoundCopy();
-            winKeyEnabled.ValueChanged += toggleWinKey;
+            allowScreenSuspension = host.AllowScreenSuspension.GetBoundCopy();
+            allowScreenSuspension.ValueChanged += toggleWinKey;
 
             disableWinKey = config.GetBindable<bool>(OsuSetting.GameplayDisableWinKey);
-            disableWinKey.BindValueChanged(t => winKeyEnabled.TriggerChange(), true);
+            disableWinKey.BindValueChanged(t => allowScreenSuspension.TriggerChange(), true);
         }
 
         private void toggleWinKey(ValueChangedEvent<bool> e)
--- a/osu.Desktop/Windows/GameplayWinKeyHandler.cs
+++ b/osu.Desktop/Windows/GameplayWinKeyHandler.cs
@@ -25,7 +25,7 @@ namespace osu.Desktop.Windows
             winKeyEnabled.ValueChanged += toggleWinKey;
 
             disableWinKey = config.GetBindable<bool>(OsuSetting.GameplayDisableWinKey);
-            disableWinKey.BindValueChanged(t => winKeyEnabled.TriggerChange());
+            disableWinKey.BindValueChanged(t => winKeyEnabled.TriggerChange(), true);
         }
 
         private void toggleWinKey(ValueChangedEvent<bool> e)
--- a/osu.Desktop/Windows/GameplayWinKeyHandler.cs
+++ b/osu.Desktop/Windows/GameplayWinKeyHandler.cs
@@ -1,0 +1,39 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Platform;
+using osu.Game.Configuration;
+
+namespace osu.Desktop.Windows
+{
+    public class GameplayWinKeyHandler : Component
+    {
+        private Bindable<bool> winKeyEnabled;
+        private Bindable<bool> disableWinKey;
+
+        private GameHost host;
+
+        [BackgroundDependencyLoader]
+        private void load(GameHost host, OsuConfigManager config)
+        {
+            this.host = host;
+
+            winKeyEnabled = host.AllowScreenSuspension.GetBoundCopy();
+            winKeyEnabled.ValueChanged += toggleWinKey;
+
+            disableWinKey = config.GetBindable<bool>(OsuSetting.GameplayDisableWinKey);
+            disableWinKey.BindValueChanged(t => winKeyEnabled.TriggerChange());
+        }
+
+        private void toggleWinKey(ValueChangedEvent<bool> e)
+        {
+            if (!e.NewValue && disableWinKey.Value)
+                host.InputThread.Scheduler.Add(WindowsKey.Disable);
+            else
+                host.InputThread.Scheduler.Add(WindowsKey.Enable);
+        }
+    }
+}
--- a/osu.Desktop/Windows/GameplayWinKeyHandler.cs
+++ b/osu.Desktop/Windows/GameplayWinKeyHandler.cs
@@ -11,26 +11,26 @@ namespace osu.Desktop.Windows
 {
     public class GameplayWinKeyHandler : Component
     {
-        private Bindable<bool> allowScreenSuspension;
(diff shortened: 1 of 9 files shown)


osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -76,6 +76,11 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Score display mode",
                     Bindable = config.GetBindable<ScoringMode>(OsuSetting.ScoreDisplayMode)
+                },
+                new SettingsCheckbox
+                {
+                    LabelText = "Disable Win key during gameplay",
```
peppy, line 82:
```suggestion
                    LabelText = "Disable windows key during gameplay",
```

osu.Desktop/Windows/WindowsKey.cs
```diff
@@ -0,0 +1,86 @@
...
+            public readonly int Flags;
+        }
+
+        private static int lowLevelKeyboardProc(int nCode, int wParam, ref KdDllHookStruct lParam)
+        {
+            if (wParam >= wm_keydown && wParam <= wm_syskeyup)
+            {
+                switch (lParam.VkCode)
+                {
+                    case 0x09 when lParam.Flags == 32: // alt + tab
```
peppy, line 39: shouldn't disable anything but the windows key.

peppy (changes requested):

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -76,6 +76,11 @@ private void load(OsuConfigManager config)
...
+                    Bindable = config.GetBindable<bool>(OsuSetting.GameplayDisableWinKey)
                 }
```
frenzibyte, line 84: This would appear in mobile platforms as well from the looks of it

frenzibyte (changes requested):

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
Game4all: yep, forgot to add a conditional check to have that setting appear only on Windows.

 But going forward, I guess having a way to register platform-specific settings maybe the right way to dothis instead of polluting the game config manager with platform-specific settings and having platform checks scattered across the code to hide / show these settings depending on the platform, if we were to add more platform-specific stuff in the future.

frenzibyte: That's another thing to deal with later on, for now keep it simple and just check for it to only appear in windows.

osu.Desktop/Windows/GameplayWinKeyHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        private GameHost host;
+
+        [BackgroundDependencyLoader]
+        private void load(GameHost host, OsuConfigManager config)
+        {
+            this.host = host;
+
+            allowScreenSuspension = host.AllowScreenSuspension.GetBoundCopy();
+            allowScreenSuspension.ValueChanged += toggleWinKey;
```
bdach, line 25: My reservations towards this whole thing aside, I find this awkward given that `AllowScreenSuspension` was originally intended for mobile platforms, so now one setting has two semantic meanings on two different platforms.

bdach (commented):
I'll say that I'm personally opposed to this existing at all and think this is better suited to customisation tools like powertoys, but I think I'm done with windows as an OS for the time being anyway, so :shrug:

osu.Desktop/Windows/GameplayWinKeyHandler.cs
smoogipoo: Yeah I agree. Should just use the `GameplayDisableWinKey` setting and call it a day. Probably should be something more like `ScreenSuspensionHandler` that's added to the hierarchy whenever needed rather than being at the `Game` level.

peppy: Hmm, are you proposing rewriting the screen suspension logic osu!-side? Or saying that the windows key isn't always going to be disabled when screen suspension is disabled (potentially editor?)

smoogipoo: > saying that the windows key isn't always going to be disabled when screen suspension is disabled

This basically. Editor (maybe), replays, match screen (while in a live room), probably will disable screen suspension but shouldn't disable windows key.

peppy: Fair enough. So the plan is to dial back and disable windows key game-wide? If that's the case, the default needs to be `false` but it will need to be reset to a new setting when we change it to be player-only, and the default should be `true`. Also the setting text is wrong if we are going to do this.

I'd propose it should be fixed in this PR to only occur during gameplay.

Game4all: My initial implementation (wrongly) assumed that all the screens that disabled screen suspension would disable the windows key as well (which in practice isn't right at all, according to smoogipoo's previous answer).

I was thinking of extending `ScreenSuspensionHandler` with a bindable boolean to allow disabling the windows key independently of the screen suspension and add a way to have the state globally available through caching (basically an instance would be globally cached and the instances added to a screen drawable hierarchy would bind to the global instance bindables). The `osu.Desktop` part would just have to bind to the bindables of the globally cached instance.

Using alternatively a `SessionStatic` may work aswell.

bdach:
I don't think `SessionStatics` was designed for this kind of usage; up to now it was mostly transient session state (i.e. used to track whether something was already done once in a session). I'd be wary of allowing this usage as it sets a dangerous precedent (the new static is essentially a global variable, and those are universally bad).

Game4all:
Huh, I'm quite unsure of the direction to take to store the state like implementing some kind of component that is cached at the game-level to store the activation state of the windows key, this feels quite overkill (it could maybe house screen suspension stuff in aswell). `SessionStatics` XMLDoc says that it stores per-session global data which isn't persisted to disk; this looked to me like a quite lightweight way of storing the key activation state altough I understand your concerns about using it to store data like a global variable.

peppy:
I'm still all for having it use the `AllowScreenSuspension` flag until we come across issues with that. It covers every scenario we want currently.

There's no point in over-complicating something so isolated and easy to fix down the track.

`GameplayWinKeyHandler` DI's host, consumes the `AllowSuspend` bindable and does its thing. no changes elsewhere in the code.

bdach:
I suppose we could split this functionality from `AllowScreenSuspension` when it becomes necessary due to that bindable being used for different purposes. Can't say I care too much about this functionality existing all things considered and it's fairly self-contained as-is anyway so I guess I'd be fine with letting it be for now.

osu.Desktop/Windows/GameplayWinKeyHandler.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        private void toggleWinKey(ValueChangedEvent<bool> e)
+        {
+            var isPlayer = typeof(Player).IsAssignableFrom(currentScreenType) && currentScreenType != typeof(ReplayPlayer);
+
+            if (!e.NewValue && disableWinKey.Value && isPlayer)
+                host.InputThread.Scheduler.Add(WindowsKey.Disable);
+            else
+                host.InputThread.Scheduler.Add(WindowsKey.Enable);
```
smoogipoo, line 49: @peppy I'm letting you review this one, because this still feels wrong to me, and is the entire reason for my suggestion https://github.com/ppy/osu/pull/9439#discussion_r450615586, but it seems like you've gone back on that in your recent comments.

peppy: I don't think this conditional needs to exist. I was hoping we could implement in the most simplistic and isolated way to begin with, and revisit in the future.

basically hard coupling to `AllowScreenSuspection` for now. @smoogipoo you okay with that? If so I'll make changes.

smoogipoo: I'm fine with hard coupling yes.

PR #9440, opened 2020-07-06
Disable activity hardware acceleration on Android

Rsplwe:
The hardware acceleration only applies to the native UI and has no effect on osu, but it still uses resources and creates separate rendering threads that are slow and cause performance degradation.

### DIFF
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -9,7 +9,7 @@ using osu.Framework.Android;
 
 namespace osu.Android
 {
-    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullSensor, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = true)]
+    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullSensor, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false)]
     public class OsuGameActivity : AndroidGameActivity
     {
         protected override Framework.Game CreateGame() => new OsuGameAndroid();


bdach:
Can you provide any explanation for this change?

bdach:
I would like to see concrete evidence to the claims made in the edited PR description, via benchmarking, FPS numbers or otherwise.

peppy:
or even existing documentation or best practices guides that cite this. searching quickly it looks like it is recommended to leave this on unless there are incompatibilities in your 2d (app level) drawing code.

it is also the default and has been for some time.

Rsplwe:
> or even existing documentation or best practices guides that cite this. searching quickly it looks like it is recommended to leave this on unless there are incompatibilities in your 2d (app level) drawing code.
> 
> it is also the default and has been for some time.

Games published with Unity3D, AndroidManifest options are mostly Disable HardwareAccelerated

bdach:
Given that @Rsplwe is being very cagey with the responses above and still has not produced any substantial evidence backing the claims made, I did rudimentary research into this myself.

[There seems to be some merit](https://forum.unity.com/threads/android-hardwareaccelerated-is-forced-false-in-all-activities.532786/) to the claim that Unity disables hardware acceleratiion in their activities. However this is about the only piece of evidence I can find that would support the assertion that no accel = better for games.

I also made some notes of frame rates on various screens on my Samsung Galaxy J7 (2016). Values are ballpark and not measured very accurately, mostly taken at a period of relative stability after effects/transforms have stopped. I got this:

| | accel on | accel off |
| :- | :-: | :-: |
| Main menu, collapsed (big cookie) | ~45fps | ~47fps |
| Main menu, uncollapsed (buttons & toolbars showing) | ~40fps | ~38fps |
| Song select | ~15fps | ~15fps |
| Gameplay, standard settings | ~18fps | ~18fps |
| Gameplay, full dim, UI disabled | ~30fps | ~32fps |
| osu!direct opened at song select | ~10fps | ~9fps |

So at a naked eye's view, all within margin of error.

As to my opinion whether this is worth merging, currently I'm leaning "no".

Rsplwe:
@bdach My native language is not English, so the expression may be wrong

peppy:
As this does not show performance regression on either of our devices, I'm willing to take on the unity post as a valid reason to try this further.

Will hopefully get some user feedback as to whether this takes the game from unplayable->playable for any users.

PR #9444, opened 2020-07-07
Start track after seeking if it was not paused explicitly

### LINKED ISSUE #9398: Music controller stops playing after dragging the progress bar to the end
**Describe the bug:** After you drag the progress bar in music controller to the end and then leave it in a different spot, the music won't play anymore and the progress bar won't bar. To resume playing the music, you have to pause and unpause the music.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=76dNIyzTX3g

**osu!lazer version:** 2020.629.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/4845911/database.log)
[network.log](https://github.com/ppy/osu/files/4845914/network.log)
[performance.log](https://github.com/ppy/osu/files/4845916/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4845917/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4845918/updater.log)

UselessToucan:
resolves #9398 
The idea is that the track should be paused only if it was explicitly paused by the user before seeking.

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -129,7 +129,11 @@ namespace osu.Game.Overlays
             seekDelegate = Schedule(() =>
             {
                 if (!beatmap.Disabled)
+                {
                     current?.Track.Seek(position);
+                    if (!IsUserPaused)
+                        current?.Track.Start();
+                }
             });
         }
 


bdach:
I'm reasonably certain this solution is incorrect and in fact papering over a larger issue.

As per ppy/osu-framework#2499, BASS does not allow seeking to the end of the track and will both stop the track and return the `Position` error code when told to seek to `pos == Bass.ChannelGetLength(activeStream)`. This is reflected in the return value of [`TrackBass.SeekAsync()`](https://github.com/ppy/osu-framework/blob/6b257f4c2f771d70cacc2945bd7f6972c0fc5ed4/osu.Framework/Audio/Track/TrackBass.cs#L260-L280), which in this case will be false (and which the music controller happily ignores, and tries to seek to the very end anyway).

Now on one side the API seems to indicate that the music controller should be checking the return value and instead seeking back a little bit, but I don't think that's a reasonable expectation to be putting on a framework consumer. I think what should happen is that `SeekAsync` should be clamping the time offset to the last seekable position of the track and only *then* attempt to seek, but I'm not 100% on that.

bdach (changes requested):
As above.

peppy:
Sounds fair.

PR #9445, opened 2020-07-08
Fix seeding numbers being unreadable due to design changes

MiraiSubject:
Before:

dotnet_2020-07-08_03-29-41

After: 
dotnet_2020-07-08_03-29-29

The two colours previously being used were both white. This also makes the colouring consistent with `TournamentSpriteTextWithBackground`

Edit: just noticed my commit text says it the other way around, oops.


peppy:
Seeding screen hasn't been updated with new design logic, which is why this has gone unnoticed. I think it's fine to fix like this for now.

PR #9447, opened 2020-07-08
Make toolbar opaque

### LINKED ISSUE #6710: Retry hotkey overlay doesn't cover top toolbar
Opening mostly for tracking, as it's not an instantaneous fix.

**Describe the bug:** The hold-to-retry hotkey overlay doesn't cover the top toolbar, leading to a gap where the beatmap background shows despite the rest of the screen fading to black.

**Screenshots or videos showing encountered issue:** [See video](https://streamable.com/dku5b)

**osu!lazer version:** latest `master`, anything past merge of #6691 really

**Logs:** Not required - this is purely visual.

**Investigation:**
After looking into this in more detail the gap is caused by `OsuScreenStack` having padding for the top toolbar itself. This has already been a problem elsewhere, and I attempted to apply negative padding to the hotkey overlay in a similar manner to what `SettingsPanel` does:
https://github.com/ppy/osu/blob/f43e4bfd703d4cd3a9f747a7ead88ac4e690f08c/osu.Game/Overlays/SettingsPanel.cs#L183-L189
Unfortunately applying the same fix in the same way just makes the gap smaller and doesn't eliminate it:

image

Upon inspecting the visual hierarchy this seems to be caused by a `ScalingContainer` that makes the top bar's absolute size invalid in the context of the hotkey overlay lower down:

image

Notice the 40px discrepancy in height sizing of the first `ScalingContainer` and then the root `Container` for the other content. This 40px difference is the same regardless of scaling; adjusting for that manually seems to suggest this is the actual cause (as in calculating `768 / 728 * 40` in the above case yields about `42.19`, which in experiments seemed to be about as much of a difference as is needed to cover the entire screen).

As a final solution we might potentially want to look into moving the container somewhere higher up the hierarchy. This might also be resolved by a future design revision if the top bar is no longer transparent.

peppy:
- Closes #6710

This is the general direction we're going with future designs. Just applying this now because it makes a lot of screens feel much better (multiplayer lobby / match, song select etc. where there are elements
adjacent to the bar which cause the transparency to feel a bit awkward).

### DIFF
--- a/osu.Game/Overlays/Toolbar/Toolbar.cs
+++ b/osu.Game/Overlays/Toolbar/Toolbar.cs
@@ -28,9 +28,6 @@ namespace osu.Game.Overlays.Toolbar
 
         private const double transition_time = 500;
 
-        private const float alpha_hovering = 0.8f;
-        private const float alpha_normal = 0.6f;
-
         private readonly Bindable<OverlayActivation> overlayActivationMode = new Bindable<OverlayActivation>(OverlayActivation.All);
 
         // Toolbar components like RulesetSelector should receive keyboard input events even when the toolbar is hidden.
@@ -103,7 +100,6 @@ namespace osu.Game.Overlays.Toolbar
 
         public class ToolbarBackground : Container
         {
-            private readonly Box solidBackground;
             private readonly Box gradientBackground;
 
             public ToolbarBackground()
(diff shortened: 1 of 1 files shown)


bdach:
This will (presumably) close #6710, right? Can't check myself right now.

peppy:
I *think* this hides the issue pointed out, even when the toolbar is hidden... Probably best you double-check though.

bdach (approved):
seems good :+1:

PR #9448, opened 2020-07-08
Remove nesting of components inside overlined component

peppy:
I think this makes things a bit more readable. The only weird case is the transfer of details from the component to the `OverlinedHeader`, but bindables make it not too bad.

- [x] Depends on https://github.com/ppy/osu-framework/pull/3699


osu.Game/Screens/Multi/Match/MatchSubScreen.cs
```diff
@@ -126,9 +136,10 @@ private void load()
                                                                 {
                                                                     new Drawable[]
                                                                     {
-                                                                        new OverlinedPlaylist(true) // Temporarily always allow selection
```
smoogipoo, line 129: Header is missing

smoogipoo (approved):
Seems ok, added back the playlist header.

PR #9454, opened 2020-07-08
Remove slider tail circle judgement requirements

peppy:
In the opposite direction this time, as a trial. Tail circles no longer contribute to score, combo or accuracy. They still require tracking for health bonus and hitsound playback, which I think feels quite good.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/SliderTailCircle.cs
+++ b/osu.Game.Rulesets.Osu/Objects/SliderTailCircle.cs
@@ -4,6 +4,7 @@
 using osu.Framework.Bindables;
 using osu.Game.Rulesets.Judgements;
 using osu.Game.Rulesets.Objects;
+using osu.Game.Rulesets.Osu.Judgements;
 using osu.Game.Rulesets.Scoring;
 
 namespace osu.Game.Rulesets.Osu.Objects
@@ -24,6 +25,13 @@ namespace osu.Game.Rulesets.Osu.Objects
 
         protected override HitWindows CreateHitWindows() => HitWindows.Empty;
 
-        public override Judgement CreateJudgement() => new SliderRepeat.SliderRepeatJudgement();
+        public override Judgement CreateJudgement() => new SliderTailJudgement();
+
+        public class SliderTailJudgement : OsuJudgement
+        {
+            protected override int NumericResultFor(HitResult result) => 0;
+
+            public override bool AffectsCombo => false;
+        }
     }
 }


smoogipoo (approved):

veke66:
this is awesum 👍

MasterCritics:
I tested the changes somewhat extensively and now here's my verdict. 

Ultimately, it plays and feels the same way, and in some cases, feels better than the other end of the spectrum. I can *tap* rapid-fire sliders and not fear accidentally breaking my combo. The same goes to longer sliders with huge jumps to other sliders in half a beat - there's no fear of accidentally leaving the sliders too early to make the jump.

That said, I think not punishing for slider breaks would be a negative influence on learning players, because they can't learn what they can't get punished by. Learning players tend to break sliders more often than not and even in Stable they are only partly punished with a 100 and no combo. For this case my feelings lean towards the other end of the spectrum, where slider breaks penalise you completely, as in the long run it should cultivate good habits for most players in all skill levels. I also err on the side of not cheesing if we don't have to, and in most cases, even in slider-jump-heavy maps, can be done without the cheese.

On the other hand, there are some slider patterns (rapid-fire sliders and quick jumps) which become significantly more difficult to do if we judge sliders based on the previous configurations. It's does feel a little counterintuitive to hold down keys for rapid-fire sliders/slider streams, and I wouldn't try to play with the slider break leeway (30ms?) to hit my sliders. As for the jump sliders, it's a problem that I've had since Stable, but it wasn't so bad then since it doesn't punish you that much compared to Lazer. 

My proposal then would be to do it somewhere down the middle: Slider break registers as a miss and breaks combo (the previous iteration), but with a dynamic slider break leeway timing which is inversely affected by the timing between the end of the slider and the next note (with a min/max leeway timing). This idea could allow most sliders to be reasonably doable, and only punishes the player if he completely screws the slider up. And since players are still punished rather than "rewarded" for missing a slider end, learning players especially can adopt good slider habits early on in their osu! playtime.

peppy:
We are trying to make things simple and understandable. The above suggestion feels way too complex.

For transparency, the current plan is to add back tail judgements in the form of visible ticks. These can be set by the mapper, and will make it obvious when the player is supposed to hold until the end of the slider.

MasterCritics:
It sounds reasonable. I'll look forward to testing that one out

kitty-cloud:
> For transparency, the current plan is to add back tail judgements in the form of visible ticks. These can be set by the mapper, and will make it obvious when the player is supposed to hold until the end of the slider.

do you plan on having slider ends with these ticks break combo if missed, or will they be like how slider ends are judged now? if the former, this has some pretty large implications for mapping

peppy:
As above, it is not worth considering "implications for mapping" until it becomes an actual thing, implemented in an actual .osu version update alongside a client update.

kitty-cloud:
fair, but which of these is it planned to let mappers choose between?
1) tails not effecting score/combo/acc, and tails effecting score/combo/acc but _not_ breaking combo when missed
2) tails not effecting score/combo/acc, and tails effecting score/combo/acc _and_ breaking combo when missed

thanks for responding so fast, btw

peppy:
Please wait for implementation.

PR #9459, opened 2020-07-08
News overlay implementaton

EVAST9919:
Screenshot_1


osu.Game/Overlays/News/Displays/FrontpageDisplay.cs
```diff
@@ -0,0 +1,131 @@
...
+        private void load()
+        {
+            fetchPage();
+        }
+
+        private void fetchPage()
+        {
+            request = new GetNewsRequest(lastCursor);
+            request.Success += response => Schedule(() => createContent(response));
+            api.PerformAsync(request);
```
bdach, line 81: A little unnerved by the lack of checks for potential double requests here.

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -49,39 +53,62 @@ private void load(OsuColour colours)
...
+
+            if (current.NewValue == null)
+            {
+                LoadDisplay(new FrontpageDisplay());
+                return;
+            }
 
-            loadContentCancellation?.Cancel();
+            LoadDisplay(Empty());
+        }
```
bdach, line 96: So this stuff was here before, more or less, but I dislike it. Displaying the front page by setting a `null` value to a `string` bindable is not acceptable in my view (and why are both `ShowFrontPage()` and `Current` public?) I'd rather just have actual methods like `ShowFrontPage()` and `ShowArticle(string slug)` or something.

osu.Game/Overlays/News/Displays/FrontpageDisplay.cs
```diff
@@ -0,0 +1,131 @@
...
+        public FrontpageDisplay()
+        {
+            RelativeSizeAxes = Axes.X;
+            AutoSizeAxes = Axes.Y;
+            Padding = new MarginPadding
+            {
+                Top = 20,
+                Bottom = 10,
+                Left = 35,
+                Right = 55
```
bdach, line 36: as per web:
```suggestion
                Bottom = 20,
                Left = 30,
                Right = 50
```

osu.Game/Overlays/News/Displays/FrontpageDisplay.cs
```diff
@@ -0,0 +1,131 @@
...
+                        Direction = FillDirection.Vertical,
+                        Spacing = new Vector2(0, 10)
+                    },
+                    showMore = new FrontpageShowMoreButton
+                    {
+                        Anchor = Anchor.TopCentre,
+                        Origin = Anchor.TopCentre,
+                        Margin = new MarginPadding
+                        {
+                            Vertical = 15
```
bdach, line 62:
```suggestion
                            Top = 15
```

osu.Game/Overlays/News/Displays/FrontpageDisplay.cs
```diff
@@ -0,0 +1,131 @@
...
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            request?.Cancel();
+            cancellationToken?.Cancel();
+            base.Dispose(isDisposing);
+        }
+
+        private class FrontpageShowMoreButton : ShowMoreButton
```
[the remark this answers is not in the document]
bdach: also:

* lazer-side button has bold text, and web-side it doesn't
* lazer-side button has fixed 140px width, and web-side it has just 20px of padding on the sides

not 100% whether those are conscious decision or just design inconsistencies

osu.Game/Overlays/News/Displays/FrontpageDisplay.cs
```diff
@@ -0,0 +1,131 @@
...
+        {
+            [BackgroundDependencyLoader]
+            private void load(OverlayColourProvider colourProvider)
+            {
+                Height = 20;
+
+                IdleColour = colourProvider.Background3;
+                HoverColour = colourProvider.Background2;
```
bdach, line 126: should be
```suggestion
                IdleColour = colourProvider.Background2;
                HoverColour = colourProvider.Background1;
```

bdach (changes requested):
Please take more care when comparing against web layout & design. I don't really want to have to check every one of these for correctness.

osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
```diff
@@ -1,66 +1,27 @@
...
-                    new NewsArticleCover(article)
-                    {
-                        RelativeSizeAxes = Axes.None,
-                        Size = new osuTK.Vector2(400, 200),
-                    }
-                };
-            }
+            AddStep("Show front page", () => news.ShowFrontPage());
+            AddStep("Custom article", () => news.ShowArticle("Test Article 101"));
+            AddStep("Custom article", () => news.ShowArticle("Test Article 102"));
```
bdach, line 24: what are these steps even supposed to do now?

osu.Game.Tests/Visual/Online/TestSceneNewsOverlay.cs
```diff
@@ -1,66 +1,27 @@
...
-using osu.Framework.Graphics;
 using osu.Game.Overlays;
-using osu.Game.Overlays.News;
 
 namespace osu.Game.Tests.Visual.Online
 {
     public class TestSceneNewsOverlay : OsuTestScene
     {
-        private TestNewsOverlay news;
+        protected override bool UseOnlineAPI => true;
```
bdach, line 10: I would rather avoid having more scenes that rely on online API and instead use the dummy API to mock requests. We could have an online scene too but to me its usefulness is doubtful anyway as there aren't any asserts here at all.

osu.Game/Overlays/News/NewsHeader.cs
```diff
@@ -3,44 +3,46 @@
...
-                TabControl.AddItem(e.NewValue);
-                Current.Value = e.NewValue;
+                TabControl.AddItem(post.NewValue);
+                Current.Value = post.NewValue;
             }
             else
             {
-                Current.Value = front_page_string;
+                Current.Value = FRONT_PAGE_STRING;
             }
```
bdach, line 46: breadcrumb code should not be relying on a magic string to tell whether or not it should be going back to the front page. `null` was actually better than what is here currently but both ways are definitively worse than, again, not having a public bindable in which one magic value represents the front page, and instead actually having proper separate methods to show the front page and to show a single post.

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -7,81 +7,106 @@
...
                     },
                 },
+                loading = new LoadingLayer(content),
             };
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            header.Post.BindValueChanged(onPostChanged, true);
```
bdach, line 65: the fact that the *header* now controls the content display of its parent overlay is backwards. it should be the opposite (overlay controls content and reports changes to header, which is only for display).

bdach (dismissed):

smoogipoo:
May want to wait before going ahead with this, because flyte has different designs for the client's frontpage. I'm not sure what the plans are re: will osu-web eventually use the same design as the client's frontpage, etc.

@arflyte It's not a huge priority so either can happen, but do you think the client should use osu-web's frontpage for now or wait for your designs to be finished?

peppy:
It's still fine to get this in as-is. There is still a news listing page and the cards will not be changing.

bdach:
still don't like the string bindables and null juggling too much but at least it's private now. I imagine at some point it will probably get removed anyway, so I don't think I have major issues with this code anymore, but am not approving outright and would like a second opinion on how this PR looks.

EVAST9919:
Also we can probably add this to the game and make panels clickable with a link to the web-site until we have news page implemented

peppy:
I'd agree. Adding the news page itself can likely be delayed until a time in the future, as it requires a few features we don't have implemented yet (images in markdown, audio previews, etc.).

EVAST9919:
Okay, so now this overlay is in-game and clicking on panels will send you to the web-site

osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -36,6 +36,7 @@ public GlobalActionContainer(OsuGameBase game)
             new KeyBinding(new[] { InputKey.Control, InputKey.O }, GlobalAction.ToggleSettings),
             new KeyBinding(new[] { InputKey.Control, InputKey.D }, GlobalAction.ToggleDirect),
             new KeyBinding(new[] { InputKey.Control, InputKey.N }, GlobalAction.ToggleNotifications),
+            new KeyBinding(new[] { InputKey.Control, InputKey.A }, GlobalAction.ToggleNews),
```
bdach, line 39: <kbd>Ctrl</kbd>+<kbd>A</kbd> seems like a pretty random binding, and I'm honestly not convinced it's going to get used.

osu.Game/Overlays/News/NewsCard.cs
```diff
@@ -31,24 +31,28 @@ public class NewsCard : CompositeDrawable
...
 
-            NewsBackground bg;
+        [BackgroundDependencyLoader]
+        private void load(OverlayColourProvider colourProvider, GameHost host)
+        {
+            if (post.Slug != null)
+            {
+                TooltipText = "view in browser";
+                Action = () => host.OpenUrlExternally("https://osu.ppy.sh/home/news/" + post.Slug);
+            }
```
bdach, line 48: this has weird UX with the post date; if you hover over the date, the date details tooltip will replace the `view in browser` one, but clicking will still open the post in the browser. not sure what to do with that; I guess I'd be fine with the date suppressing clicks if it's deemed a problem.

osu.Game/Input/Bindings/GlobalActionContainer.cs
EVAST9919: I thought that every overlay has a shortcut, now removed

osu.Game/Overlays/News/NewsCard.cs
EVAST9919: Now date suppressing clicks

osu.Game/Input/Bindings/GlobalActionContainer.cs
Joehuu: It also conflicts with select all from textboxes, so if it does get a keybind, it can't be this.

osu.Game/Overlays/OverlayView.cs
```diff
@@ -18,6 +18,11 @@ namespace osu.Game.Overlays
     public abstract class OverlayView<T> : CompositeDrawable, IOnlineComponent
         where T : class
     {
+        /// <summary>
+        /// Whether we should perform fetch on api state change to online (true by default).
+        /// </summary>
+        protected virtual bool PerformFetchOnApiStateChange => true;
```
bdach, line 24: not sure I see the point. this flag essentially guts the only actual non-obvious shared logic `OverlayView` adds. the rest might as well be duplicated locally since it's obvious.

the problem here is trying to reuse `OverlayView` which is tied to login state, for something that *shouldn't* be tied to login state.

EVAST9919: I guess, making `FrontPageDisplay` inherit `OverlayView` was a bad decision?

bdach: that would be the logical conclusion to me, yes.

EVAST9919: Will change it back then

bdach:
Currently the fetching stops working if you log out. I assume the appropriate endpoints need to be marked public web-side (and I also assumed this was already the case from the `OverlayView` changes which were reverted yesterday), but I'd say it's a pretty huge blocker for this, as if you open the game while logged out, the news overlay will be empty, and if you log out at a later date and click "Load More", it will get stuck trying to load.

EVAST9919:
> Currently the fetching stops working if you log out. I assume the appropriate endpoints need to be marked public web-side (and I also assumed this was already the case from the `OverlayView` changes which were reverted yesterday), but I'd say it's a pretty huge blocker for this, as if you open the game while logged out, the news overlay will be empty, and if you log out at a later date and click "Load More", it will get stuck trying to load.

Created an [issue](https://github.com/ppy/osu-web/issues/6361) web-side

bdach (approved):
confirmed as working offline.

fifth time's the charm I guess, although probably this could use from having another pair of eyes look this through.

smoogipoo (approved):

peppy:
Was halfway through reviewing this. 

This is running requests on startup. Needs to be fixed before next release.

EVAST9919:
Will do rn

PR #9464, opened 2020-07-08
Fix spinner spinning in the wrong direction

### LINKED ISSUE #9455: Spinner ring (and its sprite) spins in the wrong direction
**Describe the bug:** As title. The inner ring (and its `*` sprite) spins in the opposite direction (when spinning counterclockwise).

(I used a random graveyard map with spinners for this gif)
spinner

**osu!lazer version:** 2020.707.0, but I tested this on the previous release ([2020.629.0](https://github.com/ppy/osu/releases/tag/2020.629.0)), and everything spins in the **correct** direction.

bdach:
Resolves #9455.

# Summary

So, this one is entirely on me.

As mentioned in the issue thread I broke it in #9434 by using `RotationAbsolute`. As it turns out `Rotation` *was* the correct thing to use. The reason why the symbol would pause at rewind were these two lines:

https://github.com/ppy/osu/blob/d229993e5c727ff9c10328a1360ad776606053b2/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs#L210-L211

The lerps in `SpinnerDisc` and `DrawableSpinner` during rewind wouldn't interpolate within the range of (rotation from last frame, rotation in current frame). They would instead interpolate within the range of (-720, rotation in current frame). This is why the rotation would seemingly pause.

You could ask however, how is it that when I changed to `RotationAbsolute` it looked less wrong. See, when you test with autoplay, it spins *counter-clockwise*, which means negative `Rotation`.

* When using `Rotation`, the old code would lerp from -720 to a negative number, hence the spinner would barely move.
* When using `RotationAbsolute`, the old code would lerp from -720 to a *positive* number, twice as big in absolute value as `Rotation`, hence the spinner would rotate faster due to a wider interpolation range, but *still* slower than normal autoplay, which is wrong, as it should spin *faster* than normal autoplay in rewind.

To resolve, remove the two offending lines and revert to using `Rotation` correctly.

[Here's a video of the corrected behaviour](https://drive.google.com/file/d/1QDg52R3TeuCwlOk6FeV2zIHs8KB_Grsb/view?usp=sharing). Note: the manual slow testing in both directions and the speed changes with autoplay (when scrubbing forward or backward through the replay the spinner *always* moves faster regardless of direction). Due to capture frame rate that is not too visible on the video but should be very visible locally.

# Remarks

Using `RotationAbsolute` was my mistake as I misunderstood what it was. I won't bother with explaining of what I thought it was as it makes no sense, but I added xmldoc so that hopefully the next person doesn't make the same mistake.

I've added some new asserts and a test. They cover both the fact that the lerps were broken, and spinning the spinner in both directions. The latter is hacky in a sense but it works and is not that long so I figure we might as well have it.

I should really have taken more time and care to examine what was going on with the rotations pausing in the previous PR instead of just eyeballing it.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
@@ -67,7 +67,7 @@ namespace osu.Game.Rulesets.Osu.Tests
                 if (auto && !userTriggered && Time.Current > Spinner.StartTime + Spinner.Duration / 2 && Progress < 1)
                 {
                     // force completion only once to not break human interaction
-                    Disc.RotationAbsolute = Spinner.SpinsRequired * 360;
+                    Disc.CumulativeRotation = Spinner.SpinsRequired * 360;
                     auto = false;
                 }
 
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
@@ -14,6 +14,7 @@ using osu.Game.Rulesets.Osu.Objects.Drawables;
 using osuTK;
 using System.Collections.Generic;
 using System.Linq;
+using osu.Framework.Graphics.Sprites;
 using osu.Game.Storyboards;
 using static osu.Game.Tests.Visual.OsuTestScene.ClockBackedTestWorkingBeatmap;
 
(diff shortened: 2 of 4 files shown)


frenzibyte:
To be fair, the property name for `RotationAbsolute` is kinda misleading reading it alone, would be nice if the name could be improved (I once came up with `BidirectionalRotation` but that actually is illogical), but a clarifying XMLDoc would do for now.

frenzibyte (dismissed):

peppy:
I've renamed the variable to CumulativeRotation to make it 100% obvious what the intent is.

peppy (approved):

frenzibyte (approved):

smoogipoo (approved):

PR #9467, opened 2020-07-09
Replace poo icon at disclaimer screen

peppy:
It was a placeholder as I couldn't find a more appropriate font-awesome icon, but due to way-too-many complaints about the poo I found a better one.


smoogipoo (approved):
I liked it but /shrug.

frenzibyte (approved):
This one looks fantastic though, good search 👌

PR #9471, opened 2020-07-09
Ensure directories are deleted before migration tests run

peppy:


bdach (approved):
resolves the problem for me and seems correct :+1:

smoogipoo (approved):
Seems fine

PR #9476, opened 2020-07-09
Scroll selected room into view on selection

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/9475

Adding for future keyboard support, but is even useful for click selection (if the room is half off-screen).

### DIFF
--- a/osu.Game.Tests/Visual/Multiplayer/RoomManagerTestScene.cs
+++ b/osu.Game.Tests/Visual/Multiplayer/RoomManagerTestScene.cs
@@ -1,0 +1,60 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Allocation;
+using osu.Game.Beatmaps;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Rulesets;
+using osu.Game.Screens.Multi;
+using osu.Game.Users;
+
+namespace osu.Game.Tests.Visual.Multiplayer
+{
+    public class RoomManagerTestScene : MultiplayerTestScene
+    {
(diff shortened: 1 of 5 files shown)


osu.Game.Tests/Visual/Multiplayer/RoomManagerTestScene.cs
```diff
@@ -0,0 +1,60 @@
...
+using osu.Framework.Allocation;
+using osu.Game.Beatmaps;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Rulesets;
+using osu.Game.Screens.Multi;
+using osu.Game.Users;
+
+namespace osu.Game.Tests.Visual.Multiplayer
+{
+    public class RoomManagerTestScene : MultiplayerTestScene
```
bdach, line 14: Shows up in the test scene right now - I suppose this was meant to be

```suggestion
    public abstract class RoomManagerTestScene : MultiplayerTestScene
```

right?

osu.Game.Tests/Visual/Multiplayer/TestSceneLoungeSubScreen.cs
```diff
@@ -0,0 +1,59 @@
...
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+        }
+
+        public override void SetUpSteps()
+        {
+            base.SetUpSteps();
+
+            AddStep("clear rooms", () => RoomManager.Rooms.Clear());
```
bdach, line 29: minor point, but

```suggestion
```

as it actually looks to fire twice:

2020-07-09-235204_228x393_scrot
2020-07-09-235324_491x148_scrot

bdach (commented):
A few nits to the tests, but otherwise looks good. Diff looks scary but only looks that way because of class extractions/renames/ctor -> `load()` transitions.

smoogipoo (approved):

PR #9477, opened 2020-07-09
Change multiplayer tests to have null room by default

peppy:
Since the lounge has a null room by default, I think this works better. Also removes unnecessary `CopyFrom` logic.


bdach (approved):
looks straightforward enough :+1:

smoogipoo (approved):

PR #9478, opened 2020-07-09
Add keyboard selection to multiplayer lounge

### LINKED ISSUE #9473: Browse using arrow key in multiplayer lounge
**Describe the new feature:** 
Just like in song selection, you can use arrow key up and down to browse every room in multiplayer lounge and press enter to enter the room.

peppy:
- Closes #9473
- [x] Depends on #9476
- [x] Depends on #9477 
- [x] Investigate fast up/down selection causing a crash

### DIFF
--- a/osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
+++ b/osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
@@ -19,24 +19,18 @@ namespace osu.Game.Tests.NonVisual
     [TestFixture]
     public class CustomDataDirectoryTest
     {
-        [SetUp]
-        public void SetUp()
-        {
-            if (Directory.Exists(customPath))
-                Directory.Delete(customPath, true);
-        }
-
         [Test]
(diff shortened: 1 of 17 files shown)


smoogipoo:
That was super dodgy - last commit I pushed fixes both the crash and duplicate playlist items (it seems).

smoogipoo:
Doesn't fix the duplicate items, definitely happens after a web req though. Continuing to look into it.

peppy:
Planning to make a framework change in line with that last "fix"? Seems flaky as it is.

smoogipoo:
What framework change are you proposing to make?

peppy:
I have no idea because I'm not sure what the fix is actually doing. But if `private set` is not valid in the `DependencyContainer` context due to breaking on use, it should likely be disallowed.

smoogipoo:
Fair point.

smoogipoo (approved):
Looks fine to me. Last issue is a framework one (see mentioning PR).

PR #9484, opened 2020-07-09
Add osu! playfield border display styles

### LINKED ISSUE #9198: Visible playfield boundary
**Describe the new feature:** 
An optional setting that when enabled will display a rectangle containing the playfield. This will enable players to make adjustments to their tablet play-area to fit the playfield.
Bonus points if it's also skinnable.

screenshot010

FlashyReese:
Preview: https://www.youtube.com/watch?v=9qHiQKU4NF4
Fixes #9198


osu.Game/Configuration/OsuConfigManager.cs
```diff
@@ -227,6 +230,8 @@ public enum OsuSetting
         IntroSequence,
         UIHoldActivationDelay,
         HitLighting,
-        MenuBackgroundSource
+        MenuBackgroundSource,
+        ShowPlayfieldArea,
+        PlayfieldAreaDimLevel
```
bdach, line 235: I'm not convinced this setting should exist. At least not until we add extended support for UI skinning which is a long way away. It's arguably out of scope of the linked issue anyway.

The name is misleading, too. It controls dim of the *border* and not the actual area inside.

osu.Game/Screens/Play/PlayerSettings/VisualSettings.cs
```diff
@@ -39,12 +41,21 @@ public VisualSettings()
                     DisplayAsPercentage = true
                 },
                 new OsuSpriteText
+                {
+                    Text = "Playfieldd area dim:"
```
bdach, line 45: if this does indeed stay
```suggestion
                    Text = "Playfield area dim:"
```

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -51,6 +53,10 @@ public abstract class Playfield : CompositeDrawable
         /// </summary>
         public readonly BindableBool DisplayJudgements = new BindableBool(true);
 
+        private Bindable<bool> showPlayfieldArea;
+        private Bindable<double> playfieldAreaDimLevel;
+        private EditorPlayfieldBorder playfieldArea;
```
bdach, line 58: Gameplay elements should not be using editor-specific components. If this is to be reused then `EditorPlayfieldBorder` should be renamed/moved to reflect its more general purpose.

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -76,6 +82,38 @@ private void load()
...
             }
+
+            showPlayfieldArea = config.GetBindable<bool>(OsuSetting.ShowPlayfieldArea);
+            playfieldAreaDimLevel = config.GetBindable<double>(OsuSetting.PlayfieldAreaDimLevel);
+            showPlayfieldArea.ValueChanged += _ => UpdateVisuals();
+            playfieldAreaDimLevel.ValueChanged += _ => UpdateVisuals();
+            UpdateVisuals();
+        }
+
+        protected virtual void UpdateVisuals()
```
bdach, line 93: why is this `protected virtual`?

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -76,6 +82,38 @@ private void load()
...
+        {
+            if (playfieldArea == null)
+            {
+                if (showPlayfieldArea.Value)
+                {
+                    AddInternal(playfieldArea = new EditorPlayfieldBorder
+                    {
+                        RelativeSizeAxes = Axes.Both,
+                        Alpha = (float)playfieldAreaDimLevel.Value,
+                    });
```
bdach, line 103: should not be dynamically added like this, but probably added once at the point of `load()` and just hidden/shown depending on the settings here.

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -76,7 +76,19 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Score display mode",
                     Bindable = config.GetBindable<ScoringMode>(OsuSetting.ScoreDisplayMode)
-                }
+                },
+                new SettingsCheckbox
+                {
+                    LabelText = "Show playfield area",
```
bdach, line 82: would probably reword to

```suggestion
                    LabelText = "Show playfield area bounds",
```

bdach (changes requested):
Thanks for the contribution, but this change has several shortcomings which need to be addressed before this is in a mergeable state.

I'm also not convinced this is necessarily the best moment for adding support for this (maybe better to hold off until UI skinning is the focus).

(Also, the area seems awfully small compared to what people use on stable... it does seem to be consistent with the editor, however)

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
peppy: i'd call it "playfield border"

osu.Game/Screens/Play/PlayerSettings/VisualSettings.cs
peppy: This should not be here. These settings are meant to be setting you regularly change. Which this is not.

peppy:
Not sure this should be implemented at a game level. It should probably be osu! ruleset. The way you've done it, it's going to show on other rulesets where it makes little sense.

cupppu:
Regardless of whether this should be implemented later until UI-skinning or ruleset-specific (which I agree), using the editor object placement border as the playfield boundary is not correct IMO

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -56,6 +62,19 @@ public OsuPlayfield()
...
+        private void load(OsuConfigManager config)
+        {
+            showPlayfieldBorder = config.GetBindable<bool>(OsuSetting.ShowPlayfieldBorder);
+            if (showPlayfieldBorder.Value)
+            {
+                AddInternal(new PlayfieldBorder
+                {
+                    RelativeSizeAxes = Axes.Both
+                });
+            }
```
bdach, line 75: We aim to have all settings take effect immediately, even during gameplay. The way this is done makes it so that if the setting is changed after gameplay starts, it won't take effect until a restart.

As I said before, please add the border unconditionally (with zero alpha), and then register a `BindValueChanged(..., true);` callback in `loadComplete()` to toggle visibility of the playfield border upon any setting change.

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -76,6 +76,11 @@ private void load(OsuConfigManager config)
                 {
                     LabelText = "Score display mode",
                     Bindable = config.GetBindable<ScoringMode>(OsuSetting.ScoreDisplayMode)
+                },
+                new SettingsCheckbox
+                {
+                    LabelText = "Show playfield border",
+                    Bindable = config.GetBindable<bool>(OsuSetting.ShowPlayfieldBorder)
                 }
```
bdach, line 84: Seeing that this is now a osu!-specific setting the location of it doesn't make much sense. This should be in `OsuSettingsSubsection`.

osu.Game/Screens/PlayfieldBorder.cs
```diff
@@ -6,14 +6,14 @@
 using osu.Framework.Graphics.Shapes;
 using osuTK.Graphics;
 
-namespace osu.Game.Screens.Edit.Compose.Components
+namespace osu.Game.Screens
```
bdach, line 9: This is not a good namespace for this component. I would suggest

```suggestion
namespace osu.Game.Rulesets.UI
```

bdach (dismissed):

peppy:
Adjusted this slightly, to expose a couple of styles* for the borders. The main reason for doing this was to avoid the overdraw incurred by having a huge transparent box with a border.

* not final designs. needs further work to avoid overlaps in the corners at very least, so allow for better fading / additive usage.

peppy (dismissed):

bdach (dismissed):
looks cool 👍

peppy (approved):

PR #9488, opened 2020-07-10
Make default skin cursor more visible against bright backgrounds

peppy:
- Adds a dark colour to the centre (similar to stable) to make the default cursor more usable.
- Fixes cursor scale potentially not being read (only happens in tests).
- Fixes cursor tests not correctly showing transforms due to calling `OnPressed`/`OnReleased` too often.
- Adds tests against random background colours (and user input driven tests).
- [x] Depends on https://github.com/ppy/osu-framework/pull/3706

This is not a final design, just a temporary usability improvement until we have a more definitive cursor design.


smoogipoo (approved):
lgtm, can read much better with this

PR #9491, opened 2020-07-10
Implement multiplayer categories

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-web/pull/6278

Required some changes around `SearchableListFilterControl`, which is only used in two places - here and `SocialOverlay` (woops, we've forgotten to remove it).


peppy:
Seems okay. I find the location a bit hidden but it'll do for now. Maybe we can decorate the spotlight room panels with a border?

smoogipoo:
We couldm that would require them to be included in the "Normal" category, or for there to be an "all" request option. cc @nanaya

PR #9493, opened 2020-07-10
Highlight spotlight timeshift rooms with special colour

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/9491


smoogipoo:
Fails on creating room with:
```
2020-07-10 10:47:39 [verbose]: ---> System.ArgumentNullException: Value cannot be null. (Parameter 'value')
2020-07-10 10:47:39 [verbose]: at System.Enum.ToObject(Type enumType, Object value)
2020-07-10 10:47:39 [verbose]: at Newtonsoft.Json.Serialization.JsonSerializerInternalReader.EnsureType(JsonReader reader, Object value, CultureInfo culture, JsonContract contract, Type targetType)
```
Is this what's expected - to have a nullable value here? I'd expect to receive "normal".

peppy:
Normal sounds better as a default, yep.

smoogipoo (approved):

PR #9494, opened 2020-07-10
Replace large "show results" button with embedded button each playlist item

peppy:
image

### DIFF
--- a/osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
+++ b/osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
```diff
@@ -91,6 +96,8 @@ protected override void LoadComplete()
 
         private ScheduledDelegate scheduledRefresh;
 
+        public FillFlowContainer ButtonsContainer { get; private set; }
```
smoogipoo, line 99: Maybe `private` instead?

smoogipoo (changes requested):
looks fine otherwise

osu.Game/Screens/Multi/DrawableRoomPlaylistItem.cs
peppy: right. i tried to do this another way then backtracked.

smoogipoo (approved):

PR #9497, opened 2020-07-10
Fix null reference if hit lighting is disabled

peppy:


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuJudgement.cs
```diff
@@ -60,10 +60,10 @@ protected override void PrepareForUse()
 
             lightingColour?.UnbindAll();
 
-            if (JudgedObject != null)
+            if (JudgedObject != null && lighting != null)
             {
                 lightingColour = JudgedObject.AccentColour.GetBoundCopy();
-                lightingColour.BindValueChanged(colour => lighting.Colour = Result.Type == HitResult.Miss ? Color4.Transparent : colour.NewValue, true);
+                lightingColour.BindValueChanged(colour => lighting.Colour = Result?.Type == HitResult.Miss ? Color4.Transparent : colour.NewValue, true);
```
smoogipoo, line 66: Nullable result should be impossible (and asserted by base).

smoogipoo (changes requested):

smoogipoo (approved):

PR #9503, opened 2020-07-10
Sort beatmaps by date submitted

### LINKED ISSUE #9468: Add the ability to sort songs by date uploaded
**Add the ability to sort songs by date created.**

I have always wanted to do this on the original osu! client, but from what I see, only "date added", which I am sure refers to when you downloaded the song, is even remotely close to this feature. I sometimes want to go back and play really old maps just to get a taste for certain years, and for people like WubWoofWolf who plays extremely old maps all the time, this feature would probably be neat. 

It would simply just go next to all the other sorting options, so no need to send in a draft. I don't know how technical you could get with this without making it too over the top, but sorting by years seems to be the most simple way to do it - and then sorting by months as sub categories within the years, while maybe being too much, I think would make it a lot better. This would look similar to sorting by difficulty, which is already in the game.

vruss:
Resolves #9468 
This PR is still WIP. 

Tests for sorting are passing but sorting real data isn't working due to the `Submitted` variable in `BeatmapSetInfo` is set to the same value on all maps. I can't figure out how to properly update this variable.


mcendu:
What is the intended behavior if the collection has unsubmitted sets?

vruss:
I'd suggest placing them last in Titel order or something like that.

peppy:
What's the point of this PR if it doesn't work?

peppy:
Please reopen if you actually make it work. You will need to look into EF core migrations, reading from the online database (to get correct data in for stable data migration etc.)

Issue #9502, opened 2020-07-10
Add a hotkey to open notifications sidebar


vntxx:
**Describe the new feature:** 
Add the hotkey Ctrl+N (ambiguous with the known "new" command, so I'd suggest something else) to open the notifications sidebar similar to how you can open the options sidebar with Ctrl+O.

**Why should this be considered ?** 
Convenience. Just like the options hotkey.

*Edit*: minor spelling mistakes

PR #9517, opened 2020-07-10
Add notifications keybinding

vntxx:
Implementation of #9502

### DIFF
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -35,6 +35,7 @@ namespace osu.Game.Input.Bindings
             new KeyBinding(new[] { InputKey.Control, InputKey.T }, GlobalAction.ToggleToolbar),
             new KeyBinding(new[] { InputKey.Control, InputKey.O }, GlobalAction.ToggleSettings),
             new KeyBinding(new[] { InputKey.Control, InputKey.D }, GlobalAction.ToggleDirect),
+            new KeyBinding(new[] { InputKey.Control, InputKey.N }, GlobalAction.ToggleNotifications),
 
             new KeyBinding(InputKey.Escape, GlobalAction.Back),
             new KeyBinding(InputKey.ExtraMouseButton1, GlobalAction.Back),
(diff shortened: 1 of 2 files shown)


osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -157,5 +158,8 @@ public enum GlobalAction
 
         [Description("Home")]
         Home,
+
+        [Description("Toggle notifications")]
+        ToggleNotifications
```
frenzibyte, line 163: This should reside right after `ToggleDirect` above, to respect the order given to this enum.

frenzibyte (commented):
Thanks for the contribution, the code looks pretty good, just a small nit however

osu.Game/Input/Bindings/GlobalActionContainer.cs
vntxx: I actually had it there, but in that order the hotkeys will switch allocations. Basically it won't migrate well from an older version. So I guess having it down there is messing with the logical order but it can't be helped considering how the values are stored.

vntxx: I would implement some kind of migration if the project has this kind of system in place. Just give me a hint on how to do it and I will take care of it.

bdach: @vntxx is correct, this has to be last for now to avoid breaking existing keybindings.

We will eventually be moving away from storing enums as integer values in the database but that change will be done globally and is out of scope here.

frenzibyte: Yeah I see, my bad not knowing about that.

bdach:
This seems fine. The only annoying part is the enum value ordering issue but I'd say we can hold off on that for now and fix it at a later date when we're migrating the entire enum. Would like further opinions on that, though.

bdach:
there is no need to rebase this PR onto `master`. please refrain from force-pushing in general, if this is decided to be good to go one of the maintainers will merge `master` back here and merge this into `master` afterward without squashing/rebasing.

vntxx:
Sure. Just had mistakenly merged master which created an unnecessary merge commit.

bdach:
there is no reason to merge `master` either, just let it be unless there are conflicts to resolve (but once that got pushed out then you might as well have left it there).

PR #9522, opened 2020-07-11
Update comment buttons to match web

### LINKED ISSUE #9511: Update comment replies button to match web
lazer: 
image
web:
image

EVAST9919:
Closes https://github.com/ppy/osu/issues/9511
Screenshot_1

### DIFF
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
@@ -33,7 +33,10 @@ namespace osu.Game.Tests.Visual.UserInterface
                 Children = new Drawable[]
                 {
                     button = new TestButton(),
-                    new LoadRepliesButton(),
+                    new LoadRepliesButton
+                    {
+                        Action = () => { }
+                    },
                     new ShowRepliesButton(1),
                     new ShowRepliesButton(2)
                 }
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
@@ -1,0 +1,54 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Game.Overlays.Comments.Buttons;
+using osu.Framework.Graphics;
+using osu.Framework.Allocation;
+using osu.Game.Overlays;
+using osu.Framework.Graphics.Containers;
+using osuTK;
+
+namespace osu.Game.Tests.Visual.UserInterface
+{
+    public class TestSceneCommentRepliesButton : OsuTestScene
+    {
+        [Cached]
+        private readonly OverlayColourProvider colourProvider = new OverlayColourProvider(OverlayColourScheme.Blue);
+
+        public TestSceneCommentRepliesButton()
+        {
+            Child = new FillFlowContainer
+            {
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+                AutoSizeAxes = Axes.Both,
+                Direction = FillDirection.Vertical,
+                Spacing = new Vector2(0, 10),
+                Children = new Drawable[]
+                {
+                    new TestButton
+                    {
+                        Action = () => { }
+                    },
+                    new LoadRepliesButton
+                    {
+                        Action = () => { }
+                    },
+                    new ShowRepliesButton(1)
+                    {
+                        Action = () => { }
+                    },
+                    new ShowRepliesButton(2)
+                    {
+                        Action = () => { }
+                    }
+                }
+            };
+        }
+
+        private class TestButton : CommentRepliesButton
+        {
+            protected override string GetText() => "sample text";
+        }
+    }
+}
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
@@ -26,22 +26,10 @@ namespace osu.Game.Tests.Visual.UserInterface
                 Spacing = new Vector2(0, 10),
                 Children = new Drawable[]
                 {
-                    new TestButton
-                    {
-                        Action = () => { }
-                    },
-                    new LoadRepliesButton
-                    {
-                        Action = () => { }
-                    },
-                    new ShowRepliesButton(1)
-                    {
-                        Action = () => { }
-                    },
+                    new TestButton(),
+                    new LoadRepliesButton(),
+                    new ShowRepliesButton(1),
                     new ShowRepliesButton(2)
-                    {
-                        Action = () => { }
-                    }
                 }
             };
         }
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
@@ -10,6 +10,7 @@ using osuTK;
 using NUnit.Framework;
 using System.Linq;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Testing;
 
 namespace osu.Game.Tests.Visual.UserInterface
 {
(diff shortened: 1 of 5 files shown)


osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -449,5 +438,31 @@ private string getParentMessage()
                 return parentComment.HasMessage ? parentComment.Message : parentComment.IsDeleted ? @"deleted" : string.Empty;
             }
         }
+
+        private class Loading : Container
```
bdach, line 442: does this really need to be a dedicated class? couldn't you just have the `Container` in the hierarchy in `load()` and just operate on the container itself instead of adding all this?

osu.Game/Overlays/Comments/Buttons/ShowRepliesButton.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            Expanded.BindValueChanged(onExpandedChanged, true);
+        }
+
+        private void onExpandedChanged(ValueChangedEvent<bool> expanded)
+        {
+            Icon.ScaleTo(new Vector2(1, expanded.NewValue ? -1 : 1));
```
bdach, line 31: you could move this to the base (as a protected method you can call, like `ToggleIcon()`) and therefore not have to expose `Icon` as protected.

bdach (changes requested):

osu.Game/Overlays/Comments/Buttons/CommentRepliesButton.cs
```diff
@@ -0,0 +1,113 @@
...
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Input.Events;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
+using osu.Game.Graphics.UserInterface;
+using osuTK;
+
+namespace osu.Game.Overlays.Comments.Buttons
+{
+    public abstract class CommentRepliesButton : CompositeDrawable
```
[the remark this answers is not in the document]
EVAST9919: We have multiple colour changes here, not only for background. Anyway we will override `OnHover` and `OnHoverLost`

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: Loading is always hidden on start and I didn't wan to add more stuff to `DrawableComment`

osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
```diff
@@ -0,0 +1,42 @@
...
+using osu.Game.Overlays.Comments.Buttons;
+using osu.Framework.Graphics;
+using osu.Framework.Allocation;
+using osu.Game.Overlays;
+using osu.Framework.Graphics.Containers;
+using osuTK;
+
+namespace osu.Game.Tests.Visual.UserInterface
+{
+    public class TestSceneCommentRepliesButton : OsuTestScene
```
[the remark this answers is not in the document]
EVAST9919: There's nothing really to test, but, as being said, each component should have a test scene.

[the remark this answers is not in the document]
bdach: now that you moved the icon toggling icon to the base there is arguably something to test.

osu.Game.Tests/Visual/UserInterface/TestSceneCommentRepliesButton.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            AddStep("Toggle icon up", () => button.ToggleIcon(true));
+            AddAssert("Icon facing upwards", () => button.Icon.Scale.Y == -1);
+            AddStep("Toggle icon down", () => button.ToggleIcon(false));
+            AddAssert("Icon facing downwards", () => button.Icon.Scale.Y == 1);
+        }
+
+        private class TestButton : CommentRepliesButton
+        {
+            public SpriteIcon Icon => InternalChildren.OfType<CircularContainer>().First().Children.OfType<Container>().First().Children.OfType<FillFlowContainer>().First().Children.OfType<SpriteIcon>().First();
```
bdach, line 53: this is horrendous.

as a heads up, `ChildrenOfType<T>` works recursively.

```suggestion
            public SpriteIcon Icon => this.ChildrenOfType<SpriteIcon>().First();
```

EVAST9919: Yes, but `CompositeDrawable` don't have this method. Well, we can use `Container` then, but I'm not sure

bdach: i *literally* just checked that method and it worked. it's an extension method.

EVAST9919: Well, I'm just being sleepy, you are right

osu.Game/Overlays/Comments/DrawableComment.cs
bdach: I still don't think splitting this out is worth it instead of just inlining the `LoadingSpinner` inside of `DrawableComment` as this is just one tree level extra. the new fillflow added also in this PR is more of a concern to me due to the depth of the hierarchy in that spot.

osu.Game/Overlays/Comments/Buttons/CommentRepliesButton.cs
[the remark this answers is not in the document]
bdach: you wouldn't have to manually add the `HoverClickSounds` that way, but sure, I suppose.

osu.Game/Overlays/Comments/Buttons/CommentRepliesButton.cs
```diff
@@ -0,0 +1,123 @@
...
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            background.Colour = colourProvider.Background2;
+            icon.Colour = colourProvider.Foreground1;
+        }
+
+        protected void ToggleIcon(bool upwards) => icon.ScaleTo(new Vector2(1, upwards ? -1 : 1));
```
peppy, line 100: this isn't a toggle but a `Set` or something

osu.Game/Overlays/Comments/DrawableComment.cs
```diff
@@ -339,14 +365,16 @@ private void updateButtonsState()
...
-            repliesButton.FadeTo(loadedReplesCount != 0 ? 1 : 0);
+            showRepliesButton.FadeTo(loadedReplesCount != 0 ? 1 : 0);
 
             if (Comment.IsTopLevel)
                 chevronButton.FadeTo(loadedReplesCount != 0 ? 1 : 0);
 
-            showMoreButton.IsLoading = loadMoreCommentsButton.IsLoading = false;
+            showMoreButton.IsLoading = false;
+            loading.Hide();
+            loading.FinishTransforms();
```
peppy, line 377: is this to make it go away immediately?

peppy:
I feel like this will look/work better if the loading spinner is inside the "show replies" button. Right now the spinner is in the middle of nowhere, aligned in a very weird way.

osu.Game/Overlays/Comments/DrawableComment.cs
EVAST9919: right

EVAST9919:
> I feel like this will look/work better if the loading spinner is inside the "show replies" button. Right now the spinner is in the middle of nowhere, aligned in a very weird way.

I made it to look like in web, but sure, will do

EVAST9919:
> I feel like this will look/work better if the loading spinner is inside the "show replies" button. Right now the spinner is in the middle of nowhere, aligned in a very weird way.

Reworked, looks much better

osu.Game/Overlays/Comments/Buttons/ShowRepliesButton.cs
```diff
@@ -0,0 +1,31 @@
...
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            Expanded.BindValueChanged(expanded => SetIconDirection(expanded.NewValue), true);
+        }
+
+        protected override bool OnClick(ClickEvent e)
+        {
+            Expanded.Toggle();
+            return base.OnClick(e);
```
bdach, line 28: nitpick: this should be

```suggestion
            return true;
```

to prevent potential propagation up the tree (otherwise this delegates to `Drawable.Handle(UIEvent)` which is `false` by default).

bdach (approved):
I *think* this is ok now...?

smoogipoo (approved):
lgtm

PR #9533, opened 2020-07-11
Changed DatabasedKeyBinding to use string instead of int for action storage

vntxx:
Allows #9517 to move the enum value up to ensure logical order.
Also this might break everything and is not quite ready yet as the migration is missing, thus I am opening this PR as a draft first.
If this concept is completely out of line feel free to close this completely.
At least I really had fun getting into the osu code and exploring various concepts of c# :)

### DIFF
--- a/osu.Game/Database/OsuDbContext.cs
+++ b/osu.Game/Database/OsuDbContext.cs
@@ -134,7 +134,7 @@ namespace osu.Game.Database
             modelBuilder.Entity<SkinInfo>().HasIndex(b => b.DeletePending);
 
             modelBuilder.Entity<DatabasedKeyBinding>().HasIndex(b => new { b.RulesetID, b.Variant });
-            modelBuilder.Entity<DatabasedKeyBinding>().HasIndex(b => b.IntAction);
+            modelBuilder.Entity<DatabasedKeyBinding>().HasIndex(b => b.StringAction);
 
             modelBuilder.Entity<DatabasedSetting>().HasIndex(b => new { b.RulesetID, b.Variant });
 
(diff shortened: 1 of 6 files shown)


bdach:
Please don't open pull requests for functionality that is not yet ready. Draft PRs are for things that are *complete*, but require discussion as to the direction taken within them.

Additionally if this is to be completed, tests demonstrating the correctness of the migration would be very welcome.

I will close this for now; if you come to a state in which this is functionally complete, feel free to open this again.

vntxx:
The core functionality is actually already in place and needs no further adjustments (as of now). It is just quite impossible to migrate the integer values into the new string values as this would require the correct order of the enum (which requires old game code if changed).
If there is a possibility to overcome this issue, I will finalize this pr immediately.

bdach:
I'm not sure I see why the migration would be "impossible". The way I see it, we should be using the current enum values to the migration (so ship some piece of code that will perform the appropriate adjustments to the database model), leave it there to be for a couple of releases, and then break compatibility after a reasonable period as to not have to maintain that code path indefinitely.

Now the migration part is not trivial (how and when it should be ran?), but hardly impossible, and seeing that it's the core part that is needed for it to work, I can't see any PR that doesn't address that part as functionally complete.

vntxx:
I added a first iteration of the migration. Two options to be exact and one has to go. The first implementation uses the RulesetStore to load existing Rulesets and migrate everything. That might actually not be possible though because there is no way to inject parameters into the Migration class. The second implementation comes close to what you had in mind. Current values of the 4 available default rulesets are implemented as a lookup. This would actually work but it doesn't support third party ruleset migration. I'm starting to realize that i may be easier to drop this pr and leave the order of the enum (which was the issue in the first place, but this concept generally provides a way to switch them up however you'd like) as it is, as I am currently out of ideas and I many don't have enough experience to implement something that can handle each scenario.

vntxx:
It seems like the commit isn't listed because the PR is closed. If you like to, have a look at it [here](https://github.com/vntxx/osu/commit/81bde46b917ca88e60ab8f2041000e03bf488eb6).

smoogipoo:
You'll need to re-open the PR. Not sure why it was closed, we don't usually close PRs even if they're drafts.

PR #9534, opened 2020-07-12
Fix color and underline of tab control checkboxes when initially checked

Joehuu:
After going to song select with show converted beatmaps enabled:

| Before | After |
| --- | --- |
| dotnet_cE9oVCb6mq | dotnet_ZkYdW942Rr |

The `updateFade` addition after value changed is actually a side change for when the setting is changed elsewhere. It is needed because on hover won't happen.

### DIFF
--- a/osu.Game/Graphics/UserInterface/OsuTabControlCheckbox.cs
+++ b/osu.Game/Graphics/UserInterface/OsuTabControlCheckbox.cs
@@ -21,7 +21,6 @@ namespace osu.Game.Graphics.UserInterface
     {
         private readonly Box box;
         private readonly SpriteText text;
-        private readonly SpriteIcon icon;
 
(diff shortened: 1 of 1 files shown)


peppy:
I'm confused as to where the icon is getting its colour updated now.

Joehuu:
The icon's color should've never been updated. The only other usage of this checkbox is the `selected mods` and the color never changed (it's always white), unless it was always broken.

The code was never reached before because `selected mods` is always initially off. Now it has because the recent `show converts` toggle can be initially on.

peppy:
Right, I see. Got confused by your screenshots (I thought it was showing unchecked/checked, not before/after).

peppy (commented):
Looks to match expectations

peppy (approved):

PR #9536, opened 2020-07-12
Add ability to select mods from a specific score in song select

### LINKED ISSUE #9481: "Use these mods" option when right-clicking a score
**Describe the new feature:** As the title suggests, if someone wants to re-use the mods of a score (made by himself or someone else in any leaderboard) without having to manually select them.

The user will have to right-click in the mods area of a score (or in any position).
I would also suggest renaming "delete" to "delete score" to make it clear what it refers to.

Use These Mods concept

And in the mods button of the footer, you can see those mods selected from the score (previously selected mods will be deselected):

Mods

This feature could may also automatically select the customisations of each mod from the score (but not for now: see #8569)

Joehuu:
Closes https://github.com/ppy/osu/issues/9481.

### DIFF
--- a/osu.Game/Online/Leaderboards/Leaderboard.cs
+++ b/osu.Game/Online/Leaderboards/Leaderboard.cs
@@ -170,36 +170,36 @@ namespace osu.Game.Online.Leaderboards
         {
             InternalChildren = new Drawable[]
             {
-                new GridContainer
+                new OsuContextMenuContainer
                 {
                     RelativeSizeAxes = Axes.Both,
-                    RowDimensions = new[]
+                    Child = new GridContainer
                     {
-                        new Dimension(),
-                        new Dimension(GridSizeMode.AutoSize),
-                    },
-                    Content = new[]
-                    {
-                        new Drawable[]
+                        RelativeSizeAxes = Axes.Both,
+                        RowDimensions = new[]
                         {
-                            new OsuContextMenuContainer
+                            new Dimension(),
+                            new Dimension(GridSizeMode.AutoSize),
+                        },
+                        Content = new[]
+                        {
+                            new Drawable[]
                             {
-                                RelativeSizeAxes = Axes.Both,
-                                Child = scrollContainer = new OsuScrollContainer
+                                scrollContainer = new OsuScrollContainer
                                 {
                                     RelativeSizeAxes = Axes.Both,
                                     ScrollbarVisible = false,
                                 }
+                            },
+                            new Drawable[]
+                            {
+                                content = new Container
+                                {
+                                    AutoSizeAxes = Axes.Y,
+                                    RelativeSizeAxes = Axes.X,
+                                },
                             }
                         },
-                        new Drawable[]
-                        {
-                            content = new Container
-                            {
-                                AutoSizeAxes = Axes.Y,
-                                RelativeSizeAxes = Axes.X,
-                            },
-                        }
                     },
                 },
                 loading = new LoadingSpinner(),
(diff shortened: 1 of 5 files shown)


osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -373,11 +376,21 @@ public MenuItem[] ContextMenuItems
...
 
                 return items.ToArray();
             }
         }
+
+        private void getMods()
+        {
+            songSelect.ModSelect.DeselectAll(true);
+
+            songSelect.Mods.Value = score.Mods;
```
bdach, line 393: This is DI'd as `CanBeNull = true` but not null-checked.

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -58,6 +58,9 @@ public class LeaderboardScore : OsuClickableContainer, IHasContextMenu
         [Resolved(CanBeNull = true)]
         private DialogOverlay dialogOverlay { get; set; }
 
+        [Resolved(CanBeNull = true)]
+        private SongSelect songSelect { get; set; }
```
bdach, line 62: I'm very not much a fan of mod changes being done through `SongSelect` - it necessitates a lot of introductions of the `immediate` flag and the exposition of `ModSelect`. I would instead propose to inject

```csharp
[Resolved(CanBeNull = true)]
private Bindable<IReadOnlyList<Mod>> selectedMods { get; set; }
```

directly from `OsuGame` which seems to work as well. Not sure whether there are any particular reasons against doing that, though.

bdach (changes requested):

osu.Game/Online/Leaderboards/LeaderboardScore.cs
Joehuu: Actually I don't need to expose `ModSelect` and do any of the deselection stuff. `songSelect.Mods.Value = score.Mods` already does replace the old mods.

bdach: that works too, I suppose.

osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -170,36 +170,36 @@ protected Leaderboard()
         {
             InternalChildren = new Drawable[]
             {
-                new GridContainer
+                new OsuContextMenuContainer
```
smoogipoo, line 173: This should be inside the scroll container instead, otherwise the menu appears outside the masking area after a scroll:

image

smoogipoo (changes requested):

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -373,11 +376,20 @@ public MenuItem[] ContextMenuItems
...
+                    items.Add(new OsuMenuItem("Use these mods", MenuItemType.Highlighted, getMods));
+
                 if (score.ID != 0)
                     items.Add(new OsuMenuItem("Delete", MenuItemType.Destructive, () => dialogOverlay?.Push(new LocalScoreDeleteDialog(score))));
 
                 return items.ToArray();
             }
         }
+
+        private void getMods()
```
smoogipoo, line 389: Probably unnecessary to be extracted into a method, and the menu item should not be added in the first place if `songSelect` is null.

smoogipoo (changes requested):

osu.Game/Online/Leaderboards/Leaderboard.cs
Joehuu: Need the user top score container to be inside too, so just adding `Masking = true` to `OsuContextMenuContainer` works.

smoogipoo: Sounds fine ^^

smoogipoo (approved):
Looks good

PR #9538, opened 2020-07-12
Implement "Fade In" and "Hidden" mod for osu!mania

### LINKED ISSUE #6711: Add osu!mania Fade-in & Hidden mod support
Probably a "good first issue": Implement FI and HD for mania which are the last mods from stable missing in lazer (as soon as #4756 gets merged). Tracking both mods in the same issue as they are very similar and their implementations will likely be tightly related.

For design decisions, refer to #3687

LastExceed:
closes #6711 

- [x] Depends on #9571


osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +29,86 @@ public class ManiaModFadeIn : Mod
...
+        private const float lanecover_size_filled = 0.5f;
+        private const float lanecover_size_gradient = 0.25f;
+
+        public void ApplyToDrawableRuleset(DrawableRuleset<ManiaHitObject> drawableRuleset)
+        {
+            ManiaPlayfield maniaPlayfield = (ManiaPlayfield)drawableRuleset.Playfield;
+
+            foreach (Column column in maniaPlayfield.Stages.SelectMany(stage => stage.Columns))
+            {
+                column.hitObjectArea.ChildrenOfType<BufferedContainer>().First().Add(new LaneCover(false)
```
LastExceed, line 42: I am well aware that this particular line is awful, I just couldn't figure out how else to do it. Please advise

bdach:
On a rudimentary look, bar lines seem to be exempt from the fade-in effect, while on stable they don't seem to be:

osu_2020-07-12_14-23-10

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
bdach: * should be a `Single()` if anything to begin with
* should not use `ChildrenOfType()` but a named field to which the buffered container is assigned

that said, I would explore options as to whether the buffered container can be slotted in solely for the fade-in mod, instead of being added globally as done here.

LastExceed:
This is intended. Stable uses a black shadow that covers everything (also the background while we're at it), while this implementation just makes the notes invisible.

bdach:
Not a mania player really but doesn't not hiding the bar lines make things easier in any capacity?

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +29,86 @@ public class ManiaModFadeIn : Mod
         public override double ScoreMultiplier => 1;
         public override bool Ranked => true;
         public override Type[] IncompatibleMods => new[] { typeof(ModFlashlight<ManiaHitObject>) };
+
+        private const float lanecover_size_filled = 0.5f;
+        private const float lanecover_size_gradient = 0.25f;
```
bdach, line 34: don't declare single-use constants. just inlining is fine.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +29,86 @@ public class ManiaModFadeIn : Mod
...
+            }
+        }
+
+        private class LaneCover : CompositeDrawable
+        {
+            private readonly Box gradient;
+            private readonly Box filled;
+            private readonly bool reversed;
+
+            public LaneCover(bool reversed)
```
bdach, line 57: this breaks if you change mania scrolling direction mid-gameplay. I assume this is what you intended the `reversed` ctor parameter for but it is only ever invoked with `false` as the value.

bdach (changes requested):

LastExceed:
As [the world's only FI player](https://youtu.be/0ocSADb4gTg) I can assure you: no it doesn't. The lack of combo scaling however certainly does so that is something I definitely want to address when I get the information needed.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
mcendu: You can bind the scrolling direction to the stage (maybe look at the implementation of keys for reference).

LastExceed:
In fact it might actually be the opposite, in StepMania the lines are visible when playing Sudden (mod equivalent to FI) and I personally find them a bit distracting which is why I turned them off. Bar lines are skinnable though so this won't an issue

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
LastExceed: > I would explore options as to whether the buffered container can be slotted in solely for the fade-in mod

Adding the buffered container globally was suggested by smoogipoo, so I just went with that. I have no idea how I would even go about inserting it just for FI and HD (though that might just be my inexperience).

> a named field to which the buffered container is assigned

Good idea, will go with that.

> don't declare single-use constants.

I was reluctant about that at first too, but the same thing was done in `CatchModHidden`

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
LastExceed: Oh I completely forgot about up-scroll players. The ctor param was actually intended for re-use in HD, but I guess it can be used for that as well

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
bdach: >Adding the buffered container globally was suggested by smoogipoo

was not aware; in that case I suppose it can stay.

smoogipoo: I didn't suggest that it should always be here, and agree that it should only be present for FI/HD.

smoogipoo: As for how to do this, maybe you can do something with the `IApplicableToDrawableRuleset` interface (attach it to your mod). Think you'll probably need to wrap the `hitObjectContainer` in another container and expose it from `HitObjectArea` and then do something like...:
```
var hoc = drawbleRuleset.Playfield.HitObjectArea.HitObjectContainer;
hoc.Parent.Remove(hoc);
hoc.Parent.Add(new BufferedContainer { ... Child = hoc });
```

LastExceed: > I didn't suggest that

my bad, misinterpreted then

>you'll probably need to wrap the hitObjectContainer in another container

Doesn't that defeat the purpose of ***not*** wrapping the `hitObjectContainer` globally ?

smoogipoo: `BufferedContainer` has an overhead, which is the suggestion above is trying to avoid. So without HD/FI the hierarchy will be something like:
```
- Container
  - hitObjectContainer
```
But then with FI/HD it'll become:
```
- Container
  - BufferedContainer
    - hitObjectContainer
```
The overhead of having that `Container` wrapping everything is negligible.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+
+                    reversed = value;
+                    Coverage = coverage; //re-apply coverage to update visuals
+                }
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(ManiaRulesetConfigManager configManager)
+            {
+                var scrollDirection = configManager.GetBindable<ManiaScrollingDirection>(ManiaRulesetSetting.ScrollDirection);
```
smoogipoo, line 125: This bindable needs to be stored, otherwise the binding will be lost. Do something like:

```csharp
private readonly IBindable<ManiaScrollingDirection> scrollDirection = new Bindable<ManiaScrollingDirection>();

private void load()
{
    scrollDirection.BindTo(otherBindable);
}
```

I think `otherBindable` in this case should come from `IScrollingInfo`. See how `HitObjectArea` does things for reference.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+                        Color4.White.Opacity(value ? 1f : 0f)
+                    );
+
+                    reversed = value;
+                    Coverage = coverage; //re-apply coverage to update visuals
+                }
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(ManiaRulesetConfigManager configManager)
```
smoogipoo, line 123: This should be placed right below the ctor

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+            {
+                var scrollDirection = configManager.GetBindable<ManiaScrollingDirection>(ManiaRulesetSetting.ScrollDirection);
+
+                if (scrollDirection.Value == ManiaScrollingDirection.Up)
```
smoogipoo, line 127: Bind to the value change event so that you can change the scrolling direction mid-gameplay. Again, `HitObjectArea` is a good reference.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+                {
+                    filled.Anchor = value ? Anchor.BottomLeft : Anchor.TopLeft;
+                    filled.Origin = value ? Anchor.BottomLeft : Anchor.TopLeft;
+                    gradient.Colour = ColourInfo.GradientVertical(
+                        Color4.White.Opacity(value ? 0f : 1f),
+                        Color4.White.Opacity(value ? 1f : 0f)
+                    );
+
+                    reversed = value;
+                    Coverage = coverage; //re-apply coverage to update visuals
```
smoogipoo, line 118: We don't usually do this. Make an `updateCoverage()` method, and implement `Coverage` something like:
```csharp
private float coverage;

public float Coverage
{
    get => coverage;
    set
    {
        if (coverage == value)
            return;

        coverage = value;
        filled.Height = ...
        ...

        updateCoverage()
    }
}
```

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+                {
+                    filled.Height = value;
+                    gradient.Y = reversed ? 1 - value - gradient.Height : value;
+                    coverage = value;
+                }
+            }
+
+            private bool reversed;
+
+            public bool Reversed
```
smoogipoo, line 106: If this is going to be controlled by the scrolling direction coming from settings, it shouldn't be public.

It's ambiguous as to what it means (is the default location at the top or at the bottom?). Instead, I think it would be better to have the position be based on the config value and then `ManiaModFadeIn` would apply `Scale = new Vector2(1, -1)` to invert it (you'd need appropriate anchors origins, `Centre` will work).

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+                    {
+                        RelativeSizeAxes = Axes.Both,
+                        RelativePositionAxes = Axes.Both,
+                        Height = 0.25f
+                    },
+                    filled = new Box
+                    {
+                        RelativeSizeAxes = Axes.Both
+                    }
+                };
```
smoogipoo, line 87:
```suggestion
                };

```

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+            {
+                Blending = new BlendingParameters
+                {
+                    RGBEquation = BlendingEquation.Add,
+                    Source = BlendingType.Zero,
+                    Destination = BlendingType.One,
+                    AlphaEquation = BlendingEquation.Add,
+                    SourceAlpha = BlendingType.Zero,
+                    DestinationAlpha = BlendingType.OneMinusSrcAlpha
+                };
```
smoogipoo, line 74:
```suggestion
                };

```

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -19,5 +30,103 @@ public class ManiaModFadeIn : Mod
...
+                });
+            }
+        }
+
+        private class LaneCover : CompositeDrawable
+        {
+            private readonly Box gradient;
+            private readonly Box filled;
+
+            public LaneCover(float initialCoverage, bool reversed)
```
smoogipoo, line 64: As below, `reversed` should probably not be here. Also it's unnecessary for `initialCoverage` to be here, make the creator assign it instead.

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -2,14 +2,25 @@
...
 using osu.Game.Rulesets.Mania.Objects;
+using osu.Game.Rulesets.Mania.UI;
 using osu.Game.Rulesets.Mods;
+using osu.Game.Rulesets.UI;
+using osuTK.Graphics;
 
 namespace osu.Game.Rulesets.Mania.Mods
 {
-    public class ManiaModFadeIn : Mod
+    public class ManiaModFadeIn : Mod, IApplicableToDrawableRuleset<ManiaHitObject>
```
smoogipoo, line 23: I know this PR is intended to implement this for fadein, but implementing for hidden should be 5 LOC max, and I think it starts with this mod inheriting from `ManiaModHidden`. From there, move all the code below into `ManiaModHidden` and expose a virtual method `CreateLaneCover()` to apply the scale I mentioned.

smoogipoo (changes requested):

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,129 @@
...
+                    },
+                    filled = new Box
+                    {
+                        RelativeSizeAxes = Axes.Both
+                    }
+                };
+            }
+
+            [BackgroundDependencyLoader]
+            private void load(IScrollingInfo configManager)
```
bdach, line 89: looks like a leftover parameter name - should be renamed to `scrollingInfo`

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,129 @@
...
+            {
+                filled.Anchor = reversed ? Anchor.BottomLeft : Anchor.TopLeft;
+                filled.Origin = reversed ? Anchor.BottomLeft : Anchor.TopLeft;
+                filled.Height = coverage;
+
+                gradient.Y = reversed ? 1 - filled.Height - gradient.Height : coverage;
+                gradient.Colour = ColourInfo.GradientVertical(
+                    Color4.White.Opacity(reversed ? 0f : 1f),
+                    Color4.White.Opacity(reversed ? 1f : 0f)
+                );
```
bdach, line 105: thinking out loud here but maybe this could be simplified if instead of doing all this you just rotated the `LaneCover` 180 degrees whenever necessary.

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
LastExceed: oops, leftover indeed.

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
LastExceed: ...i feel stupid now

LastExceed:
not quite ready yet, there's a bug that prevents me from selecting HD (it just switched back to FI)

LastExceed:
It seems the bug is caused by inheriting FI from HD, it first occurred after 
fcda4d9 . I am unable to track down why exactly it happens so I'll need some help here.

[video](https://streamable.com/jwa7tc) (I changed the FI icon to make it more clear)

smoogipoo:
Indeed it was a but with mod selection. I've fixed it in the mentioning PR (and taken some of the changes from this PR for testing purposes).

smoogipoo:
Thanks. I've made a few adjustments to encapsulate the coverage logic, and think this is good to go now.

@peppy please review. Not super happy about exposing the hitobjectarea / hitobjectcontainer, but I don't know of a better way to do this while not creating code paths specific to mods in UI classes. Maybe you have a better idea?

smoogipoo (dismissed):

smoogipoo (dismissed):

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,50 @@
...
+        }
+
+        protected virtual PlayfieldCoveringContainer CreateCover() => new ModHiddenCoveringContainer();
+
+        private class ModHiddenCoveringContainer : PlayfieldCoveringContainer
+        {
+            public ModHiddenCoveringContainer()
+            {
+                // This cover extends outwards from the hit position.
+                Cover.Scale = new Vector2(1, -1);
```
peppy, line 47: not quite clear what this is doing

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,50 @@
...
+            foreach (Column column in maniaPlayfield.Stages.SelectMany(stage => stage.Columns))
+            {
+                HitObjectContainer hoc = column.HitObjectArea.HitObjectContainer;
+                Container hocParent = (Container)hoc.Parent;
+
+                hocParent.Remove(hoc);
+                hocParent.Add(CreateCover().WithChild(hoc).With(c =>
+                {
+                    c.RelativeSizeAxes = Axes.Both;
+                    c.Coverage = 0.5f;
```
peppy, line 35: isn't this going to override any local setting in a custom implementation? or are you expecting custom impl. to set in BDL? feels a bit undefined.

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,50 @@
...
+
+            foreach (Column column in maniaPlayfield.Stages.SelectMany(stage => stage.Columns))
+            {
+                HitObjectContainer hoc = column.HitObjectArea.HitObjectContainer;
+                Container hocParent = (Container)hoc.Parent;
+
+                hocParent.Remove(hoc);
+                hocParent.Add(CreateCover().WithChild(hoc).With(c =>
+                {
+                    c.RelativeSizeAxes = Axes.Both;
```
peppy, line 34: is there a reason for doing this here rather than in the cover class?

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
```diff
@@ -2,15 +2,50 @@
...
+        {
+            ManiaPlayfield maniaPlayfield = (ManiaPlayfield)drawableRuleset.Playfield;
+
+            foreach (Column column in maniaPlayfield.Stages.SelectMany(stage => stage.Columns))
+            {
+                HitObjectContainer hoc = column.HitObjectArea.HitObjectContainer;
+                Container hocParent = (Container)hoc.Parent;
+
+                hocParent.Remove(hoc);
+                hocParent.Add(CreateCover().WithChild(hoc).With(c =>
```
peppy, line 32: rather than setting the HOC as a child, could it be a ctor argument, making `PlayfieldCoveringContainer` a `CompositeDrawable` instead (and maybe renaming to `PlayfieldCoveringWrapper` or something).

osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
```diff
@@ -12,5 +13,7 @@ public class ManiaModFadeIn : ManiaModHidden
         public override string Acronym => "FI";
         public override IconUsage? Icon => OsuIcon.ModHidden;
         public override string Description => @"Keys appear out of nowhere!";
+
+        protected override PlayfieldCoveringContainer CreateCover() => new PlayfieldCoveringContainer();
```
peppy, line 17: maybe deserves a comment about why this uses the default while the less derived class doesn't.

peppy (changes requested):

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
smoogipoo: I generally disagree with putting `RelativeSizeAxes` inside components unless they absolutely need it (i.e. can't work without).

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
smoogipoo: This is a bit temporary since the cover should expand based on the combo.

osu.Game.Rulesets.Mania/Mods/ManiaModHidden.cs
smoogipoo: I hope my enum values resolve this issue for you.

peppy (approved):

LastExceed:
Considering how much activity this PR had, does my contribution still outweigh the time you guys spend reviewing/changing it? Please be cold if necessary

smoogipoo:
I think the initial implementation before I put my hands on it wasn't bad. From there it was mostly stylistic choices so that the component can be reused to implement real lane covers more easily in the future + a test to observe how easy it is to use and to make sure it behaves correctly with scrolling direction changes.

You'd have gotten it in the end, but these stylistic decisions are sometimes a subjective opinion that even I had to revise. tl;dr yes your contribution was worthwhile and overweighed the time spend reviewing/changing it.

Issue #7301, opened 2019-12-20
Background sometimes flashes when entering SongSelect


smoogipoo:
See: https://streamable.com/7l3zx (may help if you watch in slow motion)

There's a frame where it's faded to alpha = 0 and then faded back to the intended alpha at the very start of entering song select.

Joehuu:
Not sure if related, but spamming random after entering song select doesn't blur the background.
![UJUC8j2jq1](https://user-images.githubusercontent.com/35318437/76818014-b7e7eb00-67c1-11ea-859d-4698152dcfc0.gif)

peppy:
It's weird that on the same frame in your video, the stats on the left appear. Almost feels like a `BufferedContainer` issue (that we may have since solved?)

smoogipoo:
This has not been resolved.

PR #9539, opened 2020-07-12
Fix correct filter criteria not being applied to beatmap carousel if beatmaps take too long to load

### LINKED ISSUE #9408: First carousel load doesn't sort beatmaps with the user's choice
**Describe the bug:** In my case, the carousel should sort by `Title`, but it seems to sort by `Artist`.
Also, looks like this issue happens when entering Solo right after starting the game (but I can't be precise on this).

**Screenshots or videos showing encountered issue:** [Video](https://drive.google.com/file/d/1mQnQZP-Ft6i-PtW7mXfs9AbHIbVt4w0P/view?usp=sharing)

**osu!lazer version:** 2020.629.0

peppy:
- Closes https://github.com/ppy/osu/issues/9408.

The case here was `FilterControl` loading before `loadBeatmapSets` completes. `SongSelect` would do a non-debounced filter application

https://github.com/ppy/osu/blob/57cfdb82ce47965e1f3654c8460e4be296a0715c/osu.Game/Screens/Select/SongSelect.cs#L307-L310

but this is cancelled at 

https://github.com/ppy/osu/blob/57cfdb82ce47965e1f3654c8460e4be296a0715c/osu.Game/Screens/Select/BeatmapCarousel.cs#L430

That was the only filter application pathway.

This change ensure the filter is always applied at the end of the load process, and only applied once (via checking the `BeatmapSetsLoaded` flag). You may jump to the conclusion that this isn't thread safe (what happens if the `FilterControl` applies the filter *after* the added Flush call?), but it will actually work correctly due to the filter *always* being scheduled until `BeatmapSetsLoaded` is true.

I make this change reluctantly, but I have tested quite thoroughly. Not too easy to write test coverage (open to suggestion, maybe converting the async part to a `protected virtual` and add a semaphore?), but to test locally the reported bug, add a `Thread.Sleep(10000)` in `loadBeatmapSets` before the `root = newRoot` line.

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -114,7 +114,7 @@ namespace osu.Game.Tests.Visual.SongSelect
         {
             loadBeatmaps();
 
-            AddStep("select last", () => carousel.SelectBeatmap(carousel.BeatmapSets.First().Beatmaps.First()));
+            AddStep("select first", () => carousel.SelectBeatmap(carousel.BeatmapSets.First().Beatmaps.First()));
             waitForSelection(1, 1);
 
             advanceSelection(direction: 1, diff: true);
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -80,9 +80,9 @@ namespace osu.Game.Tests.Visual.SongSelect
         [Test]
         public void TestRecommendedSelection()
         {
-            loadBeatmaps();
+            loadBeatmaps(carouselAdjust: carousel => carousel.GetRecommendedBeatmap = beatmaps => beatmaps.LastOrDefault());
 
-            AddStep("set recommendation function", () => carousel.GetRecommendedBeatmap = beatmaps => beatmaps.LastOrDefault());
+            AddStep("select last", () => carousel.SelectBeatmap(carousel.BeatmapSets.Last().Beatmaps.Last()));
 
             // check recommended was selected
             advanceSelection(direction: 1, diff: false);
@@ -114,7 +114,7 @@ namespace osu.Game.Tests.Visual.SongSelect
         {
             loadBeatmaps();
 
-            advanceSelection(direction: 1, diff: false);
+            AddStep("select last", () => carousel.SelectBeatmap(carousel.BeatmapSets.First().Beatmaps.First()));
             waitForSelection(1, 1);
 
             advanceSelection(direction: 1, diff: true);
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -108,6 +107,9 @@ private void loadBeatmapSets(IEnumerable<BeatmapSetInfo> beatmapSets)
             itemsCache.Invalidate();
             scrollPositionCache.Invalidate();
 
+            // the filter criteria may have changed since the above filter operation.
```
bdach, line 110: I don't see a filter operation anywhere above, in fact, one seems to have been removed...?

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -321,6 +323,9 @@ private void selectNextDifficulty(int direction)
         /// <returns>True if a selection could be made, else False.</returns>
         public bool SelectNextRandom()
         {
+            if (!AllowSelection)
+                return false;
```
bdach, line 327: what's the reasoning for this one? removing it seems to have changed nothing in my quick test.

bdach (commented):
looked at this although as usual I'm reluctant to look at this code as my brain is too tiny for it. looks to be behaving better in the scenario provided, although sometimes I seem to be getting a big screen-wide one-frame flicker once the load completes ([see video](https://streamable.com/0gbz2c)). not sure if related but haven't seen anything like it before as far as I remember.

osu.Game/Screens/Select/BeatmapCarousel.cs
peppy: I doubt it will change anything, just feels more correct.

osu.Game/Screens/Select/BeatmapCarousel.cs
peppy: Ah yep, this comment is no longer valid. The above filter was incorrect in most every case so I removed it.

smoogipoo:
The flicker is probably the same issue as https://github.com/ppy/osu/issues/7301

smoogipoo (dismissed):
Seems pretty sane

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -114,7 +114,7 @@ public void TestTraversal()
         {
             loadBeatmaps();
 
-            advanceSelection(direction: 1, diff: false);
+            AddStep("select last", () => carousel.SelectBeatmap(carousel.BeatmapSets.First().Beatmaps.First()));
```
smoogipoo, line 117:
```suggestion
            AddStep("select first", () => carousel.SelectBeatmap(carousel.BeatmapSets.First().Beatmaps.First()));
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #9540, opened 2020-07-12
Allow pausing gameplay via middle mouse button

### LINKED ISSUE #9531: Add back ability to pause gameplay on middle mouse click
**Describe the new feature:** 
clicking the middle mouse button should instantly pause the current gameplay meaning, the button should react exactly like it did in stable.

peppy:
- Closes https://github.com/ppy/osu/issues/9531

Note that for the moment this just triggers standard "quit" behaviour (for cases like replays). This can change in the future once we have the ability to pause replays as well, but having the binding in place earlier is favourable IMO.

### DIFF
--- a/osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
+++ b/osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
@@ -265,6 +265,7 @@ namespace osu.Game.Screens.Play.HUD
                 switch (action)
                 {
                     case GlobalAction.Back:
+                    case GlobalAction.PauseGameplay:
                         AbortConfirm();
                         break;
                 }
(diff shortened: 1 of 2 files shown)


bdach (dismissed):
seems straightforward enough

Joehuu:
In stable, gameplay mouse buttons disables this. I think lazer should do the same.

Also the hold to confirm is not working right. Pressing the middle button and releasing it will not cancel it.

osu.Game/Screens/Play/HUD/HoldForMenuButton.cs
```diff
@@ -251,6 +251,7 @@ public bool OnPressed(GlobalAction action)
                 switch (action)
                 {
                     case GlobalAction.Back:
+                    case GlobalAction.PauseGameplay: // in the future this behaviour will differ for replays etc.
```
bdach, line 254: as correctly pointed out by @Joehuu, this same case label should be also added in `OnReleased(GlobalAction)`.

bdach (dismissed):

smoogipoo (dismissed):
TIL this is a thing

smoogipoo (approved):

PR #9543, opened 2020-07-13
Only allow link clicking and tooltips of multi room drawables when selected

### LINKED ISSUE #4395: Multi room overlays shouldn't have any linked text
Not good at design or anything, but having linked text on a background that is clickable is going to cause problems (i.e. misclicks).

Anyway, they are also linked on the right side of the screen, so redundant?

Joehuu:
Fixes https://github.com/ppy/osu/issues/4395.


smoogipoo (approved):
lgtm

peppy:
We really need to fix that font colour too.

PR #9545, opened 2020-07-13
Fix judgement animations not resetting on re-use

### LINKED ISSUE #9507: Animated judgements can start playing from wrong frame
**Describe the bug:** You can see that the accuracy on the HUD drops, but visually, everything looks like it's being hit perfectly.
Important note, not every single judgement went away, in the start of the play there's a couple of 100-s and 50-s, and later during the map they became invisible. Also, it seems that misses are fine throughout the map

Here's the skin: http://bit.ly/35cLsDR

**Screenshots or videos showing encountered issue:** https://youtu.be/n7VSg7Quz5g

**osu!lazer version:** 2020.710.0

**Logs:**
[performance.log](https://github.com/ppy/osu/files/4904044/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4904045/runtime.log)
[database.log](https://github.com/ppy/osu/files/4904046/database.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9507

Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3715
- [x] https://github.com/ppy/osu-framework/pull/3716
- [x] https://github.com/ppy/osu-framework/pull/3717

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -52,6 +52,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.622.1" />
-    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.710.0" />
+    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.714.0" />
   </ItemGroup>
 </Project>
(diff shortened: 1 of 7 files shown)


smoogipoo:
It's kinda hard to test this one unfortunately. But I've constructed the test such that if you press the buttons at the right moment, you'll see the second frame being displayed first at slow testbrowser playback speed.

bdach:
will hold off with my changes fixing toggling lighting in the middle of gameplay until this is merged as they conflict pretty heftily with this at the test scene.

peppy (approved):

PR #9549, opened 2020-07-13
Rework DatabasedKeyBinding to use a string instead of an integer for action storage

vntxx:
_Reopen of #9533_

Also this might break everything and is not quite ready yet as the migration is missing, thus I am opening this PR as a draft first.
If this concept is completely out of line feel free to close this completely.
At least I really had fun getting into the osu code and exploring various concepts of c# :)

## Additional Informations
I may also need to say sorry first. I should had taken a look into the [Contribution Guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md) for this repository prior to doing anything (as one normally would). In that regard I hereby deeply apologize.  

Furthermore the migrations are currently in an incomplete state (I think) and I would like (if possible) some help with them.
If so, please have a look [here](https://github.com/ppy/osu/pull/9533#issuecomment-657149405) first to better understand what my intention was.


osu.Game/Migrations/20200712012841_KeyBindingActionValueChange.cs
```diff
@@ -0,0 +1,125 @@
...
+using Microsoft.EntityFrameworkCore.Migrations;
+using osu.Framework.Platform;
+using osu.Game.Database;
+using osu.Game.Input.Bindings;
+using osu.Game.Rulesets;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+namespace osu.Game.Migrations
+{
+    public partial class KeyBindingActionValueChange : Migration
```
bdach, line 15: this migration can't even *begin* to run since it's missing a corresponding `20200712012841_KeyBindingActionValueChange.Designer.cs` file with a model snapshot. was this tested in any capacity?

osu.Game/Migrations/20200712012841_KeyBindingActionValueChange.cs
```diff
@@ -0,0 +1,125 @@
...
+                // Probably is missing mania dual stage migrations
+            }
+        }
+    }
+
+    internal class KeybindActionLookup
+    {
+        private readonly IDictionary<int?, Type> lookup = new Dictionary<int?, Type>
+        {
+            { null!, typeof(GlobalAction) },
```
bdach, line 67: this is not valid. null keys are not permitted in dictionaries, and the `!` has no meaning as we don't have nullable reference types enabled. this crashes at run-time.

osu.Game/Migrations/20200712012841_KeyBindingActionValueChange.cs
```diff
@@ -0,0 +1,125 @@
...
+            // OR
+
+            foreach (var actionEnum in actionLookup.Lookup())
+            {
+                foreach (var enumValue in actionEnum.Value.GetEnumValues())
+                {
+                    migrationBuilder.Sql($"UPDATE KeyBinding SET Action={enumValue} WHERE Action={(int)enumValue} AND RulesetID={actionEnum.Key}");
+                }
+
+                // Probably is missing mania dual stage migrations
```
bdach, line 58: what does "probably" mean?

osu.Game/Migrations/20200712012841_KeyBindingActionValueChange.cs
```diff
@@ -0,0 +1,125 @@
...
+{
+    public partial class KeyBindingActionValueChange : Migration
+    {
+        private readonly RulesetStore rulesetStore;
+
+        // OR
+
+        private readonly KeybindActionLookup actionLookup;
+
+        public KeyBindingActionValueChange(DatabaseContextFactory databaseContextFactory, Storage storage)
```
bdach, line 23: as you stated, migrations expect a parameterless ctor, so option 1 is not even an option unless the dependencies are manually created.

bdach (changes requested):

bdach:
leaving the above comments aside: as you probably have already noticed the major problem with this migration is the fact that the main game assembly is decoupled from the ruleset assemblies, so you might not necessarily even be *aware* of what keybindings there are to migrate. reflection also does not help here as unfortunately rulesets are loaded well after EF migrations are ran.

to this end I would propose, instead of entirely nuking the old `int` column and substituting it with a `string` one, having the two co-exist alongside one another for a while and then deprecating the `int` one in a future release. as to where the migration should be done, I wouldn't necessarily try to jam it into what EF provides, but maybe look into running it during initialisation of `KeyBindingStore`. however I would like opinions on this direction first (@peppy, @smoogipoo).

vntxx:
Thank you for the review @bdach! I will hold off for now as it doesn't seem to make any sense to work on this until your proposal got reviewed.

peppy:
If we are going to do this while still on EF and without breaking rulesets (too much backwards compatibility for something too new? i dunno, whatever floats boats) I think the simplest way is to just make a new table, migrate across from old as things update.

Not sure if wasted effort

osu.Game/Input/Bindings/DatabasedKeyBinding.cs
```diff
@@ -29,5 +29,8 @@ public int IntAction
             get => (int)Action;
             set => Action = value;
         }
+
+        [Column("SAction")]
```
bdach, line 33: that's not a great column name; `ActionName` maybe better? (goes for the field too I suppose).

osu.Game/Database/OsuDbContext.cs
```diff
@@ -134,7 +134,7 @@ protected override void OnModelCreating(ModelBuilder modelBuilder)
             modelBuilder.Entity<SkinInfo>().HasIndex(b => b.DeletePending);
 
             modelBuilder.Entity<DatabasedKeyBinding>().HasIndex(b => new { b.RulesetID, b.Variant });
-            modelBuilder.Entity<DatabasedKeyBinding>().HasIndex(b => b.IntAction);
```
bdach, line 137: I'd say both indices need to co-exist for now.

vntxx:
As the new title states: this pr currently just prepares and migrates old values into a StringAction placeholder. The core functionality is already implemented in another branch/commit. I will probably open a seperate pr as it should not be merged into the same release.

osu.Game/Input/Bindings/DatabasedKeyBinding.cs
vntxx: Sounds good. Will be changed.

osu.Game/Database/OsuDbContext.cs
vntxx: Will need to add the indexing of StringAction into the migration too

osu.Game/Input/KeyBindingStore.cs
```diff
@@ -60,6 +63,18 @@ private void insertDefaults(IEnumerable<KeyBinding> defaults, int? rulesetId = n
                         usage.Context.SaveChanges();
                     }
                 }
+
+                // for now this should be fine here
```
peppy, line 67: what should be fine where?

comments should explain what code is doing, not make statements about something you're thinking at the time. an explanation of why this is here of what it's doing should precede your "should be fine" statement, at very least.

osu.Game/Input/KeyBindingStore.cs
```diff
@@ -60,6 +63,18 @@ private void insertDefaults(IEnumerable<KeyBinding> defaults, int? rulesetId = n
...
+                foreach (var group in defaults.GroupBy(k => k.Action))
+                {
+                    foreach (var databasedKeyBinding in usage.Context.DatabasedKeyBinding.Where(b => b.RulesetID == rulesetId && b.Variant == variant && b.IntAction == (int)group.Key))
+                    {
+                        if (databasedKeyBinding.ActionName != null) continue;
+
+                        databasedKeyBinding.ActionName = group.Key.ToString();
```
peppy, line 74: this is writing every time, regardless of current state. if this is supposed to be migrating, it should only target the required rows, in the `where` statement.

osu.Game/Overlays/KeyBinding/KeyBindingsSubsection.cs
```diff
@@ -41,7 +41,7 @@ private void load(KeyBindingStore store)
                 int intKey = (int)defaultGroup.Key;
 
                 // one row per valid action.
-                Add(new KeyBindingRow(defaultGroup.Key, bindings.Where(b => ((int)b.Action).Equals(intKey)))
+                Add(new KeyBindingRow(defaultGroup.Key, bindings.Where(b => b.IntAction == intKey))
```
peppy, line 44: I'm not sure I follow. You've added a new column and prepared migrations, but are still using the old column in all usages?

Also why store both? `KeyBinding`'s `Action` is already an object, so you could just store either an int or string there and match accordingly?

peppy (changes requested):

osu.Game/Input/KeyBindingStore.cs
vntxx: Will expand this, sorry. I think that code doesn't belong in this method at all, but for the time being it fits there just fine. It would get removed anyways in a future release.

osu.Game/Input/KeyBindingStore.cs
vntxx: Did not think about that. Will fix, thanks!

osu.Game/Overlays/KeyBinding/KeyBindingsSubsection.cs
vntxx: I tried nearly everything in that regard and I couldn't get it to work with populating the Action either by an integer value or string dynamically. This is partially because I can't really control in what order EF sets initial values. If you have an Idea, please outline it to me.

vntxx: Also the implementation and usage of the migrated value will follow in a seperate commit but I might try another approach again.

vntxx: Unless

vntxx:
I think now it is in a viable state. Me changing my mind about things exploded this pr. I'd consider squashing everything into one or two logical commits (would force push for that). And while I'm at it maybe rebase onto master to see if suff breaks but that might not be necessary. I do still have the follow-up commit in place but it misses the old column removal as the SQLite dialect doesn't support this in EF.

bdach:
Opposed to squashing & rebasing. Let's just keep it simple.

vntxx:
If you like to keep all the unnecessary commits, fine by me.

bdach:
If that were really to be an issue, a new PR would be the better way to go. Rewriting history of pushed commits is a risky business and should be kept to a minimum.

peppy:
Is this in a state ready for review?

vntxx:
I guess. But I don't know why the build is failing though.

peppy:
It's not failing, I re-ran it.

vntxx:
Huh ok. Then sure.

peppy:
Apologies for never getting this merged. Going to close it for now since the `KeyBinding` storage is our target for testing / migrating to realm (#11461), and there's no chance that these changes will get merged until that happens.

PR #9553, opened 2020-07-13
Change difficulty adjust mod to match stable range of 0-10

### LINKED ISSUE #9438: allow setting all difficulty customization mod values to 0
the lowest setting value allowed is 1 

you can use 0 in the value setting in the editor so you should be able to set the setting in the mod to 0 for consistency

peppy:
- Closes https://github.com/ppy/osu/issues/9438


Joehuu:
Circle size in stable's editor is 2-7.

peppy:
on a quick test, I think we can probably let CS remain 0-10 for now.

smoogipoo (approved):

PR #9556, opened 2020-07-14
Disable osu!catch hit "explosion" when hit lighting is turned off

### LINKED ISSUE #9515: combo-lightning in osu!catch is blinding
**Describe the bug:** Note: hit lightning is disabled
The lightning you get when catching a fruit is very blinding. I don't know what this is exactly called, since it isn't hit lightning, but rather an effect that appears for every fruit you catch.

This sort of 'combo'-lightning is also present in stable except it is way less blinding. 
It also is a lot bigger in stable, (as in the lightning effect goes as far as about the middle of the screen instead of just being a small 'ball' above the catcher)

The way to fix this would be to make it similar to the 'combo'-lightning in stable.
alternatively a way to turn this effect off could be added.

https://youtu.be/JYBZ-zW6iv0 (lazer catch with skin)
https://youtu.be/INGFHIrtC1U (lazer catch with default skin)
It doesn't look as bad on video, but when playing these light are very blinding

Compared to the 'combo'-lightning in stable
https://www.youtube.com/watch?v=BkE6h8RSJio (stable osu!catch, you can see it here or in any other osu!catch video)

**osu!lazer version:** 
2020.710.0

gagahpangeran:
Closed #9515


peppy:
This will not work correctly if the setting is changed during gameplay. Please look into how other usages use config / bindable flow and copy by example.

peppy (changes requested):
.

gagahpangeran:
Oops sorry my bad.
I only see the example from this
https://github.com/ppy/osu/blob/e0d9e70404fd5beabf944d886cfe7544d5aad907/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuJudgement.cs#L34-L43

I'll change it.

peppy:
That is a special case because that load function would get called every time the judgement is displayed, but in your case you're only checking once ever (when the catcher is loaded).

bdach:
the code looks OK to me but I would like to see a test for this. it should be relatively painless to add.

gagahpangeran:
Which test file can I use as an example?

smoogipoo:
You can use mania's `TestSceneDrawableJudgement` as a starting point. You'll then want to add another step that toggles it, something like:
```csharp
[BackgroundDependencyLoader]
private void load(OsuConfigManager configManager)
{
    AddToggleStep("toggle hit lighting", enabled => configManager.Set(OsuSetting.HitLighting, enabled));
}
```

gagahpangeran:
I'm not sure this is the right test implementation.

I modify from https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs and the code a bit messy.

bdach:
I don't think having a separate test scene for hit lighting was the best idea, especially seeing that it copies most of `TestSceneCatcherArea`.

I'd probably rather use that test scene and add a

```csharp
[Test]
public void TestHitLighting()
{
    // actual testing code here
}
```

(it should work fine alongside the ctor test steps). You could refactor the parts you need from that test scene out of the constructor into separate methods you can call.

gagahpangeran:
What if I add `AddToggleStep` of hit lighting before all of the step in `TestSceneCatcherArea`?

peppy:
Something like

```csharp
AddAssert("hit lighting was not displayed", () => !this.ChildrenOfType<HitLightingDrawable>().Any());
```

bdach:
I've applied some cosmetic changes here; otherwise looks good to go to me.

PR #9559, opened 2020-07-14
Fix certain overlays and screens blocking home and alt-f4

### LINKED ISSUE #3812: ModSelectOverlay causes the Home button to work incorrectly

### LINKED ISSUE #9430: Using quick-retry hotkey in results screen when stats are expanded results in black screen

Joehuu:
- Resolves https://github.com/ppy/osu/issues/3812
- Fixes https://github.com/ppy/osu/issues/9430, supersedes and closes https://github.com/ppy/osu/pull/9436

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -573,7 +573,9 @@ namespace osu.Game
                             Origin = Anchor.BottomLeft,
                             Action = () =>
                             {
-                                if ((ScreenStack.CurrentScreen as IOsuScreen)?.AllowBackButton == true)
+                                var currentScreen = ScreenStack.CurrentScreen as IOsuScreen;
+
+                                if (currentScreen?.AllowBackButton == true && !currentScreen.OnBackButton())
                                     ScreenStack.Exit();
                             }
                         },
--- a/osu.Game/Screens/IOsuScreen.cs
+++ b/osu.Game/Screens/IOsuScreen.cs
@@ -57,6 +57,9 @@ namespace osu.Game.Screens
         /// </summary>
         bool AllowRateAdjustments { get; }
 
+        /// <summary>
+        /// Whether there are sub overlays/screens that need closing with the back button before this <see cref="IOsuScreen"/> can be exited.
+        /// </summary>
         bool OnBackButton();
     }
 }
--- a/osu.Game/Screens/IOsuScreen.cs
+++ b/osu.Game/Screens/IOsuScreen.cs
@@ -56,5 +56,7 @@ namespace osu.Game.Screens
         /// Whether mod rate adjustments are allowed to be applied.
         /// </summary>
         bool AllowRateAdjustments { get; }
+
+        bool OnBackButton();
     }
 }
--- a/osu.Game/Screens/IOsuScreen.cs
+++ b/osu.Game/Screens/IOsuScreen.cs
@@ -58,8 +58,12 @@ namespace osu.Game.Screens
         bool AllowRateAdjustments { get; }
 
         /// <summary>
-        /// Whether there are sub overlays/screens that need closing with the back button before this <see cref="IOsuScreen"/> can be exited.
+        /// Invoked when the back button has been pressed to close any overlays before exiting this <see cref="IOsuScreen"/>.
         /// </summary>
+        /// <remarks>
+        /// Return <c>true</c> to block this <see cref="IOsuScreen"/> from being exited after closing an overlay.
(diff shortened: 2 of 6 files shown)


osu.Game/Screens/IOsuScreen.cs
```diff
@@ -56,5 +56,7 @@ public interface IOsuScreen : IScreen
         /// Whether mod rate adjustments are allowed to be applied.
         /// </summary>
         bool AllowRateAdjustments { get; }
+
+        bool OnBackButton();
```
bdach, line 60: needs xmldoc

osu.Game/OsuGame.cs
```diff
@@ -573,7 +573,9 @@ protected override void LoadComplete()
                             Origin = Anchor.BottomLeft,
                             Action = () =>
                             {
-                                if ((ScreenStack.CurrentScreen as IOsuScreen)?.AllowBackButton == true)
+                                var currentScreen = ScreenStack.CurrentScreen as IOsuScreen;
+
+                                if (currentScreen?.AllowBackButton == true && !currentScreen.OnBackButton())
```
bdach, line 578: I'm not sure I'm following why both `AllowBackButton` and `OnBackButton` need to exist. `OnBackButton` seems like the more general mechanism and can probably supersede all usages of `AllowBackButton`, unless I'm missing something.

osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -269,6 +263,25 @@ public override bool OnExiting(IScreen next)
...
 
+        public override bool OnBackButton()
+        {
+            var songSelect = screenStack.CurrentScreen as MatchSongSelect;
+
+            if (songSelect?.ModSelect.State.Value == Visibility.Visible)
+            {
+                songSelect.ModSelect.Hide();
+                return true;
+            }
```
bdach, line 274: not sure `Multiplayer` should be prodding its sub-screen states directly like this. maybe something an `OnBackButton()` in `IMultiplayerSubScreen` would be better.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -86,7 +86,7 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
         [Resolved]
         private BeatmapManager beatmaps { get; set; }
 
-        protected ModSelectOverlay ModSelect { get; private set; }
+        public ModSelectOverlay ModSelect { get; private set; }
```
bdach, line 89: ideally this shouldn't be done.

bdach (dismissed):

osu.Game/OsuGame.cs
Joehuu: `AllowBackButton` is there to hide the back button and disable its action on certain screens (like player doesn't need a back button, and has local exit logic).

`OnBackButton` is there to override the action, and only logical to implement if `AllowBackButton` is true.

osu.Game/Screens/Multi/Multiplayer.cs
Joehuu: I think it needs to be done here, as sub screens can't control the back button.

```
if ((screenStack.CurrentScreen as IMultiplayerSubScreen).OnBackButton())
	return true;
```

I thought of this, seems more better?

bdach: yeah, that's what I mean; the check itself being there I don't have much of an issue with, but accessing the subscreens' controls in here (some of which need to be exposed for this to work) doesn't seem too great.

smoogipoo:
I reworded the xmldoc a bit and am fine with the implementation. Only thing I'm a little iffy about is the `OnBackButton()` naming but I can convince myself it makes sense.

Thoughts, @peppy (or merge)

smoogipoo (approved):

peppy:
The naming matches other similar `OsuScreen` methods so I think it's fine.

peppy (approved):

PR #9563, opened 2020-07-14
 Add filtering by friends to spotlights in RankingsOverlay

EVAST9919:
Screenshot_1


bdach:
I know this is what the web design is, but it looks off to me (in both places). Unless there are plans that aren't revealed for this yet I don't really understand what all that spacing is for.

EVAST9919:
Yeah. Would be better to just move it to the right of the selected spotlight info, in case there are no plans regarding this part yet

smoogipoo:
I would do that - make it match the performance page. It looks pretty bad as is.

EVAST9919:
Reworked

smoogipoo (approved):
I think this is fine for now.

PR #9566, opened 2020-07-15
Add ability to seek replays/auto with arrow keys

### LINKED ISSUE #9479: Seek with arrow keys during replay/auto
**Describe the new feature:** It would be nice if you could move in a replay with arrow keys for an amount of time, maybe 3/5 seconds (could be chosen by the user in the settings maybe).

Joehuu:
Fixes https://github.com/ppy/osu/issues/9479. Was too hidden before, so reintroducing the ability.


peppy:
Would be great to have a test for this (I think this is important because there's a high chance of it regressing if we add other elements which handle the arrow keys in the future).

Should be simple to add to `TestSceneAutoplay`.

peppy:
Your tests were relying on unsafe conditions (until step on something which may eventually become true). Changed to be based on time and not use until.

peppy (approved):

PR #9570, opened 2020-07-15
Fix catch hiding entire streams instead of contained fruits

### LINKED ISSUE #8354: osu!catch hidden is not same as osu!stable
**Describe the bug:** 
for some reason it seems that fruit/note become hidden instantly when first fruit/note of same colorcombo(i dont know mapping term) or when it sliders it turn all to hidden

https://youtu.be/ydNr6IdCKtU issue is really viable at 0:25 in osu!stable you can see stack fruit while osu!lazer is already completely gone

**osu!lazer version:** 
2020.317.0.0

**Logs:**
[logs.zip](https://github.com/ppy/osu/files/4352688/logs.zip)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/8354


peppy (approved):
Seems logical.

Issue #7155, opened 2019-12-11
Improve incompatible mods logic


peppy:
- There should be a central method for checking and removing incompatible mods.
- A mod should be able to list itself as incompatible but still be selectable. Consider a case where you have `abstract ModTimeAdjust` and want to make it incompatible with all other `ModTimeAdjust` inheritors. Or maybe even `IApplicableToX`.
- Look into whether we can list incompatible pairs, rather than requiring it to be specified in two different places (prone to error).

huoyaoyuan:
I thought about this several days ago.
```csharp
abstract class Mod
{
    abstract bool IsCompatibleWith(Mod other);
}

abstract class ModTimeAdjust
{
    override bool IsCompatibleWith(Mod other) => !(other is ModTimeAdjust);
}
```

MightyHelper:
Maybe create "incompatibility groups"?
They would consist of an array of mods that are incompatible between each other, in other words, **only one** of the group can be active at **any point in time**.
Haven't read the code, just an idea I came up with after reading the issue.

Flutterish:
In general mods are definitely incompatible if they modify the same values plus some other cases. You could somehow use reflections to mark certain fields / properties as mod modifiable, then mods that modify the same values are incompatible.
```cs
// some file
[Moddable( nameof( SpeedModifier ) )]
private double SpeedModifier = 1;

// mod file
[Modifies( /*nameof( MyOtherFile ),*/ nameof( MyOtherFile.SpeedModifier ) )]
private double SpeedModifier => 2;
```
Do you think this idea is viable?

bdach:
That just sounds like inviting unnecessary complexity to me.

smoogipoo:
What about for cases that mods aren't inter-compatible between themselves without touching an external property?

I think as mentioned in the OP we just need somewhere to define incompatibility pairs - probably a method from `Ruleset`.

Flutterish:
Then you need an additional method which explicity states that, like the current one or huoyaoyuan's does.
As of bdach's comment, I think this is pretty similar to `[Cached]` and `[Resolved]` in the sense that it solves a dependency in the background.

smoogipoo:
There isn't a direct mod-to-property relationship. Mods are extensible to the point they can do literally anything anywhere from the beatmap to the draw hierarchy _without_ setting flags in random places. An automated system just does not have enough knowledge to determine incompatibilities with such a system - it requires developer knowledge.

Flutterish:
[...]

PR #9571, opened 2020-07-15
Fix mod selection not working with base types

smoogipoo:
Fixes the issue mentioned in https://github.com/ppy/osu/pull/9538

I ported back the mod inheritance from the PR above. This is a case where we're trying to select `HD`, but since `FI : HD` and `FI` is the first visible button, `FI` is re-selected. I'd say it makes sense to do exact matching for mod selection, and keep `InstanceOfType` for deselection only.

### DIFF
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModFadeIn.cs
@@ -1,23 +1,16 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Graphics;
-using osu.Game.Rulesets.Mania.Objects;
-using osu.Game.Rulesets.Mods;
 
 namespace osu.Game.Rulesets.Mania.Mods
 {
-    public class ManiaModFadeIn : Mod
+    public class ManiaModFadeIn : ManiaModHidden
     {
         public override string Name => "Fade In";
         public override string Acronym => "FI";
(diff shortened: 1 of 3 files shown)


bdach:
> I'd say it makes sense to do exact matching for mod selection, and keep `InstanceOfType` for deselection only.

Not sure I'd agree with that notion; I'd say it needs to either always respect type hierarchy or always match by outermost type. That said I see that changing to that model utterly breaks mod incompatibility logic right now, but I would argue that's a shortcoming of the incompatibility rules' design in general and should be properly addressed with #7155, so not sure if this is something we want to deal with right now.

smoogipoo:
The way I see it is deselection rules are user defined - a mod can be incompatible with a specific mod or with an entire subset of mods without ambiguity, whereas selection can't be defined in the same way and introduces ambiguity when you request selection on a subset.

tl;dr the only way selection can work is by exact matching. As for deselection, is there a reason why you'd limit it to exact matching? There are times when it may be useful - if a ruleset made a mod incompatible with `ModRateAdjust` and listed `DT,HD`, would you expect the ruleset to the have to explicitly add more exclusions if we added another game-wide `ModRateAdjust` mod?

peppy:
I think this is fine for now, until a case comes up where we need otherwise (I can't think of one).

bdach:
The reason I'd want to see deselection on concrete types too is mostly consistency/principle of least astonishment. I do realise that inheritance is often leveraged to avoid having to specify subtypes in incompatibility rules, but again, I'd mostly say it's a shortcoming of how these rules are implemented right now.

Having one central place that declares them, perhaps in a more expressive (and less repetitive) manner than done currently, could deal with that, but I'd have to take one long hard look at what mods we have right now to say for sure.

peppy:
The whole mod manager class needs to be split out as it is managing drawable and non-drawable aspects, so we can definitely revisit and better define the API at that point.

PR #9572, opened 2020-07-15
Fix osu!catch dropping fruit appearing above the plate instead of behind

### LINKED ISSUE #7938: Falling fruits shouldn't be on a higher level than the catcher
https://streamable.com/n5bqb

**osu!lazer version:** 
2020.221.0

peppy:
- Closes https://github.com/ppy/osu/issues/7938

I was considering changing `ExplodingFruitTarget` to also user proxying, but turns out that's not what we want due to it no tracking the catcher position.

### DIFF
--- a/osu.Game.Rulesets.Catch/UI/Catcher.cs
+++ b/osu.Game.Rulesets.Catch/UI/Catcher.cs
@@ -46,6 +46,12 @@ namespace osu.Game.Rulesets.Catch.UI
 
         public Container ExplodingFruitTarget;
 
+        public Container<DrawableHitObject> CaughtFruitContainer { get; } = new Container<DrawableHitObject>
+        {
+            Anchor = Anchor.TopCentre,
+            Origin = Anchor.BottomCentre,
+        };
+
         [NotNull]
         private readonly Container trailsTarget;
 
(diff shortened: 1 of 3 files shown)


bdach:
Visually checked this, seems better.

Worth adding xmldoc to the new public members or no? I know that they are public mostly out of necessity but still. I could be convinced either way to be fair.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -46,6 +46,12 @@ public class Catcher : SkinReloadableDrawable, IKeyBindingHandler<CatchAction>
 
         public Container ExplodingFruitTarget;
 
+        public Container<DrawableHitObject> CaughtFruitContainer { get; } = new Container<DrawableHitObject>
```
smoogipoo, line 49: I'd have preferred these to be exposed via a `public Drawable CreateProxiedContent()` method instead, rather than exposing containers directly.

smoogipoo (changes requested):

osu.Game.Rulesets.Catch/UI/Catcher.cs
peppy: in catcherarea too?

smoogipoo: That one is probably okay since it's not a container.

peppy: so `CatcherArea.Catcher.CreateProxiedContent()`?

smoogipoo: Yeah, that looks fine.

smoogipoo (approved):

PR #9577, opened 2020-07-16
Add an exponent to the accuracy portion of score

smoogipoo:
Awards higher accuracy values exponentially more score. This is the same curve as ScoreV2.

### DIFF
--- a/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
@@ -18,6 +18,7 @@ namespace osu.Game.Rulesets.Scoring
     {
         private const double base_portion = 0.3;
         private const double combo_portion = 0.7;
+        private const double accuracy_exponent = 10;
         private const double max_score = 1000000;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/Scoring/ScoreProcessor.cs
```diff
@@ -18,6 +18,7 @@ public class ScoreProcessor : JudgementProcessor
     {
         private const double base_portion = 0.3;
```
peppy, line 19: think we should rename this to `accuracy_portion`? not sure if it works better than "base"..

peppy (dismissed):

osu.Game/Rulesets/Scoring/ScoreProcessor.cs
smoogipoo: I was going to do that in the upcoming mania PR but can do it here if you'd rather that.

peppy: Either or.

smoogipoo:
Let's not do this for now because it feels bad, and change mania instead.

PR #9579, opened 2020-07-16
Fix storyboard video playback when not starting at beginning of beatmap

### LINKED ISSUE #9561: Videos starting at the middle of the beatmap don't load
Some beatmaps with video that starts at middle part of the beatmap doesn't load at all.

peppy:
- Closes https://github.com/ppy/osu/issues/9561

I started adding a test but it seems like we will need a test resource with video to make it work properly. Not sure I'm ready to put in the time required to make that work right now, maybe next regression?


smoogipoo (approved):
lgtm

Issue #8767, opened 2020-04-15
auto fail mod/customization options


iSimpFor:
this mod will force a fail to occur if the requirements are not met

the customization options are:
accuracy(if below, map is failed)
hp(if below, map is failed) 
a total amount/percentage of any judgement rating(if amount/percentage of not-best ratings is higher than threshold, map is failed or if amount/percentage of best ratings is lower than threshold, map is failed)

all options are disabled by default
an option to make a check on breaks or at the end of the map

these options can be put into customization box without the mod(meaning that the customization button will always be available to access these options)

doughn0:
Also, an option to fail if it is no longer possible to beat your highscore on the map with the current play.

enoslayd:
For mania (and maybe other modes as well), have at least x combo when finishing a beatmap (ex. 200 combo at the end to consider a pass)

Joehuu:
> Also, an option to fail if it is no longer possible to beat your highscore on the map with the current play.

Idea already in https://github.com/ppy/osu/issues/6642.

GSculerlor:
Looking into this one. Should it included on fun mod or can be included on same group with SD/PF? And also any idea about the name? Since I don't think "Auto Fail" is good name here

bdach:
I'm not sure all of these proposed mechanics should be one customisable mod; seems like the mod would be doing too much to give it a single intelligible name or purpose. splitting per proposed fail condition (i.e. something like *Score Challenge*, *Accuracy Challenge*) might fare better. those then could be combined if need be, I suppose.

genericthing:
maybe we should put these options into a new section called "challenge section" and also move sd/pf there

it's a section where you challenge yourself to do better after all so all those mods will fit

peppy:
New section sounds fine.

MitsukaiDokuro:
Shouldn't be a new mod. Would this implemented as Sudden Death mod options.

PR #9584, opened 2020-07-16
Introduce Health Challenge mod

GSculerlor:
First implementation from three proposed challenge mod (#8767). This also introduce new mod section called "Challenge" but for now it only contain this mod. This mod force you to fail if your HP below threshold at the end of the map. Overall it's ready to review, with some note below.

**Proof-of-Concept:** 
https://www.youtube.com/watch?v=_6l3ja-QFRo (skip to the end of the map to see the action)

**Note:**
Still not sure about how to check if the beatmap is complete or not by manually check on mod instead on processor (since bool `HasComplete` is updated after `ApplyResultInternal` so relying on `HasComplete` is no result).

Also for "check on break" proposal, it actually easy to check if it's break time or not (I proposed new mod interface, `IAppliedToBreakTracker` and pass `BreakTracker` to mod level) but can't find any sane approach to trigger fail outside `ApplyResultInternal`

### DIFF
--- a/osu.Game/Overlays/Mods/Sections/ChallengeSection.cs
+++ b/osu.Game/Overlays/Mods/Sections/ChallengeSection.cs
@@ -13,7 +13,7 @@ namespace osu.Game.Overlays.Mods.Sections
 
         public ChallengeSection()
         {
-            Header = @"Challenge";
+            Header = "Challenge";
         }
     }
 }
--- a/osu.Game/Overlays/Mods/Sections/ChallengeSection.cs
+++ b/osu.Game/Overlays/Mods/Sections/ChallengeSection.cs
@@ -1,0 +1,19 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Game.Rulesets.Mods;
+using osuTK.Input;
+
+namespace osu.Game.Overlays.Mods.Sections
+{
+    public class ChallengeSection : ModSection
+    {
+        protected override Key[] ToggleKeys => null;
+        public override ModType ModType => ModType.Challenge;
+
+        public ChallengeSection()
+        {
+            Header = @"Challenge";
+        }
+    }
+}
--- a/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
+++ b/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
@@ -15,7 +15,7 @@ namespace osu.Game.Rulesets.Mods
         public override string Description => "Keep your health above the target!";
 
         [SettingSource("Health Target", "Minimal percentage of health needed to pass.")]
-        public Bindable<double> HealthThreshold { get; } = new BindableDouble
+        public Bindable<double> MinimumHealth { get; } = new BindableDouble
         {
             Precision = 0.01,
             MinValue = 0.50,
@@ -24,7 +24,7 @@ namespace osu.Game.Rulesets.Mods
             Value = 0.96,
         };
 
-        public override string SettingDescription => $"{HealthThreshold.Value:0%}";
+        public override string SettingDescription => $"{MinimumHealth.Value:0%}";
 
         public void ApplyToHealthProcessor(HealthProcessor healthProcessor)
         {
--- a/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
+++ b/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
@@ -38,7 +38,7 @@ namespace osu.Game.Rulesets.Mods
 
             if (!allJudged) return false;
 
-            return result.HealthAtJudgement < HealthThreshold.Value;
+            return healthProcessor.Health.Value < MinimumHealth.Value;
         }
     }
 }
--- a/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
+++ b/osu.Game/Rulesets/Mods/ModHealthChallenge.cs
@@ -1,0 +1,49 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Bindables;
+using osu.Game.Configuration;
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
+
+namespace osu.Game.Rulesets.Mods
+{
+    public class ModHealthChallenge : ModChallenge, IApplicableToHealthProcessor
+    {
+        public override string Name => "Health Challenge";
+        public override string Acronym => "HC";
+        public override string Description => "Hold your health or you fail";
+
+        [SettingSource("Health Threshold", "Set your own health threshold")]
+        public Bindable<double> HealthThreshold { get; } = new BindableAccuracy
+        {
+            Precision = 0.01,
+            MinValue = 0.50,
+            MaxValue = 1.00,
+            Default = 0.96,
+            Value = 0.96,
+        };
+
+        public override string SettingDescription => $"{HealthThreshold}";
+
+        public void ApplyToHealthProcessor(HealthProcessor healthProcessor)
+        {
+            healthProcessor.FailConditions += checkThresholdCondition;
+        }
+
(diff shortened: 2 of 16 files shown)


osu.Game/Overlays/Mods/Sections/ChallengeSection.cs
```diff
@@ -0,0 +1,19 @@
...
+namespace osu.Game.Overlays.Mods.Sections
+{
+    public class ChallengeSection : ModSection
+    {
+        protected override Key[] ToggleKeys => null;
+        public override ModType ModType => ModType.Challenge;
+
+        public ChallengeSection()
+        {
+            Header = @"Challenge";
```
bdach, line 16: nit: raw string prefix is not required
```suggestion
            Header = "Challenge";
```

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace osu.Game.Rulesets.Mods
+{
+    public class ModHealthChallenge : ModChallenge, IApplicableToHealthProcessor
+    {
+        public override string Name => "Health Challenge";
+        public override string Acronym => "HC";
+        public override string Description => "Hold your health or you fail";
+
+        [SettingSource("Health Threshold", "Set your own health threshold")]
```
bdach, line 17: doesn't really explain what the setting means. maybe something like this is better?

```suggestion
        [SettingSource("Health Target", "Minimal percentage of health needed to pass.")]
```

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        public void ApplyToHealthProcessor(HealthProcessor healthProcessor)
+        {
+            healthProcessor.FailConditions += checkThresholdCondition;
+        }
+
+        private bool checkThresholdCondition(HealthProcessor healthProcessor, JudgementResult result)
+        {
+            // Need to manually checked here since HasCompleted is updated after ApplyResultInternal
+            bool allJudged = healthProcessor.JudgedHits == healthProcessor.MaxHits;
```
bdach, line 37: this _is_ the correct way to check for map completion, but as far as I recall this is already done the same way in other places and as such I think it should be exposed in a better way than just making `MaxHits` internal.

also not sure whether this health threshold should be checked only at the end instead of being enforced all the time. not sure which one is more playable.

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
```diff
@@ -0,0 +1,44 @@
...
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
+
+namespace osu.Game.Rulesets.Mods
+{
+    public class ModHealthChallenge : ModChallenge, IApplicableToHealthProcessor
+    {
+        public override string Name => "Health Challenge";
+        public override string Acronym => "HC";
+        public override string Description => "Hold your health or you fail";
```
bdach, line 15: this description ought to be better; not a copywriter by any means, but maybe this is more understandable

```suggestion
        public override string Description => "Keep your health above the target!";
```

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
GSculerlor: It's actually same logic that `JudgementProcessor` use to check beatmap completion and exposed via `HasCompleted` bindable. But since `HasCompleted` [updated after](https://github.com/ppy/osu/blob/master/osu.Game/Rulesets/Scoring/JudgementProcessor.cs#L55-#L61) `ApplyResultInternal` it basically no use here. I'm not sure if there's another exposed variable to check beatmap completion anywhere else.

Also I guess it's a good idea to make threshold checking toggle-able so player can switch between beatmap completion and during gameplay. Will try to make it toggle-able and let's see the feedback.

smoogipoo: I don't mind the way it's done here.

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        private bool checkThresholdCondition(HealthProcessor healthProcessor, JudgementResult result)
+        {
+            // Need to manually checked here since HasCompleted is updated after ApplyResultInternal
+            bool allJudged = healthProcessor.JudgedHits == healthProcessor.MaxHits;
+
+            if (!allJudged) return false;
+
+            return result.HealthAtJudgement < HealthThreshold.Value;
```
smoogipoo, line 41: Isn't this going to use the previous judgement? The xmldoc for `HealthAtJudgement` states "health prior to this <see cref="JudgementResult"/> occurring.". Maybe you should use `Health` instead?

osu.Game/Rulesets/Scoring/JudgementProcessor.cs
```diff
@@ -21,7 +21,7 @@ public abstract class JudgementProcessor : Component
         /// <summary>
         /// The maximum number of hits that can be judged.
         /// </summary>
-        protected int MaxHits { get; private set; }
+        protected internal int MaxHits { get; private set; }
```
smoogipoo, line 24: Should be public (avoid internals unless you really really really need it).

smoogipoo (changes requested):

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
bdach: I am aware that it's the same logic. I'm just saying it should maybe be exposed better overall (I had a need for that same logic once before and just opted not to bother there). `IsLastJudgement` as a property, maybe? Although maybe that's not too good of a property to have public... Not sure.

osu.Game/Rulesets/Mods/ModHealthChallenge.cs
GSculerlor: Ah right! Will change

peppy:
Are you sure we don't want the fail to trigger immediately, rather than at the end of the map?

peppy:
Also, this likely needs a bit further thought: `DrainingHealthProcessor` currently internally finds a minimum allowable health. This health could be where the beatmap ends on, and it could be something *low* (like 25%). So the settings for this mod could actually result in an impossible scenario (which I think we want to avoid).

Maybe this mod should provide its own type of health processor, or maybe it should find the target health for the current beatmap (not sure if this is possible) and set the min/max values based on this.

GSculerlor:
> Are you sure we don't want the fail to trigger immediately, rather than at the end of the map?

I'm thinking about make it toggleable so player can switch between auto trigger or only trigger on completion and made it only available for ruleset with drain HP
image

> Maybe this mod should provide its own type of health processor, or maybe it should find the target health for the current beatmap (not sure if this is possible) and set the min/max values based on this.

I think providing a new type of health processor is more viable solution but not sure if its possible for mod to override ruleset's health processor

GSculerlor:
Looking more into it I think trigger fail only on completion is kinda useless and reduce the challenge itself. Will update to check beatmap completion only for taiko ruleset and trigger fail immediately for other rulesets.

peppy:
not sure i'd give special treatment to taiko.

GSculerlor:
Special treatment here is because taiko's hp is accumulative and it'll trigger fail on first hitobject. Or is there other better approach for it?

peppy:
Heavily depends on implementation. See how you go, I guess, and we'll work from there.

GSculerlor:
I'm thinking about remove passive drain for this mod as a solution for minimum allowable health issue (I tried to search any approach to set min/max value based on min allowable health from health processor but have no luck). Not sure if it's good enough with this proposal. Note that this changes only applied on ruleset with drain health.

https://streamable.com/7robqb

bdach:
Removing drain to make this work seems like a bad workaround as it makes the mod no longer just a "health challenge". Maybe the health threshold should be moving (in being a percentage of the amount of health on a perfect play)?

GSculerlor:
That could work but not sure if it easy for players to understand what's that mean since the number on mod setting not represent real threshold

GSculerlor:
closing this PR for now due to inactivity. might reopen it when I have time to look into it again.

PR #9590, opened 2020-07-17
Fix reversing scroll direction not always behaving as expected

peppy:
This fixes two similar issues:

- Using left and right arrows in the editor can someone (still) get stuck on the first reverse direction press. This was due to the `AlmostEquals` check not being lenient enough.
- Scrolling with precise scroll didn't feel good when reversing direction due to the completed portion of the "precision" tick not being counted for the inverse movement. This would result in a large movement requirement before the seek started to feel responsive again.

Added some more documentation around this code since it is a bit custom.

Previously:

Note "sticking" behaviour when reversing direction while fractionally close to the snap point.

Now:

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -275,11 +275,22 @@ namespace osu.Game.Screens.Edit
 
         protected override bool OnScroll(ScrollEvent e)
         {
-            scrollAccumulation += (e.ScrollDelta.X + e.ScrollDelta.Y) * (e.IsPrecise ? 0.1 : 1);
+            const double precision = 1;
 
-            const int precision = 1;
+            double scrollComponent = e.ScrollDelta.X + e.ScrollDelta.Y;
 
-            while (Math.Abs(scrollAccumulation) > precision)
+            double scrollDirection = Math.Sign(scrollComponent);
+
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Edit/Editor.cs
```diff
@@ -275,11 +275,22 @@ protected override bool OnKeyDown(KeyDownEvent e)
...
+            double scrollComponent = e.ScrollDelta.X + e.ScrollDelta.Y;
 
-            while (Math.Abs(scrollAccumulation) > precision)
+            double scrollDirection = Math.Sign(scrollComponent);
+
+            // this is a special case to handle the "pivot" scenario.
+            // if we are precise scrolling in one direction then change our mind and scroll backwards,
+            // the existing accumulation should be applied in the inverse direction to maintain responsiveness.
+            if (Math.Sign(scrollAccumulation) != scrollDirection)
+                scrollAccumulation = scrollDirection * (precision - Math.Abs(scrollAccumulation));
```
smoogipoo, line 288: Did you want to abs this again? like `scrollDirection * Math.Abs(precision - Math.Abs())`?

Just thinking if you have a large negative accumulation and change to positive, the calc is going to be `1 * (1 - (>1)) = 1 * (<0) = <0`

smoogipoo (approved):

PR #9592, opened 2020-07-17
Add texture wrapping support to skinning, fix taiko and mania

smoogipoo:
I think this may be a breaking change for the karaoke ruleset (hopefully not but haven't tested), but I don't think there's a way around it.

Notable decisions I've made:
- Implemented interface method. Most use-cases will use the `ISkin` or `ISkinSource` interface.
- Exposed interface method from classes where it's likely to be used from the outside (`LegacySkinTransformer`, `Skin`).


peppy (approved):

enoslayd:
Is this about the discoloration at the end of mania and taiko sliders?

bdach:
Correct

PR #9595, opened 2020-07-17
Add slider to the grow/deflate mod

salmonslay:
Implements a slider for the grow and deflate mod to customize the object starting size. 

Video of the mod showcasing different object sizes: https://youtu.be/PO29f-mZ_XE
Video of the deflate mod with 25x start size: https://youtu.be/o84RBTEmOGM

Default values are left the same, and the slider values are:
- Grow:
  - Min: x0
  - Max: x0.99
  - Precision: 0.01
- Deflate:
  - Min: x1
  - Max: x25
  - Precision: 0.1
  - Higher values works a lot better at lower AR (<7)


osu.Game.Rulesets.Osu/Mods/OsuModObjectScaleTween.cs
```diff
@@ -23,7 +23,7 @@ public abstract class OsuModObjectScaleTween : Mod, IReadFromConfig, IApplicable
 
         public override double ScoreMultiplier => 1;
 
-        protected virtual float StartScale => 1;
+        public abstract BindableNumber<float> StartScale { get; }
```
bdach, line 26: This could have stayed `protected` and just delegated to the bindable's `Value` in grow/shrink mods.

osu.Game.Rulesets.Osu/Mods/OsuModDeflate.cs
```diff
@@ -15,6 +17,14 @@ public class OsuModDeflate : OsuModObjectScaleTween
 
         public override string Description => "Hit them at the right size!";
 
-        protected override float StartScale => 2f;
+        [SettingSource("Starting size modifier", "The object starting size modifier")]
```
bdach, line 20: doesn't read too well. maybe this is better?

```suggestion
        [SettingSource("Starting Size", "The initial size multiplier applied to all objects.")]
```

osu.Game.Rulesets.Osu/Mods/OsuModGrow.cs
```diff
@@ -15,6 +17,14 @@ internal class OsuModGrow : OsuModObjectScaleTween
...
 
-        protected override float StartScale => 0.5f;
+        [SettingSource("Starting size modifier", "The object starting size modifier")]
+        public override BindableNumber<float> StartScale { get; } = new BindableFloat
+        {
+            MinValue = 0f,
+            MaxValue = 0.99f,
+            Default = 0.5f,
+            Value = 0.5f,
+            Precision = 0.01f,
```
bdach, line 27: not sure I like this having a `0.01` step and being from `0-0.99` too much. I kind of think having this work as an inverse of `OsuModShrink` (so allowing initial scales of 1/2, 1/3, 1/4, ..., 1/25) might be better. would like opinions, however.

osu.Game.Rulesets.Osu/Mods/OsuModDeflate.cs
LiterallyFabian: Tried to making it clear that the number is a modifier and not approach rate, but there shouldn't be any problems. I prefer the "initial size multiplier" one over the "object starting size", although I never figured out where this is shown in-game

bdach: Upon inspection it doesn't seem to be shown in-game, although it could be (as a tooltip, for example).

osu.Game.Rulesets.Osu/Mods/OsuModObjectScaleTween.cs
LiterallyFabian: I made the StartScale public & abstract as those were the modifiers used by `ModRateAdjust` that also uses a slider for changing its value, I could also not get it to work with `protected` as the customization button was grayed while using it, I'm not quite sure what to edit here

bdach: Right now it doesn't matter too much. If [this suggestion](https://github.com/ppy/osu/pull/9595#discussion_r456827279) is deemed the correct direction then my proposal will make more sense. You would do

```csharp
public BindableNumber<float> StartScaleSetting { get; }
protected override float StartScale => StartScaleSetting.Value;
```

osu.Game.Rulesets.Osu/Mods/OsuModGrow.cs
LiterallyFabian: I agree that 0.01 is too small and it doesn't make much sense having it close to 1, mostly went with the values used by the HT/DT mods as they have the same precision and allows you to modify the speed by 0.99/1.01. 
If I have understood everything correctly the `OsuModGrow` already is an inverse of `OsuModDeflate`, as giving the grow mod the value 2 for example makes it deflate instead?

osu.Game.Rulesets.Osu/Mods/OsuModObjectScaleTween.cs
peppy: I think we can go with this as a starting point and tweak the UX at a later stage.

peppy (approved):

PR #9598, opened 2020-07-17
Split rankings overlay tabs into separate components

EVAST9919:
Depends on:
- [x] https://github.com/ppy/osu/pull/9639
- [x] https://github.com/ppy/osu/pull/9641


osu.Game/Overlays/Rankings/Displays/CountriesDisplay.cs
```diff
@@ -0,0 +1,23 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics;
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests;
+using osu.Game.Overlays.Rankings.Tables;
+
+namespace osu.Game.Overlays.Rankings.Displays
+{
+    public class CountriesDisplay : RankingsDisplay
```
bdach, line 11: I was very confused as to what this is. needs to be called `CountryRankingsDisplay`

osu.Game/Overlays/Rankings/Displays/PerformanceDisplay.cs
```diff
@@ -0,0 +1,65 @@
...
+using osu.Framework.Bindables;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Online.API.Requests;
+using osu.Game.Overlays.Rankings.Tables;
+using osu.Game.Users;
+using osu.Game.Online.API;
+
+namespace osu.Game.Overlays.Rankings.Displays
+{
+    public class PerformanceDisplay : RankingsDisplay
```
bdach, line 14: `PerformanceRankingsDisplay`

osu.Game/Overlays/Rankings/Displays/PerformanceDisplay.cs
```diff
@@ -0,0 +1,65 @@
...
+                {
+                    RelativeSizeAxes = Axes.X,
+                    AutoSizeAxes = Axes.Y,
+                    Direction = FillDirection.Vertical,
+                    Children = new Drawable[]
+                    {
+                        new CountryFilter
+                        {
+                            Current = Country
+                        },
```
bdach, line 45: has the same awkward spacing as the spotlights ranking display used to have

2020-07-18-230529_1247x166_scrot

osu.Game/Overlays/Rankings/Displays/ScoresDisplay.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using osu.Framework.Graphics;
+using osu.Framework.Bindables;
+using osu.Game.Online.API.Requests;
+using osu.Game.Overlays.Rankings.Tables;
+using osu.Game.Online.API;
+
+namespace osu.Game.Overlays.Rankings.Displays
+{
+    public class ScoresDisplay : RankingsDisplay
```
bdach, line 12: `ScoreRankingsDisplay`

osu.Game/Overlays/Rankings/Displays/SpotlightsDisplay.cs
```diff
@@ -0,0 +1,79 @@
...
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests;
+using osu.Game.Overlays.Rankings.Tables;
+using System.Linq;
+using osu.Game.Overlays.BeatmapListing.Panels;
+using System.Collections.Generic;
+
+namespace osu.Game.Overlays.Rankings.Displays
+{
+    public class SpotlightsDisplay : RankingsDisplay
```
bdach, line 20: `SpotlightsRankingsDisplay`

osu.Game/Overlays/UserProfileOverlay.cs
```diff
@@ -26,7 +26,7 @@ public class UserProfileOverlay : FullscreenOverlay
         private ProfileSectionsContainer sectionsContainer;
         private ProfileSectionTabControl tabs;
 
-        public const float CONTENT_X_MARGIN = 70;
+        public const float CONTENT_X_MARGIN = 50;
```
bdach, line 29: this change is not valid. the user profile design still uses 70px padding, and rankings use 50px. you can't conflate two constants (especially in two disparate overlay namespaces) like this.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -90,154 +78,112 @@ private void load()
...
+                case RankingsScope.Spotlights:
+                    return new SpotlightsDisplay();
             }
 
-            return null;
+            throw new NotImplementedException($"Display for {Scope.Value} is not implemented.");
         }
 
-        private Drawable createTableFromResponse(APIRequest request)
+        public void ShowCountry(Country requested)
```
bdach, line 167: the flag clicks use this method anyway, so why not get rid of the `Country` bindable entirely?

bdach (changes requested):
this reviews dreadfully and probably could have used splitting. this is a rudimentary first review, there might be more issues to fix

osu.Game/Overlays/Rankings/Displays/PerformanceDisplay.cs
EVAST9919: It will be changed anyway in the near future (in line with the web)

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: Well, we don't want to reload the whole display by changing the country. If current display is `Performance` - we just need to pass the `Country` value so it will reload it's content. Probably we can check if selected overlay is performance one, but it will be worse imo

gagahpangeran:
Few days ago I made a change in ranking header on website https://github.com/ppy/osu-web/pull/6306

The new sort tab looks like this
image

EVAST9919:
> Few days ago I made a change in ranking header on website [ppy/osu-web#6306](https://github.com/ppy/osu-web/pull/6306)
> 
> The new sort tab looks like this
> image

yeah, I've seen it, thanks. Will be reworked later, but not in this pr for sure

gagahpangeran:
> yeah, I've seen it, thanks. Will be reworked later, but not in this pr for sure

More important for now, it must be hidden when user is not login.

EVAST9919:
> > yeah, I've seen it, thanks. Will be reworked later, but not in this pr for sure
> 
> More important for now, it must be hidden when user is not login.

Already is after latest commit

osu.Game/Overlays/RankingsOverlay.cs
bdach: you should be able to move out three logic in the `BindValueChanged` callback to this method, no? at which point the bindable has no more purpose.

EVAST9919: Right, but there's another purpose of this bindable: to pass `Country` value to the `PerformanceDisplay`

osu.Game/Overlays/OverlayHeader.cs
```diff
@@ -12,7 +12,7 @@ namespace osu.Game.Overlays
 {
     public abstract class OverlayHeader : Container
     {
-        public const int CONTENT_X_MARGIN = 50;
+        protected virtual int ContentSideMargin() => 70;
```
bdach, line 15: should be a property instead of being a method, and should have xmldoc as well.

osu.Game.Tests/Visual/Online/TestSceneRankingsPerformanceDisplay.cs
```diff
@@ -0,0 +1,47 @@
...
+using osu.Game.Rulesets.Osu;
+using osu.Framework.Allocation;
+using osu.Game.Overlays;
+using osu.Game.Overlays.Rankings.Displays;
+using osu.Game.Users;
+using NUnit.Framework;
+
+namespace osu.Game.Tests.Visual.Online
+{
+    public class TestSceneRankingsPerformanceDisplay : OsuTestScene
```
bdach, line 15: name is backwards

```suggestion
    public class TestScenePerformanceRankingsDisplay : OsuTestScene
```

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -9,32 +9,35 @@
...
+using osu.Game.Online.API;
 
 namespace osu.Game.Overlays
 {
     public class RankingsOverlay : FullscreenOverlay
     {
-        protected Bindable<Country> Country => header.Country;
+        public const float CONTENT_X_MARGIN = 50;
+
+        protected readonly Bindable<Country> Country = new Bindable<Country>();
```
bdach, line 25: let's put it this way: ideally I'd like for this to not exist here. the rankings overlay should not need to know what country its sub-view is displaying.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -90,52 +82,56 @@ private void load()
...
+        {
+            base.PopIn();
+
+            // We don't want to create a new display on every call, only when exiting from fully closed state.
+            if (displayUpdateRequired)
+            {
+                header.Current.TriggerChange();
+                displayUpdateRequired = false;
+            }
+        }
```
bdach, line 118: as far as I can see this was copied verbatim from `DashboardOverlay`. in that case maybe it should be added to the base in some way - but not necessarily in this PR as this one is bordering on too big already.

also, `displayUpdateRequired` looks like something `Cached` was designed for.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -144,100 +140,104 @@ public void ShowSpotlights()
...
-            lastRequest = request;
+            var display = selectDisplay();
 
-            if (request == null)
+            if (display is SpotlightsRankingsDisplay)
             {
-                loadContent(null);
+                loadSpotlightsDisplay(display);
                 return;
             }
```
bdach, line 162: this is weird. `selectDisplay` is a generic factory method, but then whatever gets created gets pattern-matched anyway (as spotlights have different logic). that logic should arguably reside in the displays themselves.

bdach (changes requested):
reviews even worse now...

bdach:
this just hit XXL size, which means splitting is _direly_ needed if this is to be reviewable. this PR does three or four disparate things (some visual changes, some structural changes due to refactoring separate displays for separate ranking types, some reuse of pre-existing abstract classes) and those would all be better served as separate PRs.

EVAST9919:
> this just hit XXL size, which means splitting is _direly_ needed if this is to be reviewable. this PR does three or four disparate things (some visual changes, some structural changes due to refactoring separate displays for separate ranking types, some reuse of pre-existing abstract classes) and those would all be better served as separate PRs.

Will do

osu.Game/Overlays/Rankings/Displays/ScoreRankingsDisplay.cs
```diff
@@ -0,0 +1,34 @@
...
+        {
+            base.LoadComplete();
+            sort.BindValueChanged(_ => PerformFetch());
+        }
+
+        protected override APIRequest<GetUsersResponse> CreateRequest() => new GetUserRankingsRequest(Current.Value, UserRankingsType.Score, sort.Value);
+
+        protected override Drawable CreateHeader() => new RankingsSortTabControl
+        {
+            Margin = new MarginPadding { Vertical = 20, Right = RankingsOverlay.CONTENT_X_MARGIN },
```
bdach, line 26: I think this shouldn't be applied locally, but to the entire header created in the base `RankingsDisplay`.

osu.Game.Tests/Visual/Online/TestSceneSpotlightsLayout.cs
```diff
@@ -32,9 +33,17 @@ public TestSceneSpotlightsLayout()
                 Origin = Anchor.Centre,
                 RelativeSizeAxes = Axes.Both,
                 Width = 0.8f,
-                Child = new SpotlightsLayout
+                Child = new SpotlightsRankingsDisplay
```
bdach, line 36: scene needs renaming to `TestSceneSpotlightsRankingsDisplay`.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -9,31 +9,35 @@
...
+        private readonly OverlayScrollContainer scrollFlow;
 
-        private APIRequest lastRequest;
+        private GetSpotlightsRequest spotlightsRequest;
         private CancellationTokenSource cancellationToken;
 
-        [Resolved]
-        private IAPIProvider api { get; set; }
+        private PerformanceRankingsDisplay performanceDisplay;
+        private Country lastRequested;
```
bdach, line 40: why does this even exist?

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -144,100 +139,101 @@ public void ShowSpotlights()
...
+                var display = new SpotlightsRankingsDisplay
+                {
+                    Current = ruleset,
+                    StartLoading = loading.Show,
+                    FinishLoading = loading.Hide,
+                    Spotlights = response.Spotlights
+                };
+                loadDisplayAsync(display);
+            });
+            API.Queue(spotlightsRequest);
```
bdach, line 179: this logic should ideally be in the spotlights display, not out here like this. the overlay should only be managing navigation between displays.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -9,31 +9,35 @@
...
         private readonly RankingsOverlayHeader header;
+        private readonly OverlayScrollContainer scrollFlow;
 
-        private APIRequest lastRequest;
+        private GetSpotlightsRequest spotlightsRequest;
         private CancellationTokenSource cancellationToken;
 
-        [Resolved]
-        private IAPIProvider api { get; set; }
+        private PerformanceRankingsDisplay performanceDisplay;
```
bdach, line 39: I'd be more accepting of this if it were of a type that encapsulates all possible displays and pattern-matches to check whether it's a performance display wherever necessary.

bdach (changes requested):

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: we need to store somwhere requested country while we loading a new display

osu.Game/Overlays/Rankings/Displays/ScoreRankingsDisplay.cs
EVAST9919: Not every display has a header. Applying margin/padding to the header container will show empty container 40px in height if the content is empty.

osu.Game/Overlays/RankingsOverlay.cs
EVAST9919: Moved to spotlights display

osu.Game/Overlays/RankingsOverlay.cs
bdach: having stared at this code for the better part of an hour i suppose the two fields can stay as the data flow is severely limited by hard-coupling to the header tabs and their associated enum and also all the schedules areound here.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -144,100 +137,84 @@ public void ShowSpotlights()
             Show();
         }
 
-        private void loadNewContent()
+        private void selectDisplayToLoad()
```
bdach, line 140: this naming is confusing; there is `selectDisplayToLoad()` and `selectDisplay()`. this should be `changeDisplay()` if anything, and the other methods should be named `createDisplay()` as that's what it actually does.

bdach (changes requested):

osu.Game/Overlays/Rankings/Displays/ScoreRankingsDisplay.cs
bdach: there are ways to circumvent this issue. i do not believe it should ever be valid for any header to just completely disregard the horizontal padding. (vertical padding can stay local as it varies.)

osu.Game/Overlays/Rankings/Displays/RankingsDisplay.cs
```diff
@@ -0,0 +1,127 @@
...
+        public Bindable<RulesetInfo> Current
+        {
+            get => current.Current;
+            set => current.Current = value;
+        }
+
+        /// <summary>
+        /// Whether we should create content on request success.
+        /// </summary>
+        protected virtual bool CreateContentOnSucess => true;
```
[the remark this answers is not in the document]
EVAST9919: The reason for this is that spotlights display has 2 requests in it and first request is only to get available spotlights, so there's no content to add. But if we will still try to add `Empty()` - animation will look incorrect since if we already have a table - it will be removed at first and added another table later. With this wew are kinda blocking content addition on first request success, so there will be no flickering and one table will softly transit to another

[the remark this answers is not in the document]
bdach: then why isn't `OnSuccess` overridden instead of adding this flag which complicates matters even more?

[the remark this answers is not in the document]
EVAST9919: wow, that's a nice point of view

EVAST9919:
Closing for now, too big for decent review, will split.

PR #9603, opened 2020-07-18
Use static instances in all fallback ControlPoint lookups to reduce allocations

peppy:
Before:

image

After:

image

I know we want to make these kinds of static 100% safe, but this is the best we can do in this context without larger changes (I think the disabled flag makes it very unlikely to ever cause issue).

### DIFF
--- a/osu.Game/Beatmaps/ControlPoints/ControlPointInfo.cs
+++ b/osu.Game/Beatmaps/ControlPoints/ControlPointInfo.cs
@@ -64,14 +64,14 @@ namespace osu.Game.Beatmaps.ControlPoints
         /// </summary>
         /// <param name="time">The time to find the difficulty control point at.</param>
         /// <returns>The difficulty control point.</returns>
-        public DifficultyControlPoint DifficultyPointAt(double time) => binarySearchWithFallback(DifficultyPoints, time);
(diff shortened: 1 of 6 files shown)


osu.Game/Beatmaps/ControlPoints/ControlPointInfo.cs
```diff
@@ -170,12 +170,12 @@ public void RemoveGroup(ControlPointGroup group)
...
-        /// <param name="prePoint">The control point to use when <paramref name="time"/> is before any control points. If null, a new control point will be constructed.</param>
+        /// <param name="fallback">The control point to use when <paramref name="time"/> is before any control points.</param>
         /// <returns>The active control point at <paramref name="time"/>, or a fallback <see cref="ControlPoint"/> if none found.</returns>
-        private T binarySearchWithFallback<T>(IReadOnlyList<T> list, double time, T prePoint = null)
-            where T : ControlPoint, new()
+        private T binarySearchWithFallback<T>(IReadOnlyList<T> list, double time, T fallback)
+            where T : ControlPoint
         {
-            return binarySearch(list, time) ?? prePoint ?? new T();
+            return binarySearch(list, time) ?? fallback;
```
bdach, line 178: The removal of the `new T()` final-chance-fallback is causing the test failures here.

bdach:
Looks a little dicey but this is probably the best that can be done unless we add a truly immutable bindable to the framework... Not overly sure about this one so I'll abstain from approving for now.

smoogipoo (approved):
This is probably ok..

PR #9604, opened 2020-07-18
Remove remaining per-Update transform in OsuLogo to reduce allocations

peppy:
Before:

image

After:

image


osu.Game/Screens/Menu/OsuLogo.cs
```diff
@@ -330,7 +330,7 @@ protected override void Update()
             if (Beatmap.Value.Track.IsRunning)
             {
                 var maxAmplitude = lastBeatIndex >= 0 ? Beatmap.Value.Track.CurrentAmplitudes.Maximum : 0;
-                logoAmplitudeContainer.ScaleTo(1 - Math.Max(0, maxAmplitude - scale_adjust_cutoff) * 0.04f, 75, Easing.OutQuint);
+                logoAmplitudeContainer.Scale = new Vector2((float)Interpolation.Damp(logoAmplitudeContainer.Scale.X, 1 - Math.Max(0, maxAmplitude - scale_adjust_cutoff) * 0.04f, 0.5f, Time.Elapsed));
```
bdach, line 333: This is going to be nitpicky as all hell, but I had a look at the math. As far as I can see `Interpolation.Damp` is using a curve of 1-0.5<sup>*x*</sup> wherein *x* is `Time.Elapsed`. Plotting that yields the following:

2020-07-18-112826_415x456_scrot

As you can see calling this with a value bigger than ~8 pretty much just yields a straight 1. As `Time.Elapsed` is in milliseconds, this means that this transition will last 8ms, which is basically unnoticeable on anything with a refresh rate <120Hz; at that point you could just skip the interpolation and just assign the final value instead.

I've added rudimentary logging of `Scale` right after this line and it seems to confirm all of the above (there are noticeable "snaps" of scale, as indicated by the following excerpt:)

```
2020-07-18 09:12:30 [verbose]: scale = 0.99084646
2020-07-18 09:12:30 [verbose]: scale = 0.9908462
2020-07-18 09:12:30 [verbose]: scale = 0.9908462
2020-07-18 09:12:30 [verbose]: scale = 0.99330133
2020-07-18 09:12:30 [verbose]: scale = 0.97678745 <- snapped here
2020-07-18 09:12:30 [verbose]: scale = 0.97607374
2020-07-18 09:12:31 [verbose]: scale = 0.97600335
2020-07-18 09:12:31 [verbose]: scale = 0.9760002
2020-07-18 09:12:31 [verbose]: scale = 0.976
```

peppy: Yeah, I'm not sure Damp is the function we want. I was just using the nearby one as a reference. Probably `Lerp` instead.

peppy: The arguments specified here were actually back-to-front. Because I wasn't 100% on the underlying math I made [a sheet](https://docs.google.com/spreadsheets/d/1EKiH1schGlL7s9dGsHY9YUTP2i4kqiymDsM3d63OEcQ/edit?usp=sharing) to confirm it's working as we expect. A bit confusing that the "exponent" in this case is the elapsed time.

bdach (approved):
looks good to me :+1:

smoogipoo (approved):

PR #9606, opened 2020-07-18
Add "Alternate" mod

### LINKED ISSUE #3773: A mod to only allow alternating

LeNitrous:
Closes https://github.com/ppy/osu/issues/3773 and supersedes https://github.com/ppy/osu/pull/7953. This mod forces the player to alternate between keys for osu and taiko rulesets. It injects a keybind handler to the drawable ruleset and intercepts and reads input allowing the mod to allow or disallow certain actions.

Considerations:
* Key states are reset after break times as usually people tend to forget what last key they pressed causing unintended misses.
* In taiko, big drum hits can be hit with 2 keys. I dislike the idea of a mod preventing basic gameplay mechanics such as this. However, if you hit one with 1 key, you must alternate with another key for the next drum hit.

Possible considerations:
* I'd like to keep the key counters as is to find out why they missed in replays or recordings.

Issues:
* [x] ~~In osu, sliders still judge tracking thus giving score to ticks, repeats, and ends. This may be because it listens for a button down input instead of listening to an `InputAction`.~~ This is intentional to allow switching between keys while tracking inside a slider.
* [x] In taiko, strong hit objects may have not reset the input state yet causing other hit objects inside their hit window to possibly miss (as we don't know what key was last used)

Prerequisites: #10472


osu.Game/Rulesets/Mods/ModAlternate.cs
```diff
@@ -0,0 +1,82 @@
...
+{
+    public abstract class ModAlternate : Mod
+    {
+        public override string Name => "Alternate";
+        public override string Acronym => "AL";
+        public override ModType Type => ModType.Fun;
+        public override double ScoreMultiplier => 1;
+        public override string Description => "Never use the same key twice!";
+        public override Type[] IncompatibleMods => base.IncompatibleMods.Append(typeof(ModAutoplay)).ToArray();
+    }
```
bdach, line 24: Couldn't this just be inside the generic variant?

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,77 @@
...
+                {
+                    if (!hitObject.IsStrong) continue;
+
+                    var window = hitObject.HitWindows.WindowFor(HitResult.Miss);
+                    interceptor.Delay(hitObject.StartTime - window).Schedule(() => resetStates());
+                    interceptor.Delay(hitObject.StartTime + window).Schedule(() => resetStates());
+
+                    foreach (TaikoHitObject h in hitObject.NestedHitObjects)
+                    {
+                        interceptor.Delay(h.StartTime - 50).Schedule(() => resetStates());
```
bdach, line 37: What is `50` and why is it hard-coded?

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
```diff
@@ -0,0 +1,37 @@
...
+        private OsuAction? lastActionReleased;
+
+        protected override void OnBreakEnd()
+        {
+            lastActionPressed = lastActionReleased = null;
+        }
+
+        protected override bool OnPressed(OsuAction action)
+        {
+            if (lastActionPressed != null && lastActionPressed == action)
```
bdach, line 21: The null-check is redundant. The incoming `OsuAction` is of a value type, so it can't ever be equal to `null`.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,77 @@
...
+
+            using (Interceptor.BeginAbsoluteSequence(0))
+            {
+                foreach (TaikoHitObject hitObject in HitObjects)
+                {
+                    if (!hitObject.IsStrong) continue;
+
+                    var window = hitObject.HitWindows.WindowFor(HitResult.Miss);
+                    interceptor.Delay(hitObject.StartTime - window).Schedule(() => resetStates());
+                    interceptor.Delay(hitObject.StartTime + window).Schedule(() => resetStates());
```
bdach, line 33: Not sure this is completely correct; what if there are two objects with overlapping hit windows?

osu.Game/Rulesets/Mods/ModAlternate.cs
```diff
@@ -0,0 +1,82 @@
...
+
+    public abstract class ModAlternate<THitObject, TAction> : ModAlternate, IApplicableToDrawableRuleset<THitObject>
+        where THitObject : HitObject
+        where TAction : struct
+    {
+        public List<BreakPeriod> Breaks;
+        public List<THitObject> HitObjects;
+        public InputInterceptor Interceptor;
```
bdach, line 32: should be definitely `readonly` and probably `protected` to boot

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
LeNitrous: It should be intended for drum rolls only. I'll add a type check before that.

bdach:
The kkdd vs kddk discussion from the previous PR seems to be unaddressed, although I'm not sure how important that is. The transform usage seems somewhat strange as well; I'd probably construct a list of times upon which the button state should be reset, and then iterate over it during gameplay in `Update()` (first frame that crosses the given time causes a reset and moves the iterator pointing at the next time instance that should cause a reset).

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
LeNitrous: `lastActionPressed` and `lastActionReleased` are nullables. They are both set to null at the start of the map and after breaks.

bdach: They are, but what I'm getting at is that `lastActionPressed == action` can't ever be true if `lastActionPressed` is null.

enoslayd:
I know this comment might get deleted but....


What if instead of failing, it will become miss instead (or add this as an option?)

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
LeNitrous: The hit window is based off of here.
https://github.com/ppy/osu/blob/fb5a54d2422171d8f2cb40ee1cb4b9b96dabc4f7/osu.Game.Rulesets.Taiko/Objects/DrumRoll.cs#L53-L57

bdach:
There *is* some feedback already to the user in that the taiko drum parts won't light up if the player fails to alternate. Misses will also still register, but not immediately upon mistakenly not alternating - they work as if you just didn't press the key (whether they should is arguable, as they add the ability to recover by alternating again).

enoslayd:
Does it apply to osu ruleset too?

bdach:
Yes, with the same caveats.

LeNitrous:
Currently I just based off of "alternating" as alternating between keys. Also using transforms is overkill too. I like the idea of storing timestamps in an array and is much more safer as it accounts for rewinding. However I don't think I have a way to access the gameplay clock's current time.

@enoslayd  It currently blocks your input thus making you miss. You can just add Sudden Death or Perfect if you want the option to fail.

bdach:
Isn't it just `Time.Current`? As you're adding the interceptor to the drawable ruleset hierarchy, it *should* be under the gameplay clock.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
bdach: That value is not always valid, though; it's merely a default. Check other usages in `DrumRoll` (it is mutated based on timing points and tick rate).

enoslayd:
Some suggestions (delete if this comment not necessary):

1. In my opinion, It should ignore swells (spinners). You just generally bash them
2. After spinners, i think the keys should return to normal. What i mean is that they will be reset like when entering a break (maybe also consider drumrolls (sliders) too)

LeNitrous:
Applied bdach's suggestion of iterating over the current time instead of using a lot of transforms. Plus this should properly track strong hit objects now as it doesn't rely on their hit windows (it checks whether the hit object is a strong object and resets the input states to guarantee we can use either key in the next object).

As for the suggestions by enoslayd and the concern about **kkdd** and **kddk** in the previous PR, it would be up for debate by other people for now.

LeNitrous:
I've wrapped up the correct way on handling strong hit objects now although I seem to be confused on how to implement the key checks in taiko. I would love to hear the opinions of taiko players on this one:

Currently I have two options on key checking (L - left / R - Right / D - Don / K - Kat):
* Strict checking: LD LK RD RK (remembers the last don and kat used)
* Lenient checking: LD LK LD RK (ignores the last don used when the next is a kat and vice-versa)

* For normal dons and kats use the lenient checking.
* For normal drum rolls including swells (spinner), use strict checking.

LeNitrous:
OK after some considerations this should handle cases with strong hit objects (big drum hits) better. It's difficult to disambiguate between 1 key press and 2 key presses as we can't anticipate what the player will do on the first frame. As a solution, this is how it handles those cases:

1. If the next hit object is a strong hit object, allow any key to be used. This allows 2 keys to be used.
2. If the strong hit object is hit with 2 keys, any key can be used for the next hit object.
3. However, if it is hit with 1 key, that key can't be used for the next hit object.

This should avoid cheesing but not in all cases due to the complexity of the problem as stated above.

Next, I added two modes as stated in the first PR. These are __KDDK__ and __KKDD__ respectively, referring to the key bindings Kat and Don respectively. The names are temporary for the two modes aren't very friendly yet as they're taiko player jargons. If there are better names, it should be replaced.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,116 @@
...
+            });
+
+            if (hitObject?.IsStrong ?? false)
+            {
+                if ((hitObject == lastHitObject) && (Interceptor.Time.Current - lastActionTime <= strong_hit_window))
+                {
+                    if ((lastAction == TaikoAction.LeftRim) && (action == TaikoAction.RightRim) ||
+                        (lastAction == TaikoAction.RightRim) && (action == TaikoAction.LeftRim) ||
+                        (lastAction == TaikoAction.LeftCentre) && (action == TaikoAction.RightCentre) ||
+                        (lastAction == TaikoAction.RightCentre) && (action == TaikoAction.LeftCentre))
```
bdach, line 50: These parens are very weird to me; they don't really help me read this (I *sort of* remember operator precedence rules for `&&` and `||` but it would be more readable if they were explicit). Would probably change to
```suggestion
                    if ((lastAction == TaikoAction.LeftRim && action == TaikoAction.RightRim) ||
                        (lastAction == TaikoAction.RightRim && action == TaikoAction.LeftRim) ||
                        (lastAction == TaikoAction.LeftCentre && action == TaikoAction.RightCentre) ||
                        (lastAction == TaikoAction.RightCentre && action == TaikoAction.LeftCentre))
```

(You could also use a `compareUnordered` here which I wrote a version of in the other suggestion.)

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var hitObject = HitObjects.FirstOrDefault((h) =>
+            {
+                var time = Interceptor.Time.Current;
+                var window = h.HitWindows.WindowFor(HitResult.Miss);
+                var endTime = (h as IHasDuration)?.EndTime ?? h.StartTime;
+                return (time > h.StartTime - window) && (time < endTime + window);
+            });
+
+            if (hitObject?.IsStrong ?? false)
```
bdach, line 43: I'd extract the `if` branch to a method (`handleStrongHit`) to reduce nesting & complexity.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,116 @@
...
+    {
+        [SettingSource("Playstyle", "Preferred Alternate Playstyle")]
+        public Bindable<Playstyle> Style { get; } = new Bindable<Playstyle>();
+
+        private TaikoAction? lastAction1;
+        private TaikoAction? lastAction2;
+        private TaikoAction lastAction;
+        private TaikoHitObject lastHitObject;
+        private double lastActionTime = 0;
+        private const double strong_hit_window = 30;
```
bdach, line 24: Where does this number come from?

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,116 @@
...
+namespace osu.Game.Rulesets.Taiko.Mods
+{
+    public class TaikoModAlternate : ModAlternate<TaikoHitObject, TaikoAction>
+    {
+        [SettingSource("Playstyle", "Preferred Alternate Playstyle")]
+        public Bindable<Playstyle> Style { get; } = new Bindable<Playstyle>();
+
+        private TaikoAction? lastAction1;
+        private TaikoAction? lastAction2;
+        private TaikoAction lastAction;
```
bdach, line 21: These need an explanation at worst and a proper rename at best. Having three `lastAction`s with numbers in them is quite unreadable.

If I were to write something that doesn't necessitate these two, I'd go for:

```csharp
private bool shouldBlock(TaikoAction currentAction)
{
    var actionsToCheck = (lastAction, currentAction);

    switch (Style.Value)
    {
        case Playstyle.KDDK:
            return compareUnordered((TaikoAction.LeftCentre, TaikoAction.LeftRim), actionsToCheck)
                   || compareUnordered((TaikoAction.RightCentre, TaikoAction.RightRim), actionsToCheck);

        case Playstyle.KKDD:
            return compareUnordered((TaikoAction.LeftCentre, TaikoAction.RightCentre), actionsToCheck)
                   || compareUnordered((TaikoAction.LeftRim, TaikoAction.RightRim), actionsToCheck);

        default:
            throw new ArgumentOutOfRangeException(nameof(currentAction));
    }
}

private bool compareUnordered((TaikoAction, TaikoAction) expected, (TaikoAction, TaikoAction) actual)
    => (expected.Item1 == actual.Item1 && expected.Item2 == actual.Item2)
       || (expected.Item1 == actual.Item2 && expected.Item2 == actual.Item1);
```

I *think* this works but haven't tested. Buyer beware. (Mostly posting for the overall structure.)

osu.Game/Rulesets/Mods/ModAlternate.cs
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    public abstract class ModAlternate<THitObject, TAction> : ModAlternate, IApplicableToDrawableRuleset<THitObject>
+        where THitObject : HitObject
+        where TAction : struct
+    {
+        protected List<THitObject> HitObjects;
+        protected InputInterceptor Interceptor;
+        private readonly List<double> breakEndTimes = new List<double>();
+        private double nextreakTime;
```
bdach, line 33: typo

```suggestion
        private double nextBreakTime;
```

(+ usages)

osu.Game/Rulesets/Mods/ModAlternate.cs
```diff
@@ -0,0 +1,101 @@
...
+            }
+
+            try
+            {
+                nextreakTime = breakEndTimes.First();
+            }
+            catch
+            {
+                checkForBreaks = false;
+            }
```
bdach, line 55: Don't use exceptions for performing logic like this. Aside from this being a code smell, CLR exceptions are slow - compared to a normal branch, at least - so they're a very bad idea here. This is sort of a pythonesque idiom but in C# land this sort of thing is frowned upon.

(Also, `FirstOrDefault()` exists.)

osu.Game/Rulesets/Mods/ModAlternate.cs
```diff
@@ -0,0 +1,101 @@
...
+                ResetActionStates();
+
+                try
+                {
+                    nextreakTime = breakEndTimes.First((stamp) => stamp > time);
+                }
+                catch
+                {
+                    checkForBreaks = false;
+                }
```
bdach, line 77: Same note about exceptions applies here.

Also, it would be nice if this could avoid enumerating an increasing number of elements of `breakEndTimes` every call, but I suppose it shouldn't have too many elements so maybe not really all that relevant here...

bdach (changes requested):

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
LeNitrous: Defined here. It won't play well if ScoreV2 is properly implemented but I'm sure that it'll be a retrievable if it is.
https://github.com/ppy/osu/blob/860f10ceee2d1b18fd1578d2ad190499dc140394/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableHit.cs#L208-L212

LeNitrous:
@Waidxd changes as suggested have been applied. Please check whether this behavior is correct.

Waidxd:
@LeNitrous behaviors works perfectly now however I ran into a separate issue regarding big dons and big kats.

In the following scenario:
small don --- big don --- small don

If I were to do as `PlayStyle.KDDK` or as `PlayStyle.KKDD`
`TaikoAction.LeftCentre` --- `(TaikoAction.LeftCentre + TaikoAction.RightCentre)` --- `TaikoAction.LeftCentre`

I will end up missing the 2nd small don

but in the scenario of:
big don --- big don --- big don

If I do 
`(TaikoAction.LeftCentre + TaikoAction.RightCentre)` -- `(TaikoAction.LeftCentre + TaikoAction.RightCentre)` -- `(TaikoAction.LeftCentre + TaikoAction.RightCentre)`

This seems to work fine... so I am not really too sure what is causing the issue yet or how to resolve it

bdach:
Strong hits should probably reset the last hand used? Not sure what is the expectation here.

LeNitrous:
It's intended whenever the next hit object is a strong, any keys can be used regardless if the last key is the same. This is to not block the mechanic of requiring 2 keys for a full score on strong hit objects.

Waidxd:
so the more common approach among full alternators from kddk players given the following scenario (capital d for large don):
d --- D --- d --- D --- d 
l --- l+r --- l --- l+r --- l
or
r --- l+r --- r --- l+r --- r
essentially we reset after every large don or kat

The following is the less common approach and what current implementation enforces, its not wrong by any means.
d --- D --- d --- D --- d 
l --- l+r --- r --- l+r --- l
or
r --- l+r --- l --- l+r --- r
There is very little to no stamina saved from the second approach I mentioned which is why most full alternators tend to go with
the first approach as it is a lot easier to not mess up

I guess the first approach could later added as an option in the future

LeNitrous:
I forgot to mention that when the player successfully uses two key on a strong hit object, the key states will be reset allowing for any key to be used on the next. However if the player uses only one key, the key states won't be reset and checks will be performed on the next.

@Waidxd for your case, the key states are always reset before the next hit object's judgement **if it is a strong** and should allow both scenarios to be accepted input.

Waidxd:
Can confirm after testing manually that the latest change by @LeNitrous fixed/allowed this approach.
> so the more common approach among full alternators from both kddk and ddkk players given the following scenario (capital d for large don):
> d --- D --- d --- D --- d
> l --- l+r --- l --- l+r --- l
> or
> r --- l+r --- r --- l+r --- r
> essentially we reset after every large note

I tried both play styles again and I can't really find any bugs/flaws with the current implementation after going through a number of different scenarios. I think its good to go!

LeNitrous:
Moved the mod to Difficulty Increase as it plays similar to Perfect and Sudden Death where it also adds a layer of difficulty to the gameplay. Also added a keyboard icon to it.

LeNitrous:
Key checks are more lenient during drum rolls with the added option to ignore the last key used after one and key checks are fully ignored during swells and key states will always reset after one. This should be ready for review at this state.

Layerex:
I think all keypresses before the first note should be ignored, since some players(including me) are used to spamming keys randomly before the map starts; missing first note because of it with 50% chance is very disturbing

LeNitrous:
Had to do a soft reset on this PR as I managed to screw up its commit history. However with the changes done, it should ignore key checks during break times and as well as before the start of gameplay more obvious now as it is in-sync with the in-game UI now.

peppy:
This doesn't work. If you press a key before you hit a circle it counts as a press. Many players will mash keys before starting, and this interaction shouldn't count towards an actual press.

Should probably be binding to on-result-occurred then finding out the currently pressed actions at that point in time? Maybe the way you are storing them is fine but you should only be applying the "press" if a successful hit occurs (for the first object in the map at very least).

LeNitrous:
This assumes `GameplayStartTime` from `Player` but then again it wasn't the wisest decision. I'll narrow the constraints down further to the first and last hit objects of a beatmap.

peppy:
That won't help. You can't base it on time. It needs to be based on result. Alternating should start counting from the first non-miss result in the beatmap IMO.

LeNitrous:
I've based the "first non-miss" hit result by checking the highest combo achieved since that would look more simpler compared to calling `ScoreProcessor.GetStatistic()` multiple times for each non-miss `HitResult` since combo is based on successful hit results.

peppy:
I'm abstaining from further review of this PR until inline comments and a code quality pass is done. Currently it reads worse than osu-stable code, especially `TaikoModAlternate`.

Go through your code, extract methods where possible (and name them logically) and add inline comments wherever the code doesn't immediately explain what is being done.

peppy (changes requested):
code quality needs work

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
```diff
@@ -0,0 +1,36 @@
...
+        {
+            lastActionPressed = null;
+        }
+
+        protected override bool OnPressed(OsuAction action)
+        {
+            if (IsBreakTime.Value)
+                return false;
+
+            if (HighestCombo.Value == 0)
```
peppy, line 23: why?

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
```diff
@@ -0,0 +1,36 @@
...
+                lastActionPressed = action;
+                return false;
+            }
+
+            if (lastActionPressed == action)
+                return true;
+
+            lastActionPressed = action;
+            return false;
+        }
```
peppy, line 34: `return false false true false` in that order. this is some convoluted logic alright.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,135 @@
...
+                ResetActionStates();
+
+            if (hitObject?.IsStrong ?? false)
+                blockInput = handleStrongHit(hitObject, action);
+            else if (hitObject is Swell)
+                blockInput = false;
+            else if (hitObject is IHasDuration)
+                blockInput = action == prevPressedAction;
+            else
+                blockInput = shouldBlock(action);
```
peppy, line 68: `maybe false maybe maybe`

use switch? comment what these are doing?

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,135 @@
...
+        {
+            if (IsBreakTime.Value)
+                return false;
+
+            var hitObject = HitObjects.FirstOrDefault(h =>
+            {
+                var time = Interceptor.Time.Current;
+                var window = (h is DrumRoll) ? ((h.NestedHitObjects.FirstOrDefault() as DrumRollTick)?.HitWindow ?? 0) : h.HitWindows.WindowFor(HitResult.Miss);
+                var endTime = (h as IHasDuration)?.EndTime ?? h.StartTime;
+                return (time > h.StartTime - window) && (time < endTime + window);
```
peppy, line 43: i don't even know where to start here. extract to a method and explain what is going on. and don't use var for primitive types.

osu.Game/Screens/Play/Player.cs
```diff
@@ -92,10 +92,7 @@ public class Player : ScreenWithBeatmapBackground, ISamplePlaybackDisabler
 
         public BreakOverlay BreakOverlay;
 
-        /// <summary>
-        /// Whether the gameplay is currently in a break.
-        /// </summary>
-        public readonly IBindable<bool> IsBreakTime = new BindableBool();
+        public IBindable<bool> IsBreakTime => breakTracker?.IsBreakTime;
```
peppy, line 95: needs to be reverted

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,135 @@
...
+                return shouldBlock(action);
+
+            bool compareUnordered((TaikoAction?, TaikoAction) actual, (TaikoAction, TaikoAction) expected)
+                => (actual.Item1 == expected.Item1 && actual.Item2 == expected.Item2) ||
+                   (actual.Item1 == expected.Item2 && actual.Item2 == expected.Item1);
+        }
+
+        public enum Playstyle
+        {
+            KDDK,
```
peppy, line 131: what do these mean?

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
```diff
@@ -0,0 +1,135 @@
...
+
+                case Playstyle.KKDD:
+                    blockInput = action == prevPressedAction;
+                    break;
+            }
+
+            return blockInput;
+        }
+
+        private bool handleStrongHit(TaikoHitObject hitObject, TaikoAction action)
```
peppy, line 104: what is there to handle?

peppy (changes requested):

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
LeNitrous: Highest combo is based on the number of non-miss hits. It is possible to do `ScoreProcessor.GetStatistic()` multiple times for each hit result type, this is much simpler.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
LeNitrous: I don't have proper names for them since I'm not really a taiko player. Though they're taiko jargons for keybind arrangements. KDDK is Rim-Centre-Centre-Rim and KKDD is Rim-Rim-Centre-Centre (or vice versa).

LeNitrous:
I'm still not confident whether I refactored enough of the code. Feedback is appreciated.

osu.Game.Rulesets.Osu/Mods/OsuModAlternate.cs
bdach: I can't parse this comment. What does this aim to do exactly? Not process inputs until the first user hit? I'd actually think using just `Combo` would be more correct, to allow recovery from any position after a miss.

bdach:
Not good enough still, sorry. The state management is (at least for me) easily the worst part of this PR - the worst parts of inheritance show here. Some parts of the logic are in the base mod, some in the ruleset-specific ones. The taiko logic also needs less intertwining between execution paths (ideally, you'd have a straight switch across hitobject types, and then handle each cleanly in a separate method).

I'll give a proper attempt at taking over this PR tomorrow.

osu.Game.Rulesets.Taiko/Mods/TaikoModAlternate.cs
bdach: FWIW, this is what I have locally and intend to push in here when I'm done:

```csharp
        public enum Playstyle
        {
            /// <summary>
            /// Each hand has a rim and a centre button, so alternating is done by changing hands.
            /// Also known in community vernacular as "kddk".
            /// </summary>
            [Description(@"Alternate hands (""kddk"")")]
            AlternateHands,

            /// <summary>
            /// One hand has both rim buttons and the other both centre buttons.
            /// Alternating is done using fingers only.
            /// Also known in community vernacular as "kkdd".
            /// </summary>
            [Description(@"Alternate fingers (""kkdd"")")]
            AlternateFingers
        }
```

Still a little arcane, but maybe a little better?

bdach:
I've made decent headway into straightening this out but haven't managed to finish yet. [Preview branch here](https://github.com/ppy/osu/compare/master...bdach:mod-alternate-wip). It has some wonkiness still (consecutive strong hits aren't behaving as expected) but I think it looks way better than what's here.

I tried writing tests for this but they're way too involved and verbose even in the simple cases, and even when writing headless (tried simulating the state changes by exposing methods as protected). I'm not even sure the taiko one is the correct direction; it might honestly be easier to take the beatmap, run an algorithm over it to determine actions permitted for each hitobject and then read from that as we go in the mod itself.

peppy:
fwiw, i don't think the user should need to track the "expected" key for a specific object and weakly agree with how it's done at the moment (adapting to the user).

if we're to do it as a preprocess step there 100% needs to be visual feedback on the objects or HUD as to which key is next. as it stands, visual feedback would still be appreciated but isn't imperative.

bdach:
The way I was thinking of wouldn't differ much from what's here in practice - it wouldn't be just "press this one and only expected key for this object", I'd be doing some fudging to make sure the user can pick up alternating again from any given point in time. But hopefully that's not a bridge I'll have to cross if I can get the current version working correctly.

bdach:
I fiddled with my version some more today but I'm still not entirely happy with how it works, so I'm putting this on the backburner due to relatively low impact/priority.

LeNitrous:
> if we're to do it as a preprocess step there 100% needs to be visual feedback on the objects or HUD as to which key is next. as it stands, visual feedback would still be appreciated but isn't imperative.

I've taken the time to review the code back and see what I can do about it and in this scenario. Its true there should be some visual feedback whenever the user presses the wrong key and what better place would it be is on somewhere the player always has their eyes on which is usually either the cursor or the next object.

For standard, we can make it flash red whenever the next input is invalid. For taiko, the area where the judgement line is. I'll have to implement it to see it in practice still.

> I'm not even sure the taiko one is the correct direction; it might honestly be easier to take the beatmap, run an algorithm over it to determine actions permitted for each hit object and then read from that as we go in the mod itself.

While the standard implementation is relatively simple enough given there are only two keys to keep track of and only one type of input expected to be received, there are readability problems with taiko's end given how complex the requirements are. Even I had issues writing tests before. I'm not too sure on how this can be further improved.

Here are my findings about taiko's input:

- The strong hit objects have the hit object itself and one child hit object that its only purpose is to detect whether a two-key press was done within a set amount of time which is usually short. This implementation was also rewritten in the mod as the hit object itself doesn't expose this.
- The rim and drum are simple and handled similarly like standard's however they expect a specific key press.
- Drum rolls (yellow sliders) accept both rim and drum hits but must be hit in time consistently. However this case still needs to be handled similarly to the point above.
- Swells are ignored and can be mashed.

ribbanya:
@bdach 
> I fiddled with my version some more today but I'm still not entirely happy with how it works, so I'm putting this on the backburner due to relatively low impact/priority.

I know this was a while ago so maybe you don't remember but what did you find unsatisfactory about your branch?

As a strictly alternating player who also wants to learn the osu codebase and contribute if I have time, I'd like to take a crack at this.

bdach:
I don't recall anymore, sorry. I think it just didn't work 100% correctly even after my attempted changes.

mk56-spn:
i think imma nuke the taiko part and PR this for STD alone

peppy:
It’s called “osu!” not “STD”.

LeNitrous:
I think taiko's issue was code quality and not to mention tries to accommodate all the nuances of alternating in taiko alone. I don't want it to be wasted effort honestly but if its really the case then I can close this PR and make a new one branching from this for the osu! part to be implemented so the original branch wont get affected.

mk56-spn:
Up to you it's your code ig. Dunno what peppy or bdach will think. Eitherway something seems broken with on pressed and on released in the main file so yeah . Probably caused by updates to things around it

StayBlueee:
Can confirm that this mod basically works outside of the box on osu! (standard) (see below for issue and fix). I have not tested the taiko implementation.

Adding onto the comment posted by @mk56-spn, I simply needed to update `OnPressed` and `OnReleased` to use `KeyBindingPressEvent<TAction>` and `KeyBindingReleaseEvent<TAction>` respectively. The reason being is that `IKeyBindingHandler` was updated back in July. Other than that, seemingly nothing else needs to be touched for the mod to work.

One thing that maybe wasn't actively considered when making this mod is the problem of keys still being registered by the key overlay even when they aren't supposed to be. This is not really an issue, but just thought it was worth mentioning.

LeNitrous:
Superseded by https://github.com/ppy/osu/pull/16701

PR #9611, opened 2020-07-18
Remove SocialOverlay component as never being used

EVAST9919:
Split from https://github.com/ppy/osu/pull/9610


bdach:
what about the classes from the `osu.Game.Overlays.Social` namespace?

bdach (changes requested):
as above

bdach (approved):
seems fine

smoogipoo (approved):

PR #9612, opened 2020-07-19
Input handler updates in line with framework changes

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3736

### DIFF
--- a/osu.Game.Rulesets.Osu/Replays/OsuFramedReplayInputHandler.cs
+++ b/osu.Game.Rulesets.Osu/Replays/OsuFramedReplayInputHandler.cs
@@ -36,19 +36,10 @@ namespace osu.Game.Rulesets.Osu.Replays
             }
         }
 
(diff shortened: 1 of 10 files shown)


osu.Game.Rulesets.Osu/Replays/OsuFramedReplayInputHandler.cs
```diff
@@ -36,19 +36,19 @@ public OsuFramedReplayInputHandler(Replay replay)
...
-                },
+                });
+            input.Add(
                 new ReplayState<OsuAction>
                 {
                     PressedActions = CurrentFrame?.Actions ?? new List<OsuAction>()
-                }
-            };
+                });
+            ;
```
bdach, line 51: codefactor is yelling at this one

```suggestion
```

smoogipoo (approved):

PR #9613, opened 2020-07-19
Reduce memory allocations in MenuCursorContainer

peppy:
Before:

image

After:

image


osu.Game/Graphics/Cursor/MenuCursorContainer.cs
```diff
@@ -55,7 +54,15 @@ protected override void Update()
                 return;
             }
 
-            var newTarget = inputManager.HoveredDrawables.OfType<IProvideCursor>().FirstOrDefault(t => t.ProvidingUserCursor) ?? this;
+            IProvideCursor newTarget = this;
+
+            foreach (var d in inputManager.HoveredDrawables)
+            {
+                if (!(d is IProvideCursor p) || !p.ProvidingUserCursor) continue;
```
bdach, line 61: I'd actually be in favour of inverting this check (either by storing the non-negated value to a variable beforehand:

```csharp
var providingUserCursor = d is IProvideCursor p && p.ProvidingUserCursor;

if (!providingUserCursor)
    continue;
```

or by inverting this condition and `break`ing inside of it). This seems to be _correct_, but I don't like having to compute de Morgan's law in my head with conditions like these.

peppy: Your suggestion actually results in a "not initialized" issue (due to the `foreach` i guess?) so I went with inverting conditional.

smoogipoo (approved):

PR #9614, opened 2020-07-19
Fix hard crash on notifications firing before NotificationOverlay is ready

peppy:
There were several usages of `notifications` before it was constructed (the one I ran into was `LogEntry` - ie. an exception occurred before BDL load).

This would cause a startup crash.

Also fixes a stray access to Toolbar in a similar way (was the actual crash I ran into).

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -63,7 +63,8 @@ namespace osu.Game
 
         private ChannelManager channelManager;
 
-        private NotificationOverlay notifications;
+        [NotNull]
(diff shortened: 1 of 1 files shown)


bdach (approved):
I think this is acceptable :+1: 

There seems to be a theoretical possibility of the notification overlay dropping notifications if they're posted between ctor and `load()`, but that seems edge-case'y enough to disregard. Every usage of `ToolbarOffset` should be reading every frame too (and all the existing ones do so already as far as I can tell), so that is also fine to return 0 sometimes at the start. (Possibly worth commenting to watch out for that? Unsure.)

smoogipoo (approved):

PR #9619, opened 2020-07-19
Implement OsuModNinja

EVAST9919:
Inspired by FruitNinja

Preview: https://www.youtube.com/watch?v=ts_XkVjbJ-Q

### DIFF
--- a/osu.Game.Rulesets.Osu/Mods/OsuModNinja.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModNinja.cs
@@ -9,21 +9,20 @@ using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Osu.Objects;
 using osu.Game.Rulesets.Osu.Objects.Drawables;
+using osu.Game.Rulesets.Osu.UI;
 using osuTK;
 
 namespace osu.Game.Rulesets.Osu.Mods
 {
-    internal class OsuModTransform : Mod, IApplicableToDrawableHitObjects
+    internal class OsuModNinja : Mod, IApplicableToDrawableHitObjects
     {
-        public override string Name => "Transform";
-        public override string Acronym => "TR";
-        public override IconUsage? Icon => FontAwesome.Solid.ArrowsAlt;
+        public override string Name => "Ninja";
+        public override string Acronym => "NJ";
+        public override IconUsage? Icon => FontAwesome.Solid.UserNinja;
         public override ModType Type => ModType.Fun;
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Osu/Mods/OsuModNinja.cs
```diff
@@ -0,0 +1,63 @@
...
+
+namespace osu.Game.Rulesets.Osu.Mods
+{
+    internal class OsuModNinja : Mod, IApplicableToDrawableHitObjects
+    {
+        public override string Name => "Ninja";
+        public override string Acronym => "NJ";
+        public override IconUsage? Icon => FontAwesome.Solid.UserNinja;
+        public override ModType Type => ModType.Fun;
+        public override string Description => "Slice the circles!";
```
bdach, line 23: this name and description seem misleading. when I think of what is the core mechanic in fruit ninja, I think of slicing. this mod doesn't *actually* add slicing; from what I can see it solely has the hit objects emulate the way fruits fly into the playfield. (they don't fly back out of the screen either, just hang in their final position.)

(also not sure of potential legal repercussions of naming this after that game.)

osu.Game.Rulesets.Osu/Mods/OsuModNinja.cs
```diff
@@ -0,0 +1,63 @@
...
+                case DrawableSliderRepeat _:
+                    return;
+
+                default:
+                    var hitObject = (OsuHitObject)drawable.HitObject;
+
+                    var originalPosition = drawable.Position;
+                    var appearPosition = new Vector2(originalPosition.X, OsuPlayfield.BASE_SIZE.Y);
+
+                    // the - 1 and + 1 prevents the hit objects to appear in the wrong position.
```
bdach, line 49: mind explaining this?

bdach:
a proper implementation mimicking mechanics of fruit ninja could actually be a decent idea for its own ruleset, but I think it wouldn't be suited to be a mod due to implementation complexity.

osu.Game.Rulesets.Osu/Mods/OsuModNinja.cs
EVAST9919: Same as for Transform mod

EVAST9919:
I just thought it'd be nice to have (more like another variation of Transform mod)

EVAST9919:
> a proper implementation mimicking mechanics of fruit ninja could actually be a decent idea for its own ruleset, but I think it wouldn't be suited to be a mod due to implementation complexity.

sounds good. It's easier to make a new ruleset with this mechanics in mind rather then trying to implement something very complex

PR #9622, opened 2020-07-20
Disallow touch device mod selection from leaderboard scores

Joehuu:


osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -376,7 +377,9 @@ public MenuItem[] ContextMenuItems
             {
                 List<MenuItem> items = new List<MenuItem>();
 
-                if (score.Mods.Length > 0 && modsContainer.Any(s => s.IsHovered) && songSelect != null)
+                var scoreMods = score.Mods;
+
+                if (scoreMods.Length > 0 && modsContainer.Any(s => s.IsHovered) && songSelect != null && scoreMods.All(m => m.Type != ModType.System))
```
peppy, line 382: could this not be simplified to `scoreMods.Any(m => m.Type != ModType.System)` and handle the length check at the same time?

Joehuu: Length check was already redundant as `modsContainer` would be 0x0 size with no mods and will not be hoverable. Will remove.

`scoreMods.Any(m => m.Type != ModType.System)` returns true if there are other non system mods with touch device (i.e. can select HDTD).

Should the behavior be to hide the menu on any touch device mod combination (my implementation), or show and ignore the touch device mod and select the other mods and hide when touch device only?

peppy: Hiding the menu just because touch is in there seems wrong. System mods should be automatically excluded from any user selection (probably at a different level to this local code).

Joehuu: Will try to fix this tomorrow, but after investigation, the touch device mod gets selected after other mods are selected (i.e. you can press "use these mods" on a HDTD score and it'll select HD, and then pressing "use these mods" again will add TD to HD). And selecting a mod from the mod select overlay will deselect the touch device mod.

bdach:
What's happening with this PR?

Joehuu:
Can't seem to find the root cause of this behavior, will make an issue instead. Closing.

PR #9626, opened 2020-07-20
Marginally improve spinner visuals

peppy:
Note that this isn't supposed to be a final solution, just an iterative improvement aimed at making spinners feel less ugly and a bit more reactive.

- Added a constant passive spin
- Reduced "ticks" count but made more visible
- Removed fill background
- Ensure it's spinning when disappearing on a successful hit
- Fixed colour fading back to blue on successful hit
- Improved animation code quality slightly

Next step is to add skinning, which may change the structure further.

Against light:

Against dark:


smoogipoo (approved):
Looks fine I think.

Issue #2854, opened 2018-06-21
Overlapping timing windows in osu!mania


voxlz:
### Hello, long time mania player here. 

An issue a lot of more experienced players run into is overlapping timing windows, were simply not playing a single note could make you lose a map. This issue arises when a lot of notes are very close to each other.

![overlapping timingwindows](https://user-images.githubusercontent.com/19476494/41750563-0bc64756-75bd-11e8-85b4-8b384c3bbeb3.png)

Sorry for the ugly picture. I have tried to "show" the score you get depending on when you hit the note. The colors match the color of the score you would get. Orange is rainbow 300. 

Now please take a look at this. Let's say you play a hard map and get distracted and don't play note 1 in the picture. Usually, the timing window disappears after pressing the note. But notice that if I don't press the note, the timing window will extend into the next note. If the speed is high enough, it will end up like in the picture. This means that if you miss note one and hit note 2 and 3 perfectly, you will still get 3 misses. (Notice how the orange is almost completely gone and replaced with red now) If this is a trill or extended stream this means you will die within a second, if you don't purposely skip a note. This is extremely hard and unrealistic though. 

I think this is an issue worth looking at. There is no worse feeling then dying because if this when you know you would survive otherwise.

**Updated**
I'm proposing that in the situation of two very close notes and overlapping timingwindows, as soon as the hit window of the second note reaches the first 300 (non-rainbow) point window, the first note should be regarded as a miss. This would make the game feel better and more accurate to the rhythm.

Adrriii:
What would be a good threshold for deciding which note will be counted as played when you press your key in an interval ?

Also, having this red part doesn't seem realistic to me. I would agree to it being removed !

Tripp1n:
It feels like there is a red part. Not sure if there actually is. Threshold needs to be pretty high, as to not acidentally giving more points. But i think if you hit a note perfectly you should never get a miss.

Adrriii:
I would be in favor of the "combo breaking" one tho, as in a rhythmic point of view, it makes more sense for a band to skip one beat and continue playing, than trying to make up for their lateness (speeding up).

Code wise, I don't know how it is handled and maybe the notes are stored in a heap (for each column in mania), if so then it would be better having a simple array (hoping it doesn't cause performance issues).

LastExceed:
[...]

PR #9627, opened 2020-07-20
Fix hold notes accepting presses during release lenience

### LINKED ISSUE #9608: Missing a hold note release causes random misses in inverse maps
When missing a hold note release, specifically on inverse maps or maps with a lot of LNs, there will be random misses throughout the rest of the map. I think it misses every perfect release in the specific column you missed on before, until you get a great, good, etc. (not a perfect/marvelous) on a release in that same column. 
It's hard to explain, please watch this video where I intentionally miss the first LN if it isn't clear: https://youtu.be/he3tHxYS7eU

map: https://osu.ppy.sh/beatmapsets/502203#mania/1111137
version: 2020.717.0
no logs

smoogipoo:
Mostly resolves https://github.com/ppy/osu/issues/9608

Keep in mind that note lock is still a thing, which is discussed in this issue: https://github.com/ppy/osu/issues/2854. This means that two notes with overlapping windows can encounter the same issue as above (and probably still do in that example).
This isn't meant to address that - it's a bit of a bigger issue.


peppy (approved):
looks good

PR #9633, opened 2020-07-21
Fix keybind clear button always clearing first keybind regardless of target

Joehuu:
Before:
fHD4nyQ6JP

After:
UqVJsVQqXD


peppy:
- This needs test coverage.
- I'm not sure how this fix works (or if it's the correct approach). Would be very helpful to add xmldoc to the `updateBindTarget` and the reasoning for the conditional added as a local comment.

peppy (changes requested):
as proposed

bdach:
What's the status of this PR?

Joehuu:
Was thinking if I should write the code here as I'm not sure of the changes, but I pushed commits.

Test is probably wrong because I had to expose `KeyBindingRow`'s `FillFlowContainer<KeyButton> buttons` which in turn had to expose `KeyBindingRow.KeyButton` to public. Not sure how else to check if the keybinding is cleared. And the test scene right now is using the game's keybindings instead of a dummy, but that's another issue.

smoogipoo:
I've re-privatised the buttons. Seems ok.

smoogipoo (approved):

peppy (approved):

PR #9635, opened 2020-07-21
Make perfect mod ignore all non-combo-affecting hitobjects

peppy:
As per discord discussion, this seems to make more sense from a consistency perspective.

Community response seems roughly positive ([70-30](https://twitter.com/ppy/status/1285478489848602624) but no real feedback to advocate reason for not making the change). Probably don't merge until we get some more feedback, just throwing this up for review/dependency reasons.


bdach (approved):
I also think this makes sense :+1:

smoogipoo (approved):

PR #9636, opened 2020-07-21
Add epilepsy warning

### LINKED ISSUE #3114: missing epilepsy warning
epilepsy warning isn't implemented.

gagahpangeran:
Closed #3114

Need feedback because I know this is not the best of my code.


osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
```diff
@@ -49,6 +49,44 @@ public MetadataLine(string left, string right)
...
         }
 
+        private class EpilepsyWarning : FillFlowContainer
+        {
+            public EpilepsyWarning(bool warning)
+            {
+                if (warning)
+                    this.Show();
+                else
+                    this.Hide();
```
bdach, line 59: I don't think it makes much sense for this component to hide itself like this. I'd say it should be hidden (or just not created at all) at a higher level, in `BeatmapMetadataDisplay` itself.

osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
```diff
@@ -49,6 +49,44 @@ public MetadataLine(string left, string right)
...
+                Direction = FillDirection.Vertical;
+                Children = new Drawable[]
+                {
+                    new SpriteIcon
+                    {
+                        Anchor = Anchor.TopCentre,
+                        Origin = Anchor.TopCentre,
+                        Icon = FontAwesome.Solid.ExclamationTriangle,
+                        Size = new Vector2(40),
+                    },
```
bdach, line 71: UX-wise this warning doesn't really... warn that much. It blends in too much with the rest of the metadata.

I'd say it needs to be jump out more; the simple way would be applying some kind of colour to the icon, for instance:

2020-07-21-234918_619x134_scrot

Not sure whether any further animations should be added (thought about flashing the icon from white to yellow for a couple of seconds, but on second thought flashing an icon in an *epilepsy warning* seems... not wise...)

osu.Game.Tests/Visual/Gameplay/TestScenePlayerLoader.cs
```diff
@@ -240,6 +243,15 @@ private void addVolumeSteps(string volumeName, Action beforeLoad, Action afterLo
             AddUntilStep("wait for player load", () => player.IsLoaded);
         }
 
+        [TestCase(true)]
+        [TestCase(false)]
+        public void TestEpilepsyWarning(bool warning)
```
bdach, line 248: This test should be checking whether the epilepsy warning is visible (by finding it via `this.ChildrenOfType` and inspecting its `Alpha`).

Also not sure whether this is the best place for it. Thus far `BeatmapMetadataDisplay` didn't really have actual logic so a test scene was maybe a bit extravagant, but in this case it would be certainly much easier to test in a new separate scene.

bdach (dismissed):
In addition to the comments below, please see [CI error messages](https://ci.appveyor.com/project/peppy/osu/builds/34214237/messages) and fix the code style & static analysis issues listed therein. I would also recommend either running `InspectCode.ps1` (if on Windows) or setting up appveyor on your fork to be able to catch these kinds of issues before pushing.

osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
frenzibyte: for the UX, In my opinion, the epilepsy warning should not be embedded with the beatmap metadata display, it should rather be displayed on the center of the screen, the track volume should get lowered and background gets dimmed, all to get proper and serious attention from the user; I've checked out the branch and made some changes to get to the idea above, here is what I got: https://streamable.com/uevzep

frenzibyte (changes requested):

osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
bdach: Looks ok but the transition away needs to be longer. I could barely read the text in time on that video.

bdach:
@gagahpangeran Are you planning to address the above comments?

gagahpangeran:
> @gagahpangeran Are you planning to address the above comments?

Yeah, I'll check it tonight or tomorrow. My dotnet dependency is kinda broken and there's conflict for past week, need reinstall.

bdach:
I've taken the liberty of asking @frenzibyte for his variant, cherrypicking it here and then adding some of my tweaks onto it. Unfortunately to work properly the flag has to be persisted to database, as `WorkingBeatmap.loadBeatmapAsync()` can discard the decoded values:

https://github.com/ppy/osu/blob/b9c1d21743087e253321debef5018bfb30344088/osu.Game/Beatmaps/WorkingBeatmap.cs#L220-L221

Not sure if you'll agree on all the decisions made here (particular areas of note are the volume adjustment handling and the transition chaining), but I figured I'd rather try to get this closer to merge than not.

osu.Game/Screens/Play/EpilepsyWarning.cs
```diff
@@ -0,0 +1,107 @@
...
+            };
+
+            track = beatmap.Value.Track;
+            track.AddAdjustment(AdjustableProperty.Volume, trackVolumeOnEpilepsyWarning);
+        }
+
+        protected override void PopIn()
+        {
+            this.FadeIn(500, Easing.InQuint)
+                .TransformBindableTo(trackVolumeOnEpilepsyWarning, 0.25, 500, Easing.InQuint);
```
peppy, line 92: what's the reasoning for this? just to "look/sound cool"? just want to make sure i'm not missing something

osu.Game/Screens/Play/EpilepsyWarning.cs
```diff
@@ -0,0 +1,107 @@
...
+            RelativeSizeAxes = Axes.Both;
+            Alpha = 0f;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(OsuColour colours, IBindable<WorkingBeatmap> beatmap)
+        {
+            Children = new Drawable[]
+            {
+                new Box
```
peppy, line 39: not sure we want to do this. adding +1 fill rate while displayed.

osu.Game/Screens/Play/EpilepsyWarning.cs
frenzibyte: more like for getting serious attention from the user, but yeah nothing else than that.

bdach: I liked it as it clearly diverted user attention, but I can't say I'm too attached to it myself. Same thing with the dim, really.

osu.Game/Screens/Play/EpilepsyWarning.cs
frenzibyte: hmm, the screen background could expose a bindable to dim it or otherwise. but could also remove the dim effect and call it a day.

peppy:
Which beatmap are you testing this with?

bdach:
I was personally checking on this map: https://osu.ppy.sh/beatmapsets/11788#osu/45308

All diffs should have the warning I believe.

peppy:
Made some further changes to avoid overdraw layer and transition smoother.

Removed the volume restoration because it felt bad.

peppy (dismissed):

smoogipoo:
I'm not entirely sure how I repro'd this, and I don't _think_ it's related to this branch, but:

```
System.InvalidOperationException: A screen should not be loaded before being pushed.
   at osu.Framework.Screens.ScreenStack.Push(IScreen source, IScreen newScreen)
   at osu.Framework.Screens.ScreenExtensions.Push(IScreen screen, IScreen newScreen)
   at osu.Game.Screens.Play.PlayerLoader.<pushWhenLoaded>b__60_1() in /home/smgi/Repos/osu/osu.Game/Screens/Play/PlayerLoader.cs:line 359
   at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal()
   at osu.Framework.Threading.Scheduler.Update()
   at osu.Framework.Graphics.Drawable.UpdateSubTree()
```

Aside from that, is there anything that still needs to happen here @peppy or is this good to go?

peppy:
This is good to go

smoogipoo (approved):
I'm going to ignore that error as there's no way it's related to this PR. I can't trigger it a second time.

It's also a really freaky error. Not sure how it's possible after a rough look at the code.

PR #9637, opened 2020-07-21
Add dynamic computation of beatmap difficulty

smoogipoo:
This introduces the new `BeatmapDifficultyManager` class. I imagine this to take on the responsibility of databased background calculation in the future.

You can retrieve the beatmap difficulty in four ways via `BeatmapDifficultyManager`:
```csharp
// Via an "untracked" bindable. Does not follow the user's current ruleset/mods.
IBindable<StarDifficulty> GetBindableDifficulty(BeatmapInfo, RulesetInfo, IEnumerable<Mod>);

// Via a "tracked" bindable. Follows the user's current ruleset/mods.
IBindable<StarDifficulty> GetBindableDifficulty(BeatmapInfo);

// Asynchronously.
async Task<StarDifficulty> GetDifficultyAsync(...);

// Synchronously. May be used for quick prototyping/testing.
StarDifficulty GetDifficulty(...);
```

Tracking of the bindable versions stops either when the given `CancellationToken` is cancelled, or when references are lost. I've chosen to explicitly cancel the token in `SongSelect` and `AdvancedStats` even though it's not required, just so that calculation doesn't continue after exiting to main menu with references still alive.

The async version can also be cancelled.

All of these will, in some way, give a `StarDifficulty` object. I imagine this object to contain more than just the star rating in the future - for example the current `BeatmapInfo.DifficultyRating` could maybe be moved in there, along with the attributes provided by the difficulty calculator (to display graphs in advanced stats, etc).

As for quirky implementation details, I've chosen to permanently keep the difficulties in the cache. I originally had it clear after 60s (although that didn't do much since it's a dictionary), however upon profiling I've found that 230 difficulty values take up ~13KB space anyway, which I'd say is insignificant for the time being.
Screenshot_2020-07-21_23-50-18


osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmap.cs
```diff
@@ -181,6 +188,17 @@ protected override void ApplyState()
             if (Item.State.Value != CarouselItemState.Collapsed && Alpha == 0)
                 starCounter.ReplayAnimation();
 
+            if (Item.State.Value == CarouselItemState.Collapsed)
+                starDifficultyCancellationSource?.Cancel();
```
peppy, line 192: move outside of the if/else

peppy:
Just to confirm, this is only showing in the expanded panels' star displays? So the tooltips are using previous databased values for the time being intentionally, correct?

smoogipoo:
Difficulty icons are interesting since they:
1. Specify a beatmap + ruleset so it wouldn't make sense to follow the currently-selected ones - it probably needs a special ctor for that.
2. Appear for when the beatmap isn't local, but then should update when the beatmap does become locally-available.

I have a WIP branch for it, but I'd like to do that in a separate PR.

peppy:
Rather than tracked/untracked, how about making it a bit more explicit?

```
CreateBindableForCurrent()
CreateBindableFor(BeatmapInfo...)
```

Get or Fetch can be used in place of Create if you prefer.

smoogipoo:
Are you saying to use the current beatmap rather than to pass in a beatmap?

peppy:
by "current" i meant current ruleset/mods

or maybe even combine the methods, and mention that if ruleset / mods is null, the game-wide settings are used and tracked?

peppy:
I haven't gone through this PR completely but "why would i want even want a bindable if it's untracked?" is a question I still have to figure out. Would be good if the naming can convey the meaning a bit better.

Also since there's `GetDifficulty` method, how about we call the bindable ones `GetBindableDifficulty`?

smoogipoo:
> or maybe even combine the methods

I think I'd be fine with having to specify the ruleset and mods every time.

> why would i want even want a bindable if it's untracked

Three reasons off the top of my head:
1. You don't receive the difficulty immediately since it's asynchronous. Handling the async operation is quite difficult - start task, schedule continuation, check cancellation token inside scheduler callback (basically everything inside `updateBindable` right now).
2. Eventually the untracked bindable will respond to changes in the beatmap - e.g. if the beatmap was imported or updated. I believe this to be expected in 100% of cases.
3. Maybe you have a component that can house both tracked and untracked bindables. I believe `DifficultyIcon` is going to do this, where in the carousel it'll use tracked bindables and in, say, multiplayer playlist panels it'll use untracked bindables.

I think a better question is when someone would use the async version. Perhaps that could be an internal usage - for example maybe the async version should be invoked on import.

And yeah, `GetDifficulty` is a thing, which as I said is mostly for prototyping, where you don't care about tracking/untracking, asynchronous callbacks, or cancellations, and you just want the difficulty. I don't expect that method to be used in production code.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <param name="rulesetInfo">The <see cref="RulesetInfo"/> to compute the difficulty with.</param>
+        /// <returns>The <see cref="StarDifficulty"/>.</returns>
+        private StarDifficulty computeDifficulty(in DifficultyCacheLookup key, BeatmapInfo beatmapInfo, RulesetInfo rulesetInfo)
+        {
+            // In the case that the user hasn't given us a ruleset, use the beatmap's default ruleset.
+            rulesetInfo ??= beatmapInfo.Ruleset;
+
+            try
+            {
+                var ruleset = rulesetInfo.CreateInstance();
```
bdach, line 184: Is there a reason why `RulesetStore` isn't used? I considered thread safety for a second but I would hope `CreateDifficultyCalculator()` implementations should always be thread-safe and pure, and we could save on some allocations here using the store.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// <summary>
+        /// Attempts to retrieve an existing difficulty for the combination.
+        /// </summary>
+        /// <param name="beatmapInfo">The <see cref="BeatmapInfo"/>.</param>
+        /// <param name="rulesetInfo">The <see cref="RulesetInfo"/>.</param>
+        /// <param name="mods">The <see cref="Mod"/>s.</param>
+        /// <param name="existingDifficulty">The existing difficulty value, if present.</param>
+        /// <param name="key">The <see cref="DifficultyCacheLookup"/> key that was used to perform this lookup. This can be further used to query <see cref="computeDifficulty"/>.</param>
+        /// <returns>Whether an existing difficulty was found.</returns>
+        private bool tryGetGetExisting(BeatmapInfo beatmapInfo, RulesetInfo rulesetInfo, IReadOnlyList<Mod> mods, out StarDifficulty existingDifficulty, out DifficultyCacheLookup key)
```
bdach, line 207:
```suggestion
        private bool tryGetExisting(BeatmapInfo beatmapInfo, RulesetInfo rulesetInfo, IReadOnlyList<Mod> mods, out StarDifficulty existingDifficulty, out DifficultyCacheLookup key)
```

(+ usages)

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -0,0 +1,281 @@
...
+        {
+            // In the case that the user hasn't given us a ruleset, use the beatmap's default ruleset.
+            rulesetInfo ??= beatmapInfo.Ruleset;
+
+            // Difficulty can only be computed if the beatmap and ruleset are locally available.
+            if (beatmapInfo.ID == 0 || rulesetInfo.ID == null)
+            {
+                // If not, fall back to the existing star difficulty (e.g. from an online source).
+                existingDifficulty = new StarDifficulty(beatmapInfo.StarDifficulty);
+                key = default;
```
bdach, line 217: I don't think I can fabricate a scenario to confirm my suspicions for sure right now, but this seems unsafe. If I'm compiling in my head correctly, `default` is `{BeatmapId = 0, RulesetId = 0, Mods = null}`; callers of this method use the value of `key` without checks for `default`, so there's a potential for garbage stores to the cache - which in itself wouldn't be a problem, but the garbage reads that would be possible later on (if I'm not misreading) due to reusing that same `default` key is.

The returned `key` should be probably nullable or something.

bdach (changes requested):
Some notes from reading source. (Mostly nits.)

Not sure whether I have something to say right now when it comes to the discussion on whether tracked/untracked/async makes sense. I am of the opinion that both untracked and async have potential for being used but I'm bad at future-proofing.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -0,0 +1,281 @@
...
+        /// </summary>
+        /// <remarks>
+        /// This <see cref="Bindable{StarDifficulty}"/> will not update to follow the currently-selected ruleset and mods.
+        /// </remarks>
+        /// <param name="beatmapInfo">The <see cref="BeatmapInfo"/> to get the difficulty of.</param>
+        /// <param name="rulesetInfo">The <see cref="RulesetInfo"/> to get the difficulty with.</param>
+        /// <param name="mods">The <see cref="Mod"/>s to get the difficulty with.</param>
+        /// <param name="cancellationToken">An optional <see cref="CancellationToken"/> which stops updating the star difficulty for the given <see cref="BeatmapInfo"/>.</param>
+        /// <returns>An <see cref="IBindable{StarDifficulty}"/> that is updated to contain the star difficulty when it becomes available.</returns>
+        public IBindable<StarDifficulty> GetUntrackedBindable([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo rulesetInfo = null, [CanBeNull] IReadOnlyList<Mod> mods = null,
```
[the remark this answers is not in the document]
smoogipoo: It was mostly for consistency but sure, I think that's fine.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
smoogipoo: `RulesetStore` doesn't provide us `Ruleset`'s, unless I'm missing something.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
smoogipoo: This method is private, so I'm not concerned about callers potentially doing something like:
```
tryGetExisting(... out key)

// Assume the key hasn't existed (ignore return value)
cache[key] = ...
```

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
bdach: Oh, duh. Fair.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
bdach: Hmm, I suppose. I still would have probably restructured this so that this mistake is impossible to make type-wise but I guess as long is it private...

smoogipoo:
@peppy I've changed to `GetBindableDifficulty`, see if it fits nicer for you.

osu.Game/Screens/Select/Details/AdvancedStats.cs
```diff
@@ -132,11 +141,39 @@ private void updateStatistics()
...
+        private CancellationTokenSource starDifficultyCancellationSource;
+
+        private void updateStarDifficulty()
+        {
+            starDifficultyCancellationSource?.Cancel();
+
+            if (Beatmap == null)
+                return;
+
+            var ourSource = starDifficultyCancellationSource = new CancellationTokenSource();
```
peppy, line 162: Reason for `ourSource` local? Doesn't seem like it could be thread-unsafe to just use the field.

smoogipoo: Old code, whoops.

peppy (approved):

PR #9638, opened 2020-07-21
Make DateTooltip use OverlayColourProvider colours

EVAST9919:
Note: sometimes colours within different overlays not being updated due to https://github.com/ppy/osu-framework/issues/3231

### DIFF
--- a/osu.Game/Graphics/DateTooltip.cs
+++ b/osu.Game/Graphics/DateTooltip.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Cursor;
 using osu.Framework.Graphics.Shapes;
 using osu.Game.Graphics.Sprites;
+using osu.Game.Overlays;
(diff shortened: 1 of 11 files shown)


bdach:
I don't think it makes sense to merge this until the linked issue is resolved.

EVAST9919:
Sure

peppy:
Is there a reason the colourProvider is passed down rather than DI resolved from within the `DrawableDate`?

EVAST9919:
> Is there a reason the colourProvider is passed down rather than DI resolved from within the `DrawableDate`?

[Smoogipoo's comment](https://github.com/ppy/osu-framework/issues/3230#issuecomment-581236135)

peppy:
Right, which will probably be resolved when the underlying issue is resolved (which is blocking this).

EVAST9919:
Gonna close this for now then, since there are 2 unresolved issues required for proper implementation.

PR #9639, opened 2020-07-21
Make content side padding adjustable for OverlayHeader

EVAST9919:
Split from https://github.com/ppy/osu/pull/9598

### DIFF
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneOverlayHeader.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneOverlayHeader.cs
@@ -36,11 +36,11 @@ namespace osu.Game.Tests.Visual.UserInterface
                 }
             });
 
-            addHeader("Orange OverlayHeader (no background)", new TestNoBackgroundHeader(), OverlayColourScheme.Orange);
-            addHeader("Blue OverlayHeader", new TestNoControlHeader(), OverlayColourScheme.Blue);
+            addHeader("Orange OverlayHeader (no background, 100 padding)", new TestNoBackgroundHeader(), OverlayColourScheme.Orange);
+            addHeader("Blue OverlayHeader (default 70 padding)", new TestNoControlHeader(), OverlayColourScheme.Blue);
             addHeader("Green TabControlOverlayHeader (string) with ruleset selector", new TestStringTabControlHeader(), OverlayColourScheme.Green);
-            addHeader("Pink TabControlOverlayHeader (enum)", new TestEnumTabControlHeader(), OverlayColourScheme.Pink);
-            addHeader("Red BreadcrumbControlOverlayHeader (no background)", new TestBreadcrumbControlHeader(), OverlayColourScheme.Red);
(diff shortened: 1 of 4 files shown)


osu.Game/Overlays/Profile/ProfileHeader.cs
```diff
@@ -23,6 +23,8 @@ public class ProfileHeader : TabControlOverlayHeader<string>
 
         public ProfileHeader()
         {
+            ContentSidePadding = 70;
```
bdach, line 26: why does this not use `UserProfileOverlay.CONTENT_X_MARGIN`?

osu.Game.Tests/Visual/UserInterface/TestSceneOverlayHeader.cs
```diff
@@ -36,11 +36,11 @@ public TestSceneOverlayHeader()
                 }
             });
 
-            addHeader("Orange OverlayHeader (no background)", new TestNoBackgroundHeader(), OverlayColourScheme.Orange);
-            addHeader("Blue OverlayHeader", new TestNoControlHeader(), OverlayColourScheme.Blue);
+            addHeader("Orange OverlayHeader (no background, 100 padding)", new TestNoBackgroundHeader(), OverlayColourScheme.Orange);
+            addHeader("Blue OverlayHeader (default 70 padding)", new TestNoControlHeader(), OverlayColourScheme.Blue);
```
bdach, line 40: this is actually incorrect. the default is 50, as far as I can tell, and the draw visualiser seems to indicate the same.

bdach (changes requested):

osu.Game.Tests/Visual/UserInterface/TestSceneOverlayHeader.cs
EVAST9919: oops, forgot to rename this after latest commit

osu.Game/Overlays/TabControlOverlayHeader.cs
```diff
@@ -30,6 +31,16 @@ public Bindable<T> Current
             set => current.Current = value;
         }
 
+        protected new float ContentSidePadding
```
bdach, line 34: this is perhaps a little strange but we have so many conventions for these types of things (get-only expression-bodied member/bindable/property with setter) at this point that I guess it can stay.

bdach (approved):
seems fine

smoogipoo:
I'm not sure I agree with this conceptually, and think we should be moving towards more uniformity in overlays rather than following in osu-web's footsteps. But ok for now - should be simple to remove at any point if we want.

smoogipoo (approved):

bdach:
I was going to say the same thing but I'm not sure whether this difference was intended or not.

Looking at the exception which is the profile overlay I'd probably say yes (more space is required for the collapse/expand button for the play time statistics portion to fit). I'd ask up front but so far every attempt of me mentioning the (seemingly) right person to ask about design things has resulted in radio silence in response, so...

PR #9641, opened 2020-07-21
Fix SpotlightSelector is a VisibilityContainer without a reason

EVAST9919:
Split from https://github.com/ppy/osu/pull/9598


bdach (approved):
looks fine by me. this removes the 300ms fade-in on load but seeing that there are more changes queued in this area I'm hoping it will return.

PR #9642, opened 2020-07-22
Add user cover background to expanded version of score panel

peppy:
I was hoping we could share it between contracted and expanded, but due to the intricacies of the design (the cornered border at the bottom of the contracted version) this isn't really possible. Have left it in the main background portion because that seems to work best for now.


smoogipoo (approved):
Lgtm

PR #9643, opened 2020-07-22
Add more display modes for HUD overlay (hide during gameplay / hide during break)

peppy:
Note that we will likely add a lot more customisation as to which elements are hidden during break time; this is just bringing things in line with stable.

Hiding during breaks is ignored for replays as it is quite common to want to interact with the HUD.

Not migrating existing setting because it's a quick toggle (shift-tab). Also not doing tri-way toggle because if someone has chosen to hide the HUD, they are likely going to be expecting only the two-way switch.


smoogipoo:
This feels very weird to me. I know "we'll likely add more customisation in the future" but I don't see that happening for several months if not a year from now as we haven't even discussed how that will happen and if it depends on having the markup language.

These choices in particular make no sense to me:
- The song progress is hidden during the break and not during gameplay.
- The score is hidden.
- The timing visualiser is hidden.

peppy:
The only differing parts here are progress and score (by timing visualiser, I assume you mean judgement visualiser, which is hidden on stable too).

Not an easy one to make happen without hard-coding rules for each element, or implementing a proper configuration system. Maybe I'll just do that..

Because we have the break score display, I feel like it's okay to hide the HUD-level one but I guess that's subjective.

Would you be more okay with this change if the break info display included the missing information (song progress and score)?

smoogipoo:
> I assume you mean judgement visualiser, which is hidden on stable too

It isn't hidden in stable. This thing:
image

> Would you be more okay with this change if the break info display included the missing information (song progress and score)?

Yeah, I think so. As for song progress maybe it should be in its minimal display during gameplay and expanded during break? That would be optimal I think, but if it was always visible then that'd be fine too.

peppy:
Weird, the judgement display fades out for me.

I'll re-think this, may have to split the HUD up into two layers for now if we want to do a quick-fix.

peppy:
@smoogipoo would you be okay with this being added as-is with:

- Default to always visible
- Adding a new setting "DuringBreaks" to only show during breaks (and hide during gameplay)

Feels better than what we currently have, which is a toggle for always or never.

smoogipoo:
I'm confused, how is that different to what's already in this PR?

peppy:
The default behaviour will not change.

smoogipoo:
As in the issue with this PR from what I recalls is that it's hiding too much during breaks, right?

I don't know, make the change and I'll see how it feels - it's hard to say otherwise...

peppy:
My point is that it is hiding the hud. It *should* be hiding "too much". More individual settings may come later, but I'd rather get this in as an incremental improvement (for people that aren't looking to have individual control).

smoogipoo:
So let me just clarify. You want to change the "visible only during gameplay" part of this PR to "visible only during breaks", right?

I think that's fine. The problem I had with this PR is that it was hiding it _during_ breaks.

peppy:
Default: always (same as `master`)

Options:

During gameplay (show HUD only during gameplay; hide during breaks)
During breaks (show HUD only during breaks; hide during gameplay)

smoogipoo:
Sounds good

peppy:
Updated as proposed.

smoogipoo (approved):
Seems fine.

PR #9646, opened 2020-07-22
Add slider "sliding" sample support

### LINKED ISSUE #8790: Slider slide legacy hitsound support
**Describe the new feature:** 
When trying out a couple maps which utilise the sliderslide legacy hitsound in standard.
(e.g. [Fort's Ooi](https://osu.ppy.sh/beatmapsets/355573#osu/784791), and a [taikosu! map](https://osu.ppy.sh/beatmapsets/25836#osu/87484). The slider slide aren't utilised in the body of the sliders.

Allowing `soft/normal/drum-sliderslide` from legacy hitsounds into slider bodies. This of course would work the same way as in stable.

peppy:
- Closes #8790
- [x] Depends on https://github.com/ppy/osu/pull/9645

Please test on https://osu.ppy.sh/beatmapsets/861#osu/5680

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -94,6 +94,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             base.LoadSamples();
 
             slidingSample?.Expire();
+            slidingSample = null;
 
             var firstSample = HitObject.Samples.FirstOrDefault();
 
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Linq;
 using osuTK;
 using osu.Framework.Graphics;
 using osu.Game.Rulesets.Objects.Drawables;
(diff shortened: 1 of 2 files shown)


bdach:
From looking at the map linked `sliderslide` support seems to be working but that map also uses `sliderwhistle` which is not supported, so no full parity yet it seems. Wiki also tells me `slidertick` exists.

peppy:
Hmmm well spotted..... adding the rest is probably best for a follow-up PR. I need to encapsulate this logic better, I think (sliderslide / sliderwhistle / spinnerspin are all going to be doing very similar things, so may be better to move to a dedicated drawable component)

Do we not already support slidertick, though?

bdach:
Good point. Admittedly I haven't checked `slidertick`, just brought it up as it was in the same section as the others on the wiki. It does look to have support.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
```diff
@@ -81,6 +83,41 @@ private void load()
...
+            Tracking.BindValueChanged(updateSlidingSample);
+        }
+
+        private SkinnableSound slidingSample;
+
+        protected override void LoadSamples()
+        {
+            base.LoadSamples();
+
+            slidingSample?.Expire();
```
smoogipoo, line 96: Is this a performance optimisation to stop updates?

peppy: Check where this is called (generally on sample change, editor for now). A new instance is being constructed just below. We need to kill the last one.

smoogipoo: Aaaaaah I misread the method as `LoadComplete`

smoogipoo (commented):
I think there's more than I can comment on here like:
- Sample banks seem to only be used if beatmap version >= 14, which means this code should probably exist inside the beatmap converter.
- There's a lot of funny business around how the control point affects the sample in stable (I'm not actually 100% sure if it does in this case).

Both of these, I think, are issues that can be touched upon later if they're deemed necessary. Always having the sample here seems like a step forward.

smoogipoo (approved):

PR #9650, opened 2020-07-22
Re-namespace multiplayer requests and responses

smoogipoo:
As discussed in discord, putting everything inside the `Online/Multiplayer` namespace. At the same time, I've renamed `RoomScore` to `MultiplayerScore` to match [osu!web naming](https://github.com/ppy/osu-web/blob/2f1c50b135abba49020259f9d51fee57a14d27a9/resources/docs/source/includes/_structures.md).


bdach (approved):
In favour of this :+1: Potential for more splitting here (into models/requests) as already mentioned in the conversation on Discord but this is a good start and we'll see how we go forward.

peppy (approved):

PR #9652, opened 2020-07-22
Use greyscale instead of hue for the logo animation

jorolf:
- [x] Depends on https://github.com/ppy/osu-resources/pull/109


peppy:
Any further documentation on the process here would be appreciated.

- what is this PR aiming to achieve?
- any comparison videos or methods you used to compare so we can objectively agree it's an improvement?

bdach:
I've eyeballed it and the quality seems to have improved with less artifacts, at the cost of a lower number of colour levels available to work with (which leads to the animation looking a bit choppy in slow motion - but it plays too fast to notice with 100% speed, at least for me). An actual viewable comparison as part of the PR would be helpful however. Guess I can make one if need be.

peppy:
Hmm, unfortunate about the frame rate, as one of the remaining goals was to increase it above 60fps.

jorolf:
I couldn't increase the framerate anyway, the original mp4 only had 60 frames to work with. Making it greyscale also helped with finding all artifacts and applying some blur which helps a bit with the perceived framerate.

jorolf:
I made a google drive folder with videos of the new, current and old animations:
https://drive.google.com/drive/folders/1IRQmlA7kFkmiHyghr9RVQiEgwUIiWzpy?usp=sharing
(I think the compression hit them really hard though, you might want to download them to compare them better)

peppy:
In both new versions there's one frame which looks incorrect, right at the start of the animation:

jorolf:
The video provided by flyte does have a pixel in that spot:
frame 1

However that dot should be grey instead of white

peppy:
maybe we should omit / skip that frame since it feels out of place..

jorolf:
That also includes the dot from the circle?

peppy:
i mean it looks like the original video wasn't trimmed properly?

jorolf:
I updated the google drive video to the current versions. I didn't exactly "remove" the frame but instead just removed the dots.

PR #9662, opened 2020-07-24
Increase screenshot quality when using JPEG setting

Poliwrath:
[75 is ImageSharp's default](https://github.com/SixLabors/ImageSharp/blob/b29bb742b90cb87f9f69bfd8db9ce9c921a64cc0/src/ImageSharp/Formats/Jpeg/JpegEncoder.cs#L18)
100 quality (1028 KB): 
osu_2020-07-24_02-03-23
75 quality (default, 150 KB):
osu_2020-07-24_02-03-31
0 quality (36 KB):
osu_2020-07-24_02-03-36

### DIFF
--- a/osu.Game/Configuration/OsuConfigManager.cs
+++ b/osu.Game/Configuration/OsuConfigManager.cs
@@ -106,6 +106,7 @@ namespace osu.Game.Configuration
             Set(OsuSetting.Version, string.Empty);
 
             Set(OsuSetting.ScreenshotFormat, ScreenshotFormat.Jpg);
+            Set(OsuSetting.ScreenshotJpegQuality, 75, 0, 100);
             Set(OsuSetting.ScreenshotCaptureMenuCursor, false);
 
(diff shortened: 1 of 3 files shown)


peppy:
I don't think this should be a setting. The default should be something high enough that it doesn't need changing.

Poliwrath:
Should the quality of ``JpegEncoder`` always be 90, then?

90's file size is 254 KB, 95 is 466 KB

peppy:
90 sounds good, I think. Can you maybe include comparisons at 85, 88, 90, 95? We can choose a good number.

Poliwrath:
80:

85:

88:

90:

95:

peppy:
I think 90 is still a bit low. How about 92 or so?

Poliwrath:
92, 378 KB
osu_2020-07-24_02-20-01

peppy:
Let's go with 92 for now, I think that's a good balance that most will be happy with.

peppy (approved):

PR #9664, opened 2020-07-24
Load nested hitobjects during map load

### LINKED ISSUE #9659: long freezes at the start of spinners in catch
**Describe the bug:** 

I saw manny comments regarding this problem in the spotlights rooms, before you start a spinner freezes for about 2 seconds, 
after freezing ends, performance while spinner lasts is very low.

https://drive.google.com/file/d/1NuY64atUPQYGzws8ZaRRSvZEvHvgjKmI/view?usp=sharing

**osu!lazer version:** 

2020.723.0

**Logs:**

[performance-update.log](https://github.com/ppy/osu/files/4968050/performance-update.log)
[runtime.log](https://github.com/ppy/osu/files/4968051/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4968052/updater.log)
[database.log](https://github.com/ppy/osu/files/4968053/database.log)
[network.log](https://github.com/ppy/osu/files/4968054/network.log)
[performance.log](https://github.com/ppy/osu/files/4968055/performance.log)
[performance-audio.log](https://github.com/ppy/osu/files/4968056/performance-audio.log)
[performance-draw.log](https://github.com/ppy/osu/files/4968057/performance-draw.log)
[performance-input.log](https://github.com/ppy/osu/files/4968058/performance-input.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9659

1. Loads nested hitobjects, samples, etc, asynchronously.
Note that does increase memory usage even in osu!, but I think we want this anyway and save memory usage for DHO pooling.
2. Even after the above, I still had a stutter which was caused by multiple `computeInitialState`s running.
From: 
Screenshot_2020-07-24_19-17-15
To:
Screenshot_2020-07-24_19-29-57

### DIFF
--- a/osu.Game/Rulesets/UI/Scrolling/ScrollingHitObjectContainer.cs
+++ b/osu.Game/Rulesets/UI/Scrolling/ScrollingHitObjectContainer.cs
@@ -2,11 +2,13 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using JetBrains.Annotations;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Caching;
 using osu.Framework.Graphics;
 using osu.Framework.Layout;
+using osu.Framework.Threading;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Objects.Types;
 using osuTK;
@@ -17,7 +19,7 @@ namespace osu.Game.Rulesets.UI.Scrolling
     {
         private readonly IBindable<double> timeRange = new BindableDouble();
         private readonly IBindable<ScrollingDirection> direction = new Bindable<ScrollingDirection>();
-        private readonly Dictionary<DrawableHitObject, Cached> hitObjectInitialStateCache = new Dictionary<DrawableHitObject, Cached>();
+        private readonly Dictionary<DrawableHitObject, InitialState> hitObjectInitialStateCache = new Dictionary<DrawableHitObject, InitialState>();
 
         [Resolved]
         private IScrollingInfo scrollingInfo { get; set; }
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/UI/Scrolling/ScrollingHitObjectContainer.cs
```diff
@@ -215,16 +217,18 @@ protected override void Update()
...
-                        objCache = hitObjectInitialStateCache[obj] = new Cached();
+                    if (!hitObjectInitialStateCache.TryGetValue(obj, out var state))
+                        state = hitObjectInitialStateCache[obj] = new InitialState(new Cached());
 
-                    if (objCache.IsValid)
+                    if (state.Cache.IsValid)
                         continue;
 
+                    state.ScheduledComputation?.Cancel();
+                    state.ScheduledComputation = computeInitialStateRecursive(obj);
```
peppy, line 227: this is really hard to follow... might help to have some inline comments explaining what this is doing

peppy:
- In the above benchmark, are you saying you still feel a stutter from the 2.3ms remaining load?
- Have you tested this on some heavy osu! maps to make sure the async overhead hasn't negatively affected load performance?

smoogipoo:
1. The stutter was from the 60ms, which happens the very first update frame. The 2nd screenshot is what it currently is, which is pretty unnoticeable.
2. I've tested on the Mashiro-iro Symphony Medley, and mem usage went from 1.3GB -> 1.8GB, load time felt similar-ish.

peppy:
I'm guessing that second memory figure is what you end with at the end of the map anyway?

smoogipoo:
You would yeah. Hard to test that though since memory usage goes up naturally - VBOs, judgements/results/etc.

peppy (approved):
let's just go with this for now, local enough to not be a huge issue.

PR #9672, opened 2020-07-25
Add support for taiko mascot animation sprites without a frame index

### LINKED ISSUE #9671: taiko mascot skin animation first frame doesn't show if not suffixed with frame index
**Describe the bug:** 
Already tested this with a known good skin. It seems that certain skins don't show pippidon-miss images. I noticed it with Cinnamon Special (https://mega.nz/file/L9oRVBIQ#YPR_nXA9rQI_w3o_fChTFQruzMpnyjkwBtWV3bs1jbs)
osu_2020-07-25_04-03-09

**osu!lazer version:** 
2020.07.25
**Logs:**

[performance.log](https://github.com/ppy/osu/files/4975070/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4975071/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4975072/updater.log)
[database.log](https://github.com/ppy/osu/files/4975073/database.log)
[network.log](https://github.com/ppy/osu/files/4975074/network.log)

bdach:
Resolves #9671.

# Summary

As noted in the issue thread, some taiko skins wouldn't display mascot sprites if they had a name that didn't have a frame number as the filename suffix (i.e. `pippidonfail.png` instead of `pippidonfail0.png`). This PR adds the support for those cases, by falling back the lookup of the first frame of animation to the indexless sprite if it is present.

This is (visually) covered by the added assets to the test `special-skin`, which is then displayed in `TestSceneDrawableTaikoMascot`.

# Remarks

It would have probably been "a little more correct" to have two separate codepaths for the indexless case and for the animation case, but I think this is okay as it's simple enough and doesn't unnecessarily complicate matters.

I've also removed the guard in `TaikoLegacySkinTransformer` as it was sort of bogus already (only checking first frame of one of 4 animations possible) and is even more bogus given the linked issue, so I just figured that creating the mascot unconditionally is simpler than checking however many filenames (8 would be the number, I suppose). The mascot animations all already support the case of having no frames, so this should be mostly free.

The most annoying part of this PR was by far the test changes - had to fiddle with everything to make the tests not fail wrongly and not pass wrongly. (Also sorry for the programmer art in the test assets. Couldn't include the originals from the issue for obvious reasons and didn't want to spend too much time on doing something presentable myself.)


peppy (approved):
This changes things so there's always a `TaikoMascot`, even in cases it's not skinned at all, but I don't see this as necessarily a bad thing. Looks safe enough for now.

PR #9676, opened 2020-07-25
Make sure HitObject time is not negative

### LINKED ISSUE #9628: Objects with negative time don't respond to drags in the timeline
**Describe the bug:** 
In lazer editor, you can put an object at negative time by drag it in the timeline (and it plays correctly both manually and with autoplay). However, objects with negative time already don't respond to drags in the timeline (you can drag them in the playfield, but not in the timeline, which means not able to change their time).

This circle can be moved in the playfield, but not in the timeline, because it has a negatime time (<00:00:000).

**osu!lazer version:** 
2020.717.0

**Logs:**
Not related I guess

gary-kim:
Fixes #9628

Signed-off-by: Gary Kim <gary@garykim.dev>

### DIFF
--- a/osu.Game/Rulesets/Objects/HitObject.cs
+++ b/osu.Game/Rulesets/Objects/HitObject.cs
@@ -43,7 +43,7 @@ namespace osu.Game.Rulesets.Objects
         public virtual double StartTime
         {
             get => StartTimeBindable.Value;
-            set => StartTimeBindable.Value = value;
+            set => StartTimeBindable.Value = Math.Max(0, value);
         }
 
         public readonly BindableList<HitSampleInfo> SamplesBindable = new BindableList<HitSampleInfo>();


bdach:
Thanks for the contribution, however, CI has reported a test failure in a relevant placement test which is directly caused by this change, which means that the test has to at least be adjusted to pass.

I am additionally unconvinced that this is the correct way to fix things. I'm not sure whether there are maps out there with  objects placed at a negative time, but if there are, doing it this way might "silently" modify these maps. I also am not sure that beat snapping should work at all before the start of the first timing point (indicated by the first white vertical line in the editor timeline).

(Also, sign-off tags don't really do anything here. We're not the kernel.)

bdach (changes requested):
As above.

gary-kim:
@bdach Thank you for such a nice and detailed review!

I'll see if I can make the HitObject adjustable even when the time is negative so as to make sure to not make any silent changes to maps.

> (Also, sign-off tags don't really do anything here. We're not the kernel.)

Just habit, it's required on almost every project I'm a part of. :)

peppy:
I don't think we want to be doing this. Instead we should be supporting negative placements properly.

A starting point would be knowing whether any existing maps use negative hitobjects (ie. whether stable supports it). Then we should decide whether we want to

- Allow negative objects
- Allow changing the start time of the audio track and disallow negative objects (to handle cases where the user wants to place hitobjects before the beginning of the audio track, which should be the ultimate goal of this).

Closing for now as this doesn't seem like a good (even temporary) solution. It's not actually restricting all sources (ie. it's not setting a `MinimumValue`) on the bindable either, so is not fully protecting the value.

PR #9686, opened 2020-07-27
Fix incorrect default hitcircle font overlapping applied to legacy skins

frenzibyte:


peppy:
As a better comparison:

image

peppy (approved):

PR #9688, opened 2020-07-27
Reduce spinner tick and bonus score

### LINKED ISSUE #9667: Reduce spinner bonus weighting in osu! ruleset
Describe the bug: Not sure if this is a bug or not, but I really hope this isn't intentional. On lazer it is now possible to get more than 1 million points due to spinner bonuses. This causes a whole array of issues with the spotlights, meaning it is now substantially easier to get 1million point scores causing all previous weeks to be redundant.

Screenshots or videos showing encountered issue: https://imgur.com/gallery/Iv7zMYH - 19x 100 score over an SS

osu!lazer version: 2020.725.0

peppy:
We're not yet sure on the final usage of bonus score (and how it will affect ranking), but for now the original "100 per spin, 1100 per bonus spin" slider scoring was far too high and allowed spinning to outweigh the accuracy component.

We could also consider scaling back the maximum spin count as a countermeasure for this, but let's start by dialing back scoring.

These are values I found felt quite good (and in line with other bonus scoring) but I'm open to changing them with valid reasoning / example.

- Closes https://github.com/ppy/osu/issues/9667

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
@@ -142,7 +142,7 @@ namespace osu.Game.Rulesets.Osu.Tests
             {
                 // multipled by 2 to nullify the score multiplier. (autoplay mod selected)
                 var totalScore = ((ScoreExposedPlayer)Player).ScoreProcessor.TotalScore.Value * 2;
-                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * 100;
+                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * 10;
             });
 
             addSeekStep(0);
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
@@ -142,7 +142,7 @@ namespace osu.Game.Rulesets.Osu.Tests
             {
                 // multipled by 2 to nullify the score multiplier. (autoplay mod selected)
                 var totalScore = ((ScoreExposedPlayer)Player).ScoreProcessor.TotalScore.Value * 2;
-                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * 10;
+                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * SpinnerTick.SCORE_PER_TICK;
             });
 
             addSeekStep(0);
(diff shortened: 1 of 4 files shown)


bdach:
Not sure that it makes sense for the bonus display to keep counting in increments of 1000:

https://github.com/ppy/osu/blob/e130a0979571961de76c42def83b766a234210cf/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/SpinnerBonusDisplay.cs#L39

although having it count in increments of 40 would probably be even stranger (and I'm not sure if there are people out there that would be unreasonably intensely angry if spinners didn't keep showing 1000s despite not actually awarding that many points). Maybe it should just be `1x, 2x, 3x...` for the number of completed bonus spins?

peppy:
You make a point. I was so used to it showing x,000 that I didn't even notice. I think we should change to represent the true scoring for now and wait to see the complaints.

osu.Game.Rulesets.Osu/Objects/SpinnerBonusTick.cs
```diff
@@ -18,7 +18,7 @@ public SpinnerBonusTick()
 
         public class OsuSpinnerBonusTickJudgement : OsuSpinnerTickJudgement
         {
-            protected override int NumericResultFor(HitResult result) => 1100;
+            protected override int NumericResultFor(HitResult result) => 50;
```
mcendu, line 21: Maybe extract a few constants? The name would be like e.g.
```suggestion
            protected override int NumericResultFor(HitResult result) => SPINNER_BONUS_TICK_VALUE + SPINNER_TICK_VALUE;
```

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
```diff
@@ -142,7 +142,7 @@ public void TestSpinnerNormalBonusRewinding()
             {
                 // multipled by 2 to nullify the score multiplier. (autoplay mod selected)
                 var totalScore = ((ScoreExposedPlayer)Player).ScoreProcessor.TotalScore.Value * 2;
-                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * 100;
+                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * 10;
```
mcendu, line 145:
```suggestion
                return totalScore == (int)(drawableSpinner.Disc.CumulativeRotation / 360) * SPINNER_TICK_VALUE;
```

osu.Game.Rulesets.Osu/Objects/SpinnerTick.cs
```diff
@@ -17,7 +17,7 @@ public class OsuSpinnerTickJudgement : OsuJudgement
         {
             public override bool AffectsCombo => false;
 
-            protected override int NumericResultFor(HitResult result) => 100;
+            protected override int NumericResultFor(HitResult result) => 10;
```
mcendu, line 20:
```suggestion
            protected override int NumericResultFor(HitResult result) => SPINNER_TICK_VALUE;
```

osu.Game.Rulesets.Osu/Objects/SpinnerBonusTick.cs
peppy: considered this but not sure it's necessary. at that point we should just construct a judgement in the test

smoogipoo (approved):

PR #9695, opened 2020-07-27
Fix hit lighting setting not having any effect if toggled during gameplay

bdach:
This broke with #9431; finally got around to finishing it.

# Summary

With introduction of pooling, switching the hit lighting setting mid-gameplay got broken on the osu! ruleset, as the judgements were no longer created and loaded on-demand. This brings back support for toggling that setting mid-gameplay, by always adding the hit lighting to the hierarchy regardless of setting value, and determining if it should be shown at the point of `ApplyHitAnimations()`.

To be extra sure, I checked allocations on [Centipede](https://osu.ppy.sh/beatmapsets/150945) (one playthrough of the map per configuration, on a fresh lazer run), and this is what I got:

| | `master` | this PR |
| :- | :-: | :-: |
| lighting off | image | image |
| lighting on | image | image |

There is a bit of a hit, especially in the case if hit lighting is off (see the allocations in `load()`), but I suppose that (a) it is the price to pay for having that setting work during gameplay and (b) it's a test on possibly *the* benchmark map for osu!, so considering that I don't think those allocs are that egregious.

Additionally hit lighting is now covered with tests.

# Remarks

The reason for storing a reference to `OsuConfigManager` and reading directly from it instead of getting a bindable and subscribing to `BindValueChanged()` is to prevent firing a lot of callbacks when not necessary. (I also prefer the way that it lets the existing lighting animations play out instead of interrupting them.)

There are unfortunately a few satellite changes in here. They are all a result of attempting to write tests, upon which I found a couple of breakages (namely the fact that the lighting would get removed from the hierarchy if the judgement was reused, which is addressed in 5fc7039, and then that always creating the lighting broke the pre-existing fade duration logic, which is addressed in 7ad3101). Not sure if you'll agree with the way I opted to resolve them.

This is sort of the part where the downsides of pooling are starting to show - at the cost of getting pooling and eager initialisation, we get state and state-related bugs like the aforementioned ones. I suppose this again is another intrinsic cost of trying to optimise in this way.

### DIFF
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -130,11 +130,16 @@ namespace osu.Game.Rulesets.Judgements
             if (type == currentDrawableType)
                 return;
 
-            InternalChild = JudgementBody = new Container
+            // sub-classes might have added their own children that would be removed here if .InternalChild was used.
+            if (JudgementBody != null)
+                RemoveInternal(JudgementBody);
+
+            AddInternal(JudgementBody = new Container
             {
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 RelativeSizeAxes = Axes.Both,
+                Depth = -float.MaxValue,
                 Child = bodyDrawable = new SkinnableDrawable(new GameplaySkinComponent<HitResult>(type), _ => JudgementText = new OsuSpriteText
                 {
                     Text = type.GetDescription().ToUpperInvariant(),
@@ -142,7 +147,7 @@ namespace osu.Game.Rulesets.Judgements
                     Colour = colours.ForHitResult(type),
                     Scale = new Vector2(0.85f, 1),
                 }, confineMode: ConfineMode.NoScaling)
-            };
+            });
 
             currentDrawableType = type;
         }
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -139,7 +139,6 @@ namespace osu.Game.Rulesets.Judgements
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 RelativeSizeAxes = Axes.Both,
-                Depth = -float.MaxValue,
                 Child = bodyDrawable = new SkinnableDrawable(new GameplaySkinComponent<HitResult>(type), _ => JudgementText = new OsuSpriteText
                 {
                     Text = type.GetDescription().ToUpperInvariant(),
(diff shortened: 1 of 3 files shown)


osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -130,19 +130,24 @@ private void prepareDrawables()
             if (type == currentDrawableType)
                 return;
 
-            InternalChild = JudgementBody = new Container
+            // sub-classes might have added their own children that would be removed here if .InternalChild was used.
+            if (JudgementBody != null)
```
peppy, line 134: not 100% sure on this (as opposed to having a `CreateBody()` body which also handles things like lighting) but let's go with it for now.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -130,19 +130,24 @@ private void prepareDrawables()
...
+                RemoveInternal(JudgementBody);
+
+            AddInternal(JudgementBody = new Container
             {
                 Anchor = Anchor.Centre,
                 Origin = Anchor.Centre,
                 RelativeSizeAxes = Axes.Both,
+                Depth = -float.MaxValue,
```
peppy, line 142: i'd hope this isn't required?

peppy: Removed as it seems to work as expected without.

peppy (approved):

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
bdach: Mostly added it in an attempt to future-proof the `Remove/Add` hack. It isn't strictly required (and possibly indicates that I *should* go for `CreateBody()`).

peppy: Arguably, `CreateBody` would add more overhead where it could potentially be avoided (for things like lighting which don't need to be refreshed on judgement type change).

I think it's fine as-is. Also, by setting `Depth` you'd potentially break usages where a derived class is trying to draw above.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
smoogipoo: Is this required at all? What's the purpose of remove-then-add? It should probably be documented if there's a specific reason for it.

As in, could this be just:
```csharp
if (JudgementBody == null)
    AddInternal(JudgementBody = new ...);
```

smoogipoo: If this is something that _has_ to happen, be it for ordering purposes, it may even be worthwhile to split this out into separate methods like `RemoveBody()` _and_ `CreateBody()`.

peppy: See the other part of this PR, where hit lighting is added once, ever.

peppy: The remove-then-add is keeping the existing logic, which is to update the `JudgementBody` to the new judgement type (see previous code, `InternalChild = `).

smoogipoo (approved):
Seems ok

PR #9700, opened 2020-07-28
Display low health fading effect vertically rather than horizontally

### LINKED ISSUE #9376: Mania (or catch) red tint should display to the sides
**Describe the feature:**

_Originally posted by @LittleEndu in https://github.com/ppy/osu/issues/8789#issuecomment-615744131_

Essentially red tint when failing should be on the side like the comment.

And maybe catch too.

ynjoe:
I decreased the gradient size to avoid overlaps with game objects
image
Closes #9376

### DIFF
--- a/osu.Game/Screens/Play/HUD/FailingLayer.cs
+++ b/osu.Game/Screens/Play/HUD/FailingLayer.cs
@@ -29,7 +29,7 @@ namespace osu.Game.Screens.Play.HUD
 
         private const float max_alpha = 0.4f;
         private const int fade_time = 400;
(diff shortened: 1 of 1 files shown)


bdach:
Source looks fine. Visually I'd also say it's fine (even though for some reason horizontal looked better to me, but I suppose visibility is more important) - not approving outright though as I'd want to make sure we agree on the visual part.

peppy:
Let's go with this for now. I think we can iterate on this in the future to make it better fit the rest of the UI (after the next HUD design refresh hits).

peppy (approved):

PR #9701, opened 2020-07-28
Fix tracked bindable updates not getting cancelled correctly

### LINKED ISSUE #9682: Difficulty doesnt load between mode changes with incompatible mods
**Describe the bug:** 
Enter song select
Switch to mania mode
Open Mods dialog
Select 4k
Without closing mods dialog change modes to standard
Now you can close the mods dialog and see that for the currently selected map all star ratings are empty. Once you select a difficulty it will load its rating like normal.

https://imgur.com/c6WScF0

**osu!lazer version:** 
2020.725.0 Linux Appimage

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/4978194/runtime.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9682

I've tried to write a test for this, but I haven't got it to fail on master since it's quite timing dependent...

In short, because the linked cancellation sources were being disposed, cancelling the `trackedUpdateCancellationSource` did not propagate to the linked sources and the updates therefore also didn't get cancelled.
To fix this I've stored the linked sources in a list that is disposed when the next request to update tracked bindables comes in.

I've also added a disposal of the `TaskScheduler` in this PR, to fix potentially having multiple schedulers running through multiple gamehost tests (similar to https://github.com/ppy/osu/pull/9470).

### DIFF
--- a/osu.Game/Beatmaps/BeatmapDifficultyManager.cs
+++ b/osu.Game/Beatmaps/BeatmapDifficultyManager.cs
@@ -146,6 +146,22 @@ namespace osu.Game.Beatmaps
         }
 
         /// <summary>
+        /// Creates a new <see cref="BindableStarDifficulty"/> and triggers an initial value update.
+        /// </summary>
+        /// <param name="beatmapInfo">The <see cref="BeatmapInfo"/> that star difficulty should correspond to.</param>
+        /// <param name="initialRulesetInfo">The initial <see cref="RulesetInfo"/> to get the difficulty with.</param>
+        /// <param name="initialMods">The initial <see cref="Mod"/>s to get the difficulty with.</param>
+        /// <param name="cancellationToken">An optional <see cref="CancellationToken"/> which stops updating the star difficulty for the given <see cref="BeatmapInfo"/>.</param>
+        /// <returns>The <see cref="BindableStarDifficulty"/>.</returns>
+        private BindableStarDifficulty createBindable([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo initialRulesetInfo, [CanBeNull] IEnumerable<Mod> initialMods,
+                                                      CancellationToken cancellationToken)
+        {
+            var bindable = new BindableStarDifficulty(beatmapInfo, cancellationToken);
+            updateBindable(bindable, initialRulesetInfo, initialMods, cancellationToken);
+            return bindable;
+        }
+
+        /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -220,6 +237,14 @@ private bool tryGetExisting(BeatmapInfo beatmapInfo, RulesetInfo rulesetInfo, IE
             return difficultyCache.TryGetValue(key, out existingDifficulty);
         }
 
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            cancelTrackedBindableUpdate();
+            updateScheduler.Dispose();
```
bdach, line 245: Probably safer to null-guard this as this can potentially be called from the finaliser (if `isDisposing == false`) and the object is not guaranteed to be fully initialised in that case.

```suggestion
            updateScheduler?.Dispose();
```

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -109,43 +109,40 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
...
+            foreach (var c in linkedCancellationSources)
             {
-                // We're on a threadpool thread, but we should exit back to the update thread so consumers can safely handle value-changed events.
-                Schedule(() =>
-                {
-                    if (!cancellationToken.IsCancellationRequested)
-                        bindable.Value = t.Result;
-                });
-            }, cancellationToken);
+                c.Cancel();
```
bdach, line 141: Not sure what this second cancel is for - shouldn't all `linkedCancellationSources` get cancelled once `trackedUpdateCancellationSource` is?

bdach (commented):
I'm wondering whether we need to be enforcing thread safety/re-entrancy of `cancelTrackedBindableUpdate`, although it should throw errors loudly if it turns out it needs to be thread-safe (i.e. because of the possibility of trying to cancel a disposed token source). I always get a little uneasy when this sort of thing is done in a `BindValueChanged()` callback but not 100% sure whether I'm worrying about things I don't need to.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
smoogipoo: Sure, it was mostly for safeguarding.

peppy (approved):

bdach (approved):

PR #9702, opened 2020-07-28
Adjust mania scoring to be 95% based on accuracy

### LINKED ISSUE #9683: Combo is still too heavily biased & 320s (Perfects) should reward more accuracy score
image

The Accuracy/Combo split should be like 95%:5%
 
If lazer scoring counts pefects towards accuracy, then I feel like higher accuracy scores should always be higher, regardless of combo. Perfects could hold more weight of your score.

Combo could effect scoring but only when two plays have around the same accuracy (and the person with the higher combo would get the bonus)

**Current**
Cryolien
`(800000 * .9632) + (200000 * (3394/4181)) = 932,913`
Gonzaalooo
`(800000 * .9659) + (200000 * (2378/4181)) = 886,472`
KcHecKa
`(800000 * .9695) + (200000 * (1950/4181)) = 868,879`

**95:5 Split**
Cryolien
`(950000 * .9632) + (050000 * (3394/4181)) = 955,628`
Gonzaalooo
`(950000 * .9659) + (050000 * (2378/4181)) = 946,043`
KcHecKa
`(950000 * .9695) + (050000 * (1950/4181)) = 944,344`

I don't want to make it seem like I'm nitpicking because I'm unhappy with my scores, but in my opinion, combo should have little to no effect on score at all. It's so arbitrary on where you miss and the accuracy of the score entirely is a better indicator on your 'consistency.' Each note could simply have their own weight, depending on your judgments, and 320s/perfects could give a bit more weight than 300s/greats. Combo could be the deciding factor between similar accuracy scores; or a minuscule 1-2% combo weight of the total score.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9683

---

- Changed to 95:5 accuracy to combo ratio.
- Increased PERFECT score reward from 320 to 350.

### DIFF
--- a/osu.Game.Rulesets.Mania/Judgements/ManiaJudgement.cs
+++ b/osu.Game.Rulesets.Mania/Judgements/ManiaJudgement.cs
@@ -28,7 +28,7 @@ namespace osu.Game.Rulesets.Mania.Judgements
                     return 300;
 
                 case HitResult.Perfect:
-                    return 320;
+                    return 350;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


bdach (approved):
looks straightforward 👍

PR #9704, opened 2020-07-28
Delay loading of cover backgrounds in score panels

smoogipoo:
Stops firing off potentially hundreds of web reqs (with pagination).


peppy:
Is there a case where this *isn't* what we want from `UserCoverBackground`? Just thinking maybe it should be applied to the class itself..

smoogipoo:
The only other use that doesn't have this is `ProfileHeader`.

peppy:
Sounds like we should definitely embed the behaviour if feasible..

bdach (approved):
looks alright to me 👍

peppy (approved):

PR #9706, opened 2020-07-28
Implement multiplayer score pagination

smoogipoo:
Preview: https://drive.google.com/file/d/1eMZehxQ8bCbvVXETxnuOqHtV7umqFvjR/view?usp=sharing

Probably need a better UI for this but currently:
1. If the user has a score, it will show the user's score and all scores around it.
2. If the user does not have a score, it will show from the highest score.

Consider merging https://github.com/ppy/osu/pull/9704 before this to not thrash connections.

I'm not super keen on my implementation around `FetchNextPage`, but I don't know how to implement this without severely overcomplifying it. Hopefully the per-update code is fine?


peppy:
- How do you know when you hit the end?
- How do you know when it's loading?

Can we add loading spinners when loading is occurring?

bdach:
Other than my review comments above (mostly nitpicky), I can't find much fault with the code itself (although I agree with the aforementioned UX concerns about indication of loading and end of results).

I also *think* that the fetching in `Update()` is fine. I'd hope the runtime can make those branches disappear for us most of the time (and I can't see any indication that this is not the case in global statistics). The only thing that worries me a little is the concurrency around `nextPageRequest` reads/writes but I can't see a potential for any race that could break things (although I *am* a little tired today reading this, so yeah).

At the end I'll highlight one more potential UX concern - the scrollbar does not really play well with the pagination ([see video](https://drive.google.com/file/d/18RitM4aXPsFB6r0Mc2iLvuQQXYcmxEz-/view?usp=sharing) - scrolling to either end gets very jumpy). I'm not too sure whether something reasonable can be done about that, though...

PR #9709, opened 2020-07-29
Include executable hash when submitting multiplayer scores

peppy:
Temporary measure to avoid scores being submitted from an earlier build where there was a breaking score change made.

Blocking until osu-web changes are made and can be tested.


smoogipoo:
Is this going to handle correctly for mobile devices?

peppy:
No idea!

smoogipoo:
Let me rephrase: are we currently storing the hashes of all builds (macOS/Windows/Linux/iOS/Android) for the osu-web check to see them?

peppy:
We can store any hashes required, yes.

peppy:
I'll test ios / android now to make sure the hash is being retrieved correctly.

peppy:
Seems to work as expected, but I did notice that we should probably be hashing `osu.Game.dll` as executables for net core are regularly just wrappers and may not change between builds.

In the future we will also need to consider hashing individual rulesets as they may have their own changes, but for now `osu.Game.dll` should suffice, based on how we do releases.

smoogipoo (approved):

PR #9711, opened 2020-07-29
Add pooling for mania hit explosions

smoogipoo:
Initialised with 5 drawables per pool, one pool per column, with no upper limit (though around 10-15 per column is a viable upper limit).


osu.Game.Rulesets.Mania/UI/IHitExplosion.cs
```diff
@@ -0,0 +1,19 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Game.Rulesets.Judgements;
+
+namespace osu.Game.Rulesets.Mania.UI
+{
+    /// <summary>
+    /// Common interface for all hit explosion bodies.
+    /// </summary>
+    public interface IHitExplosion
```
peppy, line 11: thinking this may be needed in many more places / rulesets, but i guess it's fine to be local for now.

PR #9712, opened 2020-07-29
Merge GlobalSkinConfiguration settings into the LegacySetting enum

frenzibyte:
Got me confused on where to put a couple of legacy settings I was going to support (`ComboPrefix` and `ComboOverlap` for catch-specific combo counter).


bdach (dismissed):
seems reasonable, not a fan of the `when` syntax still but as long as it's only this one lookup I suppose it's not too bad right now

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -142,8 +133,11 @@ protected override void Dispose(bool isDisposing)
 
                     break;
 
+                case LegacySkinConfiguration.LegacySetting s when s == LegacySkinConfiguration.LegacySetting.Version:
```
peppy, line 136: please remove use of `when` since we have decided no to use it for now.

peppy (changes requested):

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -142,8 +133,17 @@ protected override void Dispose(bool isDisposing)
...
                     break;
 
+                case LegacySkinConfiguration.LegacySetting legacy:
+                    switch (legacy)
+                    {
+                        case LegacySkinConfiguration.LegacySetting.Version:
+                            return SkinUtils.As<TValue>(new Bindable<decimal>(Configuration.LegacyVersion ?? LegacySkinConfiguration.LATEST_VERSION));
+                    }
+
+                    goto default;
```
peppy, line 143: you replaced one avoided syntax with another.

peppy (dismissed):

osu.Game/Skinning/LegacySkin.cs
frenzibyte: got any alternatives? I can't `break;` because other settings need to be parsed inside the `default` case.

peppy: i don't follow

frenzibyte: How can I achieve the same behaviour with a different syntax (i.e. how can I resolve this conversation)?

peppy: I don't follow what I should be looking at. Can you provide a link to the code where `HitCirclePrefix` is doing this?

bdach: I would split handling the entire `LegacySkinConfiguration` enum to a switch in a separate method and call that inside of this main switch. That would also require extracting a method for the generic lookup path (so that it can be called in both the main switch and the legacy-specific one).

bdach: I have pushed in the aforementioned refactoring here. Let me know if it is more to your liking.

smoogipoo (approved):

smoogipoo:
Tbqh the original (as in, the `when A == B` version) read fine to me, and the new code looks overengineered by comparison. But I don't want to dwell on this PR any further.

PR #9713, opened 2020-07-29
Add pooling for mania judgements

smoogipoo:
Keeping it simple for now. I believe even pooling might be excessive here but only because of the transform - it might look better to not fade-in the text every time, so a singular instance might be preferable.

Can do that in the future though, this preserves current logic.


peppy:
Against single instance (this always bugs me about VSRG), but we can apply some transform to previous instance to make sure overlap is not offensive. Pooling is definitely the right direction IMO.

bdach (approved):
seems good :+1: 

Visually seems inoffensive to me even on some pretty dense maps but I don't play mania all that much so probably not the person to speak on that. As mentioned that can be adjusted later on, though.

PR #9719, opened 2020-07-30
Add spinner skinning support

### LINKED ISSUE #8845: Add spinner skinning support
Since most if not all modes are now skinnable, I'd think implementing custom spinner support would be pretty neat for those of us who have a favorite spinner skin.

peppy:
In the interest of keeping this PR manageable (for myself) I'm getting this out in a mostly-complete state.

Missing:
- `RotationTracker` should probably be a dumb object which doesn't actually track completion. That should be moved to `DrawableSlider`
- SPIN! and CLEAR! displays are not drawn.
- Bonus display is not skinnable (probably will come a lot later, since it's not really a huge deal).
- Spinner spinning sound is still missing, but only vaguely related.
- Hidden and traceable behaviour is kinda regressed, due to the background elements no longer being exposed. Working out a sane way forward for these. Not a huge deal IMHO.

- Closes https://github.com/ppy/osu/issues/8845


smoogipoo:
Is this intended to be reviewed and/or merged, or is it awaiting on further changes?

peppy:
I think it's fine as-is. Only adds functionality and in a way that it can improve game experience already (hidden/traceable mod is slightly regressed but I don't think that will bother anyone).

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
```diff
@@ -0,0 +1,170 @@
...
+            var colour = complete ? completeColour : normalColour;
+
+            ticks.FadeAccent(colour.Darken(1), duration);
+            fill.FadeAccent(colour.Darken(1), duration);
+
+            background.FadeAccent(colour, duration);
+            centre.FadeAccent(colour, duration);
+        }
+
+        private bool updateCompleteTick() => completeTick != (completeTick = (int)(drawableSpinner.RotationTracker.CumulativeRotation / 360));
```
bdach, line 148: wow, this, uh... I thought I had brain damage for a little bit when reading this as to how this can ever be true (but it can!).

[analogous, simpler example for the curious](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYGYAE9MGFMDeqmJ2CAbJmAHbCYAOA3MaSyTBXgBQCUhbpEvUwBeTOmYpBAXwECMZSgCMA9ioiYAsr35TBpGAHYGmAIRiuwsQFYekmXL3s4cXfoMIAnF2087AwQ5vX38nfSCfXlD3di9Iv3sYiJDE92So1MxZFGkgA===). CLR implicitly stores the old value of the variable onto the bottom of the stack before proceeding with execution (below is the explanation for the first run that returns `true`; every other run after this will return `false`):

```
                                        p = 3   stack: (empty)
        IL_0000: ldsfld int32 C::p      p = 3   stack: 3
        IL_0005: ldc.i4.5               p = 3   stack: 3 5
        IL_0006: dup                    p = 3   stack: 3 5 5
        IL_0007: stsfld int32 C::p      p = 5   stack: 3 5
        IL_000c: ceq                    p = 5   stack: 0
        IL_000e: ldc.i4.0               p = 5   stack: 0 0
        IL_000f: ceq                    p = 5   stack: 1
        IL_0011: ret                    p = 5   function returns 1
```

so this effectively translates to, "store the value on the right side of the assignment to `completeTick`, then compare the value assigned to the old value and return the result of that comparison".

I don't know about you but I've literally never seen this pattern before. might be worth an inline comment or something? it read to me like a bug at first.

osu.Game.Rulesets.Osu/Skinning/LegacyNewStyleSpinner.cs
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        protected override void Update()
+        {
+            base.Update();
+            spinningMiddle.Rotation = discTop.Rotation = drawableSpinner.RotationTracker.Rotation;
+            discBottom.Rotation = discTop.Rotation / 3;
+
+            Scale = new Vector2(final_scale * 0.8f
+                                + (float)Interpolation.ApplyEasing(Easing.Out, drawableSpinner.Progress) * (final_scale * 0.2f));
```
bdach, line 97: You could hoist the multiplication by `finalScale` to the outer scope like so:
```suggestion
            Scale = new Vector2(final_scale * (0.8f + (float)Interpolation.ApplyEasing(Easing.Out, drawableSpinner.Progress) * 0.2f));
```

Unsure if less or more readable.

bdach (commented):
not a proper review, only managed to find the time to skim the source, haven't had a proper visual look yet

osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
```diff
@@ -4,37 +4,32 @@
...
     {
-        private readonly Container content;
-        protected override Container<Drawable> Content => content;
-
         private int depthIndex;
 
         public TestSceneSpinner()
         {
-            base.Content.Add(content = new OsuInputManager(new RulesetInfo { ID = 0 }));
+            //            base.Content.Add(content = new OsuInputManager(new RulesetInfo { ID = 0 }));
```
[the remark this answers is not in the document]
peppy: oh this was on my list of things to check. by removing this you can no longer interact with the test scene. i'll see if i can add it back.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
peppy: i'll split it out to read better

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
```diff
@@ -102,6 +102,14 @@ public override Drawable GetDrawableComponent(ISkinComponent component)
                             Scale = new Vector2(0.8f),
                             Spacing = new Vector2(-overlap, 0)
                         };
+
+                case OsuSkinComponents.SpinnerBody:
+                    if (Source.GetTexture("spinner-top") != null)
+                        return new LegacyNewStyleSpinner();
+                    else if (Source.GetTexture("spinner-background") != null)
+                        return new LegacyOldStyleSpinner();
```
smoogipoo, line 110:
```suggestion
                    bool hasBackground = Source.GetTexture("spinner-background") != null;

                    if (Source.GetTexture("spinner-top") != null && !hasBackground)
                        return new LegacyNewStyleSpinner();
                    else if (hasBackground)
                        return new LegacyOldStyleSpinner();
```

Before:
Screenshot_2020-07-31_23-57-15

After:
Screenshot_2020-08-01_00-00-54

Skin: https://drive.google.com/drive/folders/1ck7IcwCJSC0QbQowneyTUhCIxMFqZTJR

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Skinning/OsuLegacySkinTransformer.cs
peppy: Yeah, wasn't sure about this. Looks like a better check.

smoogipoo: For future reference, full conditional is:
`newStyleSpinner = SkinManager.UseNewLayout && (SkinManager.IgnoreBeatmapSkin || TextureManager.Load(@"spinner-circle", SkinSource.Beatmap) == null) && TextureManager.Load(@"spinner-background", SkinSource.Skin) == null;`

So this should check if version > 1 and do another check against the beatmap skin too. But I think this is MVP for now.

smoogipoo:
Something in here seems to have botched the colours and scales of the default spinner:

Master:
Screenshot_2020-08-01_00-12-23

This PR:
Screenshot_2020-08-01_00-13-07

I'm not so much concerned about the outer colour and ticks, but I am concerned about the inner fill, which looks almost invisible to me.

peppy:
Scale changes were kinda "intended". Alpha should be fixed now.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
```diff
@@ -100,13 +100,20 @@ protected override void Update()
...
+                    fill.FinishTransforms(false, nameof(Alpha));
+                    fill
+                        .FadeTo(tracking_alpha + 0.2f, 60, Easing.OutExpo)
+                        .Then()
+                        .FadeTo(tracking_alpha, 250, Easing.OutQuint);
+                }
+            }
+            else
+            {
+                fill.Alpha = (float)Interpolation.Damp(fill.Alpha, drawableSpinner.RotationTracker.Tracking ? tracking_alpha : idle_alpha, 0.98f, (float)Clock.ElapsedFrameTime);
```
smoogipoo, line 116: I got this while rewinding:
```
[runtime] 2020-07-31 15:31:29 [error]: An unhandled error has occurred.
[runtime] 2020-07-31 15:31:29 [error]: System.ArgumentOutOfRangeException: exponent has to be bigger than 0, but is -16.66666603088379. (Parameter 'exponent')
[runtime] 2020-07-31 15:31:29 [error]: at osu.Framework.Utils.Interpolation.Damp(Double start, Double final, Double base, Double exponent)
[runtime] 2020-07-31 15:31:29 [error]: at osu.Game.Rulesets.Osu.Objects.Drawables.Pieces.DefaultSpinnerDisc.Update() in /home/smgi/Repos/osu/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs:line 116
[runtime] 2020-07-31 15:31:29 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree()
```

smoogipoo (changes requested):

smoogipoo (approved):

smoogipoo (commented):
I've tested this on a bunch of skins using various legacy elements that I've seen throughout the years, and all seem to work as expected :+1:

Vinigho:
can we have an option to use the osu!lazer spinner even with a skin equiped? because I really liked the new spinner.

frenzibyte:
You can locally do that by exporting the skin, removing all the spinner elements, then reimport back. osu!lazer will eventually have a full skin editor where you can modify the elements of it (removing for them to use default).

PR #9723, opened 2020-07-30
Add osu!catch banana catching sounds

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/110
- Uses new "catch-banana" but falls back to legacy "metronomelow" for compatibility.

Not sure if we want to add bank support. Shouldn't be a huge reach, but also was not previously supported and has never been requested.


smoogipoo (approved):
lgtm

PR #9729, opened 2020-07-30
Ignore custom sample bank when performing sample lookups from user skin

### LINKED ISSUE #8928: Hitsound discrepancy on legacy skins
**Describe the bug:** 
- Certain hitsounds are different on lazer compared to stable
- Reproduce:
  - Skin: [chocomint's skin](https://shigeskinss.s-ul.eu/gEfiNoTv)
  - Map: [Panda Eyes & Teminite - Highscore [Game Over]](https://osu.ppy.sh/beatmapsets/332532#osu/736215)
  - Pick any replay, hitsound should be different on both versions immediately

**osu!lazer version:** 2020.429.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/4568479/database.log)
[network.log](https://github.com/ppy/osu/files/4568480/network.log)
[performance-audio.log](https://github.com/ppy/osu/files/4568481/performance-audio.log)
[performance-draw.log](https://github.com/ppy/osu/files/4568482/performance-draw.log)
[performance-update.log](https://github.com/ppy/osu/files/4568483/performance-update.log)
[performance.log](https://github.com/ppy/osu/files/4568484/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4568485/runtime.log)

bdach:
Resolves #8928.

# Summary

As outlined in the issue thread, stable does not take into account the custom sample set/bank when looking up samples from the user skin, but does take it into account on the beatmap skin. To resolve, add a local check in `LegacySkin` to replicate behaviour.

In addition to the test cases attached, the following videos demonstrate the fail case with beatmap/user skin combination as reported in the issue:

* [stable](https://drive.google.com/file/d/14YWxgZoZM9xcOYSqzsptk_Il8aBQGxOG/view?usp=sharing) for reference,
* [`master`](https://drive.google.com/file/d/1G9ok1vAzqqJJPLjUZ9RgauHg41gIpMC7/view?usp=sharing),
* [this PR](https://drive.google.com/file/d/14BBxl7M_Dzh1Y-2NbRHTDKX9HHKRVtEK/view?usp=sharing).

# Remarks

The code differs slightly from what I posted in the issue. I have discarded [this commit](https://github.com/ppy/osu/commit/a4dba5910b9a3a022f0daccb5e91b3eda7fb170a) after reading [this old discussion](https://discordapp.com/channels/188630481301012481/188630652340404224/613972199442874368) that I wasn't around to witness yet, and kept the bank-less lookup local to `LegacySkin` too.

Additionally includes coverage for some untested parts of the lookup code (taiko lookups as well as bank-less ones). A careful reader might notice that the test cases for taiko aren't checking `{taiko-}?normal-hitnormal2`, and that is because, as it turns out, that is currently unsupported in lazer. Adding rudimentary support is easy enough, but, *unfortunately* there's another edge case there - stable skips those lookups on converted beatmaps, which I don't even know how to properly deal with right now.

I don't remember the last time I went back and forth on a branch as many times as this one. I hate this code but I don't really see a better way out; not sure if I left enough comments to explain the hackery. I'm sort of worried this will regress something I don't expect to, but I don't see a way out of that other than merging and dealing with potential reports later (which I will be looking out for and investigating/handling as soon as possible).

### DIFF
--- a/osu.Game.Rulesets.Taiko.Tests/Resources/SampleLookups/taiko-hitobject-beatmap-custom-sample-bank.osu
+++ b/osu.Game.Rulesets.Taiko.Tests/Resources/SampleLookups/taiko-hitobject-beatmap-custom-sample-bank.osu
@@ -1,0 +1,10 @@
+osu file format v14
+
+[General]
+Mode: 1
+
+[TimingPoints]
+0,300,4,1,2,100,1,0
+
+[HitObjects]
+444,320,1000,5,0,0:0:0:0:
--- a/osu.Game.Rulesets.Taiko.Tests/TestSceneTaikoHitObjectSamples.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/TestSceneTaikoHitObjectSamples.cs
@@ -1,0 +1,52 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Reflection;
+using NUnit.Framework;
+using osu.Framework.IO.Stores;
+using osu.Game.Tests.Beatmaps;
+
+namespace osu.Game.Rulesets.Taiko.Tests
+{
+    public class TestSceneTaikoHitObjectSamples : HitObjectSampleTest
+    {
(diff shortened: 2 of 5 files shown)


smoogipoo (approved):
lgtm!

PR #9745, opened 2020-08-01
Implement DashboardBeatmapListing component for use in Social Overlay

EVAST9919:
Screenshot_1


osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
```diff
@@ -0,0 +1,164 @@
...
+using osu.Framework.Graphics.Sprites;
+using osu.Framework.Input.Events;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.Drawables;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Containers;
+using osu.Game.Graphics.Sprites;
+using osuTK;
+
+namespace osu.Game.Overlays.Dashboard.Dashboard
```
bdach, line 17: pretty bad namespace choice (why does it have `Dashboard` twice?)

would suggest using something more descriptive (`Dashboard.Components`? `Dashboard.Panels`?)

osu.Game/Overlays/Dashboard/Dashboard/DrawableBeatmapsList.cs
```diff
@@ -0,0 +1,57 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Beatmaps;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
+using osuTK;
+
+namespace osu.Game.Overlays.Dashboard.Dashboard
+{
+    public abstract class DrawableBeatmapsList : CompositeDrawable
```
bdach, line 16:
```suggestion
    public abstract class DrawableBeatmapList : CompositeDrawable
```

(+ usages)

osu.Game/Overlays/Dashboard/Dashboard/DrawableBeatmapsList.cs
```diff
@@ -0,0 +1,57 @@
...
+                        Text = CreateTitle(),
+                        Padding = new MarginPadding { Left = 10 }
+                    }
+                }
+            };
+
+            flow.AddRange(beatmaps.Select(CreateBeatmapPanel));
+        }
+
+        protected abstract string CreateTitle();
```
bdach, line 53:
```suggestion
        protected abstract string Title { get; }
```

the ballpark rule is: if it's simple data, there's no reason to have it as a method; methods generally *do* or construct stuff rather than just return stuff

osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            RelativeSizeAxes = Axes.X;
+            Height = 60;
+            Children = new Drawable[]
+            {
+                background = new Box
```
bdach, line 44: would rename to `hoverBackground` or something

osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
```diff
@@ -0,0 +1,164 @@
...
+                                        AutoSizeAxes = Axes.Y,
+                                        RelativeSizeAxes = Axes.X,
+                                        Direction = FillDirection.Vertical,
+                                        Children = new Drawable[]
+                                        {
+                                            new OsuSpriteText
+                                            {
+                                                RelativeSizeAxes = Axes.X,
+                                                Truncate = true,
+                                                Font = OsuFont.GetFont(weight: FontWeight.SemiBold),
```
bdach, line 101: web-side these are all `Regular` but using that here is pretty bad due to font AA

would have been nice if you had mentioned that (if you were aware)

osu.Game/Overlays/Dashboard/Dashboard/DrawableBeatmapsList.cs
```diff
@@ -0,0 +1,57 @@
...
+            FillFlowContainer flow;
+
+            RelativeSizeAxes = Axes.X;
+            AutoSizeAxes = Axes.Y;
+            InternalChild = flow = new FillFlowContainer
+            {
+                RelativeSizeAxes = Axes.X,
+                AutoSizeAxes = Axes.Y,
+                Direction = FillDirection.Vertical,
+                Spacing = new Vector2(0, 10),
```
bdach, line 37: not sure why all these horizontal margins are being enforced locally everywhere - the entire dashboard sidebar has the same common 10px padding on the sides

bdach (dismissed):

osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
TheWildTree: AFAIK font weights for some reason are offset by one (600 here looks like 500 on web), or at least that was the case last time I checked - quite a few overlay elements have 'wrong' font weights in place to match web visually. Font AA probably makes it a bit harder to compare, too.

might be the reason for the different weight being chosen here.

bdach: doubt it, as other weights used in other places in this PR match (i.e. the `Bold` usages)

osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
EVAST9919: Made this because we have dashboard overlay and dashboard tab in it

bdach: it reads pretty wrong to me.

on another note, how about `Dashboard.Home`? (as that's where the `home` link leads to, the URL web-side is also `/home`?)

smoogipoo (dismissed):
lgtm

osu.Game/Overlays/Dashboard/Dashboard/DashboardBeatmapPanel.cs
smoogipoo: We should probably keep these consistent, even if it looks bad for the time being, and figure out a solution that works universally in the future.

smoogipoo (approved):

PR #9747, opened 2020-08-01
Apply height constraints to all settings dropdowns

### LINKED ISSUE #9744: Settings -> Audio -> Devices scrollable list doesnt work
**Describe the bug:** 
When you have more than 18 devices in the list it goes off screen, you can scroll and see the devices past the bottom but they snap back below the screen. Similar to what happens if you try to scroll past the bottom of a list.

typical_suffering_linux_user

In total I guess osu! is finding 20 or so devices but I can only see the devices past PulseAudio briefly while scrolling and cannot select them without using keyboard controls. Though it's hard to say where the thing you want to select is since the list still doesnt scroll even when using keyboard controls.

**osu!lazer version:** 
2020.725.0 Linux appimage

**Logs:**

frenzibyte:
Closes #9744


bdach (approved):
seems reasonable

Joehuu (approved):

PR #9751, opened 2020-08-01
Add option to disable certain global actions while playing

### LINKED ISSUE #9743: Add an option to disable "ctr + key actions" in game
**Describe the new feature:** 

As a mania player, I use s-d-f-space-j-k-l for playing, and on many occasions, I have pressed without intention the ctrl key with my left hand while playing, opening direct overlay, (ctr + d) and that has led me to lose several times.
I would like to have an option in the menu to disable ctr + key actions (just as the windows key) to disable these actions in gameplay.

As an option in the main menu where you can choose to disable these actions if they bother you as it does for me.

vntxx:
Fixes #9743 

_I wasn't quite sure about two actions so removed them from this logic in in the second commit._ **OUTDATED**

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -114,9 +114,6 @@ namespace osu.Game
 
         private readonly List<OverlayContainer> visibleBlockingOverlays = new List<OverlayContainer>();
 
-        private bool bindingDeactivateGlobalActions;
-        private bool screenDeactivateGlobalActions;
-
         public OsuGame(string[] args = null)
         {
             this.args = args;
@@ -217,11 +214,6 @@ namespace osu.Game
 
             SelectedMods.BindValueChanged(modsChanged);
             Beatmap.BindValueChanged(beatmapChanged, true);
-
-            Bindable<bool> deactivateGlobalActions = LocalConfig.GetBindable<bool>(OsuSetting.GameplayDisableGlobalActions);
-            bindingDeactivateGlobalActions = deactivateGlobalActions.Value;
-            deactivateGlobalActions.ValueChanged += value => bindingDeactivateGlobalActions = value.NewValue;
-            screenDeactivateGlobalActions = false;
         }
 
         private ExternalLinkOpener externalLinkOpener;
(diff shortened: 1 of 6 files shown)


osu.Game/OsuGame.cs
```diff
@@ -114,6 +114,9 @@ public class OsuGame : OsuGameBase, IKeyBindingHandler<GlobalAction>
 
         private readonly List<OverlayContainer> visibleBlockingOverlays = new List<OverlayContainer>();
 
+        private bool bindingDeactivateGlobalActions;
+        private bool screenDeactivateGlobalActions;
```
vntxx, line 118: not sure about those variable names

osu.Game/OsuGame.cs
```diff
@@ -214,6 +217,11 @@ private void load()
 
             SelectedMods.BindValueChanged(modsChanged);
             Beatmap.BindValueChanged(beatmapChanged, true);
+
+            Bindable<bool> deactivateGlobalActions = LocalConfig.GetBindable<bool>(OsuSetting.GameplayDisableGlobalActions);
```
vntxx, line 221: this variabe name could also get a better name

osu.Game/Screens/Play/Player.cs
```diff
@@ -50,6 +50,8 @@ public class Player : ScreenWithBeatmapBackground
 
         public override bool HideOverlaysOnEnter => true;
 
+        public override bool AllowBlockedGlobalActions => true;
```
vntxx, line 53: might have to move this elsewhere but unsure for now

frenzibyte: if disabling certain global action in-game is the way to go, then judging by the name of this property alone, this shouldn't really exist anywhere, sounds like you just want to attach an `IKeyBindingHandler<GlobalAction>` to `Player` and swallow them.

that way you don't have to complicate `OsuGame` and weirdly (talking from someone who just reads that class) allowing `ResetInputSettings` and `ToggleGameplayMouseButtons` but blocking anything else.

frenzibyte (changes requested):

osu.Game/Screens/Play/Player.cs
vntxx: Oh wow I didn't even think of that. Will rewrite and push now.

vntxx:
I might want to rename `gameplayDisableGlobalActions` to `gameplayDisableOverlays` now

osu.Game/Screens/Play/Player.cs
vntxx: And I just noticed that this doesn't work as intended after I've already pushed

bdach: which probably wouldn't have happened if you had added tests for this. please do so.

Game4all: If the goal is to disable potential activation of overlays during gameplay, wouldn't it be better to use the `InitialOverlayActivation` property (see in `OsuScreen`), and possibly the game level bindable `OverlayActivationMode` to re-enable use of overlays when gameplay is paused?

vntxx:
Closed in favor of #9766

PR #9757, opened 2020-08-02
Fix spinner test failures

bdach:
Resolves spinner rewind test failures.

# Summary

The recent spinner PR caused reproducible test failures on the tests that checked the rotation after rewind. The failure itself was related to switching a lerp to a damp with a quite high exponentiation base of 0.99. This meant that the rotation values were further off from the expected ones than before, at the cost of the rotation being less rigid and more momentum-based.

This PR changes the tolerance values in failing test cases to be relative to the marginal values instead of absolute, therefore allowing for more tolerance when checking the value. This means that the rotation value might not necessarily be 100% correct, but I don't think it matters for it to be correct rather than *feel* okay, which I think the current one does (it kind of works like a ~300ms transform right now). The value that we really care about correctness-wise (cumulative rotation, as it affects scoring) is not damped and therefore needs no adjustment.

The reason why the wait steps had no effect is that the seeks prevent time from elapsing, and therefore the rotation stays frozen. This is consistent with every other gameplay element, so I'd say it's expected behaviour.

# Remarks

I went back and forth on whether to lower the damp, but decided against it. If you think that would be better then I'll close.

[Here's a spreadsheet](https://docs.google.com/spreadsheets/d/14enVlvJ2DfqAstUPtk9MSWyJOzgDJxBdgLhTR6ER2Yc/edit?usp=sharing) that shows how the current damp works. Probably the most sketchy-looking case is in the "backward to halfway" sheet but I think it's ok.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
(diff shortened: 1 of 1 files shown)


peppy:
Without looking into this, is there a reason we can't base the tests on `CumulativeRotation`, which doesn't have any interpolation applied?

peppy:
Never mind, I can see that is *also* being tested. In which case I think your change is satisfactory.

peppy (approved):
Will run this a few times to ensure we are in a good state.

bdach:
I also briefly considered exposing the private `SpinnerRotationTracker.currentRotation` (which is the signed "ground-truth" value), but decided against it because doesn't seem all that necessary - just mostly need the rotation in the right ballpark. (Also this way we're testing the closest thing exposed to the user.)

PR #9759, opened 2020-08-03
Fix NewsOverlay running request on startup

### LINKED ISSUE #9760: News overlay doing web request before requested

EVAST9919:
Same as for DashboardOverlay. Probably should be a thing for every online overlay in the future.

- Closes #9760


osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -67,7 +67,26 @@ private void load()
         protected override void LoadComplete()
         {
             base.LoadComplete();
-            article.BindValueChanged(onArticleChanged, true);
+            article.BindValueChanged(onArticleChanged);
```
frenzibyte, line 70: should probably be renamed to `updateDisplay(), then `PopIn` would call it. better than article change trigger.

osu.Game/Overlays/NewsOverlay.cs
```diff
@@ -67,7 +67,26 @@ private void load()
...
+            {
+                article.TriggerChange();
+                displayUpdateRequired = false;
+            }
+        }
+
+        protected override void PopOutComplete()
+        {
+            base.PopOutComplete();
+            displayUpdateRequired = true;
```
peppy, line 89: this going to trigger a reload every time? is that intentional?

EVAST9919: Yes, because we always want to see latest news every time we open the overlay

peppy: We don't do this in other overlays, is there any reason to only do it here? And with no limit on how often? If someone spams the button it's going to send tens of requests per second.

I'd suggest not adding such behaviour until we have a refresh button, or some kind of debounce logic.

Joehuu: https://github.com/ppy/osu/blob/8a9036938f30da45e3fa37b3446a636abad45cec/osu.Game/Overlays/DashboardOverlay.cs#L96

You can't spam the button as the overlay has to close completely (i.e. pop out complete).

EVAST9919: And also we do the same thing for friends overlay

peppy: Fair enough.

peppy (approved):

PR #9763, opened 2020-08-03
Add vertical offset to osu! playfield to match stable

### LINKED ISSUE #9457: Storyboard is not aligned perfectly to the playfield
**Describe the bug:** 
circle size works differently in lazer than in stable, causing issues with maps that use storyboard for fake circles (disconnected hardkore, aleph-0, storyboard-based aspire maps)

Screen Shot 2020-07-08 at 12 09 03 PM
From Aspire V's "RiraN - Unshakable". The slider that the cursor is on is real, as is the one fading in on the left, but the one in the bottom right is not.

**osu!lazer version:** 
2020.707.0

Wieku:
Fixes #9457

### DIFF
--- a/osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
+++ b/osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
@@ -11,16 +11,17 @@ namespace osu.Game.Rulesets.Osu.UI
     public class OsuPlayfieldAdjustmentContainer : PlayfieldAdjustmentContainer
     {
         protected override Container<Drawable> Content => content;
-        private readonly Container content;
+        private readonly ScalingContainer content;
 
         private const float playfield_size_adjust = 0.8f;
 
         /// <summary>
-        /// Whether an osu-stable playfield offset should be applied (8 osu!pixels)
+        /// When true, an offset is applied to allow alignment with historical storyboards displayed in the same parent space.
+        /// This will shift the playfield downwards slightly.
         /// </summary>
-        public bool PlayfieldShift
+        public bool AlignWithStoryboard
         {
-            set => ((ScalingContainer)content).PlayfieldShift = value;
+            set => content.PlayfieldShift = value;
         }
 
         public OsuPlayfieldAdjustmentContainer()
@@ -65,7 +66,7 @@ namespace osu.Game.Rulesets.Osu.UI
                 // Scale = 819.2 / 512
                 // Scale = 1.6
                 Scale = new Vector2(Parent.ChildSize.X / OsuPlayfield.BASE_SIZE.X);
-                Position = new Vector2(0, (PlayfieldShift ? 8f : 0f) * Parent.ChildSize.Y / OsuPlayfield.BASE_SIZE.Y);
+                Position = new Vector2(0, (PlayfieldShift ? 8f : 0f) * Scale.X);
                 // Size = 0.625
                 Size = Vector2.Divide(Vector2.One, Scale);
             }
--- a/osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
+++ b/osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
@@ -15,6 +15,14 @@ namespace osu.Game.Rulesets.Osu.UI
 
         private const float playfield_size_adjust = 0.8f;
 
+        /// <summary>
+        /// Whether an osu-stable playfield offset should be applied (8 osu!pixels)
+        /// </summary>
+        public bool PlayfieldShift
+        {
+            set => ((ScalingContainer)content).PlayfieldShift = value;
+        }
+
         public OsuPlayfieldAdjustmentContainer()
         {
             Anchor = Anchor.Centre;
@@ -39,6 +47,8 @@ namespace osu.Game.Rulesets.Osu.UI
         /// </summary>
         private class ScalingContainer : Container
         {
+            internal bool PlayfieldShift { get; set; }
+
             protected override void Update()
             {
                 base.Update();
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
```diff
@@ -55,6 +55,7 @@ protected override void Update()
                 // Scale = 819.2 / 512
                 // Scale = 1.6
                 Scale = new Vector2(Parent.ChildSize.X / OsuPlayfield.BASE_SIZE.X);
+                Position = new Vector2(0, 8 * Parent.ChildSize.Y / OsuPlayfield.BASE_SIZE.Y);
```
smoogipoo, line 58: This should probably have a disable for the editor, as stable does. Exposing it as a property seems to be the easiest way to do so and then setting it in `DrawableOsuEditRuleset`.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
Wieku: I completely forgot that this also affects the editor.
Do you mean as a `PlayfieldShift` property with `8` set as default which is then set to 0 by `DrawableOsuEditRuleset`?

smoogipoo: Yeah, that would be fine.

peppy: for future proofing, how would this work with the design mode in editor?

smoogipoo: Does design mode turn this off somehow? Or is the exclusion for the editor just because of the menu bar (in which case this probably shouldn't exist)?

peppy: as in we want the storyboard to align correctly in design mode too. so i'm not sure it should ever be turned off?

smoogipoo: Yeah that makes sense. Let's go with this as is then.

Wieku: So maybe better solution is to use bool `PlayfieldShift` as a property? It will be better than passing arbitrary values to the object. I think we want to turn on shift in design mode because aligning storyboard objects to hitobjects in osu-stable editor is a huge pain.

peppy: If we're looking to fix historically awkward offsets, this could potentially be applied at the legacy storyboard parsing step (moving all elements to their correct location) rather than adding local offsets.

I'd ask that we don't move too quickly on merging this and actually get it right.

smoogipoo: That was brought up in the original thread, with the issue that it creates a black border at the bottom of the screen.

smoogipoo: Am I missing something? stable applies this offset to gameplay too. So if we apply the same offset here, and change nothing about the editor and keep the offset, then things will align correctly and the saved positions would be post-offset so that gameplay should still match up when imported into stable?

peppy: if we are comfortable with keeping the offset as-is going forward then yes, applying at the playfield is the correct solution. given it's something like 12 "gameplay" pixels i'm not sure we want to, but maybe more effort to break compatibility than keep it at this point?

Wieku:
I made it to be only turned on in DrawableOsuRuleset because leaving this always on lead to misaligned compose tab:
osu_2020-08-03_13-40-21

bdach:
Not sure doing this just for the osu! ruleset is correct. Wouldn't this mean that the storyboards would remain visually broken when playing converts in other game modes?

Wieku:
This mostly affects maps which storyboards are made with osu! ruleset in mind so regardless of that shift those storyboards would be broken on converts anyway.

osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
```diff
@@ -55,6 +65,7 @@ protected override void Update()
                 // Scale = 819.2 / 512
                 // Scale = 1.6
                 Scale = new Vector2(Parent.ChildSize.X / OsuPlayfield.BASE_SIZE.X);
+                Position = new Vector2(0, (PlayfieldShift ? 8f : 0f) * Parent.ChildSize.Y / OsuPlayfield.BASE_SIZE.Y);
```
peppy, line 68: Not a huge fan of this being done in Update but I guess it's in line with what is already there.

osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
```diff
@@ -15,6 +15,14 @@ public class OsuPlayfieldAdjustmentContainer : PlayfieldAdjustmentContainer
 
         private const float playfield_size_adjust = 0.8f;
 
+        /// <summary>
+        /// Whether an osu-stable playfield offset should be applied (8 osu!pixels)
```
peppy, line 19: This should mention the reasoning for applying the offset (ie. for storyboard alignment).

Also we have not used the term "osu!pixels" anywhere else so please don't introduce it here. The value doesn't even need to be mentioned in the xmldoc because the description of what it does should be enough to satisfy someone setting this externally.

peppy (dismissed):

osu.Game.Rulesets.Osu/UI/OsuPlayfieldAdjustmentContainer.cs
Wieku: So reword it to something like this?
```
Whether an osu-stable playfield offset should be applied.
Set to true if correct alignment with storyboards is needed.
```

peppy: I'd rename the property to `AlignWithStoryboard`.

"When true, an offset is applied to allow alignment with historical storyboards displayed in the same parent space. This will shift the playfield downwards slightly."

Wieku: Yeah, that sounds much better

bdach (approved):
seems good :+1:

smoogipoo (approved):
lgtm

PR #9766, opened 2020-08-03
Disable activation of overlays while in gameplay

### LINKED ISSUE #9743: Add an option to disable "ctr + key actions" in game

Game4all:
Closes #9743
Alternative to / supersedes  #9751

# Summary

This adds the ability to disable activation of overlays while in gameplay through a setting that defaults to true. Overlay activation is re-enabled back if the gameplay is paused, enters a break period or the current gameplay is running a replay.


osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
```diff
@@ -0,0 +1,54 @@
...
+using osu.Game.Rulesets;
+
+namespace osu.Game.Tests.Visual.Gameplay
+{
+    public class TestSceneOverlayActivation : OsuPlayerTestScene
+    {
+        private OverlayTestPlayer testPlayer;
+
+        [Resolved]
+        private OsuConfigManager mng { get; set; }
```
bdach, line 17: let's use proper variable names please

```suggestion
        private OsuConfigManager config { get; set; }
```

(+ usages)

osu.Game/Screens/Play/Player.cs
```diff
@@ -85,6 +90,8 @@ public class Player : ScreenWithBeatmapBackground
 
         private SkipOverlay skipOverlay;
 
+        protected readonly Bindable<OverlayActivation> OverlayActivationMode = new Bindable<OverlayActivation>(OverlayActivation.Disabled);
```
bdach, line 93: seeing that you're binding to `OsuGame`'s bindable bi-directionally, there's no need to expose this for the test - just resolve `OsuGame` in the test itself, get a bound copy from it and inspect that - it should work the same

osu.Game/Screens/Play/Player.cs
```diff
@@ -197,6 +205,19 @@ private void load(AudioManager audio, OsuConfigManager config)
...
+
+            gameplayOverlaysDisabled.ValueChanged += disabled =>
+            {
+                if (DrawableRuleset.HasReplayLoaded.Value)
+                    OverlayActivationMode.Value = OverlayActivation.UserTriggered;
+                else
+                    OverlayActivationMode.Value = disabled.NewValue && !DrawableRuleset.IsPaused.Value ? OverlayActivation.Disabled : OverlayActivation.UserTriggered;
+            };
+            DrawableRuleset.IsPaused.BindValueChanged(_ => gameplayOverlaysDisabled.TriggerChange());
+            DrawableRuleset.HasReplayLoaded.BindValueChanged(_ => gameplayOverlaysDisabled.TriggerChange());
```
bdach, line 219: this is super awkwardly written to me; if you wanted to centralise logic of changing `OverlayActivationMode` I would suggest to follow the example of `updatePauseOnFocusLostState()` below by extracting the callback to a `updateOverlayActivationMode()` method, and then using it in the callbacks of all these three bindables - this way there's no weird daisy-chaining of value changed callbacks

osu.Game/Configuration/OsuConfigManager.cs
```diff
@@ -100,6 +100,7 @@ protected override void InitialiseDefaults()
 
             Set(OsuSetting.IncreaseFirstObjectVisibility, true);
             Set(OsuSetting.GameplayDisableWinKey, true);
+            Set(OsuSetting.GameplayDisableOverlayActivation, true);
```
bdach, line 103: seeing that this is given a default of `true` this tip might need wording adjustment

https://github.com/ppy/osu/blob/fd56db21568a7c71e4390a4b20a69e0922538a88/osu.Game/Screens/Menu/Disclaimer.cs#L193

bdach (dismissed):

osu.Game/Screens/Play/Player.cs
Game4all: While I agree this should have been the way to do things, this won't work since trying to resolve `OsuGame` in game testing code returns null ( OsuTestBrowser derives from `OsuGameBase`), hence the existence of this protected bindable.

bdach: fair enough, would have been good to mention it at the start - this however means that this needs xmldoc

also a tiny bit confused at why this is specifying `Disabled` as its initial value but there is also this

https://github.com/ppy/osu/blob/25ebb8619d2cbaa6fc232ed742d33da91b80d703/osu.Game/Screens/Play/Player.cs#L53

Game4all: Bindable initial value should be the same as `InitialOverlayActivationMode` of the screen, I missed that one.

osu.Game/Configuration/OsuConfigManager.cs
Game4all: IMO, I don't think the tip needs to be adjusted. I mean, the tip is still half valid in the sense that you can change the skin while playing but you'll have to pause to open the settings and switch skins, but I guess that's what most of players (to not say all) would do.

osu.Game/Screens/Play/Player.cs
```diff
@@ -341,6 +359,14 @@ private void onBreakTimeChanged(ValueChangedEvent<bool> isBreakTime)
             HUDOverlay.KeyCounter.IsCounting = !isBreakTime.NewValue;
         }
 
+        private void updateOverlayActivationMode()
+        {
+            if (DrawableRuleset.HasReplayLoaded.Value)
+                OverlayActivationMode.Value = OverlayActivation.UserTriggered;
+            else
+                OverlayActivationMode.Value = gameplayOverlaysDisabled.Value && !DrawableRuleset.IsPaused.Value ? OverlayActivation.Disabled : OverlayActivation.UserTriggered;
```
bdach, line 367: can't say I like the way this is written too much, feels like too much branching; would probably rewrite to
```csharp
bool canTriggerOverlaysDuringGameplay = DrawableRuleset.IsPaused.Value || !gameplayOverlaysDisabled.Value;
if (DrawableRuleset.HasReplayLoaded.Value || canTriggerOverlaysDuringGameplay)
    OverlayActivationMode.Value = OverlayActivation.UserTriggered;
else
    OverlayActivationMode.Value = OverlayActivation.Disabled;
```

feels more legible that way to me

osu.Game/Configuration/OsuConfigManager.cs
bdach: mostly I'd just mention that pausing is necessary to avoid people claiming that the tip is incorrect:

> All settings are dynamic and take effect in real-time. Try pausing and changing the skin while playing!

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
```diff
@@ -0,0 +1,54 @@
...
+
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Game.Configuration;
+using osu.Game.Overlays;
+using osu.Game.Rulesets;
+
+namespace osu.Game.Tests.Visual.Gameplay
+{
+    public class TestSceneOverlayActivation : OsuPlayerTestScene
```
bdach, line 12: would also like a test case for allowing the overlays to be opened by disabling the setting here for completeness

Game4all: In fact, I had a test to ensure that overlays can be opened when the setting is disabled, but I ended up not committing it as for some reason the setting update doesn't propagate to the loaded test player (whereas this works just fine in game), making the test always fail.

bdach: That is quite unsettling. I'd like to know why exactly it fails in the test.

Game4all: For reference the test was:
```csharp
        [Test]
        public void TestGameplayOverlayActivationSettingDisabled()
        {
            AddStep("enable overlay activation during gameplay", () => config.Set(OsuSetting.GameplayDisableOverlayActivation, false));
            AddAssert("activation mode is user triggered", () => testPlayer.OverlayActivationMode == OverlayActivation.UserTriggered);
        }
```
and would fail on the assertion of the overlay activation mode (always returning `OverlayActivationMode.Disabled` whether it should be `OverlayActivationMode.UserTriggered`).
I tried replacing with an `AddUntil` step but I keep getting the same results.

It looks like the setting update doesn't get propagated to the test player.

You can easily replicate by copy pasting into the test cases file.

bdach: the test doesn't work because it's using the wrong instance of `OsuConfigManager`. `PlayerTestScene` creates and caches its own:

https://github.com/ppy/osu/blob/c09868780e1d03703a18f37e512777fba3afe89c/osu.Game/Tests/Visual/PlayerTestScene.cs#L25-L32

using that one instead of the locally-`[Resolved]` config manager fixes the failure.

on that note I'd also like to say that I find the mentality of discarding the test if it doesn't work correctly quite dangerous. we shouldn't become accustomed to accepting that things just don't work sometimes for unknown reasons. if I hadn't prodded continuously reviewing this, I probably wouldn't have found out that the tests were half-broken due to using the wrong config instance, especially seeing that they *read* sane.

(also not sure how to feel about having to debug this myself, but, alas.)

Game4all: I apologize for having left the debugging of this to you. I feel really bad for doing it, especially when the cause is that I didn't paid attention enough to the existing code.

bdach:
I've pushed in a menial change that aims to avoid having `OverlayActivation.UserTriggered` specified in two places (now they both refer to a single constant). Unfortunately this can't be done without introducing the constant.

Otherwise this seems ok. In a quick gameplay test it works roughly as expected - the only issue is that the hotkeys make a click sound even if disabled, but that should probably be fixed globally as I imagine it'll be a global issue and is probably out of scope here.

Not approving since I pushed.

frenzibyte (changes requested):
The current logic looks fine to me for now otherwise, but going forward, I would suggest making `OverlayActivationMode` immutable from any component, and letting `OsuGame` get the overlay activation mode of the current active screen or otherwise, this is to avoid changing `OverlayActivationMode` in the wrong time, in this case, while the screen isn't yet pushed as caught below.

---

Also, there's another issue arisen because of this first time using `OverlayActivationMode.Disabled`, toggling the toolbar by `Ctrl+T` displays it frozen: https://streamable.com/rllzwd. 

Given that these small issues currently exist, and that this PR doesn't really need to be merged as soon as possible, maybe better to fix these issues first before merging this.

osu.Game/Screens/Menu/Disclaimer.cs
```diff
@@ -190,7 +190,7 @@ private string getRandomTip()
             {
                 "You can press Ctrl-T anywhere in the game to toggle the toolbar!",
                 "You can press Ctrl-O anywhere in the game to access options!",
-                "All settings are dynamic and take effect in real-time. Try changing the skin while playing!",
+                "All settings are dynamic and take effect in real-time. Try pausing and changing the skin while playing!",
```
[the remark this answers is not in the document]
bdach: that's not really proper English

osu.Game/Screens/Play/Player.cs
```diff
@@ -197,6 +209,13 @@ private void load(AudioManager audio, OsuConfigManager config)
                 skipOverlay.Hide();
             }
 
+            if (game != null)
+                OverlayActivationMode.BindTo(game.OverlayActivationMode);
+
+            gameplayOverlaysDisabled.BindValueChanged(_ => updateOverlayActivationMode());
+            DrawableRuleset.IsPaused.BindValueChanged(_ => updateOverlayActivationMode());
+            DrawableRuleset.HasReplayLoaded.BindValueChanged(_ => updateOverlayActivationMode());
```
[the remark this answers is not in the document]
bdach: sounds like we might want to take a lease on the bindable instead of directly binding

osu.Game/Screens/Menu/Disclaimer.cs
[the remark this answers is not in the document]
frenzibyte: of course, that's what the **maybe** is for, I'm pointing that there's kind of an oxymoron here that shouldn't really exist in such place. (or anywhere)

frenzibyte:
Oh, and also this should allow overlay activation on break times as well, it's a kind of a normal habit for a player to open the chat overlay and talk there in such periods.

osu.Game/Screens/Play/Player.cs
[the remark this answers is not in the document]
frenzibyte: oh just realized I didn't mention this, but the settings overlay closing is the result of me setting `Disable overlays during gameplay` on, I'm suggesting to move these binds inside `OnEntering()` where it makes sense to start allowing to change the `OverlayActivationMode`.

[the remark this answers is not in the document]
Game4all: Alternatively, adding a check that `Player` is the current screen should do the trick aswell.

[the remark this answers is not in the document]
frenzibyte: Oh, now that you mention it, this'll also change `OverlayActivationMode` on the results screen, tbh I think moving all this overlay activation mode logic into an `OsuGame` level would be a better choice as I mentioned above https://github.com/ppy/osu/pull/9766#pullrequestreview-463824776 (in a separate PR this would depend on)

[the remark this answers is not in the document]
bdach: taking a lease on the bindable would also work as desired by rolling the changes back, and require no other changes.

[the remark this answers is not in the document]
Game4all: > 
> 
> Oh, now that you mention it, this also changes `OverlayActivationMode` on the results screen, tbh I think moving all this overlay activation mode logic into an `OsuGame` level would be a better choice as I mentioned above [#9766 (review)](https://github.com/ppy/osu/pull/9766#pullrequestreview-463824776) (in a separate PR this would depend on)
> 
> Edit: Adding the `Player` check would fix this as well but it would just make the whole thing look like it shouldn't be there.

Did some debugging as this sounded quite weird to me. Turns out that the the behaviour described in results screen is
because of this: https://github.com/ppy/osu/blob/9a09f97478f063384a03c0d35d23f359cc9d1353/osu.Game/Screens/Ranking/ResultsScreen.cs#L33

image

As seen on the screen, the activation mode is `All`, which is the default activation mode of an `OsuScreen`. Would've been either `UserTriggered` or `Disabled` if  `Player` was to change the `OverlayActivationMode`.

osu.Game/Screens/Play/Player.cs
```diff
@@ -362,7 +363,10 @@ private void onBreakTimeChanged(ValueChangedEvent<bool> isBreakTime)
 
         private void updateOverlayActivationMode()
         {
-            bool canTriggerOverlays = DrawableRuleset.IsPaused.Value || !gameplayOverlaysDisabled.Value;
+            if (!this.IsCurrentScreen())
+                return;
```
bdach, line 367: why is this check required?

Game4all: This check has been added to prevent `OverlayActivationMode` from being updated when the `Player` isn't the current screen (could be triggered previously by changing the value of the setting while in the player loader, according to this comment: https://github.com/ppy/osu/pull/9766#pullrequestreview-463824776)

bdach: if so then I think I would prefer deferring setting up the subscriptions until `OnEntering()` as proposed above.

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
```diff
@@ -0,0 +1,68 @@
...
+        public override void SetUpSteps()
+        {
+            AddStep("disable overlay activation during gameplay", () => LocalConfig.Set(OsuSetting.GameplayDisableOverlayActivation, true));
+            base.SetUpSteps();
+        }
+
+        [Test]
+        public void TestGameplayOverlayActivation()
+        {
+            AddAssert("activation mode is disabled", () => testPlayer.OverlayActivationMode == OverlayActivation.Disabled);
```
peppy, line 25: why are the tests only covering the disabled case?

osu.Game/Screens/Play/Player.cs
```diff
@@ -50,7 +50,13 @@ public class Player : ScreenWithBeatmapBackground
...
         public override bool HideOverlaysOnEnter => true;
 
-        public override OverlayActivation InitialOverlayActivationMode => OverlayActivation.UserTriggered;
+        private const OverlayActivation initial_overlay_activation_mode = OverlayActivation.UserTriggered;
+        public override OverlayActivation InitialOverlayActivationMode => initial_overlay_activation_mode;
+
+        /// <summary>
+        /// The current activation mode for overlays.
+        /// </summary>
+        protected readonly Bindable<OverlayActivation> OverlayActivationMode = new Bindable<OverlayActivation>(initial_overlay_activation_mode);
```
peppy, line 59: exposing this is really confusing. i'd hope it was private.

osu.Game/Screens/Play/Player.cs
```diff
@@ -627,6 +652,11 @@ public override void OnEntering(IScreen last)
 
             foreach (var mod in Mods.Value.OfType<IApplicableToHUD>())
                 mod.ApplyToHUD(HUDOverlay);
+
+            DrawableRuleset.IsPaused.BindValueChanged(_ => updateOverlayActivationMode());
```
peppy, line 656: i'm pretty sure the state is not guaranteed to stop mutating when exiting player. probably needs a call to `updateOverlayActivationMode()` in exiting.

osu.Game/Screens/Menu/Disclaimer.cs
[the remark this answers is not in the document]
peppy: I don't really think this needed to change.

peppy (changes requested):

peppy:
Honestly, with the amount of complexity here, I wonder if this even needs to be a setting? Is there a use case for setting the setting to false?

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
bdach: The next test under this one covers disabling the setting.

peppy: ah, i see. got confused by the `SetUpStep` setting one state. maybe not the best way to structure tests (move the set into each test instead, or via `[TestCase(bool)]`)

osu.Game/Screens/Menu/Disclaimer.cs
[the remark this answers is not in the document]
bdach: maybe not... I'm just super not looking forward to potentially having someone claim that the tip is wrong because "settings menu can't be opened during gameplay"

osu.Game/Screens/Play/Player.cs
Game4all: This bindable is marked as protected only for testing purposes, as we can't resolve an `OsuGame` instance through DI in tests. (see this https://github.com/ppy/osu/pull/9766#discussion_r464548375 for reference)

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
Game4all: I pushed a commit to make the test steps hopefully clearer

osu.Game/Screens/Play/Player.cs
Game4all: I'm not sure at all with that one. Isn't the updating of the overlay mode with the change of screens sufficient?

frenzibyte: i.e. enabling the "disable overlays during gameplay" setting while in results screen will affect the global value, because `Player` still exists, just suspended.

frenzibyte:
About the first paragraph in https://github.com/ppy/osu/pull/9766#pullrequestreview-463824776, here's what I think of how should `OverlayActivationMode` be mutable (with `OsuGame` itself being responsible for it): https://github.com/frenzibyte/osu/compare/1398234b1bc54fe2e09158ef7aaac929b7209ec8...6285cd4b108985c02461ddf58590745e89d105b7

That would get rid of any unorganized insurances inside `Player` (binding at `OnEntering`, ensuring `OverlayActivationMode` isn't change while not in current screen), would like to hear if it's going the right direction before PRing though.

peppy:
Changing InitialOverlayActivation to just be a screen bindable and is one-way bound with the game one should work amicably.

osu.Game/Screens/Play/Player.cs
Game4all: ah it's again that case. Guess I'll wait for your proposed changes to get merged and rebase, as it'll fix that case (aswell as the binding on entering)

Game4all:
> 
> Honestly, with the amount of complexity here, I wonder if this even needs to be a setting? Is there a use case for setting the setting to false?

I'm fine with dropping the setting and having this as a default behaviour, as it would also resolve the issue where the overlay activation mode would change because the setting value was updated despite the player being suspended.

smoogipoo (approved):
Seems pretty fine to me. I'm also not 100% sure on the disclaimer change, but I think the concern pointed out by @bdach is fairly reasonable.

PR #9768, opened 2020-08-03
Allow providing custom sprite text for RollingCounter<T>

frenzibyte:
Required for catch-specific combo counter as it uses `RollingCounter<T>`, and the sprite text of the counter for legacy skins must be a `LegacySpriteText` of course.

### DIFF
--- a/osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
+++ b/osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
 using osu.Game.Graphics.UserInterface;
 using osu.Game.Tournament.IPC;
 using osu.Game.Tournament.Models;
@@ -127,20 +128,27 @@ namespace osu.Game.Tournament.Screens.Gameplay.Components
 
         private class MatchScoreCounter : ScoreCounter
         {
+            private OsuSpriteText displayedSpriteText;
+
             public MatchScoreCounter()
             {
                 Margin = new MarginPadding { Top = bar_height, Horizontal = 10 };
-
-                Winning = false;
-
-                DisplayedCountSpriteText.Spacing = new Vector2(-6);
             }
 
             public bool Winning
             {
-                set => DisplayedCountSpriteText.Font = value
+                set => displayedSpriteText.Font = value
                     ? OsuFont.Torus.With(weight: FontWeight.Bold, size: 50, fixedWidth: true)
                     : OsuFont.Torus.With(weight: FontWeight.Regular, size: 40, fixedWidth: true);
+            }
+
+            protected override OsuSpriteText CreateSpriteText()
+            {
+                displayedSpriteText = base.CreateSpriteText();
+                displayedSpriteText.Spacing = new Vector2(-6);
+                Winning = false;
+
+                return displayedSpriteText;
             }
         }
     }
--- a/osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
+++ b/osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
@@ -137,19 +137,20 @@ namespace osu.Game.Tournament.Screens.Gameplay.Components
 
             public bool Winning
             {
-                set => displayedSpriteText.Font = value
+                set => updateFont(value);
+            }
+
+            protected override OsuSpriteText CreateSpriteText() => base.CreateSpriteText().With(s =>
+            {
+                displayedSpriteText = s;
+                displayedSpriteText.Spacing = new Vector2(-6);
+                updateFont(false);
+            });
+
+            private void updateFont(bool winning)
+                => displayedSpriteText.Font = winning
                     ? OsuFont.Torus.With(weight: FontWeight.Bold, size: 50, fixedWidth: true)
                     : OsuFont.Torus.With(weight: FontWeight.Regular, size: 40, fixedWidth: true);
-            }
-
-            protected override OsuSpriteText CreateSpriteText()
-            {
-                displayedSpriteText = base.CreateSpriteText();
-                displayedSpriteText.Spacing = new Vector2(-6);
-                Winning = false;
-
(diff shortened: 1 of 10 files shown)


frenzibyte:
Also, the reason `CreateSpriteText()` isn't called in the constructor is to allow passing data for the custom sprite text, like in my case, `LegacySpriteText` is only constructible by `new LegacySpriteText(skin, fontName)`, thus the following must be applied:
```csharp
private class LegacyRollingCounter : RollingCounter<int>
{
    public LegacyRollingCounter(ISkin skin, string fontName)
    {
        this.skin = skin;
        this.fontName = fontName;
    }

    protected override OsuSpriteText CreateSpriteText() => new LegacySpriteText(skin, fontName);
}
```

osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
```diff
@@ -127,21 +128,28 @@ protected override void UpdateAfterChildren()
...
             }
+
+            protected override OsuSpriteText CreateSpriteText()
+            {
+                displayedSpriteText = base.CreateSpriteText();
+                displayedSpriteText.Spacing = new Vector2(-6);
+                Winning = false;
+
+                return displayedSpriteText;
+            }
```
bdach, line 152: a little verbose; can't this just be something of this sort?

```csharp
protected override OsuSpriteText CreateSpriteText() => displayedSpriteText = base.CreateSpriteText().With(text =>
{
    text.Spacing = new Vector2(-6);

    Winning = false;
});
```

osu.Game/Graphics/UserInterface/RollingCounter.cs
```diff
@@ -46,57 +47,71 @@ public abstract class RollingCounter<T> : Container, IHasAccentColour
...
-        protected override void LoadComplete()
+        [BackgroundDependencyLoader]
+        private void load()
         {
-            base.LoadComplete();
-
-            DisplayedCountSpriteText.Text = FormatCount(Current.Value);
+            displayedCountSpriteText = CreateSpriteText();
+            displayedCountSpriteText.Text = FormatCount(displayedCount);
+            Child = displayedCountSpriteText;
```
bdach, line 114: really confused as to why `displayedCount` and `DisplayedCount` are used interchangeably here - would argue that only the property should access its backing field directly and everything else should use the getter

osu.Game/Graphics/UserInterface/RollingCounter.cs
```diff
@@ -46,57 +47,71 @@ public abstract class RollingCounter<T> : Container, IHasAccentColour
...
+            set
+            {
+                if (TextSize == value)
+                    return;
+
+                textSize = value;
+                if (displayedCountSpriteText != null)
+                    displayedCountSpriteText.Font = displayedCountSpriteText.Font.With(size: value);
+            }
         }
```
bdach, line 77: I sorta get why this is necessary (`DisplayedCountSpriteText` is no longer guaranteed to be initialised at call time) but this is almost too ugly to let live in its current state IMO - this looks like it's not encapsulated well enough

maybe something like a `SpriteTextFactory` (a dumb object that can create a sprite text on demand) is something we might want here - unsure whether that's too enterprise™ or if it fits without seeing what the actual use case is for the legacy one properly

osu.Game/Graphics/UserInterface/RollingCounter.cs
frenzibyte: whoops I forgot checking this

osu.Game.Tournament/Screens/Gameplay/Components/MatchScoreDisplay.cs
frenzibyte: to work properly, it must be:
```csharp
protected override OsuSpriteText CreateSpriteText() => (displayedSpriteText = base.CreateSpriteText()).With(text =>
{
     text.Spacing = new Vector2(-6);
     Winning = false;
}
```
which to me, tbh doesn't sound much beneficial than the original one.

bdach: this one maybe not too much as it assigns in the middle of it but the others I'd argue would benefit (didn't comment on those as not to hammer it home too much)

osu.Game/Graphics/UserInterface/RollingCounter.cs
```diff
@@ -46,57 +47,71 @@ public abstract class RollingCounter<T> : Container, IHasAccentColour
...
 
         public abstract void Increment(T amount);
 
+        private float textSize = 40f;
+
         public float TextSize
         {
-            get => DisplayedCountSpriteText.Font.Size;
-            set => DisplayedCountSpriteText.Font = DisplayedCountSpriteText.Font.With(size: value);
+            get => displayedCountSpriteText?.Font.Size ?? textSize;
```
peppy, line 67: why not just remove all this complexity and leave it to the overriding of `CreateSpriteText` to adjust it? is this used at runtime anywhere?

frenzibyte: `HUDOverlay` uses that on its counter components (combo, score, etc..), though even after removing that there's also `AccentColour` which still needs to exist.

peppy: i can't read your response

frenzibyte: So, there are `TextSize` usages in `HUDOverlay`:
https://github.com/ppy/osu/blob/96c4a30facc6bd2b0cdc8415852624e4a8e2c835/osu.Game/Screens/Play/HUDOverlay.cs#L235
https://github.com/ppy/osu/blob/96c4a30facc6bd2b0cdc8415852624e4a8e2c835/osu.Game/Screens/Play/HUDOverlay.cs#L244
https://github.com/ppy/osu/blob/96c4a30facc6bd2b0cdc8415852624e4a8e2c835/osu.Game/Screens/Play/HUDOverlay.cs#L251

they're just initial text sizes so I can move them to inside their `CreateSpriteText()` method.

As for `AccentColour`, I guess it can be removed for now as it's not used anywhere and likely won't be.

peppy: Sounds good.

bdach (approved):
I think this is fine - not too sure about the multiplicative `Colour` usage replacing `AccentColour` but I suppose if that ever goes wrong it'll go wrong very visibly

PR #9769, opened 2020-08-03
Add osu!catch-specific combo counter support

### LINKED ISSUE #5103: Add option for combo following catcher
In stable combo is counted right above the ryuuta/center screen ish and moves back and forth (left/right) as you move the ryuuta.  Not having this at least as an option makes ctb play feel very uncomfortable as I cannot easily tell when I've dropped a note or just see how far I've made it etc.  

This is compounded by issue #5102 making it so the only way to tell "did I just miss that?" is by looking at the top of the screen to see the combo count, which is not possible in the middle of playing any fast/difficult song

frenzibyte:
Closes #5103 
- [x] Depends on #9768 

Adds support for displaying osu!catch-specific combo counter, with a legacy implementation **almost** matching osu!stable's implementation in terms of what's and how's displayed. 

Implemented this by adding a `CatchComboDisplay` component handling new and reverted judgements and is responsible of updating the skinned counter accordingly.

here's a comparsion GIF to save everyone's day 😛:

([source video for better per-frame look](https://streamable.com/5aac1q))

## Remarks

- Couldn't exactly match the way osu!stable rolls down and fades its counter on miss, here's how it currently looks:
   
- Intentionally added "hide on gameplay break" logic inside `LegacyComboCounter` instead of its parent `CatchComboDisplay` to avoid conflicting with the "hide after 1 second" transform, but can be moved if osu!lazer's combo counter animation design indicates to.
- This PR may be a bit complex for review, commit-by-commit should hopefully be better in doing so. didn't want to split this into very small untestable changes.


osu.Game/Skinning/LegacySkinTransformer.cs
```diff
@@ -47,5 +47,7 @@ public virtual SampleChannel GetSample(ISampleInfo sampleInfo)
         }
 
         public abstract IBindable<TValue> GetConfig<TLookup, TValue>(TLookup lookup);
+
+        protected bool HasFont(string fontPrefix) => GetTexture($"{fontPrefix}-0") != null;
```
bdach, line 51: think this is better suited to something like `LegacySkinExtensions`

osu.Game/Tests/Visual/SkinnableTestScene.cs
```diff
@@ -152,25 +153,38 @@ private class TestLegacySkin : LegacySkin
...
             {
                 // extrapolate frames to test longer animations
-                if (extrapolateAnimations)
-                {
-                    var match = Regex.Match(componentName, "-([0-9]*)");
-
-                    if (match.Length > 0 && int.TryParse(match.Groups[1].Value, out var number) && number < 60)
-                        return base.GetTexture(componentName.Replace($"-{number}", $"-{number % 2}"), wrapModeS, wrapModeT);
-                }
+                if (extrapolateAnimations && isAnimationComponent(componentName, out var number) && number < 60)
```
bdach, line 171: where's this `number < 60` check coming from?

osu.Game/Tests/Visual/SkinnableTestScene.cs
```diff
@@ -152,25 +153,38 @@ private class TestLegacySkin : LegacySkin
...
+            private readonly HashSet<string> legacyFontPrefixes = new HashSet<string>();
+
             public TestLegacySkin(SkinInfo skin, IResourceStore<byte[]> storage, AudioManager audioManager, bool extrapolateAnimations)
                 : base(skin, storage, audioManager, "skin.ini")
             {
                 this.extrapolateAnimations = extrapolateAnimations;
+
+                legacyFontPrefixes.Add(GetConfig<string, string>("HitCirclePrefix")?.Value ?? "default");
+                legacyFontPrefixes.Add(GetConfig<string, string>("ScorePrefix")?.Value ?? "score");
+                legacyFontPrefixes.Add(GetConfig<string, string>("ComboPrefix")?.Value ?? "score");
```
bdach, line 165: can't these be actual lookups rather than magic strings? `ComboPrefix` was added in this PR already

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
```diff
@@ -0,0 +1,156 @@
...
+using osu.Game.Graphics.UserInterface;
+using osu.Game.Rulesets.Catch.UI;
+using osu.Game.Screens.Play;
+using osu.Game.Skinning;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Catch.Skinning
+{
+    internal class LegacyComboCounter : CompositeDrawable, ICatchComboCounter
```
bdach, line 19: if this is in the ruleset assembly already does it make much difference whether it's `public` or `internal`? would just make `public` as we're trying to avoid internals for dynamic compilation reasons

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
```diff
@@ -0,0 +1,156 @@
...
+            // Combo fell to zero, roll down and fade out the counter.
+            if (combo == 0)
+            {
+                counter.Current.Value = 0;
+                if (lastExplosion != null)
+                    lastExplosion.Current.Value = 0;
+
+                this.FadeOut(immediate ? 0.0 : 400.0, Easing.Out);
+                return;
+            }
```
bdach, line 81: Sometimes this path doesn't seem to work so well. It seems that it is possible for the explosion swap to somehow interfere with this, which means that the first explosion after the miss that brought combo down to 0 the old combo count is displayed. See [reproduction video](https://drive.google.com/file/d/1cGS56djY7IiXW2gWCmiSxQX6iZEFTx8C/view?usp=sharing).

Also, shouldn't all this stuff be inside of a transform sequence of some sort?

osu.Game.Rulesets.Catch/UI/CatchComboDisplay.cs
```diff
@@ -0,0 +1,59 @@
...
+
+            updateCombo(result.ComboAtJudgement + 1, judgedObject.AccentColour.Value);
+        }
+
+        public void OnRevertResult(DrawableCatchHitObject judgedObject, JudgementResult result)
+        {
+            if (!result.Judgement.AffectsCombo || !result.HasResult)
+                return;
+
+            updateCombo(result.ComboAtJudgement, judgedObject.AccentColour.Value);
```
bdach, line 50: This doesn't seem to revert completely correctly (in the miss case the judged object's colour is passed instead of `null`).

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using osu.Framework.Graphics;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Catch.UI
+{
+    /// <summary>
+    /// An interface providing a set of methods to update the combo counter.
+    /// </summary>
+    public interface ICatchComboCounter : IDrawable
```
bdach, line 12: Was there some particular intent behind this interface? Right now it seems superfluous as it has a single implementation within the scope of this PR.

osu.Game.Rulesets.Catch.Tests/TestSceneComboCounter.cs
```diff
@@ -0,0 +1,79 @@
...
+            public TestDrawableCatchHitObject(CatchHitObject hitObject)
+                : base(hitObject)
+            {
+                AccentColour.Value = Color4.White;
+            }
+        }
+
+        private class TestCatchHitObject : CatchHitObject
+        {
+        }
```
bdach, line 77: not too sure what these are for

bdach (changes requested):
new public classes need xmldoc in general

osu.Game/Tests/Visual/SkinnableTestScene.cs
frenzibyte: not sure what you mean, `isAnimationComponent(...)` is fulfilling its job, `number < 60` is took from above.

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
frenzibyte: had the habit of marking any legacy class internal, but that's an absolutely good point, will change it back to public.

osu.Game/Tests/Visual/SkinnableTestScene.cs
frenzibyte: didn't make them lookups as they're legacy settings, therefore their names shouldn't really change or something, and making them lookups may not really be worth it, as `HitCirclePrefix` is defined by `OsuSkinConfiguation` in osu!standard ruleset assembly.

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
frenzibyte: Yes, the interface is going to be used for `DefaultComboCounter` as well.

osu.Game.Rulesets.Catch.Tests/TestSceneComboCounter.cs
frenzibyte: Added them to avoid passing null drawable hit objects, guess should allow for now.

frenzibyte:
> new public classes need xmldoc in general

Had been asking myself whether to do it or not, to my knowledge, XMLDoc is 100% required for osu!framework and probably osu!lazer for ruleset consumers (and skin authors in the future).

But in rulesets themselves, and legacy ones to be precise, not sure it's worth the effort. (unless I'm completely missing the point of xmldocs)

osu.Game.Rulesets.Catch.Tests/TestSceneComboCounter.cs
bdach: couldn't you use any of the existing non-abstract classes though?

osu.Game/Tests/Visual/SkinnableTestScene.cs
bdach: fair enough, not sure how I missed this

bdach:
I'd argue that documenting code is always worth it. xmldoc might be necessary here anyway in the long run (if the new drawable-based skinning system happens).

osu.Game.Rulesets.Catch.Tests/TestSceneComboCounter.cs
frenzibyte: hmm, will try them.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
```diff
@@ -48,12 +49,22 @@ public CatchPlayfield(BeatmapDifficulty difficulty, Func<CatchHitObject, Drawabl
                 Origin = Anchor.TopLeft,
             };
 
+            comboDisplay = new CatchComboDisplay
```
peppy, line 52: why is this in the playfield and not part of the legacy catcher?

frenzibyte: `CatcherArea` limits it's height to the catcher height itself, so it's not possible to include this inside there. And to me, it makes sense to put the catch-specific combo counter in the `CatchPlayfield` itself. 

But reading this piece of code again, the only thing I missed thinking about at is how the members in here are very specific to the legacy counter instead of being set to full size and let `LegacyComboCounter` set its size properly, I'll change that.

peppy: Even though it always tracks the catcher position? I'm not so sure about that.

Also just because its high is limited doesn't mean that's an issue unless it's masking.

frenzibyte: I'm making `CatchComboDisplay` an own component not tied directly to catcher because I wasn't sure lazer's combo counter implementation would be the same as legacy's tracking the catcher position.

That was one of the things I was worried about while implementing this, I could cache `CatcherArea` and let `LegacyComboCounter` do the tracking itself if that's what you would like to see removed from here.

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
frenzibyte: Hmm, good catch. Not quite sure what to do here, it sounds feasible to fix this, but it will make things pretty much complicated than now (the transforms to the new combo second pop up must be tracked, then somehow removed when the user miss). 

Thinking about just moving the `FinishTransforms(true)` above and not showing the explosion at all. (also side effect from this is that when missing after a beatmap break, the combo will display immediately and roll down, because `FadeOutFromOne` now must be used).

bdach: I'm not sure how to respond to this comment unless I go and try to fix this myself. A proposal in code with some video would probably be better.

osu.Game/Tests/Visual/SkinnableTestScene.cs
bdach: Sure, I guess this is not worth spending much time on.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
bdach: I think it'd be better (or in other words: less complicated) if it was in `CatcherArea` if feasible. Would save the transfer of `X` in `Update()` at the least (speaking of which, it seems to me that it can be one frame late as done currently). Being outside bounds as long as masking is off is fine with me.

frenzibyte: `CatcherArea` will not get rid of the X transfer though, moving it into `Catcher` will do (which would be far worse given that the catcher component is already complex enough).

bdach: Fair enough, doing it more locally is better anyway in my book.

frenzibyte: I've attempted to move it into `Catcher`, had to add a 300 bottom-margin but that was fine, until I realized that `Catcher` has a logic to change the entire drawing direction of its content based on the player's desired direction, which affects the combo counter and I believe it shouldn't happen. and honestly, it sounds kinda illogical for `Catcher` to isolate its direction changing logic to its *content* only, away from the combo counter, as if the counter shouldn't be inside at all.

Anyways I've pushed a branch for this as my opinion alone is probably not enough to decide, here's a two-dots diff between current commits and my attempt of moving to `Catcher`:  https://github.com/frenzibyte/osu/compare/catch-combo-counter..combo-counter-alternate-way

bdach: You have a point, I'm not 100% sure on `Catcher` any more looking at it... I still wouldn't say `Playfield` is good either, as it's way too general a component to have the counter in - it should be either `CatcherArea` or `Catcher`, preferably the former.

I think the ultimate solution might be to do something like

```
CatcherArea
   Container
       LegacyComboCounter
       Catcher
```

The container would be to make adjusting this:

https://github.com/ppy/osu/blob/25ee1fa7a231c2f2627193da8df957b51b51f783/osu.Game.Rulesets.Catch/UI/CatcherArea.cs#L106-L107

less painful (otherwise you'd have to transfer to both `MovableCatcher` and the counter).

frenzibyte: From a legacy perspective, yes it's too general to be in playfield, but from lazer's perspective which is what I envision, I think it being at the playfield makes sense as an extra HUD thing specific to the ruleset (heck I think moving it to `DrawableCatchRuleset` makes more sense for that).

But I would defer that until an actual design for osu!lazer's default skin exists and the combo counter behavior is decided (for OML skinning and otherwise), alright will move it to `CatcherArea`.

frenzibyte-alt:
As my main account has been locked due to 2FA linked to my broken phone (I apologize on my stupidity there), I don't have push access to this PR, and would be grateful if someone here with write access pushes the following two commits in order by cherry-picking or otherwise:
1. https://github.com/ppy/osu/commit/e1b4e833f717835358bac96988d6c30c818bce65
2. https://github.com/ppy/osu/commit/4f1204ba4298d61d96f2b5c81d877ae2d4c0b4ab

Hope I haven't wasted your time, again I apologize for this.

bdach:
I haven't done this because I'm still not sure whether having it in `CatcherArea` or `Catcher` is better.

peppy:
I definitely think being in the `CatcherArea` is better.

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
```diff
@@ -0,0 +1,162 @@
...
+
+            explosion.SetCountWithoutRolling(combo);
+            explosion.ScaleTo(1.9f, 400.0, Easing.Out)
+                     .FadeOut(400.0)
+                     .Expire(true);
+
+            lastExplosion = explosion;
+        }
+
+        private class LegacyRollingCounter : RollingCounter<int>
```
bdach, line 132: not entirely sure about this being an inner class - I can see uses for this elsewhere (legacy counters in mania and taiko, for one)

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
```diff
@@ -29,6 +29,8 @@ public class CatchPlayfield : ScrollingPlayfield
 
         internal readonly CatcherArea CatcherArea;
 
+        private CatchComboDisplay comboDisplay => CatcherArea.ComboDisplay;
```
bdach, line 32: this should not be dragged here out of the catcher area like this - `GameplayBeatmap`'s `LastJudgementResult` was designed for this sort of use case (albeit right now it seems to lack support for reverting, but I'm not sure that's too much of an issue, can even be done as part of this PR most likely).

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
bdach: I think this is overengineering for one but I guess I can accept it for future usage in other rulesets. Not sure why this interface is catch-specific though; I don't think it really has to be.

bdach (dismissed):

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
frenzibyte: thought about that, but went with leaving it until legacy counters are implemented on the legacy HUD overlay, but sure can move this.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
frenzibyte: I did it like this because the judgement events are in here, and because I didn't want to really pass the judgements through each component's `OnNewResult` and `OnRevertResult` (yeah `CatcherArea` would have to make a revert result method).

Didn't know about `LastJudgementResult`, TIL, will use it.

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
frenzibyte: Also because I didn't want to think way more ahead and expose it "for later", but I guess it's fine to move this one out as `IComboCounter`

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
frenzibyte: actually the whole `LegacyComboCounter` thing seems to be movable outside, and makes sense to do so.

bdach: i don't know about the entire thing - aren't the animations ruleset-specific? pretty sure other rulesets animate the counter differently in stable

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
frenzibyte: Actually not sure about moving this one, as it asks for the combo colour of the object that was hit, which is catch-specific to my knowledge.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
bdach: checked myself - using `LastJudgementResult` in its current state is infeasible because (a) no support for reverting and (b) information about the DHO is not in `JudgementResult`, which is needed in this case for combo colour, so for now I just moved the calls to the counter's `On{New,Revert}Result` to `CatcherArea`. reasoning is that it's pretty WTF for `CatcherArea`'s `OnNewResult` method update only *half* of its children.

this could be improved upon if `GameplayBeatmap` had a way to proxy `On{New,Revert}Result` entirely as an event instead of just exposing the last judgement result. unsure if that's what we want, however.

osu.Game.Rulesets.Catch/UI/ICatchComboCounter.cs
bdach: left it for now because there's no better way to do this for now, and it's a catch-specific interface as is for the reason above. if there was a way for `GameplayBeatmap` to relay `On{New,Revert}Result` calls this could be entirely removed.

bdach:
pushed commits here to hopefully push this along further. I don't entirely like how this is written still, but it seems at least passable to me. another pair of eyes would be appreciated.

osu.Game/Screens/Play/Player.cs
```diff
@@ -632,6 +632,7 @@ public override void OnEntering(IScreen last)
 
             // bind component bindables.
             Background.IsBreakTime.BindTo(breakTracker.IsBreakTime);
+            gameplayBeatmap.IsBreakTime.BindTo(breakTracker.IsBreakTime);
```
smoogipoo, line 635: This is very questionable. `Player` should not be taking on logic to ensure that the beatmap is in the correct state.

I'd much rather that beatmap contain the tracker itself than do this. Is there a reason this can't be done?

smoogipoo (changes requested):

osu.Game/Screens/Play/Player.cs
frenzibyte: Haven't thought about that, makes sense by logic. will look at making `BreakTracker` part of the gameplay beatmap, but I'll open a separate PR as it may make things a bit complicated and add more irrelevant changes.

bdach: Not sure about containing as the break tracker is placed in a frame stable context right now. I wouldn't mind making it DI-able, though.

frenzibyte: > I wouldn't mind making it DI-able, though.

@bdach Sounded good at first, but after attempting to do it, I realized it would become hard to test the combo counter's behaviour on break time, as I can't just set `IsBreakTime` to whatever value I want like I did preivously, I would be forced to hack through the `BreakTracker` with a manipulated clock or otherwise, which doesn't sound like a fine thing here.

---
As for merging `BreakTracker` to `GameplayBeatmap`, as mentioned above, the break tracker is placed in a frame stable context currently, moving the tracker to inside `GameplayBeatmap` would not be that easy.

At first, I thought about passing `DrawableRuleset` in and adding a private break tracker component inside their exposed frame stable components container, but that sounded shoddy mostly for the part that `GameplayBeatmap` would no longer own or be an ancestor to the private break tracker component, therefore could potentially result in unexpected issues.

I've then took a bit to think about how to do this properly, and I came up with the following solution: caching `GameplayBeatmap` right where it was previously, but without it being added to the hierarchy, and then load it after `DrawableRuleset` is loaded inside the frame stable components container where the score and health processors are added.

This will make `GameplayBeatmap` obtain the frame-stable clock (which is logical for a "gameplay beatmap"), allowing it to track breaks very easily, without having to change its dependency caching position.
 - Could cache `GameplayBeatmap` right where it is loaded, but it is possible not to change the cache position so I didn't do that, tell me if that's wrong to do.

@smoogipoo does that sound good?

bdach: how would it be hard to test? can always expose a public `IsBreakTime` bindable on the combo counter (even if read-only), bind to it in the test and manipulate as required. there is precedent for this (see taiko mascot and scroller tests)

continuing that thread of thinking, you could also make the counter a `VisibilityContainer` or something.

frenzibyte: Because the test scene is dealing with `CatchComboDisplay`, not directly the legacy piece where the break logic resides. (`LegacyComboCounter`)

Not sure what you mean by making it a `VisibilityContainer`?

smoogipoo: Given the amount of complexity around this issue, I'm tempted to overlook it and fix it another day. Perhaps at some point we'll put a frame stable container inside the gameplay beatmap? Or inside player? Not sure...

@peppy what's your thoughts on leaving this as is?

peppy: My view is that we should settle on anything because at this point it feels like overthinking such a simple component.

I'd apply my usual thinking: as long as it's isolated enough that it's not polluting other systems or components (more than a ctor and maybe a couple of property updates) code quality or structure is not important for a first pass.

osu.Game.Rulesets.Catch/Skinning/LegacyComboCounter.cs
```diff
@@ -0,0 +1,129 @@
...
+
+        [Resolved(canBeNull: true)]
+        private GameplayBeatmap beatmap { get; set; }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            isBreakTime = beatmap?.IsBreakTime.GetBoundCopy();
+            isBreakTime?.BindValueChanged(b =>
```
peppy, line 60: is this honestly required? seems like we could just drop this code and reduce the complexity of this PR by miles. the only difference is the combo counter will stay on screen for up to 1 second extra during the start of a break?

frenzibyte: Yeah thought about that initially, but didn't want to change how it behaves on osu!stable, and didn't think it would make the PR that complex (but turns out it did), will remove it when I have time to.

peppy: it can of course be added back once we know what we're doing with break time knowledge.

osu.Game/Tests/Visual/SkinnableTestScene.cs
peppy: this is shocking and should not be in this PR. or in this test class alone.

peppy (approved):

peppy:
Going to merge this as most of my changes were just code quality improvements.

PR #9771, opened 2020-08-04
Separate "pause sample on gameplay pause" logic to a specific GameplaySkinnableSound class

### LINKED ISSUE #9739: Pause loop does not stop playing immediately after pressing continue
The new pause loop doesn't stop playing after pressing continue in the pause menu, it continues for about 2 more seconds instead. Seems like a small oversight.
https://youtu.be/CW6e-9jHk_M
**osu!lazer version:** 2020.801.0

frenzibyte:
https://github.com/ppy/osu/issues/9739 is silently fixed as the issue was about another sample channel created and the old one left playing in the background without any control from the game (which is the reason there's another pause loop playing even when resuming), but this PR resolves the entire issue from ever happening again (another playback being invoked while resuming). Closes #9739 

---

Code-side, a better fix that would get rid of this logic would be to cache a sample store somewhere at a `DrawableRuleset` level adjusting it to match the gameplay clock's rate, but would require https://github.com/ppy/osu-framework/issues/2760 to be fixed first. (added a todo comment for that)

Had to permit null if gameplay clock is not cached to DI, due to some tests failing because of it (`TestSceneNightcoreBeatContainer`)

Not testable as the existing tests are correct enough, this is only splitting the logic to avoid this issue ever happening again.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -87,7 +87,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             Tracking.BindValueChanged(updateSlidingSample);
         }
 
-        private SkinnableSound slidingSample;
+        private GameplaySkinnableSound slidingSample;
 
         protected override void LoadSamples()
         {
@@ -103,7 +103,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 var clone = HitObject.SampleControlPoint.ApplyTo(firstSample);
                 clone.Name = "sliderslide";
 
-                AddInternal(slidingSample = new SkinnableSound(clone)
+                AddInternal(slidingSample = new GameplaySkinnableSound(clone)
                 {
                     Looping = true
                 });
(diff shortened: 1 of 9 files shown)


bdach:
While this change seems sane, there [was actually a test for pause sample playback](https://github.com/ppy/osu/blob/d42a4e943ab1d158cbbbda99fa61709bf1c23154/osu.Game.Tests/Visual/Gameplay/TestScenePause.cs#L225-L248) that didn't catch this. I tried to make it fail on `master` but scarily enough I can't seem to find a way to make it fail (even though it's clearly audibly wrong in the test browser). I would like to see that test hardened as part of this PR in a way that would have caught that issue to begin with.

frenzibyte:
After investigating further, turns out the skinnable sound "pause on gameplay pause" logic (when resuming) is being invoked before `pauseLoop.Stop()` is invoked (because of it being delayed until the pause fade out transformation is complete). 

Calling `SampleChannel.Play()` on a channel that is already playing would cause another sample channel to be created and the previous one gets discarded away unsafely without it being stopped and freed, and thus you'll hear another pause loop playing in the background. (https://github.com/ppy/osu-framework/pull/3792 aims to fix this case).

frenzibyte:
Now that https://github.com/ppy/osu-framework/issues/2760 is fixed, I'll attempt to apply the better fix I proposed above, this can be blocked for now.

peppy:
I think this is no longer required, right? We have `PausableSkinnableSound` now which seems to fulfill the purposes outlined in this PR.

frenzibyte:
Yeah, I was aiming on replacing all the sample pause logic with an internal logic in `DrawableSample` that would automatically pause the channel if the clock gets stopped (or `Rate` becomes 0).

But currently, seems that having an `ISamplePlaybackDisabler` component that chooses whether to disable sample seems like a better choice (for combining "stop samples on seeking" logic in `Editor` without hacking around).

I will close this for now, the framework PRs are probably gonna conflict with the current `ISamplePlaybackDisabler` logic, no reason for them to stay as well.

PR #9778, opened 2020-08-05
Always place spinners behind hitcircles/sliders

### LINKED ISSUE #2823: Spinners should always be behind other hit objects
Updated images from https://github.com/ppy/osu/issues/5450

You can't see the hit circle, only the approach circle
With Hidden, the hit circle is completely invisible

smoogipoo:
Resolves https://github.com/ppy/osu/issues/2823

Pretty hard to case this to only skins for now. o!f will need additional support to be able to remove proxies, which should be done anyway for `Remove(obj)`...


peppy (approved):

peppy:
I actually think it's fine even with the default being back-most.

PR #9779, opened 2020-08-05
Fix very short spinners being impossible to complete

### LINKED ISSUE #5826: Spinners too short are in-completeable
If a spinner is too short, it will not be able to be completed, making a map un-fcable, even with auto
an example of a map that has this issue is : https://osu.ppy.sh/beatmapsets/600897#osu/1269591
in stable if a spinner is too short to complete the spinner will auto complete itself, it should do the same in lazer, if a spinner can't be completed normally it should just complete itself and give a 300

peppy:
- Closes #5826

If the calculated spin requirement is lower than 1, implicit completion is expected.

Can be tested on https://osu.ppy.sh/beatmapsets/132#osu/296.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
@@ -23,12 +23,10 @@ namespace osu.Game.Rulesets.Osu.Tests
         [TestCase(true)]
         public void TestVariousSpinners(bool autoplay)
         {
-            AddStep("Miss Big", () => SetContents(() => testSingle(2)));
-            AddStep("Miss Medium", () => SetContents(() => testSingle(5)));
-            AddStep("Miss Small", () => SetContents(() => testSingle(7)));
-            AddStep("Hit Big", () => SetContents(() => testSingle(2, autoplay)));
-            AddStep("Hit Medium", () => SetContents(() => testSingle(5, autoplay)));
-            AddStep("Hit Small", () => SetContents(() => testSingle(7, autoplay)));
+            string term = autoplay ? "Hit" : "Miss";
+            AddStep($"{term} Big", () => SetContents(() => testSingle(2, autoplay)));
+            AddStep($"{term} Medium", () => SetContents(() => testSingle(5, autoplay)));
+            AddStep($"{term} Small", () => SetContents(() => testSingle(7, autoplay)));
         }
 
         [TestCase(false)]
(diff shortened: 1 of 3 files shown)


frenzibyte:
Sounds like it closes https://github.com/ppy/osu/issues/5826.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
```diff
@@ -17,32 +17,54 @@ public class TestSceneSpinner : OsuSkinnableTestScene
...
-        public TestSceneSpinner()
+        private TestDrawableSpinner drawableSpinner;
+
+        [TestCase(false)]
+        [TestCase(true)]
+        public void TestVariousSpinners(bool autoplay)
         {
             AddStep("Miss Big", () => SetContents(() => testSingle(2)));
             AddStep("Miss Medium", () => SetContents(() => testSingle(5)));
             AddStep("Miss Small", () => SetContents(() => testSingle(7)));
```
bdach, line 28: Seeing the `[TestCase]` attributes added on this method, did you mean to remove these and replace them with the `false` test case? As it stands these are covered doubly.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
```diff
@@ -17,32 +17,54 @@ public class TestSceneSpinner : OsuSkinnableTestScene
...
+            AddStep("Very short spinner", () => SetContents(() => testSingle(5, autoplay, 200)));
+            AddUntilStep("Wait for completion", () => drawableSpinner.Result.HasResult);
+            AddUntilStep("Short spinner implicitly completes", () => drawableSpinner.Progress == 1);
         }
 
-        private Drawable testSingle(float circleSize, bool auto = false)
+        private Drawable testSingle(float circleSize, bool auto = false, double length = 3000)
         {
-            var spinner = new Spinner { StartTime = Time.Current + 2000, EndTime = Time.Current + 5000 };
+            var spinner = new Spinner { StartTime = Time.Current + 2000, EndTime = Time.Current + +2000 + length };
```
bdach, line 54: Not sure what the unary `+` is for:

```suggestion
            var spinner = new Spinner { StartTime = Time.Current + 2000, EndTime = Time.Current + 2000 + length };
```

bdach (commented):
looks good otherwise :+1:

osu.Game.Rulesets.Osu.Tests/TestSceneSpinner.cs
peppy: typo

smoogipoo (approved):

PR #9786, opened 2020-08-06
Allow storyboard elements to load textures from skins

### LINKED ISSUE #8472: Storyboards should have access to classic skin files
**Describe the bug:** 
Beatmaps don't pull sprites from skins.

No screenshots available, but can be reproduced by attempting to play Supairo's aspire!taiko map "Function Phantom - Algebra" in osu!stable and comparing it to playing in osu!lazer.

**osu!lazer version:** 
2020.327.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/4395360/database.log)
[network.log](https://github.com/ppy/osu/files/4395361/network.log)
[performance.log](https://github.com/ppy/osu/files/4395362/performance.log)
[runtime.log](https://github.com/ppy/osu/files/4395363/runtime.log)

Wieku:
~PR is not yet ready for merge.~

~In its current state it only works for sprites and only for custom skins as default skins don't use textures.~ - now it doesn't work for osu!lazer skin, it's fixed for osu!classic

~I don't know how to tackle animations though. [Animation](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/Animations/Animation.cs) class doesn't expose anything to remove already added frames. DrawableStoryboardAnimation could be changed to inherit DrawableAnimation instead of TextureAnimation and then changing the texture of each Sprite but I think that's suboptimal. Any suggestions appreciated.~ - resolved by framework fix

Preview: https://youtu.be/zLu52AYAJVw

Fixes #8472


osu.Game/Storyboards/Drawables/DrawableStoryboardSprite.cs
```diff
@@ -114,14 +121,36 @@ public DrawableStoryboardSprite(StoryboardSprite sprite)
...
+            skinChanged();
 
-            Texture = textureStore.Get(path);
             Sprite.ApplyTransforms(this);
         }
+
+        private void onChange() =>
+            // schedule required to avoid calls after disposed.
+            // note that this has the side-effect of components only performing a possible texture change when they are alive.
+            Scheduler.AddOnce(skinChanged);
```
frenzibyte, line 144: Using `SkinReloadableDrawable` tackles all these for you.

frenzibyte (commented):
For the animation case, you could as well make `DrawableStoryboardAnimation` a composite (as you should use `SkinReloadableDrawable` anyways) and recycle the `Animation` composite every skin change. 

But having `ClearFrames` framework-side is definitely a valid and better option for much simplicity, and looks pretty straightforward to implement so you can have a look at doing so if you would like to. (you can test your changes by writing a simple case at `TestSceneAnimation` to ensure you have things correct)

osu.Game/Storyboards/Drawables/DrawableStoryboardSprite.cs
```diff
@@ -10,13 +10,20 @@
...
 namespace osu.Game.Storyboards.Drawables
 {
-    public class DrawableStoryboardSprite : Sprite, IFlippable, IVectorScalable
+    public class DrawableStoryboardSprite : SkinReloadableDrawable, IFlippable, IVectorScalable
     {
         public StoryboardSprite Sprite { get; }
 
+        private Sprite drawableSprite;
+
+        private TextureStore storyboardTextureStore;
```
bdach, line 23: pretty sure you could make this a

```suggestion
        [Resolved]
        private TextureStore storyboardTextureStore { get; set; }
```

instead of adding this local and assigning to it in BDL

smoogipoo:
It would be nice for `Animation` to expose a `ClearFrames()` method.

osu.Game/Storyboards/Drawables/DrawableStoryboardSprite.cs
Wieku: Yep, that works

peppy:
What's the intention here? Generally we don't use draft PRs - if you are happy with the state of this please unmark it as a draft.

Wieku:
Ah, I marked it as a draft because animation skinning is not implemented yet

peppy:
@Wieku are you planning to make the framework change, or should we look at doing that?

Wieku:
I will look into it this week

Wieku:
I think it's done. IIRC storyboard samples can't load skin samples in stable.

osu.Game/Storyboards/Drawables/DrawableStoryboardAnimation.cs
```diff
@@ -108,28 +119,48 @@ public DrawableStoryboardAnimation(StoryboardAnimation animation)
...
+
+        protected override void SkinChanged(ISkinSource skin, bool allowFallback)
+        {
+            base.SkinChanged(skin, allowFallback);
+
+            drawableTextureAnimation.ClearFrames();
+
+            for (var frame = 0; frame < Animation.FrameCount; frame++)
+            {
+                var texture = skin?.GetTexture(texturePathsRaw[frame]) ?? storyboardTextureStore?.Get(texturePaths[frame]);
```
bdach, line 157: Doing things this way causes storyboard sprites to unnecessarily reload every time a skin changes. We probably want to bring back the old lookup logic in `load()`, but if the lookup fails, just fall back to a nested `SkinnableDrawable` which should handle the skin-changing logic cleanly on its own.

Same thing goes for sprites.

bdach (dismissed):

osu.Game/Storyboards/Drawables/DrawableStoryboardAnimation.cs
Wieku: We can't do that with `SkinnableDrawable`. `SkinnableDrawable` is for loading whole gameplay components (like hitcircles) if I'm not mistaken. We need to be able to load individual skin textures. Default font (the one on the hitcircles) as an animation for example. Also, it needs to be done during skin reload because stable replaces missing frames from skin with storyboard ones.

peppy: You should be able to nest one. Also using `SkinnableSprite` for simplicity.

For individual textures in an animation, I'd argue we shouldn't bother supporting this.

Wieku: Somehow I missed SkinnableSprite. My bad.

bdach:
I've started work on getting this finished but it actually turns out there is more to that functionality than given here - the option to fall back to skin sprites is actually modulated by [the `UseSkinSprites` setting](https://osu.ppy.sh/help/wiki/osu!_File_Formats/Osu_(file_format)#general), so going to have to think how to go about supporting this.

Immediate thought is to have the root `Storyboard` DI-able, so that it can be resolved in the sprites/animations to read the setting value, but don't know for sure if it'll work with the current method of construction of the storyboard (I *think* it should, though).

Wieku:
I think passing it through element controllers also might be a good approach in case storyboarding would be expanded to allow only certain elements to be skinnable. But I also think it's too far-fetched.

bdach:
I considered that briefly but seeing that some [beatmap](https://osu.ppy.sh/beatmapsets/1130581#osu/2361608) [storyboards](https://osu.ppy.sh/beatmapsets/470977#osu/1006822) use tens of thousands of sprites for effects I'm not sure whether incurring that memory cost is worth it.

bdach:
I've pushed in the changes alongside with a simple test scene. Aside from that this can also be tested with the map linked in the video in the OP, which is https://osu.ppy.sh/beatmapsets/832364#osu/1743837.

The ugliest part is probably the `DrawableStoryboard{Sprite,Animation}` logic duplication. I wanted to use `DrawableStoryboardSprites` as frames of `DrawableStoryboardAnimation`, but the former has some lifetime logic related to events baked into it and as such wouldn't work (spotted this while debugging why my sprites weren't showing up in the test scene). Unsure if relevant.

I would have also added a test for standard storyboard sprites, just for completeness' sake - I even have one stashed, but it requires having files mirroring our hashed storage setup (so i.e. `a/ab/abcdef`) in one of the test resource folders, so decided against it for now. Can push in on request.

osu.Game/Storyboards/Drawables/DrawableStoryboardAnimation.cs
```diff
@@ -115,18 +117,18 @@ public DrawableStoryboardAnimation(StoryboardAnimation animation)
...
         [BackgroundDependencyLoader]
-        private void load(IBindable<WorkingBeatmap> beatmap, TextureStore textureStore)
+        private void load(IBindable<WorkingBeatmap> beatmap, TextureStore textureStore, Storyboard storyboard)
         {
             for (var frame = 0; frame < Animation.FrameCount; frame++)
             {
                 var framePath = Animation.Path.Replace(".", frame + ".");
 
-                var path = beatmap.Value.BeatmapSetInfo.Files.Find(f => f.Filename.Equals(framePath, StringComparison.OrdinalIgnoreCase))?.FileInfo.StoragePath;
-                if (path == null)
```
peppy, line 125: does this work as expected without this check? kinda undefined behaviour now.. (the case where `UseSkinSprites == false`)

bdach: Good point. While setting a null texture on a sprite is supported as far as I can tell it should probably not happen if it can be avoided.

peppy: Yeah it does look to work. Maybe we do want to initialise the sprites regardless thinking forward to storyboard editor implementation? They look to show up in the draw visualiser and behave as normal (but not display due to missing texture).

I think it's fine to leave this way since it doesn't crash or anything.

peppy: Actually never mind, will change as i think it improves readability (and what I said doesn't make much sense – this is already encapsulated in a drawable representing the storyboard sprite).

peppy: Ended up combining the lookup logic. Feels quite a bit cleaner.

osu.Game/Storyboards/Drawables/DrawableStoryboardSprite.cs
```diff
@@ -114,13 +112,13 @@ public DrawableStoryboardSprite(StoryboardSprite sprite)
...
-        private void load(IBindable<WorkingBeatmap> beatmap, TextureStore textureStore)
+        private void load(TextureStore textureStore, Storyboard storyboard)
         {
-            var path = beatmap.Value.BeatmapSetInfo?.Files?.Find(f => f.Filename.Equals(Sprite.Path, StringComparison.OrdinalIgnoreCase))?.FileInfo.StoragePath;
-            if (path == null)
-                return;
+            var drawable = storyboard.CreateSpriteFromResourcePath(Sprite.Path, textureStore);
+
+            if (drawable != null)
+                InternalChild = drawable.With(s => s.Anchor = s.Origin = Anchor.Centre);
```
peppy, line 120: this is completely wrong, and ignored any origin changes applied to the sprite

peppy (approved):
seems good

osu.Game/Storyboards/Drawables/DrawableStoryboardSprite.cs
bdach: ouch. not sure where i got that from...

PR #9788, opened 2020-08-06
Make SkinnableSound an IAdjustableAudioComponent

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3790

Similar to the framework PR, this adds support for transform sequencing and generic easings.


bdach:
I've merged `master` here - there were conflicts, but they were already resolved in the dependent PR, so I followed that resolution. Diffs look identical to me but I'd appreciate a double-check.

Other than that, this seems to be good to go as far as I can tell :+1:

peppy (approved):

PR #9791, opened 2020-08-06
Add keybinding to toolbar button's tooltip

### LINKED ISSUE #9147: Tooltips with keybindings
**Describe the new feature:** 
There's many global keybindings and names that are not common knowledge. A tooltip on hover with the name + keybinding would be a nice quality-of-life improvement.

osu_tooltip
(please ignore the inaccurate keybinding, I messed up)

Joehuu:
- Resolves https://github.com/ppy/osu/issues/9147

Went with putting it with sub text, as said in https://github.com/ppy/osu/issues/9147#issuecomment-636862990, seems less noisy.

image

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -67,8 +67,6 @@ namespace osu.Game
         [NotNull]
         private readonly NotificationOverlay notifications = new NotificationOverlay();
 
-        private NowPlayingOverlay nowPlaying;
-
         private BeatmapListingOverlay beatmapListing;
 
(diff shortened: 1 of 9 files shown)


peppy:
I've made some minor changes to avoid querying the key binding store many times each binding change (even though this is a relatively cheap operation right now, having it happen for every toolbar button at once would potentially cause a minor stutter).

In the future we may want to use a better method of querying, as the current one is always retrieving every binding every time. But seems fine for now.

peppy (approved):

PR #9792, opened 2020-08-06
Move beatmap track into MusicController

smoogipoo:
- [x] Depends on https://github.com/ppy/osu/pull/9788
- [x] Depends on https://github.com/ppy/osu-framework/pull/3809
- [x] Depends on https://github.com/ppy/osu-framework/pull/3810

I apologise for the size of this one. It's really hard to make such a change without touching almost every file in the game, which probably spells bad news for the reach of the (now) `MusicController` in general.  
Going forward this can probably be fixed on a per-case basis by either passing down the track or caching it locally at overarching components, which is probably a good idea in general for component isolation. I've already done this for `Player` because it felt wrong to DI `MusicController` in `GameplayClockContainer` and `FailAnimation`.

The idea of this PR is to define a single owner for the beatmap's track. Specifically, the track which plays the sound and which is stored as an on-going reference.  
The goal is to eventually be able to fade between tracks when changed in song select, which needs both the single owner idea and for the track to be able to be placed in the draw hierarchy - it needs to also become a `DrawableTrack`.

### Changes of particular note
- "Recycling" of working beatmap tracks has been removed. It is now handled by `MusicController`.
- The "transferring" of working beatmaps has been removed. Both the track and texture stores are passed to new `WorkingBeatmap` instances by `BeatmapManager` and the disposal of the track store is now also handled by the `BeatmapManager`.
- The cycling of tracks once one has complete has been moved from `OsuGame` to `MusicController`.
- `MusicController` has been moved to `OsuGameBase`. It is now loaded as part of the async procedure but before the game's `Loader`. I haven't fully assessed the performance impact of this since it was previously in a `loadComponentSingleFile()` but it doesn't seem to have a noticeable impact when loading osu! from an external USB drive with ~1000 sets.
    - Another way this can be done is to move the track handling part into its own separate component, like a "TrackController" that is DI'd in by the `MusicController`.

### Testing
Beyond the automated CI tests, I've also tested a whole bunch of scenarios, including:
- Selecting multiplayer item with no local beatmap.
- Selecting multiplayer item with a local beatmap.
- Completing a beatmap import with the selected playlist item.
- All the intros.
- Track loops in song select.
- Track doesn't loop in main menu.
- Track loops in match subscreen.
- Track doesn't change when another beatmap from the same set is selected.
- Failing in gameplay.
- Restarting in gameplay.
- DT/NC/WU/WD mods applying at song select.
- DT/NC/WU/WD mods applying in gameplay.
- DT/NC/WU/WD mods applying after map is restarted.
- DT/NC/WU/WD mods applying in multiplayer (match subscreen and gameplay).
- Playing preview tracks to mute audio.
- Changing track while a preview track is playing.


bdach:
Mostly out of curiosity, what is the end goal here? This seems like a pretty drastic change in direction.

smoogipoo:
The goal is to have a dedicated owner of the track, rather than it being referenced/potentially created by everything in the game. One immediate use for this is to fade between tracks at song select, which is a 2-liner now that `MusicController` and .Expire() and transform the new track.

smoogipoo:
Alright, PR description updated and this PR is now ready for review.

frenzibyte (changes requested):
Getting an "invalid thread for child mutation" exception by just clicking on `TestSceneNowPlayingOverlay` in the visual test browser:
```
[runtime] 2020-08-07 20:37:36 [error]: osu.Framework.Graphics.Containers.CompositeDrawable+InvalidThreadForChildMutationException: Cannot mutate the children of a Loaded CompositeDrawable while not on the update thread. Consider using Schedule to schedule the mutation operation.
[runtime] 2020-08-07 20:37:36 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.ensureChildMutationAllowed()
[runtime] 2020-08-07 20:37:36 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.AddInternal(Drawable drawable)
[runtime] 2020-08-07 20:37:36 [error]: at osu.Game.Overlays.MusicController.changeTrack() in /Users/salman/Desktop/osu/osu.Game/Overlays/MusicController.cs:line 324
[runtime] 2020-08-07 20:37:36 [error]: at osu.Game.Overlays.MusicController.beatmapChanged(ValueChangedEvent`1 beatmap) in /Users/salman/Desktop/osu/osu.Game/Overlays/MusicController.cs:line 309
[runtime] 2020-08-07 20:37:36 [error]: at osu.Framework.Bindables.Bindable`1.TriggerValueChange(T previousValue, Bindable`1 source, Boolean propagateToBindings, Boolean bypassChecks)
```

---

Other than that this looks good to me and I like the direction of it, read the entire changed files twice, but would definitely request another pair of eyes to this as one can't ensure everything is right.

osu.Game/Overlays/MusicController.cs
```diff
@@ -21,7 +24,7 @@ namespace osu.Game.Overlays
     /// <summary>
     /// Handles playback of the global music track.
     /// </summary>
-    public class MusicController : Component, IKeyBindingHandler<GlobalAction>
+    public class MusicController : CompositeDrawable, IKeyBindingHandler<GlobalAction>
```
bdach, line 27: In general I'd be in favour of moving this out of the `osu.Game.Overlays` namespace out to `osu.Game.Audio`, but I wouldn't necessarily muddy this PR with that change as it's hard enough to read through as-is.

osu.Game/Overlays/MusicController.cs
```diff
@@ -307,13 +304,36 @@ private void beatmapChanged(ValueChangedEvent<WorkingBeatmap> beatmap)
...
             queuedDirection = null;
         }
 
+        private void changeTrack()
+        {
+            CurrentTrack.Expire();
+            CurrentTrack = new DrawableTrack(current.GetTrack());
+            CurrentTrack.Completed += () => onTrackCompleted(current);
+
+            AddInternal(CurrentTrack);
```
bdach, line 324: As mentioned by @frenzibyte this is causing crashes in test scenes - as far as I'm aware value changed callbacks run on the thread that changed the value, which is not necessarily the update thread, so it seems this needs a schedule.

A normal schedule still doesn't always work (crashes i.e. in `TestSceneNightcoreBeatContainer`); `Scheduler.AddOnce()` seems to not crash any of the test scenes I checked but I'm unsure whether it's a correct thing to do, or if it should be a normal schedule and all crashing sites should apply local fixes.

osu.Game/Overlays/NowPlayingOverlay.cs
```diff
@@ -234,7 +234,7 @@ protected override void Update()
                 pendingBeatmapSwitch = null;
             }
 
-            var track = beatmap.Value?.TrackLoaded ?? false ? beatmap.Value.Track : null;
+            var track = musicController.TrackLoaded ? musicController.CurrentTrack : null;
```
bdach, line 237: I'm aware this is the minimal and henceforth "safe" change, but unless I'm missing something it seems to me that you could drop this entirely and just inspect `CurrentTrack`.

osu.Game/Screens/Edit/Compose/Components/Timeline/Timeline.cs
```diff
@@ -57,7 +62,9 @@ private void load(IBindable<WorkingBeatmap> beatmap, OsuColour colours)
             Beatmap.BindValueChanged(b =>
             {
                 waveform.Waveform = b.NewValue.Waveform;
-                track = b.NewValue.Track;
+                track = musicController.CurrentTrack;
+
+                Debug.Assert(track != null);
```
bdach, line 67: This assert feels a little strange seeing that `musicController.CurrentTrack` is already annotated as `[NotNull]`. I know it's a jetbrains annotation and therefore non-enforceable on a compiler level (although I *think* we have inspections for that enabled), but why guard this place but not the others?

osu.Game/Screens/Edit/Editor.cs
```diff
@@ -346,7 +348,7 @@ public override bool OnExiting(IScreen next)
 
         private void resetTrack(bool seekToStart = false)
         {
-            Beatmap.Value.Track?.Stop();
+            musicController.Stop();
```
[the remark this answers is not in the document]
bdach: While I agree that stopping the track globally is probably not desired I don't think `IsUserPaused` should be exposed as it violates encapsulation. I'd rather have the editor resume the track if it stopped it on `OnExiting()` or otherwise.

osu.Game/Screens/Edit/EditorClock.cs
```diff
@@ -25,8 +25,13 @@ public class EditorClock : Component, IFrameBasedClock, IAdjustableClock, ISourc
 
         private readonly DecoupleableInterpolatingFramedClock underlyingClock;
 
-        public EditorClock(WorkingBeatmap beatmap, BindableBeatDivisor beatDivisor)
-            : this(beatmap.Beatmap.ControlPointInfo, beatmap.Track.Length, beatDivisor)
+        public EditorClock(WorkingBeatmap beatmap, double trackLength, BindableBeatDivisor beatDivisor)
```
bdach, line 28: This feels a little awkward in that `trackLength` is passed *alongside* the beatmap. Can't the track length be accessed by just doing `beatmap.GetTrack().Length`?

osu.Game/Screens/Menu/IntroScreen.cs
```diff
@@ -111,7 +115,7 @@ bool loadThemedIntro()
                 if (setInfo != null)
                 {
                     initialBeatmap = beatmaps.GetWorkingBeatmap(setInfo.Beatmaps[0]);
-                    UsingThemedIntro = !(Track is TrackVirtual);
+                    UsingThemedIntro = initialBeatmap.GetTrack().IsDummyDevice == false;
```
bdach, line 118: nit
```suggestion
                    UsingThemedIntro = !initialBeatmap.GetTrack().IsDummyDevice;
```

osu.Game/Rulesets/Mods/ModDaycore.cs
```diff
@@ -27,11 +27,11 @@ protected ModDaycore()
...
         }
 
-        public override void ApplyToTrack(Track track)
+        public override void ApplyToTrack(ITrack track)
         {
             // base.ApplyToTrack() intentionally not called (different tempo adjustment is applied)
-            track.AddAdjustment(AdjustableProperty.Frequency, freqAdjust);
-            track.AddAdjustment(AdjustableProperty.Tempo, tempoAdjust);
+            (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Frequency, freqAdjust);
+            (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Tempo, tempoAdjust);
```
bdach, line 34: It feels a little clumsy that `ITrack` is not an `IAdjustableAudioComponent`, even more so seeing that `ITrack` even has [a specific method for removing speed adjustments](https://github.com/ppy/osu-framework/blob/05c118dbc94fd2e21e71595f8cf62631c94efe15/osu.Framework/Audio/Track/ITrack.cs#L78-L81).

osu.Game/Screens/Menu/IntroTriangles.cs
```diff
@@ -59,7 +59,7 @@ protected override void LogoArriving(OsuLogo logo, bool resuming)
                 LoadComponentAsync(new TrianglesIntroSequence(logo, background)
                 {
                     RelativeSizeAxes = Axes.Both,
-                    Clock = new FramedClock(UsingThemedIntro ? Track : null),
+                    Clock = new FramedClock(UsingThemedIntro ? (IAdjustableClock)Track : null),
```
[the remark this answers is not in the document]
bdach: Agreed. The only thing that `ITrack` doesn't have compared to `IAdjustableClock` is `Rate`.

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -63,17 +60,17 @@ public class GameplayClockContainer : Container
 
         private readonly FramedOffsetClock platformOffsetClock;
 
-        public GameplayClockContainer(WorkingBeatmap beatmap, IReadOnlyList<Mod> mods, double gameplayStartTime)
+        public GameplayClockContainer(ITrack track, WorkingBeatmap beatmap, IReadOnlyList<Mod> mods, double gameplayStartTime)
```
bdach, line 63: Seeing that all call sites seem to be calling this with a non-null `track` I'd stick an annotation on it and eliminate all the null-checking cases.

osu.Game/Beatmaps/WorkingBeatmap.cs
```diff
@@ -250,10 +249,9 @@ public IBeatmap Beatmap
         protected abstract Texture GetBackground();
         private readonly RecyclableLazy<Texture> background;
 
-        public virtual bool TrackLoaded => track.IsResultAvailable;
-        public Track Track => track.Value;
-        protected abstract Track GetTrack();
-        private RecyclableLazy<Track> track;
+        public Track GetTrack() => GetBeatmapTrack() ?? GetVirtualTrack(1000);
```
bdach, line 252:
```suggestion
        [NotNull]
        public Track GetTrack() => GetBeatmapTrack() ?? GetVirtualTrack(1000);
```

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -8,16 +8,14 @@
...
 using osu.Framework.Graphics.Audio;
 using osu.Framework.Graphics.Containers;
-using osu.Framework.Graphics.Transforms;
 using osu.Game.Audio;
 using osu.Game.Screens.Play;
 
 namespace osu.Game.Skinning
 {
-    public class SkinnableSound : SkinReloadableDrawable
+    public class SkinnableSound : SkinReloadableDrawable, IAdjustableAudioComponent
```
bdach, line 18: this one in particular seems like it could have been split, as far as I can tell this change causes a bunch of weirdly-looking unrelated changes (one-liner import additions that cause compilation failures if removed)

bdach (changes requested):
I read through this and checked all concrete non-abstract test scenes that were explicitly changed in this PR. Some crash in the test browser, as detailed in the comments above & below. I haven't ran comprehensive manual tests on the game itself yet, as it's already taken me about 2 hours to go through this diff.

As far as the general direction is concerned I feel that code quality has improved in places (mainly due to placing non-null constraints on things) and the rationale for the change has merit, but we're going to have to be really careful with this direction going forward, as I'm seeing a risk of `MusicController` becoming a god class if we're not vigilant.

osu.Game/Screens/Edit/Editor.cs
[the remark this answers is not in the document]
frenzibyte: The problem here is that calling `MusicController.Play()` / `MusicController.Stop()` modifies `IsUserPaused` state, which is not an expected behaviour here and the name is misleading for what's gonna happen (the track is expected to be paused internally without the user being the reason of), I would say at the very least rename:
```csharp
MusicController.Play();
MusicController.Stop();
...
```
to
```csharp
MusicController.UserPlay();
MusicController.UserStop();
...
```
to avoid such an issue like the one above.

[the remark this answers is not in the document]
bdach: While I agree this should be signalled better (and is perhaps maybe indicative of the fact that `MusicController` is mixing concerns too much) I still don't think exposing internal state is a proper solution to that. `User{Play,Stop}` would be ok with me I suppose.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -246,20 +247,36 @@ public WorkingBeatmap GetWorkingBeatmap(BeatmapInfo beatmapInfo, WorkingBeatmap
...
-                if (working == null)
-                {
-                    beatmapInfo.Metadata ??= beatmapInfo.BeatmapSet.Metadata;
+                beatmapInfo.Metadata ??= beatmapInfo.BeatmapSet.Metadata;
 
-                    workingCache.Add(working = new BeatmapManagerWorkingBeatmap(Files.Store,
-                        new LargeTextureStore(host?.CreateTextureLoaderStore(Files.Store)), beatmapInfo, audioManager));
-                }
+                ITrackStore trackStore = workingCache.FirstOrDefault(b => b.BeatmapInfo.AudioEquals(beatmapInfo))?.TrackStore;
+                TextureStore textureStore = workingCache.FirstOrDefault(b => b.BeatmapInfo.BackgroundEquals(beatmapInfo))?.TextureStore;
```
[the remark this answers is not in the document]
frenzibyte: I don't think so, there may be a beatmap with the same audio file as `beatmapInfo` but in the end another `WorkingBeatmap` will be cached targetting `beatmapInfo` specifically regardless of the one with same audio file.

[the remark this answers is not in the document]
bdach: Oh I see... That's super confusing. Almost feels like we should be collecting track/texture stores into separate caches.

osu.Game/Skinning/SkinnableSound.cs
bdach: oh wait, I just noticed it *is* split, I just didn't notice due to a lack of dependency listed in the PR description. my bad. opening post edited to reflect that

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
```diff
@@ -152,6 +140,11 @@ protected override ISkin GetSkin()
                     return null;
                 }
             }
+
+            ~BeatmapManagerWorkingBeatmap()
+            {
+                dereferenceAction?.Invoke(TrackStore);
+            }
```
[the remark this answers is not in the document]
peppy: Historically this was required because we are not sure when a working beatmap was finished being used. I'm not sure if that has changed at all with this PR but I'd hope we have a more defined lifetime now.

Iceotaku (dismissed):

osu.Game/Rulesets/Mods/ModDaycore.cs
smoogipoo: I agree, but I'm not sure what the best solution is to this. I'm not 100% sure `ITrack` should be an `IAdjustableAudioComponent` and would like more thoughts on that, with the obvious alternative being to remove `ResetSpeedAdjustments` from `ITrack`.

@peppy any opinion?

peppy: maybe just nuke the method and use the full interface here..

osu.Game.Tests/Visual/Navigation/TestSceneScreenNavigation.cs
```diff
@@ -62,30 +60,27 @@ public void TestSongContinuesAfterExitPlayer(bool withUserPause)
...
 
             PushAndConfirm(() => songSelect = new TestSongSelect());
 
-            AddUntilStep("wait for no track", () => track() is TrackVirtual);
+            AddUntilStep("wait for no track", () => Game.MusicController.CurrentTrack.IsDummyDevice);
 
             AddStep("return to menu", () => songSelect.Exit());
 
-            AddUntilStep("wait for track", () => !(track() is TrackVirtual) && track().IsRunning);
+            AddUntilStep("wait for track", () => Game.MusicController.CurrentTrack.IsDummyDevice == false && Game.MusicController.IsPlaying);
```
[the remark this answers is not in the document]
smoogipoo: The track used to be nullable, I just missed this (and possibly more) cases :(

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -225,8 +222,8 @@ private void updateRate()
             speedAdjustmentsApplied = true;
             track.ResetSpeedAdjustments();
 
-            track.AddAdjustment(AdjustableProperty.Frequency, pauseFreqAdjust);
-            track.AddAdjustment(AdjustableProperty.Tempo, UserPlaybackRate);
+            (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Frequency, pauseFreqAdjust);
+            (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Tempo, UserPlaybackRate);
 
             foreach (var mod in mods.OfType<IApplicableToTrack>())
                 mod.ApplyToTrack(track);
```
frenzibyte, line 229: I'm not quite sure about this one but I think it's worth mentioning: since `AllowRateAdjustments` is forced to `true` on `Player`, and now that `MusicController` is moved to an `OsuGameBase` level, this logic can be removed, right?

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
```diff
@@ -19,13 +19,21 @@ public partial class BeatmapManager
...
         protected class BeatmapManagerWorkingBeatmap : WorkingBeatmap
         {
+            public readonly TextureStore TextureStore;
+            public readonly ITrackStore TrackStore;
+
             private readonly IResourceStore<byte[]> store;
+            private readonly Action<ITrackStore> dereferenceAction;
 
-            public BeatmapManagerWorkingBeatmap(IResourceStore<byte[]> store, TextureStore textureStore, BeatmapInfo beatmapInfo, AudioManager audioManager)
+            public BeatmapManagerWorkingBeatmap(IResourceStore<byte[]> store, TextureStore textureStore, ITrackStore trackStore, BeatmapInfo beatmapInfo, AudioManager audioManager,
```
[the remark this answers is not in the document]
smoogipoo: Instance re-use is the reason for textureStore/trackStore. Beyond that, they each load specific extensions. It's also possible, given caching, that the texture/track stores point to different "instances" (i.e. they're not necessarily both from the same cache key). Perhaps this ties into https://github.com/ppy/osu/pull/9792#discussion_r467459273 ?

How do you propose changing this?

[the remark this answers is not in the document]
smoogipoo: Just as a bit of an update: I've reworked this stuff to remove the reference counting, but the separation of the stores still exists.

osu.Game/Overlays/MusicController.cs
smoogipoo: This maaaay be fine. Will have to check it behaves correctly with `restartTrack()`.

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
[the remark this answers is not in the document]
bdach: Hm, I see. The fact the stores are separated is ok but having to pass them down in threes separately maybe not so much. Maybe something like a `BeatmapSharedStoreCache` could be passed down instead and `WorkingBeatmap` could request the stores it needs from it? Not sure.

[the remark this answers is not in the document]
smoogipoo: I've just made the class private, so I'm inclined to keep it as it is instead of complicating it further.

osu.Game/Overlays/MusicController.cs
smoogipoo: Let's keep that for a separate PR.

osu.Game/Screens/Play/GameplayClockContainer.cs
smoogipoo: I'm hesitant to make changes in this PR. This doesn't immediately work because of the `Reset()` call(s). The following demonstrates a way to make this work, but there's quite a few things to consider (e.g what was the deadlock we were experiencing?):

```diff
diff --git a/osu.Game/Screens/Play/GameplayClockContainer.cs b/osu.Game/Screens/Play/GameplayClockContainer.cs
index f3466a562..cb3eb429e 100644
--- a/osu.Game/Screens/Play/GameplayClockContainer.cs
+++ b/osu.Game/Screens/Play/GameplayClockContainer.cs
@@ -122,13 +122,10 @@ private void load(OsuConfigManager config)

         public void Restart()
         {
-            // The Reset() call below causes speed adjustments to be reset in an async context, leading to deadlocks.
-            // The deadlock can be prevented by resetting the track synchronously before entering the async context.
-            track.ResetSpeedAdjustments();
-
             Task.Run(() =>
             {
-                track.Reset();
+                track.Seek(0);
+                track.Stop();

                 Schedule(() =>
                 {
@@ -216,14 +213,13 @@ protected override void Update()

         private void updateRate()
         {
-            speedAdjustmentsApplied = true;
-            track.ResetSpeedAdjustments();
+            if (speedAdjustmentsApplied)
+                return;

             (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Frequency, pauseFreqAdjust);
             (track as IAdjustableAudioComponent)?.AddAdjustment(AdjustableProperty.Tempo, UserPlaybackRate);

-            foreach (var mod in mods.OfType<IApplicableToTrack>())
-                mod.ApplyToTrack(track);
+            speedAdjustmentsApplied = true;
         }

         protected override void Dispose(bool isDisposing)
@@ -234,11 +230,13 @@ protected override void Dispose(bool isDisposing)

         private void removeSourceClockAdjustments()
         {
-            if (speedAdjustmentsApplied)
-            {
-                track.ResetSpeedAdjustments();
-                speedAdjustmentsApplied = false;
-            }
+            if (!speedAdjustmentsApplied)
+                return;
+
+            (track as IAdjustableAudioComponent)?.RemoveAdjustment(AdjustableProperty.Frequency, pauseFreqAdjust);
+            (track as IAdjustableAudioComponent)?.RemoveAdjustment(AdjustableProperty.Tempo, UserPlaybackRate);
+
+            speedAdjustmentsApplied = false;
         }
```

osu.Game/Rulesets/Mods/ModDaycore.cs
smoogipoo: So as it turns out the method isn't actually coming from `ITrack`, it's coming from `IAdjustableClock`... `ITrack` only duplicates it.

I went with the solution proposed above, which is to make `ITrack` an `IAdjustableAudioComponent`.

osu.Game/Screens/Edit/Editor.cs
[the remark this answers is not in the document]
smoogipoo: Just so we're all on the same page with the way I intended these changes to go down: the above change is a mishap from the way this PR took place - in two passes.

I intended for all non-test code of the form `beatmap.Track` to take on the form `music.CurrentTrack`, regardless of whether or not it is more correct than calling Play/Stop on `MusicController`. I'm trying to avoid logical changes as much as possible for this round of changes such that most such changes appear only in OsuGame/OsuGameBase/MusicController/TestScenes/BeatmapManager/WorkingBeatmap, which are either "overarching" components or ones that can't exist without change.

peppy:
I feel very uneasy about referencing `MusicController.CurrentTrack` from so many places. Before I get too deep, what's the reason for not leaving the referencing the track via the beatmap in places that makes sense (like the editor)? And where's the guarantee that `MusicController.CurrentTrack` isn't going to change? Are we relying on bindable leases?

smoogipoo:
> I feel very uneasy about referencing `MusicController.CurrentTrack` from so many places

Hence why I asked you about this before going ahead with it ;)

> where's the guarantee that `MusicController.CurrentTrack` isn't going to change

There isn't any. Sure, we're relying on bindable leases but it still feels wrong even with that. As I said in the OP, maybe we can cache the track locally or pass it down via ctor, and I've already done that for player since it felt wrong.

> what's the reason for not leaving the referencing the track via the beatmap

Mostly because it's very easy to abuse that. `GetTrack()` returns a new instance, every single time. In the case of `BeatmapManagerWorkingBeatmap`, it's also an instance that is attached to the `AudioManager`, meaning some unsuspecting user can do `beatmap.GetTrack().Play()` and suddenly have two playing tracks.

In the case of Editor, if you were to pass that track into the editor clock I believe you'll encounter the situation above. It also won't be linked to the `MusicController`'s track adding to the weirdness. And speaking of the Editor, where are you suggesting for the references to the beatmap track to remain?

Just for the record, I don't like that `beatmap.GetTrack()` is used in the two (I think) other cases either, but they're hard to remove because the `MusicController` track isn't being set.

peppy:
So what's the plan with the other elements of `WorkingBeatmap`, which still return single instance (`Waveform`, `Beatmap` etc)? Also what's the reason for not having workingbeatmap keep the reference (weak if necessary)?

smoogipoo:
> Also what's the reason for not having WorkingBeatmap keep the reference

Now that I'm thinking about it more, it technically is already cached since it uses `trackStore.Get()` so it's not a new instance every time, but there's still the potential for a leak to happen if you're not 100% sure the track is current in the `MusicController` (that handles the disposal).

> So what's the plan with the other elements of `WorkingBeatmap`

There are none.

peppy:
Could we potentially restore the previous behaviour in WorkingBeatmap using a weak reference or otherwise? Like, keeping MusicController in control of disposing the instance (and throwing if it's used via `working.Track` after disposal?) but still have it retained in the `WorkingBeatmap` for use in places like the editor as before?

smoogipoo:
That feels like a step backwards and a question as to whether there's any purpose to this PR anymore. Let me just reiterate, the entire point of this PR is for there to be a singular owner of the track - one thing that manages its lifetime, one track that plays for a beatmap, etc.

You're essentially asking for a complete revert of that. For bringing back the transfer of track stores, disposal of workingbeatmap, recycling of track by osugamebase, etc.

`WeakReference` doesn't work because the track is part of the track store.

peppy:
I don't think that's what I was proposing. The "owner" would still be the music controller, but other game components would access via the `WorkingBeatmap` still.

As safety, it could not get reconstructed if disposed (just remain in a throwing state).

I'm not saying any of the other logic should be reverted.

smoogipoo:
So consider a component that calls `beatmapManager.GetWorkingBeatmap(...).GetTrack().Play()` or similar. Maybe even calling a `.Stop()`.
There are two things that happen in this scenario:
1. The track has to be received from a cache (i.e. not a new `TrackStore`).
2. The track is immediately added to said `TrackStore`.

But this track is not current in the `MusicController`, so it doesn't handle disposal. What do you do in this scenario? How do you safeguard against this?

The way I was going was making `GetTrack()` never used outside of `MusicController`, except for very specific purposes where that is the functionality you want and you have to handle disposal yourself (can't really think of any off the top of my head, external/test code mostly). But that's too much to handle do in every case that references the track.

peppy:
I was thinking the loading of track would still be done by the `MusicController`. But read-only access to the track was still there after that point.

something like

```csharp
Assert.Throws<LoadViaMusicControllerException>(workingBeatmap.Track);
musicController.Load(workingBeatmap)
Assert.DoesNotThrow(workingBeatmap.Track);

musicController.Load(otherWorkingBeatmap);
Assert.Throws<LoadViaMusicControllerException>(workingBeatmap.Track); // or ObjectDisposedException if we don't want to allow reloading, not sure.
```

osu.Game/Beatmaps/IWorkingBeatmap.cs
```diff
@@ -59,5 +54,10 @@ public interface IWorkingBeatmap
         /// <returns>The converted <see cref="IBeatmap"/>.</returns>
         /// <exception cref="BeatmapInvalidForRulesetException">If <see cref="Beatmap"/> could not be converted to <paramref name="ruleset"/>.</exception>
         IBeatmap GetPlayableBeatmap(RulesetInfo ruleset, IReadOnlyList<Mod> mods = null, TimeSpan? timeout = null);
+
+        /// <summary>
+        /// Retrieves the <see cref="Track"/> which this <see cref="WorkingBeatmap"/> provides.
```
peppy, line 59: Was about to suggest this reads "Retrieves a new instance of the..." but I think due to trackStore doing caching internally this generally isn't always true. I guess it reads fine as-is.

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -63,17 +62,17 @@ public class GameplayClockContainer : Container
 
         private readonly FramedOffsetClock platformOffsetClock;
 
-        public GameplayClockContainer(WorkingBeatmap beatmap, IReadOnlyList<Mod> mods, double gameplayStartTime)
+        public GameplayClockContainer([NotNull] ITrack track, WorkingBeatmap beatmap, IReadOnlyList<Mod> mods, double gameplayStartTime)
         {
             this.beatmap = beatmap;
             this.mods = mods;
             this.gameplayStartTime = gameplayStartTime;
+            this.track = track;
```
peppy, line 70: In this usage, is there a reason we can't just call `GetTrack()` on the `WorkingBeatmap`, since it will return the same instance?

peppy:
I'm at 55/68 files reviewed for this – components which can generally be thought of dealing with the "global" track work really nicely.

The remaining are `Multiplayer`, `Editor` and `Player`. I've noticed there's already a few usages of `beatmap.GetTrack()`, and need to check the correct direction before going forward further. Is it safe to call this? Should I be calling this once in the top level screens and cache for components as an `ITrack`?

peppy:
@smoogipoo can i get a reply here so i can move forward? thanks!

smoogipoo:
Calling `beatmap.GetTrack()` is not safe because it's added to the track store.

osu.Game/Screens/Play/GameplayClockContainer.cs
smoogipoo: Yeah, probably fine.

peppy:
So to clarify, when is it safe vs non-safe? Because you've said I can call it in one place but also that it's unsafe. Is it when it's called before `musicController` has called it? Is there any safety against potential track leaks in that case (considering as I said, it's already called in several places)?

smoogipoo:
It's unsafe when `MusicController` isn't going to also call it. So any time it's called when the beatmap is not the current game-wide beatmap it's potentially unsafe. That's why I was going down the path of trying to reduce the scope of this method and making everything go through `MusicController`.

I suppose it could be made safe by manually disposing when using in such cases, but there's no such automatic safety.

peppy:
Right.

So as a minimum change (ensuring some degree of safety), how about renaming `GetTrack` to `LoadTrack` and adding back the Track property, which will throw if `LoadTrack` has not yet been called? Then adding documentation to LoadTrack that specifies the consumer is responsible for disposing of the track after use.

smoogipoo:
Sounds fair.

peppy:
I've gone ahead with removing the `ResetTrackAdjustments` as proposed. It seems fine. The deadlock reports is avoided for sure, as the code that would have caused it was removed (ie. the `track.Reset()`) call.

smoogipoo:
But what was specific about `track.Reset()` that caused the deadlock? Because all I've done in my changes (not sure if you've used them) was inlining the relevant things that `Reset()` does.

peppy:
From what I can tell, it was to do with the re-application of speed adjustments, which has been removed.

peppy:
@smoogipoo i [added the track transfer back](https://github.com/ppy/osu/pull/9792/commits/083bcde3cf260147304b110d4a0956285690e4ba), make sure you are okay with this. as discussed in discord, track stores do no caching and the logic you had only managed to work because everything was routed via `MusicController.CurrentTrack`.

smoogipoo:
I suppose it's fine, it's quite scary though..

smoogipoo:
I worry that someone calling `LoadTrack()` expecting to have to dispose the track locally, then changing the global beatmap could lead to the music controller transferring the wrong (now/potentially-in-the-future disposed) track.

peppy:
Who is calling `LoadTrack()`?

smoogipoo:
In the end it's a public member of a base class that can be used/abused, and I'd say is guaranteed to be used when `Track` has the potential to throw an exception without calling `LoadTrack()` first.

I need a few more days to wrap my head around the best way to do this, it just feels flimsy.

peppy:
I'd be content with making it internal or limited to single fire, with an internal method to force reload (to account for the fact we cache `WorkingBeatmaps`)

smoogipoo:
Single fire would definitely ease most of my concerns. Not sure what you mean by force reload/when it'd be used?

peppy:
Because we cache WorkingBeatmaps, there needs to be a way for `MusicController` to forcefully reload the track if it is in a disposed state (or clear it on disposal). I guess the transfer method could be repurposed for this too, if the `[NotNull]` is removed.

smoogipoo:
Right, I was thinking down a similar path of what happens if a component is referencing the disposed track.

peppy:
I definitely agree this needs further brainstorming. `Track`s still technically have two owners and not enough safety in its current state.

I'm pretty happy with the direction and do think (after intro and DI changes) it could be merged for long-term testing. Maybe with making track access private if you're worried about consumers of the game (I don't think we should be just yet). We are competent enough during review to catch calls to `LoadTrack()`, for instance.

But it does feel like we're still doing something wrong.

I'm still gonna (hopefully) complete the remaining changes to get this in a mergeable state from my perspective.

bdach (commented):
Here's a few notes that jumped out at me, but this is a superficial review - as I said on discord I haven't really gotten close to the intricacies of the problem at hand.

Sometimes the `Beatmap` -> `MusicController` changes feel a little strange (i.e. in the multiplayer room - it reads a little weirder that the room would care about the global controller's track length instead of the beatmap's). Intro code's a little strange now too but I'm not going to pretend I could do much better without trying (and probably failing) to take a shot at this myself.

peppy:
I've reverted that usage too. I thought it made sense to leave it as `MusicController` but looking at the code in the same method using `gameBeatmap`, revert does seem better.

bdach:
I don't think I have any issues left with this any more, and I've also briefly play-tested this branch and had nothing funny happen, so going to throw in a tentative approve. Should get a second pair of eyes on this before merging though (hi @smoogipoo)

bdach (dismissed):

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
```diff
@@ -72,7 +65,7 @@ public void TestSnakingEnabled(int sliderIndex)
         {
             AddStep("enable autoplay", () => autoplay = true);
             base.SetUpSteps();
-            AddUntilStep("wait for track to start running", () => track.IsRunning);
+            AddUntilStep("wait for track to start running", () => MusicController.IsPlaying);
```
smoogipoo, line 68: Is there a reason this is still using `MusicController`? There's tests that use both `Beatmap.Value.Track` and `MusicController` and it feels kinda weird... I'm going to mark down all tests that have `MusicController`, since I believe the idea was to move back to `Beatmap.Value.Track`.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
```diff
@@ -55,7 +48,7 @@ public override void SetUpSteps()
         {
             base.SetUpSteps();
 
-            AddUntilStep("wait for track to start running", () => track.IsRunning);
+            AddUntilStep("wait for track to start running", () => MusicController.IsPlaying);
```
smoogipoo, line 51: .

osu.Game.Tests/Visual/Menus/TestSceneIntroWelcome.cs
```diff
@@ -15,11 +15,9 @@ public class TestSceneIntroWelcome : IntroTestScene
 
         public TestSceneIntroWelcome()
         {
-            AddUntilStep("wait for load", () => getTrack() != null);
-
-            AddAssert("check if menu music loops", () => getTrack().Looping);
+            AddUntilStep("wait for load", () => MusicController.TrackLoaded);
```
smoogipoo, line 18: .

osu.Game/Graphics/Containers/BeatSyncedContainer.cs
```diff
@@ -47,16 +51,16 @@ public class BeatSyncedContainer : Container
...
-            Track track = null;
+            ITrack track = null;
             IBeatmap beatmap = null;
 
             double currentTrackTime = 0;
             TimingControlPoint timingPoint = null;
             EffectControlPoint effectPoint = null;
 
-            if (Beatmap.Value.TrackLoaded && Beatmap.Value.BeatmapLoaded)
+            if (musicController.TrackLoaded && Beatmap.Value.BeatmapLoaded)
```
smoogipoo, line 61: This too, should probably use `Beatmap.Track`?

osu.Game/Overlays/Music/PlaylistOverlay.cs
```diff
@@ -82,7 +85,7 @@ private void load(OsuColour colours, Bindable<WorkingBeatmap> beatmap)
                 if (toSelect != null)
                 {
                     beatmap.Value = beatmaps.GetWorkingBeatmap(toSelect);
-                    beatmap.Value.Track.Restart();
+                    musicController.CurrentTrack.Restart();
```
smoogipoo, line 88: .

osu.Game/Overlays/NowPlayingOverlay.cs
```diff
@@ -234,9 +234,9 @@ protected override void Update()
                 pendingBeatmapSwitch = null;
             }
 
-            var track = beatmap.Value?.TrackLoaded ?? false ? beatmap.Value.Track : null;
+            var track = musicController.CurrentTrack;
```
smoogipoo, line 237: .

osu.Game/Screens/Edit/Compose/Components/Timeline/Timeline.cs
```diff
@@ -26,6 +27,31 @@ public class Timeline : ZoomableScrollContainer, IPositionSnapProvider
         [Resolved]
         private EditorClock editorClock { get; set; }
 
+        [Resolved]
+        private MusicController musicController { get; set; }
```
smoogipoo, line 31: can be removed now?

osu.Game/Screens/Menu/LogoVisualisation.cs
```diff
@@ -103,15 +107,15 @@ private void load(ShaderManager shaders, IBindable<WorkingBeatmap> beatmap)
 
         private void updateAmplitudes()
         {
-            var effect = beatmap.Value.BeatmapLoaded && beatmap.Value.TrackLoaded
-                ? beatmap.Value.Beatmap?.ControlPointInfo.EffectPointAt(beatmap.Value.Track.CurrentTime)
+            var effect = beatmap.Value.BeatmapLoaded && musicController.TrackLoaded
```
smoogipoo, line 110: .

osu.Game/Screens/Menu/MainMenu.cs
```diff
@@ -177,15 +174,14 @@ public override void OnEntering(IScreen last)
             base.OnEntering(last);
             buttons.FadeInFromZero(500);
 
-            Track = Beatmap.Value.Track;
             var metadata = Beatmap.Value.Metadata;
 
-            if (last is IntroScreen && Track != null)
+            if (last is IntroScreen && musicController.TrackLoaded)
```
smoogipoo, line 179: .

osu.Game/Screens/Menu/OsuLogo.cs
```diff
@@ -327,9 +330,9 @@ protected override void Update()
             const float velocity_adjust_cutoff = 0.98f;
             const float paused_velocity = 0.5f;
 
-            if (Beatmap.Value.Track.IsRunning)
+            if (musicController.CurrentTrack.IsRunning)
```
smoogipoo, line 333: .

smoogipoo (commented):
As I mentioned in discord, I'm really not sure I like these.

Components that use both the beatmap and the track should outright crash if the track isn't loaded, rather than silently using the music controller's (wrong) track.

osu.Game/Overlays/NowPlayingOverlay.cs
peppy: i don't think we want to revert this one, right..?

smoogipoo: This one can stay as is.

osu.Game/Screens/Menu/MainMenu.cs
peppy: `Track` doesn't exist any more. did you want it added back? i think it's kinda cleaner as it is.

osu.Game/Screens/Menu/OsuLogo.cs
peppy: Not sure about this one. Shouldn't the logo be responding to the current track, not the beatmap? it's not using the beatmap elsewhere.

smoogipoo: This one can stay as is too.

osu.Game/Screens/Menu/MainMenu.cs
smoogipoo: Let's keep as is...

osu.Game.Rulesets.Osu.Tests/TestSceneSliderSnaking.cs
peppy: This is another weird one. Can't revert because the loading is being done by the music controller in the base scene. Will have to check where the track is valid to be fetched.

peppy: Actually seems okay.

osu.Game.Tests/Visual/Menus/TestSceneIntroWelcome.cs
peppy: As below, will leave as-is.

smoogipoo:
Seems like https://github.com/ppy/osu/pull/9843 has broken this branch a bit. Since mods no longer preserve references, `ModTimeRamp` doesn't have the correct track set during gameplay...

smoogipoo:
I think this is ok, but would appreciate a second pass @peppy / @bdach , especially with my recent changes.

osu.Game/Screens/Play/Player.cs
```diff
@@ -627,6 +630,10 @@ public override void OnEntering(IScreen last)
 
             foreach (var mod in Mods.Value.OfType<IApplicableToHUD>())
                 mod.ApplyToHUD(HUDOverlay);
+
+            Beatmap.Value.Track.ResetSpeedAdjustments();
+            foreach (var mod in Mods.Value.OfType<IApplicableToTrack>())
+                mod.ApplyToTrack(Beatmap.Value.Track);
```
bdach: In the failing tests, doing this causes the adjustment to get re-applied twice because of this:

https://github.com/ppy/osu/blob/7e1844ed773368a4b932ea0e2d7fc87fa0fc53b4/osu.Game/Tests/Visual/PlayerTestScene.cs#L33

Disabling this back again locally makes them pass, but this makes me quite uneasy... Maybe that `true` override is no longer necessary with the recent changes?

peppy: i would hope it's not needed.. but also hope that it didn't break anything.

peppy:
@smoogipoo did you find the reason that was breaking things? kinda need some resolution there as it really shouldn't in my head.

smoogipoo:
I haven't, but I can imagine why. `musicController.ResetTrackAdjustments()` doesn't _just_ reset the adjustments, it also adds the MC's own adjustments via `mod.ApplyToTrack()`. So `Player` does the same and the end result is that adjustments get applied twice.
Of note, `MusicController.AllowRateAdjustments` is `false` by default, and is only set to `true` by `OsuGame` when an `OsuScreen` wants it (pretty much everywhere except Editor, Player, MainMenu).

I'm not super confident about this code

peppy:
But play is also resetting?

image

smoogipoo:
That's correct, check the implementation of `ResetTrackAdjustments`:

https://github.com/ppy/osu/blob/dda776d15f03ca5f64e41c183dfdf04398f8c031/osu.Game/Overlays/MusicController.cs#L335-L348

peppy:
ah......................... at very least let's add a mentio to the xmldoc for `ResetTrackAdjustments` that the behaviour differs depending on `AllowRateAdjustments`

peppy (approved):

PR #9793, opened 2020-08-06
Add more resilient logic for whether to avoid playing SkinnableSound on no volume

peppy:
Removes need for weird `minimum_volume` constants.

Looping samples, by default are always played because this seems like better default behaviour, but I am open to changing it to be always false unless specified.

@frenzibyte this should probably be set to `=> true` for `GameplaySkinnableSound`. I haven't added that class as you did specify you were actively working on this.


frenzibyte:
> @frenzibyte this should probably be set to => true for GameplaySkinnableSound. I haven't added that class as you did specify you were actively working on this.

`GameplaySkinnableSound` is for all skinnable samples played in gameplay, and `DrawableSpinner.spinningSample` which is suffering from the "skip play on zero volume" logic is one of them, so it has to be `=> !Looping` in there as well.

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -28,6 +28,16 @@ public class SkinnableSound : SkinReloadableDrawable
         public override bool RemoveWhenNotAlive => false;
         public override bool RemoveCompletedTransforms => false;
 
+        /// <summary>
+        /// Whether to play the underlying sample when aggregate volume is zero.
+        /// Note that this is checked at the point of calling <see cref="Play"/>; changing the volume post-play will not begin playback.
+        /// Defaults to false unless <see cref="Looping"/>.
+        /// </summary>
+        /// <remarks>
+        /// Can serve as an optimisation if it is known ahead-of-time that this behaviour will not negatively affect behaviour.
```
bdach, line 37: this reads a little loopy (*behaviour will not negatively affect behaviour*). How about something like

```suggestion
        /// Can serve as an optimisation if it is known ahead-of-time that this behaviour is allowed in a given use case.
```

(also not 100% sure what sort of optimisation would this be but it might just be that it's late and my brain is gone).

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -28,6 +28,16 @@ public class SkinnableSound : SkinReloadableDrawable
...
+        /// </remarks>
+        protected bool SkipPlayWhenZeroVolume => !Looping;
```
bdach, line 39: Not sure if I like this being a *negation* in nature (by which I mean this flag is `true` if samples with 0 volume should *not* be played). Kinda think it makes things harder to read (esp. the `!SkipPlayWhenZeroVolume` check - gotta flip it twice in my head to determine what will happen); I'd probably flip semantics and make it a `PlayWhenZeroVolume`.

osu.Game/Skinning/SkinnableSound.cs
peppy: It's a bit of an odd one.

- Doesn't play samples at a bass level when fast forwarding (helps with performance)
- Doesn't play zero volume samples (causing "clicks" for some audio hardware)

bdach (approved):
looks good to go :+1:

smoogipoo (approved):

PR #9816, opened 2020-08-10
Add context menu option for duplicating multi rooms

### LINKED ISSUE #9589: Create a multiplayer room with a list of a closed one.
**Describe the new feature:** 
A quick way to copy a previously created room.

voidedWarranties:
Implements and closes #9589

- [x] Depends on ppy/osu-framework#3805 for tests to pass reliably 

Allows duplicating a multi room by right clicking the room in the lounge

Might need an alternative to the current approach of passing actions around


osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -51,7 +51,7 @@ public class Multiplayer : OsuScreen, IOnlineComponent
         [Cached]
         private readonly Bindable<FilterCriteria> currentFilter = new Bindable<FilterCriteria>(new FilterCriteria());
 
-        [Resolved]
+        [Resolved(CanBeNull = true)]
```
bdach, line 54: mind explaining these `MusicController` changes? they don't seem to have much to do with the actual functionality added in this PR.

osu.Game/Online/Multiplayer/Room.cs
```diff
@@ -103,38 +103,48 @@ private int duration
         [JsonIgnore]
         public readonly Bindable<int> Position = new Bindable<int>(-1);
 
-        public void CopyFrom(Room other)
+        /// <summary>
+        /// Copies the properties from another <see cref="Room"/> to this room.
+        /// </summary>
+        /// <param name="other">The room to copy</param>
+        /// <param name="duplicate">Whether the copy should exclude information unique to a specific room (i.e. when duplicating a room)</param>
+        public void CopyFrom(Room other, bool duplicate = false)
```
bdach, line 111: I really dislike this `duplicate` param. maybe the actual "duplicating" functionality should be an entirely separate method? something like `CreateCopyOf(Room)`?

osu.Game/Screens/Multi/Lounge/LoungeSubScreen.cs
```diff
@@ -62,7 +62,18 @@ private void load()
...
                                     Padding = new MarginPadding(10),
-                                    Child = roomsContainer = new RoomsContainer { JoinRequested = joinRequested }
+                                    Child = roomsContainer = new RoomsContainer
+                                    {
+                                        JoinRequested = joinRequested,
+                                        DuplicateRoom = room =>
+                                        {
+                                            Room newRoom = new Room();
+                                            newRoom.CopyFrom(room, true);
+                                            newRoom.Name.Value = $"Copy of {room.Name.Value}";
```
bdach, line 72: we might want this to be a bit smarter, to avoid a *Copy of Copy of Copy of (...) __[actual room name]__* sort of situation

osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
```diff
@@ -88,6 +96,7 @@ private void addRooms(IEnumerable<Room> rooms)
             {
                 roomFlow.Add(new DrawableRoom(room)
                 {
+                    DuplicateRoom = DuplicateRoom,
```
bdach, line 99: is there some reason I'm missing as to why this delegate is passed down a level instead of declared locally here?

osu.Game/Screens/Multi/Lounge/Components/DrawableRoom.cs
```diff
@@ -232,5 +236,10 @@ private void load()
                 Current = name;
             }
         }
+
+        public MenuItem[] ContextMenuItems => new MenuItem[]
+        {
+            new OsuMenuItem("Duplicate", MenuItemType.Standard, () => DuplicateRoom?.Invoke(Room))
```
bdach, line 242: not sure with naming this option *Duplicate* - would prefer something like *Create copy* myself

osu.Game/Screens/Multi/Multiplayer.cs
voidedWarranties: The `MultiScreen` test scene would crash immediately after selecting it since `MusicController` is not available, and again during navigation in `LoungeSubScreen`

osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
voidedWarranties: The problem was getting access to `LoungeSubScreen.Open`, I was unsure on the best way to do that so I passed the delegate down 2 levels

My second idea was to cache `LoungeSubScreen` and resolve it in `RoomsContainer`, would that be better?

bdach: caching would be preferable as far as I'm concerned, yeah.

osu.Game/Screens/Multi/Lounge/Components/DrawableRoom.cs
```diff
@@ -36,6 +38,8 @@ public class DrawableRoom : OsuClickableContainer, IStateful<SelectionState>, IF
 
         public event Action<SelectionState> StateChanged;
 
+        public Action DuplicateRoom;
```
bdach, line 41: Not sure about having this as a publically settable action and setting it outwardly rather than locally DI'ing whatever is necessary here and setting the action within. Is this ever going to have a different implementation?

Also, all references to "duplicating rooms" in code should be also renamed to `CreateCopy` or similar if we're renaming the action on the UI.

osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
```diff
@@ -88,6 +97,14 @@ private void addRooms(IEnumerable<Room> rooms)
             {
                 roomFlow.Add(new DrawableRoom(room)
                 {
+                    DuplicateRoom = () =>
+                    {
+                        Room newRoom = room.CreateCopy();
+                        if (!newRoom.Name.Value.StartsWith("Copy of "))
+                            newRoom.Name.Value = $"Copy of {room.Name.Value}";
```
bdach, line 104: this should be inside `CreateCopy`

osu.Game/Online/Multiplayer/Room.cs
```diff
@@ -103,6 +103,24 @@ private int duration
         [JsonIgnore]
         public readonly Bindable<int> Position = new Bindable<int>(-1);
 
+        /// <summary>
+        /// Create a copy of this room, without information specific to it, such as Room ID or host
```
bdach, line 107:
```suggestion
        /// Creates a new room whose name, settings and playlist are based on this room instance,
        /// but without copying data which should be unique to this instance such as room, channel and host IDs.
```

bdach (changes requested):

osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
smoogipoo: Also this shouldn't be "Copy of ...". This should be the "{user}'s awesome room" text.

peppy: For the time being I'm going to copy the name from the original room. It's up to the user to change this if they want to, in my opinion.

peppy (approved):

PR #9818, opened 2020-08-10
Improve messaging when timeshift token retrieval fails

### LINKED ISSUE #9813: Improve error message upon failing to retrieve multiplayer score submission token
**Describe the bug:** 
When trying to play multiplayer, the error notification appears and the map doesn't load.

Using arch linux with Xorg, and i3 as window managaer. Lazer compiled using '[pikaur](https://github.com/actionless/pikaur)' and the '[osu-lazer-git](https://aur.archlinux.org/packages/osu-lazer-git/)' package from AUR.

https://youtu.be/t_1DFnhVumc

**osu!lazer version:**
Lazer 2020.806.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/5045983/database.log)
[network.log](https://github.com/ppy/osu/files/5045984/network.log)
[performance.log](https://github.com/ppy/osu/files/5045985/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5045986/runtime.log)

peppy:
Obviously not a final solution, but should better help self-compiling
(or unofficial package) users better understand why this is happening.

- Closes #9813

### DIFF
--- a/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
+++ b/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
@@ -64,7 +64,7 @@ namespace osu.Game.Screens.Multi.Play
             {
                 failed = true;
 
-                Logger.Error(e, "Failed to retrieve a score submission token.");
+                Logger.Error(e, "Failed to retrieve a score submission token.\n\nThis may happen if you are not running an official release of osu! (ie. you are self-compiling).");
 
                 Schedule(() =>
                 {
--- a/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
+++ b/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
@@ -64,7 +64,7 @@ namespace osu.Game.Screens.Multi.Play
             {
                 failed = true;
 
-                Logger.Error(e, "Failed to retrieve a score submission token.\n\nThis may happen if you are not running an official release of osu! (ie. you are self-compiling).");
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
```diff
@@ -64,7 +64,7 @@ private void load()
             {
                 failed = true;
 
-                Logger.Error(e, "Failed to retrieve a score submission token.");
+                Logger.Error(e, "Failed to retrieve a score submission token.\n\nThis may happen if you are not running an official release of osu! (ie. you are self-compiling).");
```
bdach, line 67: Wondering whether we want an android-specific message too as I imagine someone will inevitably roll along and ask why downloading the official APK causes this to pop up.

peppy: android should work fine. there's already server-side allowance for this (see https://github.com/ppy/osu-web/pull/6449).

bdach (dismissed):
works good :+1:

osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
bdach: Oh, right. Missed that as I don't look at the web channel that often.

smoogipoo, line 67: This should probably also mention running an old version of osu!?

smoogipoo (changes requested):

smoogipoo (approved):

PR #9819, opened 2020-08-10
Always show newly presented overlay at front

### LINKED ISSUE #9815: Overlay pop in/out animations have inconsistent ordering
**Describe the bug:** The News overlay has inconsistent ordering and always opens under other overlays.

**Screenshots or videos showing encountered issue:** https://youtu.be/h4xxZOPUVZI
Notice that when opening the overlays from right to left, the next overlay always go over the previous, except for News.

**osu!lazer version:** 2020.806.0

peppy:
This feels much better. Does not change order if the overlay to be shown
is not yet completely hidden.

- Closes #9815.

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -702,6 +702,9 @@ namespace osu.Game
                     if (state.NewValue == Visibility.Hidden) return;
 
                     singleDisplayOverlays.Where(o => o != overlay).ForEach(o => o.Hide());
+
+                    if (!overlay.IsPresent)
+                        overlayContent.ChangeChildDepth(overlay, (float)-Clock.CurrentTime);
                 };
             }
 


bdach (approved):
Interesting solution... It does work, and `float.MinValue` milliseconds works out to about 10<sup>28</sup> years, so seeing you'd have to run the game presumably longer than the remaining lifetime of the Solar System to see this fail, I think this is good to go.

(This is leaving aside the possibilities of loss of precision on the way there, but again, I think any realistic timespan is unlikely to see this fail.)

smoogipoo (approved):

Joehuu:
This regressed `informationalOverlays`' depth (i.e. user profile and beatmap info). For example, when clicking a username or map in beatmap listing, their overlay is shown below the beatmap listing.

OE0HvnoXiQ

bdach:
Good point... It seems that the only thing required to fix that though is to reapply the same technique to the informational overlays.

PR #9842, opened 2020-08-13
Beatmap selection screen regex enhancements

ghost:
I wasn't finding current behavior of regex search convenient enough so I decided to enhance it a little, so now instead of doing this:
image
You now can do this:
image
I find such a way of searching very convenient and would like to see it in the main tree, so what do you think?

### DIFF
--- a/osu.Game.Tests/NonVisual/Filtering/FilterQueryParserTest.cs
+++ b/osu.Game.Tests/NonVisual/Filtering/FilterQueryParserTest.cs
@@ -169,7 +169,7 @@ namespace osu.Game.Tests.NonVisual.Filtering
             Assert.AreEqual(3, filterCriteria.SearchTerms.Length);
             Assert.AreEqual("name with space", filterCriteria.Artist.SearchTerm);
         }
-
+    
         [Test]
         public void TestApplyArtistQueriesOneDoubleQuote()
         {
@@ -179,6 +179,22 @@ namespace osu.Game.Tests.NonVisual.Filtering
             Assert.AreEqual("weird", filterCriteria.SearchText.Trim());
             Assert.AreEqual(1, filterCriteria.SearchTerms.Length);
             Assert.AreEqual("double\"quote", filterCriteria.Artist.SearchTerm);
+        }
+
+        [Test]
+        public void TestApplyDoubleConditionQueries()
+        {
+            const string query = "give me the maps with stars>=6<8 please";
+            var filterCriteria = new FilterCriteria();
+            FilterQueryParser.ApplyQueries(filterCriteria, query);
+            Assert.AreEqual("give me the maps with  please", filterCriteria.SearchText.Trim());
+            Assert.AreEqual(1, filterCriteria.SearchTerms.Length);
+            Assert.IsNotNull(filterCriteria.StarDifficulty.Max);
+            Assert.Greater(filterCriteria.StarDifficulty.Max, 7.99d);
+            Assert.Less(filterCriteria.StarDifficulty.Max, 8.00d);
+            Assert.IsNotNull(filterCriteria.StarDifficulty.Min);
+            Assert.Greater(filterCriteria.StarDifficulty.Max, 5.99d);
+            Assert.Less(filterCriteria.StarDifficulty.Max, 6.00d);
         }
     }
 }
--- a/osu.Game.Tests/NonVisual/Filtering/FilterQueryParserTest.cs
+++ b/osu.Game.Tests/NonVisual/Filtering/FilterQueryParserTest.cs
@@ -184,17 +184,17 @@ namespace osu.Game.Tests.NonVisual.Filtering
         [Test]
         public void TestApplyDoubleConditionQueries()
         {
-            const string query = "give me the maps with 4<stars<6 please";
+            const string query = "give me maps with 6<stars<8 please";
             var filterCriteria = new FilterCriteria();
             FilterQueryParser.ApplyQueries(filterCriteria, query);
-            Assert.AreEqual("give me the maps with   please", filterCriteria.SearchText.Trim());
-            Assert.AreEqual(1, filterCriteria.SearchTerms.Length);
+            Assert.AreEqual("give me maps with  please", filterCriteria.SearchText.Trim());
+            Assert.AreEqual(5, filterCriteria.SearchTerms.Length);
             Assert.IsNotNull(filterCriteria.StarDifficulty.Max);
             Assert.Greater(filterCriteria.StarDifficulty.Max, 7.99d);
             Assert.Less(filterCriteria.StarDifficulty.Max, 8.00d);
             Assert.IsNotNull(filterCriteria.StarDifficulty.Min);
-            Assert.Greater(filterCriteria.StarDifficulty.Max, 5.99d);
-            Assert.Less(filterCriteria.StarDifficulty.Max, 6.00d);
+            Assert.Greater(filterCriteria.StarDifficulty.Min, 6.00d);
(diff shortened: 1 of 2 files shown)


bdach:
This needs test coverage, and the regex should be split out of where you put it as it's unreadable.

claike:
Did not account for quite a few things, sorry. I'll fix that.

jorolf:
Have you considered using a `a < x < b` syntax? That's how it's usually written mathematically.

claike:
I did not, but that is a good idea. I'll work on it as soon as I fix all the issues with the current implementation.

claike:
I think I'm going in the right direction. Now I'm even able to detect constructions such as «artist="someone"5\<stars>=4» correctly and not detect ones such as «stars="something"». I decided to split fields that can contain any text values such as "artist", "creator" and "status" and fields that can only contain digits (and metric prefixes such as ms, s, m etc.). And so far this works pretty good. I did not run tests yet because they take significant amount of time to compute. I'm not a dotnet dev so I'm very curious how to run just the tests described in the "osu.Game.Tests/NonVisual/Filtering/FilterQueryParserTest.cs", can someone please explain me how to do this? Again, can't afford running all of the tests on my machine.
изображение

bdach:
How are you editing code? Any IDE should be able to just selectively run just the tests in `FilterQueryParserTest` (and so can `dotnet test`). Also the whitespace issues you seem to be struggling with should be indicated in rider or resharper. (If you don't own either, you can also use the `InspectCode.ps1` script in the root of this repository on Windows).

That said looking at the complexity of the newest changes I'm becoming doubtful of whether we want to maintain this going forward. Unless this can be broke down into constituent parts to be legible maybe we want a simpler interval notation like `[4,5)`/`[4,)`/`(,5)` or similar (open brace is exclusive bound, square bracket is inclusive).

claike:
Well... I'm on linux and using vim. Generally YouCompleteMe is good enough highlighting mistakes and errors, but I don't think it can monitor every wrongly put whitespace character and tell me about them:) I have eclipse which I use for my java stuff but I'm not sure if it's going to work with c# the way you need it to.
This exact regex if not finished yet, and I'm thinking of the ways to simplify it if possible. It is still going to be somewhat big and look bulky in order to reach such functionality, and my question is: would it suit the needs of the future maintainers if I just write a full detailed explanation of the regex in the comments after I finish tinkering with it?

bdach:
`vim` is pretty hardcore, all things considered... You might struggle going forward with that setup. In that case perhaps setting up appveyor on a fork and delegating CI builds to it might be the best choice for you when it comes to code formatting. (Tests you should be able to run locally still, all things considered.)

As for the regex - a proper explanation in the PR description would go a long way to help review, but doesn't guarantee that we'll be fine for it; it might turn out that to work properly/reliably/be maintainable the current query parser should be rewritten to use a more conventional lexer-parser structure and a context-free grammar definition to go along with it.

claike:
This is my explanation of the regex and why it has to be this way.

```
//Keys with text are matched here:
(
    //Capture one of the keys that can contain textual values into this group.
    (?<keyText>(status|creator|artist)\b)
    
    //Since text can't be compared using operators such as less or more we only allow capturing
    //one of [=:] operators.
    (?<opText>[=:])
    
    //Every word or any sequence of characters enclosed in double quotes gets
    //captured in this group
    (?<valueText>("".*?"")|(\S*))
) //End of textual match
| //OR
//Keys with numbers are matched here:
(
    (
        //Capture every match that contains an integral part of the number and/or
        //a single dot separator and/or fractional part of the number
        //and/or a quantifier (seconds, minutes, hours, days, years) to this group.
        //This is done so we don't capture whole constructions such as e.g. 
        //"15.3m.42e.163" and only capture the meaningful part which 15.3m in this case.
        (?<valueBeforeNumerical>([0-9]+\.?[0-9]*[mshdy]?))
    
        //Capture operator to this group. 
        //First we match one of [=:><] operators, then we check if the previous
        //operator is one of [><], and if it is, then we may want to look for one of [=:] 
        //operators in order to capture constructions such as >= <= = : > < and not
        //capture => and =< constructions.
        (?<opBeforeNumerical>([=:><](?(?<=[><])([=:]?))))
    )? //We may not have this part so we make it optional.
    
    //Capture one of the keys that can contain numerical values into the KeyNumerical group.
    //We should have this part.
    (?<keyNumerical>(stars|ar|dr|cs|divisor|length|objects|bpm)\b)
    (
        //Same as the opBeforeNumerical.
        (?<opAfterNumerical>([=:><](?(?<=[><])([=:]?))))
        //Same as the valueBeforeNumerical.
        (?<valueAfterNumerical>([0-9]+\.?[0-9]*[mshdy]?))
    ) // And we should have at least this part.
) //End of numerical match
```
image

I renamed and changed a few bits here and there, but it is not in the code yet.

bdach:
That explanation doesn't ease my concerns too much.

* Having number/operator token recognition duplicated in the `before`/`after` groups is prone to mistakes in the future. This I don't think can be resolved without having token recognition split from interpretation by using a lexer as mentioned before, but I suppose it's at the limit of acceptable right now.
* That operator recognition fragment sounds way convoluted what with the lookbehind expressions. Couldn't it just explicitly match all possible combinations instead of having to tack on the `=` sometimes?

claike:
Yep, I also don't see the way to avoid using `before`/`after` with just the regex either. But at the same time in the current state it is good enough and I think this solution does better overall then the previous one. And if there will be a need to make a better query parser you will probably rewrite the whole thing in a lexer-parser manner and reject this solution, which will be even better, but for now there is actually no need for it. :)

As of the lookbehinds. Yesterday I tried a simple match of all possible combinations but for some reason it took way more steps to compute. About 1200 more which I considered weird and immediately rejected. But now when I check it once again it works about as good. I think the problem was that I had a long space-less character sequence (as seen there https://github.com/ppy/osu/pull/9842#issuecomment-674031672) and this is probably not going to be the case most of the time. I'm still not sure which one is better, I personally don't find the lookbehinds complicated or convoluted. Also I don't think the individual who's going to maintain this part of the code will be the one not familiar with regular expressions, there will also be an explanation that is supposed to ease the pain reading it ¯\\\_(ツ)_/¯

bdach:
You have a point in that the lookbehinds might be slightly more optimal but I don't think it's worth both the readability cost. Unless it's shown that doing it the obvious way is orders of magnitude slower than the non-obvious way I'd prefer the former.

peppy:
This doesn't seem to be blocked on anything, and apart from maybe missing more test coverage (ie `>` and `>=` where relevant) seems like something we could get in. @claike are you still interested in fixing this one up? It seems like the remaining tasks are just filling in a `//TODO` comment.

bdach:
Sure you're fine with the positively enormous/arcane regex in the diff? Because I'm not entirely sure I am.

peppy:
I'm not even looking to understand it, but as long as it's tested and does the job I don't really have any issue with it either.

As it's very self contained code, we can switch away from regex to something else if/when that need appears.

bdach:
Well, the current regex in this diff has at least two errors (partly caused by duplication of subexpressions) and is actually different than the one in the "explanation" comment above, so this isn't really tested - I'm not even sure the pull has even been ran through tests on CI yet. And the test coverage included is not exhaustive.

peppy:
Closing for now may be the best option then. This is a pretty low priority / edge case that doesn't really need attention at the moment, was just looking to either close or quick merge.

bdach:
Given that this is PR'd off the submitter's `master` and has merge conflicts I'll skip trying to fix it up and just close for now.

PR #9843, opened 2020-08-13
Fix mod instances being shared between multiplayer playlist items

### LINKED ISSUE #9727: Adjusting a mod's settings while in a playlist affects all songs in the playlist
**Describe the bug:** 
If you have multiple beatmaps in a playlist with the same customizable mod (Nightcore, Double Time, Wind Up, etc.) changing the setting in the "customize" panel changes it for all other beatmaps.

Ex. If you have three beatmaps in the playlist with Nightcore, and you then set Nightcore to 1.7x speed, all other beatmaps with Nightcore now have 1.7x speed as well. Or at least, they appear to.

https://streamable.com/36qtjj

**osu!lazer version:** 
commit https://github.com/ppy/osu/commit/e60c085bf2e6889bcc51ca6b47b3a50e460a5b8c

**Logs:**
[database.log](https://github.com/ppy/osu/files/5002449/database.log)
[network.log](https://github.com/ppy/osu/files/5002450/network.log)
[performance.log](https://github.com/ppy/osu/files/5002451/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5002452/runtime.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9727

### DIFF
--- a/osu.Game.Tests/Visual/Multiplayer/TestSceneMatchSongSelect.cs
+++ b/osu.Game.Tests/Visual/Multiplayer/TestSceneMatchSongSelect.cs
@@ -16,7 +16,9 @@ using osu.Framework.Utils;
 using osu.Game.Beatmaps;
 using osu.Game.Online.Multiplayer;
 using osu.Game.Rulesets;
+using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Osu;
+using osu.Game.Rulesets.Osu.Mods;
 using osu.Game.Screens.Multi.Components;
 using osu.Game.Screens.Select;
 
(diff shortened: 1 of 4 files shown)


peppy:
I think this is fine, but an alternative may be still calling `MemberwiseClone()` then replacing only the bindables on the new instance, rather than manually doing the full clone procedure (might read better). Thoughts?

smoogipoo:
That’s also the way I started it but doesn’t work because the properties are readonly so I’d need to find the backing fields and replace them. This is neater because we already have methods to find the properties.

peppy:
right, fair enough.

PR #9849, opened 2020-08-14
Fix beatmap covers not being unloaded in most overlays

peppy:
Eventually we'll probably want something smarter than this, but for the time being this helps stop runaway memory usage. Most noticeable in Beatmap Listing.

### DIFF
--- a/osu.Game/Beatmaps/Drawables/UpdateableBeatmapSetCover.cs
+++ b/osu.Game/Beatmaps/Drawables/UpdateableBeatmapSetCover.cs
@@ -67,19 +67,18 @@ namespace osu.Game.Beatmaps.Drawables
 
             if (beatmapSet != null)
             {
-                BeatmapSetCover cover;
-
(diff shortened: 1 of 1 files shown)


smoogipoo:
Interesting, for some reason some textures aren't cleaned up...

image

peppy:
Potentially because the overlay has since closed? Can you confirm they do if you search for nothing?

peppy:
Actually even after closing it's fine for me. Are you on local-framework? If not, try hitting the clear caches button?

smoogipoo:
Yeah still happens:

image

I've had to scroll quite a way's down for this. Maybe an issue with `DelayedLoadWrapper`?

peppy:
That's worrying. Are they stuck in that state forever?

smoogipoo:
Oof. Looks like this is due to scheduler-hell in DLW/DLUW.

Will merge this PR since it works in theory, it's just that DLW/DLUW is bugged.

smoogipoo (approved):

PR #9854, opened 2020-08-14
Add "Metronome" mod

### LINKED ISSUE #9569: Metronome as a standard element of the game
**Describe the new feature:** 
When using the NC mod, we hear extra bits that work like a metronome. Maybe it’s worth adding the ability to turn on/off the metronome for NC and other modes? It’s more convenient for someone to play without these “duts-duts-duts”, and for some it helps to keep the rhythm.
Let it work the same way as for NC.

BananeVolante:
- Resolves #9569 

I am unsure about the compatibility with nightcore. Metronome and nightcore work together, but it sounds bad. I could make them incompatible, remove nightcore's samples and add it as an option in metronome or i could try to make this mod work differently when used with nightcore.

For now : 
Tick sounds use the nightcore hat sound and the mod icon use modAuto. 
The score multiplier is at 0 and this mod is unranked.


osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,89 @@
...
+using osu.Framework.Bindables;
+using osu.Game.Audio;
+
+namespace osu.Game.Rulesets.Mods
+{
+    public abstract class ModMetronome : Mod
+    {
+        public override string Name => "Metronome";
+        public override string Acronym => "MT";
+        public override IconUsage? Icon => OsuIcon.ModAuto; // temporary icon
```
bdach, line 23:
```suggestion
        public override IconUsage? Icon => FontAwesome.Solid.TachometerAlt;
```
probably closest you can get to an actual metronome

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,89 @@
...
+        [SettingSource("Tick frequency", "Number of metronome ticks per beat")]
+        public Bindable<TickFrequency> Frequency { get; } = new Bindable<TickFrequency>(TickFrequency.One);
+    }
+
+    public abstract class ModMetronome<TObject> : ModMetronome, IApplicableToDrawableRuleset<TObject>
+        where TObject : HitObject
+    {
+        public void ApplyToDrawableRuleset(DrawableRuleset<TObject> drawableRuleset)
+        {
+            drawableRuleset.Overlays.Add(new MetronomeBeatContainer(Frequency.GetBoundCopy()));
```
bdach, line 45: Generally for safety you would call `GetBoundCopy()` in the receiver, not pass it down like this. That ensures a lack of memory leaks due to possibly keeping references alive through the callback closure.

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,89 @@
...
+                    Divisor = (int)tickFrequency.Value;
+                }, true);
+            }
+
+            private int? firstBeat;
+
+            [BackgroundDependencyLoader]
+            private void load()
+            {
+                InternalChild = metronomeSample = new SkinnableSound(new SampleInfo("nightcore-hat")); // temporary sample
```
bdach, line 65: Metronomes sometimes make a different sound on the first beat of the bar than on the other ones ([see video example](https://www.youtube.com/watch?v=PXy0l68vptg)). Could be neat to have that as a customisation option at least. The quadruple metronome feels really distracting otherwise (especially with high BPM).

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,89 @@
...
+using osu.Game.Rulesets.Objects;
+using osu.Game.Rulesets.UI;
+using osu.Game.Skinning;
+using osu.Game.Configuration;
+using osu.Framework.Bindables;
+using osu.Game.Audio;
+
+namespace osu.Game.Rulesets.Mods
+{
+    public abstract class ModMetronome : Mod
```
bdach, line 19: Would agree with marking as incompatible with nightcore.

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,89 @@
...
+                    firstBeat = null;
+                    return;
+                }
+
+                if (!firstBeat.HasValue || beatIndex < firstBeat)
+                    // decide on a good starting beat index if once has not yet been decided.
+                    firstBeat = beatIndex < 0 ? 0 : (beatIndex / segmentLength + 1) * segmentLength;
+
+                if (beatIndex >= firstBeat)
+                    metronomeSample.Play();
```
bdach, line 85: This is pretty much a copy of the implementation of `NightcoreBeatContainer`. The two differ only in the samples they play - you could make one generic container that fits both purposes and does not duplicate this code.

bdach (changes requested):

osu.Game.Rulesets.Catch/CatchRuleset.cs
```diff
@@ -98,7 +98,8 @@ public override IEnumerable<Mod> GetModsFor(ModType type)
                     {
                         new CatchModEasy(),
                         new CatchModNoFail(),
-                        new MultiMod(new CatchModHalfTime(), new CatchModDaycore())
+                        new MultiMod(new CatchModHalfTime(), new CatchModDaycore()),
+                        new CatchModMetronome(),
```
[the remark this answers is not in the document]
BananeVolante: Nightcore offer something similar to a metronome, and is ranked, so i don't have any problems settings this mod as ranked, with a small score reduction.

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,97 @@
...
+        {
+            One = 1,
+            Two = 2,
+            Four = 4
+        }
+
+        [SettingSource("Tick frequency", "Number of metronome ticks per beat")]
+        public Bindable<TickFrequency> Frequency { get; } = new Bindable<TickFrequency>(TickFrequency.One);
+
+        [SettingSource("Different Sound on the first beat of a bar")]
```
bdach, line 42:
```suggestion
        [SettingSource("Highlight the first beat of each bar", "A different sample will be played on the first beat.")]
```

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,97 @@
...
+        public Bindable<bool> SpecialSampleForFirstBeatOfBar { get; } = new Bindable<bool>(true);
```
bdach, line 43:
```suggestion
        public Bindable<bool> HighlightFirstBeat { get; } = new Bindable<bool>(true);
```

(+ usages)

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,97 @@
...
+                    {
+                        Frequency = { Value = 2.0f }
+                    },
+                    otherBeatOfBarSample = new SkinnableSound(new SampleInfo("nightcore-hat")),
+                };
+            }
+
+            protected override void PlayOnBeat(int beatIndex, TimeSignatures signature)
+            {
+                if ((beatIndex % ((int)signature * Divisor)) == 0 && specialSampleForFirstBeatOfBar)
```
bdach, line 90: inner braces are redundant

```suggestion
                if (beatIndex % ((int)signature * Divisor) == 0 && specialSampleForFirstBeatOfBar)
```

osu.Game/Graphics/Containers/SoundOnBeatContainer.cs
```diff
@@ -0,0 +1,42 @@
...
+                firstBeat = beatIndex < 0 ? 0 : (beatIndex / segmentLength + 1) * segmentLength;
+
+            if (beatIndex >= firstBeat)
+                PlayOnBeat(beatIndex % segmentLength, timingPoint.TimeSignature);
+        }
+
+        //This function will be executed on every beat of the track
+        protected virtual void PlayOnBeat(int beatIndex, TimeSignatures signature)
+        {
+        }
```
bdach, line 40: please actually xmldoc this method, with an explanation of what the parameters mean.

would also argue that this should be `abstract`, as what's the point of deriving this class without overriding this method

bdach (changes requested):

peppy:
This doesn't seem to play any sounds.

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,97 @@
...
+        public override Type[] IncompatibleMods => base.IncompatibleMods.Append(typeof(ModNightcore)).ToArray();
+
+        public enum TickFrequency
+        {
+            One = 1,
+            Two = 2,
+            Four = 4
+        }
+
+        [SettingSource("Tick frequency", "Number of metronome ticks per beat")]
```
peppy, line 39: i don't understand this setting. it's not actually doing what it says (and only delays the "accent" beat more?)

osu.Game/Rulesets/Mods/ModMetronome.cs
```diff
@@ -0,0 +1,97 @@
...
+        public Bindable<TickFrequency> Frequency { get; } = new Bindable<TickFrequency>(TickFrequency.One);
+
+        [SettingSource("Highlight the first beat of each bar", "A different sample will be played on the first beat.")]
```
peppy, line 42: Accent sounds better than highlight

peppy (changes requested):

bdach:
The sounds here broke with the merge of #10619. After I fixed that though and played a track to hear how this behaves,  I'm gonna be honest, I'm not sure I see the point. It's so audibly busy that it's almost unplayable and not a difficulty reduction at all, so I'm going to abstain from putting more time into this unless @BananeVolante wants to continue.

peppy:
I agree it should probably be in "fun" instead of difficulty reduction. And potentially, to make it actually interesting as a mod, force the audio track to be muted?

BananeVolante:
I'm sorry, but i no longer have the time to work on this.
The idea sounded good, but the result isn't as interesting as expected.
If you want to close this or change it to make it more useful, I'm fine with it. However, i won't be able to contribute.

PR #9867, opened 2020-08-15
Fix pause menu selection not resetting on visibility change

### LINKED ISSUE #9856: Multiple selection in pause screen
**Describe the bug:** 
- Pause the game
- Select and continue the game with your keyboard (down arrow)
- Try to pause shortly after again
- Select the exit option with the keyboard (up arrow)
- The game selects both continue and exit
- Press enter and the game will continue rather than exit.

osu_2020-08-15_01-33-09

**osu!lazer version:** 
2020.806.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5077797/runtime.log)

bdach:
Resolves #9856.

# Summary

This is actually a pretty old regression from back in December (#7149). The following changes in `GameplayMenuOverlay` were applied:

* the private property backing field `_selectionIndex` was renamed to `selectionIndex`,
* the private property `selectionIndex` was removed in favour of having reads come from `selectionIndex` (which is now a field) and writes through `setSelection`.

However, the following write access to `selectionIndex` was missed:

https://github.com/ppy/osu/blob/d42a4e943ab1d158cbbbda99fa61709bf1c23154/osu.Game/Screens/Play/GameplayMenuOverlay.cs#L62

and because `selectionIndex` changed from a setter with side effects to a dumb property, the behaviour changed.

The fix is a refactor of the entire button container to encapsulate the index better. A visual test has also been added to cover the regression in the future.

# Remarks

Initially I was just going to fix the one access but realised the overall interface kinda sucks and being able to mistakenly set the index without the side effects is bad, so I went a bit more enterprise:tm:. I sort of think it makes the code look nice though, so maybe not all that bad.

I dropped some `internal` access modifiers since we don't like those (dynamic compilation).

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneGameplayMenuOverlay.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneGameplayMenuOverlay.cs
@@ -272,7 +272,21 @@ namespace osu.Game.Tests.Visual.Gameplay
             AddAssert("Overlay is closed", () => pauseOverlay.State.Value == Visibility.Hidden);
         }
 
+        [Test]
+        public void TestSelectionResetOnVisibilityChange()
+        {
+            showOverlay();
+            AddStep("Select last button", () => InputManager.Key(Key.Up));
+
+            hideOverlay();
+            showOverlay();
+
+            AddAssert("No button selected",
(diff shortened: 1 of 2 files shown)


peppy (approved):
looks good

PR #9869, opened 2020-08-15
Add support for SpinnerFrequencyModulate skin config option 

### LINKED ISSUE #9738: SpinnerFrequencyModulate skin setting is not implemented
**Describe the new feature:** 
Implement the SpinnerFrequencyModulate skin setting

Turn off SpinnerFrequencyModulate in the skin for the spinner sound to not pitch up.

**osu!lazer version:** 
2020.801.0
**Logs:**
[performance.log](https://github.com/ppy/osu/files/5008454/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5008455/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5008456/updater.log)
[database.log](https://github.com/ppy/osu/files/5008457/database.log)
[network.txt](https://github.com/ppy/osu/files/5008471/network.txt)

ckiee:
Closes #9738.
VSCode helpfully tried to turn it back into tabs when I tried to put spaces instead so that's the reason for the double commit, squash if you wish.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
@@ -85,8 +85,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
         }
 
         private SkinnableSound spinningSample;
-        private const float SPINNING_SAMPLE_INITIAL_FREQUENCY = 1.0f;
-        private const float SPINNING_SAMPLE_MODULATED_BASE_FREQUENCY = 0.5f;
+        private const float spinning_sample_initial_frequency = 1.0f;
+        private const float spinning_sample_modulated_base_frequency = 0.5f;
 
         protected override void LoadSamples()
         {
@@ -106,7 +106,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 {
                     Volume = { Value = 0 },
                     Looping = true,
-                    Frequency = { Value = SPINNING_SAMPLE_INITIAL_FREQUENCY }
+                    Frequency = { Value = spinning_sample_initial_frequency }
                 });
             }
         }
@@ -233,7 +233,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 RotationTracker.Tracking = !Result.HasResult && (OsuActionInputManager?.PressedActions.Any(x => x == OsuAction.LeftButton || x == OsuAction.RightButton) ?? false);
 
             if (spinningSample != null && spinnerFrequencyModulate)
-                spinningSample.Frequency.Value = SPINNING_SAMPLE_MODULATED_BASE_FREQUENCY + Progress;
+                spinningSample.Frequency.Value = spinning_sample_modulated_base_frequency + Progress;
         }
 
         protected override void UpdateAfterChildren()
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
@@ -104,6 +104,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 {
                     Volume = { Value = 0 },
                     Looping = true,
+                    Frequency = { Value = 1.0f }
                 });
             }
         }
@@ -228,8 +229,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             if (HandleUserInput)
                 RotationTracker.Tracking = !Result.HasResult && (OsuActionInputManager?.PressedActions.Any(x => x == OsuAction.LeftButton || x == OsuAction.RightButton) ?? false);
 
-            if (spinningSample != null)
-                spinningSample.Frequency.Value = spinnerFrequencyModulate ? 0.5f + Progress : 0.5f;
+            if (spinningSample != null && spinnerFrequencyModulate)
+                spinningSample.Frequency.Value = 0.5f + Progress;
         }
 
         protected override void UpdateAfterChildren()
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
@@ -85,7 +85,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
         }
 
         private SkinnableSound spinningSample;
-        private const float SPINNING_SAMPLE_INITAL_FREQUENCY = 1.0f;
+        private const float SPINNING_SAMPLE_INITIAL_FREQUENCY = 1.0f;
         private const float SPINNING_SAMPLE_MODULATED_BASE_FREQUENCY = 0.5f;
 
         protected override void LoadSamples()
@@ -106,7 +106,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
                 {
                     Volume = { Value = 0 },
                     Looping = true,
-                    Frequency = { Value = SPINNING_SAMPLE_INITAL_FREQUENCY }
+                    Frequency = { Value = SPINNING_SAMPLE_INITIAL_FREQUENCY }
                 });
             }
         }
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -221,8 +225,7 @@ protected override void Update()
                 RotationTracker.Tracking = !Result.HasResult && (OsuActionInputManager?.PressedActions.Any(x => x == OsuAction.LeftButton || x == OsuAction.RightButton) ?? false);
 
             if (spinningSample != null)
-                // todo: implement SpinnerFrequencyModulate
-                spinningSample.Frequency.Value = 0.5f + Progress;
+                spinningSample.Frequency.Value = spinnerFrequencyModulate ? 0.5f + Progress : 0.5f;
```
bdach, line 228: Do you have a source for this formula? Wiki claims this frequency [can vary between 500 Hz and 80 kHz](https://osu.ppy.sh/help/wiki/Skinning/Sounds), although I'm not sure how accurate that is. This doesn't seem to have much to do with that range.

I also believe that `Progress` can be above 1, so this should be probably clamped from the top somehow as not to allow frequency to rise too much.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -165,10 +168,11 @@ protected override DrawableHitObject CreateNestedHitObject(HitObject hitObject)
         }
 
         [BackgroundDependencyLoader]
-        private void load(OsuColour colours)
+        private void load(OsuColour colours, ISkinSource skin)
         {
             positionBindable.BindValueChanged(pos => Position = pos.NewValue);
             positionBindable.BindTo(HitObject.PositionBindable);
+            spinnerFrequencyModulate = skin.GetConfig<OsuSkinConfiguration, bool>(OsuSkinConfiguration.SpinnerFrequencyModulate)?.Value ?? true;
```
bdach, line 175: This can't be done on `load()` as it'll break if you change between skins that have different values of this configuration. See `SkinReloadableDrawable.SkinChanged()` for how to do this properly.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -31,6 +32,8 @@ public class DrawableSpinner : DrawableOsuHitObject
 
         private readonly IBindable<Vector2> positionBindable = new Bindable<Vector2>();
 
+        private bool spinnerFrequencyModulate;
```
bdach, line 35: I believe this shouldn't reside in the main `DrawableSpinner`, but in the legacy spinner implementations.

bdach (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
peppy, line 175: This should be done inside `ApplySkin` instead. The way you're doing it won't work correctly when the skin is changed.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
peppy, line 228: move the initial set of frequency to 0.5f to the point the sample is constructed, then

```csharp
if (spinningSample != null && spinnerFrequencyModulate)
    spinningSample.Frequency.Value = 0.5f + Progress;
```

also move 0.5f to a constant.

peppy (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
peppy: not sure about this one. also though so but would likely involve moving the sample playback completely. maybe something which can be done at a later point.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
peppy: they haven't changed the formula here, but i'm also not sure that disabling "modulation" is supposed to leave the sound at half frequency.

bdach: I meant this comment as a "parity with stable" sort of thing. Presumably someone might complain if it's not the exact same one.

peppy: i definitely checked when implementing. stable is (roughly) `0.5 + completion`.

but as i expected, when the skins setting is off it should *always be 1*. not 0.5

ronthecookie: should I still make (now two) consts for 1 and 0.5 like your review said?

ronthecookie: Alright, made two new consts for it and it runs fine so it should be all good now.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -165,12 +171,17 @@ protected override DrawableHitObject CreateNestedHitObject(HitObject hitObject)
...
+        private void load(OsuColour colours, ISkinSource skin)
         {
             positionBindable.BindValueChanged(pos => Position = pos.NewValue);
             positionBindable.BindTo(HitObject.PositionBindable);
         }
 
+        protected override void ApplySkin(ISkinSource skin, bool allowFallback)
+        {
+            spinnerFrequencyModulate = skin.GetConfig<OsuSkinConfiguration, bool>(OsuSkinConfiguration.SpinnerFrequencyModulate)?.Value ?? true;
+        }
```
bdach, line 183: missing call to `base`

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -165,12 +171,17 @@ protected override DrawableHitObject CreateNestedHitObject(HitObject hitObject)
         }
 
         [BackgroundDependencyLoader]
-        private void load(OsuColour colours)
+        private void load(OsuColour colours, ISkinSource skin)
```
bdach, line 174: param can now be removed

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
```diff
@@ -82,6 +85,8 @@ protected override void LoadComplete()
         }
 
         private SkinnableSound spinningSample;
+        private const float SPINNING_SAMPLE_INITAL_FREQUENCY = 1.0f;
```
bdach, line 88: typo in name
```suggestion
        private const float SPINNING_SAMPLE_INITIAL_FREQUENCY = 1.0f;
```
(+ usages)

bdach (dismissed):

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSpinner.cs
bdach: also see notes from CI (both consts should be lower snake-case)

ronthecookie:
Sorry for all the hassle, haven't touched a project with such high standards in a while.

bdach:
Source looks saner now but not sure how to test. Either a link to a skin that has this off or a test scene would do I suppose.

ronthecookie:
I can do the test skin if you want but if you want me to do the test scene then please do elaborate (in-game? headless? how to automate?)

bdach:
Well, I presume you tested this change in some capacity?

As the thing being tested is auditory it shouldn't be headless. I suppose a barebones scene with a skin that has it on and another one that has it off so that there is an audible change would be enough, but again, I don't really insist on it as long as there is *some* way to check that this is working provided as part of this PR.

ronthecookie:
> P.S: So far I've just rigged up two versions of my favorite skin
(and it works)

ronthecookie:
alright, got it [uploaded](https://i.ronthecookie.me/pull9869.zip). (was too big for github)

bdach (approved):
seems to be working correctly looking at the skins provided :+1:

smoogipoo (approved):

PR #9878, opened 2020-08-16
Fix a couple of new Resharper inspections

peppy:


bdach (approved):
looks good :+1: 

Can't possibly imagine the CI failure is related.

Joehuu (approved):

PR #9883, opened 2020-08-17
Add "During Gameplay" option for mouse confining

### LINKED ISSUE #7021: Option to confine mouse cursor only while playing
It would be nice to have a option to lock the cursor on screen only when playing, and let it free to switch between monitors in the song selection.

swoolcock:
Adds the ability to select "During Gameplay" from the confine cursor options.  "Gameplay" is considered to be any time that `Player` is active and has not been paused.  Replays/autoplay are not considered gameplay, and do not confine the cursor.

The `DuringGameplay` enum case was added to a new enum in osu! rather than the framework's `ConfineMouseMode`, as other framework consumers may not have the concept of "gameplay" in their projects.

Implemented by adding a cached tracking element to transfer `OsuSetting.ConfineMouse` to `FrameworkSetting.ConfineMouse` while allowing `Player` to resolve and indicate whether gameplay is active.  `ConfineMouseTracker` is resolved allowing nulls to ensure unit tests can still run.

Resolves #7021


peppy:
This is adding a *third* method of doing the same thing.

https://github.com/ppy/osu/pull/9766/files
https://github.com/ppy/osu/blob/master/osu.Game/Screens/Play/ScreenSuspensionHandler.cs

swoolcock:
> This is adding a _third_ method of doing the same thing.
> 
> https://github.com/ppy/osu/pull/9766/files
> https://github.com/ppy/osu/blob/master/osu.Game/Screens/Play/ScreenSuspensionHandler.cs

Would it be acceptable to make a a game-wide cached tracker for handling gameplay state and update the other uses accordingly?  It's not quite the same as `ScreenSuspensionHandler` as it needs to ignore replays too.

bdach:
Not sure about a tracker, but it does seem that a central place to do this sort of stuff is needed. The rough idea of an API I would have is to do something like

```csharp
screenMode.EnableDistractionFreeMode(DistractionFreeMode.KeepScreenActive); // disables dim/overlays - for replays
screenMode.EnableDistractionFreeMode(DistractionFreeMode.KeepScreenActive | DistractionFreeMode.ConfineInput); // for actual gameplay
screenMode.DisableDistractionFreeMode(); // re-enables everything for now - can be scoped more granularly in the future if need be
```

Names are awful, but I think it could clean up the mess somewhat nicely. Would have to see in practice, however - it might turn out horrible too.

osu.Game/Screens/Play/Player.cs
```diff
@@ -225,6 +229,8 @@ private void load(AudioManager audio, OsuConfigManager config)
 
             DrawableRuleset.HasReplayLoaded.BindValueChanged(_ => updatePauseOnFocusLostState(), true);
 
+            confineMouseTracker?.OverlayActivationMode.BindTo(OverlayActivationMode);
```
peppy, line 232: First of all, this is incorrect. You're binding here but never unbinding, relying on luck as to whether the original state revert is ever being sent to the tracker instance (depends on GC luck as bindable relationships are using weak references).

Why does this need to be done inside `Player`? The target bindable (`OverlayActivationMode`) is available in `OsuGame`, the same place you're making this tracker instance. I think binding at a game-wide level seems fine for now? Any case where we are disabling all overlays from displaying sounds like a case where we would be confining mouse input (should the user want that) anyways.

I'd also add xmldoc in the `OverlayActivation` enum class on a per-entry level, including mentioning the confine mode logic.

osu.Game/Input/ConfineMouseTracker.cs
```diff
@@ -0,0 +1,65 @@
...
+using osu.Framework.Input;
+using osu.Game.Configuration;
+using osu.Game.Overlays;
+using osu.Game.Screens.Play;
+
+namespace osu.Game.Input
+{
+    /// <summary>
+    /// Connects <see cref="OsuSetting.ConfineMouseMode"/> with <see cref="FrameworkSetting.ConfineMouseMode"/>,
+    /// while optionally binding an <see cref="OverlayActivation"/> mode, usually that of the current <see cref="Player"/>.
```
peppy, line 17: remove the part about optional.

osu.Game/Input/OsuConfineMouseMode.cs
```diff
@@ -0,0 +1,37 @@
...
+        /// The mouse cursor will be locked to the window bounds while in fullscreen mode.
+        /// </summary>
+        Fullscreen,
+
+        /// <summary>
+        /// The mouse cursor will be locked to the window bounds during gameplay,
+        /// but may otherwise move freely.
+        /// </summary>
+        [Description("During Gameplay")]
+        DuringGameplay,
```
peppy, line 30: Rename to `WhenOverlaysDisabled` for clarity? I think it should work better.

peppy (changes requested):

osu.Game/Input/OsuConfineMouseMode.cs
swoolcock: Leave the `Description` as is I assume?

peppy: yep

swoolcock:
Note that this now means that the mouse is confined during intro/outro screens.

peppy:
Hmm, that sounds bad. Maybe OsuGame should forward the "isPlaying" state then?

swoolcock:
Could potentially enable/disable the `WhenOverlaysDisabled` functionality when `MainMenu` gets pushed/popped?  Or if we want to explicitly limit it to `Player` we could do it there.

peppy:
At the point `OsuGame` is maintaining a state, do it properly. Create an `IsGameplay` `BindableBool` and hook that up.

osu.Game/OsuGame.cs
```diff
@@ -90,11 +90,18 @@ public class OsuGame : OsuGameBase, IKeyBindingHandler<GlobalAction>
...
+
         /// <summary>
         /// Whether overlays should be able to be opened game-wide. Value is sourced from the current active screen.
         /// </summary>
         public readonly IBindable<OverlayActivation> OverlayActivationMode = new Bindable<OverlayActivation>();
 
+        /// <summary>
+        /// Whether gameplay is currently active.
+        /// </summary>
+        public readonly Bindable<bool> IsGameplay = new BindableBool();
```
peppy, line 103:
```suggestion
        /// <summary>
        /// Whether the local user is currently interacting with the game in a way that should not be interrupted.
        /// </summary>
        public readonly Bindable<bool> LocalUserPlaying = new BindableBool();
```

rename other usages / privates to match.

osu.Game/OsuGame.cs
```diff
@@ -542,6 +549,7 @@ protected override void LoadComplete()
             BackButton.Receptor receptor;
 
             dependencies.CacheAs(idleTracker = new GameIdleTracker(6000));
+            dependencies.Cache(confineMouseTracker = new ConfineMouseTracker());
```
peppy, line 552: this is never used

osu.Game/Screens/Play/Player.cs
```diff
@@ -68,6 +68,8 @@ public class Player : ScreenWithBeatmapBackground
 
         private readonly Bindable<bool> storyboardReplacesBackground = new Bindable<bool>();
 
+        private readonly Bindable<bool> isGameplay = new Bindable<bool>();
```
peppy, line 71: as above, rename

osu.Game/Screens/Play/Player.cs
```diff
@@ -156,7 +158,7 @@ protected virtual void PrepareReplay()
         }
 
         [BackgroundDependencyLoader]
-        private void load(AudioManager audio, OsuConfigManager config)
+        private void load(AudioManager audio, OsuConfigManager config, OsuGameBase game)
```
peppy, line 161:
```suggestion
        private void load(AudioManager audio, OsuConfigManager config, OsuGame game)
```

`[BackgroundDependencyLoader(true)]` to allow nulls

osu.Game/Screens/Play/Player.cs
```diff
@@ -172,6 +174,9 @@ private void load(AudioManager audio, OsuConfigManager config)
 
             mouseWheelDisabled = config.GetBindable<bool>(OsuSetting.MouseDisableWheel);
 
+            if (game is OsuGame osuGame)
```
peppy, line 177:
```suggestion
            if (game != null)
```

osu.Game/Screens/Play/Player.cs
```diff
@@ -693,6 +695,9 @@ public override bool OnExiting(IScreen next)
 
             musicController.ResetTrackAdjustments();
 
+            // Ensure we reset the IsGameplay state
+            isGameplay.Value = false;
```
peppy, line 699: If this is intended as a catch-all, it should probably be run in `OnSuspending` as well.

Hard to guarantee this behaviour is correct as there's no test coverage.

peppy (changes requested):
Crash on startup due to setting being unavailable on some platforms. Need to check `.Disabled` before updating the framework config:

image

Also, if we ever have more than one player running at the same time this is going to 100% provide an incorrect state (ie. if we are showing other players in a multiplayer game while the local player is also playing, they will fight over updating the state). Maybe something we shouldn't worry about for now, but worth mentioning.

osu.Game/Screens/Play/Player.cs
swoolcock: I can add a test case for this, but it will only be verifiable in the test browser, since `OsuGame.IsGameplay` is not available headless.

peppy: You can make `isGameplay` protected and test it directly.

swoolcock: As in, it's defined in `OsuGame` but we don't use that headless.

swoolcock:
As mentioned in Discord, since this no longer requires being a cached `Component`, the confine logic can just be moved to `OsuGame`.  To test the `LocalUserPlaying` state headless, the bindable will need to be moved to `OsuGameBase`.

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
```diff
@@ -23,40 +24,49 @@ public override void SetUpSteps()
...
 
         [Test]
         public void TestGameplayOverlayActivationPaused()
         {
+            AddAssert("local user playing", () => Player.LocalUserPlaying.Value);
             AddAssert("activation mode is disabled", () => Player.OverlayActivationMode == OverlayActivation.Disabled);
             AddStep("pause gameplay", () => Player.Pause());
+            AddAssert("local user not playing", () => !Player.LocalUserPlaying.Value);
             AddUntilStep("activation mode is user triggered", () => Player.OverlayActivationMode == OverlayActivation.UserTriggered);
         }
```
bdach, line 39: This test doesn't cover resume, but adding resume here is not instant because of the existence of the resume "click orange cursor" overlay. Infrastructure to test is already in other tests (`TestScenePause`, for one) - may be wiser to move this there as in its current state it feels like it doesn't really belong in this scene basing on its name anyway.

swoolcock: I placed them in this test scene given that the logic that assigns both `LocalUserPlaying` and `OverlayActivationMode` during play are in the same place in `Player`.  If one breaks, it's likely the other will also break.

If you think it would be better to put the `LocalUserPlaying` asserts somewhere that tests pause/resume functionality, the `OverlayActivationMode` asserts should probably go there too.

Alternatively I could rename `TestSceneOverlayActivation` to something like `TestScenePlayingState` that covers both, and additionally add a `LocalUserPlaying` assert to `TestScenePause`.

peppy:
I've moved the reset logic out to `OsuGame` because I think it feels saner.

I think this is okay. Adding an extra test in the screen navigation style (actually checking the game-wide) state would probably be a good stretch goal but I'm not going to dwell on it.

osu.Game.Tests/Visual/Gameplay/TestSceneOverlayActivation.cs
peppy: i think it'll do for now.

peppy (approved):

PR #9884, opened 2020-08-17
Improve visibility of toolbar tooltips against bright backgrounds

peppy:
Either this needs to happen, or the music controller needs to be in front of the toolbar. And I think we already decided it should be behind so I went with this for now.

Before:

image

image

After:

image

image


smoogipoo (approved):
The overlap on results screen still looks shocking, but this is fine as a stopgap measure - I think this has all been redesigned already.

peppy:
Which results screen overlap?

smoogipoo:
Oh I mistook your second screenshot - that's the music controller.

peppy:
Aah right. Yeah, it's not perfect but at least it's kind of legible now. Going any darker just looks bad.

PR #9891, opened 2020-08-17
Implement mania invert mod

smoogipoo:
Just fulfilling my quota to the mania community. This is one of the most requested mod ideas (https://osu.ppy.sh/community/forums/topics/148237).

Just one of about 3 mod ideas I have in mind.


bdach:
How about `FontAwesome.Solid.YinYang` for the icon?

obraz

bdach (approved):
Looks really cool with the increased density - really similar to the one inverse map I had locally downloaded because of some issue :+1:

peppy:
@smoogipoo can you confirm that this is ready for review? I think you mentioned something about changing it further.

smoogipoo:
Already done in the last commit. This is ready for review.

peppy:
As someone who has serious issues with timing and reading hold note releases and presses, this mod is my worst nightmare.

peppy (approved):

PR #9896, opened 2020-08-17
Rework ToolbarRulesetSelector to fix potential rulesets overflow

EVAST9919:
Depends on:
- [ ] https://github.com/ppy/osu-framework/pull/3840

Preview(updated): https://streamable.com/ald6q2
TODO:
- [ ] Add tests (once behavior confirmed)
- [ ] Change colours

### DIFF
--- a/osu.Game.Tests/Visual/UserInterface/TestSceneToolbarRulesetSelector.cs
+++ b/osu.Game.Tests/Visual/UserInterface/TestSceneToolbarRulesetSelector.cs
@@ -28,7 +28,7 @@ namespace osu.Game.Tests.Visual.UserInterface
                 {
                     Anchor = Anchor.Centre,
                     Origin = Anchor.Centre,
-                    AutoSizeAxes = Axes.X,
+                    Width = 400,
                     Height = Toolbar.HEIGHT,
                     Child = selector = new ToolbarRulesetSelector()
                 };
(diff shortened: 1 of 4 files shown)


bdach:
Design-wise this looks pretty horrific. White text on pink background (which also changes to black on hover), and I'm not sure the drop-down overflow should look like it's a selected item.

peppy:
Generally when a selection happens, it should be moved to the "pinned" items temporarily (just after the default rulesets), resulting in a current selection never being inside the dropdown itself. I thought this is how overflow dropdowns already worked but I guess not.

bdach (changes requested):
see above comments

peppy:
Not sure I agree with the rulesets flying to first place when the window is resized. Can you keep them as close to the dropdown without too much effort?

peppy:
Also as I mentioned, we probably want the "official" rulesets to be pinned to the left in all cases. Although I guess if it resets on restart it's not a huge deal for now.

EVAST9919:
> Not sure I agree with the rulesets flying to first place when the window is resized. Can you keep them as close to the dropdown without too much effort?

It's not that easy: we can't use `TabContainer.SetLayoutPosition` because initial layout position of each item is 0, so setting any layout position will just move them to the far right

Joppe27:
This is a good fix for this issue, but I feel like the design still needs some work.

Aantekening 2020-08-24 190800
There is too much unused space here, only 5 game modes can be displayed in this dropdown while it's taking up a significant amount of screen space. 
Listing things like this can definitely be functional, but it doesn't really follow the design of lazer in general and isn't that nice to look at.


This is just a suggestion. Using this layout solves my two major complaints about the current design. 4 times as many game modes are displayed while remaining the same size, which makes it easier and faster to look for modes without accidently scrolling past the one you're looking for. This layout also uses the same design as the current toolbar (it expands the toolbar) which looks better. I don't know if this would be hard to implement or not since this is just a quick mockup, but i hope this gives you some ideas about how to improve the design of this feature.

More pictures:

EVAST9919:
Closing since initial design has been updated

peppy:
how come?

EVAST9919:
Either way can't do much with the problem described above

peppy:
@EVAST9919 as a heads up, flyte is working on a more final design for this control.

PR #9897, opened 2020-08-18
Fix PlayerLoader test failures due to too many steps

peppy:
Fixes local test failures in `TestEarlyExitBeforePlayerConstruction` by reducing the number of test steps before performing the null check. It seems like the construction was happening sooner than expected when loading completed immediately (ie. the induced delayed was being exceeded).

Have tested this headless on loop for a good 5 minutes with zero failures. Was previously failing 100% of the time locally for me.

The removed step is still tested in the next test in the file, so nothing lost here.


bdach (approved):
Ran the test 1k times using `[Repeat]` - no failures :+1:

smoogipoo (approved):

PR #9898, opened 2020-08-18
Added statistic in mania ruleset for key counts

lanpai:
Currently there is no way to tell the key count for a mania beatmap without playing the map. This PR adds the the count as a statistic in the statistics wedge at the left.

image


bdach:
Not sure how much of an improvement this is. Key count was already visible on the left in the "Details" tab; I guess this way it's always visible but still. Visibility on carousel is to be addressed by an upcoming redesign, I think.

smoogipoo:
Yeah I'm not sure about this either. It's probably best to wait for an update from flyte in the original issue.

lanpai:
My bad here. Yeah, we should wait until it's addressed.

PR #16, opened 2016-08-31
Beatmap framework

peppy:
Opening this PR earlier rather than later to get more eyes on it.

The aim of this pull request is to lay out the basic framework for which we will build implementations around, for BeatmapSets, Beatmaps, HitObjects and everything related.

PR #9899, opened 2020-08-18
Show precise number tooltip of leaderboard score rank when 1000 or higher

Joehuu:
Seems to be a good UX improvement.

image

### DIFF
--- a/osu.Game/Online/Leaderboards/LeaderboardScore.cs
+++ b/osu.Game/Online/Leaderboards/LeaderboardScore.cs
@@ -350,7 +350,7 @@ namespace osu.Game.Online.Leaderboards
         {
             public RankLabel(int? rank)
             {
-                TooltipText = rank == null || rank < 1000 ? null : $"{rank}";
+                TooltipText = rank == null || rank < 1000 ? null : $"#{rank:N0}";
 
                 Child = new OsuSpriteText
                 {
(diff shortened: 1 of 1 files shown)


peppy:
Should probably be formatted to "#16,483" (see other examples of formatting to ensure you support local culture).

osu.Game/Online/Leaderboards/LeaderboardScore.cs
```diff
@@ -356,6 +346,24 @@ public ScoreComponentLabel(LeaderboardScoreStatistic statistic)
             }
         }
 
+        private class RankLabel : Container, IHasTooltip
+        {
+            public RankLabel(int? rank)
+            {
+                TooltipText = rank == null || rank < 1000 ? null : $"#{rank:N0}";
```
peppy, line 353:
```suggestion
                if (rank > 1000)
                    TooltipText = $"#{rank:N0}";
```

peppy (changes requested):

osu.Game/Online/Leaderboards/LeaderboardScore.cs
Joehuu: I've edited your suggestion as we also need 1000 to be included. `ToMetric` will format numbers 1000-1050 to `1k`.

peppy (approved):
seems ok

Issue #9689, opened 2020-07-27
Improve usability of mania scroll speed slider


mas-alone:
**Describe the new feature:** 
I hope Lazer’s setting about notes’ falling speed can be the same as stable version.I wish Mania's falling speed can be used with a wild speed (the fixed) instead of the current slider. It is very inaccurate and it is very troublesome.
**Proposal designs of the feature:** 
just like stable >< please
![QQ截图20200727192808](https://user-images.githubusercontent.com/52590027/88537113-84ef1700-d03f-11ea-8032-bb32e7f0647d.png)
![QQ截图20200727193030](https://user-images.githubusercontent.com/52590027/88537190-aea83e00-d03f-11ea-815a-1023d299c3a4.png)

peppy:
Are you asking for the "checked" version of the option you point at? Right now, scroll speed does *not* scale (ie. the unchecked state on stable) which is I think what you're asking for...?

Joppe27:
I think they're asking for the specific stable BPM scaling (with numbers) instead of doing it in milliseconds. They can only be adjusted with the slider in settings in steps of 25ms or so (if you move the mouse VERY slowly) which is annoying and too sensitive when it comes to faster scroll speeds. You can't solve this by using the arrow keys since those adjustements are even bigger.

I think the BPM slider setting should not be linear and instead make smaller ms adjustements in the faster scroll speed area and bigger ms adjustements in the slow scroll speed area. There's no reason to make precise adjustements to very slow scroll speeds (e.g. 1050 -> 1000ms) because the difference there is almost unnoticable, however it is very much noticable on faster scroll speeds (e.g. 450 -> 400ms). There could also be a field where the user can just type the scroll speed they want which would solve this issue more easily (?)

There's also the in-game scroll adjustment that does use numbers like on stable, but the numbers represent completely different scroll speeds in comparison to stable which is pretty irritating.

peppy:
As you've mentioned, if you use the keyboard controls, like stable, it shows numbers, like stable.

The new numbers are scaled in a similar way and should allow quit precise adjustment. @mas-alone could you please try that method first so we can ascertain what changes need to be made to make people feel at home?

mas-alone:
> As you've mentioned, if you use the keyboard controls, like stable, it shows numbers, like stable.
> The new numbers are scaled in a similar way and should allow quit precise adjustment. @mas-alone could you please try that method first so we can ascertain what changes need to be made to make people feel at home?

[...]

PR #9902, opened 2020-08-18
Update mania scroll speed range to match stable

### LINKED ISSUE #9868: The scroll speed setting for mania on osu!Lazer isn't same as on osu!stable
**Describe the bug:** 
On osu!Lazer, the 28 scroll speed is really, **REALLY** fast compared to osu!stable despite (I think?) using same numbering for the speeds. For example, I am used to 28 scroll speed on stable but I have to experiment and pick somewhere around 12-14 speed to get more or less close to the scroll speed of stable.
https://www.youtube.com/watch?v=_sbDgPIFyvo
**osu!lazer version:** 
2020.814.0
**Logs:**
 

[logs.zip](https://github.com/ppy/osu/files/5078841/logs.zip)

lanpai:
Resolves #9868.

Using `DrawableManiaRuleset.MAX_TIME_RANGE` to calculate mania's scroll speed resolves to a value unlike osu!stable's.

### DIFF
--- a/osu.Game.Rulesets.Mania/Configuration/ManiaRulesetConfigManager.cs
+++ b/osu.Game.Rulesets.Mania/Configuration/ManiaRulesetConfigManager.cs
@@ -27,7 +27,7 @@ namespace osu.Game.Rulesets.Mania.Configuration
         public override TrackedSettings CreateTrackedSettings() => new TrackedSettings
         {
             new TrackedSetting<double>(ManiaRulesetSetting.ScrollTime,
-                v => new SettingDescription(v, "Scroll Speed", $"{(int)Math.Round(13720.0 / v)} ({v}ms)"))
+                v => new SettingDescription(v, "Scroll Speed", $"{(int)Math.Round(DrawableManiaRuleset.MAX_TIME_RANGE / v)} ({v}ms)"))
         };
     }
 
--- a/osu.Game.Rulesets.Mania/Configuration/ManiaRulesetConfigManager.cs
+++ b/osu.Game.Rulesets.Mania/Configuration/ManiaRulesetConfigManager.cs
@@ -27,7 +27,7 @@ namespace osu.Game.Rulesets.Mania.Configuration
         public override TrackedSettings CreateTrackedSettings() => new TrackedSettings
         {
             new TrackedSetting<double>(ManiaRulesetSetting.ScrollTime,
-                v => new SettingDescription(v, "Scroll Speed", $"{(int)Math.Round(DrawableManiaRuleset.MAX_TIME_RANGE / v)} ({v}ms)"))
+                v => new SettingDescription(v, "Scroll Speed", $"{(int)Math.Round(13720.0 / v)} ({v}ms)"))
         };
     }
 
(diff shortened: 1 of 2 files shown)


smoogipoo:
You've only changed the display here. You need to change the actual constants (that you've removed from the calculation here).

```
DrawableManiaRuleset.MAX_TIME_RANGE = 13720
DrawableManiaRuleset.MIN_TIME_RANGE = 340
```

lanpai:
Thought that it might have unintended consequences. Changed the min and max values for the range as suggested.

smoogipoo:
The menu should probably not be using millisecond-level values in the first place. It probably should also be using 1..40 ticks.

But that's another issue: https://github.com/ppy/osu/issues/9689

smoogipoo:
@peppy Think it's fine if we break this? Or do you want a migration?

peppy:
We already have migration support in `OsuConfigManager` so let's add one. It should take a few minutes.

peppy:
@lanpai please do not use force push in the future.

smoogipoo:
This is `ManiaRulesetManager` which requires a database migration.

lanpai:
How about we use the equation `13380 * ((x-340)/13380)^2 + 340` this would create more control in the lower end of the range but less in the higher end. For example, at the half-way point where 6690 would be, it would now be mapped to 3354 while the lower bound and upper bound are kept the same.

bdach:
People will complain it's not the same as stable. Let's keep this 100% consistent IMO; finer grained control is possible via the slider in settings.

peppy:
If it's a ruleset config, you can actually migrate with a manual sql statement, I believe. Should also be pretty easy to do?

lanpai:
I'm referring to the slider in the settings, @bdach . I play at a speed around 24-26 in osu!stable and I found it really difficult to get the slider to the correct position as the slider doesn't allow for very precise movements and the equivalent to the range I play at is only around 5% of the entire slider.

bdach:
That would be a separate PR, and has already been noted by users in #9689.

smoogipoo:
Nevermind, looks like a migration isn't required since the database stores ms values.

bdach:
Not 100% sure about that - the new minimum is higher than the old one, would we not need to clip it to be safe?

smoogipoo:
It's already going to be clipped by the config manager, so it'll be fine. I've also never seen anyone play at speed 40 in stable, so I don't worry about potentially clipping people off either.

bdach:
Right, I checked and you're correct. Seems to clip automatically with no intervention needed.

bdach:
So I double-checked this in game and this is actually a little buggy at the extremes; if I go from scroll speed 1 to 40 I don't arrive at 340ms, I arrive at 343 (as 13720/40 = 343). Then I can actually increase once more to 340ms, and then if I go back to scroll speed 1, I arrive at 12169ms. Not sure whether that's something we want to address right here and now.

smoogipoo:
I think that's fine for the time being? Probably want to change the relative calculations to consider both endpoints but that can be done in a separate PR and isn't suuuuper important imo.

smoogipoo (approved):

PR #9903, opened 2020-08-18
Implement mania hold note body recycling

smoogipoo:
This is a very brute-force way of doing things, but it's already done for the same purpose in the osu! ruleset. Hold notes can take up a lot of FBOs (I was getting 2GB+ using https://github.com/ppy/osu/pull/9891).

Obviously in the future hitobject pooling is going to take care of this completely.

The changes look daunting but they're mostly just moving the parts specific to the "foreground" to a separate class that can be reconstructed (thereby disposing the previous instance).


bdach (approved):
Seems correct. I'm seeing a ~200MB GPU memory drop with this PR on one inverted map with min scroll speed, so definitely looks like a win. :+1:

peppy (approved):

PR #9905, opened 2020-08-18
Fix some judgements potentially giving non-zero score for misses

### LINKED ISSUE #9904: Hold note ticks gives score even if not pressed
**Describe the bug:** Hold note ticks gives score even if not pressed in my screenshot, i didnt press any buttons as evident with the key overlay. HP also goes down and displays miss. Accuracy sometimes goes up (depending on the current position in map or if there are notes present together with the hold)

Centipede (which has a Hold at the start) start (accuracy Low because of Head note)
osu_2020-08-18_18-11-09

Centipede (8 secs after, Accuracy goes higher)
osu_2020-08-18_18-11-15

A 7k Mania map with Holds at the middle
osu_2020-08-18_18-15-58

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5089745/runtime.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9904

Ensured by `ScoreProcessor`, but updated the members anyway for potential external use (outside `ScoreProcessor`).

### DIFF
--- a/osu.Game.Rulesets.Mania/Judgements/HoldNoteTickJudgement.cs
+++ b/osu.Game.Rulesets.Mania/Judgements/HoldNoteTickJudgement.cs
@@ -7,7 +7,7 @@ namespace osu.Game.Rulesets.Mania.Judgements
 {
     public class HoldNoteTickJudgement : ManiaJudgement
     {
-        protected override int NumericResultFor(HitResult result) => 20;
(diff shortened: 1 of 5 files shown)


bdach (approved):
footgun eliminated :+1:

peppy (approved):

PR #9909, opened 2020-08-18
Implement hold note freezing

### LINKED ISSUE #8754: Long Notes do not mask away (at the judgement line) when being hit
Feature Request/Bug (if not intentional):
Long Notes in mania do not disappear when being hit. If intentional, it should be a setting in skin.ini.
This is not noticeable on skins with solid keys (like 'osu!classic'), because they cover long notes that fall past them (so they seem to disappear regardless of whether they were hit), but on skins with transparent keys (like 'osu!lazer', or the skin I'm testing with) it is clear they don't disappear.
osu!stable:
LNs disappear past the mania-stage-hint. (played with Auto)
screenshot651
When not hit, or when the key is released prematurely (not shown here), they do not. 
screenshot652
osu!lazer:
In lazer, they do not disappear, regardless of whether they are hit or not. (1st picture utilizes Auto)
osu_2020-04-14_14-54-42
osu_2020-04-14_15-09-24

These are with the same skin on the same map.
osu!lazer version: 2020.413.0

smoogipoo:
Resolves https://github.com/ppy/osu/issues/8754

- If the head is hit and a release has not occurred, the head "freezes" onto the hit target and the body gets shortened.
- If the hold note is released, it continues scrolling past the hit target forevermore. I'm unsure whether to tie this to having _broken_ on the hold note, but I didn't since right now ticks are sparse enough to the point where a hold can jump from off-screen to on-screen.
- If the head is missed, the hold note also continues scrolling past the hit target forevermore.

There are two main components to this:
1. The body needs to resize, and the head's position needs to move along with it.
2. The tail needs to be masked along with the body. Without masking the tail you get this sort of effect:
image

It looks a little hacky but I think it makes sense? Suggestions accepted.


bdach:
So while the freezing part works pretty cool, what I'm not entirely sold on is scaling down the long note background sprite's height when the head is frozen. On skins that don't have, for a lack of a better phrasing, *vertical tonal variation*, that works well. But on some skins, [like this one](https://drive.google.com/file/d/1vi6i_KApJCzRKQxlXAXzT3uM3YYbNF2P/view), shrinking the height of the long note background looks super off and would probably look a lot better if the texture was the same height, but clipped.

smoogipoo:
Ah you're right about that, good catch! This needs to have another set of masking logic, hmmmmm...

Maybe that's a good thing though, it sounds like it may be possible to combine the tail + body masking and simplify some of this code.

smoogipoo:
Took me a little while, but I've changed it to use masking. See if that works better for your skin!

bdach (commented):
Checked a few skins, and this definitely looks a lot better across the board - nice work :+1: I think I also get how this works. Mostly a few nits/comment things left.

bdach (dismissed):

bdach (approved):
playtested, seems good - hopefully this approve will stick

PR #9911, opened 2020-08-18
Fix toolbar not respecting current overlay activation mode

Game4all:
This was observed while reviewing #9766

# Summary

This attempts to fix the toolbar not respecting the current overlay activation mode and being toggleable even though the current overlay activation mode doesn't permit it (== Disabled).

The logic previously in place for this turned out to not even be working since it was missing a schedule call. Even when it was fixed, it would allow unwanted transitions, making the toolbar potentially visible for a very few moments.

This moves the check of the current overlay activation state inside `UpdateState()` to prevent the toolbar from displaying when not allowed. A test also has been added.

### DIFF
--- a/osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
+++ b/osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
@@ -4,8 +4,10 @@
 using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Testing;
+using osu.Game.Overlays;
 using osu.Game.Overlays.Toolbar;
 using osu.Game.Rulesets;
 using osuTK.Input;
@@ -15,7 +17,7 @@ namespace osu.Game.Tests.Visual.Menus
     [TestFixture]
     public class TestSceneToolbar : OsuManualInputManagerTestScene
     {
-        private Toolbar toolbar;
+        private TestToolbar toolbar;
 
         [Resolved]
         private RulesetStore rulesets { get; set; }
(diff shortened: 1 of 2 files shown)


osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
```diff
@@ -72,5 +74,26 @@ public void TestRulesetSwitchingShortcut(bool toolbarHidden)
...
+                AddUntilStep("toolbar still hidden", () => toolbar.Visibility == Visibility.Hidden);
+            else
+                AddAssert("toolbar is visible", () => toolbar.Visibility == Visibility.Visible);
+        }
+
+        public class TestToolbar : Toolbar
+        {
+            public new Bindable<OverlayActivation> OverlayActivationMode => base.OverlayActivationMode;
+
+            public Visibility Visibility => State.Value;
```
bdach, line 96: what's the point? `State` is `public`.

osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
```diff
@@ -72,5 +74,26 @@ public void TestRulesetSwitchingShortcut(bool toolbarHidden)
...
+        public void TestRespectsOverlayActivation(OverlayActivation mode)
+        {
+            AddStep($"set activation mode to {mode}", () => toolbar.OverlayActivationMode.Value = mode);
+            AddStep("hide toolbar", () => toolbar.Hide());
+            AddStep("try to show toolbar", () => toolbar.Show());
+
+            if (mode == OverlayActivation.Disabled)
+                AddUntilStep("toolbar still hidden", () => toolbar.Visibility == Visibility.Hidden);
+            else
+                AddAssert("toolbar is visible", () => toolbar.Visibility == Visibility.Visible);
```
bdach, line 89: could also be parameterised:

```suggestion
            AddUntilStep($"toolbar is {expectedVisibility}", () => toolbar.State.Value == expectedVisibility);
```

also why is one of these an assert and the other an until step?

peppy (approved):

peppy (approved):

PR #9926, opened 2020-08-20
Fix osu!catch playfield bounds not matching expectations

### LINKED ISSUE #8221: osu!catch playfield is wider than in stable
**Describe the bug:** osu!catch playfield is widher than in stable
I don't know if it's intended.

comp

bottom is osu!lazer

**osu!lazer version:** 
2020.311.0

peppy:
- Playfield was scaled using 0.86f, not 0.8f (cross-check with `OsuPlayfieldAdjustmentContainer`). Changing this fixes horizontal sizing.
- After changing to 0.8f, the vertical alignment did not match correctly. I ported over stable's method, which is using a playfield range of -100..340 instead of the standard 0..384 (see the `FruitCircle` transform for reference, in osu-stable codebase)
- Closes #8221

Note: below visual is not what I used to make these changes, and only provided as a rough reference. I was going by actual size data so it should be 1:1 now.

A few things I'm not 100% on, but can be fixed as follow-up if/when necessary:
- The catcher looks visually smaller on lazer, but the difference is <2% and that can be fixed separately (is not affected by this PR). It may just be sprite resizing differences.
- In stable, the playfield is offset more from the top than the bottom of the screen (this goes for osu! mode too). It's doing a vertical offset from screen-top of `(parentHeight - playfieldHeight) * 4 / 3`, so 3/4 of the spacing is at that top while 1/4 is at the bottom. I don't believe we're doing that anywhere in lazer (correct me if i've missed it), so I'm not sure why vertical alignment is correct without making this change.

### DIFF
--- a/osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
+++ b/osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
@@ -10,6 +10,8 @@ namespace osu.Game.Rulesets.Catch.UI
 {
     public class CatchPlayfieldAdjustmentContainer : PlayfieldAdjustmentContainer
     {
+        private const float playfield_size_adjust = 0.8f;
+
         protected override Container<Drawable> Content => content;
         private readonly Container content;
 
@@ -18,7 +20,7 @@ namespace osu.Game.Rulesets.Catch.UI
             Anchor = Anchor.TopCentre;
             Origin = Anchor.TopCentre;
 
-            Size = new Vector2(0.86f); // matches stable's vertical offset for catcher plate
+            Size = new Vector2(playfield_size_adjust);
 
             InternalChild = new Container
             {
--- a/osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
+++ b/osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
@@ -17,8 +17,12 @@ namespace osu.Game.Rulesets.Catch.UI
 
         public CatchPlayfieldAdjustmentContainer()
         {
-            Anchor = Anchor.TopCentre;
-            Origin = Anchor.TopCentre;
+            // because we are using centre anchor/origin, we will need to limit visibility in the future
+            // to ensure tall windows do not get a readability advantage.
+            // it may be possible to bake the catch-specific offsets (-100..340 mentioned below) into new values
+            // which are compatible with TopCentre alignment.
+            Anchor = Anchor.Centre;
+            Origin = Anchor.Centre;
 
             Size = new Vector2(playfield_size_adjust);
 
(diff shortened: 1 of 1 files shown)


bdach:
As an "empirical review" I made [this video](https://drive.google.com/file/d/1mVgvSiRjave87QAky_B3g_ybrihv_mVg/view) - I overlaid lazer gameplay over stable's. It's not perfect (because of frame offset, droplet rotation, etc) but the horizontal range seems spot on. Vertical might be slightly off but that's way more difficult to accurately line up, and looks close enough to me anyway.

osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
```diff
@@ -10,15 +10,21 @@ namespace osu.Game.Rulesets.Catch.UI
 {
     public class CatchPlayfieldAdjustmentContainer : PlayfieldAdjustmentContainer
     {
+        private const float playfield_size_adjust = 0.8f;
```
bdach, line 13: If we're now sharing this constant with `OsuPlayfieldAdjustmentContainer`, maybe it should be somewhere common?

osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
```diff
@@ -10,15 +10,21 @@ namespace osu.Game.Rulesets.Catch.UI
...
+
         protected override Container<Drawable> Content => content;
         private readonly Container content;
 
         public CatchPlayfieldAdjustmentContainer()
         {
-            Anchor = Anchor.TopCentre;
-            Origin = Anchor.TopCentre;
+            // because we are using centre anchor/origin, we will need to limit visibility in the future
+            // to ensure tall windows do not get a readability advantage.
```
bdach, line 21: I'm a bit lost with this comment (probably because I've barely read catch code) - why does anchor impact readability?

osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
peppy: Yeah, not sure about this just yet. I think keeping it local is fine for now since it's only two rulesets which use this.

osu.Game.Rulesets.Catch/UI/CatchPlayfieldAdjustmentContainer.cs
peppy: The difference is if you make the window very tall. You will be able to see fruit earlier than intended. If you compare a release build or master to this PR, you should notice the difference (this PR will show more view range, compared to master).

bdach: Oh, tall windows as in literal game windows, gotcha. Not sure why I interpreted it the way I did at first as it makes no sense in retrospect...

bdach (approved):
Yeah, I think this is ok for now. The smaller kinks can be iterated upon, more important to get the baseline right first and foremost.

smoogipoo (approved):

PR #9927, opened 2020-08-20
Force tournament client to run in windowed mode

peppy:
Brought up in https://github.com/ppy/osu/issues/9200#issuecomment-677516773.

We generally haven't tested in other modes, and it doesn't really make sense as you wouldn't be able to use it in a meaningful way otherwise.

- [x] Test on windows.

### DIFF
--- a/osu.Game.Tournament/TournamentGame.cs
+++ b/osu.Game.Tournament/TournamentGame.cs
@@ -31,6 +31,7 @@ namespace osu.Game.Tournament
         public static readonly Color4 TEXT_COLOUR = Color4Extensions.FromHex("#fff");
         private Drawable heightWarning;
         private Bindable<Size> windowSize;
+        private Bindable<WindowMode> windowMode;
 
         [BackgroundDependencyLoader]
         private void load(FrameworkConfigManager frameworkConfig)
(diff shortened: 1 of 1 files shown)


bdach:
I've checked it on windows. It works but it's a little spazzy if you try to cycle modes via <kbd>Alt</kbd>+<kbd>Enter</kbd> - it goes borderless for a split-second and then back to windowed.

Doing something like the following "fixes" that:

```diff
diff --git a/osu.Game.Tournament/TournamentGame.cs b/osu.Game.Tournament/TournamentGame.cs
index 307ee1c773..5552c5f217 100644
--- a/osu.Game.Tournament/TournamentGame.cs
+++ b/osu.Game.Tournament/TournamentGame.cs
@@ -10,6 +10,8 @@
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Colour;
+using osu.Framework.Input;
+using osu.Framework.Input.Bindings;
 using osu.Game.Graphics.Cursor;
 using osu.Game.Tournament.Models;
 using osu.Game.Graphics;
@@ -32,10 +34,13 @@ public class TournamentGame : TournamentGameBase
         private Drawable heightWarning;
         private Bindable<Size> windowSize;
         private Bindable<WindowMode> windowMode;
+        private readonly BindableList<WindowMode> supportedModes = new BindableList<WindowMode> { WindowMode.Windowed };

         [BackgroundDependencyLoader]
         private void load(FrameworkConfigManager frameworkConfig)
         {
+            Window?.SupportedWindowModes.BindTo(supportedModes);
+
             windowSize = frameworkConfig.GetBindable<Size>(FrameworkSetting.WindowedSize);
             windowSize.BindValueChanged(size => ScheduleAfterChildren(() =>
             {
```

but (a) unfortunately it does not suffice on its own (it doesn't migrate if someone already has a wrong mode from a previous run) and (b) it's hacky as heck (binding to a `IBindableList`), so unsure if worth.

peppy:
Thanks for testing, you beat me. Given that it's quite an edge case I think a-bit-spazzy is fine as long as the final state is correct.

bdach (approved):

PR #9929, opened 2020-08-20
Fix catch not applying HR offset correctly

smoogipoo:
Regression from https://github.com/ppy/osu/pull/9421
Found on https://osu.ppy.sh/beatmapsets/944317#fruits/2106490

If a diffcalc deploy hasn't happened since the above PR was merged, this does not need sr/pp recalcs.


peppy:
I think we should be good in terms of deploys. Haven't updated in a while.

peppy (approved):

bdach (approved):

PR #9930, opened 2020-08-20
Remove release samples in invert mod

smoogipoo:


osu.Game.Rulesets.Mania/Mods/ManiaModInvert.cs
```diff
@@ -60,12 +60,7 @@ public void ApplyToBeatmap(IBeatmap beatmap)
                         Column = column.Key,
                         StartTime = locations[i].startTime,
                         Duration = duration,
-                        Samples = locations[i].samples,
-                        NodeSamples = new List<IList<HitSampleInfo>>
-                        {
-                            locations[i].samples,
-                            locations[i + 1].samples
-                        }
+                        NodeSamples = new List<IList<HitSampleInfo>> { locations[i].samples, new List<HitSampleInfo>() }
```
bdach, line 63: Could use `Array.Empty<HitSampleInfo>()` to avoid allocs here.

bdach (approved):

peppy (approved):

PR #9931, opened 2020-08-20
Add hyper droplets

smoogipoo:
This was discovered on https://osu.ppy.sh/beatmapsets/822552#fruits/1723939 - droplets can be hyperfruit.

Not sure about the design but it's going to change soon already soooo... Preview: https://drive.google.com/file/d/1zcfDJGATarP88NQvDDpVYwnYY064B7C5/view?usp=sharing

The legacy skin already implemented the display logic, it's only the default skin that hadn't, and obviously gameplay also assumed it didn't exist.

I'm not super happy about the conditional, suggestions appreciated.

I've also kinda clobbered it into the existing `TestSceneHyperDash`, but I think this testscene should be rewritten in the future to test smaller scenarios. The fact that it overrides `CreateBeatmap` makes it extremely unwieldy to accurately test such scenarios.


frenzibyte (changes requested):
Few nits, otherwise looks good, the design seems fitting for now.

frenzibyte (approved):
LGTM 👍

peppy (approved):

PR #9932, opened 2020-08-20
Use full catcher width for hyperdash calculation

smoogipoo:
More than anything, I'm looking for discussion here. This can be replicated on https://osu.ppy.sh/beatmapsets/511538#fruits/1087385, which is a hdash-less map on stable.

I've tried to make a failure scenario happen, but I can't seem to. Since everyone seems to want this to happen, maybe we should just do it until a later point in time (beatmap v128+) where we use the 80% area instead?

This requires a re-diffcalc.


peppy:
where is the discussion where "everyone seems to want this"? hard to discuss if there's not even a linked issue or forum thread..

smoogipoo:
Initially happened on the osu!catch BN server https://discord.gg/Fz45Cm, but spilled over into #difficulty-catch

peppy (dismissed):

peppy (approved):

smoogipoo (commented):
Seems fine

PR #9933, opened 2020-08-20
Fix juice streams sometimes playing double hitsounds

### LINKED ISSUE #8266: Juice streams in catch play end hitsound too loud
It's likely getting triggered twice.

bdach:
Resolves #8266.

# Summary

Sporadically, some catch hitsounds at the end of juice streams would sound really loud in comparison to the rest of the map. This behaviour was not consistent and would vary from stream to stream.

The reason for the doubling up was that in some cases, both the ending `DrawableFruit` and the `DrawableJuiceStream` would play their samples. This, in turn, was caused by the fact that `Catcher` could actually catch `DrawableJuiceStream`s, but not in a way you would suspect. If by the end of the stream the catcher is in a position that would have captured *the stream's first fruit*, it would register a hit on the juice stream:

https://github.com/ppy/osu/blob/02678c04d59986ae555db2a817ad7ea18dea8395/osu.Game.Rulesets.Catch/UI/Catcher.cs#L217-L231

This does nothing in terms of scoring, because `DrawableJuiceStream` has an `IgnoreJudgement` on it, but [it *does* fire `ApplyResult`](https://github.com/ppy/osu/blob/02678c04d59986ae555db2a817ad7ea18dea8395/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs#L92-L93), which in turn causes sample playback to happen in the [base `DrawableHitObject` itself](https://github.com/ppy/osu/blob/02678c04d59986ae555db2a817ad7ea18dea8395/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L267-L268).

This in fact means that not only the end sample could play twice, it could [also play when it shouldn't if the catcher was where the stream started](https://drive.google.com/file/d/1CUWMnOr7CTwxSxjNPGfA8ruOsSwkqeCL/view?usp=sharing) - and this would apply to banana showers as well, I'm pretty sure.

Now, as for the fix... There's about a dozen ways you could program this if you wanted. The way I'm proposing comes from the fact that after some thinking I decided that distinguishing whether a hit object is palpable or not on the drawable level is too high up; that same distinction should apply to the data/domain level. Therefore, I changed the hierarchy of catch hit objects as follows:

```
CatchHitObject
  JuiceStream
  BananaShower
  PalpableCatchHitObject
    Fruit
    Droplet
```

and adjusted the drawable hierarchy to match. This way, the fix at 9546fbb makes semantic sense, at least to me; if the object can't be caught, don't try to catch it, therefore don't make it a hit, therefore never play hit samples for it.

Here's a pair of videos showing the effects of this in actual gameplay:

* [before](https://drive.google.com/file/d/1WTzcdUl-92kE2Mqb5e5qtjhQK1p_8muH/view?usp=sharing),
* [after](https://drive.google.com/file/d/1hObTEymWz46_Slxk8IPcFrxXollLQgOe/view?usp=sharing).

# Remarks

There is a *semblance* of a test here, but a flimsy one (as in you have to run it yourself and know what to listen for to be able to tell if there's breakage). I can't figure out a good way to test this. I could test that all juice streams end in a miss, but that would be a pretty WTF assertion to write... Suggestions welcome.

Unsure if the fix is too enterprise™️. In my thinking the `HitObject` layer is supposed to represent gameplay as data, and it makes sense to me that you would distinguish palpable and non-palpable objects there. Maybe you prefer the simpler variants like sticking a conditional on there. I dunno, there's honestly too many possible ways out.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneAutoJuiceStream.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneAutoJuiceStream.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
 using System.Linq;
+using osu.Game.Audio;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Catch.Objects;
 using osu.Game.Rulesets.Catch.UI;
@@ -38,7 +40,11 @@ namespace osu.Game.Rulesets.Catch.Tests
                         new Vector2(width, 0)
                     }),
                     StartTime = i * 2000,
-                    NewCombo = i % 8 == 0
+                    NewCombo = i % 8 == 0,
+                    Samples = new List<HitSampleInfo>(new[]
+                    {
+                        new HitSampleInfo { Bank = "normal", Name = "hitnormal", Volume = 100 }
+                    })
                 });
             }
 
--- a/osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
+++ b/osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
@@ -28,6 +28,11 @@ namespace osu.Game.Rulesets.Catch.Objects
         }
 
         /// <summary>
+        /// Whether this object can be placed on the catcher's plate.
+        /// </summary>
+        public virtual bool CanBePlated => false;
+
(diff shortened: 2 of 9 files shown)


peppy:
I like the new structure, definitely feels more appropriate to have the property available in the `HitObject` itself. Still not sure about using a `virtual` rather than passing down in the `ctor` + `readonly`, but it should do.

peppy:
This makes catch actually playable for me. The volume discrepancies were *so* offputting!

PR #9936, opened 2020-08-20
Fix multi screen test crashing

bdach:
As seen in #9933 ([CI build result](https://ci.appveyor.com/project/peppy/osu/builds/34767670/tests)).

Crash due to missing dependency. Would actually crash without fail in test browser, presumably didn't fail on CI due to the ctor step completing too fast before the crash could take place. Should no longer be possible to have it silently fail due to the added until step.

### DIFF
--- a/osu.Game.Tests/Visual/Multiplayer/TestSceneMultiScreen.cs
+++ b/osu.Game.Tests/Visual/Multiplayer/TestSceneMultiScreen.cs
@@ -14,7 +14,8 @@ namespace osu.Game.Tests.Visual.Multiplayer
         {
             Screens.Multi.Multiplayer multi = new Screens.Multi.Multiplayer();
 
-            AddStep(@"show", () => LoadScreen(multi));
+            AddStep("show", () => LoadScreen(multi));
+            AddUntilStep("wait for loaded", () => multi.IsLoaded);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


osu.Game.Tests/Visual/Multiplayer/TestSceneMultiScreen.cs
```diff
@@ -10,11 +12,15 @@ public class TestSceneMultiScreen : ScreenTestScene
     {
         protected override bool UseOnlineAPI => true;
 
+        [Cached]
+        private MusicController musicController { get; set; } = new MusicController();
```
frenzibyte, line 16: Missing `Add(musicController)`, honestly not sure about caching a local music controller at all, you can just allow permitting null from the multiplayer screen just like `MainMenu`, but it's fine either way I suppose.

frenzibyte (changes requested):
.

osu.Game.Tests/Visual/Multiplayer/TestSceneMultiScreen.cs
peppy: The direction of https://github.com/ppy/osu/pull/9792 will make this obsolete anyway.

peppy (approved):

Issue #9500, opened 2020-07-10
Some mania skin.ini settings are not supported yet or are not the same as in stable


Joppe27:
- [x] Colour# (lane colour): when alpha is set to 0, lazer makes the colour fully opaque, while stable makes the colour completely transparent. (https://github.com/ppy/osu/pull/9945)
- [x] mania-stage-hint only applies to the individual columns, not the entire playfield like it should. (https://github.com/ppy/osu/pull/9971)
- [x] lightingN custom file paths through skin.ini are not supported yet. (https://github.com/ppy/osu/pull/9976)
- [x] KeysUnderNotes is not supported yet. (https://github.com/ppy/osu/pull/9983)
- [x] (When notes get hit, they fade out instead of immediately dissapearing. This might be intended?) (https://github.com/ppy/osu/pull/9909)

version: 2020.707.0

enoslayd:
Barlineheight and scoreposition too

enoslayd:
The last item was solved with hold note head freezing pr (also notes quickly disappears when hit in that pr)

enoslayd:
The last item was solved with hold note head freezing pr (also notes quickly disappears when hit in that pr)

bdach:
Closing as all tasks have been completed.

GrilledCheeese:
I got these errors while importing skins, are they related to this?

2020-08-29 16:12:15 [important]: Failed to process line "Colours" into "System.Collections.Generic.List`1[osu.Game.Skinning.LegacyManiaSkinConfiguration]": Color specified in incorrect format (should be R,G,B or R,G,B,A):
2020-08-29 16:12:15 [important]: Failed to process line "Colour1: 0,0,0,225" into "System.Collections.Generic.List`1[osu.Game.Skinning.LegacyManiaSkinConfiguration]": Color specified in incorrect format (should be R,G,B or R,G,B,A):
2020-08-29 16:12:15 [important]: Failed to process line "Colour2: 0,0,0,225" into "System.Collections.Generic.List`1[osu.Game.Skinning.LegacyManiaSkinConfiguration]": Color specified in incorrect format (should be R,G,B or R,G,B,A):
2020-08-29 16:12:15 [important]: Failed to process line "Colour3: 0,0,0,225" into "System.Collections.Generic.List`1[osu.Game.Skinning.LegacyManiaSkinConfiguration]": Color specified in incorrect format (should be R,G,B or R,G,B,A):
[...]

PR #9947, opened 2020-08-21
Add padding to mania column borders to match stable

### LINKED ISSUE #9547: column bars between mania columns reaches behind mania-keys
**Describe the bug:** column bars are drawn behind mania keys. The skin uses a transparent mania-keys skin (but it has a colored part at the bottom). Stable doesnt draw them

Lazer
osu_2020-07-13_20-51-30

Stable
screenshot070

**Skin:** https://github.com/ppy/osu/files/4912515/oks.zip
**osu!lazer version:** 2020.712.0

**Logs:**
_not important? but_
[runtime.log](https://github.com/ppy/osu/files/4912572/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4912573/updater.log)
[database.log](https://github.com/ppy/osu/files/4912574/database.log)
[network.log](https://github.com/ppy/osu/files/4912578/network.log)

bdach:
Resolves #9547.

# Summary

As can be seen from the screenshots attached to the aforementioned issue, on legacy skins on stable column borders aren't drawn on the entire height of the screen - they stop at the hit position.

To resolve, apply padding locally to the relevant border boxes. The effect of this change can be seen on [this video](https://drive.google.com/file/d/1178k5i3cbx8Tk1153jp2GSALjgCVdZOB/view?usp=sharing), and also in visual tests:

| `master` | this PR |
| :-: | :-: |
| before | after |

Note that the above screenshots aren't still 100% accurate, due to the fact that (among other issues) `mania-stage-hint` is still drawn for individual columns instead of for the entire stage at once, but that is out of scope of this particular change and tracked at #9500.

# Remarks

After too much time spent thinking I decided to keep this one simple (no proxying, no complications, just a singular local padding application). A possible gripe with that way of doing things is that the fallback lookup of `HitPosition` is now duplicated in here and in `HitObjectArea`:

https://github.com/ppy/osu/blob/56480d16654ec2455bc17c3fa09f2a68606e52d0/osu.Game.Rulesets.Mania/UI/Components/HitObjectArea.cs#L49-L51

but unsure how much of a deal this is. Will split into an extension method/move to `ManiaLegacySkinTransformer` on request, but honestly it felt like a bit of an over-engineer (and I recall similar objections being raised in other skinning PRs before).

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyColumnBackground.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyColumnBackground.cs
@@ -8,6 +8,7 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
 using osu.Framework.Input.Bindings;
+using osu.Game.Rulesets.Mania.UI;
 using osu.Game.Rulesets.UI.Scrolling;
 using osu.Game.Skinning;
 using osuTK;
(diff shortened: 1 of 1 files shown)


peppy (approved):
Seems like an amicable solution. Simplicity is best.

smoogipoo (approved):

PR #9948, opened 2020-08-21
Add hotkey for user profile overlay

### LINKED ISSUE #9941: Add "open profile" hotkey
**Describe the new feature:** 
Ctrl + F (maybe another). 
The first press opens the status bar, and the second press will open the profile itself. 
If you are without a login, then let it open the login window.

Airkek:
Resolves https://github.com/ppy/osu/issues/9941

### DIFF
--- a/osu.Game/Overlays/Toolbar/ToolbarUserButton.cs
+++ b/osu.Game/Overlays/Toolbar/ToolbarUserButton.cs
@@ -21,13 +21,15 @@ namespace osu.Game.Overlays.Toolbar
 
         public ToolbarUserButton()
         {
+            TooltipMain = "User";
+            TooltipSub = "Who are you?";
+            Hotkey = GlobalAction.ShowProfile;
+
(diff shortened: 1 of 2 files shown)


bdach (changes requested):

peppy:
Not sure this is useful in its current state? As mentioned above, it doesn't open the profile and really doesn't help at all since you can't continue using keyboard input to do anything at all.

osu.Game/Overlays/Toolbar/ToolbarUserButton.cs
```diff
@@ -20,6 +21,10 @@ public class ToolbarUserButton : ToolbarOverlayToggleButton, IOnlineComponent
 
         public ToolbarUserButton()
         {
+            TooltipMain = "User";
+            TooltipSub = "Who are you?";
```
peppy, line 25: These do not work in english language.

PR #9950, opened 2020-08-21
Add "hp" filter keyword in song select search bar

Airkek:
I think it would be correct to use both names of this criteria (`hp` and `dr`)


bdach (approved):
seems to work, I guess

peppy (approved):

PR #9959, opened 2020-08-23
Fix osu!catch relax mod

orozso:
The catcher no moving in Relax mod.
```cs
protected override bool OnMouseMove(MouseMoveEvent e)
{
    catcher.UpdatePosition(e.MousePosition.X / DrawSize.X * CatchPlayfield.WIDTH);
    return base.OnMouseMove(e);
}
```

The catcher without `CatchPlayfield.WIDTH` move between 0 and 1, but the playfield size is `CatchPlayfield.WIDTH` (512).

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModRelax.cs
+++ b/osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModRelax.cs
@@ -49,7 +49,7 @@ namespace osu.Game.Rulesets.Catch.Tests.Mods
                     {
                         X = CatchPlayfield.CENTER_X,
                         StartTime = 750,
-                        Path = new SliderPath(PathType.Linear, new Vector2[] { Vector2.Zero, Vector2.UnitY * 200 })
+                        Path = new SliderPath(PathType.Linear, new[] { Vector2.Zero, Vector2.UnitY * 200 })
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModRelax.cs
```diff
@@ -0,0 +1,84 @@
...
+                    },
+                    new Fruit
+                    {
+                        X = CatchPlayfield.WIDTH,
+                        StartTime = 500
+                    },
+                    new JuiceStream
+                    {
+                        X = CatchPlayfield.CENTER_X,
+                        StartTime = 750,
```
frenzibyte, line 51: Due to the closeness between each fruit's start time, running the test with 2x the rate will fail, distancing them out will fix it:
```suggestion
                        StartTime = 0
                    },
                    new Fruit
                    {
                        X = 0,
                        StartTime = 500
                    },
                    new Fruit
                    {
                        X = CatchPlayfield.WIDTH,
                        StartTime = 1000
                    },
                    new JuiceStream
                    {
                        X = CatchPlayfield.CENTER_X,
                        StartTime = 1500,
```

frenzibyte (changes requested):

osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModRelax.cs
bdach: not sure how relevant that is

frenzibyte: Better mentioning about it rather than staying silent, though forgot to mention, my suggestion will only hide it, it can be replicated with quadruple the speed, haven't gone through what exactly is going on yet.

smoogipoo (approved):

osu.Game.Rulesets.Catch.Tests/Mods/TestSceneCatchModRelax.cs
smoogipoo: I think it's fine for now.

PR #9969, opened 2020-08-24
Fix catch hyper dash colour defaults not being set correctly

### LINKED ISSUE #9918: HyperDashAfterImage custom color support doesn't work
**Describe the bug:** 
Copied from my comment on https://github.com/ppy/osu/pull/8200

Currently there doesn't seem to be support for the HyperDashAfterImage line in the skin.ini file. All hyperdash afterimages are a bright red, while this was changeable in the stable version. Not entirely sure if the cause is HyperDashAfterImage since I can't locate skin files in lazer. Regardless, would like to be able to customize the color of the glow. bdach suggested that it might be because of the beatmap skin option, I've tried playing with it both on and off and the issue persists.

**osu!lazer version:** 
20.814.0

**Logs:**

peppy:
As the defaults were not set, if a skin happened to specify 0,0,0,0 it would be ignored due to the early returns in property setters.

- Closes #9918.

### DIFF
--- a/osu.Game.Rulesets.Catch/UI/Catcher.cs
+++ b/osu.Game.Rulesets.Catch/UI/Catcher.cs
@@ -285,8 +285,6 @@ namespace osu.Game.Rulesets.Catch.UI
 
         private void runHyperDashStateTransition(bool hyperDashing)
         {
-            trails.HyperDashTrailsColour = hyperDashColour;
-            trails.EndGlowSpritesColour = hyperDashEndGlowColour;
             updateTrailVisibility();
 
             if (hyperDashing)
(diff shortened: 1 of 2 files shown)


bdach:
This is gonna be very weird but I had a quick look and I'm still seeing red on the skin mentioned in the issue, but *only* the first few times, after which it fixes itself without intervention on my part...

peppy:
May be related to the `FadeColour`... does it resolve if you remove that or set the duration to zero?

bdach:
So it seems. Should I just nuke the fade? I don't think we guarantee smooth transitions across skin changes anywhere else.

peppy:
Yep, no reason to have them.

smoogipoo (approved):

PR #9970, opened 2020-08-24
Change TeamFlag to fill a container instead of stretch

### LINKED ISSUE #9963: Don't force aspect ratio for custom flags in tournament client
**Describe the new feature:** 
Title already explains it; reason for the request being that this would be very nice to have for 1v1 tournaments, since you could use peoples profile pictures as custom "flags". As it is right now, they get squished and look strange when you try to do this.
Section doesn't really apply for this I think. Felt wrong to submit as a bug though

MiraiSubject:
Resolves #9963

Before: 

image

image

After: 

image

image

Visual tests using custom flags might be desired, but that would require custom assets (maybe in osu-resources?) to test the filling of the container. 

For now I've tested this locally by pointing `Texture` in `DrawableTeamFlag` to a flag in my `%localappdata%\osulazer\tournament\flags`


MiraiSubject:
Updated the PR because GroupTeam missed its original size and in some scenes were some leftover size properties that could be replaced with scale or weren't needed at all any more.

I double checked all the scenes that use the flags using a custom flag and the built-in ones from the tests. For demonstration purposes, I've put screenshots of the different places that use them and also to show that the sizing, at least visually, hasn't changed. 

dotnet_2020-09-13_20-01-57

I made a lot of screenshots so I made it collapsible so it doesn't flood this comment section. 

<details>
<summary>Screenshots of various screens that use flags</summary>

TeamWinScreen

image

image


TeamIntro

image

image


Gameplay

image

image

Schedule

image

image
</details>

osu.Game.Tournament/Screens/Ladder/Components/DrawableMatchTeam.cs
```diff
@@ -63,7 +63,7 @@ public DrawableMatchTeam(TournamentTeam team, TournamentMatch match, bool losers
             this.losers = losers;
             Size = new Vector2(150, 40);
 
-            Flag.Scale = new Vector2(0.9f);
```
peppy, line 66: how did you calculate this one?

osu.Game.Tournament/Screens/Gameplay/Components/TeamDisplay.cs
```diff
@@ -29,7 +29,6 @@ public TeamDisplay(TournamentTeam team, TeamColour colour, Bindable<int?> curren
             var anchor = flip ? Anchor.TopLeft : Anchor.TopRight;
 
             Flag.RelativeSizeAxes = Axes.None;
-            Flag.Size = new Vector2(60, 40);
```
peppy, line 32: won't this be bigger now?

osu.Game.Tournament/Screens/TeamWin/TeamWinScreen.cs
```diff
@@ -90,8 +90,6 @@ private void matchChanged(ValueChangedEvent<TournamentMatch> match)
             {
                 new DrawableTeamFlag(match.Winner)
                 {
-                    Size = new Vector2(300, 200),
```
peppy, line 93: this is definitely wrong. check how it looks on master/release.

peppy (changes requested):

peppy:
If you need a reference, you can check any osulive stream vod (like [this one](https://www.twitch.tv/videos/739869464))

peppy:
@MiraiSubject planning to look into fixing this?

MiraiSubject:
> @MiraiSubject planning to look into fixing this?

Yes, I have some time again this week to look into it. Haven't been able to code much due to the start of university.

MiraiSubject:
I think scales and the values used in code make much more sense now. I used the size of `300, 200` and made it `75, 50` in `DrawableTeamFlag` and then calculated respective scales for the scenes that originally had sizes for the flags. 

One component I can't quite justify the `Scale` value for is `DrawableMatchTeam`

https://github.com/ppy/osu/blob/c24a29d1acdde0228ffdd24f7726e0a0100f9e16/osu.Game.Tournament/Screens/Ladder/Components/DrawableMatchTeam.cs#L66

I tried to get this one as accurate as possible by measuring the pixels at 1080p on `2020.1020.0` and this PR.

bdach:
It was closer, but not entirely without change yet. From what I can tell, for 100% matching the values should be:

| usage | `master` | current | should be to match |
| :- | :-: | :-: | :-: |
| `DrawableTeamFlag` | 70x47 | 75x50 | *could stay as baseline* |
| `DrawableTournamentTeam` (abstract) | 70x47 (default) | 75x50 (default) | 0.94 * 75x50 = 70.5x47 |
| `GroupTeam` | 36x24 | 75x50 * 0.5 = 37.5x25 | 75x50 * 0.48 = 36x24 |
| `TeamDisplay` | 60x40 | 75x50 * 0.8 = 60x40 | *correct* |
| `DrawableMatchTeam` | 45x30 * 0.9 = 40.5x27 | 75x50 * 0.55 = 41.25x27.5 | 75x50 * 0.54 = 40.5x27 |
| `SeedingScreen` | 300x200 * 0.3 = 90x60 | 75x50 * 1.2 = 90x60 | *correct* |
| `TeamWinScreen` | 300x200 * 0.5 = 150x100 | 75x50 * 2 = 150x100 | *correct* |

Overall I'm not a 100% sure this is the best method for specifying those sizes...

peppy:
I think I'm going to go with 75x50 just because the ratio looks nice in code. Visually I've checked with OWC designs and the flags fit fine.

Will change 0.55 to 0.54 just because it's already a weird number (and 0.5 to 0.48 because the resultant sizing feels better).

peppy (approved):

PR #9971, opened 2020-08-24
Fix legacy hit target being layered incorrectly

### LINKED ISSUE #9546: mania-stage-light is displayed under mania-stage-hint
**Describe the bug:** mania-stage-light is displayed under mania-stage-hint instead of above like in stable

lazer:
osu_2020-07-13_20-41-05

stable:
screenshot068

**osu!lazer version:** 2020.712.0

**Logs:**
_not relevant imo but_
[runtime.log](https://github.com/ppy/osu/files/4912503/runtime.log)
[updater.log](https://github.com/ppy/osu/files/4912504/updater.log)
[database.log](https://github.com/ppy/osu/files/4912505/database.log)
[network.log](https://github.com/ppy/osu/files/4912507/network.log)

edit: wrong screenshot

smoogipoo:
Fixes https://github.com/ppy/osu/issues/9546

As explained in the issue thread, the osu-stable hierarchy ordered by draw order looks like:
```
- ContainerBelow:
    - stage: mania-stage-light
    - stage: mania-stage-left
    - stage: mania-stage-right
- Container
    - column: columnBackground
    - column: leftColumnLine
    - column: rightColumnLine
    - stage: mania-stage-hint
    - stage: judgementLine
    - column: mania-stage-light
    - column: mania-keyN (if KeysUnderNotes == true)
    - column: mania-keyND (if KeysUnderNotes == true)
- ContainerNotes
- ContainerAbove
    - stage: HD/FI masks
    - column: mania-keyN (if KeysUnderNotes == false)
    - column: mania-keyND (if KeysUnderNotes == false)
    - column: keyText
    - stage: mania-stage-bottom
    - column: lightingL
    - column: lightingN
    - stage: judgementText
```

osu-lazer master looks something like:
```
- Stage
    - StageBackground
        - stage: mania-stage-light
        - stage: mania-stage-left
        - stage: mania-stage-right
                                              <-----|
                                                    |<---|
    - Columns                                       |    |
        - Column                                    |    |
            - ColumnBackground                      |    |
                - column: columnBackground    >-----|    |
                - column: leftColumnLine      >-----|    |
                - column: rightColumnLine     >-----|    |
                - column: mania-stage-light              |
            - HitTargetArea                              |
                - HitTarget                              |
                    - stage: mania-stage-hint >----------|
                    - stage: judgementLine    >----------|
                - HitObjectContainer
                - Explosions
                    - column: lightingN
            - KeyArea
                - column: mania-keyN
                - column: mania-keyND
    - stage: mania-stage-bottom
    - stage: judgementText
```
With the movement of components in this PR denoted by the `>--|-->`'s.

Which results in the following structure:
```
- Stage
    - StageBackground
        - stage: mania-stage-light
        - stage: mania-stage-left
        - stage: mania-stage-right
        - column: columnBackground
        - column: leftColumnLine
        - column: rightColumnLine
        - stage: mania-stage-hint
        - stage: judgementLine
    - Columns
        - Column
            - ColumnBackground
                - column: mania-stage-light
            - HitTargetArea
                - HitTarget
                - HitObjectContainer
                - Explosions
                    - column: lightingN
            - KeyArea
                - column: mania-keyN
                - column: mania-keyND
    - stage: mania-stage-bottom
    - stage: judgementText
```
Which pretty closely resembles osu-stable except for the higher layers.

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
@@ -66,7 +66,7 @@ namespace osu.Game.Rulesets.Mania.Skinning
             };
 
             for (int i = 0; i < stageDefinition.Columns; i++)
-                columnBackgrounds.SetColumn(i, new ColumnBackground(i, i == stageDefinition.Columns - 1));
+                columnBackgrounds.SetContentForColumn(i, new ColumnBackground(i, i == stageDefinition.Columns - 1));
         }
 
         protected override void Update()
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
@@ -14,6 +14,7 @@ namespace osu.Game.Rulesets.Mania.Skinning
     {
         private Drawable leftSprite;
         private Drawable rightSprite;
+        private Container columnBackgroundContainer;
 
         public LegacyStageBackground()
         {
@@ -44,7 +45,8 @@ namespace osu.Game.Rulesets.Mania.Skinning
                     Origin = Anchor.TopLeft,
                     X = -0.05f,
                     Texture = skin.GetTexture(rightImage)
-                }
+                },
+                columnBackgroundContainer = new Container { RelativeSizeAxes = Axes.Both }
             };
         }
 
@@ -58,5 +60,7 @@ namespace osu.Game.Rulesets.Mania.Skinning
             if (rightSprite?.Height > 0)
                 rightSprite.Scale = new Vector2(1, DrawHeight / rightSprite.Height);
         }
+
+        public void AddColumnBackground(Drawable background) => columnBackgroundContainer.Add(background);
     }
 }
--- a/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
@@ -2,22 +2,31 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
 using osu.Framework.Graphics.Sprites;
+using osu.Game.Rulesets.Mania.Beatmaps;
+using osu.Game.Rulesets.Mania.UI;
+using osu.Game.Rulesets.UI.Scrolling;
 using osu.Game.Skinning;
 using osuTK;
+using osuTK.Graphics;
 
 namespace osu.Game.Rulesets.Mania.Skinning
 {
     public class LegacyStageBackground : CompositeDrawable
     {
+        private readonly StageDefinition stageDefinition;
+
         private Drawable leftSprite;
         private Drawable rightSprite;
-        private Container columnBackgroundContainer;
+        private ColumnFlow<Drawable> columnBackgrounds;
 
-        public LegacyStageBackground()
+        public LegacyStageBackground(StageDefinition stageDefinition)
         {
+            this.stageDefinition = stageDefinition;
             RelativeSizeAxes = Axes.Both;
         }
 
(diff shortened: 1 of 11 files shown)


osu.Game.Rulesets.Mania/Skinning/LegacyStageBackground.cs
```diff
@@ -58,5 +60,7 @@ protected override void Update()
             if (rightSprite?.Height > 0)
                 rightSprite.Scale = new Vector2(1, DrawHeight / rightSprite.Height);
         }
+
+        public void AddColumnBackground(Drawable background) => columnBackgroundContainer.Add(background);
```
peppy, line 64: Was thinking this should just be a `.Children =` but i guess this is local enough to not really matter.

smoogipoo: It can't be. Columns are added in sequence, not all at once.

peppy: right, i see. it's shared between all columns here. bit of a convoluted situation but visually looks correct and is legacy class.

bdach:
I also had the "extend first column's hit target across remaining ones" workaround in mind when thinking about this. Shame it's a bit convoluted but as long as it's constrained to legacy classes I'd say it's all good.

osu.Game.Rulesets.Mania/Skinning/ManiaLegacySkinTransformer.cs
```diff
@@ -88,10 +90,11 @@ public override Drawable GetDrawableComponent(ISkinComponent component)
                     switch (maniaComponent.Component)
                     {
                         case ManiaSkinComponents.ColumnBackground:
-                            return new LegacyColumnBackground(maniaComponent.TargetColumn == beatmap.TotalColumns - 1);
+                            return new LegacyColumnBackground(maniaComponent.TargetColumn == beatmap.TotalColumns - 1, stageBackground);
```
peppy, line 93: i guess this code is the dodgiest of it all, storing a reference based on access order. i can imagine this potentially falling over in the future if this was to change for whatever reason. i guess we can't use DI since the columns aren't contained within the background..

smoogipoo: Initially I had DI of `Stage` and accessing via `(Stage.Background.Drawable as LegacyStageBackground)?.AddColumnBackground()`, but I didn't feel happy exposing the background like that.

Yeah, it's a bit unfortunate that the columns aren't contained within the background. I had another idea in mind that is to extract the backgrounds from the columns, but I'm not sure how I feel about that path either... I'd like columns to be constructible by themselves for e.g. tests, so that'd have to change.

peppy: I think I'm okay with it this way if you make sure it fails at least one test on null (ie. remove the `CanBeNull` allowance and cause an exception if ever called out of order). Else we may regress this without even knowing, since it's a legacy-only pathway.

smoogipoo: I don't understand what you mean by making sure it fails at least one test - it does fail every test in which a stage isn't created (e.g. `TestSceneColumnBackground`).

smoogipoo: I'll look into rethinking how column backgrounds are added for legacy skins to instead add them directly into `LegacyStageBackground` rather than proxying them in.

smoogipoo: Done, changes kinda blew up but see if you like it more.

osu.Game.Rulesets.Mania/UI/ColumnFlow.cs
```diff
@@ -0,0 +1,105 @@
...
+                    columns[i].Width = width.Value;
+            }
+        }
+
+        /// <summary>
+        /// Sets the content of one of the columns of this <see cref="ColumnFlow{TContent}"/>.
+        /// </summary>
+        /// <param name="column">The index of the column to set the content of.</param>
+        /// <param name="content">The content.</param>
+        public void SetColumn(int column, TContent content) => columns[column].Child = content;
```
peppy, line 89: `SetContentForColumn` or `SetColumnContent` maybe?

peppy:
i... *think* this feels better..

peppy (approved):

PR #9976, opened 2020-08-25
Add support for custom LightingN paths

smoogipoo:
Can be tested in 4K with https://drive.google.com/file/d/1veLeOWwl28JMEMl1tQQSIlQxl-3osxPy/view?usp=sharing

Fixes one item in https://github.com/ppy/osu/issues/9500


bdach (approved):
seems correct :+1:

PR #9977, opened 2020-08-25
Fix snapped distances potentially exceeding the source distance

peppy:
This results in slider placement including "excess" length, where the curve is not applied to the placed path. This is generally not what we want.

I considered adding a bool parameter (or enum) to change the floor/rounding mode, but on further examination I think this is what we always expect from this function.

Before:

After:

### DIFF
--- a/osu.Game/Rulesets/Edit/HitObjectComposer.cs
+++ b/osu.Game/Rulesets/Edit/HitObjectComposer.cs
@@ -293,7 +293,13 @@ namespace osu.Game.Rulesets.Edit
 
         public override float GetSnappedDistanceFromDistance(double referenceTime, float distance)
         {
-            var snappedEndTime = BeatSnapProvider.SnapTime(referenceTime + DistanceToDuration(referenceTime, distance), referenceTime);
+            double actualDuration = referenceTime + DistanceToDuration(referenceTime, distance);
+
+            double snappedEndTime = BeatSnapProvider.SnapTime(actualDuration, referenceTime);
+
+            // we don't want to exceed the actual duration and snap to a point in the future.
+            if (snappedEndTime > actualDuration)
+                snappedEndTime -= BeatSnapProvider.GetBeatLengthAtTime(referenceTime);
 
             return DurationToDistance(referenceTime, snappedEndTime - referenceTime);
         }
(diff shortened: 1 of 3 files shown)


osu.Game/Rulesets/Edit/HitObjectComposer.cs
```diff
@@ -293,7 +293,13 @@ public override double GetSnappedDurationFromDistance(double referenceTime, floa
 
         public override float GetSnappedDistanceFromDistance(double referenceTime, float distance)
         {
-            var snappedEndTime = BeatSnapProvider.SnapTime(referenceTime + DistanceToDuration(referenceTime, distance), referenceTime);
+            double actualDuration = referenceTime + DistanceToDuration(referenceTime, distance);
+
+            double snappedEndTime = BeatSnapProvider.SnapTime(actualDuration, referenceTime);
+
+            // we don't want to exceed the actual duration and snap to a point in the future.
+            if (snappedEndTime > actualDuration)
```
smoogipoo, line 301: Is this potentially going to lead to precision issues? Something like:
```
actualDuration = 1.73
snappedEndTime = 1.7330000163
```
And then it'd go back one beatlength and under-snap? Maybe this should be written like `snappedEndTime > actualDuration + beatLength / 2`?

peppy: In practice, this is only ever invoked on user interaction, so even if it was a precision issue the cursor movement would cancel it out.

We could definitely change to your logic with no issue, but i'm worried it doesn't read as clear...

smoogipoo: At the very least I think using `Precision.DefinitelyBigger` would be in good form.

peppy: Ah, never mind, I see what you mean. It has the potential for catastrophic failure.

peppy: div2 can't work because the input distance is not necessarily snapped. how about using `Precision.AlmostGreater` with 1ms or something?

smoogipoo: 1ms is also fine in my books, if that works.

peppy: Just added 1ms in the end. Precision isn't really usable in this case without some very backwards logic.

smoogipoo (approved):
Seems to work fine :+1:

bdach (approved):

PR #9979, opened 2020-08-25
Remove padding from columns

smoogipoo:
The movement of background in https://github.com/ppy/osu/pull/9971 has made a pixel gap become more apparent with legacy skins.

Stable doesn't have this padding, and neither do flyte's new designs. So it's probably fine to remove it straight up.

Master before the above PR:
Screenshot_2020-08-25_20-19-35

Master after the above PR:
image

After this PR (+ https://github.com/ppy/osu/pull/9945 to fix column line colours):
image

### DIFF
--- a/osu.Game.Rulesets.Mania/UI/Stage.cs
+++ b/osu.Game.Rulesets.Mania/UI/Stage.cs
@@ -79,7 +79,6 @@ namespace osu.Game.Rulesets.Mania.UI
                         columnFlow = new ColumnFlow<Column>(definition)
                         {
                             RelativeSizeAxes = Axes.Y,
-                            Padding = new MarginPadding { Left = COLUMN_SPACING, Right = COLUMN_SPACING },
                         },
                         new Container
                         {


bdach:
The const has other usages left, such as

https://github.com/ppy/osu/blob/30e4be8015a8910dc0446a36741790bd59efc670/osu.Game.Rulesets.Mania/UI/ColumnFlow.cs#L65-L67

and

https://github.com/ppy/osu/blob/30e4be8015a8910dc0446a36741790bd59efc670/osu.Game.Rulesets.Mania/UI/Column.cs#L133-L135

Are those ok to remain?

smoogipoo:
Legacy skin has inter-column spacing. The fallback is only for the default skin there, which does and still has spacing in updated designs.

smoogipoo:
Technically the second one is incorrect - it should be anchored for the leftmost/rightmost columns and should use the legacy skin spacing, but given it's half a pixel I think it's fine for now.

That override only exists for the editor anyway.

bdach:
Right, those usages seem OK to remain, but I'm admittedly a little confused as to what I'm looking at with the images in the OP. Pictures no. 2 and 3 don't have column borders any more, and something funky is going on with the stage padding (goes from 1px either side to apparently nothing on the left and some on the right, and then none at all at the end).

Probably has no bearing on the actual change, though. I've looked at the legacy default skin in the test browser and it seems better. :+1:

bdach (approved):

smoogipoo:
I think I had #9945 merged into master for that second one. Updated the screenshot (SS 3 looks like what I expect/what stable shows).

Issue #8314, opened 2020-03-17
Result panel missing unstable rate stats


stanriders:
Life graph as well but it's less important

Theighlin:
Is there work going on on life graph? Trying to resurface this if not as it is a major feature I am (and probably others are) looking forward to. It could be put in the detailed statistics view now that it's there.

bdach:
Is there a design for this? Asking partially because I would want to see if I could take a shot at it, and partially because if not we might want to take this out of August milestone.

smoogipoo:
![image](https://user-images.githubusercontent.com/1329837/91057022-a86aa900-e661-11ea-8ee9-18fc2c21ef35.png)

I think it'd be one of the texts at the bottom below the timing distribution.

bdach:
So the basic display for osu!/taiko/mania is in, but spinner UR is not in yet, neither is UR in catch (which is spatial-based in being the fruit-to-centre-of-catcher distance, and not time-based). I'm not entirely sure what to do with this issue at this point - are the things I mentioned above that aren't yet implemented important enough to enough people to bring back? I'm not sure I see much point in either myself, if I'm completely honest.

smoogipoo:
I think this issue can be closed off for now. Those things are like extras, and in the case of catch, probably a different statistic altogether.

PR #9981, opened 2020-08-25
Add component for displaying simple statistics on result screen

bdach:
First step toward #8314.

# Summary

This PR adds a component that can be used on the results screen to present simple, textual statistics to the end user. The design follows the bottom-most visible component [shown here](https://user-images.githubusercontent.com/1329837/91057022-a86aa900-e661-11ea-8ee9-18fc2c21ef35.png).

The API allows users of this component to customise both the number of columns to lay out the statistics into, as well as control the formatting of the values passed in via the `DisplayValue()` virtual method.

The test scene attached demonstrates the usage of the component. [Here's a video showing what it looks like.](https://drive.google.com/file/d/1fbZ0XXgUd-5g6pZD0Yu3hCA44DMenpPQ/view?usp=sharing)

# Remarks

I'm PRing this separately for multiple reasons:

- this on its own is somewhat large and the design is mostly eyeballed, so I wanted to keep any potential issues with that separated from the rest,
- I'm unsure whether I haven't overengineered this with the customisation options (both felt very likely to be needed at some point and also relatively free),
- integration is still a bit away for two reasons:

  - The way statistic rows are done right now doesn't allow me to just drop in this component in there, as [`StatisticsPanel` is tightly coupled to `StatisticsRow`, which is tightly coupled to `StatisticsItem`](https://github.com/ppy/osu/blob/c21b1e358c55a03f74e7e02a74042ec7a25e2712/osu.Game/Screens/Ranking/Statistics/StatisticsPanel.cs#L99-L119). I *could* add the new row alongside them in `StatisticsPanel` itself, but that feels like breaking the nice declarativeness of `CreateStatisticsForScore` somewhat. Right now I'm leaning toward decoupling via an `IStatisticsRow` interface which you could ask for a drawable representation of the row.
  - Then there's the actual computation of the UR. Current best idea is to have an `UnstableRate : SimpleStatisticsItem<double>` which receives hit events and does the computation during load.

  Both reasons are why I'm PRing the "safest" part first. Let me know if you have better ideas on how to proceed.

Naming is a bit ehh. Couldn't think of a better name (`TextualStatisticItem`?), so alternate suggestions welcome. Also thinking of renaming current `StatisticItem` to `DetailedStatisticItem` for added contrast.

The column creation is a tad hacky, but I don't think `GridContainer` has something like HTML `{row,col}span` to use for the spacers.

### DIFF
--- a/osu.Game.Tests/Visual/Ranking/TestSceneSimpleStatisticRow.cs
+++ b/osu.Game.Tests/Visual/Ranking/TestSceneSimpleStatisticRow.cs
@@ -1,0 +1,68 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using Humanizer;
+using NUnit.Framework;
+using osu.Framework.Extensions.Color4Extensions;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.Shapes;
+using osu.Framework.Utils;
+using osu.Game.Screens.Ranking.Statistics;
+
+namespace osu.Game.Tests.Visual.Ranking
+{
+    public class TestSceneSimpleStatisticRow : OsuTestScene
+    {
+        private Container container;
+
+        [SetUp]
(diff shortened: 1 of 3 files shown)


smoogipoo (approved):
lgtm :+1:

PR #9983, opened 2020-08-26
Implement mania "KeysUnderNotes" skin config

smoogipoo:
Resolves one of the items in https://github.com/ppy/osu/issues/9500

Kinda hard to test this atm, because:
1. The skin instances are shared - if it's added to the special skin it'll affect all special skins, and
2. The mania skin test scenes don't usually add a column to the hierarchy, only keeping a column for reference purposes. This means that `KeysUnderNotes=1` would cause the keys to get proxied into the column and then disappear because the column itself isn't in the hierarchy.

Can be tested with arrow skins in particular: https://osu.ppy.sh/community/forums/topics/512453 (https://puu.sh/z7raA.osk)


bdach (approved):
looks good to me :+1:

PR #9984, opened 2020-08-26
Fix lingering small ring in circles! intro

Poliwrath:
Before: https://cdn.discordapp.com/attachments/188630652340404224/748063158249193503/Ed4cgBFN2q.mp4 (notice the small ring under "c")
After: https://cdn.discordapp.com/attachments/188630652340404224/748070803287048312/54Phbxded0.mp4

### DIFF
--- a/osu.Game/Screens/Menu/IntroSequence.cs
+++ b/osu.Game/Screens/Menu/IntroSequence.cs
@@ -205,7 +205,7 @@ namespace osu.Game.Screens.Menu
                 const int line_end_offset = 120;
 
                 smallRing.Foreground.ResizeTo(1, line_duration, Easing.OutQuint);
-                smallRing.Delay(400).FadeOut();
+                smallRing.Delay(400).FadeColour(Color4.Black, 300);
 
                 lineTopLeft.MoveTo(new Vector2(-line_end_offset, -line_end_offset), line_duration, Easing.OutQuint);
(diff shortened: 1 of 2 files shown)


bdach:
Not sure the `MessageFormatter` change is bundled with this...

Poliwrath:
That's intentional, I wasn't sure if I should make a whole PR for such a minor thing, so here's two minor things!

bdach:
At this point you might as well leave it, but it's really confusing to open the diff and see a change both entirely unrelated to the PR title and not mentioned at all in the description.

osu.Game/Screens/Menu/IntroSequence.cs
```diff
@@ -205,6 +205,7 @@ public void Start(double length)
                 const int line_end_offset = 120;
 
                 smallRing.Foreground.ResizeTo(1, line_duration, Easing.OutQuint);
+                smallRing.Delay(400).FadeOut();
```
bdach, line 208: On closer inspection in test browser at say, 20% speed, this looks wrong as it's an instant fade. Unsure if worth fixing it but fading to black might be a better solution.

```suggestion
                smallRing.Delay(400).FadeColour(Color4.Black, 300);
```

(Can't be a nonzero duration fade either because of how `Ring` is implemented - its individual layers are getting the alpha change together instead of just the background)

smoogipoo (approved):
Looks fine for the time being. Maybe a little slow but I can only notice it by frame peeking.

PR #9985, opened 2020-08-26
Add support for LightingL

### LINKED ISSUE #8559: Legacy hold note lighting is not implemented
Describe the bug: mania lightingn and lightingl is not applied form the skin. it uses the default lighting from lazer skin

image

osu!lazer version: 2020.402.0

**Logs:**
not needed, i think

smoogipoo:
Resolves https://github.com/ppy/osu/issues/8559

Prereqs:
- [x] https://github.com/ppy/osu/pull/9976

Not 100% sure about the hierarchy manipulation from LegacyHoldNoteBody. Another way of doing this could be to expose an event from hold notes that is bound to by the playfield in order to add/remove lights in a similar way to normal hit explosions. The new flyte skins may require it, but I'm not sure just yet.

Let me know your thoughts/suggestions, or if you'd like me to pursue the above path instead.


osu.Game.Rulesets.Mania/Skinning/LegacyBodyPiece.cs
```diff
@@ -60,28 +95,68 @@ private void load(ISkinSource skin, IScrollingInfo scrollingInfo, DrawableHitObj
...
         {
-            if (!(sprite is TextureAnimation animation))
-                return;
+            if (bodySprite is TextureAnimation bodyAnimation)
+            {
+                bodyAnimation.GotoFrame(0);
+                bodyAnimation.IsPlaying = isHitting.NewValue;
+            }
+
+            if (lightContainer != null)
```
bdach, line 104: Could flip this check to save on nesting.

osu.Game.Rulesets.Mania/Skinning/LegacyBodyPiece.cs
```diff
@@ -60,28 +95,68 @@ private void load(ISkinSource skin, IScrollingInfo scrollingInfo, DrawableHitObj
...
+                    if (light is TextureAnimation lightAnimation)
+                        lightAnimation.GotoFrame(0);
+
+                    lightContainer.FadeIn(80);
+                }
+                else
+                {
+                    lightContainer.FadeOut(120)
+                                  .OnComplete(d => Column.TopLevelContainer.Remove(d));
+                }
```
bdach, line 125: Yeah, this does feel maybe not *bad*, but a little unorthodox, especially the add/remove... Not sure if the other approach would be better to be fair; would have to compare the end results most likely. I could potentially see presses/releases before the fade is done being more of a problem with the event approach since this code handles them correctly without much hassle.

osu.Game.Rulesets.Mania/Skinning/LegacyBodyPiece.cs
```diff
@@ -19,7 +20,9 @@ public class LegacyBodyPiece : LegacyManiaColumnElement
         private readonly IBindable<ScrollingDirection> direction = new Bindable<ScrollingDirection>();
         private readonly IBindable<bool> isHitting = new Bindable<bool>();
 
-        private Drawable sprite;
+        private Drawable bodySprite;
+        private Drawable lightContainer;
+        private Drawable light;
```
bdach, line 25: Would mark these `[CanBeNull]` for safety maybe

osu.Game.Rulesets.Mania/Skinning/LegacyBodyPiece.cs
smoogipoo: I think the biggest tell is whether it's going to have to be done in the default skin too. Though the default skin may also put it inside the hold note itself rather than at the explosion layer, so we'll have to see...

peppy:
Doesn't look like this is testable with visual tests, shall we add some resouces for it?

smoogipoo:
`TestSceneHoldNote` already tests this.

peppy:
Ah, right, it's already in the default legacy. Pefect.

peppy (approved):

PR #9988, opened 2020-08-26
Add basic hit unstable rate display to results screen

bdach:
Continues (resolves?) #8314.

# Summary

Following up on #9981, integrate the new simple statistics display into the results screen and add hit unstable rate computation for osu!, taiko and mania as the first simple statistic item. Here are the caveats for now, to keep this *relatively* small (hence the question mark in the first line):

- The results achieved will *not* be 100% matching stable in all respects; they are either matching or *at least* very close on a few random scores I checked, but I haven't given much attention to what in particular stable is doing with UR in respect to which object time offsets it does and doesn't include. The actual calculation should be matching, it's just 10x standard deviation. If need be, this can be iterated on further. For now the hit events taken into consideration are consistent with the distribution graph.
- Catch does *not* receive an unstable rate stat yet as from what I've seen it uses spatial offsets from the middle of the catcher instead of time offsets, and I'm unsure of how useful that is or whether it should be named an "unstable rate" (feels like conflating terms, and from some googling it seems that other folks get confused by it too).
- This also does *not* implement the unstable rate stat for spinners. Unsure if that's something players consider useful and something that we want to bring over, and anyway I'm pretty sure we need `HitEvent` work to make it happen.
- Same as in stable, I'm pretty sure right now the UR will *not* account for mods. That is something that might need to be worked on at a `HitEvent` level (maybe have two reference offsets, one rate-adjusted and one not, but I haven't thought about this much yet).

Aside from the test case attached, here's a screenshot showing how it looks in practice:

osu_2020-08-26_22-36-26

# Remarks

The diff is scary, but menial. The worst part by far is f5e52c8 - I miscalculated with the original PR and should have planned better for how the final API should look, therefore having to resort to class renames to salvage the existing property-based declarative structure of `CreateDrawableStatisticRows`. I tried to split commits such that the flow of changes/refactorings can be easily followed, but I might have failed in that respect. Let me know if you want something split out of this.

Aside from that there's just quite a bit of restructuring.


smoogipoo:
My main concern is that this is going to be a breaking change for a few rulesets:
- https://github.com/EVAST9919/lazer-swing/blob/master/osu.Game.Rulesets.Swing/SwingRuleset.cs
- https://github.com/Altenhh/tau/blob/master/osu.Game.Rulesets.Tau/TauRuleset.cs
- https://github.com/LumpBloom7/sentakki/blob/master/osu.Game.Rulesets.Sentakki/SentakkiRuleset.cs

Is there a way that maybe this could forego the `IStatisticRow` change? Maybe something like:
```
new StatisticRow[]
{
    ...,
    new SimpleStatisticRow
    {
        Items = new SimeplStatisticItem[]
        {
            new UnstableRate(...)
        }
    }
}
```
That is to say, it's rows within rows rather than "flattening" the hierarchy somewhat, for the sole reason of avoiding the breaking change?

bdach:
That is a viable direction, however I'd have to shuffle it somewhat so that `StatisticItem` has a `CreateDrawableStatisticItem()` on it, probably. The single biggest pain point in this whole change that sparked `IStatisticRow` is that all the statistic items up to now were wrapped in a `StatisticContainer` which comes bundled with the green line thing and the statistic name, but now the table can't be inside of that, because of what the design looks like, so I need to abstract that away somehow. I'll do that if the above sounds ok.

smoogipoo:
Can the header be hidden empty if the text is empty, but otherwise keeping the wrapping as is?

bdach:
I've tried the proposed change and didn't see any reason not to do it, so it's now in here. The history of this branch is now a war crime, but the overall diff probably looks better now, actually. I renamed the drawable component *again* (to `SimpleStatisticTable` this time, to not have nested rows), but this time git picked up on it being a rename so it'll be easier to review.

I've dropped a padding container for simplicity, so it looks a smidge different now:

osu_2020-08-27_21-17-03

but I'm not sure it matters much.

smoogipoo (approved):
lgtm :+1:

PR #9998, opened 2020-08-28
Reduce redundant difficulty re-calculations

smoogipoo:
Only noticed this because I'm querying beatmap difficulty a lot in an upcoming PR.

Two changes:
1. Querying the difficulty again in the async context. My upcoming changes query the beatmap difficulty some 50 times simultaneously, which would enqueue an equal number of computations on maps like https://osu.ppy.sh/beatmapsets/29157#osu/156352 that don't complete fast enough.
2. Fixing the `Equals()` comparison. In a similar vein to the above, I'm querying the beatmap difficulty many times with different instances of the same mods.


bdach:
I did a double take on the duplicated `tryGetExisting` but it's probably ok to have it both sync and async to avoid both unnecessarily enqueueing and unnecessarily recomputing.

bdach (approved):

PR #10003, opened 2020-08-29
Add marker interface for beatmap skins

Craftplacer:
As part of #9957, it became apparent that we need to be able to make sure that no user game skins get applied to beatmaps classes themselves. This pull request seeks to resolve that.

I would like to point out two files that I feel like seem out of place, to head start review:

https://github.com/ppy/osu/pull/10003/files#diff-9edc2fea98d7c0d5897a054099d8d8d2L81-R95
https://github.com/ppy/osu/pull/10003/files#diff-5fa7caadf0a5dbf0601fa721a6339045L119-R120


osu.Game/Skinning/IBeatmapSkin.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Skinning
+{
+    public interface IBeatmapSkin : ISkin
```
bdach, line 6: missing xmldoc

osu.Game/Beatmaps/IWorkingBeatmap.cs
```diff
@@ -44,7 +44,7 @@ public interface IWorkingBeatmap
         /// <summary>
         /// Retrieves the <see cref="Skin"/> which this <see cref="WorkingBeatmap"/> provides.
```
bdach, line 45: xmldoc should be adjusted to match
```suggestion
        /// Retrieves the <see cref="IBeatmapSkin"/> which this <see cref="WorkingBeatmap"/> provides.
```

osu.Game.Rulesets.Osu.Tests/TestSceneSkinFallbacks.cs
```diff
@@ -80,15 +80,18 @@ public void TestBeatmapSkinDefault()
...
         {
-            private readonly ISkinSource skin;
+            private readonly ISkinSource skinSource;
 
             public CustomSkinWorkingBeatmap(IBeatmap beatmap, Storyboard storyboard, IFrameBasedClock frameBasedClock, AudioManager audio, ISkinSource skin)
                 : base(beatmap, storyboard, frameBasedClock, audio)
             {
-                this.skin = skin;
+                if (!(skinSource is IBeatmapSkin))
+                    throw new ArgumentException("The provided skin source must be of type IBeatmapSkin.");
```
bdach, line 89: this is completely bogus. you're checking the wrong thing (`skinSource` which isn't assigned to yet, instead of `skin`), and even after correcting this test doesn't pass. did you try to run tests after making these changes at all?

osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -116,7 +116,8 @@ public void TestSamplePlaybackWithRateMods(Type expectedMod, double expectedRate
             AddAssert("sample playback rate matches mod rates", () => sample.Channel.AggregateFrequency.Value == expectedRate);
         }
 
-        private class TestSkin : LegacySkin
+        // TODO: adding IBeatmapSkin to keep changes as minimal as possible, but this shouldn't exist or should be reworked to inherit LegacyBeatmapSkin
```
bdach, line 119: I see no reason for this todo. it's test code. it doesn't need to be 100% pretty.

osu.Game/Skinning/DefaultBeatmapSkin.cs
```diff
@@ -0,0 +1,9 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Skinning
+{
+    public class DefaultBeatmapSkin : DefaultSkin, IBeatmapSkin
```
bdach, line 6: not sure I see the point, but fine, I guess

bdach (changes requested):

osu.Game/Skinning/DefaultBeatmapSkin.cs
Craftplacer: WorkingBeatmap uses a default skin class, this is why it exists, see:
https://github.com/ppy/osu/pull/10003/files#diff-de69e6e0a80b6ab97364a0919bf10d2aR280

bdach: yes, I did read the entire diff. what I'm trying to say is that I would have probably just stuck `IBeatmapSkin` on `DefaultSkin` instead of doing this, but it's probably not worth splitting hairs over.

osu.Game.Rulesets.Osu.Tests/TestSceneSkinFallbacks.cs
Craftplacer: I'll take a second look about this.

Regarding testing, I forgot that that you can run the non-visual tests under Rider's Unit Tests tab, and I'm very sorry about that for forgetting this. I was told and used to run InspectCode.ps1 back then when running Windows.

bdach: for clarification you can run *any* test, including the visual ones (headless), through the nunit runner.

osu.Game.Rulesets.Osu.Tests/TestSceneSkinFallbacks.cs
```diff
@@ -85,7 +85,7 @@ public class CustomSkinWorkingBeatmap : ClockBackedTestWorkingBeatmap
             public CustomSkinWorkingBeatmap(IBeatmap beatmap, Storyboard storyboard, IFrameBasedClock frameBasedClock, AudioManager audio, ISkinSource skin)
                 : base(beatmap, storyboard, frameBasedClock, audio)
             {
-                if (!(skinSource is IBeatmapSkin))
+                if (!(skin is IBeatmapSkin))
                     throw new ArgumentException("The provided skin source must be of type IBeatmapSkin.");
```
frenzibyte: I would just make `skin` of type `IBeatmapSkin` and call it a day.

frenzibyte (approved):

bdach (approved):
seems fine to me, probably worth a double-check though

peppy (approved):
I'm not 100% on the attached PR but this seems logical.

smoogipoo:
I really don't like this. What's the difference between a normal skin and a beatmap skin? Why should the `LegacyBeatmapEncoder` not be able to take in an `ISkin`, or at the very least a `Skin`, and have to be scoped to an `IBeatmapSkin`? After all, combo colours are and will continue to be a skinnable element.

What can a beatmap skin provide that a normal skin absolutely must not? That sounds like a hole in skinning to me if ever allowed.

peppy:
Yeah, I'm not sure either. The conversation on that linked PR is way too deep in fucked logic for me to even follow. Only merged this because it's an easy rollback if we decide to not need it.

PR #10007, opened 2020-08-29
Fix replay download button always being disabled when initial score's replay is unavailable

Joehuu:
Before:
4Qa8nB2YY7
After:
kGAuB2ESr7

Also fixes the opposite (i.e. always being enabled) (e.g. finishing a play in timeshift and your score is locally available while other scores are unavailable/not saved online).


bdach:
I've made very minor changes (fixed a typo in test name and changed some `First()` calls to `Single()`s where applicable), but other than that seems good 👍

ghost (approved):
Aside from the irrelevant test failure on the latest commit (confirmed via [a local CI build](https://ci.appveyor.com/project/frenzibyte-alt/osu/builds/34934655/tests)), this looks good over here as well.

smoogipoo (approved):

PR #10010, opened 2020-08-29
Adjust old-style spinner alignment to be closer to stable

### LINKED ISSUE #9742: The display of spinners is not entirely correct
**Describe the bug:** 
The spinner-meter and spinner-background files do not overlap each other exactly. The spinner-meter is higher than the spinner-background. It is found on many skins, but here you can see most of all (the background is made in the form of soft lighting around the edges of the girl). The spinner itself is located just above the middle of the screen, which is why it almost goes to the upper border. Maybe lower it a little lower?
screenshot093
osu_2020-08-01_09-26-31
**osu!lazer version:** 
2020.801.0

### LINKED ISSUE #9894: Spinner filling animation is shifted down compared to stable
**Describe the bug:** The spinner filling animation isn't in the same position as stable in this skin. I waited posting this because i thought it was gonna be fixed by #9763 , but it didn't. You can see the white dots not lining up on the gray ones beyond.

Stable:
Screenshot_2
Lazer:
osu_2020-08-17_20-25-22

bdach:
- Resolves #9742.
- Resolves #9894.

PRing for feedback, as I'm not very certain of this solution.

# Summary

As it turns out, stable did... unorthodox things with the old-style spinner. While the background and circle do seem to be anchored centrally, `spinner-metre` is anchored... top-left, with an additional offset, which as the linked issues show, some skins depend on for visual effects.

Worse yet, on some skins `spinner-metre` and `spinner-background` have different dimensions entirely, sometimes higher [than prescribed e.g. here](https://osu.ppy.sh/help/wiki/Skinning/osu!), which the old anchor choice also exposed (and which is why height of `spinner-metre` cannot be used for blinking).

To resolve, wrap the whole spinner in a faux-absolute-coordinate container, keep the background and disc central, and hack the metre somewhat to make things look less bad.

Here are a few image comparisons of how affected spinners look with this change:

| skin | stable | this PR |
| :- | :-: | :-: |
| skin from #9742 | screenshot013 | osu_2020-08-29_23-39-55 |
| skin from #9894 | screenshot012 | osu_2020-08-29_23-40-16 |

This PR also includes some skin resources for `metrics-skin` based on the skin from #9894, as it seemed a pretty good ground truth as it has all three elements aligned visually.

# Remarks

There is one caveat to this; this change breaks `old-skin`:

image

Normally I wouldn't bother to PR seeing that, but I checked out of curiosity and stable actually has a similar thing happening if I drop in the `old-skin` assets into it:

image

so not sure what to do with that either...


peppy:
I was dreading having to go through and figure out the pixel alignment (as I mentioned in my original PR it was just an initial pass and guaranteed further tweaks would be required). It seems you've got it perfect here.

I've tested against my small library of (very old '07-'09) skins and alignment matches in all cases.

I think the old-skin misalignment is actually to be expected in order to match stable. While it may look incorrect for the default-old, it is potentially required for other skins from a similar era (and if not, we can adjust when someone complains about this issue specifically).

peppy (approved):

Issue #9992, opened 2020-08-27
Editor seeking jumps too far ahead


smoogipoo:
Scrollwheel tick seeking in the editor seems to jump ahead exactly `2 * beatLength/divisor`, regardless of the divisor. Can be replicated on https://osu.ppy.sh/beatmapsets/174381#mania/421064 and https://osu.ppy.sh/beatmapsets/502203#mania/1068787

peppy:
I can't reproduce this on either map, via precision or non-precision scrolling. Can you check your system settings aren't set to scroll two ticks at once? This is usually an OS level setting.

Alternatively, can you try using the arrow keys to reproduce as that is equal to one wheel "tick".

smoogipoo:
It's only propagating one tick at once.

![image](https://user-images.githubusercontent.com/1329837/91699490-579e0780-ebaf-11ea-8a23-e21ee6fe9148.png)

Should be everything you need to see what's going on - when `scrollAccumulation` is 0 (after a scroll), it goes into that `if()`, which adds 1, and then adds another 1 after the `if()`. Then it iterates through the `scrollAccumulation`, subtracting `precision=1` every time (so it iterates twice).

peppy:
Are you then able to reproduce this on any beatmap?

smoogipoo:
Looks like it, yes.

peppy:
Every mouse wheel scroll in one direction?

I don't believe the point above is what is causing what you see (although it also seems like an issue), rather than we are not correctly handling non-precise scrolling. On macOS non-precise for me is only incrementing 0.1 delta, but in the editor I think we expect it to still increment one tick.

peppy:
If it's only the first time you scroll in the editor, the the fix is as you say. The remaining issue I mention actually seems to be a non-issue (non-precise scrolling in macOS is intentionally set to 0.1f, while in windows/linux it will be 1 as expected).

If it's happening for more than one scroll tick then something else must be at play.

smoogipoo:
Resolved by #10017

PR #10017, opened 2020-08-31
Fix first scroll wheel in editor incorrectly advancing twice

peppy:
Resolves (hopefully) https://github.com/ppy/osu/issues/9992


smoogipoo (approved):
Confirmed to fix it

bdach (approved):
same here - had the issue, this fixes it, diff looks sane :+1:

PR #10019, opened 2020-08-31
Add a bindable to OsuScreen to allow dynamically changing its OverlayActivation mode

Game4all:
This was suggested in #9766 to prevent potential problems with messing directly with the game-wide bindable.

# Summary

This adds a `Bindable<OverlayActivation>` property to `OsuScreen` allowing to dynamically change the overlay activation mode of the screen during its lifetime; the set value takes effect once the screen becomes currrent. `InitialOverlayActivationMode` property has been kept around since most usages don't require using the bindable. The game bindable also has been made immutable, exclusively sourcing its current value from the current screen.

No tests since the very bit of logic to test is housed in `OsuGame`


osu.Game/Screens/OsuScreen.cs
```diff
@@ -44,9 +44,11 @@ public abstract class OsuScreen : Screen, IOsuScreen, IHasDescription
         public virtual bool HideOverlaysOnEnter => false;
 
         /// <summary>
-        /// Whether overlays should be able to be opened once this screen is entered or resumed.
+        /// The initial initial overlay activation mode to use when this screen is entered for the first time.
         /// </summary>
-        public virtual OverlayActivation InitialOverlayActivationMode => OverlayActivation.All;
+        protected virtual OverlayActivation InitialOverlayActivationMode => OverlayActivation.All;
+
+        public Bindable<OverlayActivation> OverlayActivationMode { get; }
```
smoogipoo, line 51: `readonly` instead of auto property

smoogipoo (changes requested):
Seems mostly fine, one minor nit

osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
```diff
@@ -91,7 +91,7 @@ public void TestRespectsOverlayActivation(OverlayActivation mode)
 
         public class TestToolbar : Toolbar
         {
-            public new Bindable<OverlayActivation> OverlayActivationMode => base.OverlayActivationMode;
+            public new Bindable<OverlayActivation> OverlayActivationMode => base.OverlayActivationMode as Bindable<OverlayActivation>;
```
frenzibyte, line 94: Would've replaced this with an `IOverlayActivationModeProvider` interface cached to all components tracking overlay activation mode current value, but I guess it's fine if it's only this case.

osu.Game/Screens/IOsuScreen.cs
```diff
@@ -39,9 +39,9 @@ public interface IOsuScreen : IScreen
         bool HideOverlaysOnEnter { get; }
 
         /// <summary>
-        /// Whether overlays should be able to be opened once this screen is entered or resumed.
+        /// Whether overlays should be able to be opened when this screen is current.
         /// </summary>
-        OverlayActivation InitialOverlayActivationMode { get; }
+        Bindable<OverlayActivation> OverlayActivationMode { get; }
```
frenzibyte, line 44: Should definitely be immutable from the interface, value changes should only be done in the screen class itself internally.

osu.Game/Screens/OsuScreen.cs
```diff
@@ -44,9 +44,11 @@ public abstract class OsuScreen : Screen, IOsuScreen, IHasDescription
         public virtual bool HideOverlaysOnEnter => false;
 
         /// <summary>
-        /// Whether overlays should be able to be opened once this screen is entered or resumed.
+        /// The initial initial overlay activation mode to use when this screen is entered for the first time.
```
frenzibyte, line 47: Doesn't sound clear enough (also double word typo, I just realized it as I was writing the suggestion), how about:
```suggestion
        /// The initial overlay activation mode, this would be set to <see cref="OverlayActivationMode"/> on constructor.
```

frenzibyte (changes requested):

osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
bdach: disagree strongly, overengineering

osu.Game/Screens/OsuScreen.cs
bdach: original is better if the double word is removed

frenzibyte: what if the overlay activation mode has been changed before the screen is entered, I've suggested it to redirect to the `OverlayActivationMode` bindable which is the one its value will be used when the screen is entered.

osu.Game.Tests/Visual/Menus/TestSceneToolbar.cs
frenzibyte: On second thoughts, actually yeah, the casting got my eyes on it, but I think it's fine given that this is only a test.

osu.Game/Screens/OsuScreen.cs
frenzibyte: actually might have been too picky about it, I just didn't like the fact that it explicitly says "it will be used when the screen is entered for the first time", maybe at least appending `, unless OverlayActivationMode changed prior to that.` to it.

frenzibyte (approved):

smoogipoo (approved):

PR #10020, opened 2020-08-31
Add top user score to timeshift leaderboard

### LINKED ISSUE #9527: Users can not see their own beatmap score and leaderboard position within Timeshift Lobbies
Right now, users can not see their own beatmap score while being in the TImeshift Lobby. They would have to return to the singleplayer map wheel to see their best score, but even then it would be impossible to discern which of the scores displayed in the singleplayer map wheel are currently counting in the timeshift lobby they are playing in.

The only way for users to check their own beatmap score in a Timeshift Lobby is to check the beatmap's individual leaderboard. However, once the leaderboard exceeds 50 players, they can not find their score anymore when they aren't in the top 50. 

Same goes for the overall timeshift lobby leaderboard. They can only see their current leaderboard position when they are within the Top 50 of the lobby's leaderboard. Once they are below the top 50, they can not see their current position, their total score, their amount of completed maps and their amount of play attempts.

The web display of a timeshift lobby does not offer this feature either.

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9527

Prereqs:
- [x] https://github.com/ppy/osu-web/pull/6564

Preview:
image

Position is 0 because of a bug in osu-web: https://github.com/ppy/osu-web/pull/6564#issuecomment-683709821. I've reduced the size of chat a little bit to make a little bit more room, otherwise only one other score could ever be displayed if there's a user top score present.

Diff is chonky but most of it is moving `UserTopScoreContainer` inside `Leaderboard` itself, rather than `BeatmapLeaderboard`.


osu.Game/Online/Leaderboards/Leaderboard.cs
```diff
@@ -87,6 +88,21 @@ public IEnumerable<TScoreInfo> Scores
...
+        {
+            get => topScoreContainer.Score.Value;
+            set
+            {
+                if (value == null)
+                    topScoreContainer.Hide();
+                else
+                {
+                    topScoreContainer.Show();
+                    topScoreContainer.Score.Value = value;
```
bdach, line 101: Although this code was just moved from its original place, it's a bit weird to me how a `null` top score won't persist to the bindable, contrary to everything else... `UserTopScoreContainer` looks to be able to handle nulls ok.

osu.Game/Screens/Multi/Match/Components/MatchLeaderboard.cs
```diff
@@ -50,7 +50,7 @@ protected override APIRequest FetchScores(Action<IEnumerable<APIUserScoreAggrega
 
         protected override LeaderboardScore CreateDrawableScore(APIUserScoreAggregate model, int index) => new MatchLeaderboardScore(model, index);
 
-        protected override LeaderboardScore CreateDrawableTopScore(APIUserScoreAggregate model) => new MatchLeaderboardScore(model, 0);
+        protected override LeaderboardScore CreateDrawableTopScore(APIUserScoreAggregate model) => new MatchLeaderboardScore(model, model.Position ?? 0);
```
bdach: Not sure about this coercion to 0 - it doesn't really ever make sense to have a 0 position. Would probably rather prefer just outright throwing, or being able to handle a `null` position and hiding the position text instead.

smoogipoo: Sure, the base class already handled nulls.

bdach:
Diff looks good, but I'm holding off approving until the web-side PR gets merged and deployed, since the web-side API change is breaking.

peppy (approved):

PR #10021, opened 2020-08-31
Exclude misses and empty window hits from unstable rate calculations

bdach:
Companion piece to #10015.

# Summary

For parity with the distribution graph, exclude misses and hits with empty windows from being factored into unstable rate calculation.

The calculation part was not tested before - it is now with some straightforward non-visual tests.

# Remarks

Also exposed a getter in `SimpleStatisticItem<T>` - might come in handy with future simple statistic items, I'd say.

### DIFF
--- a/osu.Game.Tests/NonVisual/Ranking/UnstableRateTest.cs
+++ b/osu.Game.Tests/NonVisual/Ranking/UnstableRateTest.cs
@@ -37,7 +37,7 @@ namespace osu.Game.Tests.NonVisual.Ranking
 
             var unstableRate = new UnstableRate(events);
 
-            Assert.IsTrue(Precision.AlmostEquals(0, unstableRate.Value));
+            Assert.AreEqual(0, unstableRate.Value);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


osu.Game.Tests/NonVisual/Ranking/UnstableRateTest.cs
```diff
@@ -0,0 +1,43 @@
...
+            var events = new[]
+            {
+                new HitEvent(-100, HitResult.Miss, new HitObject(), null, null),
+                new HitEvent(0, HitResult.Great, new HitObject(), null, null),
+                new HitEvent(200, HitResult.Meh, new HitObject { HitWindows = HitWindows.Empty }, null, null),
+            };
+
+            var unstableRate = new UnstableRate(events);
+
+            Assert.IsTrue(Precision.AlmostEquals(0, unstableRate.Value));
```
frenzibyte-alt, line 40:
```suggestion
            Assert.IsTrue(unstableRate.Value == 0);
```
nit: should definitely be zero?

bdach: acceptable difference is 1e-7, but sure, i suppose

smoogipoo (approved):

PR #10022, opened 2020-08-31
Do not attempt to complete spinners that are too short in autoplay

### LINKED ISSUE #10011: Auto attempts to spin spinners that are impossible to complete
**Describe the bug:** 
Not a big problem, close this if you want.
In rare cases, when the spinner is extremely short and is stacked on a hitcircle, it will cause acc drop.
Beatmap: https://osu.ppy.sh/beatmapsets/814850#osu/1901200
https://www.youtube.com/watch?v=Yu2Oy1KgTdM
**osu!lazer version:** 
2020.825.0
**Logs:**
Not needed

PajLe:
Resolves #10011 

Intentional low duration spinners will most likely have the `Duration` of `0`. Seems like adding a simple `Duration` check is enough to solve this problem.

### DIFF
--- a/osu.Game.Rulesets.Osu/Replays/OsuAutoGenerator.cs
+++ b/osu.Game.Rulesets.Osu/Replays/OsuAutoGenerator.cs
@@ -154,9 +154,9 @@ namespace osu.Game.Rulesets.Osu.Replays
             // The startPosition for the slider should not be its .Position, but the point on the circle whose tangent crosses the current cursor position
             // We also modify spinnerDirection so it spins in the direction it enters the spin circle, to make a smooth transition.
             // TODO: Shouldn't the spinner always spin in the same direction?
-            if (h is Spinner)
+            if (h is Spinner spinner)
             {
-                if ((h as Spinner).Duration == 0)
+                if (spinner.Duration < reactionTime)
                     return;
 
                 calcSpinnerStartPosAndDirection(((OsuReplayFrame)Frames[^1]).Position, out startPosition, out spinnerDirection);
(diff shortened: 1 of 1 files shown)


bdach:
What does "most likely" mean in this scenario? Have you tested with the beatmap linked?

PajLe:
> 
> 
> What does "most likely" mean in this scenario? Have you tested with the beatmap linked?

Yes. Works with the linked beatmap. "Most likely" means if `Duration` is not `0`, then it probably won't be less than autoplay's `reactionTime`

bdach:
Then why is the check explicitly `== 0` and not `< reactionTime`?

osu.Game.Rulesets.Osu/Replays/OsuAutoGenerator.cs
```diff
@@ -156,6 +156,9 @@ private void addHitObjectReplay(OsuHitObject h)
             // TODO: Shouldn't the spinner always spin in the same direction?
             if (h is Spinner)
             {
+                if ((h as Spinner).Duration == 0)
+                    return;
```
bdach, line 160: can be at least re-done to not typecheck twice (perhaps even inlined into one conditional, but I haven't checked)

```suggestion
            if (h is Spinner spinner)
            {
                if (spinner.Duration < reactionTime)
                    return;
```

PajLe:
I was conflicted on the same topic. In the end I choose explicit `== 0` check solely for that reason of "probability"

bdach (dismissed):

bdach:
Spinners too short to complete manually will auto-complete and award a "great" judgement. You should make sure, but I think checking against `reactionTime` should be safe.

bdach:
I've replaced the duration check with a `SpinsRequired` check, the same way #9779 did.

[Video demonstrating change.](https://drive.google.com/file/d/1aDdT8LM_VpQzW1WSPKt18XcXnfQwHtrS/view?usp=sharing)

peppy (approved):

PR #10023, opened 2020-08-31
Revert "Add marker interface for beatmap skins"

Craftplacer:
Reverts ppy/osu#10003


peppy:
Did you forget to add a description?

Craftplacer:
> Did you forget to add a description?

More or less. I saw no need. I want to revert this pull request because we all kinda realized that it has no use.

I'm in the process of cleaning up the mess.

smoogipoo (approved):

PR #10027, opened 2020-09-01
Fix potentially incorrect zoom level getting set on very short audio

### LINKED ISSUE #8742: Editor Timeline throws exceptions for extremely short beatmaps
Exception occurs in load method of Timeline.

When a Beatmap with a track length less than 10 seconds is loaded, MinZoom will throw an argument exception due to attempting to set it to a value less than 1.

peppy:
Tracks that are only seconds long could potentially crash due to exceeding supported bounds of `ZoomableScrollContainer`:

https://github.com/ppy/osu/blob/bb97e9632cd05c65bb826cabc7fff4e2b078e1d5/osu.Game/Screens/Edit/Compose/Components/Timeline/ZoomableScrollContainer.cs#L54-L55

https://github.com/ppy/osu/blob/bb97e9632cd05c65bb826cabc7fff4e2b078e1d5/osu.Game/Screens/Edit/Compose/Components/Timeline/ZoomableScrollContainer.cs#L74-L75

Closes #8742

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/Timeline/Timeline.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/Timeline/Timeline.cs
@@ -68,7 +68,7 @@ namespace osu.Game.Screens.Edit.Compose.Components.Timeline
             }, true);
         }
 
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):

bdach:
Should resolve #8742.

Issue #9859, opened 2020-08-15
default lazer judgement sprites appears alongside the taiko judgement images from the skin


enoslayd:
**Describe the bug:** the default lazer judgement sprites appears alongside the taiko judgement images from the skin. the skin doesnt have any hit-(x).png since the skin is a taiko only

**Screenshots or videos showing encountered issue:** 
Lazer:
![osu_2020-08-15_14-29-33](https://user-images.githubusercontent.com/41866328/90307284-35499000-df07-11ea-8d9f-e3881ee476f6.png)

Stable:
![screenshot095](https://user-images.githubusercontent.com/41866328/90306699-4a6ff000-df02-11ea-90b0-5efe764296ff.png)

**osu!lazer version:** 2020.814.0

**Skin:** https://drive.google.com/file/d/1is63tvBoQQnVybxPTla2SNvxqDY2rvZc/view?usp=sharing

Logs:
[runtime.log](https://github.com/ppy/osu/files/5078288/runtime.log)
[performance.log](https://github.com/ppy/osu/files/5078289/performance.log)

bdach:
@enoslayd FYI the skin is inaccessible (google drive says no permission to access the file).

enoslayd:
> @enoslayd FYI the skin is inaccessible (google drive says no permission to access the file).

Fixed

bdach:
So what I think is actually happening here is that the taiko-specific sprite in the skin is more specifically an *explosion*, not a judgement, in lazer code terms. The new lazer default skin has both an explosion and a judgement text, but stable actually doesn't seem to use `hit300` (without the prefix) and friends at all in taiko - it just displays explosions (with the prefix) on the drum. This can even be seen in the default legacy skin in lazer (see 300/100s/50s fading in and out above the drum - they're not there on stable).

As the skin linked doesn't have normal hit judgements skinned, the fallback goes straight to default lazer skins, leading to the double text.

It seems to me that the way to resolve this is to entirely disable showing non-explosion judgements on legacy skins in taiko and just keep the explosions if we want compatibility, which does not seem to be immediately achievable using the current state of the skin transformer approach (`LegacySkin` would have to know if it's in a taiko context to determine whether it should block lookup of `GameplaySkinComponent<HitResult>` or not).

peppy:
@bdach actually it is quite possible already!

PR #4471, opened 2019-03-14
Implement skinning for slider border thickness

santerinogelainen:
resolves #4422

Legacy parsing might be useless since legacy client doesn't have that skin.ini option implemented afaik, but i still added it.

Size is given as a number (1 = default, 0.2 = 20% of the default, 2 = 200% of the default, etc.).

Min is 0 (= no border). Max is 8 (the point where it's just all border)

PR #10041, opened 2020-09-03
Bring legacy slider border width closer to osu!stable

### LINKED ISSUE #8272: Sliderborder width is higher than in stable

frenzibyte:
Closes #8272 

| master | this |
|:-------:|:----:|
| sliderborder-before | sliderborder-after |

Doesn't fully match osu!stable's slider border size, but brings it up to the closest. (also not sure about the positioning mismatch, but it's probably different from the size)


peppy:
If we're going to go with `0.77f` it needs an inline comment explaining how the value was reached. Even if that is "based on visual comparison".

frenzibyte:
I did it solely on visual comparison, will attach a comment.

bdach:
Not overly sure about this being done in the legacy drawable itself. There's an existing skin lookup path for the slider border already.

frenzibyte:
Well yes but in `OsuLegacySkinTransformer`, there is no way to identify legacy skins from non-legacy, realized there's a legacy class for `SliderBody` so I just adjusted it from there and called it a day, don't see any issues with that.

bdach:
There are other issues (#9859 for one) wherein there is a need for the legacy skin to override the default skin in a similar manner. So I think a resolution that makes it possible to override this on a transformer level would be preferable.

frenzibyte:
Can do it a transformer level once legacy transformers are no longer sourced from osu!lazer skins.

smoogipoo:
I'm confused. Why couldn't this be in `LegacySkinConfiguration`?

frenzibyte:
Do you mean `OsuLegacySkinTransformer`? if so, skin lookups path goes through the legacy skin transformer regardless of whether a non-legacy (e.g. osu!lazer default) skin is selected or not (which is an issue and should probably be fixed), therefore osu!lazer slider border's width would also get shrunk, that's why I didn't override at the transformer at first.

smoogipoo:
No, I mean `LegacySkinConfiguration`. I.e. the skin configuration that `LegacySkin` sets.

bdach:
Just so we're on the same page here, the suggestion is to set the value directly for the `SliderBorderSize` lookup in the lookup dictionary in the ctor of `LegacySkin` or somewhere along those lines, right?

smoogipoo:
I was thinking more in a ctor for `LegacySkinConfiguration`. It only has to set a value in its own `ConfigDictionary`.

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -12,6 +14,13 @@ public class LegacySkinConfiguration : SkinConfiguration
         /// </summary>
         public decimal? LegacyVersion { get; internal set; }
 
+        public LegacySkinConfiguration()
+        {
+            // Roughly matches osu!stable's slider border portions.
+            // Can't use nameof(SliderBorderSize) as the lookup enum is declared in the osu! ruleset.
+            ConfigDictionary["SliderBorderSize"] = 0.77f.ToString(CultureInfo.InvariantCulture);
```
bdach, line 21: I'm not sure I like the magic string too much here... Doing it "properly" would require letting the ruleset patch this onto the skin configuration, but that feels almost overengineered, and moving the lookups to the main assembly is also icky... I guess maybe we let this slide because legacy?

frenzibyte: Yeah, or reverting back to [what I had at first](https://github.com/ppy/osu/pull/10041/commits/e7eaaf8b02efd43448b6dbbb11b4e98e7062aca7) which doesn't include magic string or overrides, just an adjustment for any legacy slider body, whichever is fine for you.

bdach: I will continue to insist that if we have a mechanism for looking up skin values, we should be using it. If it's not good enough for this use case, then that should be fixed.

peppy:
Has anyone checked how this works with a legacy skin that *actually specifies* `SliderBorderSize`? That should decide whether the current or previous code is correct. One is going to break that case.

frenzibyte:
Thing is, `SliderBorderSize` isn't actually a real skin lookup from osu!stable, just one created here for lazer (https://github.com/ppy/osu/pull/4471), so the behaviour has to be defined here.

peppy:
well in that case, 1.0 should probably reference "1.0 in stable" for that value, so your original change makes more sense.

frenzibyte:
Just to clarify, are you proposing to revert back to https://github.com/ppy/osu/pull/10041/commits/e7eaaf8b02efd43448b6dbbb11b4e98e7062aca7?

bdach:
i can't even fathom how that comment can ever be interpreted differently

frenzibyte:
> i can't even fathom how that comment can ever be interpreted differently

When I read it at first, I misinterpreted it for reverting the changes **but** with adding some type of comment explaining something (I know it's weird, but that's what I parsed at first), after reading it a couple of times I think I'm starting to get the *correct* interpretation, but I was hesitating to do it and just revert, so I asked just to be clear about it.

Seeing that you're here essentially saying "yes", reverting is what was meant here I suppose.

bdach (approved):
I guess this is fine with me for the reasons mentioned above (and because @peppy gave an actual reason which I can understand as to why should be done this way). Although as I said, we'll be having the exact same problem in this month's milestone with the issue linked above, and that one isn't fixable this easily.

smoogipoo (approved):

PR #10044, opened 2020-09-03
Update toolbar icons and source titles from overlays directly

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/113
- [x] Waiting on updated icons for `notifications` and `beatmap listing`
- Tidies up left-over `SearchableList` classes as they were in the way

Hugely improves code quality IMO. Not 100% on the generic addition to `FullscreenOverlay` but I think it's okay for now. A further refactor may see the `Title`/`Description`/`Texture` moved into the overlay itself, then passed down to the header. Or maybe it's fine as is.

This brings us about halfway to the toolbar design update. It's still missing a lot of improvements that will make these icons feel more at-home, but alas.


peppy:
Good to go now.

smoogipoo (approved):
Seems fine

PR #10045, opened 2020-09-03
Update and add missing beatmap statistic icons to info wedge

### LINKED ISSUE #2229: Add icons to beatmap info wedge
Web already has these icons. Spinner icon is not implemented yet tho.

peppy:
- [x] Depends on ppy/osu-resources#113
- [x] Depends on https://github.com/ppy/osu-resources/pull/114
- Closes https://github.com/ppy/osu/issues/2229

image


osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -418,10 +418,18 @@ public InfoLabel(BeatmapStatistic statistic)
                                     Anchor = Anchor.Centre,
                                     Origin = Anchor.Centre,
                                     RelativeSizeAxes = Axes.Both,
-                                    Scale = new Vector2(0.8f),
                                     Colour = Color4Extensions.FromHex(@"f7dd55"),
-                                    Icon = statistic.Icon,
+                                    Icon = FontAwesome.Regular.Circle,
```
smoogipoo, line 422: Isn't this only specifically for osu! rulesets? E.g. rush! now looks like:
image

osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -418,10 +418,18 @@ public InfoLabel(BeatmapStatistic statistic)
...
+                                    Scale = new Vector2(0.8f)
                                 },
+                                statistic.CreateIcon().With(i =>
+                                {
+                                    i.Anchor = Anchor.Centre;
+                                    i.Origin = Anchor.Centre;
+                                    i.RelativeSizeAxes = Axes.Both;
+                                    i.Size = new Vector2(1.2f);
```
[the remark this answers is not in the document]
smoogipoo: If we're going to go with this, at the very least this should be moved inside `BeatmapStatisticSprite` imo, and commented as to why it exists. I would expect this to be larger than the icon itself by reading the code.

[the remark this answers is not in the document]
peppy: There's no padding, no. It's to get the icons up to the correct size.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
peppy: Well that highly depends on expectation. Should the circle be always present or something optional? Probably need to figure that out and check with @arflyte.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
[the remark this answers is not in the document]
peppy: Turns out there is padding, as usual fucked assets provided to us. Will fix but requires a resources bump.

osu.Game/Screens/Select/BeatmapInfoWedge.cs
peppy: I've adjusted the legacy icons to display inside the circle but also be sized correctly. Seems okay for now?

image

peppy: Some further tweaks to match flyte's originals better (except for circle thickness):

image

smoogipoo:
lgtm :+1:

smoogipoo (dismissed):

smoogipoo (dismissed):

smoogipoo (approved):

PR #10047, opened 2020-09-03
Support creating new empty beatmaps

peppy:
The first steps in completing the beatmap creation and publishing workflow.

- On entering the editor from the main menu, a new beatmap will now be automatically created.
- Will automatically populate the local user's username
- Adds standard filename conformity when saving beatmaps

This PR is just a first step, making the required changes to beatmap handling to allow this to happen. Next steps is to add the ability to change the audio file and update metadata.


osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -212,10 +241,20 @@ public void Save(BeatmapInfo info, IBeatmap beatmapContent)
                 using (ContextFactory.GetForWrite())
                 {
                     var beatmapInfo = setInfo.Beatmaps.Single(b => b.ID == info.ID);
+                    var metadata = beatmapInfo.Metadata ?? setInfo.Metadata;
+
+                    // grab the original file (or create a new one if not found).
+                    var fileInfo = setInfo.Files.SingleOrDefault(f => string.Equals(f.Filename, beatmapInfo.Path, StringComparison.OrdinalIgnoreCase)) ?? new BeatmapSetFileInfo();
+
+                    // metadata may have changed; update the path with the standard format.
+                    beatmapInfo.Path = $"{metadata.Artist} - {metadata.Title} ({metadata.Author}) [{beatmapInfo.Version}]";
```
smoogipoo, line 250:
```suggestion
                    beatmapInfo.Path = $"{metadata.Artist} - {metadata.Title} ({metadata.Author}) [{beatmapInfo.Version}].osu";
```

Since beatmap exports are already supported.

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
```diff
@@ -33,6 +33,9 @@ public BeatmapManagerWorkingBeatmap(IResourceStore<byte[]> store, TextureStore t
 
             protected override IBeatmap GetBeatmap()
             {
+                if (BeatmapInfo.Path == null)
+                    return BeatmapInfo?.Ruleset.CreateInstance().CreateBeatmapConverter(new Beatmap { BeatmapInfo = BeatmapInfo }).Beatmap;
```
smoogipoo, line 37: Why not `null`? This looks to me like some code is expecting this func to never return `null`, but that isn't always the case.

smoogipoo (changes requested):

osu.Game/Beatmaps/BeatmapManager.cs
peppy: Aha, that was an oversight.

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
peppy: This is what makes a new beatmap for consumption. Should it be in a different place?

smoogipoo: Where does that happen? Doesn't BeatmapManager create a file anyway?

peppy: I don't follow. Where are you seeing it "create a file"?

smoogipoo: Ah, I was wrong, it only does that on `Save()`.

In that case, return the `new Beatmap() {...}` part?

peppy: without the conversion? for some reason i thought that was required..

smoogipoo: Yeah. Conversion is only done by `WorkingBeatmap.GetPlayableBeatmap()`

peppy: have you tested this works? because as i say there was 100% a reason i did the conversion.

bdach: I've tried removing the conversion here and nothing seemed to catch fire - unsure whether I've missed some scenario, though. Mostly just entered and exited the editor a couple of times, saved, exported a few times. The two added tests also pass with the suggested change.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -94,6 +95,34 @@ public partial class BeatmapManager : DownloadableArchiveModelManager<BeatmapSet
...
+                    {
+                        BaseDifficulty = new BeatmapDifficulty(),
+                        Ruleset = ruleset,
+                        Metadata = metadata,
+                        Version = "difficulty"
+                    }
+                }
+            };
+
+            var working = Import(set).Result;
```
bdach, line 122: Because of this a new beatmap is imported and stored every time the editor is entered and exited, even if nothing of value was actually done. Worse yet, because the author is set to the current user in `Editor`, the beatmap looks fine in the run after it was created as it stays in memory, but after a restart the change is lost and the author rolls back to `system`.

bdach (dismissed):

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
peppy: It does look to work, weird. I must have fixed whatever was requiring this in the first place 👍 .

osu.Game/Beatmaps/BeatmapManager.cs
peppy: I think moving the user providing to the ctor is a good solution for now. This will likely change going forward (to the point of maybe not even importing to the database until first editor save? not sure, we'll see the direction things take).

smoogipoo (dismissed):

bdach:
I haven't ran this with the latest changes yet, but won't this still create new beatmaps if you enter editor and then immediately exit it? I think it still will, as the import call has been left alone.

Not sure whether that's temporary for now, but I'd reckon it's something that should be resolved before the next release at the latest.

peppy:
Yep, that is intended (see https://github.com/ppy/osu/pull/10047#discussion_r483375759).

I don't see it being a huge issue even with a release. They can be easily cleaned up if/when required. But I have some upcoming changes which will resolve this in the near future.

smoogipoo (dismissed):

bdach:
I haven't approved this yet yesterday because this branch hasn't passed tests once yet, which is pretty worrying.

peppy:
Yeah, I'm planning to figure that out today, hold off for now.

bdach:
Seems that it was the test names all along. Should have checked that more carefully...

As an aside, this is an option to potentially avoid this happening in the future:

```diff
-        public CleanRunHeadlessGameHost(string gameName = @"", bool bindIPC = false, bool realtime = true)
+        public CleanRunHeadlessGameHost([CallerMemberName] string gameName = @"", bool bindIPC = false, bool realtime = true)
             : base(gameName, bindIPC, realtime)
         {
         }
````

Either that, or just sticking a guid on the end or something. I guess that can happen in a separate PR, though.

bdach (approved):

PR #10061, opened 2020-09-04
Fix potential hard crash if ruleset settings fail to construct

### LINKED ISSUE #10053: Settings will not open if there is an incorrect ruleset.
**Describe the bug:** 
If the folder with the rulesets contains a mode file that cannot be opened in this version, then for some reason the settings menu stops working (It just doesn't open in any way). If you remove the ruleset, then everything becomes normal. The only problem is that this is a completely non-obvious problem.
image
Example .dll - https://github.com/karaoke-dev/karaoke/releases/tag/2020.0802.0
https://youtu.be/rGmc74K0hrY (When you click on the settings icon, the screen shifts to the right, but nothing else happens.)
**osu!lazer version:** 
2020.093.0

peppy:
Closes https://github.com/ppy/osu/issues/10053.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/GameplaySection.cs
+++ b/osu.Game/Overlays/Settings/Sections/GameplaySection.cs
@@ -45,7 +45,7 @@ namespace osu.Game.Overlays.Settings.Sections
                 }
                 catch (Exception e)
                 {
-                    Logger.Error(e, $"Failed to load ruleset settings");
+                    Logger.Error(e, "Failed to load ruleset settings");
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/Settings/Sections/GameplaySection.cs
```diff
@@ -35,8 +37,18 @@ private void load(RulesetStore rulesets)
             foreach (Ruleset ruleset in rulesets.AvailableRulesets.Select(info => info.CreateInstance()))
             {
                 SettingsSubsection section = ruleset.CreateSettings();
```
smoogipoo, line 39: Any reason for not covering this also? The `CreateInstance()` call above this is checked by `RulesetStore`, so this looks to be the only one that's not covered here.

smoogipoo (changes requested):

osu.Game/Overlays/Settings/Sections/GameplaySection.cs
peppy: fair

smoogipoo (approved):

PR #10062, opened 2020-09-04
Fix MusicController raising track changed event twice

### LINKED ISSUE #10055: Now playing overlay doesn't animate correctly any more

frenzibyte:
Closes https://github.com/ppy/osu/issues/10055

`changeBeatmap()` could be called a second time on the same beatmap on performing `prev()` or `next()` due to the below lines:
https://github.com/ppy/osu/blob/ebed7d09e3412206660666a3231b406e4550206d/osu.Game/Overlays/MusicController.cs#L325-L328

This is guarding against doing that to avoid double `TrackChanged` event raises, and since the `queuedDirection` is flushed on the first time, the second raise would take `TrackChangeDirection.None` with it, therefore causing the background transitions to not happen as in the linked issue which this PR fixes.

This is tested at once in #10063 (`TestSceneMusicActionHandling`) as that test scene actually depends on this issue being fixed therefore covering it.

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -279,6 +279,10 @@ namespace osu.Game.Overlays
 
         private void changeBeatmap(WorkingBeatmap newWorking)
         {
+            // The provided beatmap is same as current, no need to do any changes.
+            if (newWorking == current)
+                return;
+
             var lastWorking = current;
 
             TrackChangeDirection direction = TrackChangeDirection.None;
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -279,7 +279,8 @@ namespace osu.Game.Overlays
 
         private void changeBeatmap(WorkingBeatmap newWorking)
         {
-            // If the provided beatmap is same as current, then there is no need to do any changes.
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/MusicController.cs
```diff
@@ -279,6 +279,10 @@ private void restartTrack()
 
         private void changeBeatmap(WorkingBeatmap newWorking)
         {
+            // If the provided beatmap is same as current, then there is no need to do any changes.
```
peppy, line 282: Rather than a comment that matches the code on the line below (100% useless), this should explain *how* this can happen, ie. under what scenarios the event is fired twice (and why).

This is important because it's done this way for a reason and may be accidentally changed in the future if someone isn't aware of this reasoning.

It probably should have already been commented in the calls to `changeBeatmap`, but you can find the reasoning here as it may not be immediately obvious (bindable execution order): https://ptb.discordapp.com/channels/188630481301012481/188630652340404224/746293044302577684

peppy: Something like:

```
This method can potentially be triggered multiple times as it is eagerly fired in next() / prev() to ensure correct execution order (changeBeatmap must be called before consumers receive the bindable changed event, which is not the case when called on the local bindable itself).
```

peppy (approved):

PR #10063, opened 2020-09-04
Move OsuGame-dependent music functionalities outside of MusicController

### LINKED ISSUE #10048: Media Keys no longer work
**Describe the bug:** As title. Media keys (<kbd>next track</kbd>/<kbd>previous track</kbd> and <kbd>play/pause track</kbd>) don't seem to work anymore. Trying to change their keybinds doesn't make them work anyway.

**osu!lazer version:** 2020.903.0

frenzibyte:
Closes #10048 

- [x] Depends on #10062

Since `MusicController` has been added very internally (by `AddInternal`) above the `GlobalActionContainer`, any music global bindings presses will not propagate into the music controller's `OnPressed` handler, therefore no longer working anymore.

This could be fixed by adding `MusicController` inside the `GlobalActionContainer`, but it'll only bring functionality that doesn't entirely work, since the controller used to be inside `OsuGame`, it was resolving `OnScreenDisplay` to display toasts of the music action, now that the controller has been moved up to `OsuGameBase`, resolving `OnScreenDisplay` will never work and is incorrect to have, as depending on an `OsuGame` component from `OsuGameBase` is probably wrong and very prone to error.

I've fixed this issue by moving these `OsuGame`-dependent functionalities out of `OsuGameBase` and into `OsuGame` as a separate component, named `MusicActionHandler`, not sure about the name though, suggestions open for a better name.

Wrote an `OsuGameTestScene` that ensures "pressing" the music bindings (by triggering them from an `OsuGameBase` level) performs the action works as expected now.


bdach (commented):
I guess I'd be fine with this resolution personally but I'd like to clarify something first.

osu.Game/Overlays/Music/MusicActionHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+    public class MusicActionHandler : Component, IKeyBindingHandler<GlobalAction>
+    {
+        [Resolved]
+        private IBindable<WorkingBeatmap> beatmap { get; set; }
+
+        [Resolved]
+        private MusicController musicController { get; set; }
+
+        [Resolved]
+        private OnScreenDisplay onScreenDisplay { get; set; }
```
bdach, line 26: any reason for dropping the `[CanBeNull]` constraint here after moving from `MusicController`?

frenzibyte: Hmm, was originally planning that `MusicActionHandler` shouldn't really be used if `OnScreenDisplay` isn't cached, but seems like the class name is a bit misleading to show that, fair, will return back.

bdach: i don't understand what you're trying to say. i was asking because removing the null is more restrictive what was there before, and i just wanted to ask if you had any idea why null was even allowed. if the `CanBeNull` is not necessary, it should be dropped.

frenzibyte: back when it was in `MusicController`, it was quite useful to permit null to allow test scenes that need to test `MusicController`-dependent components (e.g. `NowPlayingOverlay`) not having to cache an `OnScreenDisplay`. 

Makes sense to permit null here as well because `OnScreenDisplay` is more like giving a bonus behaviour in this component rather than a required dependency for this to work.

bdach: i would have really been fine with non-null

bdach:
I've added some cosmetic changes here (which I was going to add yesterday before asking about the `OnScreenDisplay` non-null constraint) but I think other than that this is fine.

osu.Game/Overlays/Music/MusicActionHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Input.Bindings;
+using osu.Game.Beatmaps;
+using osu.Game.Input.Bindings;
+using osu.Game.Overlays.OSD;
+
+namespace osu.Game.Overlays.Music
+{
+    /// <summary>
+    /// Handles <see cref="GlobalAction"/> relating to music playback, and displays a <see cref="Toast"/> via the cached <see cref="OnScreenDisplay"/> accordingly.
```
peppy, line 15:
```suggestion
    /// Handles <see cref="GlobalAction"/>s related to music playback, and displays <see cref="Toast"/>s via the global <see cref="OnScreenDisplay"/> accordingly.
```

osu.Game/Overlays/Music/MusicActionHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+    /// </summary>
+    public class MusicActionHandler : Component, IKeyBindingHandler<GlobalAction>
```
peppy, line 17: `MusicKeyBindingHandler` sounds better, i think.

osu.Game/Overlays/Music/MusicActionHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+                    if (musicController.TogglePause())
+                        onScreenDisplay?.Display(new MusicActionToast(musicController.IsPlaying ? "Play track" : "Pause track"));
+
+                    return true;
+
+                case GlobalAction.MusicNext:
+                    musicController.NextTrack(() =>
+                    {
+                        onScreenDisplay?.Display(new MusicActionToast("Next track"));
+                    }).RunTask();
```
peppy, line 45: we cannot have this returned task running business. i have *no idea* why this is calling `RunTask`, but it can't. there's no logical reasoning and this code reads abhorrently.

peppy (changes requested):

osu.Game/Overlays/Music/MusicActionHandler.cs
frenzibyte: did it mostly just to not do any change in behaviour, I agree that it looks bad though, will keep it scheduled then.

peppy:
I've fixed the remaining issues I had with this PR. @smoogipoo requesting a quick check.

peppy (dismissed):

osu.Game/OsuGameBase.cs
```diff
@@ -64,6 +64,8 @@ public class OsuGameBase : Framework.Game, ICanAcceptFiles
 
         protected FileStore FileStore;
 
+        protected GlobalActionContainer GlobalBinding;
```
peppy, line 67: this should be renamed or simply not exposed. we don't expose things for tests any more. please use `ChildrenOfType` instead.

frenzibyte: I see, will take this into account in the future.

smoogipoo (approved):

PR #10074, opened 2020-09-07
Hide the game-wide cursor on touch input

### LINKED ISSUE #7437: Hide menu cursor on touch devices
**Describe the new feature:** Cursor showing up is not expected and looks bad on touch devices, and should be hidden. Cursor in std mode should not be hidden.

**Proposal designs of the feature:** When touch input is used, an invisible drawable would be drawn in place of the cursor.

### LINKED ISSUE #10046: Loading Layer blocks touch input
**Describe the bug:**
The loading animation seems to block input from touch devices when you try to press on it to move the cursor.
It seems to block input from overlays in front of it, too.

> **Loading Cover in PlayerLoader:**
loading cover

> **Trying to touch the (loading) Multiplayer lounge list:**
lounge

> **Trying to scroll the settings in the loading region of the Multi lounge:**
multiplayer lounge loading + settings

**osu!lazer version:** 2020.825.0

peppy:
I looked at fixing this locally on the `LoadingLayer` and while it is possible, this makes more sense.

Closes https://github.com/ppy/osu/issues/10046.
Closes https://github.com/ppy/osu/issues/7437.

Tested on iOS simulator only, but should be ample based on simplicity of implementation.

### DIFF
--- a/osu.Game/Graphics/Cursor/MenuCursorContainer.cs
+++ b/osu.Game/Graphics/Cursor/MenuCursorContainer.cs
@@ -48,7 +48,7 @@ namespace osu.Game.Graphics.Cursor
         {
             base.Update();
 
-            var lastMouseSource = GetContainingInputManager().CurrentState.Mouse.LastSource;
+            var lastMouseSource = inputManager.CurrentState.Mouse.LastSource;
             bool hasValidInput = lastMouseSource != null && !(lastMouseSource is ISourcedFromTouch);
 
             if (!hasValidInput || !CanShowCursor)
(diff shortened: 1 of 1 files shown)


frenzibyte:
I think this should be enough to close https://github.com/ppy/osu/issues/7437 as any touchscreen device should now have their cursor hidden on first touch input.

peppy:
It will be hidden before the first touch input, too. It will only be unhidden on a non-touch input.

smoogipoo (approved):

frenzibyte (approved):

ThatsNiceGuy:
this feels like pestering honestly but can we still have it be a toggle later?

peppy:
No, there's no reason to have it displayed.

HUOYI159:
I would love it to add a switch, because since the cursor disappeared, it's getting uncomfortable, because I'm stuck with the last version (Android) where the cursor appeared, and it makes OSU! It becomes more vivid

HUOYI159:
I would love it to add a switch, because since the cursor disappeared, it's getting uncomfortable, because I'm stuck with the last version (Android) where the cursor appeared, and it makes OSU! It becomes more vivid

Issue #10067, opened 2020-09-05
Continues running as a background process after exiting


JajaUmaOfficial:
**Describe the bug:** 
osu! lazer continues running as a background process even after exiting.
I cloned the repo and built locally; running on Windows 10 64-bit.

**Screenshots or videos showing encountered issue:** 
https://imgur.com/a/n1QZZVa

**osu!lazer version:** 
2020.903.0

**Logs:**
[performance.log](https://github.com/ppy/osu/files/5178589/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5178590/runtime.log)
[database.log](https://github.com/ppy/osu/files/5178591/database.log)
[network.log](https://github.com/ppy/osu/files/5178592/network.log)

bdach:
Can reproduce, but it's not consistent. It's usually accompanied by star ratings on song select not showing before, so background calculations are the immediate suspect.

JajaUmaOfficial:
I don't think this is what's happening for me; this occurs consistently, even when I just open osu! lazer and exit without opening any menus.
CPU usage is 0 and I see no running tasks after closing and pausing execution in debug mode.
Reattaching the debugger to the process shows a constant memory usage of ~427MB and no CPU usage.

bdach:
then could you please give more details as to what's happening? any weird operations going on any of the remaining threads?

peppy:
creating a memory dump from task manager and sending to pe@ppy.sh would be appreciated

JajaUmaOfficial:
Sent memory dump.

I did some testing on another windows 10 (19041.450) machine and there's no issue.
I tried comparing thread names between the 2 machines[ just before the program exits](https://github.com/ppy/osu/blob/master/osu.Desktop/Program.cs#L63). 
Lists of thread names are attached; nothing looks out of the ordinary as far as I can tell.

[Threads-Good.pdf](https://github.com/ppy/osu/files/5179068/Threads-Good.pdf)
[Threads-Bad.pdf](https://github.com/ppy/osu/files/5179069/Threads-Bad.pdf)

Let me know if there's any better info I can provide; thanks.

bdach:
The raw number of threads in the "bad" example is higher. If I were to bet I would suspect one of the worker threads, although it's impossible to tell without stack traces, but hopefully they'll be visible in the dump you sent anyway.

JajaUmaOfficial:
[...]

PR #10076, opened 2020-09-07
Avoid endless mania beatmap conversion

smoogipoo:
Fixes the beatmap mentioned in https://github.com/ppy/osu/issues/10067


peppy (dismissed):

peppy:
Worth adding test coverage for this? Probably not much benefit..

smoogipoo:
Can do, gimme a few.

peppy (approved):

PR #10079, opened 2020-09-07
Implement very primitive collection management (imports from stable)

smoogipoo:
Preview: https://drive.google.com/file/d/1wyta0ozlhiA1cDmr0yjoZ_3fJ7exiHhs/view?usp=sharing

This isn't the final product of this in either design or functionality, just something to get out there as I think it's one of the most important aspects missing from lazer.

Imports from stable's `collection.db`.

Bit of a chunky PR, let me know if you want something split out - most of it is design work :/

The only miscellaneous change I left in is a fix to `IconButton`, where it's using transforms internally that would override the setter of `IconButton.IconColour`. I perceive this to be a pretty harmless 1-line change, but I can split it out if desired.


smoogipoo:
Before this gets asked - we decided to not use EF for this since we'll soon be switching over to Realm and using EF for this would result in a refactoring of the DB structure at the least. As such, the flow of saving the database is similar to that of `IniConfigManager` for an acceptable level of safety.

peppy:
Cool stuff. Just going from the video though, this is going to require some design and UX adjustments for sure... it makes my brain spin watching the current flow.

@arflyte are you able to watch the attached video on this issue and suggest any changes you would make or any concerns as to what is going on? Not asking for a full design here, but looking for an external opinion from someone that doesn't actually play the game as to whether you can understand what is going on and how to use it.

arflyte:
Was confused with everything until I see the collection management. The dropdown menu and its position is very confusing. Using checkboxes would be better for the dropdown menu.  The management menu is almost there, definitely need a proper design for that.

Overall, I understood what's happening in the video.

peppy:
Changing the dropdown (green / red icons) to checkboxes sounds good for now, let's go with that.

Can probably just use font-awesome https://fontawesome.com/icons/square?style=regular and https://fontawesome.com/icons/square (should be white, no colour).

peppy:
While reviewing this, I noticed the presence of

`CollectionsDropdown` <- select a collection for playing in the playlist control
`CollectionSettings` <- add the current beatmap to a collection from replay etc.

We probably want to hook these up to the new system.

peppy:
Non-code review pass:

# Usability issues (easy fixes)

- [x] Rename right-click menu to "Collections" and move above "Delete"
- [ ] Right click menu should have a ▶︎ for expanding items (separate PR)
- [ ] Add close or back button to dialog
- [ ] Give "delete" button more spacing (it's not wide enough and looks weird, should be "square" dimensions or so)
- [ ] Don't prompt for delete if collection is empty
- [ ] Show number of items in collection (preferrably on the "manage" dialog, but failing that, at least on the delete dialog)
- [ ] Clicking "new collection" in the manage dialog should focus the text field of the new collection (and have the text selected or be empty).

# "What's going on"

- [ ] Clicking "More..." from the right click menu doesn't let you actually add to a collection.
- [ ] There should be a "New collection..." option in the dropdown for discoverability. Otherwise the only way to actually make one is via the right-click menu.

Also hold off on implementing the checkboxes option in the dropdown. I am starting to disagree with this because it reads more like "checked collections are currently being displayed", not "beatmap is contained in these collections".

smoogipoo:
I've completed most of the above items, with a few caveats:

- I've made a few modifications to `OsuRearrangeableListContainer` - exposing + renaming the dragging bindable, and exposing the `ShowDragHandle` property as a bindable instead of ctor-time property.
- Regarding showing the number of beatmaps contained, I've chosen to only show it at deletion time. Adding it to the management dialog looks kinda complex from a layout perspective (a pill on the right side that overlaps the textbox?).
- I've also chosen not to complete the item to allow adding from the management dialog. I agree with in in principle, but similar to the above - placing it in the dialog seems complex, and I'd like to wait for flyte's input on how those buttons should be laid out (currently the delete button underlays the textbox a bit, so maybe it there could be two buttons or something?).

peppy:
Since the flow from the "New collection" option in the dropdown doesn't actually create the new collection (or focus the "new" field) maybe we should rename to "Manage collections..." for now?

smoogipoo:
Yeah, I actually had it that way at first :P (Manage collections...)

osu.Game/OsuGame.cs
```diff
@@ -616,6 +620,7 @@ protected override void LoadComplete()
             loadComponentSingleFile(CreateUpdateManager(), Add, true);
 
             // overlay elements
+            loadComponentSingleFile(new ManageCollectionsDialog(), overlayContent.Add, true);
```
peppy, line 623: Is there a reason this is a global thing? Seems like something we wouldn't want global since it's only used in one place.

osu.Game/OsuGameBase.cs
```diff
@@ -225,6 +228,9 @@ List<ScoreInfo> getBeatmapScores(BeatmapSetInfo set)
             dependencies.Cache(difficultyManager);
             AddInternal(difficultyManager);
 
+            dependencies.Cache(CollectionManager = new BeatmapCollectionManager(Storage));
+            AddInternal(CollectionManager);
```
peppy, line 232: This is adding an extra background thread on startup (`loadDatabase()`), which I'm not sure we want. Any reason this is done in `GameBase` and not `Game`? Just to stay in line with the other "manager"s?

osu.Game/Collections/BeatmapCollectionManager.cs
```diff
@@ -0,0 +1,318 @@
...
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Logging;
+using osu.Framework.Platform;
+using osu.Game.Beatmaps;
+using osu.Game.IO.Legacy;
+using osu.Game.Overlays.Notifications;
+
+namespace osu.Game.Collections
+{
+    public class BeatmapCollectionManager : CompositeDrawable
```
peppy: rename to `CollectionManager` maybe? not sure we will ever have another type of collection (and the variable in `OsuGameBase` is already referring to it as `CollectionManager`).

peppy (changes requested):

osu.Game/OsuGame.cs
smoogipoo: Hmm... The reasoning is the dim on `OsuFocusedOverlayContainer` doesn't behave nicely if it's placed anywhere else, I suppose I can add a local backgrop to the dialog, or maybe not dim, if you'd rather that?

peppy: Did you try using `DialogOverlay.Push`?

smoogipoo: That requires it to be a `PopupDialog`, which is structurally/visually completely different and isn't (at this point) geared towards having arbitrary content placed inside it. That was the direction I was trying for originally but I thought the complexity of doing so wasn't worth it, however now that the hierarchy is simplified a bit I can give it a shot.

peppy: Probably not going to work well since it shows a delete dialog and i don't think multiple overlapping dialogs are supported.

I'm not sure I see the issue with dimming, i'll try local so i can understand it myself.

osu.Game/OsuGameBase.cs
smoogipoo: Mostly to stay in line with other managers, yes. I think it can be moved.

osu.Game/OsuGame.cs
smoogipoo: The issue is that the dim is applied to the game's content, which means the dialog would also be dimmed itself if placed inside SongSelect.

peppy: i see, unfortunate. would really like to reduce this load-on-startup components especially in cases they may never be used.

let's leave it for now though, that can be a future optimisation. the only weird part about doing it this way is the rearrangeable list animates only on first display.

osu.Game/OsuGameBase.cs
smoogipoo: In hindsight it's probably fine to not spawn the extra thread - I mostly did it as a safety insurance because it's querying EF, but that should be fast anyway since it's querying an indexed column...

peppy: I definitely think it should be threaded, because if someone has a 100mb collection database it shouldn't delay visual startup. Probably something that could be in the `loadComponentSingleFile` queue?

smoogipoo: Yeah, that'd also be fine.

osu.Game/Screens/Select/CollectionFilter.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// Whether the collection contains a given beatmap.
+        /// </summary>
+        /// <param name="beatmap">The beatmap to check.</param>
+        /// <returns>Whether <see cref="Collection"/> contains <paramref name="beatmap"/>.</returns>
+        public virtual bool ContainsBeatmap(BeatmapInfo beatmap)
+            => Collection?.Beatmaps.Any(b => b.Equals(beatmap)) ?? true;
+    }
+
+    public class AllBeatmapCollectionFilter : CollectionFilter
```
peppy, line 49: are you okay with renaming all `CollectionFilter` suffixes to `CollectionMenuItem` or similar? really weird as it stands..

peppy (changes requested):

osu.Game/Screens/Select/CollectionFilter.cs
smoogipoo: I did it this way because it's stored and used by `FilterCriteria`, and having that store a "MenuItem" sounds even more wrong.

smoogipoo: Maybe it deserves another level of abstraction, something like `CollectionMenuItem` containing the filter as its value...

peppy (approved):
Going to go with this as an initial implementation 👍

PR #10082, opened 2020-09-07
Update section headers in ProfileOverlay in line with web

EVAST9919:
Old:
old
New:
new
Web:
web


osu.Game/Overlays/Profile/Sections/PaginatedContainerHeader.cs
```diff
@@ -0,0 +1,132 @@
...
+            };
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            current.BindValueChanged(onCurrentChanged);
+        }
+
+        private float getInitialCounterAlpha()
```
smoogipoo, line 99: Why do it this way?

How about changing to `BindValueChanged(onCurrentChanged, true)` above, and then combining this into `onCurrentChanged`? The call to this can then be removed from `load()`.

osu.Game/Overlays/Profile/Sections/PaginatedContainerWithHeader.cs
```diff
@@ -0,0 +1,34 @@
...
+using osu.Game.Users;
+
+namespace osu.Game.Overlays.Profile.Sections
+{
+    public abstract class PaginatedContainerWithHeader<TModel> : PaginatedContainer<TModel>
+    {
+        private readonly string headerText;
+        private readonly CounterVisibilityState counterVisibilityState;
+
+        protected PaginatedContainerHeader Header;
```
smoogipoo, line 15:
```suggestion
        protected PaginatedContainerHeader Header { get; private set; }
```

osu.Game/Overlays/Profile/Sections/CounterPill.cs
```diff
@@ -36,8 +33,8 @@ private void load(OverlayColourProvider colourProvider)
                 {
                     Anchor = Anchor.Centre,
                     Origin = Anchor.Centre,
-                    Margin = new MarginPadding { Horizontal = 10, Vertical = 5 },
-                    Font = OsuFont.GetFont(weight: FontWeight.Bold),
+                    Margin = new MarginPadding { Horizontal = 10, Bottom = 1 },
+                    Font = OsuFont.GetFont(size: 14 * 0.8f, weight: FontWeight.Bold),
```
smoogipoo, line 37: Replace this with the actual value?

smoogipoo (changes requested):
I suppose it's fine, but I'm not sold on the complexity of this PR. At the point where "missingText" is an optional parameter, why couldn't the header also be? In that scenario, everything needed for this was already present in the base class, with the only thing missing being the enum and the restyling.

smoogipoo (approved):
lgtm

PR #10084, opened 2020-09-08
Avoid requiring sending the calling method for CleanRunHeadlessGameHost

peppy:
Uses solution proposed [here](https://github.com/ppy/osu/pull/10047#issuecomment-688484852) to avoid an easily overlooked parameter causing test breakage.


smoogipoo (approved):
Every day you learn something new. Very cool.

PR #10101, opened 2020-09-09
Add distance snap toggle (and support for future arbitrary composer toggles)

peppy:
Adds a toggle for distance snap, as well as facilities for rulesets to add/display arbitrary toggles. We may need to expand this into a class definition in the future, rather than just a bindable list, but this is a good starting point.

- [x] Depends on https://github.com/ppy/osu/pull/10100


smoogipoo:
Slider path placement shouldn't snap (I think?) but that's for a separate PR, because the same will happen for anything that references the `GetSnapped*` methods in HOC (and I'm not exactly sure those should be straight up disabled with the toggle).

smoogipoo (approved):

peppy:
slider paths should still snap to beat snap. it's a bit different from distance snap

PR #10102, opened 2020-09-09
Prompt to save changes when exiting the editor

peppy:


smoogipoo (approved):

smoogipoo:
I don't think the previous failure (https://ci.appveyor.com/project/peppy/osu/builds/35109154) is a result of this PR, so I think I'll merge as is.

PR #10106, opened 2020-09-09
Implement dynamic score recalculation

smoogipoo:
Preview: https://drive.google.com/file/d/1WtusOV9-_C2Pgn-cWX10TrJ6pno6KCwZ/view?usp=sharing

I'm really interested to hear opinions on this. The purpose is to recalculate scores using the standardised/classic scoring algorithms to unify scoring a more in leaderboards.

Eventually we're going to have to have some osu!web-side recalculation based probably on standardised score (assuming classic <-> standardised scoring closely follow each other), which is going to help with returning players around the user for leaderboard displays, so this is only a first step.

Works in:
- Song select leaderboard
- Solo results screen
- Beatmap set overlay

Does not work in multiplayer results screen (will always display standardised scoring) due to osu!web not providing `max_combo` along with the beatmap. I haven't made an issue for this, but will if the direction proposed here is acceptable.

A few changes of note:
1. All `DifficultyAttributes` now contain max combo.
2. `ScoreProcessor` exposes a *non-static* method to recompute score using minimal statistics, in order to adhere to its own per-ruleset `DefaultAccuracyPortion`/`DefaultComboPortion` values.
3. (Most) places where `scoreInfo.TotalScore` was referenced are replaced with `scoreManager.GetBindableTotalScoreString()` (follows the API provided by `BeatmapDifficultyManager`).

Also, this doesn't re-order leaderboards, including results screen. I don't know whether this is something that should be done at this point in time rather than after osu!web-side recalculation is present, since we don't have full data.


osu.Game/Rulesets/Scoring/ScoreProcessor.cs
```diff
@@ -202,20 +202,31 @@ private void updateScore()
...
-                    double comboScore = comboPortion * HighestCombo.Value / maxHighestCombo;
+                    double accuracyScore = accuracyPortion * accuracyRatio;
+                    double comboScore = comboPortion * comboRatio;
 
                     return (max_score * (accuracyScore + comboScore) + bonusScore) * scoreMultiplier;
 
                 case ScoringMode.Classic:
                     // should emulate osu-stable's scoring as closely as we can (https://osu.ppy.sh/help/wiki/Score/ScoreV1)
-                    return bonusScore + baseScore * (1 + Math.Max(0, HighestCombo.Value - 1) * scoreMultiplier / 25);
+                    return bonusScore + (accuracyRatio * maxCombo * 300) * (1 + Math.Max(0, (comboRatio * maxCombo) - 1) * scoreMultiplier / 25);
```
smoogipoo, line 229: Note that this `* 300` is hardcoded now. It mostly matches except for mania which uses 350 as its top judgement, but this function doesn't do a good job at approximating "classic" mania in the first place (standardised is closer).

That said I think this is fine given that this is trying to match stable scoring, which is based on 300s, and saves having to compute the maximum `baseScore` which requires having immediate web implementation/processing for things like leaderboards.

peppy:
Pretty happy with the structure and feel of this. I guess the only thing missing is correct bonus score calculation (spin etc.), but that's definitely a discussion for another day.

peppy (approved):

PR #10121, opened 2020-09-10
Fix song select filter ordering

### LINKED ISSUE #10120: Collections' dropdown menu stays above "Sort by" dropdown
**Describe the bug:** This probably happens only in the carousel. I can't think of other places with overlapping dropdowns.

dropdown

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10120

`GridContainer` probably needs a way to control the depth ordering of its cells, but this'll do for now - everything was absolute size anyway.


bdach (approved):
seems good :+1:

peppy (approved):
works. one less grid yay.

PR #10126, opened 2020-09-10
Fix drag handles not showing on now playing playlist items

Joehuu:
Regressed in https://github.com/ppy/osu/commit/1260e30cde06611ca403a972a54403879c338223.

Not sure if it should be true by default in `OsuRearrangeableListItem`.


bdach:
I think *not* having drag handles on by default on a rearrangeable list is weird. `true` should probably be the default.

bdach:
I've taken the liberty of adding test coverage for this. It arguably duplicates the framework tests a little, but seeing that we now have extra functionality game-side in being able to hide the handles, I'd say it's useful to have going forward.

smoogipoo (approved):

Issue #10115, opened 2020-09-10
when importing a skin, overwrite skins with the same metadata as the skin your importing


igontusuwu:
i suggest this so that we can avoid having duplicate/entries with a element changed when the intention is to just update the skin

this reduces clutter and duplicates on the skin selection tab

enoslayd:
Iirc, If you didnt change anything in skin.ini it will overwrite the existing one

peppy:
This should already be the case, correct.

Can you provide more details of what you're seeing? Perhaps the skin in question doesn't have a `skin.ini` file?

enoslayd:
Example. You imported a skin then change just one detail (e.x. colour1 value) and reimported it. It wont overwrite the first imported one cause of different hash value (idk the right term) so lazer treats them as two different skin even though they have the same name and author (and maybe even the skin version). But if you change an asset instead (e.g. hitcircle.png) but didnt touch the skin.ini, it will be overwritten and put at the bottom of the list. Thats how i understand/think it works

peppy:
I'm hoping to get an answer from the opening poster, since changing the skin.ini is quite rare.

enoslayd:
Changing skin.ini is not a rare occurence (at least for me).  Especially since most of mania's advanced skin configs (e.g. placing the playfield to the left or center) is in skin.ini

enoslayd:
Changing skin.ini is not a rare occurence (at least for me).  Especially since most of mania's advanced skin configs (e.g. placing the playfield to the left or center) is in skin.ini

igontusuwu:
i did a bit of test and i found that:
putting a skin in with different metadata than another skin makes it a separate skin
putting a skin with the same metadata overwrites the old skin, regardless of element changes

it seems like everything is working normally but i'm not sure how there are duplicate skins in my skin tab, i think it's related to a small metadata change or something
i'll close this now

PR #10133, opened 2020-09-11
Determine uniqueness of legacy skins based on the author/name combo

peppy:
- [x] Merge https://github.com/ppy/osu/pull/10171 first

Addresses concerns raised in https://github.com/ppy/osu/issues/10115. Adds test coverage for skin imports.


enoslayd:
Can we also add the skin version? Just a suggestion

peppy:
Skin "version" is a target version, not the actual version of a skin. I've left a comment regarding this:

https://github.com/ppy/osu/pull/10133/files#diff-55e8b74687bad5e8320af07b757aee7bR96-R97

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -318,10 +321,11 @@ void rollback()
                 LogForModel(item, "Beginning import...");
 
                 item.Files = archive != null ? createFileInfos(archive, Files) : new List<TFileModel>();
-                item.Hash = computeHash(item, archive);
 
                 await Populate(item, archive, cancellationToken);
 
+                item.Hash = ComputeHash(item, archive);
```
bdach, line 327: Moving this after `Populate()` is breaking beatmap import tests, as `BeatmapManager.createBeatmapDifficulties()`, which is called by its overridden `Populate()` implementation, is relying on the hash already being there at this point to determine whether there is a need to import instead of e.g. restoring a deleted item.

osu.Game/Skinning/SkinManager.cs
```diff
@@ -86,6 +87,21 @@ public List<SkinInfo> GetAllUsableSkins()
...
+        protected override string ComputeHash(SkinInfo item, ArchiveReader reader = null)
+        {
+            if (item.Creator != null && item.Creator != unknown_creator_string)
+            {
+                // this is the optimal way to hash legacy skins, but will need to be reconsidered when we move forward with skin implementation.
+                // likely, the skin should expose a real version (ie. the version of the skin, not the skin.ini version it's targeting).
+                return item.ToString().ComputeSHA2Hash();
+            }
+
+            // if there was no creator, the ToString above would give the filename, which along isn't really enough to base any decisions on.
```
bdach, line 101:
```suggestion
            // if there was no creator, the ToString above would give the filename, which alone isn't really enough to base any decisions on.
```

bdach (changes requested):

osu.Game/Database/ArchiveModelManager.cs
peppy: I've fixed this in the simplest way possible (locally). I think for non-legacy skins this will be handled in a better way so I'm not too concerned about code quality around this logic.

osu.Game.Tests/Skins/IO/ImportSkinTest.cs
```diff
@@ -68,7 +68,7 @@ public async Task TestImportTwiceWithSameMetadata()
         [Test]
         public async Task TestImportTwiceWithNoMetadata()
         {
-            using (HeadlessGameHost host = new CleanRunHeadlessGameHost())
+            using (HeadlessGameHost host = new CleanRunHeadlessGameHost(nameof(ImportSkinTest)))
```
smoogipoo, line 71: You didn't mean this, right? Should be `nameof(TestImportTwiceWithNoMetadata)` (same as all other ones i this commit)

peppy: This is intentional, check the constructor (this creates a prefix, in addition to the method name).

It's differentiating `ImportSkinTest.TestBasic` from `ImportBeatmapTest.TestBasic`.

smoogipoo (approved):
I'm not really happy with the changes to the tests because this is definitely going to bite us again, but let's go with this for now...

peppy:
I'll look at fixing it to avoid any name specifications in a follow-up PR.

PR #10134, opened 2020-09-11
Add collections filter to the now playing overlay

### LINKED ISSUE #10125: Viewing collections from playlist overlay and adding beatmaps from replay not implemented
https://github.com/ppy/osu/pull/10079#issuecomment-688588958

Comment got lost in the huge amount of commits.

**osu!lazer version:** 2020.903.0

smoogipoo:
Resolves https://github.com/ppy/osu/issues/10125

- Removes the `CollectionSettings` player group. Do we even want this in player? Seems like too many things happening from too many places (maybe in the future you could use the music controller to do this? The button is hard disabled, but no reason why it should be.). I can add it back if so.
- Implements collection filters in the `NowPlayingOverlay`.

I've disabled the "manage collections" menu item from the now playing overlay because the overlap doesn't work well:
Screenshot_2020-09-11_16-55-16
At the same time, I don't feel great about putting the collections dialog at a higher level (basically has to be a topmost overlay), as it'd be above the volume controller too and doesn't play well with settings. Thinking this should probably be integrated into the now playing overlay somewhat in the future.


peppy:
Agree with removing from player for now.

peppy (approved):

PR #10139, opened 2020-09-11
Fix editor selected hitobjects containing the selection up to five times

peppy:
Due to doing the add in two places, having two `SelectionHandlers` (timeline + composer) and no feedback protection, any selected objects were added to the list five times. Wow.


osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -145,10 +145,16 @@ public void OnReleased(PlatformAction action)
         /// <param name="blueprint">The blueprint.</param>
         internal void HandleSelected(SelectionBlueprint blueprint)
         {
-            selectedBlueprints.Add(blueprint);
-            EditorBeatmap.SelectedHitObjects.Add(blueprint.HitObject);
+            if (!selectedBlueprints.Contains(blueprint))
```
smoogipoo, line 148: Isn't this guaranteed to be false? The only way this can happen is if `onBlueprintSelected()` is called multiple times, which seems to be impossible due to the implementation of `SelectionBlueprint.State` (where it avoids multiple events if already selected).

I.e. this code could be reduced to just
```
selectedBlueprints.Add(blueprint);

if (!EditorBeatmap.SelectedHitObjects.Contains(...)
    ...

UpdateVisibility();
```

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -157,10 +163,12 @@ internal void HandleSelected(SelectionBlueprint blueprint)
         /// <param name="blueprint">The blueprint.</param>
         internal void HandleDeselected(SelectionBlueprint blueprint)
         {
-            selectedBlueprints.Remove(blueprint);
-            EditorBeatmap.SelectedHitObjects.Remove(blueprint.HitObject);
+            if (selectedBlueprints.Remove(blueprint))
```
smoogipoo, line 166: Same here, this change isn't really required as it's guaranteed to be true.

smoogipoo (changes requested):

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
peppy: sure, can be removed.

peppy:
Don't merge this just yet, I think I found further issues.

smoogipoo (approved):

PR #10140, opened 2020-09-11
Editor clipboard support

peppy:
Tentatively pushing this as it's in a working state, but needs some fixed to serialization (@smoogipoo may get to them before I do, we'll see).

- Fixes incorrect ordering of `ApplyDefaults` for newly added objects, meaning it may not have been run in time.
- [x] Fix slider serialization (path is not transferring correctly).
- [x] Framework bump and using `BindCollectionChanged` for initial disabled states
- [x] Add tests
- [x] Select on paste
- [x] Depends on https://github.com/ppy/osu/pull/10139 to not paste five times over
- [x] Depends on https://github.com/ppy/osu/pull/10152 for slider and spinner support


osu.Game/Screens/Edit/Editor.cs
```diff
@@ -244,6 +256,17 @@ private void load(OsuColour colours, GameHost host)
             changeHandler.CanUndo.BindValueChanged(v => undoMenuItem.Action.Disabled = !v.NewValue, true);
             changeHandler.CanRedo.BindValueChanged(v => redoMenuItem.Action.Disabled = !v.NewValue, true);
 
+            // todo: BindCollectionChanged
```
bdach, line 259: looks removable

osu.Game/Screens/Edit/Editor.cs
```diff
@@ -394,6 +429,43 @@ private void confirmExit()
...
+        protected void Paste()
+        {
+            if (string.IsNullOrEmpty(clipboard.Value))
+                return;
+
+            var objects = clipboard.Value.Deserialize<ClipboardContent>().HitObjects;
+
+            Debug.Assert(objects.Any());
+
+            double timeOffset = clock.CurrentTime - objects.First().StartTime;
```
bdach, line 458: This `First()` doesn't work in terms of "beatmap time", but in terms of selection order (namely, the first object selected will be pasted in at `CurrentTime` and the rest relative to it, and not necessarily the earliest). Unsure what's the user expectation here; I think I would want timeline order and not selection order but I don't map.

peppy: wow, select objects should probably definitely be sorted..

peppy:
This should be good to go now (apart from the remaining prereq). Note that there is a bit of adjustment to `EditorBeatmap`'s schedule/debounce logic, but not as much as we discussed in discord. I think it should be easy enough to review as part of this PR, but let me know if you'd prefer it split out.

smoogipoo (approved):
Good stuff :+1:

PR #10141, opened 2020-09-11
Implement bonus result judgements/statistics

smoogipoo:
PRing for comments, not merge, because I don't like the way this is going and want to start up a conversation while brainstorming how this flow should work.

- [x] `IsHit` doesn't consider `SmallTickMiss` or `LargeTickMiss`.
- [x] `ScoreProcessor.ApplyResult`/`RevertResult` don't consider combo breaks on `SmallTickMiss` or `LargeTickMiss`.
- [ ] Performance calculator support.
- [ ] Need to fix the hacky auto statistic writing of `ScoreProcessor`.
- [ ] I don't like the `ScoreProcessor.GetNumericBonusResult()` method, but I don't know how to improve this yet.

It feels like bonuses, and potentially ticks, should be kept separate to the main hit results somehow. It's all getting to be too much of a mess.

### DIFF
--- a/osu.Game.Rulesets.Catch/Judgements/CatchBananaJudgement.cs
+++ b/osu.Game.Rulesets.Catch/Judgements/CatchBananaJudgement.cs
@@ -8,31 +8,13 @@ namespace osu.Game.Rulesets.Catch.Judgements
 {
     public class CatchBananaJudgement : CatchJudgement
     {
+        public override HitResult MaxResult => HitResult.LargeBonusHit;
+
         public override bool AffectsCombo => false;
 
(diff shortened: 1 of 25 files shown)


bdach:
I keep meaning to go through this properly instead of skimming (as it seems like the sort of change that requires that) but I seem to never be able to get to it in time... Does this need review still, or is there a direction set already? If not then I can put priority on reading through this over the coming days.

smoogipoo:
@bdach I'm still deep in thought over this and trying a few other paths.

smoogipoo:
Closing this since I have a better path forward.

PR #10146, opened 2020-09-12
Ensure track is looping in song select immediately

### LINKED ISSUE #10127: Song select tests intermittently fail
Opening for longer form discussion, as I consider this not fit for discord.

Song select tests have [begun](https://ci.appveyor.com/project/peppy/osu/builds/35136187) [to](https://ci.appveyor.com/project/peppy/osu/builds/35137301) [fail](https://ci.appveyor.com/project/peppy/osu/builds/35137619) a lot. As far as I can tell the failures are *not* due to the beatmap fail exceptions (they're a red herring, everything falls back gracefully, *although* the exceptions should probably be silenced to not mislead), but they are indirectly caused by the music controller changes.

Here's one of the relevant parts of one failing test:

https://github.com/ppy/osu/blob/8d550a7b29c90940163fd3d67546087f6f01cd2b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs#L385-L396

Because the maps in the scene don't have actual tracks, they use the default 1s virtual track. `MusicController` has a callback registered on track completion, to automatically navigate to the next track:

https://github.com/ppy/osu/blob/d09289b1ce52eefbed3dee61413187ae50e9ba52/osu.Game/Overlays/MusicController.cs#L347-L348

Because of this, it is possible for the following thing to happen:

* The first step succeeds, beatmap and rulesets are changed correctly
* In between that step's completion and one of the asserts the virtual track completes
* Once the virtual track completes, the beatmap is changed away to another one
* The assert fails because the new map is not the one the test is looking for

"Funnily" enough, because of the above scenario running normally will very often be fine locally, but if you breakpoint at the steps, it'll fail every time because the audio thread has the time to execute the track completion callback in the background. Breakpointing at the ruleset value set (so one line after the beatmap is set) pretty much guarantees losing the race every time.

The best idea I've had so far of dealing with this is the following very stupid patch:

```diff
diff --git a/osu.Game/Beatmaps/WorkingBeatmap.cs b/osu.Game/Beatmaps/WorkingBeatmap.cs
index d9780233d..ac1225234 100644
--- a/osu.Game/Beatmaps/WorkingBeatmap.cs
+++ b/osu.Game/Beatmaps/WorkingBeatmap.cs
@@ -264,7 +264,7 @@ public IBeatmap Beatmap
         private Track loadedTrack;
 
         [NotNull]
-        public Track LoadTrack() => loadedTrack = GetBeatmapTrack() ?? GetVirtualTrack(1000);
+        public Track LoadTrack() => loadedTrack = GetBeatmapTrack() ?? GetVirtualTrack(Math.Max(BeatmapInfo.Length, 1000));
 
         /// <summary>
         /// Transfer a valid audio track into this working beatmap. Used as an optimisation to avoid reload / track swap
```

which allows to simulate long tracks by passing the length in metadata, but I'm 200% aware how dumb of an idea that is. (Or, use an actual backing track for the map, I suppose...)

Because the test uses `BeatmapManager` directly, there's not much leeway to work with here without entirely ripping out the manager and swapping it for something we can instrument/mock (which sort of defeats the purpose as it was a good integration test to begin with...)

I suppose the tests affected can also take the snapshot of both bindable values and assert on that instead, but that still doesn't really eliminate the race entirely and also weakens the test to the point that I'm not entirely sure it'd be useful anymore anyway.

Edit: This also very visibly falls over in test browser.

bdach:
Resolves #10127.

# Summary

I'm PRing this mostly for feedback. As I mentioned in the issue thread, tests exposed that it was possible for tracks to complete and switch away on their own in song select instead of looping, therefore causing a potential race between asserts checking the track and the track switching away in song select tests that caused the intermittent failure.

The reason it was possible for the track to switch away is that because the looping flag was being set in the following call tree:

```
performUpdateSelected+run()
    updateComponentFromBeatmap()
```

which doesn't necessarily execute right away - it can be delayed up to 200ms if a map was previously selected:

https://github.com/ppy/osu/blob/5740fc2bd099b2141f2d5c9be7282611f769012c/osu.Game/Screens/Select/SongSelect.cs#L454-L457

The resolution aims to resolve this by subscribing to `MusicController.TrackChanged` to immediately set the loop flag on every track change handled by the music controller.

This can be tested by running `TestScenePlaySongSelect.TestPresentNewBeatmapNewRuleset` *in the visual test browser* (*not* headless with nunit runner, as in that variant asserts can pretty often win the race). In visual tests the test should fail almost every time and look wrong on `master`, and work with this branch.

# Remarks

[As discussed on discord](https://discord.com/channels/188630481301012481/188630652340404224/754297416039006269), the alternatives were:

* moving the `Looping = true` set somewhere where it would be executed immediately in the beatmap change callback, but it would have to be scheduled to be 100% safe, because the music controller *also* changes track in a beatmap change callback, so without the schedule we'd be relying on callback execution order for the fix to work,
* adding some sort of global field in `MusicController` that would have all loaded tracks loop - sucks because now there are two sources for one setting.

This resolution has some risk in it because events, so there's a potential for missed unsubscribes, but compared to the alternatives above I think it's way better. I've tested it a little bit and it seems to behave ok and pass tests, but this is the sort of change where I wouldn't necessarily trust that that's enough to catch all possible kinks.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -517,6 +517,8 @@ namespace osu.Game.Screens.Select
             FilterControl.Activate();
 
             ModSelect.SelectedMods.BindTo(selectedMods);
+
+            music.TrackChanged += ensureTrackLooping;
         }
 
         private const double logo_transition = 250;
@@ -568,6 +570,7 @@ namespace osu.Game.Screens.Select
             BeatmapDetails.Refresh();
 
             music.CurrentTrack.Looping = true;
+            music.TrackChanged += ensureTrackLooping;
             music.ResetTrackAdjustments();
 
             if (Beatmap != null && !Beatmap.Value.BeatmapSetInfo.DeletePending)
@@ -593,6 +596,7 @@ namespace osu.Game.Screens.Select
             BeatmapOptions.Hide();
 
             music.CurrentTrack.Looping = false;
+            music.TrackChanged -= ensureTrackLooping;
 
             this.ScaleTo(1.1f, 250, Easing.InSine);
 
@@ -614,9 +618,13 @@ namespace osu.Game.Screens.Select
             FilterControl.Deactivate();
 
             music.CurrentTrack.Looping = false;
+            music.TrackChanged -= ensureTrackLooping;
 
             return false;
         }
+
+        private void ensureTrackLooping(WorkingBeatmap beatmap, TrackChangeDirection changeDirection)
+            => music.CurrentTrack.Looping = true;
 
         public override bool OnBackButton()
         {
@@ -653,8 +661,6 @@ namespace osu.Game.Screens.Select
             beatmapInfoWedge.Beatmap = beatmap;
 
             BeatmapDetails.Beatmap = beatmap;
-
-            music.CurrentTrack.Looping = true;
         }
 
         private readonly WeakReference<ITrack> lastTrack = new WeakReference<ITrack>(null);
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -642,6 +642,7 @@ namespace osu.Game.Screens.Select
             base.Dispose(isDisposing);
 
             decoupledRuleset.UnbindAll();
+            music.TrackChanged -= ensureTrackLooping;
         }
 
         /// <summary>


frenzibyte (dismissed):
Sounds like a good solution, and also logical as it shows now that the current track would get looped if the user is in the song select screen.

smoogipoo:
I'd really like an unsubscription to be added to `Dispose()` just as a safeguard - screens don't necessarily have to be exited if their parenting `ScreenStack` is disposed.

Seems fine otherwise :+1:

smoogipoo (changes requested):
As mentioned

peppy (approved):

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -634,6 +642,7 @@ protected override void Dispose(bool isDisposing)
             base.Dispose(isDisposing);
 
             decoupledRuleset.UnbindAll();
+            music.TrackChanged -= ensureTrackLooping;
```
smoogipoo, line 645: this needs a `if (music != null)`.

Issue #10123, opened 2020-09-10
Adding the ability to add a map to the collection in the "options" section


Moltanicaa:
**Describe the new feature:** 
In the stable, we can click on f3 (options will open), then we can click on 1, after which the collection editing screen will open, where we can add a map. Here you can do the same, because it is quite convenient (in two clicks).
![image](https://user-images.githubusercontent.com/59089206/92761394-4ff21600-f39a-11ea-9332-6475bc0602db.png)

Joehuu:
Do we want stable's order? Right now the order is wrong:
![osu!_eVaTCtW6RY](https://user-images.githubusercontent.com/35318437/92964224-97bf7e80-f428-11ea-83f3-a0ff32542ade.png)

I would change it to this (please see if the design works):
![dotnet_0RUkih5Weu](https://user-images.githubusercontent.com/35318437/92964307-bcb3f180-f428-11ea-931d-ffdb54113f3e.png)

For multi, should we restrict it to this?
![osu!_yJil7OejRL](https://user-images.githubusercontent.com/35318437/92964399-dead7400-f428-11ea-9a51-842afe516ed9.png)

Played/unplayed will be visible from the beatmap carousel once it's implemented, but the leaderboards are not shown in multi song select now. Only edit is restricted for now.
![dotnet_buIjju1IfD](https://user-images.githubusercontent.com/35318437/92964493-0c92b880-f429-11ea-865f-4c5d68df3034.png)

peppy:
Order isn't too important, make it whatever makes sense.

PR #10149, opened 2020-09-13
Fix mod select overlay absorbing input from toolbar ruleset selector

### LINKED ISSUE #8001: Changing ruleset with key combination in ModSelectOverlay doesn't work with all numbers
**Describe the bug:** For example, if you're playing osu!mania and you try to switch to osu! with <kbd>Ctrl</kbd> + <kbd>1</kbd>, the input triggers the "Deselect All" button from ModSelectOverlay (like when you only press <kbd>1</kbd>) and doesn't change ruleset.

**osu!lazer version:** 2020.225.0

Joehuu:
Fixes https://github.com/ppy/osu/issues/8001.

This is because `ModSelectOverlay` is a `WaveOverlayContainer` and will prioritize input because it'll get focus when visible. I did a local fix in `ModSelectOverlay` for now, since I don't know the regressions this'll cause if put in `OsuFocusedOverlayContainer`.

In https://github.com/ppy/osu/issues/10123, I'm changing `BeatmapOptionsOverlay` to *handle* input of all the buttons to avoid hardcoded key params but the same exact issue happens. Once I PR that, it would have the same fix as done here.

### DIFF
--- a/osu.Game/Overlays/Mods/ModSelectOverlay.cs
+++ b/osu.Game/Overlays/Mods/ModSelectOverlay.cs
@@ -390,6 +390,9 @@ namespace osu.Game.Overlays.Mods
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
+            // don't absorb control as ToolbarRulesetSelector uses control + number to navigate
+            if (e.ControlPressed) return false;
+
             switch (e.Key)
             {
                 case Key.Number1:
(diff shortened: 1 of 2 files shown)


osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -390,6 +390,9 @@ protected override void PopIn()
 
         protected override bool OnKeyDown(KeyDownEvent e)
         {
+            // don't absorb control as ToolbarRulesetSelector uses control + number to navigate
+            if (e.ControlPressed) return false;
```
bdach, line 394: I don't really like this solution, but given that the "proper" way to do this (`KeyBindingContainer` - not even necessarily databased) is a lot more long-winded I guess I could stomach it if everyone else can.

smoogipoo (approved):
I think this is okay for now. Another way of looking at it is those buttons shouldn't handle ctrl+key inputs anyway.

PR #10156, opened 2020-09-14
Add manage collections button to beatmap options

### LINKED ISSUE #10123: Adding the ability to add a map to the collection in the "options" section

Joehuu:
Kinda resolves https://github.com/ppy/osu/issues/10123. I matched stable behavior, but the manage collections dialog/overlay doesn't support adding yet though.

Most changes are the input regression test as I made `BeatmapOptionsOverlay` handle input of the buttons now.

Copied the navigation code from `ToolbarRulesetSelector` with minor tweaks.
https://github.com/ppy/osu/blob/16d32f5246f0daf1cfaa967e909418cd336e4f8b/osu.Game/Overlays/Toolbar/ToolbarRulesetSelector.cs#L94-L109


smoogipoo (approved):
lgtm! :+1:

PR #10160, opened 2020-09-15
Add null check in SongSelect disposal for safety

peppy:
As per https://github.com/ppy/osu/pull/10146/files#r488395926


smoogipoo (approved):

bdach:
I even thought about it but then figured it might be fine without... My bad.

peppy:
The usual case this surfaces is if a test fails during construction of a drawable (before it is added to the hierarchy) or BDL prior to the DI resolution, and can cause confusion as the `Dispose` nullref appears more prominently than the true issue.

PR #10162, opened 2020-09-15
Fix mania scrollspeed slider precision

### LINKED ISSUE #9689: Improve usability of mania scroll speed slider

Morilli:
This should partly fix #9689 by adding a decently accurate keyboard step to the mania scroll speed slider, thereby making it possible to accurately change its value.

The major underlying issue imo is that the range of the slider is just way too large, but that should probably be addressed elsewhere.


smoogipoo (approved):
Sure, I think this is fine for now.

PR #10163, opened 2020-09-15
Pause playback in present tests to avoid track inadvertently changing at menu

peppy:
Similar to https://github.com/ppy/osu/issues/10127, but in these tests we return to the main menu, where looping is removed.

This seems like the best way to handle this for now..


bdach (approved):
seems to make tests behave better in visual test runner :+1:

smoogipoo (approved):

PR #10164, opened 2020-09-15
Fix tick miss judgments not breaking combo

smoogipoo:
This doesn't affect any of our rulesets, since we don't use these judgements yet. There are more issues with `IsHit`, but I'll tackle those in a separate PR...

Also fixes the miss judgements potentially awarding score, which was fixed for `HitResult.Miss` previously.

### DIFF
--- a/osu.Game.Tests/Gameplay/TestSceneScoreProcessor.cs
+++ b/osu.Game.Tests/Gameplay/TestSceneScoreProcessor.cs
@@ -31,37 +31,49 @@ namespace osu.Game.Tests.Gameplay
         [Test]
         public void TestOnlyBonusScore()
         {
-            var beatmap = new Beatmap<TestBonusHitObject> { HitObjects = { new TestBonusHitObject() } };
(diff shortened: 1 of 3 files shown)


smoogipoo:
Hold off on this a bit, want to try splitting ticks out of the results.

peppy:
I guess this needs merge/refactoring.

smoogipoo:
This is actually already fixed
https://github.com/ppy/osu/blob/c93d1631d4d2437bb4eb792943d68cc9e9c263d4/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L125-L126
https://github.com/ppy/osu/blob/c93d1631d4d2437bb4eb792943d68cc9e9c263d4/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L136

But I'll add an updated version of these tests later - should ensure the table is working as intended for all hit results.

Closing as this isn't needed anymore.

peppy:
figured that would probably be the case.

PR #10169, opened 2020-09-15
Fix gameplay samples potentially playing while player is paused

### LINKED ISSUE #10168: Changing skin settings during gameplay can trigger the miss sound
**Describe the bug:** When you change skin or the beatmap hitsound/skin toggles, the game will play the _miss_ sound (only during gameplay if at least one miss sound have been made)

**Screenshots or videos showing encountered issue:** [Video](https://drive.google.com/file/d/1MAaNaVZk-DZTD53-wSBdFScbmvRoZDJW/view?usp=sharing)

frenzibyte:
Closes #10168 

The underlying issue with this is a bug in the gameplay sample pause logic, where if the skin changes while player is paused, `SkinnableSound` would play the new created sample channels thinking they should be playing, becuase of using `requestedPlaying` as the condition for whether to do so.

I have fixed this by correcting the code to play the internal sample channels if they were already playing previously by using `IsPlaying` (which checks for `samplesContainer.Any(c => c.Playing)`). 

Also replaced `Play()` with `play()` to not interfere with the `requestedPlaying` state, as it shouldn't.

---
Added test case covering the issue.


osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
```diff
@@ -101,5 +107,58 @@ public void TestNonLoopingStopsWithPause()
...
+
+            AddStep("trigger skin change", () =>
+            {
+                skinSource.TriggerSourceChanged();
+            });
+
+            AddStep("retrieve new sample", () =>
+            {
+                DrawableSample newSample = skinnableSound.ChildrenOfType<DrawableSample>().Single();
+                Assert.IsTrue(newSample != sample, "Sample still hasn't been updated after a skin change event");
```
bdach, line 134: this is weird. why is there an assert *inside* a step? why couldn't this be an assert step (or, more reliably, a wait step)?

osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
```diff
@@ -101,5 +107,58 @@ public void TestNonLoopingStopsWithPause()
...
+                sample = newSample;
+            });
+
+            AddAssert("new sample paused", () => !sample.Playing);
```
bdach, line 138: i don't think "pausing" samples is ever a thing.

osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
```diff
@@ -101,5 +107,58 @@ public void TestNonLoopingStopsWithPause()
...
+
+            AddAssert("sample playing", () => sample.Playing);
+
+            AddStep("pause gameplay clock", () => gameplayClock.IsPaused.Value = true);
+            AddUntilStep("wait for sample to stop playing", () => !sample.Playing);
+
+            AddStep("trigger skin change", () =>
+            {
+                skinSource.TriggerSourceChanged();
+            });
```
bdach, line 129: can be a single line

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -138,8 +140,10 @@ protected override void SkinChanged(ISkinSource skin, bool allowFallback)
 
             samplesContainer.ChildrenEnumerable = channels.Select(c => new DrawableSample(c));
 
-            if (requestedPlaying)
-                Play();
+            // Sample channels have been reloaded to new ones because skin has changed.
```
bdach, line 143: redundant statement. this is inside a method called `SkinChanged`.

bdach (changes requested):

osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
frenzibyte: Pretty sure there are test scenes using it either here or at framework, but could make it an until step, sure.

osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
frenzibyte: Urgh, will try to keep an eye out next time.

osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableSound.cs
peppy: Was there a reason for doing it this way?

frenzibyte: done it this way as just quick ensuring logical assertion not exactly part of the test scene but mostly to avoid fake pass because the sample hasn't actually changed, but this seems a bit horrible to read actually, .will change it to assert step when I have time to push the requested changes.

peppy: right, missed the line below.

bdach (approved):
seems fine

peppy (approved):

PR #10171, opened 2020-09-16
Fix global bindings being lost when running tests under headless contexts

peppy:
This was a shocking one to track down. Only affects rider certain methods of headless execution (rider's test panel for one).

Basically, `ResetDatabase` call was occurring after `OsuGameBase` initialises key bindings (via `KeyBindingStore.Register`, which would cause the `KeyBindingContainer` at the `OsuManualInputManagerTest` to have no bindings.

One could argue "wouldn't registering that one against the `KeyBindingStore` fix this?" but unfortunately, due to some level of EF caching, it decides it doesn't need to register the key bindings the second time around as it can still see the old ones (hits the `continue` part of the loop inside `insertDefaults`).

### DIFF
--- a/osu.Game/Tests/Visual/OsuTestScene.cs
+++ b/osu.Game/Tests/Visual/OsuTestScene.cs
@@ -69,12 +69,26 @@ namespace osu.Game.Tests.Visual
         /// </summary>
         protected virtual bool UseOnlineAPI => false;
 
+        /// <summary>
+        /// When running headless, there is an opportunity to use the host storage rather than creating a second isolated one.
+        /// This is because the host is recycled per TestScene execution in headless at an nunit level.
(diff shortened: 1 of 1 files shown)


bdach:
I *think* this is fine but can't seem to manage to get a setup in that would replicate this behaviour on `master` to be able to double-check that it fixes it...

peppy:
The replication method is to run TestChangeSelectionViaKeyboardBeforeEnter in TestPlaySongSelect, by self, from rider

bdach:
Aha, yep, there it is. Can confirm that the fail changes to a pass.

bdach (approved):

smoogipoo (approved):
I think this is fine

PR #10173, opened 2020-09-16
Fix profile section tab control not absorbing input from behind

Joehuu:


bdach:
I guess for 100% percent correctness the active tab's `ExpandingBar` would have to do the same thing, albeit given the gap on that one is literal pixels wide I don't necessarily think it's worth it.

bdach (approved):

smoogipoo (approved):

PR #10177, opened 2020-09-17
Fix import tests not waiting long enough before performing migration

peppy:
Resolves https://ci.appveyor.com/project/peppy/osu/builds/35249434.

Also removes local `nameof()` specifications and uses the clean run game host.

### DIFF
--- a/osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
+++ b/osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
@@ -4,6 +4,7 @@
 using System;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using NUnit.Framework;
@@ -22,7 +23,7 @@ namespace osu.Game.Tests.NonVisual
         [Test]
         public void TestDefaultDirectory()
         {
-            using (HeadlessGameHost host = new CustomTestHeadlessGameHost(nameof(TestDefaultDirectory)))
+            using (var host = new CustomTestHeadlessGameHost())
             {
                 try
                 {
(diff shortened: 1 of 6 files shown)


osu.Game.Tests/NonVisual/CustomDataDirectoryTest.cs
```diff
@@ -272,8 +273,16 @@ private OsuGameBase loadOsu(GameHost host)
         {
             var osu = new OsuGameBase();
             Task.Run(() => host.Run(osu));
+
             waitForOrAssert(() => osu.IsLoaded, @"osu! failed to start in a reasonable amount of time");
```
bdach, line 277: I'm super confused as to why this is untouched, as here's where the original test failure is - was this meant to be removed in favour of the other `waitForOrAssert` added here? I can still get what seems to be the exact same failure if I stress these tests hard enough (`[Repeat(1000)]`) with this current diff.

If I remove this assert I can still get the test to sometimes die, but on something else (a nullref during game disposal [here](https://github.com/ppy/osu/blob/0dceef8cd73d9ea236a57fb10b767567647105d9/osu.Game/OsuGameBase.cs#L365), from a finaliser)

peppy: what does it fail on with this assert still in place? it's basically a pre-load step which shouldn't cause any failure due to the second check added.

this was left in place because without it, the `UpdateThread` could potentially not be ready yet (surprised this didn't immediately show for you? it was basically 100% reproduction here)

bdach: I haven't checked what this fails on, but mostly I'm not seeing how adding a second assert *after* this one is supposed to resolve the issue, if this is the check that's failing already (failure CI run linked in PR description is pointing to this exact line number on `master`).

peppy: - The first check is there to ensure that UpdateThread is non-null, so the second check can run
- The second check is waiting two frames so `LoadComplete` has been run (and KeyBindingContainer's `LoadComplete` has been run). From an async context, checking `IsLoaded` alone is not enough as it is set *before* the `LoadComplete` code is run.

smoogipoo (dismissed):

bdach (dismissed):
All explained in discord :)

smoogipoo (approved):

PR #10178, opened 2020-09-17
Fix potential endless taiko beatmap conversion

smoogipoo:
Visible on https://osu.ppy.sh/beatmapsets/321566#osu/1270556

This eventually also needs a cancellation token too, but it should suffice for now - I don't believe `tickSpacing` can get small and pass this check without the duration also being small since they're both relative to beat length.

### DIFF
--- a/osu.Game.Rulesets.Taiko/Beatmaps/TaikoBeatmapConverter.cs
+++ b/osu.Game.Rulesets.Taiko/Beatmaps/TaikoBeatmapConverter.cs
@@ -8,6 +8,7 @@ using osu.Game.Rulesets.Taiko.Objects;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using osu.Framework.Utils;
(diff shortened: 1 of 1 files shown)


peppy:
Confident this won't break any normal beatmaps?

peppy (dismissed):

smoogipoo:
Hmmm, I'll run this on my set of ranked beatmaps. Should take ~30min. May need to allow at least one hit.

smoogipoo:
There's no differences over my set of 90k ranked/loved/approved beatmaps, so I'm fairly certain this won't break normal beatmaps.

smoogipoo:
Hold off, I got another OOM on my unranked set.

smoogipoo:
Not sure what happened, I think I might've recomputed with stable when I ran into the OOM again.

image

Seems to complete on my set of unranked beatmaps too, so I'll unblock this PR.

PR #10180, opened 2020-09-17
Make beatmap conversion support cancellation tokens

smoogipoo:
This doesn't use the cancellation token just yet, it'll need additional implementation for mania/taiko in further PRs.

### DIFF
--- a/osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapConverter.cs
+++ b/osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapConverter.cs
@@ -5,6 +5,7 @@ using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Catch.Objects;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Objects;
 using osu.Framework.Extensions.IEnumerableExtensions;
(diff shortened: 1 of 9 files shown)


peppy:
Do you think it's worth adding obsoleted methods to not break compatibility?

smoogipoo:
As I mentioned in the description - I can’t seem to be able to do that without eventually causing breaking further down the track. Maybe it’d work by making the _new_  ConverHitObject method virtual instead of abstract (I thought there was an issue last time I tried), but then someone using the ruleset api wouldn’t really know how to convert hitobjects.

peppy:
someone using or someone implementing the api?

smoogipoo:
As in, the only way obsoletion can work is via:
```csharp
/// <summary>
/// Performs the conversion of a hit object.
/// This method is generally executed sequentially for all objects in a beatmap.
/// </summary>
/// <param name="original">The hit object to convert.</param>
/// <param name="beatmap">The un-converted Beatmap.</param>
/// <param name="cancellationToken">The cancellation token.</param>
/// <returns>The converted hit object.</returns>
protected virtual IEnumerable<T> ConvertHitObject(HitObject original, IBeatmap beatmap, CancellationToken cancellationToken)
{
#pragma warning disable 618
    return ConvertHitObject(original, beatmap);
#pragma warning restore 618
}

/// <summary>
/// Performs the conversion of a hit object.
/// This method is generally executed sequentially for all objects in a beatmap.
/// </summary>
/// <param name="original">The hit object to convert.</param>
/// <param name="beatmap">The un-converted Beatmap.</param>
/// <returns>The converted hit object.</returns>
[Obsolete("Use the cancellation-supporting override")]
protected virtual IEnumerable<T> ConvertHitObject(HitObject original, IBeatmap beatmap) => Enumerable.Empty<T>();
```
Which removes `abstract` from `ConvertHitObject`. Which hides that method from anyone wanting to implement the beatmap converter, which doesn't sit right with me.

smoogipoo:
I guess it might be fine since people should be using the templates anyway?

peppy:
Yep, looks good!

peppy (approved):

PR #10185, opened 2020-09-18
Fix hovered tab items not showing hover state when deselected

Joehuu:
Before:
OPOyn8gmzc
rS78Q53XUh

After:
LMvXrNGLRv
Ygvh43yzbz


Joehuu:
This also happens with other tab items (e.g. `OsuTabItem`), only ever seen if they can be controlled with keyboard. Will broaden the scope of this.

smoogipoo (approved):

PR #10189, opened 2020-09-18
Make Autoplay more lazy on fast reverse sliders

PercyDan54:
osu_2020-09-18_20-47-39

### DIFF
--- a/osu.Game.Rulesets.Osu/Replays/OsuAutoGenerator.cs
+++ b/osu.Game.Rulesets.Osu/Replays/OsuAutoGenerator.cs
@@ -1,19 +1,20 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using osuTK;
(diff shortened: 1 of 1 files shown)


bdach:
Please don't make commits using the web interface.

Have you tested this code at all? Are you able to provide an explanation as to what this change is supposed to do?

PercyDan54:
> 
> 
> Please don't make commits using the web interface.
> 
> Have you tested this code at all? Are you able to provide an explanation as to what this change is supposed to do?

Sorry this is my first pr, feel free to point out any other mistakes.
I've tested it with several beatmaps with fast reverse sliders.
As known, fast reverse sliders can be simply completed with cursor hovering above it instead of always following like Auto did.This attempts to change that behaviour

bdach:
For what it's worth I don't necessarily think that this needed fixing. Special casing this feels off to me.

PercyDan54:
> 
> 
> For what it's worth I don't necessarily think that this needed fixing. Special casing this feels off to me.

This is not a bug.
Just wanted to make auto more smooth.
Ignore this if you want.

PR #10196, opened 2020-09-20
Add support for legacy miss & strong hit explosions

### LINKED ISSUE #8891: osu!taiko miss explosions don't display
**Describe the bug:** for some reason, one of my skin (taiko only) doesnt display its miss hitburst but displays other hitbursts. other skin (includes lazer default and classic) doesnt have the problem

https://drive.google.com/open?id=10adPg55b-SvEFRuoxAW2z_V_O6wnnXdp

**osu!lazer version:** master with commit b83a41e

**Logs:**
not needed?

**Skins:**
miss displays
[Kitkat - Copy.zip](https://github.com/ppy/osu/files/4547367/Kitkat.-.Copy.zip)

miss not displays
[taiko.zip](https://github.com/ppy/osu/files/4547345/taiko.zip)

### LINKED ISSUE #9860: taiko big dons/kats dont use taiko-300k
**Describe the bug:** taiko big dons/kats doesnt seem to use taiko-300k.png and instead will use taiko-300.png. in the screenshot, they both say great but taiko-300 is just a solid colored one while taiko-300k has a semi-transparent letters while having outer glow (edit: idk if it also affects taiko-100k)

Lazer: (edit for wrong screenshot)
Screenshot (148)

Stable:
screenshot096

**Skin:** https://drive.google.com/file/d/1is63tvBoQQnVybxPTla2SNvxqDY2rvZc/view?usp=sharing

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5078297/runtime.log)
[performance.log](https://github.com/ppy/osu/files/5078298/performance.log)

bdach:
- Resolves #8891.
- Resolves #9860.

# Summary

This PR brings back support for displaying miss explosions and strong hit explosions in taiko.

Miss explosions were already theoretically supported, but not actually created and added to the playfield due to a check in `addExplosion()` excluding them from being created.

As for strong hit explosions, as I outlined in the issue thread I'm pretty sure stable only shows them if both hits are frame-perfect (as in they occur on the exact same frame), and this implementation follows the same logic for now.

Here are some videos showing the effects of this change:

| | | |
| :- | :-: | :-: |
| current `master` | [test scene](https://drive.google.com/file/d/1k0YvsZPmBYe6PC-y38Hp1CYTa0Ta229T/view?usp=sharing) | [in-game autoplay](https://drive.google.com/file/d/1WPGqH6hYs1StPt2QFUgOYytw0IncEiSN/view?usp=sharing) |
| this PR | [test scene](https://drive.google.com/file/d/1T8tcDhjaD-y9Ylptvn7wkTWoBmfQkcG3/view?usp=sharing) | [in-game autoplay](https://drive.google.com/file/d/1rwqEqHurWuaWp76KO39-LN-81TcoFItu/view?usp=sharing) |

# Remarks

Not sure about replicating the stable behaviour of strong explosions. It's easy to implement but borderline useless for anything but autoplay. I'll change to something different on request.

Easily the worst part of this diff is the schedule in `addExplosion()`. As the inline comment attempts to explain, it is done to handle both hits potentially firing in the same frame. To explain better, assume that a `LeftCentre` and a `RightCentre` are hit on the same frame. This is how the input is processed:

1. `LeftCentre` press is dispatched to `DrawableHit.DrawableStrongNestedHit`. As the main object is not yet judged, the handler early-returns and does nothing.
2. The main `DrawableHit` is next in the input queue; the hit is processed normally as if it was a single. This is where the explosion would get created and added to the hierarchy without the schedule.
3. `RightCentre` press is dispatched to `DrawableHit.DrawableStrongNestedHit`. The main object is judged, so the double hit result is applied only at this point.
4. `RightCentre` press is then dispatched to the main `DrawableHit`, which early-returns as it's already processed an action in this frame.

The purpose of the schedule is to delay the explosion creation, so that point 3. can execute before the pertinent values of the nested object are read to determine what component should be displayed. I felt this was the simplest way of resolving this, but I'd definitely be happy to see it go - that might require some changes to that action handling logic, however, and definitely warrant a separate PR.

I'm not even entirely sure how that would have to be handled - some kind of framework API to allow `KeyBindingHandler`s to receive events with higher priority, or adding that override locally game-side as part of the input queue construction?

### DIFF
--- a/osu.Game.Rulesets.Taiko.Tests/DrawableTestStrongHit.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/DrawableTestStrongHit.cs
@@ -1,0 +1,44 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Rulesets.Scoring;
+using osu.Game.Rulesets.Taiko.Objects;
+using osu.Game.Rulesets.Taiko.Objects.Drawables;
+
+namespace osu.Game.Rulesets.Taiko.Tests
+{
+    public class DrawableTestStrongHit : DrawableHit
+    {
+        private readonly HitResult type;
+        private readonly bool hitBoth;
+
+        public DrawableTestStrongHit(double startTime, HitResult type = HitResult.Great, bool hitBoth = true)
+            : base(new Hit
+            {
+                IsStrong = true,
+                StartTime = startTime,
+            })
+        {
+            // in order to create nested strong hit
+            HitObject.ApplyDefaults(new ControlPointInfo(), new BeatmapDifficulty());
+
+            this.type = type;
+            this.hitBoth = hitBoth;
+        }
+
+        protected override void LoadAsyncComplete()
+        {
+            base.LoadAsyncComplete();
+
+            Result.Type = type;
+
+            var nestedStrongHit = (DrawableStrongNestedHit)NestedHitObjects.Single();
+            nestedStrongHit.Result.Type = hitBoth ? type : HitResult.Miss;
+        }
+
+        public override bool OnPressed(TaikoAction action) => false;
+    }
+}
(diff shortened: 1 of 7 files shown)


peppy (approved):
I think the schedule is fine. Very isolated.

PR #10198, opened 2020-09-21
Add mention of ruleset templates to readme

peppy:
Apparently these weren't linked from anywhere visible. Probably deserves to be more prominent than developing osu! itself, as I think more people would be interested in that aspect when arriving at the lazer repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,13 @@ If you are looking to install or test osu! without setting up a development envi
 
 If your platform is not listed above, there is still a chance you can manually build it by following the instructions below.
 
-## Developing or debugging
+## Developing a custom ruleset
+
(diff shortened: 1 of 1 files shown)


bdach:
May be worth mentioning that ruleset API stability is not assured at this stage yet, although we try to preserve it where possible.

peppy:
probably worth mentioning that in the linked repository (and asking people to update it if they see an issue)?

i think the eventual plan is to combine that repo into this one..

bdach:
That's also fair, I suppose. Since the instructions how to actually use the templates are over in the other repository I guess it makes more sense for the warning to be there, too.

bdach (approved):

Joehuu (approved):
LGTM. *watched your stream when you were doing this (still watching as of now)*

PR #10200, opened 2020-09-21
Fix HitObjectContainer not correctly unbinding from startTime fast enough

peppy:
Because the bindables are inside a list, the automatic unbinding logic doesn't run immediately in disposal and the bindings remain bound until next GC, which is too late (causes exceptions in the `onStartTimeChanged` callback).

Found while investigating #10142.


bdach:
For the record, the reproduction scenario for the bug this fixes seems to be:

1. Edit any map, try to drag around any object on the timeline in the composer (dragging not explicitly required, just for demo purposes)
2. Change tab to any other (e.g. timing)
3. Go back to composer and try to drag any object on the timeline again, resulting in hard crash at the next nearest snap at:

```
Unhandled exception. System.InvalidOperationException: Removed a drawable (DrawableSlider) whose parent was not this (HitObjectContainer), but .
   at osu.Framework.Graphics.Containers.CompositeDrawable.RemoveInternal(Drawable drawable)
   at osu.Framework.Graphics.Containers.LifetimeManagementContainer.RemoveInternal(Drawable drawable)
   at osu.Game.Rulesets.UI.HitObjectContainer.onStartTimeChanged(DrawableHitObject hitObject) in /home/dachb/Dokumenty/opensource/osu/osu.Game/Rulesets/UI/HitObjectContainer.cs:line 59
```

bdach (approved):
tested as working :+1:

bdach:
I have no idea where this latest test failure could be coming from. Tried several time to stress the test to reproduce and got nothing...

smoogipoo (approved):
Though due to async disposal this could still go wrong, I suppose it's conceptually fine.

peppy:
Should we instead making `UnbindAllBindables` `public` or `internal protected`?

smoogipoo:
Just that isn't going to be enough in this case because it doesn't know about bindables nested inside lists, only those in fields(/properties). It would need support for arbitrary objects or need to be virtual, at which point things start to get a little sketchy.

If we decide to remove async disposal in the future and make `Dispose()` truly synchronous, then this code would be fine.

peppy:
I meant making it public or protected so it can be overridden here, with custom logic.

It's already virtual.

smoogipoo:
Sounds scary, but it may be fine.

peppy:
I guess the scary part is that `base` call may be forgotten? Should we make a separate method for this?

Honestly I really dislike the async disposal pattern in general and am thinking more and more that we should just remove it and fix the performance of disposal some other way...

smoogipoo:
Generally I agree, but I'd also need to see the performance to judge that. Even hierarchies that don't have poolable drawables can get quite large (e.g. song select), such that the only way may be to Fody the unbind process.

peppy:
Given that this may still regress, I'd rather fix it here. Can either go down the framework change path (hard) or add a local load state check inside the event (easy)?

smoogipoo:
What are you going to be checking via load state? Disposal doesn't change load state afaik.

peppy:
Right, that's an issue isn't it. Even checking `Parent` isn't enough...

Might just add an inline comment for now so we are aware of it?

smoogipoo:
Do you remember what part of `onStartTimeChanged` exceptions?

peppy:
The removal, see above https://github.com/ppy/osu/pull/10200#issuecomment-696338858

peppy:
Gonna merge this for now as it fixes a hard crash. Let's see how things play out going forward with async disposal and the likes.

PR #10201, opened 2020-09-21
Fix SPM changing incorrectly with playback rate changes

### LINKED ISSUE #10119: Replay SPM depends on playback rate
**Describe the bug:** 
(Probably duplicate of #9821 ? But this is with replays.)
When Autoplay:
If playback rate>1, SPM will be more than 477, up to 955 SPM with 2x playback rate
If playback rate<1, SPM will be less than 477, down to 238 SPM with 0.5x playback rate

You can use a random replay without Autoplay and it's also able to reproduce this
osu_2020-09-10_18-58-52
osu_2020-09-10_18-58-55

**osu!lazer version:** 
2020.903.0
**Logs:**
Not needed

peppy:
I'm not sure this is the direction we want to go, but I don't have any better ideas right now.

Closes https://github.com/ppy/osu/issues/10119.


osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
```diff
@@ -194,13 +193,7 @@ public void TestSpinUnaffectedByClockRate(double rate)
...
             addSeekStep(0);
 
-            AddStep("adjust track rate", () => MusicController.CurrentTrack.AddAdjustment(AdjustableProperty.Tempo, new BindableDouble(rate)));
-            // autoplay replay frames use track time;
-            // if a spin takes 1000ms in track time and we're playing with a 2x rate adjustment, the spin will take 500ms of *real* time.
-            // therefore we need to apply the rate adjustment to the replay itself to change from track time to real time,
-            // as real time is what we care about for spinners
-            // (so we're making the spin take 1000ms in real time *always*, regardless of the track clock's rate).
-            transformReplay(replay => applyRateAdjustment(replay, rate));
+            AddStep("adjust track rate", () => Player.GameplayClockContainer.UserPlaybackRate.Value = rate);
```
bdach, line 196: Could you explain where I went wrong with this before? With this change it feels like the test is testing something else entirely (I meant to simulate DT/HT without actually applying DT/HT as it's more involved).

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -191,7 +193,9 @@ private void setClock()
 
         private class StabilityGameplayClock : GameplayClock
         {
-            public IFrameBasedClock ParentGameplayClock;
+            public GameplayClock ParentGameplayClock;
+
+            public override IEnumerable<Bindable<double>> NonGameplayAdjustments => ParentGameplayClock.NonGameplayAdjustments;
```
frenzibyte, line 198: `ParentGameplayClock` could be null, if the parent clock wasn't a `GameplayClock`.

```suggestion
            public override IEnumerable<Bindable<double>> NonGameplayAdjustments => ParentGameplayClock?.NonGameplayAdjustments;
```

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
peppy: i find the comment here very hard to parse, will need to take another look at it.

bdach: To be fair now that I think about it, I'm pretty sure it's a hack to bypass the fact that autoplay doesn't generate spinner frames correctly when rate mods are applied. But all of this rate-affecting stuff is super confusing, so I might well be talking nonsense.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
peppy: How come this doesn't fail any tests?

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -200,11 +203,26 @@ public void StopUsingBeatmapClock()
...
         {
             if (!IsPaused.Value)
+            {
                 userOffsetClock.ProcessFrame();
+            }
 
             base.Update();
         }
 
+        private double getTrueGameplayRate()
```
smoogipoo, line 213: This is unused, right?

smoogipoo (changes requested):

osu.Game/Screens/Play/GameplayClockContainer.cs
peppy: ah yep, this looks left over from before i moved it.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
frenzibyte: Probably because there isn't any code path to this property in all existing tests, I still think adding a null conditional would be better to avoid a potential null reference.

osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerRotation.cs
smoogipoo: Has this been addressed to not be required?

peppy: I suggest you read through this and make your own judgement. As far as I can tell it was not doing what we want.

bdach: Yeah I think it is fine to go without. Spinner autoplay is broken right now with rate affecting mods and should be fixed properly.

peppy:
Probably need to investigate that test failure...

bdach:
I've been stressing the failing test for a while now but it seems to be failing exceedingly rarely... It only failed once on me and I foolishly didn't even capture any sort of data as to what was going wrong with the seek, as I tried to see if I could reproduce at all to begin with. Might be easier to reproduce on a cold start? Not sure.

I'll probably add some log calls and run it overnight, see if anything comes out of that.

peppy:
Looks stable enough to probably get this in as is and add the failure to our long list of random test failures?

bdach (approved):
I think so, yeah... Whatever it is that causes the sporadic failures I doubt this PR caused it.

smoogipoo (approved):

PR #10203, opened 2020-09-22
Fix fail animation breaking on post-fail judgements

peppy:
Before:

After:

Also fixes similar issues on the other three rulesets (tested each to look more correct than before). Can be tested using `TestSceneFailAnimation`.


bdach (approved):
Interesting that this went unnoticed as it spans across all rulesets. Confirmed that it's looking better visually :+1:

smoogipoo (approved):

peppy:
A future improvement may involve only applying the animation in the non-scrolling axis, so objects keep their velocity.

PR #10204, opened 2020-09-22
Add editor key bindings to switch between screens

peppy:
Adds expected bindings for switching between compose/design/timing/setup screens, matching osu-stable. Note that I changed the default bindings for mute to media-key-mute because it was clashing (F4). Not sure if we need a migration for this, probably not worth the effort.


enoslayd:
isnt the media-key mute clashes with the os volume?

peppy:
You have a point. Maybe not what we want? I don't have a mute key to check.

bdach:
As an idea, how about <kbd>Ctrl</kbd>+<kbd>1</kbd>..<kbd>4</kbd> for the editor tabs? Can't switch ruleset in editor anyways, right?

peppy:
Maybe. I'm just trying to match user expectations from stable (including my own 😅 )

enoslayd:
Maybe disable mute key (and maybe the mute itself) on editor? Who maps while having the sounds muted?

smoogipoo (dismissed):

smoogipoo:
My mute key is a secondary function under my print screen key (QWERTY JIS keyboard), so I'm not sure it's well defined either.

Maybe keep F4 and add a secondary bind for it?

peppy:
The reason for changing from F4 is to avoid conflict in default key bindings, so it definitely needs to be changed from that (or the editor need to be changed I guess).

Probably best not to merge as-is as it may be a regression for windows users where `InputKey.Mute` may toggle system level mute too.

smoogipoo:
Actually yeah good point, mute key toggles system level mute for me on linux too.

peppy:
Ctrl-F4 for now seem ok?

smoogipoo:
Sure, let's go with that.

smoogipoo (approved):

PR #10208, opened 2020-09-22
Reduce complexity of AllHitObjects enumerator when nested playfields are not present

peppy:
This enumerable should not exist. Probably should be replaced with a BindableList or something else that doesn't require this nested linq. The expensive part doesn't stop at this change, but there is casting and ordering occurring in the underlying call to `HitObjectContainer.Objects`.

Helps with non-mania editor compose screens a bit, but again, we really need to do some work to banish this from existence. Every use of it is bad (including removal / addition of new objects and other non-editor cases).

Before:

image

After:

image


bdach (approved):
Seems like a quick win and a good first step towards elimination as not to rip the whole thing apart immediately, I suppose :+1:

smoogipoo (approved):

PR #10224, opened 2020-09-24
Add more descriptive description and download button when statistics not available

peppy:
Before:

image

After:

### DIFF
--- a/osu.Game/Screens/Ranking/Statistics/StatisticsPanel.cs
+++ b/osu.Game/Screens/Ranking/Statistics/StatisticsPanel.cs
@@ -75,7 +75,21 @@ namespace osu.Game.Screens.Ranking.Statistics
                 return;
 
             if (newScore.HitEvents == null || newScore.HitEvents.Count == 0)
-                content.Add(new MessagePlaceholder("Score has no statistics :("));
+                content.Add(new FillFlowContainer
+                {
+                    RelativeSizeAxes = Axes.Both,
+                    Direction = FillDirection.Vertical,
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Ranking/Statistics/StatisticsPanel.cs
```diff
@@ -75,7 +75,21 @@ private void populateStatistics(ValueChangedEvent<ScoreInfo> score)
                 return;
 
             if (newScore.HitEvents == null || newScore.HitEvents.Count == 0)
-                content.Add(new MessagePlaceholder("Score has no statistics :("));
```
smoogipoo, line 78: Please wrap this in a `{ }`

```
if (...)
    x.Add(
    {
        ...
    }
else
{
...
}
```
Is almost bound to lead to errors.

smoogipoo (changes requested):

osu.Game/Screens/Ranking/Statistics/StatisticsPanel.cs
peppy: you mean just adding a pair of `{}` for the outer if, correct?

smoogipoo: yes

peppy: it's actually enforced by our style guidelines, overlooked.

smoogipoo (dismissed):

PR #10226, opened 2020-09-24
Fix score panel being incorrectly vertically aligned on screen resize

peppy:
On changing the screen aspect ratio while the extended results are visible, then returning to non-extended, the panel will be at an incorrect vertical location.

Fixed by using vertical centre at all times, and only adjusting the X positioning.

Before:

After:


Joehuu (approved):
LGTM, can't seem to break.

smoogipoo (approved):
I thought there was a reason for not centreleft anchoring the panels, but it seems to work.

PR #10231, opened 2020-09-24
Add editor object to object snapping

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10230 for testing purposes

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/Editor/TestSceneObjectObjectSnap.cs
+++ b/osu.Game.Rulesets.Osu.Tests/Editor/TestSceneObjectObjectSnap.cs
@@ -27,16 +27,20 @@ namespace osu.Game.Rulesets.Osu.Tests.Editor
             AddStep("get playfield", () => playfield = Editor.ChildrenOfType<OsuPlayfield>().First());
         }
 
-        [Test]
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Osu.Tests/Editor/TestSceneObjectObjectSnap.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            AddStep("move mouse to centre", () => InputManager.MoveMouseTo(playfield.ScreenSpaceDrawQuad.Centre));
+
+            if (!distanceSnapEnabled)
+            {
+                AddStep("disable distance snap", () =>
+                {
+                    InputManager.PressKey(Key.Q);
+                    InputManager.ReleaseKey(Key.Q);
+                });
```
bdach, line 42: We actually have shorthands for these already framework-side:

```suggestion
                AddStep("disable distance snap", () => InputManager.Key(Key.Q));
```

(not going to highlight the other places but there are some)

peppy: Was looking for this but couldn't find it.

smoogipoo (approved):

PR #10232, opened 2020-09-24
Avoid destroying editor screens when changing between modes

peppy:
I think we want it this way for as long as we can. Probably going to have to take care with cross-screen events and bindables firing unnecessarily (may need an enter/exit mode). Also really this should be using ScreenCarousel but...........................

- Fixes distance snap (and other state) being lost on changing screens
- Improves performance when switching between screens by infinite (after first load)

Consider the transition a best-effort-temporary-thing.

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -518,7 +518,7 @@ namespace osu.Game.Screens.Edit
                 .ScaleTo(0.98f, 200, Easing.OutQuint)
                 .FadeOut(200, Easing.OutQuint);
 
-            if ((currentScreen = screenContainer.FirstOrDefault(s => s.Type == e.NewValue)) != null)
+            if ((currentScreen = screenContainer.SingleOrDefault(s => s.Type == e.NewValue)) != null)
             {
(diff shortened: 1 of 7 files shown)


osu.Game/Screens/Edit/Editor.cs
```diff
@@ -512,7 +512,21 @@ private void resetTrack(bool seekToStart = false)
...
         private void onModeChanged(ValueChangedEvent<EditorScreenMode> e)
         {
-            currentScreen?.Exit();
+            var lastScreen = currentScreen;
+
+            lastScreen?
+                .ScaleTo(0.98f, 200, Easing.OutQuint)
+                .FadeOut(200, Easing.OutQuint);
+
+            if ((currentScreen = screenContainer.FirstOrDefault(s => s.Type == e.NewValue)) != null)
```
bdach, line 521: It feels like I'm beating a dead horse at this point, but I'd also use `SingleOrDefault` here. I realise it has performance implications but the container should have 4 children at absolute most and I'd really feel safer if this did check for duplicates, just in case.

bdach (dismissed):
Seems good :+1: I'll approve to not hold things up over trivialities but I do have one comment.

smoogipoo (approved):

PR #10234, opened 2020-09-24
Add audio track selection to editor setup screen

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10233

Drag & drop will come later.

### DIFF
--- a/osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
+++ b/osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
@@ -27,15 +27,15 @@ namespace osu.Game.Tests.Visual.Editing
             AddStep("set dummy", () => Beatmap.Value = new DummyWorkingBeatmap(Audio, null));
 
             base.SetUpSteps();
+
+            // if we save a beatmap with a hash collision, things fall over.
+            // probably needs a more solid resolution in the future but this will do for now.
+            AddStep("make new beatmap unique", () => EditorBeatmap.Metadata.Title = Guid.NewGuid().ToString());
         }
 
         [Test]
         public void TestCreateNewBeatmap()
         {
-            // if we save a beatmap with a hash collision, things fall over.
-            // probably needs a more solid resolution in the future but this will do for now.
-            AddStep("make new beatmap unique", () => EditorBeatmap.Metadata.Title = Guid.NewGuid().ToString());
-
             AddStep("save beatmap", () => Editor.Save());
             AddAssert("new beatmap persisted", () => EditorBeatmap.BeatmapInfo.ID > 0);
         }
--- a/osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
+++ b/osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
@@ -1,0 +1,25 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Game.Beatmaps;
+using osu.Game.Rulesets;
+using osu.Game.Rulesets.Osu;
+using osu.Game.Storyboards;
+
+namespace osu.Game.Tests.Visual.Editing
+{
+    public class TestSceneEditorBeatmapCreation : EditorTestScene
+    {
+        protected override Ruleset CreateEditorRuleset() => new OsuRuleset();
+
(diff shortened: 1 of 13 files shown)


osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
```diff
@@ -0,0 +1,66 @@
...
+using SharpCompress.Archives;
+using SharpCompress.Archives.Zip;
+
+namespace osu.Game.Tests.Visual.Editing
+{
+    public class TestSceneEditorBeatmapCreation : EditorTestScene
+    {
+        protected override Ruleset CreateEditorRuleset() => new OsuRuleset();
+
+        protected override bool EditorComponentsReady => Editor.ChildrenOfType<SetupScreen>().FirstOrDefault()?.IsLoaded == true;
```
bdach, line 23: as is tradition
```suggestion
        protected override bool EditorComponentsReady => Editor.ChildrenOfType<SetupScreen>().SingleOrDefault()?.IsLoaded == true;
```

osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            AddStep("set dummy", () => Beatmap.Value = new DummyWorkingBeatmap(Audio, null));
+
+            base.SetUpSteps();
+        }
+
+        [Test]
+        public void TestCreateNewBeatmap()
+        {
+            AddStep("add random hitobject", () => EditorBeatmap.Metadata.Title = Guid.NewGuid().ToString());
```
bdach, line 35: step name and action are two different things :wink: not sure which way you meant to go here

osu.Game/Overlays/MusicController.cs
```diff
@@ -81,6 +81,11 @@ private void load()
             mods.BindValueChanged(_ => ResetTrackAdjustments(), true);
         }
 
+        /// <summary>
+        /// Forcefully reload the current <see cref="WorkingBeatmap"/>'s track from disk.
+        /// </summary>
+        public void ForceReloadCurrentBeatmap() => changeTrack();
```
bdach, line 87: Would probably rename to `ForceReloadCurrentTrack`. Reloading beatmap in music controller sounds a little misleading.

osu.Game/Screens/Edit/Setup/SetupScreen.cs
```diff
@@ -120,4 +202,60 @@ private void onCommit(TextBox sender, bool newText)
...
+                Anchor = Anchor.Centre,
+                Origin = Anchor.Centre,
+                RelativeSizeAxes = Axes.X,
+                CornerRadius = CORNER_RADIUS,
+                OnFocused = DisplayFileChooser
+            };
+
+        public void DisplayFileChooser()
+        {
+            Target.Child = new FileSelector("/Users/Dean/.osu/Songs", new[] { ".mp3", ".ogg" })
```
bdach, line 238: not the best default for anyone else :stuck_out_tongue: Should probably be home dir if starting from scratch/editing existing, or latest song location if one has been chosen before in current editor session?

osu.Game/Screens/Edit/Setup/SetupScreen.cs
```diff
@@ -104,10 +140,56 @@ private void load(OsuColour colours)
...
+            {
+                FileInfo = osuFileInfo,
+                Filename = info.Name
+            });
+
+            Beatmap.Value.Metadata.AudioFile = info.Name;
+
+            music.ForceReloadCurrentBeatmap();
+
+            editor.UpdateClockSource();
```
bdach, line 183: `editor` is permitted to be null
```suggestion
            editor?.UpdateClockSource();
```

bdach (changes requested):

osu.Game.Tests/Visual/Editing/TestSceneEditorBeatmapCreation.cs
peppy: My line of thought was "this test is failing from saving multiple non-unique beatmaps" so I was going to add a hitobject. Then realised changing the metadata was much simpler. Oops.

osu.Game/Screens/Edit/Setup/SetupScreen.cs
peppy: Yeah, it is, but I didn't feel the need to fix this since it's likely never going to be called with a null editor. I guess there's no harm adding the check, though.

peppy:
I'm in the process of moving the file operations out into `WorkingBeatmap` / `BeatmapManager` for what it's worth.

The actually process won't change but I may push to this PR (or can just leave for a follow-up PR).

peppy:
Ended up making the changes on this branch since they weren't as large as I expected, should be good to go.

smoogipoo (approved):

PR #10235, opened 2020-09-24
Fix anchor of tournament ruleset selector dropdown

Joehuu:
Before:
cgwKyowRQ0

After:
2tKfaLTMdQ


bdach:
I'm not sure this is the correct way to go forward. It's applying a specific fix for one scenario for a general component.

To illustrate better why I think this is a problem, here's a comparison of what the component will look like if the left side is higher than the right and the other way around, with this current change:

2020-09-24-230804_661x157_scrot
2020-09-24-231005_654x165_scrot

Not sure what the design intent here but it feels like a regression. I think both sides should be either vertically top or vertically centre, but not half-and-half.

bdach (changes requested):
as above

Joehuu:
It is given that the dropdown needs to be vertical anchored top or the dropdown will try to center itself while the container autosizes. I can't change the vertical anchor only on that specific component or the layout will be inconsistent from others.

Right now, all usages of the left side are one lined and smaller in height than the right side. If I were to change the left side's vertical anchor, I would need to hack in some margin to give a center look at idle height (i.e. before the right side (e.g. dropdown) resizes) and that wouldn't work if the right side can have various heights.

peppy:
I think this is fine to go with for now, until we design a component which breaks the layout. Having the right (autosize) portion anchored to a central point is arguably going to break just as bad as the example @bdach provided so it's a lose-lose.

peppy (approved):

PR #10241, opened 2020-09-25
Rework ternary states to fix context menus not updating after already displayed

peppy:


smoogipoo:
What scenario is this covering for? Seems like a lot of added complexity for menu items that aren't on screen all that often?

peppy:
Open context menu, press "q".

With my upcoming changes this reduces complexity 10x.

smoogipoo:
Conceptually, it's weird that you're essentially pressing a menu item and it isn't closing the menu...

peppy:
Strongly disagree, you're pressing a keybind unrelated to the menu and it shouldn't change the current UI state. I would want to keep the menu open potentially to see the new state.

Unless we want to make menus block all keyboard input, which I'd argue is anti-user for the editor.

smoogipoo:
It isn't related to the menu... but it changes the menu's state? We're talking about new combo, right? Which _is_ part of the menu.

I.e. in the future I fully foresee "Q" being displayed as the keybind in the menu, at which point it makes it even more pointless to keep the menu active.
Go into your browser, right click, press inspect element. Does it keep the menu open? What's the difference with that interaction from this?

And yeah, that's typically how browsers work - they block all but page-wide keyboard input with the menu open.

peppy:
The key bind is for the button on the left. See the PR where I add that.

I'm not interested in matching browsers here *unless* we add menu interaction via keyboard. Which is for another day/year

smoogipoo (approved):
you've lost me, but ok

PR #10247, opened 2020-09-25
Forcefully re-apply DrawableHitObject state transforms on post-load DefaultsApplied

peppy:
Fixes sliders not correctly adjusting their fade when extending their path length, amongst other similar issues.


smoogipoo (approved):
Seems to be fine

PR #10251, opened 2020-09-25
Add additional bonus HitResult types

smoogipoo:
This is a non-breaking change:
1. Adds `SmallBonus` and `LargeBonus` to `HitResult`.
    * There have been no changes to hitobjects yet, this is tbd in a separate PR.
2. Adds ordering to the enum.
3. Changes displays game-wide to use the new results + ordering.
    * Trims out a few unnecessary (for now) results - bonus judgements, tick misses.
    * Provides a bit more context as to how to display the value. For ticks the display should be x/y and for bonus it should probably just be the raw score value (not done yet).
4. It defines the relationships between all `HitResult` types in terms of whether they are bonus, affect combo, represent hits, or are scorable. These are mostly unused for now, except for the new code in this PR, where it currently has no effect.
    * Going forward, we'll be removing the `AffectsCombo` and `IsBonus` properties of `Judgement`. This is tbd in a separate PR.
    * I've tabulated the decisions for what does what here: https://docs.google.com/spreadsheets/d/1WYOfjB-xjn0DZI99d2-Pxpo5CYxPyW15yzYDz9gc8Ug/edit#gid=0

I've taken a very simple approach to `ScoreTable` (the beatmap overlay one) which is to find all hitresults across all scores displayed. Eventually we'll need a ruleset-level method to find all the applicable results for a ruleset, but this should suffice for now.
As I mentioned in Discord, maybe the way to do this is to expose all retrievable hitobjects for a ruleset and enumerate over them - undecided but that's why I haven't done it here.

(Drafting while further PRs are published, don't review yet in-case I need to move things around more)

### DIFF
--- a/osu.Game/Rulesets/Scoring/HitResult.cs
+++ b/osu.Game/Rulesets/Scoring/HitResult.cs
@@ -87,15 +87,21 @@ namespace osu.Game.Rulesets.Scoring
         SmallBonus = 254,
 
         /// <summary>
-        /// Indicate a large bonus.
+        /// Indicates a large bonus.
         /// </summary>
         [Description("L Bonus")]
         [Order(8)]
         LargeBonus = 320,
 
+        /// <summary>
+        /// Indicates a miss that should be ignored for scoring purposes.
+        /// </summary>
         [Order(13)]
         IgnoreMiss = 384,
 
+        /// <summary>
+        /// Indicates a hit that should be ignored for scoring purposes.
+        /// </summary>
         [Order(12)]
         IgnoreHit,
     }
--- a/osu.Game/Rulesets/Scoring/HitResult.cs
+++ b/osu.Game/Rulesets/Scoring/HitResult.cs
@@ -14,7 +14,7 @@ namespace osu.Game.Rulesets.Scoring
         /// </summary>
         [Description(@"")]
         [Order(14)]
-        None = 0,
+        None,
 
         /// <summary>
         /// Indicates that the object has been judged as a miss.
@@ -25,7 +25,7 @@ namespace osu.Game.Rulesets.Scoring
         /// </remarks>
         [Description(@"Miss")]
         [Order(5)]
-        Miss = 64,
+        Miss,
 
         [Description(@"Meh")]
         [Order(4)]
@@ -57,7 +57,7 @@ namespace osu.Game.Rulesets.Scoring
         /// Indicates small tick miss.
         /// </summary>
         [Order(11)]
-        SmallTickMiss = 128,
+        SmallTickMiss,
 
         /// <summary>
         /// Indicates a small tick hit.
@@ -70,7 +70,7 @@ namespace osu.Game.Rulesets.Scoring
         /// Indicates a large tick miss.
         /// </summary>
         [Order(10)]
-        LargeTickMiss = 192,
+        LargeTickMiss,
 
         /// <summary>
         /// Indicates a large tick hit.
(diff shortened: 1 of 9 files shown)


bdach (commented):
looks good overall :+1: just one observation to make

osu.Game/Rulesets/Scoring/HitResult.cs
```diff
@@ -21,47 +24,156 @@ public enum HitResult
...
+        /// </summary>
+        [Description("L Bonus")]
+        [Order(8)]
+        LargeBonus = 320,
+
+        /// <summary>
+        /// Indicates a miss that should be ignored for scoring purposes.
+        /// </summary>
+        [Order(13)]
+        IgnoreMiss = 384,
```
peppy, line 100: I'm a bit confused as to where these numbers come from.. can't find a comment anywhere explaining them which is something I think would be helpful to future visitors.

smoogipoo: They’re just padding in-case we add intermediate values. E.g. what if we add tiny tick as we once discussed? Or a tick that awards less than a large tick but still affects combo?

The reason is purely as safety insurance to prevent future breakage, as enums are stored as consts at IL-level.

How so you want me to comment this? An in-line non-xmldoc?

smoogipoo: Or, should I remove it? As discussed earlier I think we can add nested objects to achieve the same effect...

peppy: Do we refer to them using the actual numbers anywhere in storage? Does it get serialised as a number somewhere I'm missing? If not, I think removing them would be best for now.

smoogipoo: No, it’s strictly for how enums are stored in DLLs.

peppy: ah, i see..

if we're already using `[Order]` is there any reason to add padding? looking at them i can't really see a *hugely* logical order anyways. may be best to just add new ones at the end for safety... or try and get a better ruleset API versioning setup by then (we probably want this eventually, requiring rulesets to update ahead of the next release, which would avoid having to make hacks like this).

smoogipoo: Sure I can work with ruleset versioning, will remove.

osu.Game/Scoring/ScoreInfo.cs
```diff
@@ -186,6 +185,48 @@ public string StatisticsJson
...
+                if (key.IsBonus())
+                    continue;
+
+                int value = Statistics.GetOrDefault(key);
+
+                switch (key)
+                {
+                    case HitResult.SmallTickHit:
+                    {
+                        int total = value + Statistics.GetOrDefault(HitResult.SmallTickMiss);
```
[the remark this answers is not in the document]
smoogipoo: Yeeeeeah... I think this is best case and nothing more that can be done here... For lazer scores (timeshift) this'll work corectly since miss/hit directly correspond.

Good catch though, noted.

bdach:
Even though CI failed three times on this branch, one failure (the spinner one) is a known one with an unknown cause, the `CompletionCancellation` one I think I've seen before and can repro on `master`, and the third one I can't repro at all, so gonna throw an approve anyway... ¯\_(ツ)_/¯

bdach (approved):

peppy (approved):

peppy:
@smoogipoo would be nice to convert that gdoc to a wiki page on this repo at some point, to start paving a path forward to documenting the various behaviours for ruleset devs (and players).

PR #10257, opened 2020-09-26
Add support for SpinnerNoBlink to LegacyOldStyleSpinner

hbnrmx:


osu.Game.Rulesets.Osu/Skinning/LegacyOldStyleSpinner.cs
```diff
@@ -25,12 +26,16 @@ public class LegacyOldStyleSpinner : CompositeDrawable
         private Sprite metreSprite;
         private Container metre;
 
+        private bool spinnerNoBlink;
```
smoogipoo, line 29: Given that this is always used in a `!spinnerNoBlink` manner, I'd like to see this variable inverted as `spinnerBlink`. You can do the inverse locally - don't need to change the name of the setting enum.

osu.Game/Skinning/LegacySkinConfiguration.cs
```diff
@@ -19,6 +19,7 @@ public enum LegacySetting
             ComboOverlap,
             AnimationFramerate,
             LayeredHitSounds,
+            SpinnerNoBlink
```
smoogipoo, line 22: This should be in `OsuSkinConfiguration`, not `LegacySetting`.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Skinning/LegacyOldStyleSpinner.cs
```diff
@@ -25,12 +25,16 @@ public class LegacyOldStyleSpinner : CompositeDrawable
...
 
+        private bool spinnerBlink;
+
         private const float sprite_scale = 1 / 1.6f;
         private const float final_metre_height = 692 * sprite_scale;
 
         [BackgroundDependencyLoader]
         private void load(ISkinSource source, DrawableHitObject drawableObject)
         {
+            spinnerBlink = !source.GetConfig<OsuSkinConfiguration, bool>(OsuSkinConfiguration.SpinnerNoBlink)?.Value ?? true;
```
bdach, line 36: Not sure if it's just a "me" problem, but this makes my brain hurt. I don't immediately know reading this if the default is `false` or `true` as I don't remember operator precedence by heart and I have to add braces to check (it does seem to be correct, however).

hbnrmx: Thanks! Added braces

smoogipoo (approved):

peppy (approved):

PR #10258, opened 2020-09-26
Add dynamic calculation of performance points to local scores

### LINKED ISSUE #8587: PP Counter In Completion Screen Not Working

Game4all:
Closes #8587
Supersedes #9121

Depends on #10338

# Summary

This adds the ability to score panels to calculate pp for a local score in a dynamic way (online score will always be using the precomputed pp field if it is populated).

The calculation returns the total pp awarded for now, might be cool to add a way to see pp awarded for each skill (this should be straightforward to implement since `PerformanceCalculator` already supports it.)

Please note that I haven't added tests to this (I'm unsure about the way to go to add them). Please let me know if you require them anyways.

DvuuHwVQJ9


osu.Game/Screens/Ranking/Expanded/Statistics/PerformanceStatistic.cs
```diff
@@ -0,0 +1,76 @@
...
+            protected override double RollingDuration => AccuracyCircle.ACCURACY_TRANSFORM_DURATION;
+
+            protected override Easing RollingEasing => AccuracyCircle.ACCURACY_TRANSFORM_EASING;
+
+            protected override OsuSpriteText CreateSpriteText() => base.CreateSpriteText().With(s =>
+            {
+                s.Font = OsuFont.Torus.With(size: 20, fixedWidth: true);
+                s.Spacing = new Vector2(-2, 0);
+            });
+        }
```
bdach, line 68: I feel like this shouldn't be a carbon copy of what is already there in `CounterStatistic`.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -114,6 +115,27 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
             return computeDifficulty(key, beatmapInfo, rulesetInfo);
         }
 
+        /// <summary>
+        /// Calculates performance for the given <see cref="ScoreInfo"/> on a given <see cref="WorkingBeatmap"/>.
+        /// </summary>
+        /// <param name="beatmap">The <see cref="WorkingBeatmap"/> to do the calculation on. </param>
+        /// <param name="score">The score to do the calculation on. </param>
+        /// <param name="token">An optional <see cref="CancellationToken"/> to cancel the operation.</param>
+        /// <returns></returns>
```
bdach, line 124: missing content - either fill out (preferred) or remove

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -114,6 +115,27 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
...
+        public async Task<double> CalculatePerformance([NotNull] WorkingBeatmap beatmap, [NotNull] ScoreInfo score, CancellationToken token = default)
```
bdach, line 125: the lack of caching on this is concerning, especially seeing that this seems to be called three times on entering the results screen and once more on every time you expand another score and then go back to the original one

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -114,6 +115,27 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
...
+        {
+            return await Task.Factory.StartNew(() =>
+            {
+                if (token.IsCancellationRequested)
+                    return default;
+
+                var calculator = score.Ruleset.CreateInstance().CreatePerformanceCalculator(beatmap, score);
+                var total = calculator.Calculate(null);
```
bdach, line 133: `null` is already the default for this param, can be omitted

bdach (changes requested):

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
Game4all: Seeing that the expanded part of the first score panel (I meant by first the panel you land on when you access results) is constructed 3 times in a row, it may be a good idea as you said to cache calculation somewhere (and move the calculation there aswell), maybe in `ScorePanel` (since it doesn't get reconstructed each time you select another score) as adding a cache lookup to the `DifficultyManager` feels overkill to me

bdach: overkill why? is there a reason to ever recalculate a hashed score's PP after it's already been computed once?

by the way I missed this entirely yesterday but the fact that performance calculation is happening in `BeatmapDifficultyManager` seems suboptimal from a code organisation standpoint. the class is supposed to be doing things related to difficulty, so it should probably be split.

osu.Game/Screens/Ranking/Expanded/Statistics/CounterStatistic.cs
```diff
@@ -17,7 +17,7 @@ public class CounterStatistic : StatisticDisplay
     {
         private readonly int count;
 
-        private RollingCounter<int> counter;
+        protected RollingCounter<int> Counter;
```
smoogipoo, line 20:
```suggestion
        protected RollingCounter<int> Counter { get; private set; }
```

osu.Game/Screens/Ranking/Expanded/Statistics/PerformanceStatistic.cs
```diff
@@ -0,0 +1,53 @@
...
+
+        public override void Appear()
+        {
+            base.Appear();
+            Counter.Current.BindTo(performance);
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            cancellationTokenSource.Cancel();
```
smoogipoo, line 49:
```suggestion
            cancellationTokenSource?.Cancel();
```

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -114,6 +115,26 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
             return computeDifficulty(key, beatmapInfo, rulesetInfo);
         }
 
+        /// <summary>
+        /// Calculates performance for the given <see cref="ScoreInfo"/> on a given <see cref="WorkingBeatmap"/>.
+        /// </summary>
+        /// <param name="beatmap">The <see cref="WorkingBeatmap"/> to do the calculation on. </param>
+        /// <param name="score">The score to do the calculation on. </param>
+        /// <param name="token">An optional <see cref="CancellationToken"/> to cancel the operation.</param>
+        public async Task<double> CalculatePerformance([NotNull] WorkingBeatmap beatmap, [NotNull] ScoreInfo score, CancellationToken token = default)
```
smoogipoo, line 124: Async methods should be named as `MethodNameAsync`

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -114,6 +115,26 @@ public StarDifficulty GetDifficulty([NotNull] BeatmapInfo beatmapInfo, [CanBeNul
...
+        {
+            return await Task.Factory.StartNew(() =>
+            {
+                if (token.IsCancellationRequested)
+                    return default;
+
+                var calculator = score.Ruleset.CreateInstance().CreatePerformanceCalculator(beatmap, score);
```
smoogipoo, line 131: In addition to the above caching, this should use the result of `GetDifficultyAsync()` too otherwise it's recomputing difficulty every time, even though we're pretty much guaranteed to already have a cached version. Maybe `StarDifficulty` should contain the full difficulty attributes.

smoogipoo (changes requested):

Game4all:
As an sidenote, I'm looking to add an "advanced" display of the performance values in the form of a tooltip, displaying awarded pp for each skill as seen on the following screenshot (screenshot is actually wrong, as it contains difficulty attributes aswell).
osu_2020-09-30_20-08-58

bdach:
That looks mightily confusing in its current state. Why is the total not immediately below the other attributes, but with the difficulty attributes wedged in between, therefore making it seem like the values should sum up but actually don't?

This might be better suited for `SimpleStatistcTable`, where UR resides currently.

Game4all:
> 
> 
> That looks mightily confusing in its current state. Why is the total not immediately below the other attributes, but with the difficulty attributes wedged in between, therefore making it seem like the values should sum up but actually don't?
> 

Yeah I agree that in its current state this looks misleading, altough this should be very straight to fix

> This might be better suited for SimpleStatistcTable, where UR resides currently.

Actually I'm fine with that, but the "major" problem with this to me is that play statistics requires the beatmap to have been played to display stats (either by playing the map or watching a replay)

bdach:
> Actually I'm fine with that, but the "major" problem with this to me is that play statistics requires the beatmap to have been played to display stats (either by playing the map or watching a replay)

I don't think this is how it will always be, as far as I'm aware? Regardless that should not be a concern for now.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        public readonly struct PerformanceCacheLookup
+        {
+            public readonly double Accuracy;
+            public readonly int BeatmapId;
+            public readonly long TotalScore;
+            public readonly int Combo;
+            public readonly Mod[] Mods;
+            public readonly int RulesetId;
```
bdach, line 69: This cache lookup is weird. It doesn't use the actual unique-ish attributes of a score (local ID, online ID, user ID, score hash), but instead uses weird heuristics like accuracy or total score. I can guarantee this will cause a mis-hit *somewhere*.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -300,11 +301,13 @@ public BindableStarDifficulty(BeatmapInfo beatmap, CancellationToken cancellatio
         public readonly double Stars;
         public readonly int MaxCombo;
 
-        public StarDifficulty(double stars, int maxCombo)
+        public readonly DifficultyAttributes Attributes;
+
+        public StarDifficulty(double stars, int maxCombo, DifficultyAttributes attributes = null)
```
bdach, line 306: This nullability of `attributes` worries me. Why can't it be guaranteed to be present?

osu.Game/Rulesets/Difficulty/PerformanceCalculator.cs
```diff
@@ -21,14 +21,14 @@ public abstract class PerformanceCalculator
 
         protected double TimeRate { get; private set; } = 1;
 
-        protected PerformanceCalculator(Ruleset ruleset, WorkingBeatmap beatmap, ScoreInfo score)
+        protected PerformanceCalculator(Ruleset ruleset, WorkingBeatmap beatmap, ScoreInfo score, DifficultyAttributes attributes = null)
```
bdach, line 24: Ideally instead of having to jam the `DifficultyAttributes` through here, `PerformanceCalculator` should interface with `BeatmapDifficultyManager` directly, therefore directly using its cache. That direction might therefore lead to it becoming a `Component` or otherwise a part of the drawable hierarchy to leverage DI.

I'm not 100% on this one - @smoogipoo, how do you see this going forward?

bdach (changes requested):

osu.Game/Rulesets/Difficulty/PerformanceCalculator.cs
smoogipoo: I don't like that path, because it kinda removes the "headless" capability of this which is used in server-side components. Don't want to instantiate a whole `GameHost` to use this.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
smoogipoo: Also may be best to wait for score standardisation PRs before this, because they add MaxCombo to `DifficultyAttributes`, so the extra params here would only be duplication.

osu.Game/Scoring/ScorePerformanceManager.cs
smoogipoo: Furthermore, you can't use the mods directly. It needs to be acronym and sorted in some fashion - check how `BeatmapDifficultyManager` does its caching.

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -31,8 +31,8 @@ public class OsuPerformanceCalculator : PerformanceCalculator
         private int countMeh;
         private int countMiss;
 
-        public OsuPerformanceCalculator(Ruleset ruleset, WorkingBeatmap beatmap, ScoreInfo score)
-            : base(ruleset, beatmap, score)
+        public OsuPerformanceCalculator(Ruleset ruleset, WorkingBeatmap beatmap, ScoreInfo score, DifficultyAttributes attributes)
+            : base(ruleset, beatmap, score, attributes)
         {
             countHitCircles = Beatmap.HitObjects.Count(h => h is HitCircle);
```
smoogipoo, line 37: The `WorkingBeatmap` is only ever used for this in any of the performance calculators, right? I'm quite concerned that retrieving the playable beatmap will hit performance very hard, and this is done for every score that's opened on the results screen.

Unsure what the path forward is here yet, just putting this out there as something that needs investigation to make sure we can actually handle this sort of load. This is similar to the reason `BeatmapDifficultyManager` uses a task scheduler.

Game4all: Actually, do we want to keep a `WorkingBeatmap` or even a `IBeatmap` arround for performance calculation ? Out of the 4 rulesets, only std seems to use the `IBeatmap` in `PerformanceCalculator` only to get the number of sliders and hit circles. Can't these calculations be moved inside the difficulty attributes calculation?

 Wouldn't doing so make the calculation lighter performance-wise since we're assured to have difficulty attributes cached in `BeatmapDifficultyManager` and we don't have the burden of loading the working beatmap ? This way we could just construct the performance calculator with a null working beatmap and the cached difficulty attributes fetched from`BeatmapDifficultyManager`

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
Game4all: See this line as to why the difficulty attributes can't be guaranted to be present: https://github.com/Game4all/osu/blob/2766cf73b492ea751723b387399a40c4b6e20bc3/osu.Game/Beatmaps/BeatmapDifficultyManager.cs#L237

osu.Game/Scoring/ScorePerformanceManager.cs
Game4all: This means that in the case of calculating performance for two exact same scores (same accuracy, total score, combo, same mod combination) on same beatmap using the same ruleset, the performance will be computed twice? Goal I had with including all these values was to cache performance calculation for scores with identical stats.

bdach: The caching won't be *perfect* with the fields I suggested, yes. The only way to get perfect caching would be to create a cache key based on all inputs to the performance calculator, but given that diffcalc differs from ruleset to ruleset doing so is probably infeasible and definitely overengineering. The attributes you chose *sort of* achieve uniqueness if you're lucky, but not necessarily (imagine two scores with the same accuracy and total score, but e.g. with PERFECTs and OKs swapped - you could mistakenly use an inapplicable cache key and return a wrong value). Score unique attributes are way safer in that regard.

osu.Game/Rulesets/Difficulty/PerformanceCalculator.cs
bdach: Ah, right, forgot about that... Maybe some sort of DI-able wrapper over `PerfomanceCalculator`, then? Not sure, but I really dislike the current form of having to have the user of `PerformanceCalculator` also require using `BeatmapDifficultyManager` and manually passing the results of one to the other to get the cache gains.

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
bdach: If we do decide that performance calculation should be based entirely on difficulty attributes then `PerformanceCalculator` should have nothing to do with any `WorkingBeatmap` on its own. (Difficulty calculation is a different thing entirely.)

smoogipoo: Yeah, I agree with completely removing the beatmap from performance calculator, and heaving the count as part of the attributes.

osu.Game/Rulesets/Difficulty/PerformanceCalculator.cs
Game4all: not sure I follow the reasoning here. wouldn't that just add another layer of indirection with very little gain ? (chances are, it probably will only be used here)

bdach: For one, if the direction is to use difficulty attributes for pp calculation then `WorkingBeatmap` should be replaced entirely with non-nullable `DifficultyAttributes`, which I'd be more fine with than this current version. I'm not sure whether having a performance calculator necessitate always instantiating a difficulty attribute calculator to be able to *do anything* with it is good design but I suppose it's no worse than what we have currently.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,83 @@
...
+        private async Task<double> computePerformanceAsync(ScoreInfo score, PerformanceCacheLookup lookupKey, CancellationToken token = default)
+        {
+            var attributes = await difficultyManager.GetDifficultyAsync(score.Beatmap, score.Ruleset, score.Mods, token);
+
+            // Performance calculation requires the beatmap and ruleset to be locally available. If not, return a default value.
+            if (attributes.Attributes == null)
+                return default;
+
+            if (token.IsCancellationRequested)
+                return default;
```
bdach, line 51: These are super suspicious. This method should probably be returning `double?` at least to distinguish actual zeroes from "we haven't run at all due to missing attributes/cancellation". Also returning an actual value on cancellation isn't usually the .NET operating method, you usually throw `OperationCanceled`.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            var calculator = score.Ruleset.CreateInstance().CreatePerformanceCalculator(attributes.Attributes, score);
+            var total = calculator?.Calculate() ?? default;
```
bdach, line 54: Same here, this should return `null` for no performance rating if there is no perf calc implementation in the first place.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,83 @@
...
+
+            performanceCache[lookupKey] = total;
+
+            return total;
+        }
+
+        public readonly struct PerformanceCacheLookup
+        {
+            public readonly string ScoreHash;
+            public readonly int LocalId;
```
bdach, line 64: not necessarily relevant right this moment but at some point we might want to be serving difficulty attributes via API at which point these should probably include online score ID.

also, maybe rename `LocalId` to `LocalScoreID`, or `ScoreInfoID`, or similar? a little unclear right now.

osu.Game/Screens/Ranking/Expanded/Statistics/PerformanceStatistic.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            this.score = score;
+        }
+
+        [BackgroundDependencyLoader]
+        private void load(ScorePerformanceManager performanceManager)
+        {
+            if (score.PP.HasValue)
+            {
+                performance.Value = (int)score.PP.Value;
```
bdach, line 30: not sure about this cast - maybe it should be a round instead, to prevent truncation

bdach (changes requested):

osu.Game/Screens/Ranking/Expanded/Statistics/PerformanceStatistic.cs
Game4all: I'm not sure we're applying any rounding everywhere else in the codebase where we are manipulating pp values, so I would say that we should truncate for now for consistency with the rest of the codebase, no?

bdach: Actually most places that I can see do `ToString()` from decimal, which rounds.

bdach:
Not sure what these CI failures are about. Probably going to have to fire up windows to check as it seems specific to the android project...

bdach:
Build seems to be [passing on my fork](https://ci.appveyor.com/project/bdach/osu/builds/35711293), for what it's worth. Pretty spooky (which is appropriate for the month, I guess?)

bdach:
Gonna shoot off a merge commit in hopes this magically starts passing instead of reporting bogus warnings...

osu.Game/Screens/Ranking/Expanded/Statistics/CounterStatistic.cs
```diff
@@ -19,7 +19,7 @@ public class CounterStatistic : StatisticDisplay
         private readonly int count;
         private readonly int? maxCount;
 
-        private RollingCounter<int> counter;
+        protected RollingCounter<int> Counter { get; private set; }
```
peppy, line 22: this is weird abuse of this class. you're initialising a `CounterStatistic` with a readonly count and then updating the count at runtime. the count variable should probably not exist.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,78 @@
...
+            var attributes = await difficultyManager.GetDifficultyAsync(score.Beatmap, score.Ruleset, score.Mods, token);
+
+            // Performance calculation requires the beatmap and ruleset to be locally available. If not, return a default value.
+            if (attributes.Attributes == null)
+                return null;
+
+            token.ThrowIfCancellationRequested();
+
+            var calculator = score.Ruleset.CreateInstance().CreatePerformanceCalculator(attributes.Attributes, score);
+            var total = calculator?.Calculate();
```
peppy, line 48: i guess this is fine non-async for now because calculation is low overhead? @smoogipoo can you confirm that you're okay with this?

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        [Resolved]
+        private BeatmapDifficultyManager difficultyManager { get; set; }
+
+        /// <summary>
+        /// Calculates performance for the given <see cref="ScoreInfo"/>.
+        /// </summary>
+        /// <param name="score">The score to do the calculation on. </param>
+        /// <param name="token">An optional <see cref="CancellationToken"/> to cancel the operation.</param>
+        public async Task<double?> CalculatePerformanceAsync([NotNull] ScoreInfo score, CancellationToken token = default)
```
peppy, line 27:
```suggestion
        public Task<double?> CalculatePerformanceAsync([NotNull] ScoreInfo score, CancellationToken token = default)
```

You're adding a level of async-await state machine that's not required. Remove the `async` keyword and return the underlying task directly.

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            var lookupKey = new PerformanceCacheLookup(score);
+
+            if (performanceCache.TryGetValue(lookupKey, out double performance))
+                return performance;
```
peppy, line 32:
```suggestion
                return Task.FromResult(performance);
```

osu.Game/Scoring/ScorePerformanceManager.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            return await computePerformanceAsync(score, lookupKey, token);
```
peppy, line 34:
```suggestion
            return computePerformanceAsync(score, lookupKey, token);
```

peppy (changes requested):

osu.Game/Scoring/ScorePerformanceManager.cs
Game4all: Since we've moved the difficulty attributes calculation out of the performance calculator (which was incurring most of the overhead), I think this can go in as is.

peppy: That's the case for the time being, sure. Just want to get confirmation. I'm fine with it being how it is.

osu.Game/Screens/Ranking/Expanded/Statistics/CounterStatistic.cs
Game4all: How about making both count variables nullable and construct `PerformanceStatistic` with null counter values and hook into the counter directly? Unless making a carbon copy of `PerformanceStatistic` is fine, which I doubt ...

bdach:
@Game4all how does 44471b4 fix tests not building? c5cf0d0 built successfully and was supposed to also fix tests?

osu.Game/Screens/Ranking/Expanded/Statistics/CounterStatistic.cs
bdach: I think `PerformanceStatistic` should inherit from `StatisticDisplay` instead of `CounterStatistic`, but also `CounterStatistic.StatisticCounter` (lol) - the inner rolling counter class - should be extracted out to avoid duplication. That was what I was mostly worried about, the inner class and only it.

Game4all: It took me 3 reads to get it correctly , but yeah on a second thought that approach seems cleaner than patching a class to control directly the counter. Will give it a shot soon ..

peppy (approved):

peppy:
@smoogipoo can you confirm you're okay with this existing as it is? kind of a temporary solution but seems okay enough to get in for now.

smoogipoo:
Looks okay at a visual pass. I haven't been prioritising looking into this (it's just been hanging in my emails), so if it works then I'm fine with it.

PR #10259, opened 2020-09-26
Fix visual brokenness in taiko tests using DrawableTestHit

bdach:
# Summary

#10250 has left `TestSceneHits` and `TestSceneDrawableTaikoMascot` looking pretty broken visually, due to changing the inheritance hierarchy from

```
DrawableTaikoHitObject
  DrawableHit
    DrawableTestStrongHit
  DrawableTestHit
```

to the more correct

```
DrawableTaikoHitObject
  DrawableHit
    DrawableTestHit
      DrawableTestStrongHit
```

In tests that interact with taiko drawable hitobjects they need to be added to the visual hierarchy so that nested objects are created properly, which at the state of current `master` leads to stuff like this:

| `TestSceneHits` | `TestSceneDrawableTaikoMascot` |
| :-: | :-: |
| 2020-09-26-210550_1920x1138_scrot | 2020-09-26-210601_1920x1138_scrot |

which looks pretty bad.

In those two tests it is possible to add the object to the playfield rather than the immediate child of the test window, which fixes the brokenness. However, `TestSceneHitExplosion` would also display the object - and there's no playfield in sight there to add to, so instead I opted for a workaround local to `DrawableTestHit` to be able to apply zero alpha.

# Remarks

`TestSceneHitExplosion` was broken like so from the start. I figured I should fix it if I'm already fixing the others, though, especially seeing that I wrote it. It's the worst fix, by far, too (suppressing the base call), but there's not much room to maneuver there...

### DIFF
--- a/osu.Game.Rulesets.Taiko.Tests/DrawableTestHit.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/DrawableTestHit.cs
@@ -22,6 +22,12 @@ namespace osu.Game.Rulesets.Taiko.Tests
             HitObject.ApplyDefaults(new ControlPointInfo(), new BeatmapDifficulty());
         }
 
+        protected override void UpdateInitialTransforms()
+        {
+            // base implementation in DrawableHitObject forces alpha to 1.
+            // suppress locally to allow hiding the visuals wherever necessary.
+        }
+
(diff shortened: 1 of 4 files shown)


peppy (approved):
Seems amicable 👍

PR #10261, opened 2020-09-26
Fix drum rolls only spawning centre flying hits

bdach:
- Resolves a regression from #10217.
- [x] Depends on #10259 (first in chain).
- [x] Depends on #10260 (for the testing infrastructure).

Not sold on the resolution at all, so PRing more for discussion.

# Summary

The bidirectional sample-hit type mapping introduced in the aforementioned PR breaks flying hits, because they get no samples assigned from the tick that created them, therefore always ending up being centres.

Resolve by suppressing the flying hit's base call to `LoadSamples()`, which is a resolution that makes the most semantic sense for me (the flying hit is not supposed to play audio anyway, and suppressing the base call disables the sample -> hit type remapping that caused the type to be changed from `Rim` to `Centre`).

# Remarks

I know this is a bad resolution. I don't see good ones, not at least without majorly reworking how the bidirectional propagation of sample-to-hit-type works right now. The data is flying back and forth in multiple places:

- beatmap decoder is setting both hit type and samples
- `Hit` ctor copies hit type from hit object
- `LoadSamples` writes same value back (which is what failed here)
- any type change affects samples
- any sample change affects type

I'm PRing this mostly as it's an immediate, *relatively* non-invasive fix. A singular source of truth would go a long way here instead of trying to keep this state consistent at all times, but as I'm typing this I'm not yet sure how to make it happen given all of the things that are happening listed above.

### DIFF
--- a/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableFlyingHit.cs
+++ b/osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableFlyingHit.cs
@@ -27,5 +27,11 @@ namespace osu.Game.Rulesets.Taiko.Objects.Drawables
             base.LoadComplete();
             ApplyResult(r => r.Type = r.Judgement.MaxResult);
         }
+
+        protected override void LoadSamples()
+        {
+            // block base call - flying hits are not supposed to play samples
+            // the base call could overwrite the type of this hit
+        }
     }
 }
--- a/osu.Game.Rulesets.Taiko.Tests/TestSceneFlyingHits.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/TestSceneFlyingHits.cs
@@ -1,0 +1,48 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
+using osu.Framework.Testing;
+using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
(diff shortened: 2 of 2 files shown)


osu.Game.Rulesets.Taiko/Objects/Drawables/DrawableFlyingHit.cs
```diff
@@ -27,5 +27,11 @@ protected override void LoadComplete()
             base.LoadComplete();
             ApplyResult(r => r.Type = r.Judgement.MaxResult);
         }
+
+        protected override void LoadSamples()
+        {
+            // block base call - flying hits are not supposed to play samples
```
peppy, line 33: It looks like this is going to stop the base(base) call of `LoadSamples`, causing it to never load any samples. This may not be an issue for taiko since the samples are played by the drum, but not sure how safe it is...

bdach: Yep, it will. I also considered exposing `DrawableHit.updateSamplesFromTypeChange()` as protected and calling it manually in the flying hit ctor for it to write the proper value to prevent overwrite later, but that feels like a worse hack than this, and sort of dangerous seeing that on current `master` it's possible that a flying hit's `PlaySamples()` will get called (the above resolution doesn't cause double sample playback as it writes the samples to the hit object and not the *drawable* hit object, but it seems like it might do so at some point in the future).

peppy: I actually think this is fine due to the comment being present here. Somehow glossed over reading it.

These are a special kind of object anyways.

peppy (approved):
good for now

PR #10262, opened 2020-09-26
Display drawable judgements over hitobjects for osu! ruleset

### LINKED ISSUE #7584: Hit explosions appear underneath hitcircles
Classic skin hitcircleoverlay always appears over hit300g etc... after hit & fade out, May cause aesthetic issues with skins that use the two as a layered effect.

image

**osu!lazer version:** 2020.118.0

**Logs:** N/A

frenzibyte:
In a closer look:

- For the default skin in osu!stable, the judgements appear behind the objects as questioned in https://github.com/ppy/osu/issues/7584#issuecomment-695880367 (which is answered in https://github.com/ppy/osu/issues/7584#issuecomment-695881843 afterwards).
For simplicity, I made it appear over objects as well, as making it appear over then under after disappearing sounds really hard to do, but if it looks bad as-is, I can make it always appear under objects only for that legacy default skin via some skin setting or otherwise. (also curious to know why its behaviour is different from custom skins, is hardcoded inside stable?)

Closes #7584

### DIFF
--- a/osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
+++ b/osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
@@ -48,7 +48,7 @@ namespace osu.Game.Rulesets.Osu.UI
                     RelativeSizeAxes = Axes.Both,
                     Depth = 1,
                 },
-                // Todo: This should not exist, buit currently helps to reduce LOH allocations due to unbinding skin source events on judgement disposal
+                // Todo: This should not exist, but currently helps to reduce LOH allocations due to unbinding skin source events on judgement disposal
                 // Todo: Remove when hitobjects are properly pooled
                 new SkinProvidingContainer(null)
                 {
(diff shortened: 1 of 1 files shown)


peppy:
The default needs to match. It's important to its style.

bdach:
In addition to the above, how is [this issue comment](https://github.com/ppy/osu/issues/7584#issuecomment-697117549) addressed in this PR?

frenzibyte:
> I can't understand the text in your comment

*uhh..., I forgot to re-read what I just typed*

> The default needs to match stable. It's important to its style.

Fair point, can do I think, but I need to know the specifics of how stable does it for default skin:
  1. does it exactly show the judgement under the object?, or is there more to it? (from visual comparsions, appearing under hit object container looks identical to stable)
  2. why is only the default skin having its judgements appear under objects, is there a specific setting or a skin version determining whether judgements appear under objects?

> In addition to the above, how is this issue comment addressed in this PR?

Oh, I forgot to mention about this: It doesn't actually need the judgements to be per-hitobject, since the order of the objects goes from nearest to deepest, so making it part of the object would not help at all in resolving this, it would allow you to either 
 1. make the judgement appear over the current object (which this alternative easy solution achieves)
 2. make the judgement appear under the current object but above future objects (which doesn't sound like what is wanted)

and the implementation is kinda complicated, as I've went through it myself, involves making the object expire only after judgement finishes animation, and other stuff hard to grasp... (see https://github.com/ppy/osu/compare/master...b7c7f4037d31fcbefb126e97e0c519bee2472287)

PR #10273, opened 2020-09-28
Trigger a hitobject update after blueprint drag ends

peppy:
Was missing, causing no undo history (or general hitobject update) to occur after moving an object in space but not time.

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
@@ -201,6 +201,10 @@ namespace osu.Game.Screens.Edit.Compose.Components
 
             if (isDraggingBlueprint)
             {
+                // handle positional change etc.
+                foreach (var obj in selectedHitObjects)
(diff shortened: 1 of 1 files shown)


smoogipoo:
I don't really disagree with the change in general, but I'm fairly certain this does work. At least I can't replicate on master.

smoogipoo (approved):

peppy:
It works because changes are triggered from everything-else. Including "new combo" changing due to ternary states (fixed in https://github.com/ppy/osu/pull/10271) and other randomness.

smoogipoo:
So this is a preventative measure for a future change that breaks this? Because as I said I'm on latest master and can't replicate (that PR was merged already).

peppy:
I'm guessing there's still other things triggering an update. You can breakpoint inside `UpdateHitObject` to check this.

Without the merged PR, I can cause a failure if being careful. You need to make sure distance snap is turned off and no toggles are triggering an update. Also it only works on some object types due to reasons.

Definitely more such cases to fix, but I'm still figuring out the endgame for how we want to trigger updates (and where), so not really reading into it too much for now.

PR #10275, opened 2020-09-28
Allow attaching a replay to a FrameStabilityContainer when FrameStablePlayback is off

peppy:
Attached replays were completely ignored when `FrameStablePlayback` was disabled. This allows them to run regardless. As always, disabling frame stability will not guarantee good output, but in testing this is good enough to be used in the editor.


smoogipoo (approved):
Seems sane

PR #10277, opened 2020-09-28
Add autoplay support to editor

peppy:
- Fixes hitsounds not playing
- Fixes hitobjects not looking like they are getting hit
- [x] Depends on https://github.com/ppy/osu/pull/10274
- [x] Depends on https://github.com/ppy/osu/pull/10275

Known issues which will be fixed in future PRs:
- Rewinding to the exact location of a hitobject then playing forwards will cause it to not explode (likely an issue with replay handling).
- Seeking forward may cause some objects in the past to not be an a hit state. Not a huge deal and fixes itself after rewinding before the objects. Pretty obvious why this happens (frame stability disabled).

This adjusts fade out extension logic to act more like stable (leaving the approach circle behind). Probably not the final look we want, but it's what people are used to and works better with autoplay causing the hits to explode (approach circle is quite a stable drawable to work with in contrast).

hitcircles:

sliders:


smoogipoo (approved):
amazing

PR #10280, opened 2020-09-28
Fix extending spinners in editor causing them to disappear temporarily

### LINKED ISSUE #10279: Spinners disappear in editor when adjusting their length

peppy:
The editor triggers `ApplyDefaults` which causes `updateState` to be run on the `DrawableHitObject` itself, but this is a case where a child component binds to that value and don't receive the change because no `ValueChanged` is triggered.

Closes https://github.com/ppy/osu/issues/10279

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
@@ -3,7 +3,6 @@
 
 using System;
 using osu.Framework.Allocation;
-using osu.Framework.Bindables;
 using osu.Framework.Extensions.Color4Extensions;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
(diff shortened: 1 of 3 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
```diff
@@ -93,7 +92,7 @@ protected override void LoadComplete()
             base.LoadComplete();
 
             drawableSpinner.RotationTracker.Complete.BindValueChanged(complete => updateComplete(complete.NewValue, 200));
-            drawableSpinner.State.BindValueChanged(updateStateTransforms, true);
+            drawableSpinner.ApplyCustomUpdateState += updateStateTransforms;
```
smoogipoo, line 95: These all need to have unbinds on disposal.

smoogipoo (changes requested):

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
peppy: Aha, right. This could be an issue on skin change.

smoogipoo (approved):

PR #10282, opened 2020-09-28
Fix hit samples playing while paused / seeking in the editor

### LINKED ISSUE #10301: Slider sounds should not play while paused in the editor
The slider slider sample keeps playing, but even more annoyingly it seems like the tick samples play with every change to the slider path resulting in some gnarly buzzing sounds: https://drive.google.com/file/d/1kVAOdayQA70pQtHt7x59HvQDg29mWn1y/view?usp=sharing (maybe turn down volume to be safe).

peppy:
Fixes https://github.com/ppy/osu/issues/10301

This fixes only the main samples (via `PlaySamples`). Sliders / spinners with ongoing sounds can still get in a stuck state due to lack of frame stability – this will be fixed separately.

Not 100% on the interface's name..


peppy:
Hold off on this one briefly, making some changes to better handle looping sounds.

peppy:
It turns out we already had some pretty solid logic for blocking sample playback via the `IsPaused` bool so i've switched to making use of this. Also found a better name for the interface.

Note that this now also resolves the bug that is mentioned in the inline comments I removed (`SkinnableSound` now doesn't start playback when in a state it shouldn't, but saves this state for next unpause in the case of looping samples).

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -37,6 +37,8 @@ public class SkinnableSound : SkinReloadableDrawable, IAdjustableAudioComponent
         /// </remarks>
         protected bool PlayWhenZeroVolume => Looping;
 
+        protected virtual bool PlayWhenPaused => false;
```
bdach, line 40: I'm not sure I entirely agree with how this is done, given that `UnpausableSkinnableSound` has to exist. Thus we have a base class that allows sample playback control via an `ISamplePlaybackDisabler`, but then a derived class that opts out of that behaviour. The name is also pretty... oxymoronic?

Ideally I'd invert this and have the base `SkinnableSound` be the unpausable one, and the derived to DI in the disabler and do the necessary pause/unpause logic as the more specific class. Maybe that's a code quality PR for later, though... Might as well rename to something like `CanBePaused`/`Disabled` or something.

peppy: Hmm, that does indeed sound better. Wasn't 100% sure about this. I guess the only argument I have for doing it this way is there's only one usage of the derived case so I felt comfortable making it the more derived class..

@smoogipoo thoughts?

smoogipoo: That sounds like a good path forward.

peppy:
Really tidied up the `SkinnableSound` class 👍 .

smoogipoo (dismissed):

smoogipoo:
Needs a bit of conflict resolution

smoogipoo (approved):

PR #10285, opened 2020-09-29
Fix storyboard samples not pausing

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10282

Note that this still doesn't fix the fact that they won't resume, which requires more local state management (this is mentioned in `SkinnableSound`'s implementation).


smoogipoo:
Needs some conflict resolution

smoogipoo (approved):
Seems ok

PR #10286, opened 2020-09-29
Fix HitObject samples getting stuck in a playing state on seeking far into the future

peppy:
This is generally avoided due to `FrameStability` in gameplay, but as the editor removes the guarantee that we are running in a frame stable context this is required.

Note that this does not fix a similar issue faced by storyboard samples, which will need event passing from their nested `LifetimeManagementContainer`.

- [x] Depends on https://github.com/ppy/osu/pull/10285 for merge ease.

### DIFF
--- a/osu.Game/Skinning/PausableSkinnableSound.cs
+++ b/osu.Game/Skinning/PausableSkinnableSound.cs
@@ -41,8 +41,14 @@ namespace osu.Game.Skinning
                         // it's not easy to know if a sample has finished playing (to end).
                         // to keep things simple only resume playing looping samples.
                         else if (Looping)
+                        {
                             // schedule so we don't start playing a sample which is no longer alive.
-                            Schedule(base.Play);
+                            Schedule(() =>
+                            {
+                                if (RequestedPlaying)
+                                    base.Play();
+                            });
+                        }
                     }
                 });
             }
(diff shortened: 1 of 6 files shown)


osu.Game/Skinning/PausableSkinnableSound.cs
```diff
@@ -41,7 +41,8 @@ private void load(ISamplePlaybackDisabler samplePlaybackDisabler)
                         // it's not easy to know if a sample has finished playing (to end).
                         // to keep things simple only resume playing looping samples.
                         else if (Looping)
-                            base.Play();
+                            // schedule so we don't start playing a sample which is no longer alive.
+                            Schedule(base.Play);
```
smoogipoo, line 45: Do we want this instead of an `IsAlive` check? Wouldn't it also require scheduling the above `base.Stop()` too?

For example, seek into the future, trigger valuechanged a few times by starting/stopping play, then seek back to the object. Wouldn't that make the object play the samples because the schedules would be then be run?

smoogipoo (changes requested):

osu.Game/Skinning/PausableSkinnableSound.cs
peppy: The `IsAlive` thing is the reason this is as it is. As discussed yesterday `IsAlive` can't be used here (unless the `DrawableHitObject` is being passed in).

Will check on the schedule during rewinds, didn't notice an issue but I can understand why there is cause for concern.

peppy: I've added a re-check of `RequestedPlaying` inside the schedule, which is enough to make this safe (the play is only issued to `Looping` samples, which need to be manually stopped, so it's guaranteed their `RequestedPlaying` will be in the correct state at the point of the scheduled callback being run).

smoogipoo (approved):

PR #10287, opened 2020-09-29
Update timeline blueprint width using event flow rather than on update

peppy:
Unfortunately the placement object makes this a bit awkward. I still think this is an improvement though? Not sure about the `Contains` check...


peppy:
Feel free to close this if it doesn't look better. It's a bit of a weird one: without pooling this will likely cause a larger overhead (the `HitObjectUpdated` event is going to get bound to by *every* timeline blueprint)...

With pooling it really doesn't matter if we're doing (mostly noop) updates for all on-screen blueprints anyway.

peppy:
Let's just close this one, I don't think it is adding value and I'm really hesitant about the number of event bindings (we've seen that this can be an issue with the disposal flow).

Issue #10144, opened 2020-09-11
Standardised score is not 1-million-based


Contextualist:
**Describe the bug:**

 When I updated to this version (2020.910.0), I noticed that all of my local scores look less than what were displayed in the previous version. (e.g. ~990k scores now become ~780k). SS a map without any mod and bonus point does not yield a score of 1 million.

**Screenshots or videos showing encountered issue:** 

This screenshot shows one of my SS record (without any mod and bonus point). Note that I have `standardised` on in the settings.

![osu_2020-09-11_14-47-49](https://user-images.githubusercontent.com/16000021/92962290-bc265580-f43e-11ea-964d-b75747a8522a.jpg)

**osu!lazer version:** 

2020.910.0

**Logs:**

[logs.zip](https://github.com/ppy/osu/files/5211035/logs.zip)

peppy:
which ruleset?

PercyDan54:
> 
> 
> which ruleset? which beatmap?
Beatmap:
https://osu.ppy.sh/beatmapsets/75890#osu/214234
Ruleset osu!

peppy:
https://github.com/ppy/osu/pull/10141 is likely relevant.

Azzirah:
i think the dynamic scoring calculation takes into account the online/stable max combo of the beatmap and since if i'm not mistaken for the time being sliderends are not accounted in the combo of a map therefore an FC in the current version of lazer would have exactly the number of sliders less combo than an FC made on stable, which i believe is what's happening here.

Contextualist:
@Azzirah 
> i think the dynamic scoring calculation takes into account the online/stable max combo of the beatmap

Agree, the score here seems to reflect what you describe:
Current (Lazer) FC: 170
Sliderends: 77
170/(77+170)\*0.7 + 1\*0.3 = 0.781781

PR #10288, opened 2020-09-29
Standardise scoring across all rulesets

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10251

Bit of a chunky one, sorry. I'm splitting things out as much as I can, but I can't really make this one smaller.

In an effort to (r)esolve https://github.com/ppy/osu/issues/10144, we need to be able to compute score with more knowledge about how much numeric score each result provides. To do this, we need to:
1. Add bonus results.
2. Make all hit results of the same type give the same score.
3. Make all hit results of the same type affect the same portion of the score.

(2) requires making all scoring relative, except for bonus score.
(3) requires resolving the mess that is `AffectsCombo` and `IsBonus`, which have caused us problems in the past.

I've tabled up all the hit results and their properties here: https://docs.google.com/spreadsheets/d/1WYOfjB-xjn0DZI99d2-Pxpo5CYxPyW15yzYDz9gc8Ug/edit#gid=0

I've tested this with the community rulesets `sentakki`, `tau`, `Rush!`, `Swing` and `Yoso`. All play fine, so this is not a ruleset-api-breaking change.

One important difference of note is that this changes scoring for osu/taiko/catch, which now award 200 for GOODs instead of 100. I plan to address this in a separate PR. If I've changed awarded results from `Good` to `Ok` at any point in this PR, please point it out since that is an unintentional change.

(Drafting while further PRs are published)

### DIFF
--- a/osu.Game/Rulesets/Judgements/Judgement.cs
+++ b/osu.Game/Rulesets/Judgements/Judgement.cs
@@ -15,12 +15,12 @@ namespace osu.Game.Rulesets.Judgements
         /// <summary>
         /// The score awarded for a small bonus.
         /// </summary>
-        public const double SMALL_BONUS_SCORE = 10;
+        public const int SMALL_BONUS_SCORE = 10;
 
         /// <summary>
         /// The score awarded for a large bonus.
         /// </summary>
-        public const double LARGE_BONUS_SCORE = 50;
+        public const int LARGE_BONUS_SCORE = 50;
 
         /// <summary>
         /// The default health increase for a maximum judgement, as a proportion of total health.
@@ -74,7 +74,7 @@ namespace osu.Game.Rulesets.Judgements
         /// <summary>
         /// The numeric score representation for the maximum achievable result.
         /// </summary>
-        public double MaxNumericResult => ToNumericResult(MaxResult);
+        public int MaxNumericResult => ToNumericResult(MaxResult);
 
         /// <summary>
         /// The health increase for the maximum achievable result.
@@ -94,7 +94,7 @@ namespace osu.Game.Rulesets.Judgements
         /// </summary>
         /// <param name="result">The <see cref="JudgementResult"/> to find the numeric score representation for.</param>
         /// <returns>The numeric score representation of <paramref name="result"/>.</returns>
-        public double NumericResultFor(JudgementResult result) => ToNumericResult(result.Type);
+        public int NumericResultFor(JudgementResult result) => ToNumericResult(result.Type);
 
         /// <summary>
         /// Retrieves the numeric health increase of a <see cref="HitResult"/>.
@@ -155,7 +155,7 @@ namespace osu.Game.Rulesets.Judgements
 
         public override string ToString() => $"MaxResult:{MaxResult} MaxScore:{MaxNumericResult}";
 
-        public static double ToNumericResult(HitResult result)
+        public static int ToNumericResult(HitResult result)
         {
             switch (result)
             {
@@ -163,25 +163,25 @@ namespace osu.Game.Rulesets.Judgements
                     return 0;
 
                 case HitResult.SmallTickHit:
-                    return 1 / 30d;
+                    return 10;
 
                 case HitResult.LargeTickHit:
-                    return 1 / 10d;
+                    return 30;
 
                 case HitResult.Meh:
-                    return 1 / 6d;
+                    return 50;
 
                 case HitResult.Ok:
-                    return 1 / 3d;
+                    return 100;
 
                 case HitResult.Good:
-                    return 2 / 3d;
+                    return 200;
 
                 case HitResult.Great:
-                    return 1d;
+                    return 300;
 
                 case HitResult.Perfect:
-                    return 7 / 6d;
+                    return 350;
 
                 case HitResult.SmallBonus:
                     return SMALL_BONUS_SCORE;
--- a/osu.Game/Rulesets/Judgements/Judgement.cs
+++ b/osu.Game/Rulesets/Judgements/Judgement.cs
@@ -82,12 +82,12 @@ namespace osu.Game.Rulesets.Judgements
         public double MaxHealthIncrease => HealthIncreaseFor(MaxResult);
 
         /// <summary>
-        /// Retrieves the numeric score representation of a <see cref="JudgementResult"/>.
+        /// Retrieves the numeric score representation of a <see cref="HitResult"/>.
         /// </summary>
-        /// <param name="result">The <see cref="JudgementResult"/> to find the numeric score representation for.</param>
+        /// <param name="result">The <see cref="HitResult"/> to find the numeric score representation for.</param>
         /// <returns>The numeric score representation of <paramref name="result"/>.</returns>
-        [Obsolete("Has no effect. Use ToNumericResult(HitResult) (standardised across all rulesets).")] // Can be removed 20210328
-        protected virtual int NumericResultFor(HitResult result) => result == HitResult.Miss ? 0 : 1;
+        [Obsolete("Has no effect. Use ToNumericResult(HitResult) (standardised across all rulesets).")] // Can be made non-virtual 20210328
+        protected virtual int NumericResultFor(HitResult result) => ToNumericResult(result);
 
         /// <summary>
         /// Retrieves the numeric score representation of a <see cref="JudgementResult"/>.
--- a/osu.Game/Rulesets/Judgements/Judgement.cs
+++ b/osu.Game/Rulesets/Judgements/Judgement.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Scoring;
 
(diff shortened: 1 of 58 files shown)


LumpBloom7:
The standardization of the score values across rulesets will affect sentakki's Break modifier in a rather big way.

So, to sum up the BREAK modifier in its current state, any HitObjects with that modifier will be weighted much higher than normal Hitobjects, and the perfect judgement would also give more points than greats. (separated into [SentakkiJudgement](https://github.com/LumpBloom7/sentakki/blob/master/osu.Game.Rulesets.Sentakki/Judgements/SentakkiJudgement.cs) and [SentakkiBreakJudgement](https://github.com/LumpBloom7/sentakki/blob/master/osu.Game.Rulesets.Sentakki/Judgements/SentakkiBreakJudgement.cs)) 

With that system, normal notes would be weighted lower, and `Perfect`s would be equal to `Great`s. While BREAK notes will be weighted much higher, and `Perfect`s would actually be higher than `Great`s. This would be most similar to Sega's maimai. With this change, the weighting between Break notes and normal notes would be equal, and `Great` will always have a lower score value than `Perfect`.

Now the only way that BREAK notes can differentiate themselves is by enabling the `Perfect` judgement only for itself, while normal notes can only go up to `Great`. Having the max judgement be visibly different between HitObjects in the same ruleset is a big consistency issue for me. Showing a `Perfect` judgement even if they are valued equally to `Great`s at least acknowledges the players hit timing, even if that particular note type doesn't reward it...

peppy:
The goal of lazer isn't to emulate other games 1:1. The point of these coming changes are to create some sense of congruency across rulesets, with enough flexibility to offer various judgements without going too far as to having weird scoring mechanisms which require documentation/learning for users to understand.

Do you think it's feasible too come up with your own scoring mechanisms that fit better with the new changes, rather than trying to copy an existing game?

LumpBloom7:
I'm not against the standardization of the judgement score values, I think it's a good path forward, and I can live with that.

The exact score values used in sentakki are actually the least of my concerns, but rather the ability to give HitObjects different accuracy/scoring weights, which isn't really possible anymore.. Some HitObjects are just worth more than others, and there should be something in place to allow for that.

Perhaps something like this would be worth considering?
```cs
protected virtual float weighting => 1;
```

smoogipoo:
Maimai actually came up in our discussions before going down this path, but I'm not sure how to structure it in a way that everyone can agree with. Let's break this down a bit.

## The fundamental problem

Score recalc requires that **bonus score is absolutely defined**. It either needs to be defined as a single total value, or it needs to be defined as a count of constant values (e.g. what it is in this PR). Unlike the bonus portion, the base and combo portions can be immediately calculated using the acc% + max combo respectively.

Thus, for bonus score there are 3 possibilities:
1. Store as a total value.
    * Context is lost as to where the bonus came from.
    * Requires osu-web + database changes.
2. Store as separate counts per-type and _somehow_ determine, at the point of recomputation, the amount of score awarded for each type.
    * Full context is preserved.
    * Requires osu-web + database changes (explained later).
3. Standardise to a single scoring value and store as a count.
    * Partial context is preserved, for the case that there are only two bonus types.
    * Doesn't require osu-web changes.

These changes follow option 3 as I tried to avoid breaking changes.

I'll remind that standardisation of the normal portions is not strictly required, however if standardisation is to be broken via such weighting then, along with the same for the bonus results, the following question needs to be answered:

## How should the result screen display judgements?

#### Option 1: Combined
```
perfect:x good:y miss:z
``` 
#### Option 2: Fully-separated
```
  hits | perfect:x good:y miss:z
breaks | perfect:x good:y miss:z
```
#### Option 3: Per-type accuracy
This is chunithm-like.
```
perfect:x good:y miss:z

  hits | total:n acc:x%
breaks | total:n acc:x%
```

This PR currently follows option 1, however if the weightings are so different then I think it is more reasonable to expect option 2 or option 3.

## Redesigning `ScoreInfo`

To summarise all of the above, we need to redesign `ScoreInfo` such that it:
1. Provides some way to retrieve the exact bonus score.
2. "Tags" hit results uniquely based on hitobjects/judgements.

Here's a few ways this can be done:

#### Option 1: Statistic tagging

```
class ScoreInfo
{
    public int MaxCombo;
    public double Accuracy;

    public Dictionary<(string judgementAcronym, HitResult result), int> HitResultCounts =
    {
        { (Hit, Great), 10 },
        { (Hit, Miss), 2 },
        { (Break, Great), 5 },
        { (Break, Ok), 1 },
        { (Break, Miss), 0 },
        { (BonusObject, SmallBonusHit), 10 },
        { (OtherBonusObject, LargeBonusHit), 5 },

    };
}

public double ComputeBonusScore()
{
    double totalBonus = 0;

    foreach (var (key, count) in HitResultCounts)
    {
        if (!key.result.IsBonus())
            continue;

        var judgement = Activator.CreateInstance(key.judgementAcronym);
        totalBonus += judgement.NumericResultFor(key.result) * count;
    }

    return totalBonus;
}
```

#### Option 2: Optional tagging

```
class ScoreInfo
{
    public int MaxCombo;
    public double Accuracy;
    public double Bonus;

    public Dictionary<HitResult, int> HitResultCounts =
    {
        { Great, 15 },
        { Ok, 1 },
        { Miss, 2 },
        { SmallBonusHit, 10 },
        { LargeBonusHit, 5 },
    };

    // This can come later. We just won't have the ability to show separated stats as in options 2 or 3.
    public Dictionary<string, (int count, int maxCount)> JudgementCounts =
    {
        { Hit, (10, 12) },
        { Break, (5, 6) },
        { BonusObject, (10, X) },
        { OtherBonusObject, (5, Y) }
    }
}
```

Out of those two, I'd prefer option 2.

At the very least, maybe we should get the `Bonus` property in as it seems a safe choice, use that to resolve immediate standardisation issues, and continue discussions for the rest? Regardless of whether we go with any of these options (and even if none of these), the `Bonus` property could always be used. This would still require database changes, but it seems to be the least intrusive of them all - just a new column.

peppy:
personally i don't think we should be changing our scoring to support some game mechanic in another game someone is trying to emulate, and go ahead with what we have.

the primary goal here is to complete reimplementation of existing rulesets to a point we can go live. custom rulesets can either adapt, or request changes after the fact.

we're going to be here forever if we keep trying to make things more and more modular. the goal is to eventually support custom rulesets in an official way (providing online services for them) and that should come with some sane limitations so we can actually make it happen in a reasonable amount of time.

smoogipoo:
@LumpBloom7 As an alternative, I'd suggest adding a bunch of nested hitobjects in your Break object, and applying result to all simultaneously. Should achieve the same effect with this PR as is.

LumpBloom7:
To think that I'll get to revisit my very first attempt at Break notes XD.

I'll also be experimenting alternative compromises for the Break notes that doesn't involve the higher score weighting, just to see what sticks.

osu.Game/Rulesets/Judgements/Judgement.cs
```diff
@@ -95,6 +153,42 @@ protected virtual double HealthIncreaseFor(HitResult result)
         /// <returns>The numeric health increase of <paramref name="result"/>.</returns>
         public double HealthIncreaseFor(JudgementResult result) => HealthIncreaseFor(result.Type);
 
-        public override string ToString() => $"AffectsCombo:{AffectsCombo} MaxResult:{MaxResult} MaxScore:{MaxNumericResult}";
+        public override string ToString() => $"MaxResult:{MaxResult} MaxScore:{MaxNumericResult}";
+
+        public static double ToNumericResult(HitResult result)
```
bdach, line 158: I feel slightly uneasy with this function. Expressing the scoring relative to greats might look elegant, but I think I'd still advocate for the scores to be whole numbers to not open ourselves up to floating-point issues. These fractions all have non-well-defined denominators when it comes to base-2 and *will* drift when summed cumulatively. The more objects (and the more distinct results), the bigger the drift.

For instance, summing `Judgement.ToNumericResult(HitResult.Meh)` 6 times yields not 1, but 0.99999999999999989.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -475,6 +475,21 @@ protected void ApplyResult(Action<JudgementResult> application)
             if (!Result.HasResult)
                 throw new InvalidOperationException($"{GetType().ReadableName()} applied a {nameof(JudgementResult)} but did not update {nameof(JudgementResult.Type)}.");
 
+            // Some (especially older) rulesets use scorable judgements instead of the newer ignorehit/ignoremiss judgements.
```
bdach, line 478: Will this code be removable with the obsoletion of the methods marked as such in this PR, or does it have to remain here for now?

bdach (commented):
I've done a first rudimentary pass over this. Huge work :+1:

I've cross-checked the scores and judgements with the doc attached, and it seemed like it all checked out, but it's late and I was going somewhat fast, so it might be good to get a confirmation on that.

swoolcock:
If some hit objects require “better than perfect” scoring, it should be easy enough to just add an optional bonus judgment nested hit object, right? I’m just thinking about how to implement fever mode in Rush!.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: In practice this could lead implementers down the wrong path of doing this as a natural thing so I want to remove it somehow.

The validation check takes care of everything just below this, but I’m not sure how to signal correct usage to the user while still allowing current code... Maybe obsoletion notice via important log entries?

osu.Game/Rulesets/Judgements/Judgement.cs
smoogipoo: Yeah that’s a good point. Not super sure why I made it relative, maybe to match HP... Will revert.

smoogipoo:
@swoolcock That’s the solution we’re going for at this point, yup.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: I've added an obsoletion notice via logger messages.

osu.Game/Rulesets/Judgements/Judgement.cs
```diff
@@ -12,30 +13,68 @@ namespace osu.Game.Rulesets.Judgements
...
+        /// <summary>
+        /// The minimum <see cref="HitResult"/> that can be achieved - the inverse of <see cref="MaxResult"/>.
+        /// </summary>
+        public HitResult MinResult
+        {
+            get
+            {
+                switch (MaxResult)
+                {
+                    case HitResult.SmallBonus:
```
peppy, line 57: might be good to document these along with the types table (for the corresponding "max" results) as they are hard coupled with the max result.

osu.Game/Rulesets/Judgements/Judgement.cs
```diff
@@ -65,6 +105,21 @@ protected virtual double HealthIncreaseFor(HitResult result)
         {
             switch (result)
             {
+                default:
+                    return 0;
+
+                case HitResult.SmallTickHit:
+                    return DEFAULT_MAX_HEALTH_INCREASE * 0.05;
```
peppy, line 112: We already kinda covered this, but one remaining concern is rulesets that choose to fire way too many ticks may have weird HP distribution. Hopefully the health processor can make sense of this though.

It shouldn't be too bad, just mappers need to be aware of how adding the equivalent of buzz sliders affect hp increase. Which is already a known thing in osu! ruleset.

peppy (approved):

osu.Game/Rulesets/Judgements/Judgement.cs
smoogipoo: I've added a wiki page: https://github.com/ppy/osu/wiki/Scoring . Please check and point out any issues you have with it.

PR #10289, opened 2020-09-29
Improve the usability of the editor timeline with combo colour and indices

### LINKED ISSUE #8870: Editor timeline doesn't display (new) combos
Mainly opening these issues to discuss how these elements should be designed.

**Describe the bug:** 
Editor timeline should give a visual representation of new combo usage and combo count within combo. Right now all objects are white with no way to know what combo they belong to when looking at only the timeline.

Lazer:
osu_2020-04-26_21-19-14
Stable:
screenshot182

**osu!lazer version:** 
2020.421.0

peppy:
- Closes #8870.

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
@@ -185,7 +185,7 @@ namespace osu.Game.Screens.Edit.Compose.Components.Timeline
                     else
                         mainComponents.Colour = drawableHitObject.AccentColour.Value;
 
-                    var col = mainComponents.Colour.AverageColour.Linear;
+                    var col = mainComponents.Colour.TopLeft.Linear;
                     float brightness = col.R + col.G + col.B;
 
                     // decide the combo index colour based on brightness?
--- a/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
@@ -13,6 +13,8 @@ using osu.Framework.Graphics.Effects;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input.Events;
+using osu.Game.Graphics;
+using osu.Game.Graphics.Sprites;
 using osu.Game.Graphics.UserInterface;
 using osu.Game.Rulesets.Edit;
 using osu.Game.Rulesets.Objects;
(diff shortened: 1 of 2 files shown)


LittleEndu:
Closes #8870 ?

osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
```diff
@@ -34,11 +39,21 @@ public class TimelineHitObjectBlueprint : SelectionBlueprint
...
+        private readonly OsuSpriteText comboIndexText;
+
+        private Bindable<int> comboIndex;
+
         private const float thickness = 5;
 
         private const float shadow_radius = 5;
 
-        private const float circle_size = 16;
+        private const float circle_size = 24;
```
bdach, line 56: Not sure if this was a conscious design choice but I feel at this size the numbers are pretty hard to see properly. How about something like 32? Although I do admit that size looks pretty silly on combo-index-less rulesets...

osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
```diff
@@ -117,12 +146,54 @@ public TimelineHitObjectBlueprint(HitObject hitObject)
...
+                comboColour = drawableHitObject.AccentColour.GetBoundCopy();
+                comboColour.BindValueChanged(colour =>
+                {
+                    if (HitObject is IHasDuration)
+                        mainComponents.Colour = ColourInfo.GradientHorizontal(drawableHitObject.AccentColour.Value, Color4.White);
+                    else
+                        mainComponents.Colour = drawableHitObject.AccentColour.Value;
+
+                    var col = mainComponents.Colour.AverageColour.Linear;
+                    float brightness = col.R + col.G + col.B;
```
bdach, line 189: Not sure this should use `AverageColour`. Due to averaging over the gradient it can lead to some funky-looking results in some edge cases (hitobjects with duration and dark combo colour):

2020-09-29-223705_693x95_scrot

(this screenshot was taken on a dirty tree with `circle_size == 32`)

About the only place where I can think always using `AccentColour` could be a problem is on extreme zoom-out.

osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
peppy: May need to change the font, not sure. Don't want to go bigger in size though, looks wrong even on osu! ruleset.

osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineHitObjectBlueprint.cs
peppy: Oh good point. I was using average before I added the gradient..

smoogipoo (approved):

PR #10290, opened 2020-09-29
Adjust hit results for osu/catch/taiko

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10288

The above PR standardised scoring across all rulesets, which means GOOD awards 200 points instead of 100. To fix this, osu! and taiko have been changed to use the judgements Miss/Ok/Great.
Similarly, PERFECT is worth 350 score but catch expects 300 for catching a fruit, so catch has been changed to use the Great judgement instead.


peppy:
Needs conflict resolution

peppy (approved):

PR #10294, opened 2020-09-29
Add selection scale and rotate support

peppy:
- [x] Improve documentation
- [x] Avoid scale/rotation outside of playfield via early returns

Probably need to also make the scaling more 1:1 with movement (seems to not always be the case?)


osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -40,10 +163,57 @@ public override bool HandleMovement(MoveSelectionEvent moveEvent)
...
+            return new Quad(minPosition.X, minPosition.Y, size.X, size.Y);
+        }
+
+        /// <summary>
+        /// Returns rotated position from a given point.
+        /// </summary>
+        /// <param name="p">The point.</param>
+        /// <param name="center">The center to rotate around.</param>
+        /// <param name="angle">The angle to rotate (in degrees).</param>
+        internal static Vector2 Rotate(Vector2 p, Vector2 center, float angle)
```
bdach, line 202: nit again but this function has the freedom spelling of the word :smile: 
```suggestion
        internal static Vector2 Rotate(Vector2 p, Vector2 centre, float angle)
```

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -40,10 +163,57 @@ public override bool HandleMovement(MoveSelectionEvent moveEvent)
...
+        {
+            angle = -angle;
+
+            p.X -= center.X;
+            p.Y -= center.Y;
+
+            Vector2 ret;
+            ret.X = (float)(p.X * Math.Cos(angle / 180f * Math.PI) + p.Y * Math.Sin(angle / 180f * Math.PI));
+            ret.Y = (float)(p.X * -Math.Sin(angle / 180f * Math.PI) + p.Y * Math.Cos(angle / 180f * Math.PI));
```
bdach, line 211: Can use framework's `MathUtils.DegreesToRadians` here. (There also really ought to be a better way to do this entire thing with framework functions but I can't see an immediate one without introducing matrix math)

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -10,7 +15,125 @@ namespace osu.Game.Rulesets.Osu.Edit
 {
     public class OsuSelectionHandler : SelectionHandler
     {
-        public override bool HandleMovement(MoveSelectionEvent moveEvent)
+        public override ComposeSelectionBox CreateSelectionBox()
+            => new ComposeSelectionBox
+            {
+                CanRotate = true,
+                CanScaleX = true,
+                CanScaleY = true,
```
bdach, line 23: These should be off if only one object is selected.

* If only a slider is selected, it doesn't break but behaves funky (grabbing right and lower handles does nothing, while grabbing left and upper handles can move the object around due to the way this is implemented).
* Worse yet trying to scale a single hitcircle crashes to desktop because of division-by-zero reasons.

osu.Game/Screens/Edit/Compose/Components/ComposeSelectionBox.cs
```diff
@@ -0,0 +1,352 @@
...
+            {
+                AddRangeInternal(new[]
+                {
+                    new DragHandle
+                    {
+                        Anchor = Anchor.TopCentre,
+                        HandleDrag = e => OnScaleY?.Invoke(e, Anchor.TopCentre),
+                        OperationStarted = operationStarted,
+                        OperationEnded = operationEnded
+                    },
```
bdach, line 137: These could be DRY'd a little. The only differing parameter is the anchor - some bit-masking should make it so that this code has to only be written once but only called 8 times with different anchors.

bdach (changes requested):

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
peppy: yep, already noted in my head.

osu.Game/Screens/Edit/Compose/Components/ComposeSelectionBox.cs
peppy: yep, definitely. kinda sad you got to review this PR before I didn't a quality pass on it, shouldn't have PR'd yet 😅 .

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
peppy: copied this function from stable and forgot to fix it up 😅

peppy:
This is ready for review now.

Note that I've added support for scaling individual sliders. It doesn't behave perfectly but I think it's fine for an initial pass (the input to result isn't 1:1, which is the only real issue). Basically, it will always hold its original start position rather than attempt to scale arbitrarily around its own centre (which is a bit troublesome):

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -10,40 +16,180 @@ namespace osu.Game.Rulesets.Osu.Edit
...
+        private static Vector2 rotatePointAroundOrigin(Vector2 point, Vector2 origin, float angle)
+        {
+            angle = -angle;
+
+            point.X -= origin.X;
+            point.Y -= origin.Y;
+
+            Vector2 ret;
+            ret.X = (float)(point.X * Math.Cos(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Sin(angle / 180f * Math.PI));
+            ret.Y = (float)(point.X * -Math.Sin(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Cos(angle / 180f * Math.PI));
```
bdach, line 187: still a couple left!
```suggestion
            ret.X = (float)(point.X * Math.Cos(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Sin(MathUtils.DegreesToRadians(angle));
            ret.Y = (float)(point.X * -Math.Sin(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Cos(MathUtils.DegreesToRadians(angle));
```

bdach (approved):

smoogipoo (approved):

PR #10297, opened 2020-09-30
Set the window icon with SDL

swoolcock:
Sets the window icon for windows created with the SDL window backend.

- [x] Depends on https://github.com/ppy/osu/pull/10493


osu.Desktop/OsuGameDesktop.cs
```diff
@@ -138,6 +138,7 @@ public override void SetHost(GameHost host)
                 // SDL2 DesktopWindow
                 case DesktopWindow desktopWindow:
                     desktopWindow.CursorState.Value |= CursorState.Hidden;
+                    desktopWindow.SetIconFromStream(Assembly.GetExecutingAssembly().GetManifestResourceStream(GetType(), "lazer.ico"));
```
bdach, line 141: Might as well hoist out the resource stream to a local, if it's being used in both branches anyway.

peppy (approved):

PR #10302, opened 2020-09-30
Fix timing screen tests and add simple view of control points to timeline

peppy:
Fixes up test scene issues, adds a very simple display of control points to the timeline and adjusts the zoom accordingly. More complex stuff will come as follow-up PRs.

PR #10303, opened 2020-09-30
Add textbox entry for speed multiplier and volume

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10302
- [x] Depends on https://github.com/ppy/osu-framework/pull/3907
- [x] Framework update

I looked at making this share logic with the BPM entry textbox/slider but that one is a bit custom (doing conversion to beatlength).

### DIFF
--- a/osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
+++ b/osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
@@ -1,0 +1,74 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Framework.Graphics.UserInterface;
+using osu.Game.Graphics.UserInterfaceV2;
+using osu.Game.Overlays.Settings;
+
+namespace osu.Game.Screens.Edit.Timing
+{
+    internal class SliderWithTextBoxInput<T> : CompositeDrawable, IHasCurrentValue<T>
+        where T : struct, IEquatable<T>, IComparable<T>, IConvertible
+    {
+        private readonly SettingsSlider<T> slider;
+
+        public SliderWithTextBoxInput(string labelText)
+        {
+            LabelledTextBox textbox;
+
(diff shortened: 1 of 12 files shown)


osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
```diff
@@ -0,0 +1,74 @@
...
+                        {
+                            RelativeSizeAxes = Axes.X,
+                        }
+                    }
+                },
+            };
+
+            textbox.OnCommit += (t, isNew) =>
+            {
+                if (!isNew) return;
```
bdach, line 49: This guard breaks things. Try entering i.e. `100a` into the textbox twice; it'll get reverted first try but not the second. It's because `isNew` means "is the text being committed now the same as the last text being committed". Maybe a framework shortcoming?

osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                try
+                {
+                    slider.Bindable.Parse(t.Text);
```
bdach, line 53: This also has a nasty failure case. Try entering `-100` twice; it'll go `0` -> `-100` again. `Bindable.Parse` will succeed both times and the value will be clipped to 0 successfully both times, but because the underlying value of the bindable doesn't change the second time, the text will not get restored.

bdach (changes requested):

osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
peppy: Wow, that sounds incredibly wrong.

peppy: Fixed via https://github.com/ppy/osu-framework/pull/3907

osu.Game/Screens/Edit/Timing/SliderWithTextBoxInput.cs
peppy: Fixed via https://github.com/ppy/osu/pull/10303/commits/e0a0902a15f9e1dac34b8795035c2ef48e47dbec (ensuring we always trigger a reformat after commit, even when the value is "valid" by clamping).

smoogipoo (approved):

PR #10306, opened 2020-09-30
Make DifficultyIcon support dynamic star difficulty

smoogipoo:
Preview: https://drive.google.com/file/d/1syCxOYoG6ci56pWy_jUZZJxzGR30CgBi/view?usp=sharing

This is set up to work for timeshift, however it doesn't work in practice since timeshift doesn't retrieve the locally-available beatmap so it falls back to whatever's coming in via `BeatmapInfo.StarDifficulty`. _Something_ needs to retrieve the local beatmap somewhere, unsure if multiplayer or `BeatmapDifficultyManager` at this point. I think there was a reason why I didn't do it for multiplayer.

This also doesn't support grouped difficulty icons yet.

The calculation is implemented through a DLUW to prevent offscreen calculations from taking place in e.g. song select.

### DIFF
--- a/osu.Game/Beatmaps/BeatmapDifficultyManager.cs
+++ b/osu.Game/Beatmaps/BeatmapDifficultyManager.cs
@@ -114,6 +114,25 @@ namespace osu.Game.Beatmaps
             return computeDifficulty(key, beatmapInfo, rulesetInfo);
         }
 
+        /// <summary>
+        /// Retrieves the <see cref="DifficultyRating"/> that describes a star rating.
+        /// </summary>
+        /// <remarks>
+        /// For more information, see: https://osu.ppy.sh/help/wiki/Difficulties
+        /// </remarks>
+        /// <param name="starRating">The star rating.</param>
+        /// <returns>The <see cref="DifficultyRating"/> that best describes <paramref name="starRating"/>.</returns>
+        public static DifficultyRating GetDifficultyRating(double starRating)
+        {
+            if (starRating < 2.0) return DifficultyRating.Easy;
+            if (starRating < 2.7) return DifficultyRating.Normal;
+            if (starRating < 4.0) return DifficultyRating.Hard;
+            if (starRating < 5.3) return DifficultyRating.Insane;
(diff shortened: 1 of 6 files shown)


Gonzalo-Bruna:
is it a known issue that the star rating in the beatmap wedge doesn't update when you apply mods?

bdach:
I'm pretty sure it is, the buffered container there doesn't update entirely correctly. It's a separate issue.

osu.Game/Beatmaps/BeatmapDifficultyManager.cs
```diff
@@ -307,5 +307,19 @@ public StarDifficulty(double stars, int maxCombo)
 
             // Todo: Add more members (BeatmapInfo.DifficultyRating? Attributes? Etc...)
         }
+
+        public DifficultyRating DifficultyRating
+        {
+            get
+            {
+                if (Stars < 2.0) return DifficultyRating.Easy;
```
peppy, line 315: where are these values coming from? osu-web? may be good to leave an inline comment linking it up.

osu.Game/Beatmaps/Drawables/DifficultyIcon.cs
```diff
@@ -82,16 +110,73 @@ private void load(OsuColour colours)
...
             };
+
+            difficultyBindable.BindValueChanged(difficulty => background.Colour = colours.ForDifficultyRating(difficulty.NewValue.DifficultyRating));
+        }
+
+        public ITooltip GetCustomTooltip() => new DifficultyIconTooltip();
+
+        public object TooltipContent => shouldShowTooltip ? new DifficultyIconTooltipContent(beatmap, difficultyBindable) : null;
+
+        private class DifficultyRetriever : Drawable
```
peppy, line 125:
```suggestion
        private class DifficultyRetriever : Component
```

not hugely important, but better defines what it is.

peppy (changes requested):

bdach (approved):
this looks real cool :+1:

peppy (approved):

Issue #9176, opened 2020-06-02
Rebind mods/random/options


boswelja:
**Describe the new feature:** 
Add the ability to rebind the mods/random/options buttons in song select. Currently they're bound by default to F1-F3 respectively, and there doesn't seem to be a way to change this.
**Proposal designs of the feature:**

boswelja:
I haven't looked at the code very much, but I feel like this would be a good place for me to start contributing. As long as it's something you guys want, can I have a go at it?

peppy:
Yep, definitely

PR #10307, opened 2020-09-30
allow rebinding of mods/random/options in song selector

charlie-gray:


bdach:
* Please don't PR off `master`. We mention this in [the contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md).
* Did you see the previous PR that was supposed to address this (#9194) or the discussion within?

osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -119,6 +126,16 @@ public enum GlobalAction
...
 
+        //Song selector bindings
+        [Description("Choose a random song")]
+        FooterRandom,
+
+        [Description("Change beatmap options")]
+        FooterOptions,
+
+        [Description("Select mods")]
+        FooterMods,
```
bdach, line 137: To expand on the above comment, doing this appears to work but actually blocks music controller keybindings in song select (previous song and play/pause) because of a binding conflict. This not what we want.

bdach (changes requested):

osu.Game/Input/Bindings/GlobalActionContainer.cs
charlie-gray: Thanks for the great feedback!I'll get on this right away.

charlie-gray:
Closing this pull request for right now. Will tidy up branches, implement your suggestions, and link to issue #9176. Sorry about that! Does this sound good to you @bdach?

PR #10309, opened 2020-10-01
Allow rebinding of mods/random/options in song selector

### LINKED ISSUE #9176: Rebind mods/random/options

charlie-gray:
Resolves #9176.

Allows the user to rebind mods/random/options buttons in the footer of the song selector.

New version of #10307 because it did not follow contributing guidelines. In addition, I moved the enums for SongSelectorBindings so that it does not interfere with existing user keybinds.

### DIFF
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -92,8 +92,7 @@ namespace osu.Game.Input.Bindings
             new KeyBinding(InputKey.F2, GlobalAction.FooterRandom),
             new KeyBinding(InputKey.F3, GlobalAction.FooterOptions),
         };
-
-
+        
         protected override IEnumerable<Drawable> KeyBindingInputQueue =>
             handler == null ? base.KeyBindingInputQueue : base.KeyBindingInputQueue.Prepend(handler);
     }
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -205,7 +205,5 @@ namespace osu.Game.Input.Bindings
 
         [Description("Timing Mode")]
         EditorTimingMode,
-
     }
-
 }
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -50,7 +50,6 @@ namespace osu.Game.Input.Bindings
             new KeyBinding(InputKey.KeypadEnter, GlobalAction.Select),
         };
 
-        
         public IEnumerable<KeyBinding> EditorKeyBindings => new[]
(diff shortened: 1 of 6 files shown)


bdach:
I'm sorry to say, but this version of the PR suffers from the exact same problem as the previous one - it breaks existing users' bindings. For instance, after checking out this PR having been on a working `master` previously, my <kbd>Enter</kbd> key switched with my <kbd>Escape</kbd> key.

This is because you've added the new bindings in between existing ones, instead of at the end, therefore breaking the mapping in the SQLite database. If you add them last as not to break old ones, I believe you'll find that they don't work as the music controller bindings will take priority. Either way, both issues are more fundamental and need to be resolved properly. To that end I'll go ahead and remove the *good first issue* tag from the original issue to avoid confusion (although the first attempt at fixing this, with full discussion, was linked both in the issue by GitHub itself and by myself manually in your previous try yesterday).

Additionally, you seem to be using CI to test code style, instead of checking locally using `./InspectCode.ps1` or otherwise. Please read the contribution guidelines I linked you yesterday completely, especially the parts concerned with testing your code before pushing (unit tests and code style analysis).

bdach (changes requested):
as above

charlie-gray:
Thank you for the great feedback. I appreciate it! I'm going to try moving the order in which key binds are added and also adding a function to prevent the keys for media control from working in the song selector. Does that sound good?

bdach:
I'm not sure you would be able to add such a function locally in this repository. `GlobalAction` internally uses `KeyBindingContainer`s to work; `GlobalActionContainer` has a `SimultaneousBindingMode` of `None`. In that mode, only [the first binding with the most keys is triggered](https://github.com/ppy/osu-framework/blob/446ee955a5dc4239bbda1030defbf27e57c4d308/osu.Framework/Input/Bindings/KeyBindingContainer.cs#L207-L209).

In this case, there's a draw between the song select bindings and the music controller bindings, as they both consist of exactly one key. Given that `OrderByDescending` is a stable sort, if the song select bindings are moved to last, they will never actually propagate to the `FooterButton`s. You theoretically *could* change the `SimultaneousBindingMode` to `Unique`, doing only that would result in simultaneously triggering *both* bindings at the same time, which is not what we want.

In my view a proper resolution would consist of:

* migrating away from int-based databased key bindings in favour of string-based to make sure that adding enum members in the middle is no longer ever a problem (see #9549),
* (optionally) adding a way to specify which key bindings should take precedence rather than rely on their member order in the enum.

Because resolving this takes a significant amount of work for relatively little gain, I hope you can understand that we probably wouldn't be looking to address this right away and is also why the previous PRs in these areas were backburnered.

charlie-gray:
Okay. Closing this PR for right now.

PR #10310, opened 2020-10-01
Add buttons to flip/rotate the current selection by known amounts

### LINKED ISSUE #3247: Proposal: Faster rotating and scaling objects in editor.
Today I was doing something in Affinity (Photoshop replacement) and then it got me. Why I can't rotate object in osu! by simply dragging cursor!
affinity-rotating

I think this option would fit perfectly with that boxes around circles and sliders that were added to lazer editor.
My concept is about adding small draggable circle above slider box and selected objects box. 
Also scaling objects this way would be awesome. 
Visualisation:
slider
objects

And yes, I know that there are two ways of rotating (by origin).

peppy:
This adds buttons to flip and rotate by 90 degrees at a time. Designs *probably* aren't final but work well enough for the time being.

This also tidies up the prerequisite PR further, so it may be better to just review this one. Not sure.

- [ ] Builds on https://github.com/ppy/osu/pull/10294.
- Fixes CanX methods triggering a recreate even when there's no change (can be seen by seeking forwards through a beatmap while having a selection active).
- Changes event flow to not pass down `DragEvents`.

I started by trying to use the scale function for doing flip operations but it really just over-complicates things, especially when slider paths are involved (they require considerably different logic).

Closes https://github.com/ppy/osu/issues/3247


osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -10,40 +16,219 @@ namespace osu.Game.Rulesets.Osu.Edit
...
+        private static Vector2 rotatePointAroundOrigin(Vector2 point, Vector2 origin, float angle)
+        {
+            angle = -angle;
+
+            point.X -= origin.X;
+            point.Y -= origin.Y;
+
+            Vector2 ret;
+            ret.X = (float)(point.X * Math.Cos(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Sin(angle / 180f * Math.PI));
+            ret.Y = (float)(point.X * -Math.Sin(MathUtils.DegreesToRadians(angle)) + point.Y * Math.Cos(angle / 180f * Math.PI));
```
turbedi, line 226:
```suggestion
            ret.X = (point.X * MathF.Cos(MathUtils.DegreesToRadians(angle)) + point.Y * MathF.Sin(angle / 180f * MathF.PI));
            ret.Y = (point.X * -MathF.Sin(MathUtils.DegreesToRadians(angle)) + point.Y * MathF.Cos(angle / 180f * MathF.PI));
```
Avoids the cast to float.

peppy: Have applied this to the [prerequisite PR](https://github.com/ppy/osu/pull/10294), thanks.

bdach (approved):
looks really cool, and I agree that it cleaned up quite nicely :+1: 

About the only minor complaint I might have is that the "handle" buttons still have `OnDrag{Start,End}` starting and ending operations - but I wouldn't say it's a big deal because the change handler will ignore operations that do nothing anyway.

smoogipoo (approved):

PR #10313, opened 2020-10-01
Display all control point information on the main editor timeline

### LINKED ISSUE #10293: Add "slider velocity" visualisation in editor
Add timing point visualization that will simplify the mapping of SV maps
Screenshot_2123
You can also add quick editing on click

peppy:
May look a bit busy, but it works quite well in most scenarios. I think it'll be fine as a first pass.

- [x] Depends on #10312 
- Closes https://github.com/ppy/osu/issues/10293

image

Volume changes are visualised by vertical fill:

image

image

image

Note that kiai time is not added yet, as I want to figure out a better way to show that (as a span).


osu.Game/Screens/Edit/Compose/Components/Timeline/SamplePointPiece.cs
```diff
@@ -0,0 +1,81 @@
...
+                    Y = -5,
+                    Anchor = Anchor.BottomLeft,
+                    Alpha = 0.9f,
+                    Rotation = -90,
+                    Font = OsuFont.Default.With(weight: FontWeight.SemiBold)
+                }
+            };
+
+            volume.BindValueChanged(volume => volumeBox.Height = volume.NewValue / 100f, true);
+            bank.BindValueChanged(bank => text.Text = $"{bank.NewValue}", true);
```
bdach, line 78: is this an unnecessary string interpolation, or just an unorthodox coercion to an empty string?

bdach:
Code-wise this looks ok to me, but I'm not 100% on the visual look. I know nothing about mapping, but in the worst cases the information density is too much for how much screen space is actually available:

2020-10-01-212313_284x147_scrot

The bpm view is sort of illegible (truncated first digit, rest obscured by the blueprint), and the sample point is actually clipped from the top by the bonus view (so you can't tell whether it's like 90% or 100%).

I experimented with something like:

2020-10-01-211505_178x141_scrot

but it can also be misleading (looks as if the bpm control point starts after the first measure). Not sure if this can be super legible without i.e. allowing to turn off individual control point type display.

I guess we could always take this as-is and iterate further if need be.

peppy:
i intentionally did it this way – the BPM text isn't really supposed to be legible, more there as a marker/colour to give the modder or mapper an idea that it has changed.

eventually we would also want a toggle for blueprint display, although you can already do this by entering timing view where the BPM is not obstructed.

osu.Game/Screens/Edit/Compose/Components/Timeline/SamplePointPiece.cs
peppy: oversight, it would seem.

bdach (dismissed):
looks good to go :+1:

smoogipoo (dismissed):

smoogipoo:
Needs some conflict resolution.

smoogipoo (approved):

PR #10316, opened 2020-10-01
Fix default spinner flashing yellow glow before completion

smoogipoo:
I'm not sure what regressed this, if this is even a regression. There's really two issues here:
1. The state event happens in `DrawableSpinner.LoadComplete`, which happens much earlier than `DefaultSpinnerDisc.LoadComplete` (a child of `DrawableSpinner`).
2. `ApplyCustomUpdateState` is propagated upwards from nested hitobjects, so it'd receive `Hit` and `Idle` events as ticks get hit and as ticks load and trigger their initial `ApplyCustomUpdateState`.

Not sure what the best way to test this is, or if it needs to be tested?

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
@@ -93,6 +93,8 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 
             drawableSpinner.RotationTracker.Complete.BindValueChanged(complete => updateComplete(complete.NewValue, 200));
             drawableSpinner.ApplyCustomUpdateState += updateStateTransforms;
+
+            updateStateTransforms(drawableSpinner, drawableSpinner.State.Value);
         }
 
         protected override void Update()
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
```diff
@@ -124,6 +126,9 @@ protected override void Update()
 
         private void updateStateTransforms(DrawableHitObject drawableHitObject, ArmedState state)
         {
+            if (!(drawableHitObject is DrawableSpinner))
```
peppy, line 129: how can this happen?

smoogipoo: Because the event is propagated upwards from nested hitobjects.

peppy: so it fires multiple times on the spinner itself? o_O

smoogipoo: Yeah:
image

peppy: wow... we may want to check other usages of that.

are you able to change the xmldoc while we're here to make it more obvious that this is the case?

smoogipoo: Done. Hopefully that's better?

peppy (approved):

Issue #6115, opened 2019-09-14
Cursortrail rate setting.


anexistantperson:
Describe the new feature:
In stable you can make the cursors really small then edit your config to make them really big to create a incredibly smooth cursor trail.
Now in lazer you can no longer edit the config because it just resets. However if there was an option to change the rate of the cursortrail and maybe even the size of it,
it would be much easier to switch skins around without having to restart and edit a file every time you do so, along with the cursor not being all pixelated and weird, Which would obviously be very convenient for those who like to have the smooth cursor trails.
It would probably be best to have the option have a maximum of 4-8.

peppy:
Is this a `skin.ini` setting? If so, it will eventually be supported again.

anexistantperson:
No it was never in stable. What i'm saying is the cursor trails seems to just place the cursortrail.png after a specific distance. The cursor size affects that distance and I was thinking it would be easier to just have the distance a separate option and making it easier to make it smooth for skinners while also having it look good.

Before you could make the cursor images small then make the cursor big ingame but it only really does its job if you have it 4x the size (atleast in my opinion) to trick the game into having a higher rate but that also cant be done in lazer because you cant change it to a value over 2 anymore. even with editing the config. Example:
![Untitled](https://user-images.githubusercontent.com/55331139/64918484-cb5fbe00-d76c-11e9-8f9a-4dc262db3599.png)

peppy:
This will definitely be added once we finish skin settings.

LittleEndu:
Will the size of the cursor trail also be customizable in the future? Right now the trail size doesn't match stable's even if you have stable's cursor size setting set to 1x (I assume that's because lazer doesn't use cursotrail.png and instead draws regular shapes)

![osu_2020-03-02_22-15-38](https://user-images.githubusercontent.com/11320483/75715854-cd93e580-5cd6-11ea-979a-d5e8fa48ab40.png)
![screenshot175](https://user-images.githubusercontent.com/11320483/75715856-cec51280-5cd6-11ea-9741-492ed57b5f19.png)

LittleEndu:
I think the issue here is that on stable, density of the trail polygons stays the same no matter what cursor size you use. So if you move your 1.0x cursor size cursor when your trail is a 8x8 file it will get a new polygon every 4px for an overlap of 4px, but if you move your 2.0x cursor size cursor keeping the same trail file dimensions, you will still get a new polygon every 4px but now you overlap 12px.

PR #10321, opened 2020-10-01
Try to match stable's cursor trail density

LittleEndu:
This addresses the issue in #6115 but does not close it (as this PR doesn't add the ability to adjust the density, it just makes it more in match with stable)

Quoting my comment from that issue:
> I think the issue here is that on stable, density of the trail polygons stays the same no matter what cursor size you use. So if you move your 1.0x cursor size cursor when your trail is a 8x8 file it will get a new polygon every 4px for an overlap of 4px, but if you move your 2.0x cursor size cursor keeping the same trail file dimensions, you will still get a new polygon every 4px but now you overlap 12px.

~~If something like this could be considered I can record comparison videos. Going to open the PR as a draft because this.~~

Comparison video here: https://youtu.be/iHeu8hPEYAY
This PR is top-left, current release is top-right, stable is bottom. 

Rendered another video with chroma key: https://youtu.be/CR88gzm0I7A
stable is left, this PR is middle, current release is right (and unkeyed, it looks bad with video compressions, but it also looks bad in-game, that's why I want this change)

### DIFF
--- a/osu.Game.Rulesets.Osu/Skinning/LegacyCursorTrail.cs
+++ b/osu.Game.Rulesets.Osu/Skinning/LegacyCursorTrail.cs
@@ -44,7 +44,7 @@ namespace osu.Game.Rulesets.Osu.Skinning
 
         protected override bool InterpolateMovements => !disjointTrail;
 
-        protected override float IntervalMultiplier => Math.Max(cursorSize.Value, 1);
+        protected override float IntervalMultiplier => 1 / Math.Max(cursorSize.Value, 1);
 
         protected override bool OnMouseMove(MouseMoveEvent e)
         {
--- a/osu.Game.Rulesets.Osu/Skinning/LegacyCursorTrail.cs
+++ b/osu.Game.Rulesets.Osu/Skinning/LegacyCursorTrail.cs
@@ -1,9 +1,12 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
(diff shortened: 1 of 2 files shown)


peppy:
Should you not be making this change in `LegacyCursorTrail`?

bdach:
Please unmark as draft whenever you deem the code is ready. I will abstain from reviewing until then.

LittleEndu:
I just wanted to know if this (changing the density of the trail to *almost* match what stable does by making it denser on higher cursor size) is something that we want to do, before marking it ready. Now that the implementation is only in legacy cursor trail, I don't see why not try to support something stable did anyway. I'll record comparison vid then mark it as ready.

bdach:
I don't really know how to read the comparison videos, given that the trail on it is... continuous. A comparison of a non-continuous trail on stable with how it looks on lazer would actually give a frame of reference as to how the two compare.

LittleEndu:
> I don't really know how to read the comparison videos, given that the trail on it is... continuous. A comparison of a non-continuous trail on stable with how it looks on lazer would actually give a frame of reference as to how the two compare.

The non-continuous trail isn't changed. That trail gets new sprites based on time (so the faster you move your cursor, the sparser it would be). I don't think cursor size or the dimensions of the cursortrail.png change the density of the non-continuous trail.

The comparison videos were meant to show that current release will leave bigger trails (that's trails with higher dimensions, mine is 24x24) more blocky looking which is the entire reason people increase their cursor size on stable to get more sprites per distance. Denser trails look more fluid.

smoogipoo (approved):
Seems to behave fine

PR #7210, opened 2019-12-14
Add drawable content for slider tails

mcendu:
Closes #7206

PR #10323, opened 2020-10-02
Add legacy skin "sliderendcircle" support

### LINKED ISSUE #7206: Display slider end and reverse arrow circle
**Proposal designs of the feature:** Disabled by default. Enabled if skin contains `hitcircle.png`. An option for that may or may not be added.

Very likely dupe & would do it myself.

peppy:
- Closes https://github.com/ppy/osu/issues/7206
- Replaces and closes https://github.com/ppy/osu/pull/7210.
- [x] Depends on #10322.

# Non-repeats

stable:

lazer:

# Repeats

stable:

lazer:


LittleEndu:
How does this work with hidden?

peppy:
No idea, you'll have to try it.

smoogipoo:
It doesn't work with HD.

peppy:
Is the expected behaviour that they should not be shown at all?

smoogipoo:
Should fade along with the head it looks like.

smoogipoo:
Same thing also happens for repeats - the circle fades out but the arrow stays, which is going to need special handling.

peppy:
I've added hidden support, but haven't visually compared it to ensure 1:1 yet. I *think* it's correct though.

smoogipoo (approved):
Seems to behave alright

Issue #10314, opened 2020-10-01
Changes to control points in editor don't trigger change handler correctly


peppy:
No prompt to save on making any changes on the timing screen.

peppy:
Priority should be given to this for control points specifically, as it can be very frustrating when trying to time a beatmap.

PR #10325, opened 2020-10-02
Add the ability to move a control point around in time

peppy:
For now, this does a reconstruction of the group, as there is no way to modify the time (I believe this was the intentional direction to keep things simple, and should be fine for the time being).

- [x] Depends on #10324

### DIFF
--- a/osu.Game/Screens/Edit/Timing/GroupSection.cs
+++ b/osu.Game/Screens/Edit/Timing/GroupSection.cs
@@ -1,0 +1,96 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Graphics.UserInterface;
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Edit/Timing/GroupSection.cs
```diff
@@ -0,0 +1,103 @@
...
+                textBox.Current.Disabled = false;
+                textBox.Text = $"{group.NewValue.Time:n0}";
+            }, true);
+        }
+
+        private void changeSelectedGroupTime(in double time)
+        {
+            var currentGroupItems = SelectedGroup.Value.ControlPoints.ToArray();
+
+            Beatmap.Value.Beatmap.ControlPointInfo.RemoveGroup(SelectedGroup.Value);
```
smoogipoo, line 95: So how do you envision this working from a change handler perspective? Is `EditorBeatmap` going to bind to `ControlPointInfo.Groups` and bind to `ItemAdded`/`ItemRemoved` events, then bind to events from the bindables?

Or, is section going to call something like `changeHandler.SaveState()`?

peppy: I have a PR in the works to fix that (https://github.com/ppy/osu/issues/10314). It just saves a state when the `ControlPointInfo.Groups` `BindableList` fires a change.

peppy: Have added the local logic required to ensure only one state saved.

smoogipoo: It doesn't seem to want to undo the changes?

smoogipoo (approved):

PR #10326, opened 2020-10-02
 Only regenerate autoplay on editor state change

peppy:
Alleviates performance concerns without having much of a visual effect (unless you're playing the beatmap while dragging a selection, which is very unlikely).

We can potentially change this in the future if/when we have partial regeneration support.

### DIFF
--- a/osu.Game/Rulesets/Edit/DrawableEditRulesetWrapper.cs
+++ b/osu.Game/Rulesets/Edit/DrawableEditRulesetWrapper.cs
@@ -40,17 +40,21 @@ namespace osu.Game.Rulesets.Edit
             Playfield.DisplayJudgements.Value = false;
         }
 
+        [Resolved]
+        private IEditorChangeHandler changeHandler { get; set; }
+
         protected override void LoadComplete()
         {
             base.LoadComplete();
 
             beatmap.HitObjectAdded += addHitObject;
-            beatmap.HitObjectUpdated += updateReplay;
             beatmap.HitObjectRemoved += removeHitObject;
+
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Edit/IEditorChangeHandler.cs
```diff
@@ -10,6 +11,11 @@ namespace osu.Game.Screens.Edit
     /// </summary>
     public interface IEditorChangeHandler
     {
+        /// <summary>
+        /// Fired whenever a state change occurs.
+        /// </summary>
+        public event Action OnStateChange;
```
smoogipoo, line 17:
```suggestion
        event Action OnStateChange;
```
Unnecessary, public by default.

osu.Game/Rulesets/Edit/DrawableEditRulesetWrapper.cs
```diff
@@ -69,7 +80,7 @@ private void removeHitObject(HitObject hitObject)
             drawableRuleset.Playfield.Remove(drawableObject);
             drawableRuleset.Playfield.PostProcess();
 
-            drawableRuleset.RegenerateAutoplay();
+            updateReplay();
```
smoogipoo, line 83: Is this still required? Because regeneration is now bound to the change handler.

smoogipoo (changes requested):

osu.Game/Rulesets/Edit/DrawableEditRulesetWrapper.cs
peppy: Likely not.

smoogipoo (approved):

PR #10329, opened 2020-10-02
Add drag-drop support in editor setup screen for audio and background images

peppy:
The background image doesn't update everywhere immediately, but I think that can come at a later point (it's not a simple one to approach, not even sure how we should be handling that flow but we probably want a `WorkingBeatmapUpdated` event or something...).

I was hoping to make visual cues for the drop events but our window backend doesn't support previewing file drags yet, so this is what it is.

Also:

- Adds ability to register import handlers game-wide.
- ~Exits import process after first handler handles a file.~ reverted this as it's slightly wrong. will address at a later point.

I'm aware that the unregistering from the handler is not completely correct – if you go to the setup screen once you will still be able to import from the compose/timing screens until exiting the editor. This will eventually be fixed by better screen management logic (maybe `ScreenCarousel` will make a comeback?)


osu.Game/Screens/Edit/Setup/SetupScreen.cs
```diff
@@ -196,6 +271,12 @@ private void onCommit(TextBox sender, bool newText)
             Beatmap.Value.Metadata.AuthorString = creatorTextBox.Current.Value;
             Beatmap.Value.BeatmapInfo.Version = difficultyTextBox.Current.Value;
         }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+            game.UnregisterImportHandler(this);
```
smoogipoo, line 278:
```suggestion
            game?.UnregisterImportHandler(this);
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #10331, opened 2020-10-02
Fix tests specifying steps in their constructors

peppy:
Not sure if this is all of them, but should be an improvement.. all the other similar (hitobject) test already used a "Various" test method or run steps in BDL/LoadComplete; just left those for now.


osu.Game.Rulesets.Mania.Tests/TestSceneNotes.cs
```diff
@@ -28,8 +28,8 @@ namespace osu.Game.Rulesets.Mania.Tests
     [TestFixture]
     public class TestSceneNotes : OsuTestScene
     {
-        [BackgroundDependencyLoader]
-        private void load()
+        [Test]
+        public void TestVariousNotes()
         {
             Child = new FillFlowContainer
```
smoogipoo, line 34: Put this inside a step.

smoogipoo (dismissed):

bdach (approved):
looks good :+1:

PR #10338, opened 2020-10-02
Remove WorkingBeatmap from PerformanceCalculator

Game4all:
# Summary

This changes the `PerformanceCalculator` to use pre-calculated `DifficultyAttributes` given in constructor instead of calculating them on instantiation, therefore requiring to have a `WorkingBeatmap` around, and removes the `Beatmap` property as well. The `OsuPerformanceCalculator` was using it to compute number of hit circles, that computation has been moved to the osu difficulty attributes calculation.

This is a breaking change for rulesets api as it changes method overloadings.


osu.Game/Rulesets/Ruleset.cs
```diff
@@ -158,7 +158,7 @@ protected Ruleset()
 
         public abstract DifficultyCalculator CreateDifficultyCalculator(WorkingBeatmap beatmap);
 
-        public virtual PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, ScoreInfo score) => null;
+        public virtual PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, DifficultyAttributes attributes, ScoreInfo score) => null;
```
bdach, line 161: Why is this still taking in a `WorkingBeatmap`? It's unused entirely.

Additionally, we will probably want to have something like

```csharp
public PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, ScoreInfo score)
{
    var difficultyCalculator = CreateDifficultyCalculator(beatmap);
    var difficultyAttributes = difficultyCalculator.Calculate(score.Mods);
    return CreatePerformanceCalculator(beatmap, difficultyAttributes, score);
}
```

It doesn't make this change non-breaking for ruleset implementors, but it *does* give time for consumers of this method to adapt. And [there *are* consumers](https://github.com/search?l=C%23&q=CreatePerformanceCalculator&type=Code), [`osu-tools`](https://github.com/ppy/osu-tools) included.

In fact we might want to add that and keep that forever as again I'm not sure whether making consumers have to create a difficulty calculator to be even able to use the performance calculator is good design.

osu.Game.Rulesets.Osu/Difficulty/OsuDifficultyAttributes.cs
```diff
@@ -11,5 +11,6 @@ public class OsuDifficultyAttributes : DifficultyAttributes
         public double SpeedStrain;
         public double ApproachRate;
         public double OverallDifficulty;
+        public int HitCirclesCount;
```
bdach, line 14: nit: rename to `HitCircleCount`

bdach (dismissed):

osu.Game/Rulesets/Ruleset.cs
Game4all: looks like I had a brain fart at the moment I committed this 😅

osu.Game/Rulesets/Ruleset.cs
```diff
@@ -158,7 +158,15 @@ protected Ruleset()
 
         public abstract DifficultyCalculator CreateDifficultyCalculator(WorkingBeatmap beatmap);
 
-        public virtual PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, ScoreInfo score) => null;
+        public virtual PerformanceCalculator CreatePerformanceCalculator(DifficultyAttributes attributes, ScoreInfo score) => null;
+
+        [Obsolete("Use the DifficultyAttributes overload instead.")]
+        public PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, ScoreInfo score)
```
bdach, line 164: As stated above, I'd like more opinions on whether we should actually be obsoleting this.

smoogipoo: Could remain as non-obsolete - may be useful for osu!tools and similar usages. My bigger concern at this point is whether this is a reasonable breaking change to make, because I think the tau ruleset already has a pp calculator.

Game4all: Just did a symbol search on all available rulesets repositories and it seems no custom ruleset implemented a performance calculator yet (didn't got any results).

bdach: I also did the same and saw no hits for overrides (but did see hits for calls of the method). I think the obsoletion warning should be removed on that basis (maybe we even want to make the overload that takes `DifficultyAttributes` `internal`? not sure)

smoogipoo: It's fine to keep the override public. And remove the obsoletion notice.

bdach (dismissed):

osu.Game/Rulesets/Ruleset.cs
```diff
@@ -158,7 +158,14 @@ protected Ruleset()
 
         public abstract DifficultyCalculator CreateDifficultyCalculator(WorkingBeatmap beatmap);
 
-        public virtual PerformanceCalculator CreatePerformanceCalculator(WorkingBeatmap beatmap, ScoreInfo score) => null;
+        public virtual PerformanceCalculator CreatePerformanceCalculator(DifficultyAttributes attributes, ScoreInfo score) => null;
```
peppy, line 161: shall we specify `[CanBeNull]` here (and below) for clarity?

peppy (approved):

Issue #8721, opened 2020-04-11
Missing the option to see (and save) your replay after failing the map.


ghost:
**Describe the new feature:** 

We can't see the replay ( in the fail screen ) of a score we just failed by pressing F1 (or F2 I don't remember) like on stable.

In addition being able to save your failed replay with another key while in the fail screen would be very nice.

szmarczak:
I think that replays of fails is crucial. That way you can easily see where you make mistakes.

memesnightmares:
Expanding on this idea a bit, it would be nice to allow quick-replay or saving to local leaderboards from the fail screen (and maybe also atomic save and quit on the pause screen). These saved replays could be labeled in the local leaderboard with some icon indicating that they're incomplete/cut/severed replays.

Joehuu:
Implemented in https://github.com/ppy/osu/pull/18785.

PR #10339, opened 2020-10-02
Add ability to save a failed score and replay to local leaderboards

santerinogelainen:
Related to #8721

Adds the ability to save your score to local leaderboards if you failed the map. Since the whole score is saved, you can also watch the replay through the scores info page.

This also means the addition of a new ScoreRank (ScoreRank.F). This could cause major problems to existing architecture but my knowledge of the code base is limited.

### DIFF
--- a/osu.Game/Scoring/ScoreRank.cs
+++ b/osu.Game/Scoring/ScoreRank.cs
@@ -7,6 +7,9 @@ namespace osu.Game.Scoring
 {
     public enum ScoreRank
     {
+        [Description(@"F")]
+        F,
+
         [Description(@"D")]
         D,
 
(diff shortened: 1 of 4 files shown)


bdach:
I'm 99% sure introducing the F rank should be a stand-alone pull request, and probably a huge one, at that. Two immediate breakages off the top of my head from what you've done here:

* No support in `DrawableRank` (how will an F rank show? probably like a D - in which case it needs to be changed to avoid visual confusion)
* Because ranks are stored as the underlying enum values in the database, doing what you've done here will cause all local scores to be bumped down one rank.

bdach (changes requested):
as above - this PR does not make sense without assessing the impact of adding the F rank properly

santerinogelainen:
@bdach Alright, I kinda figured. I'm closing this for now since assessing the impact of the F-rank is most likely too big of a task for me right now. I can reopen if the F-rank gets implemented at some point, or some other solution is found that does not require adding a new ScoreRank.

peppy:
Not sure this needs to be closed immediately. Seems like this is mostly usable apart from that oversight. Will take a look sometime this week.

PR #10343, opened 2020-10-02
Replace direct references to HitResult.Miss wherever applicable

### LINKED ISSUE #10354: Some hitsounds will play even if missed
**Describe the bug:**

When a spinner reach the end (either in a gameplay or in replay), the hit sound sounds like it only plays halfway and then stopped. This is observed every time in every spinner of any map.

Videos of a spinner in version 2020.1002.0 vs in version 2020.925.0: 
[spinner-end-hitsound.zip](https://github.com/ppy/osu/files/5322139/spinner-end-hitsound.zip)

**osu!lazer version:** 2020.1002.0

**Logs:** [logs.zip](https://github.com/ppy/osu/files/5322094/logs.zip)

bdach:
Resolves #10354 and a bunch of theoretical ones (although I haven't checked every single one).

PRing in a somewhat haphazard state for potential hotfix purposes. Tested... rudimentarily.

# Summary

With the scoring unification changes at #10288 the miss result was split into multiple (`Miss`, `LargeTickMiss`, `SmallTickMiss` and `IgnoreMiss`). Unfortunately, there were a lot of places that relied on `Miss` being the only and single hit result that indicates a miss.

This PR replaces 25 references to `HitResult.Miss`. They are of two main categories:

* comparison with `HitResult.Miss` to ascertain hit state - replaced by access to `HitResult.IsHit()` or even directly `JudgementResult.IsHit`,
* assignment of `HitResult.Miss` to apply a miss result - replaced by assignment of `JudgementResult.Judgement.MinResult`.

The remaining usages, save for singular exceptions, are:

* test code,
* anything that touches `HitWindows`.

# Remarks

Test coverage is not provided here, since if I wanted to comprehensively test this, I'd probably be easily able to rack up 10 PRs or more. The miss change was so fundamental that the tests would have to be *really* paranoid to catch this out earlier.

There's one weird change in `DrawableHitObject`, namely a pattern match with `when` to avoid duplicating the conditions in `HitResult.IsHit()`. Not sure how you'll see that one.

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -511,7 +511,7 @@ namespace osu.Game.Rulesets.Objects.Drawables
                 case HitResult.None:
                     break;
 
-                case HitResult.Miss:
+                case { } result when !result.IsHit():
                     updateState(ArmedState.Miss);
                     break;
 
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -506,19 +506,8 @@ namespace osu.Game.Rulesets.Objects.Drawables
 
             Result.TimeOffset = Math.Min(HitObject.HitWindows.WindowFor(HitResult.Miss), Time.Current - endTime);
 
-            switch (Result.Type)
-            {
-                case HitResult.None:
-                    break;
-
-                case { } result when !result.IsHit():
-                    updateState(ArmedState.Miss);
-                    break;
-
-                default:
-                    updateState(ArmedState.Hit);
-                    break;
-            }
+            if (Result.HasResult)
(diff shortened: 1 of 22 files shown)


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -511,7 +511,7 @@ protected void ApplyResult(Action<JudgementResult> application)
                 case HitResult.None:
                     break;
 
-                case HitResult.Miss:
+                case { } result when !result.IsHit():
```
peppy, line 514: does this seem okay as a replacement for this switch?

```csharp
    if (Result.Type != HitResult.None)
        updateState(Result.Type.IsHit() ? ArmedState.Hit : ArmedState.Miss);
```

peppy (changes requested):

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
bdach: It should be, yeah. I can push that version in a bit.

peppy (approved):
looks correct as far as i can tell

PR #10348, opened 2020-10-03
Allow rotation lock on Android to function properly

tytydraco:
According to Google's documentation, `fullSensor` will ignore rotation locking preferences, while `fullUser` will obey them.

See this issue for more details:
https://github.com/ppy/osu/pull/10347


tytydraco:
https://developer.android.com/guide/topics/manifest/activity-element

peppy (approved):
Look sane.

PR #10349, opened 2020-10-03
Bring back health increase/decrease values closer to previous ones

### LINKED ISSUE #10337: osu! HP drain too harsh after scoring unification changes
**Describe the bug:** 
I was playing some beatmaps and when there's a waiting section in the beatmap the HP bar drains faster than it should be and I lose. I played the beatmaps on osu!stable, there's no problem at all.

**osu!lazer version:** 2020.1002.0

**Logs:**

[logs.zip](https://github.com/ppy/osu/files/5319435/logs.zip)

bdach:
Aims to resolve #10337 "in hotfix mode", so to speak. Not tested very thoroughly in actual gameplay, so buyer beware.

# Summary

There are several issues at play here, so I lumped them into one PR to reduce noise. Taking it from the top:

* In osu!:
  - Slider ends gave no health. To make them give health, they've been changed to be small ticks (can't be large ticks, as that re-introduces sliderend combo break)
  - The only hit result that gave health was GREAT. Both OK and MEH took away health. The old values are restored here.
  - Health increases for spinner ticks have also been decreased significantly; this brings them back up again.
* In taiko:
  - taiko rolls its own HP system, so it was largely unaffected by the changes, there was however one missed reference to `Great` that should have been `SmallTickHit` (see 682b5fb).
* In catch:
  - Droplet health has been brought closer to previous values.
* In mania:
  - The new anti-mashing HP changes from #10291, as well as the new HP increases for ticks have been moved locally to `ManiaJudgement`.

Here's a [spreadsheet](https://docs.google.com/spreadsheets/d/1fM6jCZg3CmZFdqHQeS7d4PKhgqpMLa7plxmeEPfTaTk/edit#gid=0) that aims to provide a comparison of what changed where.

# Remarks

OK and GOOD are slightly different than they were before. Because now 100s are OKs and not GOODs, and GOODs gave 50%, now OKs give 50% and GOODs give the linearly-interpolated 75%. I don't think it should realistically matter for any of our scenarios, as the only ruleset that uses GOODs (mania) has a local override.

### DIFF
--- a/osu.Game.Rulesets.Mania/Judgements/ManiaJudgement.cs
+++ b/osu.Game.Rulesets.Mania/Judgements/ManiaJudgement.cs
@@ -2,10 +2,40 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using osu.Game.Rulesets.Judgements;
+using osu.Game.Rulesets.Scoring;
 
 namespace osu.Game.Rulesets.Mania.Judgements
 {
     public class ManiaJudgement : Judgement
     {
+        protected override double HealthIncreaseFor(HitResult result)
+        {
+            switch (result)
+            {
+                case HitResult.LargeTickHit:
+                    return DEFAULT_MAX_HEALTH_INCREASE * 0.1;
+
+                case HitResult.LargeTickMiss:
+                    return -DEFAULT_MAX_HEALTH_INCREASE * 0.1;
+
+                case HitResult.Meh:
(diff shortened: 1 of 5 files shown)


peppy (approved):
Tested on 10 odd beatmaps that I am pretty familiar with. Feels a lot closer to what I'd expect.

peppy:
Going to get this in for now. Further changes can come as follow-up PRs when required 👍 .

PR #10359, opened 2020-10-04
Fixed metadata text not aligning correctly

### LINKED ISSUE #4506: Source text not aligned to middle while loading a beatmap.
**Issue:** I don’t think I should be opening this issue but isn’t the source text supposed to be aligned automatically in the middle if the beatmap has a source?
**Version:** _2019.321.0_
**Image:**
GitHawk Upload by dscnnctd

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

willyao94:
Fixes https://github.com/ppy/osu/issues/4506.

Use a single TextFlowContainer to center align the metadata on beatmap load.

Output from maps mentioned in the #4506 :
blooming
Neo-Aspect
Jojo

### DIFF
--- a/osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
+++ b/osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
@@ -23,32 +23,6 @@ namespace osu.Game.Screens.Play
     /// </summary>
     public class BeatmapMetadataDisplay : Container
     {
-        private class MetadataLine : Container
-        {
-            public MetadataLine(string left, string right)
-            {
-                AutoSizeAxes = Axes.Both;
-                Children = new Drawable[]
-                {
-                    new OsuSpriteText
-                    {
-                        Anchor = Anchor.TopCentre,
-                        Origin = Anchor.TopRight,
-                        Margin = new MarginPadding { Right = 5 },
-                        Colour = OsuColour.Gray(0.8f),
-                        Text = left,
-                    },
-                    new OsuSpriteText
-                    {
-                        Anchor = Anchor.TopCentre,
-                        Origin = Anchor.TopLeft,
-                        Margin = new MarginPadding { Left = 5 },
-                        Text = string.IsNullOrEmpty(right) ? @"-" : right,
-                    }
-                };
-            }
-        }
-
         private readonly WorkingBeatmap beatmap;
         private readonly Bindable<IReadOnlyList<Mod>> mods;
         private readonly Drawable facade;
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
```diff
@@ -167,6 +138,20 @@ private void load()
...
+                t.Padding = new MarginPadding { Right = 10 };
+            });
+            metadataLine.AddText(string.IsNullOrEmpty(metadata.Source) ? @"-" : metadata.Source);
+
+            metadataLine.AddParagraph("Mapper", t =>
+            {
+                t.Colour = OsuColour.Gray(0.8f);
+                t.Padding = new MarginPadding { Right = 10 };
+            });
+            metadataLine.AddText(string.IsNullOrEmpty(metadata.AuthorString) ? @"-" : metadata.AuthorString);
```
bdach, line 153: This solution changes the design. Previously the "Source" and "Layout" texts were right-aligned with each other, and now they aren't.

From the screenshot you've provided:

2020-10-04-024047_529x364_scrot


Additionally, as stated in the issue thread, we would probably be looking for enabling text wrapping wherever possible, but I guess that's a minor issue, relatively speaking.

bdach (changes requested):

osu.Game/Screens/Play/BeatmapMetadataDisplay.cs
willyao94: Sorry, based on dscnnctd's screenshot on 12 Nov 2019, I had assumed the design change was approved.
Just to clarify, which do you mean the first or second one?


Note that the second pic ran with a small hack prepending a space in front of Source so that it seems aligned.
I don't suppose that would get approved?

Regarding the max width and wrapping, what should the max width be?
I did a quick test to set a max width of 50 and it resulted in this which is clearly incorrect.

I think it's caused by the MultilineTextBuilder using the default_never_fixed_width_characters in SpriteText?

bdach: Adding a space is indeed not correct. It relies on particular font and string used and will break with localisation.

As for the wrapping, there's a reference screenshot in the issue showing about how wide the text should be.

Make sure you're setting the max width of the actual fill flow - from what you have on the screenshot it seems to be that you might be incorrectly passing the with to sprite text creation parameters, although I can't say for sure without seeing the code.

PR #10370, opened 2020-10-05
Ensure virtual track time is long enough for test beatmaps

peppy:
This has come up before, but in my upcoming test addition I ran into a case where seeking too far into a beatmap (while there's still hitobjects) causes `MusicController` to get into an expensive loop and bring the tests to a halt for a few seconds. This should give better results.

- [x] This breaks mania out-of-order tests for unknown reasons, needs further investigation.

This doesn't fix the fact that music controller is still firing a fucktonne of next track events on screen exiting, or hitting the next end of track. Need to open an issue for that or something (wip fix branch [here](https://github.com/ppy/osu/compare/master...peppy:fix-test-music-controller-next?expand=1) but not really a fix yet).


bdach:
I know what's causing the test failure, but... it's quite a doozy to unpack.

* First off, the result of the test is inconsistent with how much you pad the end time by. So, for example, if you pad it by 1000 instead of 2000, you'll get a failure on a different judgement than with 2000.
* Having seen that, I suspected replay playback issues, or frame alignment issues. Upon verifying I confirmed that the replay actions were all applied correctly, *but* at some point the replay was getting rewound back to start.
* More importantly, the replay was being rewound back to start *while the asserts were checking the judgements*, which means that the rewind unapplied results from under the test. This is the immediate cause of the failure.
* As to why the replay was being rewound - the virtual track stopped correctly. Upon its end, `TrackVirtualManual` raises track completion:

  https://github.com/ppy/osu/blob/3630aa30d4e3cae2ee6a3e739d9169a1cb10a144/osu.Game/Tests/Visual/OsuTestScene.cs#L339-L343

  which then in turn fires `MusicController.onTrackCompleted`:

  https://github.com/ppy/osu/blob/3630aa30d4e3cae2ee6a3e739d9169a1cb10a144/osu.Game/Overlays/MusicController.cs#L384-L385

  The track in question is neither looping, nor is the beatmap disabled, therefore the track changes from under `Player`, changing `Time` back to 0 and triggering the entire chain of consequences.

peppy:
I also suspected the rewinding thing (was already looking at a fix for that) but the weird one I noticed was a miss on hitobject index 2 (well, the second result in the local list), but perfect on the rest. Did you ever see this come up in your testing? That's what made me think it *wasn't* related to rewinding / next track breakage.

bdach:
Oh yeah, that's actually yet another thing - judgement number 1 and 2 are actually reversed on the list that they get collected to. If you check `HitObject.StartTime` on them, it goes 1100, 1000, 1200, 1300, 1400, 1500, 1600.

That's because the second hit triggers the first miss explicitly, so they get stored in the order that the callbacks fire in. It's not an error, though - the way that the judgements are retrieved (`SingleOrDefault` and comparison over `HitObject`) ensures that there's no mix-up later on.

smoogipoo:
I'll just point out that that ordering technically incorrect as it means the combo count is one increment behind what I'd expect after that - if a future note is hit combo will be 0 rather than 1.

peppy:
In order to get this merged, I've updated to a less destructive version (which still ensures the track is at least 60s). In testing this seems to fix the test scene that was breaking, and also alleviates issues which are still plaguing tests (ran into this issue causing autoplay tests to run <1fps in the test browser, meaning I couldn't actually investigate an underlying issue).

smoogipoo (approved):

PR #10371, opened 2020-10-05
Fix seeking in replays not correctly pausing samples

peppy:
This was a complete oversight. Wasn't being cached and wasn't being updated correctly with the new bindable flow. I've removed the `IsSeeking` as it was only ever used for this purpose, replacing with a common method that should make it easier to understand its purpose in the future.

Comes with test coverage (slightly more comprehensive compared to editor version).

### DIFF
--- a/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
+++ b/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
@@ -59,13 +59,16 @@ namespace osu.Game.Rulesets.UI
         private int direction;
 
         [BackgroundDependencyLoader(true)]
-        private void load(GameplayClock clock)
+        private void load(GameplayClock clock, ISamplePlaybackDisabler sampleDisabler)
         {
             if (clock != null)
             {
                 parentGameplayClock = stabilityGameplayClock.ParentGameplayClock = clock;
                 GameplayClock.IsPaused.BindTo(clock.IsPaused);
             }
+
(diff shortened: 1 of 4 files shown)


peppy:
This is broken due to caching precedence (editor is not getting the correct pause state due to local caching added in `FrameStabilityContainer`).

peppy:
I've fixed this in the only simple way I can think of. Hopefully we can agree on going with this short term to fix the issue. Open to other suggestions that don't involve changing the hierarchy massively.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -207,11 +211,15 @@ private void updateClock()
 
         private void setClock()
         {
-            // in case a parent gameplay clock isn't available, just use the parent clock.
-            parentGameplayClock ??= Clock;
-
-            Clock = GameplayClock;
-            ProcessCustomClock = false;
```
smoogipoo, line 214: Sure you want to omit this?

Are these changes still required at all?

smoogipoo (changes requested):

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
peppy: yes, see the ProcessFrame method.

smoogipoo (approved):

PR #10373, opened 2020-10-05
Don't stop non-looping samples immediately when pausing

### LINKED ISSUE #10358: Hitsounds cut out abruptly at hitobject end
So, basically the problem is that no matter what skin i use or what map i play the hitsounds dont work as intended, the audio cuts when I hit a note so it's very annoying. I never had this kinda error before and I'm a casual player.

I's not the same problems other players had, like the one where the hitsound plays but its the default one, in this case its the skin hitsound that tries to play when you hit a single note but for some reason it only plays the start and then cuts to silence, and in some cases it doesn't even play a sound. The only notes not affected are the long notes who plays the normal hitsound from the skin you are using.
If someone has this issue pls let me know so I can know its not something that affects my pc.

I tried recording my problem but i'm having problems with the recording software so i'm trying to explain my problem.

osu! version: Lazer 2020.1002.0

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5322436/runtime.log)
[session.log](https://github.com/ppy/osu/files/5322437/session.log)

peppy:
Opted to go with a simple / reliable fix for this one: we only pause looping samples now. All other samples will be allowed to play out. This actually feels better in most every case, and I believe is also how stable handles things.

- Resolves #10358
- [x] Update tests

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -109,9 +109,9 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             }
         }
 
-        public override void StopAllSamples()
+        public override void StopLoopingSamples()
         {
-            base.StopAllSamples();
+            base.StopLoopingSamples();
             slidingSample?.Stop();
         }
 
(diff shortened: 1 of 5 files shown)


smoogipoo (dismissed):
I guess this is fine? Not a fan of this method changing name...

peppy:
I'll rename back to avoid api breaking change if you'd prefer.

smoogipoo:
May be best to

smoogipoo (approved):

PR #10384, opened 2020-10-06
Fix a couple of new rider inspections

peppy:
The default parameter one doesn't correctly resolve even after updating it to match the base type, but might as well make them match.

### DIFF
--- a/.idea/.idea.osu.Desktop/.idea/modules.xml
+++ b/.idea/.idea.osu.Desktop/.idea/modules.xml
@@ -2,7 +2,7 @@
 <project version="4">
   <component name="ProjectModuleManager">
     <modules>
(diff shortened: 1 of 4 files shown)


smoogipoo (dismissed):

peppy:
Oh no.

bdach:
So I wanted to look at what the inspections are, but I'm not seeing anything on `master`, even though I just updated rider... What version are you getting them on? I'm currently on

```
JetBrains Rider 2020.2.4
Build #RD-202.7660.16, built on September 30, 2020
(snip)
Runtime version: 11.0.8+10-b944.34 amd64
VM: OpenJDK 64-Bit Server VM by JetBrains s.r.o.
.NET Core 3.1.6
Linux 4.15.0-118-generic
```

peppy:
These were on EAP. The failing one has been reported [here](https://youtrack.jetbrains.com/issue/RSRP-481535)

bdach:
Ah, I see... I guess we wait this one out then.

peppy:
Both of these are fixed in EAP3 and no longer require adjustments. Looks like they may have just rolled back the r# version for now. Will close for now.

PR #10390, opened 2020-10-06
Split editor setup screen up into sections

peppy:
- [x] Depends on #10388 to avoid very bad merge conflicts.

Unlike the [design](https://user-images.githubusercontent.com/1329837/42362632-3042c92e-812f-11e8-8ee8-85ad31f83745.png) implies, I believe we want all the beatmap settings to be in a single scroll view like the user profile content.

I started to look at sharing implementation with `UserProfileOverlay` to get the tab control displaying but this requires some abstraction work (it's currently very local to that class). Definitely worth putting in the effort to abstract this as it is used in multiple other places on osu-web already.

### DIFF
--- a/osu.Game/Screens/Edit/Setup/SetupSection.cs
+++ b/osu.Game/Screens/Edit/Setup/SetupSection.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
@@ -14,7 +13,7 @@ namespace osu.Game.Screens.Edit.Setup
 {
     internal class SetupSection : Container
     {
-        protected FillFlowContainer Flow;
+        private readonly FillFlowContainer flow;
 
         [Resolved]
         protected OsuColour Colours { get; private set; }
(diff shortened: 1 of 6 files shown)


osu.Game/Screens/Edit/Setup/SetupSection.cs
```diff
@@ -0,0 +1,43 @@
...
+using osu.Framework.Graphics.Containers;
+using osu.Game.Beatmaps;
+using osu.Game.Graphics;
+using osuTK;
+
+namespace osu.Game.Screens.Edit.Setup
+{
+    internal class SetupSection : Container
+    {
+        protected FillFlowContainer Flow;
```
smoogipoo, line 17:
```suggestion
        protected FillFlowContainer Flow { get; private set; }
```

osu.Game/Screens/Edit/Setup/SetupSection.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        [Resolved]
+        protected OsuColour Colours { get; private set; }
+
+        [Resolved]
+        protected IBindable<WorkingBeatmap> Beatmap { get; private set; }
+
+        public override void Add(Drawable drawable) => throw new InvalidOperationException("Use Flow.Add instead");
```
smoogipoo, line 25: Is there a reason why this is not set as override of `Content`?

smoogipoo (changes requested):

osu.Game/Screens/Edit/Setup/SetupSection.cs
peppy: FillFlowContainer ain't a Container :(

smoogipoo: Oh, you might need to use `FillFlowContainer<Drawable>`. I'm 99% sure this works.

smoogipoo: Actually no, am I misunderstanding something? It should just work by the looks of it:
image

peppy: Weird, it does work. I may have changed the type since I determined it didn't.

smoogipoo (approved):

PR #10394, opened 2020-10-06
Add obsoleted Bindable property back to SettingsItem for compatibility

peppy:
@smoogipoo please test it works with rulesets as expected, i don't have any handy.


smoogipoo (approved):
Looks to work :)

PR #10399, opened 2020-10-06
Add editor beatmap batch change support

peppy:


osu.Game/Screens/Edit/EditorBeatmap.cs
```diff
@@ -180,22 +203,62 @@ public void RemoveAt(int index)
...
+            if (isBatchApplying)
+                throw new InvalidOperationException("Attempting to perform a batch application from within an existing batch");
+
+            isBatchApplying = true;
+
+            applyFunction(this);
+
             beatmapProcessor?.PreProcess();
-            processHitObject(hitObject);
             beatmapProcessor?.PostProcess();
```
bdach, line 237: It looks to me like `PostProcess` ought to be called after processing the batch inserts and deletes, right? Can't seem to find a way to break it, though...

peppy: Ah, nice catch. I think this won't be visible in our current rulesets as we don't use this flow, but does need to happen that way around I believe.

smoogipoo (approved):

PR #10400, opened 2020-10-06
Ensure first hitobject is always a new combo

### LINKED ISSUE #10372: Moving second object before the first one in the timeline causing unexpected behavior
anim

peppy:
- [x] Depends on #10399.
- Closes https://github.com/ppy/osu/issues/10372.


bdach:
I guess this does restore sanity, but it's still a little unintuitive how dragging an object from the same combo as the first one before the first one will split the dragged object onto a separate combo (because the previously-first-and-now-second hitobject's new combo flag is not cleared).

peppy:
A bit, yeah. Can't think of a super-simple way of avoiding that right now.

smoogipoo (approved):
I think this is fine for now.

PR #10401, opened 2020-10-06
Expand ScoreProcessor tests

### LINKED ISSUE #10318: ScoreProcessor is missing tests for all scoring relationships
`ScoreProcessorTest` should ensure that all relationships defined in https://github.com/ppy/osu/wiki/Scoring#hit-results are satisfied - that the correct score portions are affected and the correct `MinResult` is set for each corresponding `MaxResult`.

Additionally, the `HitResultExtensions` methods should also be tested.

LittleEndu:
Closes #10318

The basis of testing how score portion works is if we hit 3 out of 4 objects while achieving max combo of 2, the portions effected are different enough that the testing is meaningful but still easy to manually calculate. 

Tests for SmallTickHit are probably wrong, but I don't know how they are supposed to act in game modes where they are from (taiko and fruits) because I've not played these. [Wiki](https://github.com/ppy/osu/wiki/Scoring#hit-results) says they affect accuracy so I expected them to get score in Standard scoring but they don't. They don't get score in Classic scoring because we have a combo of 0 in this test.


osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,193 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
         }
+
+        [TestCase(ScoringMode.Standardised, HitResult.Miss, HitResult.Great, 0)]
+        [TestCase(ScoringMode.Standardised, HitResult.Meh, HitResult.Great, 387_500)]
+        [TestCase(ScoringMode.Standardised, HitResult.Ok, HitResult.Great, 425_000)]
+        [TestCase(ScoringMode.Standardised, HitResult.Good, HitResult.Perfect, 3_350_000 / 7.0)]
+        [TestCase(ScoringMode.Standardised, HitResult.Great, HitResult.Great, 575_000)]
+        [TestCase(ScoringMode.Standardised, HitResult.Perfect, HitResult.Perfect, 575_000)]
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickMiss, HitResult.SmallTickHit, 0)]
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickHit, HitResult.SmallTickHit, 0)] // TODO: this should probably be 225_000 from accuracy portion
```
bdach, line 64: because you don't seem to have looked into this in any capacity, I will explain why this is passing (probably wrongly). this is due to the following conditional in `ScoreProcessor`.

https://github.com/ppy/osu/blob/c803be2dd260ce5d1963a53c08d9f48bbc9c8d35/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L271-L275

because the maximum combo is 0 due to all judgements not affecting combo, the scoring mode is forced to `Classic`, which means that the score will be always 0 if there are no bonuses. mind you, that is probably wrong, but I don't think it's acceptable to submit this test without looking into why this is not working correctly at all.

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,193 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
+        [TestCase(HitResult.Great, HitResult.Miss)]
+        [TestCase(HitResult.Perfect, HitResult.Miss)]
+        [TestCase(HitResult.SmallTickHit, HitResult.SmallTickMiss)]
+        [TestCase(HitResult.LargeTickHit, HitResult.LargeTickMiss)]
+        [TestCase(HitResult.SmallBonus, HitResult.IgnoreMiss)]
+        [TestCase(HitResult.LargeBonus, HitResult.IgnoreMiss)]
+        public void TestMinResults(HitResult hitResult, HitResult expectedMinResult)
+        {
+            var result = new JudgementResult(new HitObject(), new TestJudgement(hitResult));
+            Assert.IsTrue(result.Judgement.MinResult == expectedMinResult);
```
bdach, line 117: use `Assert.Equals` for these (this goes for all of the other tests as well)

bdach (dismissed):

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,193 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
+        [TestCase(ScoringMode.Classic, HitResult.Perfect, HitResult.Perfect, 936)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickMiss, HitResult.SmallTickHit, 0)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickHit, HitResult.SmallTickHit, 0)]
+        [TestCase(ScoringMode.Classic, HitResult.LargeTickMiss, HitResult.LargeTickHit, 0)]
+        [TestCase(ScoringMode.Classic, HitResult.LargeTickHit, HitResult.LargeTickHit, 936)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallBonus, HitResult.SmallBonus, 30)]
+        [TestCase(ScoringMode.Classic, HitResult.LargeBonus, HitResult.LargeBonus, 150)]
+        public void TestFourVariousResultsOneMiss(ScoringMode scoringMode, HitResult hitResult, HitResult maxResult, double expectedScore)
+        {
+            var minResult = new JudgementResult(new HitObject(), new TestJudgement(hitResult)).Judgement.MinResult;
```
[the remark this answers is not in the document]
LittleEndu: Same for the MinResult test?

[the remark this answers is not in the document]
bdach: obviously, yes

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,222 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
+            }
+
+            Assert.IsTrue(Precision.AlmostEquals(expectedScore, scoreProcessor.TotalScore.Value));
+        }
+
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickHit, 6_850_000 / 7.0)]
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickMiss, 6_400_000 / 7.0)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickHit, 1950 / 7.0)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickMiss, 1500 / 7.0)]
+        public void TestSmallTicksAccuracy(ScoringMode scoringMode, HitResult hitResult, double expectedScore)
```
bdach, line 104: i don't think fixing the issue with the old test by swapping it for a different test is a proper resolution. it's skirting the problem. i think we should fix the scoring issue instead as it's clearly unintuitive behaviour.

however, to confirm for sure - @peppy, @smoogipoo, what are your views on this?

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
smoogipoo: I know there was a reason for that at some point... But I think the relevant codepath is properly handled now:
https://github.com/ppy/osu/blob/f206ea96ce6ccbce87af0e360b0f2630ceac0635/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L196-L202

Could the forcing of `Classic` just be removed?

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
smoogipoo: So I guess there's really two issues here - the first being that `Classic` scoring is forced which shouldn't be required anymore (see my other comment above). The second is that classic scoring gives 0 score if there's 0 combo, which should definitely not be the case.

https://github.com/ppy/osu/blob/f206ea96ce6ccbce87af0e360b0f2630ceac0635/osu.Game/Rulesets/Scoring/ScoreProcessor.cs#L224-L226
This should probably be something like `(accuracyRatio * (maxCombo + 1) * 300)`

LittleEndu: But should standard scoring award combo portion score if the combo is 0 achieved out of 0 max combo?

LittleEndu: Or it might make sense to have accuracy portion be 100% if max achievable combo is 0

smoogipoo: No. As it is is intended - combo should be fully awarded if max combo is 0.

Scoring should have zero edge cases that require special casing.

LittleEndu: I changed comboRatio from being 0 if maxHighestCombo is 0 to being 1. Then all the combo portion score is awarded when the map has 0 combo FC

osu.Game/Rulesets/Scoring/ScoreProcessor.cs
```diff
@@ -197,7 +197,7 @@ private double getScore(ScoringMode mode)
         {
             return GetScore(mode, maxHighestCombo,
                 maxBaseScore > 0 ? baseScore / maxBaseScore : 0,
-                maxHighestCombo > 0 ? (double)HighestCombo.Value / maxHighestCombo : 0,
+                maxHighestCombo > 0 ? (double)HighestCombo.Value / maxHighestCombo : 1,
```
LittleEndu, line 200: ``maxHighestCombo`` is a bad name for this value. Should be something like ``maxAchievableCombo`` or should be documented as such

bdach: is there something stopping you from doing either?

LittleEndu: Because I was unsure if this is what this variable stores exactly because it's undocumented. If I was already changing names, I would probably change ``Reset(bool storeResults)`` to ``Reset(bool storePerfectResults)`` as-well, because it seems to only be used after autoplay to get all the maximum values.

bdach: well, you've already inferred what the variable is by suggesting the rename, right?

also, disagree on the `Reset` rename. the fact that it's used to store maximum for autoplay is incidental.

LittleEndu: Then there needs to be some other way to get maximum achievable combo after applying a beatmap. The ``reset`` method is only ever used in there, in ``ApplyBeatmap``, though, exactly for getting all the maximum values by simulating a perfect play.

bdach: why?

LittleEndu: > the fact that it's used to store maximum for autoplay is incidental

it doesn't look incidental do me. It looks like it's the only thing it's used for (I only see reset used when beatmap is applied). If the storing of maximum score is indeed incidental, *then* there needs to be another way to store maximum values.

bdach: I meant "incidental" in the deeper sense that it doesn't *matter* why the score processor is storing results. It just can, for future use. There's no reason to make a bigger deal out of it than it needs to be.

LittleEndu: I get that, but the fact remains that after the results are stored they are expected to be stored from a perfect simulation. If something else were to call the reset function to store results that aren't perfect, score calculation for that ScoreProcessor would be inaccurate

bdach: I don't really want to continue belaboring this point as it is insignificant in my view. One look at the body of `Reset()` says what the parameter does.

If we were to *really* care that nobody should be able to store non-perfect results then the entire calculation of perfect results should be encapsulated entirely. You're welcome to try to implement that, if you care. I kind of don't - there's hundreds of more important issues open right now. Until this becomes an issue, we can deal with what we have.

bdach:
FWIW I've made slight changes and expanded coverage slightly - think this is good to go but don't want to approve & merge myself entirely as I pushed.

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,237 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
 
             Assert.IsTrue(Precision.AlmostEquals(expectedScore, scoreProcessor.TotalScore.Value));
         }
+
+        [TestCase(ScoringMode.Standardised, HitResult.Miss, HitResult.Great, 0)]
+        [TestCase(ScoringMode.Standardised, HitResult.Meh, HitResult.Great, 387_500)]
+        [TestCase(ScoringMode.Standardised, HitResult.Ok, HitResult.Great, 425_000)]
+        [TestCase(ScoringMode.Standardised, HitResult.Good, HitResult.Perfect, 3_350_000 / 7.0)]
```
peppy, line 60: where does the 7.0 come from? should likely be moved out to a `const`

peppy (dismissed):

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
peppy: Would be nice to have an inline comment (xmldoc on the method?) explaining what these tests are doing, and where the numbers come from. Will be helpful when we inevitably break them and need to update values.

LittleEndu: 7.0 is for the fractional portion. Thoguth it looked easier on the eye, could easily be replaced with double. Also the reason why the expected score is double in this test but int in the test that already existed

bdach: it's because the test is using perfect as the max judgement, which is 350 - I guess the intent was to avoid specifying inexact values (100/350 does not divide well in base 10). I figured it was tolerable but do agree that it should probably just be a whole number rounded to the nearest unit here, and the comparison should use half precision.

peppy: it's this one right?

image

should probably expose that as a public const and consume it here

bdach: Not exactly, it's the accuracy. Three OK judgements if a PERFECT is the max result gives 300/1400 = 0.2142857... which has no finite decimal expansion.

bdach:
I've reworded the xmldoc a touch. Would appreciate a check if this is readable and grokkable now.

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,257 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
+        /// </summary>
+        /// <param name="scoringMode">Scoring mode to test.</param>
+        /// <param name="hitResult">The <see cref="HitResult"/> that will be applied to selected hit objects.</param>
+        /// <param name="maxResult">The maximum <see cref="HitResult"/> achievable.</param>
+        /// <param name="expectedScore">Expected score after all objects have been judged, rounded to the nearest integer.</param>
+        /// <remarks>
+        /// This test intentionally misses the 3rd hitobject to achieve lower than 75% accuracy and exactly 50% max combo.
+        /// <para>
+        /// For standardised scoring, <paramref name="expectedScore"/> is calculated using the following formula:
+        /// 1_000_000 * ((3 * <paramref name="hitResult"/> / 4 * <paramref name="maxResult"/>) * 30% + 50% * 70%)
```
smoogipoo, line 68:
```suggestion
        /// 1_000_000 * ((3 * <paramref name="hitResult"/> / (4 * <paramref name="maxResult"/>)) * 30% + 50% * 70%)
```

But really, I don't really like all of this maths being thrown out as an exercise for the user to calculate... What do you think about putting inline comments instead:
```
[TestCase(ScoringMode.Standardised, HitResult.Miss, HitResult.Great, 0)]
[TestCase(ScoringMode.Standardised, HitResult.Meh, HitResult.Great, 387_500)] // 50/300 * 3/4 * 300000 + 350000
[TestCase(ScoringMode.Standardised, HitResult.Ok, HitResult.Great, 425_000)] // 100/300 * 3/4 * 300000 + 350000
[TestCase(ScoringMode.Classic, HitResult.Miss, HitResult.Great, 0)]
[TestCase(ScoringMode.Classic, HitResult.Meh, HitResult.Great, 156)] // 50 * 3/4 * 4 * (1 + 1/25)
[TestCase(ScoringMode.Classic, HitResult.Ok, HitResult.Great, 312)] // 100 * 3/4 * 4 * (1 + 1/25)
```
At least that way it makes it really obvious as to what's changing and can be easily verified without digging up the hitresult values.

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
```diff
@@ -53,5 +53,257 @@ public void TestSingleOsuHit(ScoringMode scoringMode, HitResult hitResult, int e
...
+            }
+
+            Assert.IsTrue(Precision.AlmostEquals(expectedScore, scoreProcessor.TotalScore.Value, 0.5));
+        }
+
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickHit, 978_571)]
+        [TestCase(ScoringMode.Standardised, HitResult.SmallTickMiss, 914_286)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickHit, 279)]
+        [TestCase(ScoringMode.Classic, HitResult.SmallTickMiss, 214)]
+        public void TestSmallTicksAccuracy(ScoringMode scoringMode, HitResult hitResult, int expectedScore)
```
smoogipoo, line 128: How does this test work? Should I be reading the test? Is this test still required? I thought the reason for this test had to do with classic scoring being broken for `maxAchievableCombo = 0`.

smoogipoo (changes requested):

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
LittleEndu: This test isn't required anymore but it still works and tests behaviour that's special to small ticks (that they only affect accuracy portion)

Probably needs documentation similar to previous test

osu.Game.Tests/Rulesets/Scoring/ScoreProcessorTest.cs
LittleEndu: > instead

Why not both?

smoogipoo: Sure.

smoogipoo (approved):
Looks fine to me!

PR #10416, opened 2020-10-08
Add local per-hitobject overrides for Sample/Difficulty control points

peppy:
This is a proof-of-concept showing the general structure for review/feedback.

If it's agreed upon tests and documentation will be improved, along with a second pass checking auxiliary usages (beatmap converters etc.)


bdach:
When it comes to samples, I don't think I can see much of a problem with this approach as they're pretty localised and nobody else but the hitobject itself should ever really need to query it. On that note I can't say I have very good foresight with these things, however, so I might well be horribly wrong.

Difficulty control points will definitely be more... interesting, though. If the future direction is to allow them universally instead of just in osu!, we should probably consider/try adding support for the overrides to a scrolling ruleset to see what issues might fall out in the future. It might lead to quite a bit of the scrolling algorithm stuff changing, although as long as overlaps are disallowed it should be relatively easy to convert the system with overrides to just a sequence of normal difficulty control points by fudging some interval math. I'm also not sure how per-object difficulty would apply or feel in mania, for example.

If it's just for osu!, though, then this current layout is probably easiest to work with as individual objects are pretty independent of each other there.

smoogipoo:
I still fundamentally disagree with per-hitobject difficulty points. I'm also pretty strongly against per-hitobject sample points - all sample points do is provide a bank + volume which hitobject samples already have?

I really think it should be done on a per-case basis. If you need velocity for sliders in Rulesets.Osu, do it in Rulesets.Osu. Expose some interface like `IHasVelocity` and switch the editor's functionality based on that, maybe, because it also makes no sense for this to be hard-tied to `IHasRepeats` out of all things.

peppy:
for samples: the idea would be to move all the weird sample stuff out, and use this instead.

i'm not sure what you mean but "do it per ruleset". are you suggesting that there should be a second level of speed multiplier override specified per ruleset? to me, speed multiplier is a common concept across all rulesets and adding an extra level of override means things like the editor can't share the benefits of this.

smoogipoo:
Yes, a second level of override is exactly what I'm suggesting.

I would agree that first level override (difficulty points) are common concepts across rulesets, but I disagree that second level overrides are as common. If you want this to be the case then it needs to be supported in scrolling container at the least, and there needs to be suitable changes made for bar lines in both mania and taiko.  
If that can't be done or is difficult to do, then I don't consider this to be a common concept.

I don't see why having an interface for the second-level override makes it hard for the editor to share the benefits of this?

peppy:
So OsuHitObject would have a VelocityOverride variable in your vision? And we would keep using the ugly stable method of storing samples haphazardly rather than using the `SampleControlPoint` structure?

smoogipoo:
But `SampleControlPoint` isn't going to be enough to handle hitobject samples. What are you going to do at the point where you need normal/whistle/finish/clap with individual sound banks/volumes for each? Store a list of `SampleControlPoint`s?

osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -104,10 +123,12 @@ public HitObject()
         /// <param name="cancellationToken">The cancellation token.</param>
         public void ApplyDefaults(ControlPointInfo controlPointInfo, BeatmapDifficulty difficulty, CancellationToken cancellationToken = default)
         {
+            DifficultyControlPoint = LocalDifficultyControlPoint ?? controlPointInfo.DifficultyPointAt(StartTime);
+
             ApplyDefaultsToSelf(controlPointInfo, difficulty);
 
             // This is done here since ApplyDefaultsToSelf may be used to determine the end time
-            SampleControlPoint = controlPointInfo.SamplePointAt(this.GetEndTime() + control_point_leniency);
+            SampleControlPoint = LocalSampleControlPoint ?? controlPointInfo.SamplePointAt(this.GetEndTime() + control_point_leniency);
```
peppy, line 131: @smoogipoo I tried replacing this with your recommendation:

```csharp
        public void ApplyDefaults(ControlPointInfo controlPointInfo, BeatmapDifficulty difficulty, CancellationToken cancellationToken = default)
        {
            updateDifficultyControlpoint(controlPointInfo);
            updateSampleControlPoint(controlPointInfo);
        }

        private void updateSampleControlPoint(ControlPointInfo controlPointInfo)
        {
            if (SampleControlPoint == beatmapSampleControlPoint)
                SampleControlPoint = null;

            beatmapSampleControlPoint = controlPointInfo.SamplePointAt(this.GetEndTime() + control_point_leniency);

            SampleControlPoint ??= beatmapSampleControlPoint;
        }

        private void updateDifficultyControlpoint(ControlPointInfo controlPointInfo)
        {
            if (DifficultyControlPoint == beatmapDifficultyControlPoint)
                DifficultyControlPoint = null;

            beatmapDifficultyControlPoint = controlPointInfo.DifficultyPointAt(StartTime);

            DifficultyControlPoint ??= beatmapDifficultyControlPoint;
        }
```

But this feels a bit unsafe / hard to work with. The method I was using to decide whether a custom override control point is required (whether the `Local` version was `null` or not) can no longer be used. So instead, an alternative approach would be needed:
- Always make a new one on each change (probably not great?)
- Expose `bool`s in `HitObject` to allow access to the `DifficultyControlPoint == beatmapDifficultyControlPoint` comparison.

For reference, see the use case at https://github.com/ppy/osu/pull/10417/files#diff-7d2412aa7d34bb3e5c9b8ab573bf0bf6dd0795224a6261c46bd4a862eed68933R207-R219

peppy: Creating a new point every change does work, so that is a direction we can take, potentially:

```csharp
                                var closestGlobalDifficulty = editorBeatmap.ControlPointInfo.DifficultyPointAt(hitObject.StartTime);

                                var localDifficulty = hitObject.DifficultyControlPoint = new DifficultyControlPoint
                                {
                                    SpeedMultiplier = hitObject.DifficultyControlPoint.SpeedMultiplier
                                };

                                localDifficulty.SpeedMultiplier *= (repeatHitObject.Duration / proposedDuration);

                                if (closestGlobalDifficulty.IsRedundant(localDifficulty))
                                    hitObject.DifficultyControlPoint = null;

                                beatmap.Update(hitObject);
```

But note that we are doing a second lookup on `ControlPointInfo` to check whether the local point is redundant. Also it's hard to know whether we should be displaying a local override any more, so every hitobject shows one as a result:

20210817 184548 (dotnet)

Previously this was also attained from the `null` case.

peppy: I've pushed the changes above to https://github.com/ppy/osu/compare/master...peppy:smgi-control-point-suggestions?expand=1 if you want to have a look around.

smoogipoo: As discussed earlier, it may be better to go the full-hog and copy the control points anyway, removing the concept of "override" and fixing the timeline display to reference the previous hitobject as an additional step.

It's going to be difficult to encode the beatmap while maintaining the concept of an "override" anyway.

peppy:
Closing in favour of #14578, which takes this one step further and is definitely a better direction than what we have here.

PR #10417, opened 2020-10-08
Add ability to adjust slider's velocity via editor timeline

### LINKED ISSUE #7882: Add the ability to change slider velocity from compose mode
**Describe the new feature:** 
Current way of setting slider velocity is very unintuitive for new mappers because of how slider velocity must be set via a number instead of snapping it to the beats in the timeline
Make slider duration setting the same way spinners are made and move reverse creation into a context menu when you right-click a slider.

peppy:
- [ ] Depends on https://github.com/ppy/osu/pull/10416
- Closes #7882

Adds the most basic implementation of being able to adjust "slider velocity" via the compose timeline, all without making control points at all. Hold shift and drag end of slider to adjust.

The override will be automatically removed when redundant.

Note that this doesn't update autoplay after making a change, as the legacy patcher does not see the local overrides as changes. If you want to test autoplay, move the adjusted slider once.


smoogipoo:
Loading and saving need to be supported before I'm ready to review this.

peppy:
Definitely. Was holding off on that until you gave your thoughts on the dependent PR (it would be part of that, not this).

peppy:
Closing this one. Will reopen fresh using the newer structural changes in #14619.

PR #10419, opened 2020-10-08
Increase reliability of editor beatmap change handling

### LINKED ISSUE #10215: EditorChangeHandler Begin/EndChange doesn't work with (now-)scheduled HitObjectUpdated calls
Due to HitObjectUpdated getting called with a schedule to debounce potential multiple operations, it turns out that the operations could look to be inside a Begin/End block but not actually grouped correctly:

image

The `UpdateHitObject` call here will not run the callback events inline. Need to figure the correct way to handle this, probably involving not doing the scheduling and fixing (via debounce?) the operations which require it in the first place.

peppy:
This makes a few important changes:

- `EditorBeatmap` employs the same transactional logic as `ChangeHandler`. I have chosen this over the recently added `ApplyBatchChanges(Action action)` as it works better in cases where we are DIing the beatmap and don't always expect it to be present. Annoying to code around if you have to null check then run the delegate locally to ensure correct behaviour.
- `ChangeHandler` binds to `EditorBeatmap`'s transactional events, ensuring 100% coverage. This means that calling only `EditorBeatmap.SaveState` is enough to save an undo state too.
- `EditorBeatmap.BeginChange`/`EndChange` is preferred when operating on the beatmap directly and in one batch (ie. without game loop running while in a transaction). `ChangeHandler` is still used in many places, and has a slightly different meaning – it will not stop the flow of `HitObject` updates for instance. The remaining usages are in areas like `DragStart`/`DragEnd`, where we want to track a change but it's a long-running operation.
- Calling `EditorBeatmap.Update` is still debounced as previously, even when not in a transaction. I tried removing this and making the debounce local (ie. in `SliderSelectionBlueprint`) but ran into order-of-execution issues. I think this is fine though.

Closes #10215.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderSelectionBlueprint.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/SliderSelectionBlueprint.cs
@@ -182,7 +182,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders
         private void updatePath()
         {
             HitObject.Path.ExpectedDistance.Value = composer?.GetSnappedDistanceFromDistance(HitObject.StartTime, (float)HitObject.Path.CalculatedDistance) ?? (float)HitObject.Path.CalculatedDistance;
-            editorBeatmap?.UpdateHitObject(HitObject);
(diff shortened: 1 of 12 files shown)


smoogipoo (approved):
lgtm

PR #10420, opened 2020-10-08
Fix taiko's HitObjectComposer not allowing movement of selected hitobjects

peppy:


bdach (approved):
this one's a no-brainer :stuck_out_tongue:

PR #10427, opened 2020-10-08
Fix first hitobject in osu! hidden mod not getting correct fade applied

### LINKED ISSUE #8153: First hitcircle not vanishing with HD

peppy:
Closes #8153

### DIFF
--- a/osu.Game/Rulesets/Mods/ModHidden.cs
+++ b/osu.Game/Rulesets/Mods/ModHidden.cs
@@ -38,7 +38,13 @@ namespace osu.Game.Rulesets.Mods
         public virtual void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> drawables)
         {
             if (IncreaseFirstObjectVisibility.Value)
+            {
+                var firstObject = drawables.FirstOrDefault();
+                if (firstObject != null)
+                    firstObject.ApplyCustomUpdateState += ApplyFirstObjectIncreaseVisibilityState;
+
                 drawables = drawables.SkipWhile(h => !IsFirstHideableObject(h)).Skip(1);
+            }
 
             foreach (var dho in drawables)
                 dho.ApplyCustomUpdateState += ApplyHiddenState;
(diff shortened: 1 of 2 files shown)


osu.Game/Rulesets/Mods/ModHidden.cs
```diff
@@ -38,7 +38,13 @@ public void ReadFromConfig(OsuConfigManager config)
         public virtual void ApplyToDrawableHitObjects(IEnumerable<DrawableHitObject> drawables)
         {
             if (IncreaseFirstObjectVisibility.Value)
+            {
+                var firstObject = drawables.FirstOrDefault();
```
bdach, line 42: I believe this should be

```suggestion
                var firstObject = drawables.FirstOrDefault(IsFirstHideableObject);
```

or, probably better, rewrite this logic a bit to be

```csharp
if (IncreaseFirstObjectVisibility.Value)
{
    drawables = drawables.SkipWhile(h => !IsFirstHideableObject(h));

    var firstObject = drawables.FirstOrDefault();
    if (firstObject != null)
        firstObject.ApplyCustomUpdateState += ApplyFirstObjectIncreaseVisibilityState;

    drawables = drawables.Skip(1);
}
```

Can be tested on any map that starts with a spinner - the second object won't fade out at all while the third will work like normal hidden.

bdach (changes requested):

bdach (dismissed):

smoogipoo:
The weird part is that the approach circle is not supposed to fade out, but I'm not good enough at HD to know how confusing this actually is. Maybe this should instead fade out just `circle.CirclePiece` if `increasedVisibility = true`?

bdach:
That's what `master` does already, doesn't it?

smoogipoo:
Master:
- Doesn't fade out anything.

This PR:
- Fades out everything.

Stable:
- Fades out only the circle.

bdach:
Oh the issue is formulated weirdly. Sure.

peppy:
That wasn't the intention.

smoogipoo:
To me it feels weird because the approach is hidden way before it touches the circle, making it kinda useless? Again I can't really provide much insight here but the reason for it not hiding in the first place is because players maintain an afterimage of the circle in their mind specifically so they can hit the first note accurately, assuming that the approach is visible.

peppy:
Okay, will change it to match stable for now.

smoogipoo (approved):

PR #10428, opened 2020-10-08
Reset new combo button state after successful placement

peppy:


osu.Game/Screens/Edit/Compose/Components/ComposeBlueprintContainer.cs
```diff
@@ -201,7 +201,12 @@ protected sealed override SelectionBlueprint CreateBlueprintFor(HitObject hitObj
         protected override void AddBlueprintFor(HitObject hitObject)
         {
             refreshTool();
+
             base.AddBlueprintFor(hitObject);
+
+            // on successful placement, the new combo button should be reset as this is the most common user interaction.
+            if (Beatmap.SelectedHitObjects.Count == 0)
```
bdach, line 208: Not sure how to feel about this relying on selection count to work... It works correctly, and I can't break it (tried with paste, nothing funny), but it does feel a smidge brittle.

smoogipoo: Does removing this check break anything? It does feel misplaced...

peppy: It can be removed. I have it here as a safety measure because I'm not sure when this method may be called in the future. It seems like currently it's only on completing a placement, but do we have a guarantee of that?

smoogipoo: Let's keep it in for now.

smoogipoo (approved):

PR #10430, opened 2020-10-08
Fix ternary button states not updating correctly after a paste operation

peppy:
Accessing selected objects via blueprints was not realiable (dependent on action flow). Such was the case for the `UpdateTernaryStates()` function, which after a paste operation could not see the new selected objects (`selectedBlueprints` had not yet been populated).

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -37,8 +37,6 @@ namespace osu.Game.Screens.Edit.Compose.Components
 
         public int SelectedCount => selectedBlueprints.Count;
 
-        public IEnumerable<HitObject> SelectedHitObjects => EditorBeatmap.SelectedHitObjects;
-
         private Drawable content;
 
         private OsuSpriteText selectionDetailsText;
@@ -309,7 +307,7 @@ namespace osu.Game.Screens.Edit.Compose.Components
         {
             ChangeHandler?.BeginChange();
 
-            foreach (var h in SelectedHitObjects)
+            foreach (var h in EditorBeatmap.SelectedHitObjects)
             {
                 // Make sure there isn't already an existing sample
                 if (h.Samples.Any(s => s.Name == sampleName))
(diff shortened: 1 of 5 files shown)


osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -37,7 +37,7 @@ public class SelectionHandler : CompositeDrawable, IKeyBindingHandler<PlatformAc
 
         public int SelectedCount => selectedBlueprints.Count;
 
-        public IEnumerable<HitObject> SelectedHitObjects => selectedBlueprints.Select(b => b.HitObject);
+        public IEnumerable<HitObject> SelectedHitObjects => EditorBeatmap.SelectedHitObjects;
```
bdach, line 40: I'm not even honestly sure exposing as `public` even makes sense anymore. Only usages are `BlueprintContainer` (which already has a resolved `EditorBeatmap`), and subclasses (which can either access the protected `EditorBeatmap`, or we could make it a convenience protected property if that's better for code quality)..

smoogipoo: I'd agree with removing and referencing the beatmap directly, if that's what you're suggesting.

peppy: Can give that a try. I think this existed before `EditorBeatmap` was a thing.

peppy: Seems to work fine.

peppy: Are you sure about this? There's roughly 20-30 references to it.

smoogipoo: I didn't realise there were that many... Fine to leave it as a helper property in that case.

peppy: Should I revert the changes then? I'm neutral on it.

bdach: Personally I'm mostly worried about someone potentially accidentally getting selected hit objects through `SelectedBlueprints` (which is also public) instead of this and hitting this same bug again.

peppy:
I think I'll just go with this for now and do further cleanups on the class in a follow-up PR.

PR #10431, opened 2020-10-08
Fix incorrect timing distribution display due to lack of rounding

### LINKED ISSUE #10426: ~0ms bar on the extended results screen is overweighted compared to others
**Describe the bug:** The central bar on the timing distribution graph takes more judgements than others and ends up being way taller as a result of bad rounding.
https://github.com/ppy/osu/blob/8dddd8aff56ac819ee8165e42155a9400b5eb9d1/osu.Game/Screens/Ranking/Statistics/HitEventTimingDistributionGraph.cs#L69
`(int)` floors the value, so on the screenshots below (`binSize == 2`) this line of code results in all judgements between (but not equal to) -2ms and +2ms to be put into the ~0ms bar (4ms worth of judgements total), whereas all the other bars only include 2ms worth of judgements.

**Screenshots or videos showing encountered issue:** I played the same map 5 times with different global offset, having noticed the bar in question peaking way out of place way too often. I normally play with +25ms offset. The issue is most visible on +6 and +36ms screenshots, but is still kind of prominent on some others.
image
image
image
image
image

**osu!lazer version:** 2020.1005.0

peppy:
Closes #10426.

Before:

image

After:

image

### DIFF
--- a/osu.Game.Tests/Visual/Ranking/TestSceneHitEventTimingDistributionGraph.cs
+++ b/osu.Game.Tests/Visual/Ranking/TestSceneHitEventTimingDistributionGraph.cs
@@ -24,6 +24,12 @@ namespace osu.Game.Tests.Visual.Ranking
         }
 
         [Test]
+        public void TestAroundCentre()
+        {
+            createTest(Enumerable.Range(-100, 100).Select(i => new HitEvent(i / 50f, HitResult.Perfect, new HitCircle(), new HitCircle(), null)).ToList());
+        }
+
+        [Test]
         public void TestZeroTimeOffset()
         {
             createTest(Enumerable.Range(0, 100).Select(_ => new HitEvent(0, HitResult.Perfect, new HitCircle(), new HitCircle(), null)).ToList());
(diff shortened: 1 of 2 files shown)


bdach:
So the image above has some visible anomalies in it, namely that:

* The middle bin is lower - this is because of `MidpointRounding.AwayFromZero`. For a bin centered at N times the bin width where N is not 0, (N-0.5) gets rounded to N, and (N+0.5) gets rounded to N+1. The zero bin, however, loses its both margin points to neighbouring bins (-0.5 to bin -1, and 0.5 to bin 1).
* The side bins are lower - that's due a fencepost error in the test itself. To generate a set of symmetrical points, with N points either side around 0, 2N+1 points are needed - in the case of this test here it'd have to be 301 instead of 300.

Neither note is relevant to the fix, just figured I'll leave an explanation to the wonkiness. Either way the discontinuities cannot be entirely eliminated, and the fix is correct as far as I can tell.

bdach (approved):

smoogipoo (approved):

PR #10432, opened 2020-10-08
Add "reverse pattern" support to editor selection handler

peppy:
Note that paths are not perfectly reversed if they have excess length at the end. Have to check but I think stable does this too.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
+++ b/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
@@ -7,6 +7,7 @@ using System.Linq;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Utils;
+using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Osu.Objects;
 using osu.Game.Screens.Edit.Compose.Components;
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -41,6 +43,54 @@ protected override void OnOperationEnded()
...
+
+                        // propagate types forwards to last null type
+                        if (i == points.Length - 1)
+                            p.Type.Value = lastType;
+                        else if (p.Type.Value != null)
+                        {
+                            var newType = p.Type.Value;
+                            p.Type.Value = lastType;
+                            lastType = newType;
+                        }
```
smoogipoo, line 84: I... think this is correct. Even though it breaks the visualiser.

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -41,6 +43,54 @@ protected override void OnOperationEnded()
...
+
+                        slider.Path.ControlPoints.Insert(0, p);
```
smoogipoo, line 86: As a side note, since this is the second time this has been done - it's much more efficient to do bulk operations with control points, including adding of control points, because the slider path is pretty much recalculated immediately with each change.

smoogipoo (approved):

PR #10440, opened 2020-10-09
Fix editor playing deleted beatmap's track on exit-without-save

### LINKED ISSUE #10434: Song selected in song setup plays when project is abandoned
If a song has been chosen in song setup but the user chooses to abandon the project, the imported track will be playing when returning to the title screen even though no beatmap was created upon exit. Skipping forward to the next track and returning plays the intro track.

video: https://youtu.be/UIN9ZCgUzFU
osu!lazer version: 2020.1005.0

peppy:
Closes #10434.

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -150,7 +150,7 @@ namespace osu.Game.Overlays
         {
             if (IsUserPaused) return;
 
-            if (CurrentTrack.IsDummyDevice)
+            if (CurrentTrack.IsDummyDevice || beatmap.Value.BeatmapSetInfo.DeletePending)
             {
                 if (beatmap.Disabled)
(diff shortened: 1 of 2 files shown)


osu.Game/Overlays/MusicController.cs
```diff
@@ -150,7 +150,7 @@ public void EnsurePlayingSomething()
         {
             if (IsUserPaused) return;
 
-            if (CurrentTrack.IsDummyDevice)
+            if (CurrentTrack.IsDummyDevice || beatmap.Value.BeatmapSetInfo.DeletePending)
```
peppy, line 153: Note that I've added this for future usage/safety. It's not required or used in this PR due to the reason mentioned in the inline comment.

smoogipoo (approved):

PR #10443, opened 2020-10-09
Improve TimelineTickDisplay performance

peppy:
Before: ~20ms (1/8 tick divisor on solace of oblivion)
After: ~0.3ms

I started by using a `DrawablePool` but it makes more sense to locally optimise this one. May turn out that we want to abstract this into `TimelinePart` for other usages (the display bounds part at least) but want to get this into the next lazer release if possible, as it's causing 90% of the editor performance issues.


smoogipoo:
This still deeeeeesperately needs pooling. Can that be easily slotted in? It's currently causing a huge number of Gen2s that make me even prefer the old performance...

peppy:
From what? No new instances are constructed unless changing tick divisor.

smoogipoo:
Ah looks like it was due to WaveformGraph. Will fix there.

smoogipoo (approved):

PR #10450, opened 2020-10-09
Fix slider samples being overwritten by the last node

smoogipoo:
Mania requires knowledge of the slider's default samples:

https://github.com/ppy/osu/blob/da0308425163b4330073fbdf4ee0250314c5ae48/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs#L368-L369

But they were being replaced during parsing by the last node's samples. The way `Slider`'s doing things is still likely completely incorrect, because it then uses these samples to determine the tick's bank/volume, but the purpose of this PR is not to fix that:
https://github.com/ppy/osu/blob/da0308425163b4330073fbdf4ee0250314c5ae48/osu.Game.Rulesets.Osu/Objects/Slider.cs#L215-L216

I've tested this against (and cross-checked with stable):
https://osu.ppy.sh/beatmapsets/692068#osu/1464590
https://osu.ppy.sh/beatmapsets/177187#osu/426550 (fully keysounded)

Furthermore, I also noticed `JuiceStream` completely disregarding the existence of per-node samples, which led to incorrect playback in maps like the second one above. I've fixed this since it's a very minor change.

### DIFF
--- a/osu.Game.Rulesets.Catch/Objects/JuiceStream.cs
+++ b/osu.Game.Rulesets.Catch/Objects/JuiceStream.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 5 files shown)


peppy (dismissed):

smoogipoo:
Took the easy path out and changed the test. Mania shouldn't care about the hold note's default samples anyway. It's already probably resulting in sound doubling.

bdach (approved):
seems correct as far as I can tell :+1:

PR #10451, opened 2020-10-09
Fix breaks being culled during beatmap parsing

smoogipoo:
Mania needs to know the total break time to calculate its "conversion difficulty":

https://github.com/ppy/osu/blob/da0308425163b4330073fbdf4ee0250314c5ae48/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/PatternGenerator.cs#L109

This needs to be an exact value that includes breaks with no effect.


osu.Game/Beatmaps/Formats/LegacyBeatmapDecoder.cs
```diff
@@ -307,12 +307,7 @@ private void handleEvent(string line)
                     double start = getOffsetTime(Parsing.ParseDouble(split[1]));
                     double end = Math.Max(start, getOffsetTime(Parsing.ParseDouble(split[2])));
 
-                    var breakEvent = new BreakPeriod(start, end);
-
-                    if (!breakEvent.HasEffect)
-                        return;
```
bdach, line 313: I'm confused - you stated in the description that breaks with no effect shouldn't be included, but this guard is gone, so I'm not sure which thing I'm supposed to trust. `Beatmap.TotalBreakTime` doesn't seem to filter by validity either...

smoogipoo: Sorry, description should've read "includes breaks with no effect". 

Imo this is a more correct value for `Beatmap.TotalBreakTime`, but whether breaks actually do anything or are used purely for informational purposes depends on `HasEffect`. Alternatively, we may be able to keep the culling and acumulate `TotalBreakTime` during decode, with the caveat being that it has to be transferred in `BeatmapConverter`.

bdach: Figured this was the case but wanted to clarify 100%. I've checked usages of `Breaks` and as far as I can see you got the ones that needed adjusting already, and the change to not cull at decode time is arguably more correct in some cases (ie. decode-encode stability/idempotence)

bdach (approved):

PR #10452, opened 2020-10-09
Make mania beatmap conversions match stable

smoogipoo:
In my quest to bring mania diffcalc in-line with stable, I found many issues with beatmap conversion.

1. Incorrect initial density. During circle conversion, density is first computed and then used inside the generator. This needs to be `int.MaxValue` for the first circle (default density value), but wasn't the case because it was always being updated.
2. Spinner conversion did not consider `ForceNotStack` (which depends on the last pattern), and also omitted the forced 0 lower bound for retrieving the random column. The omission would cause the lower bound to be `RandomStart`, which is not 0 with 8K.
3. Slider conversion has a lot of fp-precision issues. Most of these relate to when objects should be added to the "intermediate" pattern or the "end-time" pattern (which forms the set of objects for the next pattern to reference against):
https://github.com/ppy/osu/blob/da0308425163b4330073fbdf4ee0250314c5ae48/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs#L79-L82
To fix these, I've changed it to use ints everywhere.  
I've also noticed that stable mania calculates the end time in its own way, which I've replicated. This also uses the same SV calculation as taiko's converter:
https://github.com/ppy/osu/blob/da0308425163b4330073fbdf4ee0250314c5ae48/osu.Game.Rulesets.Taiko/Beatmaps/TaikoBeatmapConverter.cs#L176-L182

Here's the full suite of maps I've tested these changes against: [mania-test-beatmaps.zip](https://github.com/ppy/osu/files/5354885/mania-test-beatmaps.zip), but keep in mind that I've gone through ~95k ranked/loved/approved beatmaps and can reasonably confirm no differences across that full set. These are only the maps that have resulted in the changes in this PR.

### DIFF
--- a/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
+++ b/osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
@@ -3,8 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
-using osu.Framework.Utils;
 using osu.Game.Audio;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Mania.MathUtils;
@@ -12,6 +12,7 @@ using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Mania.Objects;
 using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Beatmaps.Formats;
 
 namespace osu.Game.Rulesets.Mania.Beatmaps.Patterns.Legacy
 {
@@ -25,8 +26,9 @@ namespace osu.Game.Rulesets.Mania.Beatmaps.Patterns.Legacy
         /// </summary>
         private const float osu_base_scoring_distance = 100;
 
-        public readonly double EndTime;
-        public readonly double SegmentDuration;
+        public readonly int StartTime;
+        public readonly int EndTime;
+        public readonly int SegmentDuration;
         public readonly int SpanCount;
 
         private PatternType convertType;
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
```diff
@@ -76,7 +83,7 @@ public override IEnumerable<Pattern> Generate()
 
             foreach (var obj in originalPattern.HitObjects)
             {
-                if (!Precision.AlmostEquals(EndTime, obj.GetEndTime()))
+                if (EndTime != (int)Math.Round(obj.GetEndTime()))
```
bdach, line 86: Could this be `Math.Floor` for consistency with the ctor? Not a huge deal really but I'm just semi-curious if there's a reason for why the two differ.

bdach (commented):
Have extracted and ran the tests attached and can confirm that with the entire PR queue I get failures on `master` and passes with all of the diffcalc branches up to this, so looks good to me. I haven't looked too hard at the actual operations involved as I guess it only matters that it should match whatever stable does.

Would approve outright but there's one thing I want to clarify/ask here and one in #10451.

osu.Game.Rulesets.Mania/Beatmaps/Patterns/Legacy/DistanceObjectPatternGenerator.cs
smoogipoo: The calculation in the ctor is directly from stable. The calculation here is just making sure that lazer's storage as `double` doesn't lead to incorrect comparisons. It passes the tests but I don't believe this can be replaced with `Math.Floor` - I'll run across my full suite of beatmaps.

bdach: If it's verbatim from stable then not worth. Maybe leave a comment to not touch at best.

smoogipoo: Changing this one to a `Math.Floor()` doesn't affect any beatmaps I have but I'm still not sure how I feel about that in principle.

bdach: Nah it's fine. As I said I was mostly curious if there was reason for the difference other than stable gospel.

bdach (approved):

peppy (approved):

PR #10453, opened 2020-10-09
Bring mania's diffcalc in-line with stable

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10500
- [x] https://github.com/ppy/osu/pull/10503

Along with #10450, #10451, and #10452, this brings mania's diffcalc in-line with stable/web. Or as best as it can be.

Here's some quick stats, out of 90k ranked/loved/approved beatmaps:
- Total mania-specific differences >0.01*: 27
    - 0.3% of mania-specific beatmaps.
- Total mania-specific differences >0.1*: 3
    - 0.04% of mania-specific beatmaps.
- Total mania-specific differences >0.2*: 1
    - 0.01% of mania-specific beatmaps.
    - Difference is -0.598*
    - The difference occurs on https://osu.ppy.sh/beatmapsets/419387#mania/908448, which is a map with mania sliders. The differences are all due to float precision in the `holdEndTime` checks in `Strain.cs`. I've tried but I can't get much closer than this.
- Total convert differences >0.01*: 13091
    - 18% of converted beatmaps.
- Total convert differences >0.1*: 118
    - 0.1% of converted beatmaps.
- Total convert differences >0.2*: 7
    - 0.009% of converted beatmaps.
    - All aspire and/or loved gimmick maps - lazer's parsing limits change the beatmaps leading to changes in conversion.

Full data: https://docs.google.com/spreadsheets/d/1pefq65DTnFOLI_BCrE3MzztsdJfcJ9Su4IOZAF9wrhs/edit#gid=1879017573


smoogipoo:
Will resolve tests after all those branches are merged.

osu.Game.Rulesets.Mania/Difficulty/Skills/Strain.cs
```diff
@@ -0,0 +1,80 @@
...
+
+                // Decay individual strains
+                individualStrains[i] = applyDecay(individualStrains[i], current.DeltaTime, individual_decay_base);
+            }
+
+            holdEndTimes[column] = endTime;
+
+            // Increase individual strain in own column
+            individualStrains[column] += 2.0 * holdFactor;
+            individualStrain = individualStrains[column];
```
bdach, line 66: This field assignment is super confusing and possibly buggy. Because `GetPeakStrain()` is called by a different code path in the base difficulty calculator than this, doesn't the value returned by that method depend on call order? It's a bit sketch if I'm honest.

smoogipoo: If you're talking about order of calls to `StrainValueOf` vs `GetPeakStrain`, then yes. But I don't know what the solution is.

If you're talking about the order of calls to `StrainValueOf` in terms of hitobject column, then also yes, but that's one of the reasons mania's diffcalc is completely bonkered. I'm just replicating stable 1:1 here.

smoogipoo: Fwiw, I don't intend to keep the virtual-ness of `GetPeakStrain`. It's just that mania's diffcalc needs a rewrite before that can happen.

peppy:
@smoogipoo please update tests and confirm when your unranked beatmap tests complete.

osu.Game.Rulesets.Mania/MathUtils/LegacySortHelper.cs
```diff
@@ -0,0 +1,164 @@
...
+using System.Collections.Generic;
+using System.Diagnostics.Contracts;
+
+namespace osu.Game.Rulesets.Mania.MathUtils
+{
+    /// <summary>
+    /// Provides access to .NET4.0 unstable sorting methods.
+    /// </summary>
+    /// <remarks>
+    /// Source: https://referencesource.microsoft.com/#mscorlib/system/collections/generic/arraysorthelper.cs
```
peppy, line 14: Probably add microsoft's copyright notice in here for good measure

```suggestion
    /// Source: https://referencesource.microsoft.com/#mscorlib/system/collections/generic/arraysorthelper.cs
    /// Copyright (c) Microsoft Corporation.  All rights reserved.
```

peppy (changes requested):

smoogipoo:
I've fixed the diffcalc, but I've noticed the score multiplier attribute missing. I'm currently fixing that, so I'll block until that's done.

PR #10462, opened 2020-10-09
Try to match stable's judgement animations

### LINKED ISSUE #10195: For legacy skins, judgements should use animations that match stable

**Describe the bug:** 

https://youtu.be/Je50_IcAmFs

**osu!lazer version:** 
2020.910.0

LittleEndu:
Closes #10195

In this PR, ``DrawableJudgment`` is made to check if its drawable is animatable legacy skin component and if so it won't get any scale effects applied. This should match what stable does, where scale effects only apply to skins with no animations.

### DIFF
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -71,8 +71,12 @@ namespace osu.Game.Rulesets.Judgements
 
         protected virtual void ApplyHitAnimations()
         {
-            JudgementBody.ScaleTo(0.9f);
-            JudgementBody.ScaleTo(1, 500, Easing.OutElastic);
+            if (bodyDrawable.Drawable is LegacySkinExtensions.SkinnableTextureAnimation) { }
+            else
+            {
+                JudgementBody.ScaleTo(0.9f);
+                JudgementBody.ScaleTo(1, 500, Easing.OutElastic);
+            }
 
             this.Delay(FadeOutDelay).FadeOut(400);
         }


bdach:
I vehemently disagree with the way this is done. Non-legacy skins should have no notions of legacy anything.

LittleEndu:
I honestly can't see any other way to disable something that's done to all skin elements regardless of being legacy or provisional after ``GetDrawableComponent`` is already called. In here ``DrawableJudgement`` creates a new ``SkinnableDrawable`` which calls ``GetDrawableComponent``, and then ``DrawableJudgement`` applies the scale effect to that ``SkinnableDrawable`` (again, not knowing if it's legacy component or not).

peppy:
You need to do it how it was specified in discord: creating a legacy judgement.

LittleEndu:
Did you even read what I just wrote? Honestly just do me a favour and close comments to collaborators...

peppy:
You would move the transform inside the legacy judgement. I'd highly suggest you look at the many other examples of the exact same pattern.

LittleEndu:
I have come to the conclusion that making ``LegacyJudgement`` is either impossible or way over my head. 

For there to be a ``DrawableJudgement`` and ``LegacyJudgement`` the consumer of ``DrawableJudgement`` would need to be made to call ``GetDrawableComponent`` (that is to select which one to use based on what skin is loaded, that's what every other gameplay element does when it needs a legacy vs provisional component). The consumer of ``DrawableJudgement`` is ``JudgementContainer<T>`` which itself is consumed by every ruleset individually (except fruits). So now I'm looking at needing to create ``LegacyJudgementContainer<T>`` that would contain legacy judgements instead of ``DrawableJudgement``s and needing to create a legacy version of all the ruleset judgements, e.g ``LegacyOsuJudgement`` that somehow needs to inherit both the behaviour of ``DrawableOsuJudgement`` and ``LegacyJudgement``. There is probably something that you could do by refactoring some of these classes into interfaces but this is the part where it gets way over my head. 

And I still fail to find any other examples of the exact same pattern anywhere in the code.

bdach:
a possible way forward that I think of from the top of my head is making `DrawableJudgement.JudgementBody` a new skin component on its own, with its default implementation being what is currently in `DrawableJudgement` (so a container with the hit result component inside), and then moving all transform application to it out from `DrawableJudgement` inside of that new component. then you can create another variant of that component for legacy skins and turn off the transform there locally without conditionals.

it would require an introduction of an interface and a cast somewhere, but there's precedent for that, I suppose (see `ICatchComboCounter`). it's perhaps a little too much of indirection, would have to see the working end result to say for sure.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -71,8 +71,12 @@ private void load()
 
         protected virtual void ApplyHitAnimations()
         {
-            JudgementBody.ScaleTo(0.9f);
-            JudgementBody.ScaleTo(1, 500, Easing.OutElastic);
+            if (bodyDrawable.Drawable is LegacySkinExtensions.SkinnableTextureAnimation) { }
```
LittleEndu, line 74: Just a passing thought, what if this line was replaced with checking if ``bodyDrawable.Drawable`` is at all animated. Would remove the legacy notation and would not replace the current default implementation.

bdach: Not viable in the long run, if the drawable-based skinning system happens. People will have animations randomly not work here in that case.

LittleEndu: If people have an animated drawable they probably want to disable scaling here anyway. If these scaling effects aren't disabled, the animation might look off (that's the reason I opened #10195). Also, the line doesn't disable all animations, it disables only those two scale transformations.

bdach: This doesn't make the proposal sound any better, sadly. What if they *do* want to resolve all animations? Then it's back to square one.

The provided direction will ensure this should not be a problem in any case going forward.

PR #10468, opened 2020-10-10
Fix windows key blocking applying when window is inactive / when watching a replay

### LINKED ISSUE #10467: windows key cannot be used in autoplay/replays
**Describe the bug:** when watching autoplay (or any replays), windows key cannot be used. if the lazer window is selected, it will open the start menu once but pressing the windows key again wont close it (essentially, the key cannot be used again). when the lazer window is not selected, pressing windows key wont open the start menu. this also affects shortcuts that has wiindows key in it (e.g. win + r for run dialog box). on sdl, windows key cannot be used. not even opening the start menu once like in osutk nor when the lazer window is delected and shortcuts is affected.

(sorry for uploading here, i cant access drive for some reason)
This was made while lazer is in osutk. i used onscreen keyboard for demonstration but its the same when using a physical keyboard
[kjkn.mkv.zip](https://github.com/ppy/osu/files/5358391/kjkn.mkv.zip)

**osu!lazer version:** 2020.1009.0 on windows 10 2004 19041.508

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5358397/runtime.log)

peppy:
By relying on the new local gameplay state we resolve both these issues in one swoop.

Closes #10467.

### DIFF
--- a/osu.Desktop/Windows/GameplayWinKeyBlocker.cs
+++ b/osu.Desktop/Windows/GameplayWinKeyBlocker.cs
@@ -5,24 +5,24 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Platform;
+using osu.Game;
 using osu.Game.Configuration;
 
 namespace osu.Desktop.Windows
 {
     public class GameplayWinKeyBlocker : Component
     {
-        private Bindable<bool> allowScreenSuspension;
         private Bindable<bool> disableWinKey;
+        private Bindable<bool> localUserPlaying;
 
-        private GameHost host;
+        [Resolved]
+        private GameHost host { get; set; }
 
-        [BackgroundDependencyLoader]
-        private void load(GameHost host, OsuConfigManager config)
+        [BackgroundDependencyLoader(true)]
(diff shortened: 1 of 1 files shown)


osu.Desktop/Windows/GameplayWinKeyBlocker.cs
```diff
@@ -5,32 +5,32 @@
...
-        [BackgroundDependencyLoader]
-        private void load(GameHost host, OsuConfigManager config)
+        [BackgroundDependencyLoader(true)]
+        private void load(OsuGame game, OsuConfigManager config)
         {
-            this.host = host;
-
-            allowScreenSuspension = host.AllowScreenSuspension.GetBoundCopy();
-            allowScreenSuspension.BindValueChanged(_ => updateBlocking());
+            localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
```
swoolcock, line 24: `game` nullable?

peppy: Ah yeah, realised halfway that it can never be null (this component is only ever used in `OsuGameDesktop`). Will remove the null allowance.

bdach (approved):
I haven't tested but source seems sane :+1:

PR #10471, opened 2020-10-10
Add the ability to search for local beatmaps via online IDs

### LINKED ISSUE #10470: Allow searching with beatmap(set) IDs in songselect
**Describe the new feature:** 
Stable allows searching for beatmaps via their ID. Lazer should too.

screenshot212
screenshot213
osu_2020-10-10_14-53-28
osu_2020-10-10_14-53-37
The map exists in my lazer library, just fyi

peppy:
Closes #10470.

Only allows searching when the search text is a single word and also a number


bdach (approved):
added tests here, looks good to me :+1:

PR #10472, opened 2020-10-10
Expose BreakTracker.IsBreakTime in Player

LeNitrous:
Required by https://github.com/ppy/osu/pull/9606. This allows others such as mods to subscribe to the break state of the player.

### DIFF
--- a/osu.Game/Screens/Play/Player.cs
+++ b/osu.Game/Screens/Play/Player.cs
@@ -92,7 +92,7 @@ namespace osu.Game.Screens.Play
         /// <summary>
         /// Whether the gameplay is currently in a break.
         /// </summary>
-        public readonly BindableBool IsBreakTime = new BindableBool();
+        public readonly IBindable<bool> IsBreakTime = new BindableBool();
 
         private BreakTracker breakTracker;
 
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Play/Player.cs
```diff
@@ -89,6 +89,8 @@ public class Player : ScreenWithBeatmapBackground
 
         public BreakOverlay BreakOverlay;
 
+        public IBindable<bool> IsBreakTime => breakTracker?.IsBreakTime;
```
peppy, line 92: Should probably add xmldoc, and it should never be null. If there's a reason `breakTracker` can be null then initialise this bindable locally and bind to the `breakTracker` one in load/`LoadComplete`

peppy (changes requested):

osu.Game/Screens/Play/Player.cs
```diff
@@ -257,6 +262,7 @@ private void load(AudioManager audio, OsuConfigManager config, OsuGame game)
                 mod.ApplyToHealthProcessor(HealthProcessor);
 
             breakTracker.IsBreakTime.BindValueChanged(onBreakTimeChanged, true);
+            IsBreakTime.BindTo((BindableBool)breakTracker.IsBreakTime);
```
bdach, line 265: what's this cast for?

peppy (approved):

PR #10476, opened 2020-10-10
Fix looping samples not pausing in gameplay

### LINKED ISSUE #10454: The spinner sound is looped even if paused
**Describe the bug:** 
There have already been several similar problems, but the spinner was not mentioned.
https://youtu.be/xx9LbFAcl_c
**osu!lazer version:** 
2020.1009.0

bdach:
Resolves #10454.

# Summary

The sample pausing logic added in `ISamplePlaybackDisabler` and `GameplayClock`:

https://github.com/ppy/osu/blob/bd44340423e5517b580cabfaa715fb8cd4fb9058/osu.Game/Screens/Play/GameplayClock.cs#L71-L76

would not run in gameplay due to the `FrameStabilityContainer` never processing it while paused:

https://github.com/ppy/osu/blob/bd44340423e5517b580cabfaa715fb8cd4fb9058/osu.Game/Rulesets/UI/FrameStabilityContainer.cs#L103-L116

Even though the container had `ProcessCustomClock` set, the call to the base `UpdateSubTree()` does not run when paused, leading to the frames never getting processed.

Ensure the `StabilityGameplayClock` is processed regardless of pause state. On its own it performs no timing logic and explicitly does not process its underlying clocks, so it does not affect frame stability in any way (only propagates updates of the sample disable for now).

# Remarks

Was not initially sure of this. A little more convinced this is OK now, but not 100%. Suggestions how to improve welcome - also not sure if the comments are all that legible.

Did not include a test as the issue was incomplete integration and any test would probably be a bit finicky. Will include one on request, though.

### DIFF
--- a/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
+++ b/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
@@ -45,6 +45,9 @@ namespace osu.Game.Rulesets.UI
             stabilityGameplayClock = new StabilityGameplayClock(framedClock = new FramedClock(manualClock = new ManualClock()));
 
             this.gameplayStartTime = gameplayStartTime;
+
+            // UpdateSubTree() does not necessarily always run, but we want to manually process the clock regardless of that.
+            ProcessCustomClock = false;
         }
 
         private readonly ManualClock manualClock;
@@ -103,6 +106,10 @@ namespace osu.Game.Rulesets.UI
             validState = !GameplayClock.IsPaused.Value;
 
             int loops = 0;
+
+            // gameplay clock does not update its underlying clocks, but we need to run its frames
+            // to propagate the state of the clocks it wraps to whichever component needs to read it
+            stabilityGameplayClock.ProcessFrame();
 
             while (validState && requireMoreUpdateLoops && loops++ < MaxCatchUpFrames)
             {


smoogipoo:
I guess my concern here is could this lead to a situation where the clock is processed multiple times in a single frame? Doing so could lead to an incorrect elapsed time value that can have other consequences.

I don't feel comfortable enough with this code and the 4 clocks it has to judge this for myself, so I'll leave it up to @peppy to handle this.

peppy:
The clocks in question here have overridden `ProcessFrame` (so calling it does nothing but update the paused bindable state). Still need to investigate if there's a better way to do this because I agree that I am not comfortable with the current structure of the clock hierarchy.

bdach:
I guess an alternative lowish-effort method would be to do something along the lines of

```diff
index eeea6777c..dc67a0cba 100644
--- a/osu.Game/Screens/Play/GameplayClock.cs
+++ b/osu.Game/Screens/Play/GameplayClock.cs
@@ -33,6 +33,8 @@ public class GameplayClock : IFrameBasedClock, ISamplePlaybackDisabler
         public GameplayClock(IFrameBasedClock underlyingClock)
         {
             this.underlyingClock = underlyingClock;
+
+            IsPaused.BindValueChanged(_ => updateSamplePlaybackDisableState());
         }
 
         public double CurrentTime => underlyingClock.CurrentTime;
@@ -72,9 +74,11 @@ public void ProcessFrame()
         {
             // intentionally not updating the underlying clock (handled externally).
 
-            samplePlaybackDisabled.Value = ShouldDisableSamplePlayback;
+            updateSamplePlaybackDisableState();
         }
 
+        private void updateSamplePlaybackDisableState() => samplePlaybackDisabled.Value = ShouldDisableSamplePlayback;
+
         public double ElapsedFrameTime => underlyingClock.ElapsedFrameTime;
 
         public double FramesPerSecond => underlyingClock.FramesPerSecond;
```

but I'm not sure it's much better (reads redundant if someone doesn't know better). I'm not even sure I prefer that method over what this PR does right now.

If there's a need to look at the clocks and restructure them then I can try but I'm not necessarily sure I'll be able to grasp all the nuance I need to in the existing implementation.

smoogipoo:
In the interests of simplifying this entire thing, could the `IsPaused` and sample playback disabling stuff be moved out from `GameplayClock` into its own component? It seems like `GameplayClock` has evolved from being a very basic thing to now serving several of these additional purposes which in turn make it very flimsy to work with.

Is there a reason why all of that stuff can't be put inside a different component that internally handles its own hierarchy via DI?

peppy:
Could even be moved to `Player` / `Editor` themselves, if the code is only a few lines...

Not against that direction.

bdach:
I'll see what I can come up with.

PR #10490, opened 2020-10-12
Fix beatmap decoder not supporting multi-segment sliders

### LINKED ISSUE #10475: New slider curve specifications don't save
Using legacy encoder = loss of data. Probably need to add basic support in legacy encoder for now.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10475

I knew I implemented the encoder part of this, but turns out I didn't implement the decoder part.

I believe there are enough tests between this and the decoder+encoder tests to ensure no breakage for existing maps, but I'm running a diffcalc in the background to confirm.


smoogipoo:
Looks good. No differences across a set of ~90k ranked beatmaps (osu/taiko/catch, converts included).

peppy (approved):
The format seems fine and the code seems to work. Not looking into it too deeply at legacy class.

We may need to port this back to stable at some point, just something to keep in mind for the future.

PR #10493, opened 2020-10-13
Update framework

peppy:


bdach:
I'm trying to look at the test failures but I'm very uncertain as to what's going on. I've had both [this](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs#L79) and [this](https://github.com/ppy/osu-framework/blob/master/osu.Framework/Graphics/Containers/DelayedLoadUnloadWrapper.cs#L113) assert fail while running tests/debugging. Neither is consistent, which makes debugging even more annoying. 

From what I can tell it is possible for a drawable to enter `EndDelayedLoad()` twice, probably due to the fact that DLUW creates one instance to pass it to DLW. That somewhat explains the first assert fail and also the "cannot add to multiple containers" exception.

As for the second assert, `checkForUnload()` can be run on a disposed drawable if the entire hierarchy goes away. In this case `SongSelect` et al. get disposed, and then despite that `checkForUnload()` is entered, failing at the `ClearInternal` call - it tries to set `Parent` to `null` (which it already is in this case) but fails since absolutely no sets to `Parent` are permitted on disposed drawables, even if the value to set is `null`.

I'm not even really sure where I would start fixing this to be fair.

bdach:
It's not ready but [here's a framework branch that passes the failed tests](https://github.com/ppy/osu-framework/compare/master...bdach:fix-dlw-potentially-loading-twice). The tl;dr of it is that both `DelayedLoadTriggered` and `DelayedLoadCompleted`, which determine whether drawables should be loaded in DLW's `Update()`, were both flimsy. If both are false, a load is queued.

* Contrary to what the xmldoc states `DelayedLoadTriggered` becomes `false` as soon as the async load task ends/is canceled.
* `DelayedLoadCompleted` could both become `true` too late (because child addition is scheduled), and then also wrongly become `false` if a DLUW gets unloaded (due to clearing contents).

Both conditions seem to fail in this case. I think what's happening is, because of the double load, a load task is dropped, leading to failed assertions (double-loaded drawables) and the "cannot add to multiple containers" error. The dropped load task also has an associated `EndDelayedLoad()` callback, which also schedules `checkForUnload()` every frame, leading to the third issue with trying to set a null `Parent` on a disposed drawable.

The fix proposal is to be a lot more stringent with both flags, but I can't say I'm confident of it at all. I'm not even 100% sure that what I typed up above is correct or makes sense.

smoogipoo (dismissed):

smoogipoo (approved):
let's go for attempt # 3

PR #10495, opened 2020-10-13
Fix mod customisation container not absorbing input

### LINKED ISSUE #10210: Mod Customisation container doesn't absorb input correctly
**Describe the bug:** (The best way to reproduce this is by increasing the UI size a lot.)
The customisation container shows the mods' tooltip (and the hover sound) when hovering a mod inside of it. I guess that's the wrong UX.
You can also select/deselect a mod by right-clicking it (but not left-clicking).

**Screenshots or videos showing encountered issue:** [Video](https://drive.google.com/file/d/1RyyK4vzkvvt4D9PIPYob0TU_JEbwtLyY/view?usp=sharing) (sorry for bad quality, but it should be enough)

FamousPig:
Fixes #10210

Previously hovering and right clicks were passed through the customisation menu to buttons behind it. Left clicks were and are absorbed by the `OsuScrollContainer` inside `ModSettingsContainer`. Mouse clicks and hover events are now absorbed by `ModSettingsContainer`.

I'm honestly not sure whether the way I have implemented the testing is correct or too hacky, if there is an easier way and if testing is even worth it for this issue.


osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -121,9 +141,21 @@ private class TestModSelectOverlay : ModSelectOverlay
 
             public bool ButtonsLoaded => ModSectionsContainer.Children.All(c => c.ModIconsLoaded);
 
+            public ModButton GetModButton(Mod mod)
+            {
+                return ModSectionsContainer.Children.Single(s => s.ModType == mod.Type)
```
bdach, line 146: Going through `Children` and `ButtonsContainer` can probably be skipped. `ChildrenOfType<>` works recursively across the entire subtree.

osu.Game.Tests/Visual/UserInterface/TestSceneModSettings.cs
```diff
@@ -121,9 +141,21 @@ private class TestModSelectOverlay : ModSelectOverlay
...
+                                           .ButtonsContainer.OfType<ModButton>().Single(b => b.Mods.Any(m => m.GetType() == mod.GetType()));
+            }
+
             public void SelectMod(Mod mod) =>
-                ModSectionsContainer.Children.Single(s => s.ModType == mod.Type)
-                                    .ButtonsContainer.OfType<ModButton>().Single(b => b.Mods.Any(m => m.GetType() == mod.GetType())).SelectNext(1);
+                GetModButton(mod).SelectNext(1);
+
+            public float SetModSettingsWidth(float newWidth)
```
bdach, line 153: Return value can probably be skipped.

osu.Game/Overlays/Mods/ModSelectOverlay.cs
```diff
@@ -495,5 +495,12 @@ protected override void Dispose(bool isDisposing)
         }
 
         #endregion
+
+        protected class MouseInputAbsorbingContainer : Container
```
bdach, line 499: This is a little too generic to live here as `protected`. I'd consider extracting `ModSettingsContainer` to a separate component and apply the input handler overrides there.

bdach (changes requested):

peppy:
I've made two further changes to get this in a mergeable state:

- 24eff8c6 Rename the class to ModSettingsContainer (all surrounding variables refer to "settings", not "cusomisation")
- 3e326a92 Reworked how you are propagating state between the two containers to use bindable flow. I figured it was easier to commit this to show how it should be done than attempt to do so in a review. Let me know if you have any questions.

peppy (dismissed):

bdach (dismissed):

bdach:
I've had to resolve a merge conflict here, but thankfully it was trivial (two new tests at same position of file). Please double check if you can, though.

bdach (approved):

smoogipoo (approved):

Issue #10484, opened 2020-10-11
OsuDbContext is not disposed in some cases, causing exceptions later on.


ms2mml:
**Describe the bug:** 
There are certain cases where `OsuDbContext` is not disposed properly. This seems to happen because the `usageCompleted()` method in *DatabaseContextFactory.cs* only disposes the underlying context if either the `usage.PerformedWrite` is true, or if `usage.Errors` is non-empty. If these conditions are false, and this happens to the last "use" of a `DatabaseWriteUsage.GetForWrite()` for that thread, then the `recycleThreadContexts()` will no longer be called, and the `OsuDbContext` is never disposed of.

For me, this ends up causing a failure during the folder migration process. At the end of the `Migrate()` method, there is a call to `deleteRecursive(source)`, which fails because the `OsuDbContext` is not disposed of.

**Repro steps (Windows, debugging osu.Desktop):**
1. Delete the `%AppData%/Roaming/osu` folder for a "fresh install"
2. Set a breakpoint at line 266 of *OsuGameBase.cs*: `KeyBindingStore.Register(globalBindings);`
3. Reach that point in the code, then set a breakpoint at line 48 of *KeyBindingStore.cs*: `foreach (var insertable in group.Skip(count).Take(aimCount - count))`
4. Note that the breakpoint in step 3 is never hit, meaning no writes are done at this step.
5. Attempt to use the *Change folder location...* feature to change the osu folder. This will throw an exception when trying to delete the files at the end.

**Screenshots or videos showing encountered issue:** 
![image](https://user-images.githubusercontent.com/70926398/95679957-f47a9a00-0b8a-11eb-8d14-34ffcf40dab6.png)
![image](https://user-images.githubusercontent.com/70926398/95680048-84204880-0b8b-11eb-8dc2-8a5fa0d80172.png)

**osu!lazer version:**
Commit 8507327fa6cc6fe03d1dccc7a34231a6234a2e80 on master branch.

**Logs:**
I have added logging to the constructor and dispose method of `OsuDbContext`, as well as a guid to each instance. The following logs are from starting up osu until the main menu loads. The log contains Open/Dispose, ThreadId, ProcessId, Guid of the instance of `OsuDbContext`, and a hash of the stack trace during the call to the constructor.
```
Open		Thread: 1	Process: 20940	Guid: 26924190-3955-41bf-8e7b-232b73ec7544	ee7edb56f4816c86f502e6d025e4349b0649a1f33838f458bb2086cd62862466
[...]

PR #10496, opened 2020-10-14
Use default storage when accessing cache data

ms2mml:
Referencing the cache issue mentioned at the end of #10484.

During folder migration, the `cache` folder is ignored in migration. However, subsequent access to the cache uses `UnderlyingStorage` instead of `defaultStorage`. So even though the `cache` files aren't copied over to the migrated location, they will be recreated there, which defeats the purpose of ignoring them during migration.

In addition, the following edge case will cause an exception to be thrown.
1. Use the migration feature
2. Try to import a beatmap without exiting the game first
3. The game will throw an exception because the cache folder was not migrated, and subsequent access to the cache is trying to access the cache under the migrated path instead of the default path:
image
4. Upon restarting the game, the cache folder will be recreated under the migrated path

Stack trace for the exception above:
```
at System.IO.Enumeration.FileSystemEnumerator`1.CreateDirectoryHandle(String path, Boolean ignoreNotFound)
at System.IO.Enumeration.FileSystemEnumerator`1.Init()
at System.IO.Enumeration.FileSystemEnumerator`1..ctor(String directory, Boolean isNormalized, EnumerationOptions options)
at System.IO.Enumeration.FileSystemEnumerable`1..ctor(String directory, FindTransform transform, EnumerationOptions options, Boolean isNormalized)
at System.IO.Enumeration.FileSystemEnumerableFactory.UserFiles(String directory, String expression, EnumerationOptions options)
at System.IO.Directory.InternalEnumeratePaths(String path, String searchPattern, SearchTarget searchTarget, EnumerationOptions options)
at System.IO.Directory.GetFiles(String path, String searchPattern)
at osu.Framework.Platform.NativeStorage.GetFiles(String path, String pattern) in D:\src\vs\osu-framework\osu.Framework\Platform\NativeStorage.cs:line 45
at osu.Game.IO.WrappedStorage.GetFiles(String path, String pattern) in D:\src\vs\osu\osu.Game\IO\WrappedStorage.cs:line 68
at osu.Game.IO.WrappedStorage.GetFiles(String path, String pattern) in D:\src\vs\osu\osu.Game\IO\WrappedStorage.cs:line 68
at osu.Framework.IO.Stores.RawCachingGlyphStore.createCachedPageInfo(Int32 page) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\RawCachingGlyphStore.cs:line 56
at osu.Framework.IO.Stores.RawCachingGlyphStore.LoadCharacter(Character character) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\RawCachingGlyphStore.cs:line 40
at osu.Framework.IO.Stores.GlyphStore.Get(String name) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\GlyphStore.cs:line 128
at osu.Framework.IO.Stores.ResourceStore`1.Get(String name) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\ResourceStore.cs:line 131
at osu.Framework.Graphics.Textures.TextureStore.getTexture(String name, WrapMode wrapModeS, WrapMode wrapModeT) in D:\src\vs\osu-framework\osu.Framework\Graphics\Textures\TextureStore.cs:line 54
at osu.Framework.Graphics.Textures.TextureStore.Get(String name, WrapMode wrapModeS, WrapMode wrapModeT) in D:\src\vs\osu-framework\osu.Framework\Graphics\Textures\TextureStore.cs:line 132
at osu.Framework.IO.Stores.FontStore.Get(String name) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\FontStore.cs:line 117
at osu.Framework.IO.Stores.FontStore.Get(String fontName, Char character) in D:\src\vs\osu-framework\osu.Framework\IO\Stores\FontStore.cs:line 144
at osu.Framework.Graphics.Sprites.SpriteText.load(ShaderManager shaders) in D:\src\vs\osu-framework\osu.Framework\Graphics\Sprites\SpriteText.cs:line 77
```

### DIFF
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -57,10 +57,8 @@ namespace osu.Game.IO
                 TryChangeToCustomStorage(out Error);
         }
 
-        public Storage GetStorageFromPath(string path)
-        {
-            return IGNORE_DIRECTORIES.Any(x => path.StartsWith(x)) ? defaultStorage : UnderlyingStorage;
-        }
+        public Storage GetStorageFromPath(string path) =>
+            IGNORE_DIRECTORIES.Any(x => path.StartsWith(x)) ? defaultStorage : UnderlyingStorage;
 
         public override string GetFullPath(string path, bool createIfNotExisting = false) =>
             GetStorageFromPath(path).GetFullPath(MutatePath(path), createIfNotExisting);
--- a/osu.Game/IO/OsuStorage.cs
+++ b/osu.Game/IO/OsuStorage.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


peppy:
Thanks for the contribution!

My first thoughts here are that maybe we want to expose a `CreateCacheStorage` at a framework level (which defaults to just calling `CreateStorage` and completely separate that flow out. It miight be a better option as we have more components that might use cache, so they don't need to know about the magic "cache" directory and can instead use `game.CacheStorage` directly.

ms2mml:
> Thanks for the contribution!
> 
> My first thoughts here are that maybe we want to expose a `CreateCacheStorage` at a framework level (which defaults to just calling `CreateStorage` and completely separate that flow out. It miight be a better option as we have more components that might use cache, so they don't need to know about the magic "cache" directory and can instead use `game.CacheStorage` directly.

Sure, I think having a framework level cache storage is a better idea as well. I will work on making the update.

peppy:
@ms2mml just checking in on whether you're still going to look at this. if not i may make the framework-side amendment for you just to move things along.

bdach:
I guess there's not much merit to these changes at this point, given that the framework change was merged?

peppy:
In *theory* this should not be required any more. Would be good to test and make sure the failure cannot be repro'd.

bdach:
I tested and as expected the `cache` directory is no longer part of any custom storage directory after migration, which means that the issue should be fixed.

Closing this PR as superseded by the framework change at ppy/osu-framework#3977.

PR #10502, opened 2020-10-14
Confine available area for HUD components to exclude the song progress area

peppy:
This change bring the available area for main HUD components to exclude the song progress region. Generally improves the relative behaviour of components as can be seen:

This is a prerequisite for legacy skinning of HUD components (ie. allowing the combo counter to exist at the bottom-left corner).

For simplicity I'm just using every-update height propagation. `SongProgress` could support `AutoSize` with some work but I don't think it's worth it at this point in time.


bdach (approved):
Looks like an OK resolution :+1:

smoogipoo (approved):

PR #10504, opened 2020-10-14
Add legacy skinning support for combo counter

peppy:
Initial pass, mostly visual (sizing may not match perfectly). Note that I'm inheriting some *really* shocking code here (`ComboCounter` class). Did my best to clean it up but it could probably do with more work in the future, if we ever have to touch it again.

- [x] Depends on #10502


smoogipoo (approved):
I think this is fine.

PR #10505, opened 2020-10-14
Move ISampleDisabler implementation to Player and FrameStabilityContainer

### LINKED ISSUE #10454: The spinner sound is looped even if paused

peppy:
I've left the logic inside the clocks for now, but moved the updating to outside. Hopefully this feels a bit better.

The implementation at `Player` is required for storyboards, which are not inside the `FrameStabilityContainer`.

- Resolves #10454 
- Alternative to (and will close #10476 if merged)


bdach (approved):
Yeah I think I could live with this too. @smoogipoo thoughts?

smoogipoo (approved):
I definitely like this a lot more

PR #10508, opened 2020-10-15
Add legacy skinning support for score display

peppy:
- [x] Depends on #10504 

- Changes top-right aligned components to respect the height of the score display (to avoid overlap)
- Fixes a couple of issues with the top-right items.

Note that this display isn't fixed-width due to a framework oversight that makes it impossible to use for now (fixed-width width is computed by the size of `m`, which is not present in skins).


osu.Game/Screens/Play/HUDOverlay.cs
```diff
@@ -214,6 +205,8 @@ protected override void LoadComplete()
         protected override void Update()
         {
             base.Update();
+
+            topRightElements.Y = ToLocalSpace(ScoreCounter.Drawable.ScreenSpaceDrawQuad.BottomRight).Y;
```
smoogipoo, line 209: I'm not a huge fan of this - it's now being done in two places, but let's go with it for now.

smoogipoo (approved):

osu.Game/Screens/Play/HUDOverlay.cs
peppy: wait until you see the next PR! but yeah, this will be replaced over time with something more appeasing to the eyes.

Issue #5331, opened 2019-07-11
Disable screen rotation while in active gameplay


ticotaco72:
**Describe the new feature:**  while in gameplay, rotating your device should be blocked on Android and ios

Joehuu:
I think the OS rotation lock is enough? This will cause a UX problem if we disable rotation in certain screens (i.e. gameplay), or settings bloat if we make it toggleable.

Also to note, I don't see a gameplay benefit with rotating in the middle of play, only a disadvantage if screen accidentally rotates (which OS rotation lock is there for).

peppy:
This has recently been fixed on android to respect the user rotation lock setting. It has always done so on iOS.

Agree that we may want to block rotation during gameplay (when `LocalUserActive`) though.

frenzibyte:
This turned out to be extremely convoluted for multiple reasons, in a cascading manner:
 - UIKit does not support direct API to lock current screen orientation
 - UIKit allows controlling the allowed screen orientations dynamically by overriding `supportedInterfaceOrientations` and returning a different mask if the game is requesting a lock, but I can't directly override on the root view controller class as it is prepared and provided by SDL.
 - SDL overrides `rootViewController.supportedIntefaceOrientations` and reads the hint value of `SDL_HINT_ORIENTATIONS` there, so sounds like we're winning right? we can just set the hint whenever we want and it'll just work...right? So that actually just doesn't work, whatever value I set there is still bypassed and the game can still freely rotate around.
 - Random google searches show that we can control the orientation on a higher level by overriding `UIApplicationDelegate.supportedInterfaceOrientationsForWindow:`, but...since we migrated to SDL, we can no longer override any method of the app delegate, as the app delegate is privately set up by SDL.

I can double down and start inventing Obj-C weirdness to make SDL use a custom app delegate subclass we provide, but that is not simple to achieve at all. As such, I've just un-assigned as not worth the time/effort for now.

peppy:
I'm guessing now that you have an `AppDelegate` again this becomes more feasible?

frenzibyte:
Yeah I'll look into this again.

PR #10516, opened 2020-10-15
Lock screen rotation while in gameplay on android

Game4all:
# Summary

This adds a `GameplayScreenRotationLocker` component which locks screen rotation on android devices while in gameplay.

Unfortunately, android (same for Xamarin) doesn't expose a nice, non-hackish way to get the current activity from anywhere (which is required to lock the screen rotation) so I had to resort to storing the game activity inside a static field upon startup.

# Testing

No automated testing for this altough from the testing I've done on 3 devices I own, this seems to work fine.

### DIFF
--- a/osu.Android/GameplayScreenRotationLocker.cs
+++ b/osu.Android/GameplayScreenRotationLocker.cs
@@ -17,7 +17,7 @@ namespace osu.Android
         private void load(OsuGame game)
         {
             localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
-            localUserPlaying.BindValueChanged(userPlaying => updateLock(userPlaying));
+            localUserPlaying.ValueChanged += updateLock;
         }
 
         private void updateLock(ValueChangedEvent<bool> userPlaying)
--- a/osu.Android/GameplayScreenRotationLocker.cs
+++ b/osu.Android/GameplayScreenRotationLocker.cs
@@ -1,0 +1,31 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using Android.Content.PM;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Game;
+
+namespace osu.Android
+{
+    public class GameplayScreenRotationLocker : Component
+    {
+        private Bindable<bool> localUserPlaying;
+
+        [BackgroundDependencyLoader]
+        private void load(OsuGame game)
+        {
+            localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
+            localUserPlaying.BindValueChanged(_ => updateLock());
+        }
+
+        private void updateLock()
+        {
+            OsuGameActivity.Activity.RunOnUiThread(() =>
+            {
+                OsuGameActivity.Activity.RequestedOrientation = localUserPlaying.Value ? ScreenOrientation.Locked : ScreenOrientation.FullUser;
+            });
+        }
+    }
+}
--- a/osu.Android/GameplayScreenRotationLocker.cs
+++ b/osu.Android/GameplayScreenRotationLocker.cs
@@ -20,7 +20,7 @@ namespace osu.Android
         private void load(OsuGame game)
         {
             localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
-            localUserPlaying.ValueChanged += updateLock;
+            localUserPlaying.BindValueChanged(updateLock, true);
         }
 
         private void updateLock(ValueChangedEvent<bool> userPlaying)
--- a/osu.Android/GameplayScreenRotationLocker.cs
+++ b/osu.Android/GameplayScreenRotationLocker.cs
@@ -17,14 +17,14 @@ namespace osu.Android
         private void load(OsuGame game)
         {
             localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
-            localUserPlaying.BindValueChanged(_ => updateLock());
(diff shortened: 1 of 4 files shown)


bdach:
I'm not sure we should be doing this. We already changed the activity to use `FullUser` so that OS-level screen rotation can be preserved in-game. And even if we were to do this forcing landscape would probably be saner.

osu.Android/OsuGameActivity.cs
```diff
@@ -12,10 +12,14 @@ namespace osu.Android
     [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false)]
     public class OsuGameActivity : AndroidGameActivity
     {
+        internal static Activity Activity;
```
bdach, line 15: also at least make this `private set` or something

Joehuu:
For reference, this will fix https://github.com/ppy/osu/issues/5331 for android.

osu.Android/GameplayScreenRotationLocker.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
+            localUserPlaying.BindValueChanged(_ => updateLock());
+        }
+
+        private void updateLock()
+        {
+            OsuGameActivity.Activity.RunOnUiThread(() =>
+            {
+                OsuGameActivity.Activity.RequestedOrientation = localUserPlaying.Value ? ScreenOrientation.Locked : ScreenOrientation.FullUser;
```
swoolcock, line 27: Would prefer that you pass `evt.NewValue` into `updateLock`, since `localUserPlaying.Value` could potentially have changed.  As it stands, the current value of the bindable is not being captured in the lambda.

osu.Android/OsuGameAndroid.cs
```diff
@@ -55,6 +55,12 @@ public override Version AssemblyVersion
             }
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            LoadComponentAsync(new GameplayScreenRotationLocker(), Add);
```
swoolcock, line 61: Is there a reason this needs to be loaded async here?  Can you not just cache it like any other component?

Game4all: There isn't a real reason for this, apart from trying to do things consistently with desktop

osu.Android/GameplayScreenRotationLocker.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    public class GameplayScreenRotationLocker : Component
+    {
+        private Bindable<bool> localUserPlaying;
+
+        [BackgroundDependencyLoader]
+        private void load(OsuGame game)
+        {
+            localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
+            localUserPlaying.BindValueChanged(userPlaying => updateLock(userPlaying));
```
swoolcock, line 20: Change to `localUserPlaying.ValueChanged += updateLock;`

bdach: Method group would also work. I'd slightly prefer that to an event subscription actually but unsure if it matters.

swoolcock: `BindValueChanged` exists mostly to allow passing `true` as the second argument to run it immediately.
The first line is literally `ValueChanged += onChange;`

swoolcock:
Would be nice to implement this for iOS too, even if it's a separate PR.

Game4all:
As an android user, I'll have to delegate implementation of this for iDevices as I don't own any of these (and lack Xamarin.iOS knowledge as well)

osu.Android/OsuGameActivity.cs
```diff
@@ -12,10 +12,14 @@ namespace osu.Android
     [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false)]
     public class OsuGameActivity : AndroidGameActivity
     {
+        internal static Activity Activity { get; private set; }
```
bdach, line 15: Just occurred to me that this `static` could potentially be dispensed with if the activity was passed to `OsuGameAndroid` via ctor, which then would cache the activity to whoever needs to use it. I'd probably prefer that from a code quality standpoint, if feasible.

osu.Android/GameplayScreenRotationLocker.cs
Game4all: Actually, I went with `BindValueChanged` as I thought we might want to run callback immediately in this scenario but I'm not sure it is actually necessary anymore.

osu.Android/GameplayScreenRotationLocker.cs
```diff
@@ -0,0 +1,34 @@
...
+        private Bindable<bool> localUserPlaying;
+
+        [Resolved]
+        private OsuGameActivity gameActivity { get; set; }
+
+        [BackgroundDependencyLoader]
+        private void load(OsuGame game)
+        {
+            localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
+            localUserPlaying.ValueChanged += updateLock;
```
smoogipoo, line 23: Not a huge issue but it's probably worthwhile to have this be `BindValueChanged(updateLock, true)` just to get it into a correct initial state.

smoogipoo (changes requested):

smoogipoo (approved):
Looks fine to me. @peppy could you test on your devices?

peppy:
Works as expected

PR #10521, opened 2020-10-16
Fix editor not seeking by full beat when track is playing

### LINKED ISSUE #10519: Make the scrolling speed in editor for 1/4 snap the same as in stable
**Describe the new feature:** 
So, right now in osu!lazer when you use the arrows to scroll through the timeline in editor mode while on 1/4 snap, it's really slow, it gets kinda annoying to map because I like to walk through the song using the keyboard arrows. If you do quick comparison, in stable it's quick while on 1/4 snap! In lazer, it's only quick if I use 1/1 snap divisor, which is kinda counter-intuitive imo.

peppy:
This is expected behaviour as my osu-stable, and I still stand behind the reasoning behind it.

Closes #10519.

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -597,10 +597,20 @@ namespace osu.Game.Screens.Edit
         {
             double amount = e.ShiftPressed ? 4 : 1;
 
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
Seems to work as expected

PR #10522, opened 2020-10-16
Add support for ScorePrefix and ScoreOverlap values in legacy skins

Morilli:
Legacy skins now use the values specified in `ScorePrefix` and `ScoreOverlap` for the score and accuracy display.

From my basic testing this does what it should.


osu.Game/Skinning/LegacyAccuracyCounter.cs
```diff
@@ -24,16 +27,19 @@ public LegacyAccuracyCounter(ISkin skin)
             Margin = new MarginPadding(10);
 
             this.skin = skin;
+            scorePrefix = skin.GetConfig<LegacySkinConfiguration.LegacySetting, string>(LegacySkinConfiguration.LegacySetting.ScorePrefix)?.Value ?? "score";
```
peppy, line 30: This is probably going to be used in more places than just here, so I guess we should revert to doing things the way I had them:

rather than `new LegacySpriteText` below, it should be `skin.GetDrawable(new HUDComponent(ScoreText)`.

then the config lookups etc. would be inside that lookup. you can see the logic in the left hand of the diff here: https://github.com/ppy/osu/pull/10504/commits/37e9f331ad78fb9ff10701888169e5ea47ddea7a

peppy (changes requested):

osu.Game/Skinning/LegacyAccuracyCounter.cs
Morilli: I'm not sure I understand. Are you suggesting to move the `GetConfig` calls into `LegacySkin.GetDrawableComponent`, which is then called by the ``CreateSpriteText()`` method? If so, would `GetDrawableComponent` have to create and return that `LegacySpriteText` (which I'd also have to cast to a `OsuSpriteText`)?
I'm kinda confused right now.

peppy: The linked commit should explain quite well (it's a full implementation of what I'm asking for here).

Morilli: Okay, I've implemented what I understood from your comment. Please tell me if this isn't what you asked for.

peppy:
Tidied up a bit and tested it's working, looks good.

peppy (dismissed):

peppy (approved):

PR #10528, opened 2020-10-16
Add legacy skinning support for health display

peppy:
- [x] Depends on https://github.com/ppy/osu-resources/pull/115 for default legacy skin
- [x] osu!resources deploy.


smoogipoo:
I think this is an ok start.

The biggest issue I have (but it's not a blocker) is that the fill shouldn't be stretched, but should be masked instead:

lazer:

image
image

stable:

image

(skin: https://osu.ppy.sh/community/forums/topics/797025)

Which causes some awkward visuals like in my skin (left side should be rounded):

image

(skin: https://drive.google.com/file/d/19BwcK6fj5PTQY0fwi02-s3F9vrAvPPjR/view)

peppy:
Uh that's weird, i made it to be masked... How does it work in my screenshots? it's clearly masked there.

smoogipoo:
Uh... Ignore all of that. Somehow I was on an old commit o_O

smoogipoo (approved):

PR #10537, opened 2020-10-16
Fix legacy score & combo counter text overflowing outside padding

### LINKED ISSUE #10536: LegacyScoreCounter text overflows outside padding
**Describe the bug:** This causes the score display to jitter too much which wasn't present on stable. Seeing it is has `Anchor` and `Origin` set to `Anchor.TopRight` it shouldn't really be an issue. I'm unsure if this is a framework-level issue or it could be something be fixed by using a `TextFlowContainer`. Making this issue for visibility.

image

**osu!lazer version:** master

LeNitrous:
Fixes https://github.com/ppy/osu/issues/10536. `LegacyScoreCounter` has `Anchor.TopRight` on `Origin` and `Anchor`. However it's child still has `Anchor.TopLeft` on both properties causing it to jitter violently and its text overflowing outside padding during gameplay.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableScoreCounter.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneSkinnableScoreCounter.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
+using osu.Framework.Extensions.IEnumerableExtensions;
 using osu.Framework.Testing;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Osu;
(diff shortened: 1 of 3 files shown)


Morilli:
Oops, that is my bad. I guess this should be done in `LegacyAccuracyCounter` as well then.

bdach:
Added visual test coverage and applied the same solution to `LegacyAccuracyCounter` as per @Morilli's note. Not sure how you'll see my approach to line continuations in a5b0307. I think it looks passable and really didn't want it to be this one huge line, but I'll change on request.

Morilli:
Hm, looking at how this affects the score (and accuracy) counter, there are still those "jitters" you mentioned in the issue visible. This is probably due to texture files potentially having differing sizes (the stable skin does, for example).
osu!stable seems to have a fixed size for each letter, which does look way better, so I'm thinking lazer should probably try and do the same eventually?

bdach:
adding back constant width is a bit more involved and requires framework-side changes. should be a separate PR as such.

smoogipoo (approved):

PR #10556, opened 2020-10-19
Allow legacy text to display fixed width correctly

peppy:
- [x] Depends on https://github.com/ppy/osu-framework/pull/3945
- [x] Depends on #10493


Joehuu:
Just noting, stable's combo counter is not fixed width. I'm fine either way.
IXwLfoI16P

peppy:
Probably should be. I guess we'll change it and see if/when people complain as usual.

smoogipoo (approved):

PR #10563, opened 2020-10-19
Fix autoplay/replay settings going off screen on some legacy skins

### LINKED ISSUE #10538: Some skins have problem with UI scaling and don't have the options in Autoplay
**Describe the bug:** 
Some skins have next problems:
- Don't scale properly
- Offset on autoplay options (huge)
UI scaling maybe, but I'm not sure

**Scale**: 1x Works fine
osu_2020-10-16_12-10-24
**Scale**: 1.15x (has some problems on the overlay)
osu_2020-10-16_12-09-55
**This skin doesn't have autoplay options**
osu_2020-10-16_12-12-49

**Skins**:
http://www.mediafire.com/file/l3573ohbh1xm9l8/skins.rar/file

**osu!lazer version:** 

2020.1017.0

Joehuu:
Fixes https://github.com/ppy/osu/issues/10538.

See https://github.com/ppy/osu/issues/10538#issuecomment-711591371 for why I did this.

Similar to:
https://github.com/ppy/osu/blob/b9c1d21743087e253321debef5018bfb30344088/osu.Game/Skinning/LegacyAccuracyCounter.cs#L36-L45

### DIFF
--- a/osu.Game/Screens/Play/HUDOverlay.cs
+++ b/osu.Game/Screens/Play/HUDOverlay.cs
@@ -208,17 +208,9 @@ namespace osu.Game.Screens.Play
         {
             base.Update();
 
-            float topRightOffset = 0;
+            // for now align with the accuracy counter. eventually this will be user customisable.
+            topRightElements.Y = ToLocalSpace(AccuracyCounter.Drawable.ScreenSpaceDrawQuad.BottomRight).Y;
 
-            // fetch the bottom-most position of any main ui element that is anchored to the top of the screen.
-            // consider this kind of temporary.
-            foreach (var d in mainUIElements)
-            {
-                if (d is SkinnableDrawable sd && (sd.Drawable.Anchor & Anchor.y0) > 0)
-                    topRightOffset = Math.Max(sd.Drawable.ScreenSpaceDrawQuad.BottomRight.Y, topRightOffset);
-            }
-
-            topRightElements.Y = ToLocalSpace(new Vector2(0, topRightOffset)).Y;
             bottomRightElements.Y = -Progress.Height;
         }
 
--- a/osu.Game/Screens/Play/HUDOverlay.cs
+++ b/osu.Game/Screens/Play/HUDOverlay.cs
@@ -16,7 +16,6 @@ using osu.Game.Rulesets.Mods;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Rulesets.UI;
 using osu.Game.Screens.Play.HUD;
-using osu.Game.Skinning;
 using osuTK;
 using osuTK.Input;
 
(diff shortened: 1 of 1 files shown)


Joehuu:
Guess it's time to get a proper IDE as I keep forgetting to run the script or let appveyor run on my repo before I PR.

bdach:
Sure, I think this makes sense to me. While it looks a bit wonky if someone hacks the percent sign to center the accuracy meter, that's already quite the abuse of the skinning system, so... As a temporary measure this surely can be good enough.

bdach (dismissed):

peppy:
This is actually incorrect for the default skin now, since the score is lower than the accuracy. Also it doesn't fix the fact that we're loading and displaying health bars that cover the whole screen (shouldn't be a thing)

Joehuu:
Just didn't want to complicate *temporary* fixes. Still works with all the arbitrary margin/padding on the top right elements. Can make it more right if we want too though.

640x480 @ 1.60x scale (max)
osu!_gViIcgOHjG

osu!_T3yVWo7xTR

> These skins will be displaying stuff on screen at locations the user is not expecting.

Sure, all the skinning workarounds will break on aspect ratios that were not supported by the skin. Toggling the toolbar will also break the skin even if at the supported aspect ratio.

With trimming, existing stable users who play lazer and use these kinds of skins will start reacting (as mentioned in the issue comment).

We can advise lazer users to remove the workaround health display playfield border if the legacy skin had an option and use lazer's playfield border once it's implemented and user customizable. If not, advise skinners to provide one without the workarounds.

This is all a transition thing between legacy and the new skinning system with much more customizability that I don't want to be involved in. If you think this isn't the best course of action and want to influence new or old legacy skins to not do such workarounds with such tactics as trimming, that should be another issue to deal with.

peppy:
The code I had there was not a temporary fix though, it was my intended path forward (all top anchored elements would be used to find the correct offset for "main area" content).

This is removing that direction so at very least that needs to be considered. Maybe a comment left in place suggesting why it is removed.

peppy (approved):
VERY reluctantly going with this for now.

Joehuu:
I wouldn't consider it a hack. Health displays can be unusually taller than the score and acc counters even without a playfield border hack.

I don't think this is abuse in skinning.

Ref: https://skins.osuck.net/index.php?newsid=1464

This is release 2020.1017.0.

osu_2020-10-19_20-48-20

peppy:
From the person that made the skinning system, I consider this abuse. But call it what you will. In lazer we will likely define skinning restrictions a lot better so this kind of thing cannot happen in the first place.

bdach:
For what it's worth I was aware this changed behaviour in the way described on the default skin, but I also checked visually and it looked fine (although I did not mention that clearly in the review).

PR #10573, opened 2020-10-21
Fix ISourcedFromTouch events being blocked by LoadingLayer

peppy:
Noticed this on iOS (log out -> details at song select behind settings overlay). Blocking until I can test on hardware (seems to be correct in simulator, but fps is so low I want to make certain).

Open to suggestions for a better method of providing this behaviour.


bdach:
If I'm understanding correctly, a plausible way forward might be to add a framework option to allow drawables to say whether an event sourced from touch should be raised if a touch is handled, but it's a sort of muddy thing to add/understand... Shoehorning that into `Handle()` will also be... unpleasant, I imagine.

peppy:
Have tested and confirmed this fixes the behaviour.

smoogipoo (approved):

PR #10016, opened 2020-08-31
Add star rating to beatmap wedge

smoogipoo:
Resolves https://github.com/ppy/osu/issues/9412

This currently doesn't update with the beatmap's modded/ruleset difficulty, I'll comment why that is below.

Previews:
Screenshot_2020-08-31_15-03-23

Screenshot_2020-08-31_15-03-02

Screenshot_2020-08-31_15-02-46

PR #10576, opened 2020-10-21
Make beatmap wedge difficulty indicators update dynamically

Game4all:
# Summary

This makes the beatmap wedge difficulty indicators on song select update dynamically according to the currently selected mods.

Note that this is calling `updateDisplay()` when the star rating changes instead of calling `ForceRedraw()` for aesthethic reasons:

 `updateDisplay()` adds a slight visual transition making the change more pleasant visually whereas the change of color / star rating on difficulty indicators when just calling `ForceRedraw()` looks very crude. 

## Visual Demo
https://streamable.com/j2k5ye

### DIFF
--- a/osu.Game/Screens/Ranking/Expanded/StarRatingDisplay.cs
+++ b/osu.Game/Screens/Ranking/Expanded/StarRatingDisplay.cs
@@ -22,18 +22,7 @@ namespace osu.Game.Screens.Ranking.Expanded
     /// </summary>
     public class StarRatingDisplay : CompositeDrawable
     {
-        private readonly BeatmapInfo beatmap;
-
-        private StarDifficulty? difficulty;
-
-        /// <summary>
-        /// Creates a new <see cref="StarRatingDisplay"/>.
-        /// </summary>
-        /// <param name="beatmap">The <see cref="BeatmapInfo"/> to display the star difficulty of.</param>
-        public StarRatingDisplay(BeatmapInfo beatmap)
-        {
-            this.beatmap = beatmap;
-        }
+        private readonly StarDifficulty difficulty;
 
         /// <summary>
         /// Creates a new <see cref="StarRatingDisplay"/> using an already computed <see cref="StarDifficulty"/>.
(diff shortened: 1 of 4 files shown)


bdach:
I don't think this is the proper resolution, which is most of the reason why this hasn't been done already. Calling `updateDisplay()` nullifies the benefits of having the wedge be a buffered container, creating a heavy drawable hierarchy each time.

In fact, @smoogipoo has already attached one patch attempting to resolve this [here](https://github.com/ppy/osu/pull/10016#issuecomment-683601685), which was seemingly held up by the buffered container not behaving as expected.

Game4all:
I hadn't assessed the overhead of reconstructing the wedge on every star rating update yeah ...

As a sidenote, I Initially wanted to make only the relevant parts update (and call `ForceRedraw()`) and I ran into that very same issue with the buffered container and the text flow container, which I somehow resolved by replacing the textflow with two sprite texts but the lack of proper visual transition due to the wedge using a buffered container of really put me off continuing that way as it was looking really crude... Unless transitions aren't an immediate concern ..

bdach:
This has been a problem before in other places. [I've suggested before in the past](https://github.com/ppy/osu/pull/7540#discussion_r376642332) that dynamic components that use transforms should be taken out of the wedge to prevent reconstruction but allow animations, but unsure if we want to go that way.

smoogipoo:
I hadn’t really thought of animations yet, but at a cursory glance I like the direction of this PR much more than my patch. Plus it resolves the animation.

I wouldn’t be concerned about dynamic difficulty here - if the beatmap is selected dynamic difficulty is computed by the difficulty icon and carousel star displays anyway. It’s all debounced/cached.

Game4all:
FWIW , the overhead of reconstructing the wedge isn't really noticeable on release builds but it can sometimes be noticeable in debug when by example selecting a few mods then hitting deselect all button (I can't however get to reproduce that consistently).

osu.Game/Screens/Ranking/Expanded/StarRatingDisplay.cs
```diff
@@ -24,27 +24,42 @@ public class StarRatingDisplay : CompositeDrawable
     {
         private readonly BeatmapInfo beatmap;
 
+        private StarDifficulty? difficulty;
```
bdach, line 27: Not entirely sure I like both these fields existing, but I also don't see a good way to eliminate either, so...

osu.Game/Screens/Ranking/Expanded/StarRatingDisplay.cs
```diff
@@ -24,27 +24,42 @@ public class StarRatingDisplay : CompositeDrawable
...
 
         [BackgroundDependencyLoader]
-        private void load(OsuColour colours)
+        private void load(OsuColour colours, BeatmapDifficultyManager difficultyManager)
         {
-            var starRatingParts = beatmap.StarDifficulty.ToString("0.00", CultureInfo.InvariantCulture).Split('.');
+            AutoSizeAxes = Axes.Both;
+
+            if (!difficulty.HasValue)
+                difficulty = difficultyManager.GetDifficulty(beatmap);
```
bdach, line 53: can write
```suggestion
            difficulty ??= difficultyManager.GetDifficulty(beatmap);
```

bdach (commented):
I guess if nobody else minds the buffered drawable reconstruction then I'd be fine with it too. Still not sure it's strictly necessary to get a good visual effect, though.

peppy:
I've removed the `BeatmpInfo` based constructor and in the process also made the star difficulty displayed on score cards actually reflect the locally applied mods. Feels pretty good.

peppy (dismissed):

peppy (approved):

PR #10579, opened 2020-10-22
Remove IOnlineComponent and change existing components to use bindable flow

peppy:
In a distant past we thought it was a good idea to maintain thread safety of api functions by exposing a limited event-based API that would ensure actions were scheduled back to the Update thread (via `APIAccess`'s `Scheduler`.

In practice, this has caused issues such as:
- Changing the expected order of execution, as the scheduled code is performed when the API component is updated, not the target component.
- Potential for references to be held to components which are already disposed (short term)
- Potential for code to be run on a drawable post-disposal of that drawable, requiring local protection against this in some cases.

In addition, this avoids the potential of registering an API component but never unregistering. By using bindables the whole way, we can be ensured that binding clean-up is automatic.

Considering the above, it seems better to leave the schedule responsibilities to the consumer. This is already what we do for other properties on `IAPIProvider` (like `LocalUser`) so there is already precedent for it.


osu.Game/Overlays/Toolbar/ToolbarUserButton.cs
```diff
@@ -44,16 +50,17 @@ public ToolbarUserButton()
...
         {
-            api.Register(this);
+            apiState.BindTo(api.State);
+            apiState.BindValueChanged(onlineStateChanged, true);
 
             StateContainer = login;
         }
 
-        public void APIStateChanged(IAPIProvider api, APIState state)
+        private void onlineStateChanged(ValueChangedEvent<APIState> state)
```
smoogipoo, line 61: Missing schedule

smoogipoo (changes requested):

smoogipoo (approved):

PR #10586, opened 2020-10-23
Add Seasonal backgrounds

### LINKED ISSUE #7055: Seasonal backgrounds
**Describe the new feature:** It would be exactly like Stable in the sense that with every season, you can enable seasonal backgrounds instead of the default background when you open Lazer. Different from Stable, however, each picture would be present for, lets say 10 seconds, then it changes to a new picture. So it basically constantly refreshes the pictures instead of staying on just one picture.

**Proposal designs of the feature:** Button under skinning? Maintenance? Since I have a Chromebook, I can't go back to the Stable client to remember where that button was on that client to know where to put it in Lazer.

hbnrmx:
Resolves #7055

PRing this as a first draft because I would like some assistance with the remaining issues.

1. stable gives the user the option Seasonal Backgrounds: `Always` / `Sometimes` (default) / `Never`
Not sure if this is worth keeping around in lazers Background source. For now, I only added `Seasonal` (default) 

1. we definitely need some way of caching or prefetching, but I would like some guidance on which level to do that.

1. In `BackgroundScreenDefault` I had to change the initial load to `LoadComponentAsync`
 because of `[LongRunningLoad]` but chaining `LoadComponentAsync`s like that looks very wrong to me.

### DIFF
--- a/osu.Game/Configuration/SeasonalBackgrounds.cs
+++ b/osu.Game/Configuration/SeasonalBackgrounds.cs
@@ -1,0 +1,12 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Configuration
+{
+    public enum SeasonalBackgrounds
+    {
+        Always,
+        Sometimes,
+        Never
+    }
+}
--- a/osu.Game/Configuration/SessionStatics.cs
+++ b/osu.Game/Configuration/SessionStatics.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
-using System.Collections.Generic;
 using osu.Game.Online.API.Requests.Responses;
 
 namespace osu.Game.Configuration
@@ -16,8 +14,7 @@ namespace osu.Game.Configuration
         {
             Set(Static.LoginOverlayDisplayed, false);
             Set(Static.MutedAudioNotificationShownOnce, false);
-            Set(Static.SeasonEndDate, DateTimeOffset.MinValue);
-            Set(Static.SeasonalBackgrounds, new List<APISeasonalBackground>());
+            Set<APISeasonalBackgrounds>(Static.SeasonalBackgroundsResponse, null);
         }
     }
 
@@ -25,7 +22,6 @@ namespace osu.Game.Configuration
     {
         LoginOverlayDisplayed,
         MutedAudioNotificationShownOnce,
-        SeasonEndDate,
-        SeasonalBackgrounds,
+        SeasonalBackgroundsResponse,
     }
 }
--- a/osu.Game/Configuration/SessionStatics.cs
+++ b/osu.Game/Configuration/SessionStatics.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
 using System.Collections.Generic;
 using osu.Game.Online.API.Requests.Responses;
 
@@ -15,6 +16,7 @@ namespace osu.Game.Configuration
         {
             Set(Static.LoginOverlayDisplayed, false);
             Set(Static.MutedAudioNotificationShownOnce, false);
+            Set(Static.SeasonEndDate, DateTimeOffset.MinValue);
             Set(Static.SeasonalBackgrounds, new List<APISeasonalBackground>());
         }
     }
@@ -23,6 +25,7 @@ namespace osu.Game.Configuration
     {
         LoginOverlayDisplayed,
         MutedAudioNotificationShownOnce,
-        SeasonalBackgrounds
+        SeasonEndDate,
+        SeasonalBackgrounds,
     }
 }
--- a/osu.Game/Configuration/SessionStatics.cs
+++ b/osu.Game/Configuration/SessionStatics.cs
@@ -22,6 +22,11 @@ namespace osu.Game.Configuration
     {
         LoginOverlayDisplayed,
         MutedAudioNotificationShownOnce,
+
+        /// <summary>
+        /// Info about seasonal backgrounds available fetched from API - see <see cref="APISeasonalBackgrounds"/>.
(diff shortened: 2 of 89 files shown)


bdach:
No offence, but I think it would have been wiser to ask your questions in an issue comment. Fetching the images every time is pretty shocking.

Here's a proposal of a completely different flow instead:

* On game load, check if there are seasonal backgrounds available
* If there are new backgrounds, fetch them all at once in the background and persist them to storage, then display a notification to the user prompting if they would like to switch
* If the backgrounds went away, clean up the old ones and restore the user's setting to the one before going to seasonal

Could also have seasonal as a separate option, I guess. It can't be as done currently as every switch to seasonal will then require users to revert back manually.

peppy:
All at once sounds like not a great idea. Stable currently downloads each time a new one is required asynchronously, and caches after retrieval.

bdach:
I guess it's a pretty huge footprint, true. The seasonal background API response should probably persist across the current session, though, to avoid re-querying to check if there is a need to fetch a new one (`SessionStatics`?)

hbnrmx:
For reference, the current Halloween event has 15 images, so ~6Mb.
`SessionStatics` only stores boolean configuration values currently, so it may not be built with textures in mind.
That said, I don't know half a thing about caching...

bdach:
I was suggesting storing the API response there, not the textures themselves. For textures, the framework `LargeTextureStore` already does some caching, but I believe it won't have much of an effect here as the textures are refcounted - so if all references to a background are lost the texture will get unloaded, which I think will happen on screen transitions rendering the caching pretty pointless in this particular case.

PercyDan54:
A suggestion: We might want to split the settings into two
Use seasonal background: Sometimes, always, never
Background Source: Beatmap, Skin

hbnrmx:
@bdach 
@PercyDan54 

Thanks for your feedback, and sorry I didn't get to this earlier.
In this iteration I
* split out Seasonal Backgrounds (Always/Sometimes/Never) from Background Source (Beatmap/Skin)
(I also feel this makes more sense in the case where no seasonal event is happening.)

* changed background fallback such that Beatmap/Skin source choice
 is respected when no Seasonal event is happening.

* made `SeasonalBackgroundLoader` read from `SessionStatics` 
(this should make sure that the API is queried only once for the index)

This still leaves:

* repeated texture fetching issue
* ugly `LoadComponentAsync` chain as described in the OP.

peppy:
"Sometimes" is based off whether we are currently "in season". This is provided by the api response [already](https://osu.ppy.sh/docs/index.html#apiv2seasonal-backgrounds) in the form of a timestamp (`ends_at`) so you should be able to consume this.

osu.Game/Configuration/SessionStatics.cs
```diff
@@ -12,12 +16,16 @@ protected override void InitialiseDefaults()
         {
             Set(Static.LoginOverlayDisplayed, false);
             Set(Static.MutedAudioNotificationShownOnce, false);
+            Set(Static.SeasonEndDate, DateTimeOffset.MinValue);
+            Set(Static.SeasonalBackgrounds, new List<APISeasonalBackground>());
```
peppy, line 20: Can probably just store the the whole response to one static?

osu.Game/Configuration/SeasonalBackgrounds.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Configuration
+{
+    public enum SeasonalBackgrounds
```
peppy, line 6: `SeasonalBackgroundMode` will probably feel better

osu.Game/Screens/Backgrounds/BackgroundScreenDefault.cs
```diff
@@ -102,6 +106,18 @@ private Background createBackground()
...
+            {
+                case SeasonalBackgrounds.Sometimes:
+                    if (seasonalBackgroundLoader.IsInSeason())
+                        goto case SeasonalBackgrounds.Always;
+                    break;
+
+                case SeasonalBackgrounds.Always:
+                    newBackground = seasonalBackgroundLoader.LoadBackground() ?? newBackground;
+                    break;
+            }
```
peppy, line 119: this should probably be above everything else in this method. as it is, a different background is constructed above, then potentially thrown away due to the current sequence.

osu.Game/Screens/Backgrounds/BackgroundScreenDefault.cs
```diff
@@ -102,6 +106,18 @@ private Background createBackground()
             else
                 newBackground = new Background(backgroundName);
 
+            switch (showSeasonalBackgrounds.Value)
+            {
+                case SeasonalBackgrounds.Sometimes:
+                    if (seasonalBackgroundLoader.IsInSeason())
+                        goto case SeasonalBackgrounds.Always;
```
peppy, line 113: rather than using `goto`,  just write this as

`if (showSeasonalBackgrounds.Value == Always || (showSeasonalBackgrounds.Value == Sometimes))`

osu.Game/Graphics/Backgrounds/SeasonalBackgroundLoader.cs
```diff
@@ -0,0 +1,76 @@
...
+        {
+            if (!backgrounds.Value.Any()) return null;
+
+            current = (current + 1) % backgrounds.Value.Count;
+            string url = backgrounds.Value[current].Url;
+
+            return new SeasonalBackground(url);
+        }
+
+        public bool IsInSeason() => DateTimeOffset.Now < endDate.Value;
```
peppy, line 56:
```suggestion
        public bool IsInSeason => DateTimeOffset.Now < endDate.Value;
```

peppy (changes requested):

bdach:
For the record I've fixed the test failures locally and am currently in the process of writing a test scene for this.

bdach:
I've fixed the test failures, made a few stylistic changes, added test coverage, and also added support for delayed fetching of the backgrounds if API is not immediately available on launch, as the game would actually hard-crash if network was not available. The side effect of that change is that even though the API endpoint seems to be guest-accessible, the backgrounds won't be loaded until login, as we don't have an online-but-not-logged-in API state.

There's one transition at load that looks a little off that I want to look at but other than that I guess this is ok if we don't want to deal with caching the downloaded assets yet.

bdach:
It should be *sliiightly* better now. The fade in `SeasonalBackground.load` is a bit dodge but unsure if there's a better way to make sure it just doesn't come in without any transition at all and look absolutely janky.

peppy (approved):

PR #10591, opened 2020-10-24
Fix epilepsy warning not restoring track volume

### LINKED ISSUE #10567: Beatmap volume is decreased after epilepsy warning
**Describe the bug:** The volume on epilepsy warning screen gets lower, but doesn't return to normal when beatmap starts. The difference is easily noticeable after you retry the map, when the warning doesn't pop up.

**Screenshots or videos showing encountered issue:** https://streamable.com/aym55g

**osu!lazer version:** 2020.1020.0

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5409536/runtime.log)

bdach:
Resolves #10567.

Volume adjustment moved out of the warning itself into the loader, [as suggested](https://github.com/ppy/osu/issues/10567#issuecomment-713282338). After some deliberation I decided against having the loader DI-able as it would probably be way too convoluted and make the transition logic more opaque.

The fade back to 1 is immediate if transitioning to `Player` and not if going back to `SongSelect`. Had to delay the fade in first case as sometimes it would fire just barely too early and there would be a sudden volume jolt right before the push.

Tests included, but can also be tested on https://osu.ppy.sh/beatmapsets/11788#osu/45308, for one example. Additionally some code simplifications (transform sequence chaining to avoid triple nested schedules) and a slight visual change (the fadeout of the warning felt too long).


peppy:
I actually looked at doing this on `MusicController` itself as you have, but decided against it because it's almost guaranteed to become an issue at a later stage (there's no way to isolate fades if two components are both acting on the track). I think it's fine existing in `PlayerLoader`, but should still use the `AddAdjustment` flow..

peppy (changes requested):
as proposed

bdach:
Sure. I'm not going to be able to salvage the transition away (because the loader screen is going away, so can't put transforms on it on exit), but maybe best to just fix the actual issue instead of worrying about transitions for now.

peppy:
Yeah. Taking this to the next level, we'd likely want to add a short fade out while entering player anyway. If that is actually *easier* to implement (it might be?) then feel free to go with that instrea.

bdach:
I've changed to use manual adjustment as proposed, but abstained from adding a fade out on exit to `Player` as the logic is finicky enough as is... Sorta difficult to work with audio on screen transitions like these.

osu.Game/Screens/Play/EpilepsyWarning.cs
```diff
@@ -18,11 +16,7 @@ namespace osu.Game.Screens.Play
 {
     public class EpilepsyWarning : VisibilityContainer
     {
-        public const double FADE_DURATION = 500;
-
-        private readonly BindableDouble trackVolumeOnEpilepsyWarning = new BindableDouble(1f);
-
-        private Track track;
+        public const double FADE_DURATION = 250;
```
peppy, line 19: was this change for stylistic reasons?

bdach: Correct, 500ms fadeout on transition to player looked weird to me.

peppy: Guess it's fine. I set it higher for the inwards transition specifically, but not a huge deal.

peppy (approved):

PR #10596, opened 2020-10-26
Add support for deleting objects when Shift is held down

### LINKED ISSUE #10485: Right click while holding SHIFT to delete circles instantly in editor
Can also right click while holding SHIFT to delete multiple selected circles

charlie-gray:
This fixes #10485

When a circle is right clicked in the editor while shift is pressed, the circle is deleted. I added a function to SelectionHandler.cs to solve this.

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -225,9 +225,7 @@ namespace osu.Game.Screens.Edit.Compose.Components
         /// <param name="state">The input state at the point of selection.</param>
         internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState state)
         {
-            if (state.Keyboard.ShiftPressed && state.Mouse.IsPressed(MouseButton.Right))
-                EditorBeatmap.Remove(blueprint.HitObject);
-            else if (state.Keyboard.ControlPressed)
+            if (state.Keyboard.ControlPressed)
                 blueprint.ToggleSelection();
             else
                 ensureSelected(blueprint);
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -24,6 +24,7 @@ using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Objects.Types;
 using osuTK;
+using osuTK.Input;
 
 namespace osu.Game.Screens.Edit.Compose.Components
 {
@@ -224,7 +225,9 @@ namespace osu.Game.Screens.Edit.Compose.Components
         /// <param name="state">The input state at the point of selection.</param>
         internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState state)
         {
-            if (state.Keyboard.ControlPressed)
+            if (state.Keyboard.ShiftPressed && state.Mouse.IsPressed(MouseButton.Right))
+                EditorBeatmap.Remove(blueprint.HitObject);
+            else if (state.Keyboard.ControlPressed && state.Mouse.IsPressed(MouseButton.Left))
                 blueprint.ToggleSelection();
             else
                 ensureSelected(blueprint);
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -24,7 +24,6 @@ using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Objects.Types;
 using osuTK;
-using osuTK.Input;
 
 namespace osu.Game.Screens.Edit.Compose.Components
 {
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -226,11 +226,11 @@ namespace osu.Game.Screens.Edit.Compose.Components
         /// <param name="state">The input state at the point of selection.</param>
         internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState state)
         {
-            shiftClickDeleteCheck(blueprint, state);
-            multiSelectionHandler(blueprint, state);
+            if (!shiftClickDeleteCheck(blueprint, state))
+                handleMultiSelection(blueprint, state);
         }
 
-        private void multiSelectionHandler(SelectionBlueprint blueprint, InputState state)
+        private void handleMultiSelection(SelectionBlueprint blueprint, InputState state)
         {
             if (state.Keyboard.ControlPressed)
             {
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -32,6 +33,8 @@ namespace osu.Game.Screens.Edit.Compose.Components
     /// </summary>
     public class SelectionHandler : CompositeDrawable, IKeyBindingHandler<PlatformAction>, IHasContextMenu
     {
+        private bool shiftPressed;
```
bdach, line 36: This field should not exist.

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -455,6 +469,12 @@ public MenuItem[] ContextMenuItems
         {
             get
             {
+                if (shiftPressed)
+                {
+                    deleteSelected();
+                    return null;
+                }
```
bdach, line 476: Hijacking the context menu to delete the selected object like this instead is an incredibly workaround-y way of doing this. You'd probably be looking to do this in `HandleSelectionRequested` instead:

```csharp
if (state.Keyboard.ShiftPressed && state.Mouse.IsPressed(MouseButton.Right))
{
    EditorBeatmap.Remove(blueprint.HitObject);
    return;
}
```

although that method is getting a little bit complex already with that addition.

bdach (changes requested):

charlie-gray:
Got it. Does this work? I tried reducing the size of the function too. Let me know what you think.

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -237,6 +226,13 @@ internal void HandleDeselected(SelectionBlueprint blueprint)
         /// <param name="blueprint">The blueprint.</param>
         /// <param name="state">The input state at the point of selection.</param>
         internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState state)
+        {
+            shiftClickDeleteCheck(blueprint, state);
+            multiSelectionHandler(blueprint, state);
```
bdach, line 230: This is not the same as what my suggestion had - that's not how `return` works. This way both methods will be called in their entirety and `multiSelectionHandler` will presumably crash due to having the hitobject/blueprint deleted from under it.

bdach (changes requested):

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
charlie-gray: I think I understand. Let me just make sure.

If the hitobject/blueprint is deleted in `shiftClickDeleteCheck`, we want the function to return so that `multiSelectionHandler` does not run and crash. Is that correct?

bdach: That's correct. If the methods are to be split as you have them, the way this is sometimes done is by having the first function return a `bool` that returns if anything was indeed deleted guarding the second call, as so:

```csharp
if (!blueprintDeletionRequested(blueprint, state))
    handleMultiSelection(blueprint, state);
```

charlie-gray: Okay. I think I understand. Let me just double check.

We want the function `HandleSelectionRequested` to return after running `shiftClickDeleteCheck` if the hitobject/blueprint is deleted. This is so `multiSelectionHandler` doesn't run and isn't confused as to why there is not a hitobject/blueprint.

I think the easiest way to go about this would be to put the code you suggested earlier directly into `HandleSelectionRequested` above where `multiSelectionHandler` is called. Is that okay?

charlie-gray: Sorry, deleted the old comment by accident. I like your solution. Will implement it now. Thank you so much for your help!

charlie-gray:
Sorry! Ignore that one! I messed something up.

charlie-gray:
Does this work?

bdach:
You really ought to be running code style analysis locally, as per [the contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md).

bdach:
I've pushed a commit that resolves the code style issues and restructures the code somewhat differently. The main goal was to be able to tell from a glance at `HandleSelectionRequested` which actions could be performed on the blueprint on which the mouse event happened, while simultaneously decreasing the length of the body of that method by a few lines by splitting other methods.

That said I'm not convinced this is 100% the right direction. Maybe the proper way forward is to start defining `EditorAction`s and add a `KeyBindingContainer` in the editor hierarchy somewhere, as I can see `HandleSelectionRequested` becoming very clunky if more things are tacked onto it.

peppy:
I'm not that offended by the code at this point. We can reconsider if that handling function grows further, as you say.

peppy (dismissed):

peppy:
The one time this doesn't work is if the blueprint is already in a selected state. Not sure if we need to cover that case though...

peppy:
I've moved this logic to the `SelectionBlueprint` to better handle the case where the blueprint is already selected. @bdach request a re-review to make sure you're okay with this.

As an aside, people are next going to ask for this to work in compose mode too. That's a bit harder to implement and will probably need to lean on the snapping algorithm somehow.

peppy (dismissed):

bdach:
I'm not 100% sure I agree with that direction... Right click on selected blueprints didn't work because `beginClickSelection` wasn't propagating higher up:

https://github.com/ppy/osu/blob/37e9e39ee85f73a466581ee99a8090f73ba6f5c6/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs#L301-L306

But my version of `HandleSelectionRequested` was already doing the deselection check (or, half of it, anyway), so it seems to me that it could have worked by doing

<details>
<summary>patch</summary>

```diff
diff --git a/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs b/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
index 7751df29c..5ac360d02 100644
--- a/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/BlueprintContainer.cs
@@ -298,13 +298,6 @@ private void beginClickSelection(MouseButtonEvent e)
         {
             Debug.Assert(!clickSelectionBegan);
 
-            // Deselections are only allowed for control + left clicks
-            bool allowDeselection = e.ControlPressed && e.Button == MouseButton.Left;
-
-            // Todo: This is probably incorrectly disallowing multiple selections on stacked objects
-            if (!allowDeselection && SelectionHandler.SelectedBlueprints.Any(s => s.IsHovered))
-                return;
-
             foreach (SelectionBlueprint blueprint in SelectionBlueprints.AliveChildren)
             {
                 if (blueprint.IsHovered)
diff --git a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
index 9cddb69d0..01e23bafc 100644
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
@@ -227,7 +227,7 @@ internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState
         {
             if (state.Keyboard.ShiftPressed && state.Mouse.IsPressed(MouseButton.Right))
                 EditorBeatmap.Remove(blueprint.HitObject);
-            else if (state.Keyboard.ControlPressed)
+            else if (state.Keyboard.ControlPressed && state.Mouse.IsPressed(MouseButton.Left))
                 blueprint.ToggleSelection();
             else
                 ensureSelected(blueprint);
```

</details>

Is there a particular reason I'm missing as to why not do it that way? The deletion logic feels out of place in `SelectionBlueprint` (and leads to the test failures, due to a new dependency on `EditorBeatmap`).

peppy:
Oh definitely, if that works I have no issue with it. I went down the path of thinking it was the context menu handling that was interrupting, but by the end of changing the implementation I realised that wasn't the case.

Feel free to revert my commit and apply your change.

bdach (approved):

smoogipoo (approved):
I think this is fine

PR #10597, opened 2020-10-27
Fix mania crashing due to spectator client handling frames with unconverted beatmap

bdach:
Not confident about this at all, but I wanted to get this out for discussion at least.

# Summary

On current `master`, trying to start a mania map crashes with a

```
[runtime] 2020-10-26 23:41:00 [error]: System.InvalidCastException: Unable to cast object of type 'osu.Game.Beatmaps.Beatmap' to type 'osu.Game.Rulesets.Mania.Beatmaps.ManiaBeatmap'.
[runtime] 2020-10-26 23:41:00 [error]: at osu.Game.Rulesets.Mania.Replays.ManiaReplayFrame.ToLegacy(IBeatmap beatmap) in /home/dachb/Dokumenty/opensource/osu/osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs:line 57
[runtime] 2020-10-26 23:41:00 [error]: at osu.Game.Online.Spectator.SpectatorStreamingClient.HandleFrame(ReplayFrame frame) in /home/dachb/Dokumenty/opensource/osu/osu.Game/Online/Spectator/SpectatorStreamingClient.cs:line 250
[runtime] 2020-10-26 23:41:00 [error]: at osu.Game.Rulesets.UI.ReplayRecorder`1.recordFrame(Boolean important) in /home/dachb/Dokumenty/opensource/osu/osu.Game/Rulesets/UI/ReplayRecorder.cs:line 96
```

This is due to `ManiaReplayFrame` being hard-coupled to the converted beatmap:

https://github.com/ppy/osu/blob/19e58dc4fc244bbadb992942f07a35dbed00142d/osu.Game.Rulesets.Mania/Replays/ManiaReplayFrame.cs#L28-L30

while the spectator client was using a plain one taken from `WorkingBeatmap`:

https://github.com/ppy/osu/blob/19e58dc4fc244bbadb992942f07a35dbed00142d/osu.Game/Online/Spectator/SpectatorStreamingClient.cs#L47-L48

Resolve by shoving the converted beatmap into the spectator client through `ReplayRecorder`, to avoid converting/applying mods twice. Unfortunately it's infeasible to use DI given the current structure without at least adding a game-level bindable - which is mostly why I'm bothering to PR this as I'm not sure I see any alternative.

# Remarks

A few test scenes need a cached `GameplayBeatmap` after this change, as evidenced by the diff.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneReplayRecorder.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneReplayRecorder.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using NUnit.Framework;
+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
@@ -12,11 +13,13 @@ using osu.Framework.Input.Events;
 using osu.Framework.Input.StateChanges;
 using osu.Framework.Testing;
 using osu.Framework.Threading;
+using osu.Game.Beatmaps;
 using osu.Game.Graphics.Sprites;
 using osu.Game.Replays;
 using osu.Game.Rulesets;
 using osu.Game.Rulesets.Replays;
 using osu.Game.Rulesets.UI;
+using osu.Game.Screens.Play;
 using osu.Game.Tests.Visual.UserInterface;
 using osuTK;
 using osuTK.Graphics;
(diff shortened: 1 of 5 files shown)


peppy:
Quite amazed that this didn't manifest in any test scene we already had..

peppy (approved):

bdach:
> Quite amazed that this didn't manifest in any test scene we already had..

Most likely didn't due to the autoplay exclusion.

PR #10601, opened 2020-10-27
Remove remaining sample disable logic from FrameStabilityContainer

### LINKED ISSUE #10554: slider slide sound keeps playing when changing to timing tab in the editor
**Describe the bug:** if you are in the editor in compose tab and change to timing tab while the slider is playing, the slider slide sound will keep playing until you change back to the compose tab. same with the spinner sound.

https://cdn.discordapp.com/attachments/748809435668938805/767508271224586250/2020-10-19_02-01-58.mp4

**osu!lazer version:** 2020.1017.0

**Logs:**
no need

peppy:
This stuff is never going to go away is it?

This is part of the changes required to fix #10554. Basically, the editor has a `FrameStabilityContainer` present (even though it doesn't really make use of it) and this causes sample disabled triggered from editor pause to not propagate to a non-alive screen. Moving this logic out feels better because it means FrameStability is no longer responsible for pausing logic at all.


smoogipoo (changes requested):
This looks to have broken samples - most gameplay samples don't work anymore and the slide sample buzzes.

peppy:
I guess this is the case where your PC is running slow enough to be considered behind realtime? Does it happen on all beatmaps or are you testing with something out-there?

peppy:
I've restored the previous 200ms lenience, can you try again now?

smoogipoo:
Works fine now

smoogipoo (approved):

PR #10605, opened 2020-10-27
Add screen hierarchy for spectating another player

peppy:
Just going to PR this to get something out there (and so I can monitor the size of the diff)

Note that there's no way to reach this from the client itself, and I've been developing it completely via visual tests. Individual commits can be split out if deemed required. A few underlying changes were required to make this work:

- `FrameStabilityContainer` needed logic changes to handle cases where replay data was not completely received. Fixing handling where there's zero frames, or where we've reached the end of streaming data.
- Player needs to act on `FrameStabilityContainer` running out of frames, and pause gameplay. Right now this is done via a simple method, which will likely need to be adjusted once there's the ability for a user to pause replay playback.

Note that some of the tests are incomplete (pending further implementation in the `Spectator` screen component). All the basic flows and behaviours seem to be correct though.

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
@@ -59,21 +59,20 @@ namespace osu.Game.Tests.Visual.Gameplay
 
             AddAssert("screen hasn't changed", () => Stack.CurrentScreen is Spectator);
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
-
+            start();
             sendFrames();
 
-            AddUntilStep("wait for player", () => Stack.CurrentScreen is Player);
-
+            waitForPlayer();
             AddAssert("ensure frames arrived", () => replayHandler.HasFrames);
 
             AddUntilStep("wait for frame starvation", () => replayHandler.NextFrame == null);
-            AddAssert("game is paused", () => player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value);
-            sendFrames();
-            AddUntilStep("game resumed", () => !player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value);
+            checkPaused(true);
 
+            sendFrames();
+
+            checkPaused(false);
             AddUntilStep("wait for frame starvation", () => replayHandler.NextFrame == null);
-            AddAssert("game is paused", () => player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value);
+            checkPaused(true);
         }
 
         [Test]
@@ -81,17 +80,19 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
-
-            AddUntilStep("wait for player", () => Stack.CurrentScreen is Player);
-
+            start();
+            waitForPlayer();
             AddUntilStep("game is paused", () => player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value);
+
+            sendFrames();
+
+            checkPaused(false);
         }
 
         [Test]
         public void TestSpectatingDuringGameplay()
         {
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
             // should seek immediately to available frames
             loadSpectatingScreen();
@@ -102,9 +103,9 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
         }
 
@@ -113,10 +114,10 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
 
-            // should replay until running out of frames then fail
+            // TODO: should replay until running out of frames then fail
         }
 
         [Test]
@@ -124,10 +125,9 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
-            AddUntilStep("wait for player", () => Stack.CurrentScreen is Player);
-
+            waitForPlayer();
             // should immediately exit and unbind from streaming client
             AddStep("stop spectating", () => (Stack.CurrentScreen as Player)?.Exit());
 
@@ -139,10 +139,17 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+            start();
             sendFrames();
             // player should never arrive.
         }
+
+        private void waitForPlayer() => AddUntilStep("wait for player", () => Stack.CurrentScreen is Player);
+
+        private void start() => AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+
+        private void checkPaused(bool state) =>
+            AddAssert($"game is {(state ? "paused" : "playing")}", () => player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value == state);
 
         private void sendFrames(int count = 10)
         {
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
@@ -127,7 +127,7 @@ namespace osu.Game.Tests.Visual.Gameplay
         }
 
         [Test]
-        public void TestHostStartsPlayingWhileAlreadyWatching()
+        public void TestHostRetriesWhileWatching()
         {
             loadSpectatingScreen();
 
@@ -171,9 +171,8 @@ namespace osu.Game.Tests.Visual.Gameplay
             sendFrames();
             waitForPlayer();
 
-            // should immediately exit and unbind from streaming client
             AddStep("stop spectating", () => (Stack.CurrentScreen as Player)?.Exit());
-            AddUntilStep("spectating stopped", () => spectatorScreen.GetParentScreen() == null);
+            AddUntilStep("spectating stopped", () => spectatorScreen.GetChildScreen() == null);
         }
 
         [Test]
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
@@ -25,6 +25,9 @@ namespace osu.Game.Tests.Visual.Gameplay
         [Cached(typeof(SpectatorStreamingClient))]
         private TestSpectatorStreamingClient testSpectatorStreamingClient = new TestSpectatorStreamingClient();
 
+        // used just to show beatmap card for the time being.
+        protected override bool UseOnlineAPI => true;
+
         private Spectator spectatorScreen;
 
         [Resolved]
@@ -134,9 +137,9 @@ namespace osu.Game.Tests.Visual.Gameplay
             start();
             sendFrames();
             waitForPlayer();
+
             // should immediately exit and unbind from streaming client
             AddStep("stop spectating", () => (Stack.CurrentScreen as Player)?.Exit());
-
             AddUntilStep("spectating stopped", () => spectatorScreen.GetParentScreen() == null);
         }
 
@@ -145,14 +148,15 @@ namespace osu.Game.Tests.Visual.Gameplay
         {
             loadSpectatingScreen();
 
-            start();
+            start(88);
             sendFrames();
-            // player should never arrive.
+
+            AddAssert("screen didn't change", () => Stack.CurrentScreen is Spectator);
         }
 
         private void waitForPlayer() => AddUntilStep("wait for player", () => Stack.CurrentScreen is Player);
 
-        private void start() => AddStep("start play", () => testSpectatorStreamingClient.StartPlay());
+        private void start(int? beatmapId = null) => AddStep("start play", () => testSpectatorStreamingClient.StartPlay(beatmapId));
 
         private void checkPaused(bool state) =>
             AddAssert($"game is {(state ? "paused" : "playing")}", () => player.ChildrenOfType<DrawableRuleset>().First().IsPaused.Value == state);
@@ -176,7 +180,7 @@ namespace osu.Game.Tests.Visual.Gameplay
 
             public readonly User StreamingUser = new User { Id = 1234, Username = "Test user" };
 
-            public void StartPlay() => sendState();
+            public void StartPlay(int? beatmapId = null) => sendState(beatmapId);
 
             public void EndPlay()
             {
@@ -218,12 +222,12 @@ namespace osu.Game.Tests.Visual.Gameplay
                 base.WatchUser(userId);
             }
 
-            private void sendState()
+            private void sendState(int? beatmapId = null)
             {
                 sentState = true;
                 ((ISpectatorClient)this).UserBeganPlaying((int)StreamingUser.Id, new SpectatorState
                 {
-                    BeatmapID = beatmaps.GetAllUsableBeatmapSets().First().Beatmaps.First(b => b.RulesetID == 0).OnlineBeatmapID,
+                    BeatmapID = beatmapId ?? beatmaps.GetAllUsableBeatmapSets().First().Beatmaps.First(b => b.RulesetID == 0).OnlineBeatmapID,
                     RulesetID = 0,
                 });
             }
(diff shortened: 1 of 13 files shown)


bdach (dismissed):
Here's a rudimentary first pass. Haven't looked at this too hard, especially given that tests are yet to pass.

osu.Game.Tests/Visual/Gameplay/TestSceneSpectator.cs
```diff
@@ -0,0 +1,236 @@
...
+
+        private int nextFrame;
+
+        public override void SetUpSteps()
+        {
+            base.SetUpSteps();
+
+            AddStep("reset sent frames", () => nextFrame = 0);
+
+            AddStep("import beatmap", () => ImportBeatmapTest.LoadOszIntoOsu(game, virtualTrack: true).Wait());
```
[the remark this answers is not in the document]
peppy: yeah, was planning on replacing this with a randomised beatmap to avoid weird edge cases. the import is very temporary.

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
```diff
@@ -124,22 +124,50 @@ private bool inImportantSection
...
+
+                        // else allow play to end.
+                    }
+                    else if (next.Time.CompareTo(time) != currentDirection)
+                    {
+                        // in the case we have more frames, block if the next frame's time is less than the current time.
+                        return null;
+                    }
+
+                    // if we didn't change frames, we need to ensure we are allowed to run frames in between, else return null.
```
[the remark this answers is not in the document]
peppy: nah, that comment is talking about the fall-through case.

and i completely agree that this method is shocking, but i spent a good hour looking at it and didn't come up with anything better. will take another look at it today, but open to any improvements you can see, even if they are minor steps in the right direction.

curious if you can repro that behaviour on master?

peppy:
Funny part is all tests are/were passing 100% locally 😅. Always the way.

bdach:
Yeah, I'm not even 100% sure what that `Push` failure is all about. Looks like perhaps the spectator client might be calling `Push` too eagerly? Cannot repro locally, so mostly guessing...

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
```diff
@@ -124,22 +124,50 @@ private bool inImportantSection
...
+                        // in the case we have no more frames and haven't received the full replay, block.
+                        if (!replay.HasReceivedAllFrames)
+                            return null;
+
+                        // else allow play to end.
+                    }
+                    else if (next.Time.CompareTo(time) != currentDirection)
+                    {
+                        // in the case we have more frames, block if the next frame's time is less than the current time.
+                        return null;
```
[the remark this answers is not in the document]
peppy: I've added this outside the if-else clause as it's `!= 0` in all cases.

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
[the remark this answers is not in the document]
peppy: Rewrote this logic, hopefully the issue is just-resolved now (I still can't see this happening locally).

bdach:
I think I've understood why I'm getting this weird oscillation issue. [Here's a demonstration video of how it looks to me.](https://drive.google.com/file/d/1EMPuogknGkJu0tBUjM80_UaM4xpVd_x7/view)

In short, there are at least two feedback mechanisms at play here.

1. `FrameStabilityContainer` is at time ~1899, and proposes to advance to ~1901. The next available frame is at time 1900 - so FSC is allowed to advance to 1900.
2. FSC is at time 1900, but it proposed to advance to ~1901, so it operates as if it is actually falling behind on a rewind and starts going backwards for a few frames.
3. At some point FSC manages to succeed after proposing a time at some point before 1900 (say, ~1890). The next frame it tries to go forward again, and is blocked on pending frames, at which point it starts going backwards again to catch up.
4. Eventually FSC succeeds in proposing a time before ~1890, and is allowed to proceed to it. Rinse and repeat this back and forth (a few frames at a time, every time), until
5. FSC rewinds past 1700 at which point there are 2 pending frames and normal playback resumes until the threshold of 1900 is crossed again.

Gonna try to rack my brain for a possible fix but at this point it might honestly be easier to just disable rewind entirely when spectating? Not sure...

bdach:
Still got nothing after ~2 hours. I feel like the logic of checking for new frames should be separated out of the time-dependent logic, as the one extra update loop is triggering this. Commenting it out makes the gameplay stop properly (but not resume, understandably).

I've tried fiddling with multiple things but I got nothing. The parent gameplay clock moving on its own thwarts all my efforts - even if I can get FSC to stay in place time-wise, the parent clock will shift back after hitting the last frame boundary and trigger all of the rewind logic that causes this to happen in the first place

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -119,93 +115,109 @@ private void updateClock()
...
                 setClock(); // LoadComplete may not be run yet, but we still want the clock.
 
-            validState = true;
-            requireMoreUpdateLoops = false;
+            // each update start with considering things in valid state.
+            state = PlaybackState.Valid;
 
-            var newProposedTime = parentGameplayClock.CurrentTime;
+            // our goal is to catch up to the time provided by the parent clock.
+            var proposedTime = parentGameplayClock.CurrentTime;
```
bdach, line 122: Actually, now that I look at it again,

```suggestion
            var proposedTime = frameStableClock.IsPaused.Value ? manualClock.CurrentTime : parentGameplayClock.CurrentTime;
```

*sorta kinda* fixes it, but I don't even know where to *begin* justifying that change...

peppy: As discussed on stream, fixed this in a very similar way. Need confirmation it works for you, as I can't reproduce locally.

peppy:
I'm very unsure how the failing test is failing. Been trying to repro it locally but can't over countless headless runs.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
bdach: I can confirm that the latest changes fix the oscillation issue.

bdach:
I think I am able to reproduce the test failure, or at least *a* failure that is similar to it. To reproduce, comment/ignore all of the tests except for the failing one (`TestPlayStartsWithNoFrames`) and run it in the test browser. Headless seems to pass. ([Don't ask how I got this "bright" idea.](https://twitter.com/Spaceman_Atlas/status/1321863418563956736))

I will attempt to describe the failure case, with log excerpts this time, to hopefully describe better what's happening.

<details>
<summary>Failure case description (or at least an attempt of one)</summary>

FSC = `FrameStabilityContainer`
FRIH = `FramedReplayInputHandler`

```
12411: step 13 send frames
```

The "send frames" step executes. FSC's `UpdateSubTree` runs:

```
[runtime] 2020-10-29 19:54:15 [verbose]: == UPDATESUBTREE START ==
[runtime] 2020-10-29 19:54:15 [verbose]: FSC waiting on frames = True, paused = True, manual @ -1043.595439, gameplay @ -1084.595439, framed @ -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = 1, next frame @ 0, current time = , proposed time = -1043.595439, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -1043.595439
```

FSC is paused, so it sets frame from the manual clock's time. FRIH wants to advance forward correctly to the frame at time 0, but the time given in FSC is actually in the past relative to 0. Since the directions don't match, and the conditions to block are not met, it suggests to stay at the same time instant instead of advancing to the newly-received first frame.

```
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 0, manual clock @ -1043.595439 (rate 0), framed clock @ -1043.595439 (rate 0)
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = 1, next frame @ 0, current time = -1043.595439, proposed time = -1043.595439, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 0, manual clock @ -1043.595439 (rate 0), framed clock @ -1043.595439 (rate 0)
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = 1, next frame @ 0, current time = -1043.595439, proposed time = -1043.595439, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 0, manual clock @ -1043.595439 (rate 0), framed clock @ -1043.595439 (rate 0)
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = 1, next frame @ 0, current time = -1043.595439, proposed time = -1043.595439, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 0, manual clock @ -1043.595439 (rate 0), framed clock @ -1043.595439 (rate 0)
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = 1, next frame @ 0, current time = -1043.595439, proposed time = -1043.595439, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 0, manual clock @ -1043.595439 (rate 0), framed clock @ -1043.595439 (rate 0)
```

This continues for the 5 times FSC is allowed to try and catch up to the gameplay clock in one frame.

On the next frame:

```
[runtime] 2020-10-29 19:54:15 [verbose]: == UPDATESUBTREE START ==
[runtime] 2020-10-29 19:54:15 [verbose]: FSC waiting on frames = False, paused = False, manual @ -1043.595439, gameplay @ -1124.6999797887659, framed @ -1043.595439
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = -1, next frame @ 0, current time = -1043.595439, proposed time = -1060.2621056666667, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed 0
```

Since `WaitingOnFrames` has become false, the gameplay clock is still going backwards, but is no longer paused. Stability is applied, clamping -1124 to -1060 (~16ms before manual clock's time). FRIH now supposes it is going backwards (wrongly, since it has actually never even played a frame), and the directions match (also wrongly), so it proposes advancing to time 0 despite technically doing playback backwards. The playback state is valid, so the catch-up logic is allowed to run.

```
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = 1, manual clock @ 0 (rate 0.021840392924351245), framed clock @ 0 (rate 0.021840392924351245)
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH direction = -1, next frame @ , current time = 0, proposed time = -16.666666666666668, important section = False, replay done = False
[runtime] 2020-10-29 19:54:15 [verbose]: FRIH proposed -16.666666666666668
[runtime] 2020-10-29 19:54:15 [verbose]: FSC direction = -1, manual clock @ -16.666666666666668 (rate -0.021840392924351245), framed clock @ -16.666666666666668 (rate -0.021840392924351245)
```

While both clocks have now moved to the first frame correctly, both FRIH and FSC want to go backwards still, therefore FRIH blocks playback at time 0 due to apparent frame starvation.

</details>

No bright ideas on how to fix right now, sorry. Even grasping this scenario mentally in its entirety is at the limit of my cognition right now.

peppy:
Does that repro occur 100%? I can't seem to make it happen. The details you provided should be enough to figure out a solution, but still weird this is the second time I haven't been able to repro an issue that you can..

peppy:
I've pushed a fix which I *think* should work to cover the case you describe (or, basically, if it used to work then it should still work after this PR). No longer running the updateClock loop in the case of waiting for frames.

I'm not sure if there's still issues with the updateClock loop in the scenario you have mentioned so be on the lookout for anything else weird.

bdach:
It wasn't entirely 100% (seemed to be timing dependent as breakpointing in places sometimes allowed the test to pass) but it was consistent if only logging output was tacked on, yeah.

Zero clue why you can't seem to reproduce any of these, and the test is still failing :( I guess I'll have to continue my insane exercise of trying to debug on appveyor if the fix does work for me locally...

peppy:
For what it's worth I can repro on windows.

bdach:
I just noticed the discord message where you mentioned offsets - I in fact do have a custom offset in my lazer config, and windows also has a hardcoded offset of 15ms I wanna say? That'll probably be the key piece as to why our execution states seem to be different.

peppy:
Oh damn. I think you may be on to something there.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -85,35 +85,52 @@ protected override void LoadComplete()
 
         public override bool UpdateSubTree()
         {
-            state = frameStableClock.IsPaused.Value ? PlaybackState.NotValid : PlaybackState.Valid;
+            if (parentGameplayClock == null)
+                setClock(); // LoadComplete may not be run yet, but we still want the clock.
+
+            double aimTime = parentGameplayClock.CurrentTime;
+
+            if (frameStableClock.WaitingOnFrames.Value)
```
smoogipoo, line 93: Am I the only one that can't follow this logic? It's split between 3 methods - updateClock(), updateReplay() and now UpdateSubTree(), all of which directly contribute to the state.

I really don't like this. Why can't this be inside updateClock() at the very least, with a further improvement being to separate it out into its own method that isolates its purpose?

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -85,35 +85,52 @@ protected override void LoadComplete()
...
+            {
+                // waiting on frames is a special case where we want to avoid doing any update propagation, unless new frame data has arrived.
```
smoogipoo, line 95: What's the issue with the "update propagation" happening? All we're doing is updating the clocks right? Over which both updateReplay() and updateClock() have full control over the proposed time and can choose to set it arbitrarily...

Regardless of how that's done the final state should be NotValid and the update "propagation" won't happen anyway?

smoogipoo (changes requested):
Can this replay stuff be split out into a separate PR?

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
peppy: can you explain which logic you mean? the fact that this variable is set inside `updateClock`?

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
peppy: I moved this out to avoid running the updateClock method while waiting on frames. By running it, it updates the manualClock's state which I wanted to avoid at all costs.

But I think you may be right in that we don't want to call SetFrameFromTime twice in a row, as it will advance two frames, potentially.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
smoogipoo: All of the new logic added in UpdateSubTree is very hard to follow. Including the parent gameplay clock now being set here instead of updateClock()

peppy: I can attempt to move it back. I initially moved it over to attempt to fix the reported issue, but it turns out that was not required.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
smoogipoo: > we don't want to call SetFrameFromTime twice in a row

That's the sort of stuff I'm concerned about. As another example, this also has no check for `ReplayInputHandler` being null, which is already checked prior to `updateReplay()`. Which is why I mentioned that it's hard to follow this logic above.

> updates the manualClock's state which I wanted to avoid at all costs

But why? What part of it would break? Couldn't this very same logic, placed inside updateReplay(), set `proposedTime = manualClock.CurrentTime` and forego such updates?

bdach:
I've gone through this code again and pushed in a few changes. Most are xmldocs, but one is pretty important (7164583) - pretty sure the lack of unsubscriptions there was keeping instances of `SpectatorPlayer` alive.

The oscillation issue [I described here](https://github.com/ppy/osu/pull/10605#issuecomment-718119011) is back, but it does seem to be caused by incorrect offset application which is now tracked separately in #10646, and this diff is large enough as it is on its own, so I'd say let's ignore for now.

I don't see much reason to hold up this PR any more, although I'm still a little cagey on the clock logic. It reads correctly to me, as in I can't spot anything that would be wrong, but there are a lot of things that *aren't* there but should be that I could be missing.

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -85,34 +85,46 @@ protected override void LoadComplete()
...
+                // it should be provided as the original value each loop.
                 updateClock();
 
                 if (state == PlaybackState.NotValid)
                     break;
 
                 base.UpdateSubTree();
                 UpdateSubTreeMasking(this, ScreenSpaceDrawQuad.AABBFloat);
-            }
+            } while (state == PlaybackState.RequiresCatchUp && loops-- > 0);
```
smoogipoo, line 101: nit

```suggestion
            } while (state == PlaybackState.RequiresCatchUp && --loops > 0);
```

smoogipoo (approved):
Looks good to me

PR #10606, opened 2020-10-27
Fix timeline zoom during drag-select causing unexpected behaviour

### LINKED ISSUE #10594: Scrolling while selecting within the timeline causes unexpected behavior
Title

https://puu.sh/GGDv1/5de23802dd.mp4

**osu!lazer version:** 2020.1020.0

FamousPig:
Fixes #10594 

Drag-selecting while zooming in the timeline would cause unexpected results. The starting point of the selection would shift and the end point behave erratically. I believe this is caused by the scaling of the zoomed content scaling the local coordinates with it. 

```csharp
private void updateZoomedContentWidth() => zoomedContent.Width = DrawWidth * currentZoom;
```

Fixed by exposing ``currentZoom`` value of ``ZoomableScrollContainer`` and compensating for zoom changes in the ``handleDrag()`` method.

I attempted to use ``ToSpaceOfOtherDrawable`` to convert coordinates to and from various related components [as suggested](https://github.com/ppy/osu/issues/10594#issuecomment-716766355). However none of my attempts yielded the expected results. I can not say for sure that a better solution does not exist, but I was unable to find one.

<details>

Before:

before

After:

after

</details>

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBlueprintContainer.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/Timeline/TimelineBlueprintContainer.cs
@@ -9,7 +9,6 @@ using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input.Events;
-using osu.Framework.Utils;
 using osu.Game.Rulesets.Edit;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Screens.Edit.Components.Timelines.Summary.Parts;
@@ -139,6 +138,7 @@ namespace osu.Game.Screens.Edit.Compose.Components.Timeline
         private class TimelineDragBox : DragBox
         {
             private Vector2 lastMouseDown;
+
             private float? lastZoom;
             private float localMouseDown;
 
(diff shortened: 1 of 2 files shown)


bdach:
I don't like this resolution code-quality wise (both exposing `CurrentZoom` and storing `localZoom` should not be necessary), but I'm not seeing any better options. I could get rid of one of the usages of `lastZoom` but not the other (the one used for `selection2`). I don't understand why any way I try to transform the coordinates it seems like *something*'s interfering and I get a bouncy scroll at the end being dragged.

Ideally `ZoomableScrollContainer` should be able to spit out some kind of zoom-agnostic coordinate which `DragBox` would then store, and then `DragBox` would pass it back for the scroll container to rescale it back using current zoom internally. As I'm failing to find a way to do this right now, skipping further attempts/review of this for now as to not hold up the queue. Will revisit later, hopefully.

FamousPig:
The reason `selection2` is bouncy is probably because the `MouseButtonEvent` contains last frame's "scaled" coordinates. Would it be alright to move the "zoomcorrection"  stuff into `ZoomableScrollContainer` and provide a method like `UpdateStaleCoordinates` which applies the correction? It's still a workaround but I feel it would be better than what we have now.

bdach:
I've arrived at an alternate solution to this, which I believe is slightly nicer to the original proposed version. It's still way more arcane than it needs to be, but it seems to work.

It still requires `CurrentZoom` to be publicly exposed, which I guess is not too bad - may be however worth to rename `ZoomableScrollContainer.Zoom` to `TargetZoom` to better convey the distinction between the two alongside the xmldoc. Not sure.

smoogipoo (approved):
Seems to work fine. There's a 1-frame issue but I believe that's coming from elsewhere, and I think this is fine for now.

PR #10608, opened 2020-10-28
Fix editor crash on exit when forcing exit twice in a row

### LINKED ISSUE #10148: make some browser tab shortcuts work for osu chat tabs
**Describe the new feature:** 
make ctrl+w, ctrl+shift+t, ctrl+t do the same thing you would expect them to do in your browser
ctrl+w - close the active chat tab if chat is opened (already works in stable i believe)
ctrl+shift+t - reopen closed chat tabs in the order in which they were closed
ctrl+t - I imagine maybe prompting a user to enter a name while user search results are showing up so they can choose if they're unsure with the name

### LINKED ISSUE #10418: Bindable error on quit exit from editor
Seems like somehow `OnResuming` is being called more than once (or `OnSuspending` was never called?)

```
[runtime] 2020-10-08 08:51:58 [error]: System.InvalidOperationException: This bindable is already bound to the requested bindable (osu.Game.Rulesets.Mods.Mod[]).
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Bindables.Bindable`1.BindTo(Bindable`1 them)
[runtime] 2020-10-08 08:51:58 [error]: at osu.Game.Screens.Select.SongSelect.OnResuming(IScreen last) in /Users/dean/Projects/osu/osu.Game/Screens/Select/SongSelect.cs:line 567
[runtime] 2020-10-08 08:51:58 [error]: at osu.Game.Screens.Select.PlaySongSelect.OnResuming(IScreen last) in /Users/dean/Projects/osu/osu.Game/Screens/Select/PlaySongSelect.cs:line 51
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Screens.ScreenStack.resumeFrom(IScreen source)
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Screens.ScreenExtensions.MakeCurrent(IScreen screen)
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal()
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Threading.Scheduler.Update()
[runtime] 2020-10-08 08:51:58 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree()
```

peppy:
Basically the editor was calling `this.Exit()` from inside `OnExiting`, which manages to trigger a double `MakeCurrent` call on screen stack. Will be looking at a framework side reproduction / fix for this (to throw a better error message) but it does also need to be fixed locally.

Closes #10418.

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -444,12 +444,21 @@ namespace osu.Game.Screens.Edit
                 if (dialogOverlay == null || dialogOverlay.CurrentDialog is PromptForSaveDialog)
                 {
                     confirmExit();
-                    return true;
+                    return false;
                 }
 
                 if (isNewBeatmap || HasUnsavedChanges)
                 {
-                    dialogOverlay?.Push(new PromptForSaveDialog(confirmExit, confirmExitWithSave));
+                    dialogOverlay?.Push(new PromptForSaveDialog(() =>
+                    {
+                        confirmExit();
+                        this.Exit();
+                    }, () =>
+                    {
+                        confirmExitWithSave();
+                        this.Exit();
+                    }));
+
                     return true;
                 }
             }
@@ -464,7 +473,6 @@ namespace osu.Game.Screens.Edit
         {
             exitConfirmed = true;
             Save();
-            this.Exit();
         }
 
         private void confirmExit()
@@ -483,7 +491,6 @@ namespace osu.Game.Screens.Edit
             }
 
             exitConfirmed = true;
-            this.Exit();
         }
 
         private readonly Bindable<string> clipboard = new Bindable<string>();


peppy:
Blocking pending framework investigation. This may be a framework-only fix after all.

peppy:
This is no longer required due to https://github.com/ppy/osu-framework/pull/3956. I'm not sure if the previous or current code reads better, so up to reviewers to make that call. The old code was weird enough that I misunderstood it to be a bug, but it turns out it actually "works" correctly due to the exit call first setting `exitConfirmed`.

peppy:
Scratch that. This logic is required else the exit process ends at song select, rather than doing the required game-exit process.

bdach (approved):
Seems reasonable :+1:

smoogipoo (approved):

PR #10609, opened 2020-10-28
Clean up FrameStabilityContainer state handling logic

peppy:
Figured this could be improved while working through the spectator logic.

This should not change any logic, just restructure to be easier to follow.

### DIFF
--- a/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
+++ b/osu.Game/Rulesets/UI/FrameStabilityContainer.cs
@@ -236,14 +236,13 @@ namespace osu.Game.Rulesets.UI
             NotValid,
 
             /// <summary>
-            /// Whether we are running up-to-date with our parent clock.
-            /// If not, we will need to keep processing children until we catch up.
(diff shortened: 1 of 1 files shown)


bdach (dismissed):
Looks correct as far as I can tell transformation-wise. Definitely a readability improvement. :+1:

smoogipoo (dismissed):

osu.Game/Rulesets/UI/FrameStabilityContainer.cs
```diff
@@ -224,6 +222,26 @@ private void setClock()
...
         public ReplayInputHandler ReplayInputHandler { get; set; }
 
+        private enum PlaybackState
+        {
+            /// <summary>
+            /// Playback is not possible. Child hierarchy should not be processed.
+            /// </summary>
+            NotValid,
+
+            /// <summary>
```
peppy, line 238: just noticed these xmldoc were copy-pasted and now read incorrectly, lemme fix.

smoogipoo (approved):

PR #10611, opened 2020-10-28
Add "currently playing" player listing and spectator UI flow

peppy:
This is all non-final designs (aka pulled from my ass) but it should be functional enough.

preview: https://www.youtube.com/watch?v=UA681Iu6HBE

- [x] Depends on https://github.com/ppy/osu/pull/10605

Issues:
- Fail won't work as it is disabled (needs to be until we are transmitting full gameplay state in frame bundles to allow correctly resuming from point-in-time).

### DIFF
--- a/osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
+++ b/osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
 using osu.Framework.Allocation;
@@ -43,6 +44,9 @@ namespace osu.Game.Overlays.Dashboard
         [Resolved]
         private IAPIProvider api { get; set; }
 
+        // temporary, should be game-global but i don't want to add more manager classes for now.
+        private static readonly Dictionary<int, User> user_cache = new Dictionary<int, User>();
+
         protected override void LoadComplete()
         {
             base.LoadComplete();
@@ -53,15 +57,24 @@ namespace osu.Game.Overlays.Dashboard
                 switch (e.Action)
                 {
                     case NotifyCollectionChangedAction.Add:
-                        foreach (var u in e.NewItems.OfType<int>())
+                        foreach (int userId in e.NewItems.OfType<int>())
                         {
-                            var request = new GetUserRequest(u);
-                            request.Success += user => Schedule(() =>
+                            if (user_cache.TryGetValue(userId, out var user))
                             {
-                                if (playingUsers.Contains((int)user.Id))
-                                    userFlow.Add(createUserPanel(user));
-                            });
+                                addUser(user);
+                                continue;
+                            }
+
+                            var request = new GetUserRequest(userId);
+                            request.Success += u => Schedule(() => addUser(u));
                             api.Queue(request);
+
+                            void addUser(User u)
+                            {
+                                user_cache[userId] = u;
+                                if (playingUsers.Contains(userId))
+                                    userFlow.Add(createUserPanel(u));
+                            }
                         }
 
                         break;
--- a/osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
+++ b/osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System.Collections.Generic;
 using System.Collections.Specialized;
 using System.Linq;
 using osu.Framework.Allocation;
@@ -44,9 +43,6 @@ namespace osu.Game.Overlays.Dashboard
         [Resolved]
         private IAPIProvider api { get; set; }
 
-        // temporary, should be game-global but i don't want to add more manager classes for now.
-        private static readonly Dictionary<int, User> user_cache = new Dictionary<int, User>();
-
         protected override void LoadComplete()
         {
             base.LoadComplete();
@@ -57,24 +53,15 @@ namespace osu.Game.Overlays.Dashboard
                 switch (e.Action)
                 {
                     case NotifyCollectionChangedAction.Add:
-                        foreach (int userId in e.NewItems.OfType<int>())
+                        foreach (var u in e.NewItems.OfType<int>())
                         {
-                            if (user_cache.TryGetValue(userId, out var user))
+                            var request = new GetUserRequest(u);
+                            request.Success += user => Schedule(() =>
                             {
-                                addUser(user);
-                                continue;
-                            }
-
-                            var request = new GetUserRequest(userId);
-                            request.Success += u => Schedule(() => addUser(u));
+                                if (playingUsers.Contains((int)user.Id))
+                                    userFlow.Add(createUserPanel(user));
+                            });
                             api.Queue(request);
-
-                            void addUser(User u)
-                            {
-                                user_cache[userId] = u;
-                                if (playingUsers.Contains(userId))
-                                    userFlow.Add(createUserPanel(u));
-                            }
                         }
 
                         break;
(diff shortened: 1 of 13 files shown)


enoslayd:
Does it work if spectating someone on stable?

camellirite:
@enoslayd no

osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        [Resolved]
+        private IAPIProvider api { get; set; }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            playingUsers = spectatorStreaming.PlayingUsers.GetBoundCopy();
```
bdach, line 50: Not necessarily a problem with this PR specifically but for some reason the collection change callbacks on this get called every now and again, which makes this display pretty hard to use before it's fully populated as things move around a lot. Probably going to need to be looked at server-side but figured I'd say it just in case.

osu.Game/Users/UserPanel.cs
```diff
@@ -68,7 +70,8 @@ private void load()
             base.Action = ViewProfile = () =>
             {
                 Action?.Invoke();
-                profileOverlay?.ShowUser(User);
+                if (ShowProfileOnClick)
+                    profileOverlay?.ShowUser(User);
```
bdach, line 74: Not a fan of this to be fair. I'd prefer this to be something like

```csharp
            base.Action = () =>
            {
                if (Action != null)
                    Action.Invoke();
                else
                    profileOverlay?.ShowUser(User);
            };
```

with xmldoc saying that specifying no `Action` defaults to opening profile overlay. Or heck, even initialise `Action` to show overlay by default and just call that.

peppy: I did it this way because the other usages of this are compounding a new action in addition to showing the profile overlay (ie. the login dropdown hiding itself on display). Also thinking that we don't really want to block the "show profile" behaviour under normal scenarios. Maybe need to re-think this display and add a spectate button below or next to the card instead.

osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
peppy: The issue is we don't have an API endpoint to lookup multiple users at once right now, so the API responses come in one at a time. Need this to be fixed before deploying this new screen.

smoogipoo:
Gonna need a little bit of conflict resolution.

osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
```diff
@@ -0,0 +1,109 @@
...
+                AutoSizeAxes = Axes.Y,
+                Spacing = new Vector2(10),
+            };
+        }
+
+        [Resolved]
+        private IAPIProvider api { get; set; }
+
+        // temporary, should be game-global but i don't want to add more manager classes for now.
+        private static readonly Dictionary<int, User> user_cache = new Dictionary<int, User>();
```
smoogipoo, line 48: I really don't like this. If you're going to hack it that's fine, but not as a `static`... Storing `User` objects to a static gives me all sorts of bad feelings.

Highly prefer either removing this and waiting for the API issue to be resolved or moving it into `OsuGame` temporarily.

smoogipoo (changes requested):

osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
peppy: what's the reasoning, other than "bad feelings"? even at a game level it's still going to be roughly the same, just in a different place. just completely against static?

i'd like to do the same with case-when for instance, but kind of gave up on that in local usages.

smoogipoo: Bad feelings are the only reasons. What do you mean by "case-when"?

osu.Game/Screens/Play/SpectatorPlayer.cs
```diff
@@ -3,33 +3,61 @@
...
     {
-        [Resolved]
-        private SpectatorStreamingClient spectatorStreaming { get; set; }
+        private readonly Score score;
+
+        protected override bool CheckModsAllowFailure() => false; // todo: better support starting mid-way through beatmap
 
         public SpectatorPlayer(Score score)
-            : base(score)
+            : base(true, true)
```
bdach, line 24: Rider seems to tell me that this `base` delegation is redundant (and so it appears to be).

osu.Game/Screens/Play/SpectatorResultsScreen.cs
```diff
@@ -0,0 +1,47 @@
...
+                });
+            }
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (spectatorStreaming != null)
+                spectatorStreaming.OnUserBeganPlaying -= userBeganPlaying;
```
bdach, line 44: Wondering if worth duplicating this in an `OnExiting()` handler - not sure what guarantees we have wrt when this will get called when the stack is cleaning up exited screens.

bdach (changes requested):

osu.Game/Screens/Play/Spectator.cs
```diff
@@ -236,17 +305,43 @@ private void attemptStart()
...
+                beatmapPanelContainer.Clear();
+                onlineBeatmap = null;
+                return;
+            }
+
+            var req = new GetBeatmapSetRequest(state.BeatmapID.Value, BeatmapSetLookupType.BeatmapId);
             req.Success += res => Schedule(() =>
             {
-                beatmapPanelContainer.Child = new GridBeatmapPanel(res.ToBeatmapSet(rulesets));
+                if (state != this.state)
```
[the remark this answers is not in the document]
peppy: yeah, meant to be reference, to match the request with response handling.

osu.Game/Screens/Play/Spectator.cs
```diff
@@ -74,47 +86,90 @@ public Spectator([NotNull] User targetUser)
...
+                                Origin = Anchor.Centre,
+                            },
+                            watchButton = new PurpleTriangleButton
+                            {
+                                Text = "Start Watching",
+                                Width = 250,
+                                Anchor = Anchor.Centre,
+                                Origin = Anchor.Centre,
+                                Action = attemptStart
+                            }
```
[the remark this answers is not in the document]
peppy: Not sure I follow, at which point in time?

osu.Game/Screens/Play/SpectatorResultsScreen.cs
peppy: I think it should be fine due to the `Schedule` call guarding against execution

osu.Game/Screens/Play/Spectator.cs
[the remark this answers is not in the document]
bdach: I'm thinking right here on construction, by doing `Enabled = { Value = false }`, mostly for safety. I noticed in visual tests that this wasn't disabled even if state hasn't even arrived, and I could get it to crash by clicking on the button.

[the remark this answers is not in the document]
peppy: Still not really following. I don't have code anywhere to enable/disable this button as it is presumed that if the panel is already displayed, the button should be clickable.

If this is a test-specific issue, I think we should be resolving it locally there by fixing the missing dependency or whatever is causing the crash.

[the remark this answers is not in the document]
bdach: Looking again maybe this concern is a bit far-fetched, but I'm a bit thrown off by the overall treatment of `SpectatorState`. Another place is checking for `state` being null (`showBeatmapPanel`), while `attemptStart` isn't. I can think of one theoretical scenario in which the button will be clickable while `state` is null (if the `userBeganPlaying` fires a few frames after `LoadComplete()`), in which case it'll hard crash, but not sure if it's actually possible.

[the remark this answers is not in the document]
peppy: Okay sorry, I didn't realise you were talking about the spectator screen itself, not the listing (I just added a similar button there so assumed the review was related to it). This is a valid concern.

bdach:
I've made two tiny changes to `CurrentlyPlayingDisplay` - one to make the layout of the `PlayingUserPanel` slightly cleaner (86d283e), and one fixing the vertical spacing between rows being seemingly larger-than-desired (211510f).

Other than that, I think this is good to go as a first step :+1:

bdach (dismissed):

smoogipoo (approved):

PR #10614, opened 2020-10-29
Only play results panel animation once (and only for the local user)

peppy:
The score still incrementing was kind of by accident, but I think it feels fine as is (and it's already displayed on the contracted cards anyway) so leaving it that way.

### DIFF
--- a/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
+++ b/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
@@ -78,7 +78,7 @@ namespace osu.Game.Screens.Ranking.Expanded.Accuracy
         private Container<RankBadge> badges;
         private RankText rankText;
 
-        public AccuracyCircle(ScoreInfo score)
+        public AccuracyCircle(ScoreInfo score, bool withFlair)
         {
             this.score = score;
         }
--- a/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
+++ b/osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
@@ -78,7 +78,7 @@ namespace osu.Game.Screens.Ranking.Expanded.Accuracy
         private Container<RankBadge> badges;
         private RankText rankText;
 
-        public AccuracyCircle(ScoreInfo score, bool withFlair)
+        public AccuracyCircle(ScoreInfo score)
         {
             this.score = score;
         }
--- a/osu.Game/Screens/Ranking/Expanded/ExpandedPanelMiddleContent.cs
+++ b/osu.Game/Screens/Ranking/Expanded/ExpandedPanelMiddleContent.cs
@@ -120,7 +120,7 @@ namespace osu.Game.Screens.Ranking.Expanded
                                     Margin = new MarginPadding { Top = 40 },
                                     RelativeSizeAxes = Axes.X,
                                     Height = 230,
-                                    Child = new AccuracyCircle(score, withFlair)
+                                    Child = new AccuracyCircle(score)
                                     {
                                         Anchor = Anchor.Centre,
                                         Origin = Anchor.Centre,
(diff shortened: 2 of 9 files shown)


osu.Game/Screens/Ranking/Expanded/Accuracy/AccuracyCircle.cs
```diff
@@ -78,7 +78,7 @@ public class AccuracyCircle : CompositeDrawable
         private Container<RankBadge> badges;
         private RankText rankText;
 
-        public AccuracyCircle(ScoreInfo score)
+        public AccuracyCircle(ScoreInfo score, bool withFlair)
```
bdach, line 81: This param doesn't appear to do anything (yet? any more?)

osu.Game/Screens/Ranking/ScorePanelList.cs
```diff
@@ -117,19 +118,24 @@ public ScorePanel AddScore(ScoreInfo score)
                 d.Origin = Anchor.Centre;
             }));
 
-            if (SelectedScore.Value == score)
-                selectedScoreChanged(new ValueChangedEvent<ScoreInfo>(SelectedScore.Value, SelectedScore.Value));
-            else
+            if (IsLoaded)
```
bdach, line 121: I'd probably invert this check, although that's sort of subjective.

osu.Game/Screens/Ranking/Expanded/ExpandedPanelMiddleContent.cs
```diff
@@ -266,6 +270,9 @@ protected override void LoadComplete()
                         delay += 200;
                     }
                 }
+
+                if (!withFlair)
+                    FinishTransforms(true);
```
bdach, line 275: This is again subjective but the D rank animation (the "broken rank" one, wherein it kinda rotates as if it was hanging off an anchor point and then stays crooked) looks a bit weird if it finishes immediately. Not sure whether not having it finish would be better or worse, though.

bdach (changes requested):
Aside from the one unused param I don't have major objections.

osu.Game/Screens/Ranking/ScorePanelList.cs
```diff
@@ -162,12 +172,15 @@ private void selectedScoreChanged(ValueChangedEvent<ScoreInfo> score)
             expandedTrackingComponent.Margin = new MarginPadding { Horizontal = expanded_panel_spacing };
             expandedPanel.State = PanelState.Expanded;
 
-            // Scroll to the new panel. This is done manually since we need:
-            // 1) To scroll after the scroll container's visible range is updated.
-            // 2) To account for the centre anchor/origins of panels.
-            // In the end, it's easier to compute the scroll position manually.
-            float scrollOffset = flow.GetPanelIndex(expandedPanel.Score) * (ScorePanel.CONTRACTED_WIDTH + panel_spacing);
-            scroll.ScrollTo(scrollOffset);
+            SchedulerAfterChildren.Add(() =>
```
smoogipoo, line 175: Kind of a scary addition. Everything seems fine so I'll let it pass, but is there a reason for this now being required?

peppy: It allows for the fixes to the trigger value change flow that accompany it. Previously there was some dodge-as-fuck inline execution of the event in just the right way to make things work. Also panels were being expanded-then-contracted-then-expanded on load.

The reasoning for this being required is because at the point in time when this is run, the flow hasn't updated its alive children yet, so while the index (and offset) will be correct the `ScrollTo` gets limited to the scroll extent (500) instead of the actual panel location.

osu.Game/Screens/Ranking/ScorePanelList.cs
peppy: I dunno, not a huge fan of the double return..

image

osu.Game/Screens/Ranking/Expanded/ExpandedPanelMiddleContent.cs
peppy: Would be nice to have the little rotation part play separately, but in the interest of keeping the implementation simple, I think I'll leave it for now (unless you have a suggestion of how to make it work).

smoogipoo (approved):

PR #10615, opened 2020-10-29
Add applause effect to results screen

### LINKED ISSUE #10571: Add "applause" support
**Describe the new feature:** 
So far there hasn't been a single request for this. But this is an important part of the skin.

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10614
- [x] Depends on https://github.com/ppy/osu-resources/pull/116
- [x] Depends on https://github.com/ppy/osu/pull/10619
- Closes https://github.com/ppy/osu/issues/10571

Note that this doesn't add beatmap skin support. Will look at that later since it's not used in many scenarios and requires a considerable amount of implementation.

### DIFF
--- a/osu.Game/Audio/HitSampleInfo.cs
+++ b/osu.Game/Audio/HitSampleInfo.cs
@@ -50,9 +50,9 @@ namespace osu.Game.Audio
             get
             {
                 if (!string.IsNullOrEmpty(Suffix))
-                    yield return $"{Bank}-{Name}{Suffix}";
+                    yield return $"Gameplay/{Bank}-{Name}{Suffix}";
 
-                yield return $"{Bank}-{Name}";
+                yield return $"Gameplay/{Bank}-{Name}";
             }
         }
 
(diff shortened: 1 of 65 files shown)


osu.Game/Audio/HitSampleInfo.cs
```diff
@@ -50,9 +50,9 @@ public virtual IEnumerable<string> LookupNames
             get
             {
                 if (!string.IsNullOrEmpty(Suffix))
-                    yield return $"{Bank}-{Name}{Suffix}";
+                    yield return $"Gameplay/{Bank}-{Name}{Suffix}";
 
-                yield return $"{Bank}-{Name}";
+                yield return $"Gameplay/{Bank}-{Name}";
```
bdach, line 55: I believe moving these prefixes locally will break every single legacy skin (including beatmap skins) with custom gameplay hitsounds, unfortunately.

bdach (changes requested):

osu.Game/Audio/HitSampleInfo.cs
peppy: legacy skins already do local manipulation to remove all prefixes

peppy: Should be resolved (via https://github.com/ppy/osu/pull/10619).

bdach (dismissed):
looks good to me :+1:

smoogipoo (approved):

PR #10619, opened 2020-10-30
Change sample lookup structure to allow specifications outside of "Gameplay" prefix

peppy:
Until now it was not too simple to do skin lookups of samples which have prefixes outside of the Gameplay folder (in lazer resources).

This change adds the namespace specification in locally to the locations that didn't specify it before and allows the above to work correctly.

### DIFF
--- a/osu.Game/Skinning/LegacySkin.cs
+++ b/osu.Game/Skinning/LegacySkin.cs
@@ -417,10 +417,14 @@ namespace osu.Game.Skinning
 
         public override SampleChannel GetSample(ISampleInfo sampleInfo)
         {
-            var lookupNames = sampleInfo.LookupNames;
+            IEnumerable<string> lookupNames = null;
 
             if (sampleInfo is HitSampleInfo hitSample)
                 lookupNames = getLegacyLookupNames(hitSample);
+            else
+            {
+                lookupNames = sampleInfo.LookupNames.SelectMany(getFallbackNames);
+            }
 
             foreach (var lookup in lookupNames)
             {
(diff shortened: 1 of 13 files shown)


osu.Game/Skinning/LegacySkin.cs
```diff
@@ -433,32 +437,44 @@ public override SampleChannel GetSample(ISampleInfo sampleInfo)
...
+            {
                 // for compatibility with stable, exclude the lookup names with the custom sample bank suffix, if they are not valid for use in this skin.
                 // using .EndsWith() is intentional as it ensures parity in all edge cases
                 // (see LegacyTaikoSampleInfo for an example of one - prioritising the taiko prefix should still apply, but the sample bank should not).
-                lookupNames = hitSample.LookupNames.Where(name => !name.EndsWith(hitSample.Suffix, StringComparison.Ordinal));
+                foreach (var l in lookupNames)
+                {
+                    if (!l.EndsWith(hitSample.Suffix, StringComparison.Ordinal))
+                    {
+                        foreach (var n in getFallbackNames(l))
```
smoogipoo, line 453: Is this necessary? Isn't it effectively `getFallbackNames(getFallbackNames(lookup))`?

I'm not fully understanding this code, could it be written as:

```
var lookupNames = hitSample.LookupNames.SelectMany(getFallbackNames);

if (!UseCustomSampleBanks && !string.IsNullOrEmpty(hitSample.Suffix))
{
    // for compatibility with stable, exclude the lookup names with the custom sample bank suffix, if they are not valid for use in this skin.
    // using .EndsWith() is intentional as it ensures parity in all edge cases
    // (see LegacyTaikoSampleInfo for an example of one - prioritising the taiko prefix should still apply, but the sample bank should not).
    lookupNames = hitSample.LookupNames.Where(name => !name.EndsWith(hitSample.Suffix, StringComparison.Ordinal));
}

foreach (var l in lookupNames)
    yield return l;

// also for compatibility, try falling back to non-bank samples (so-called "universal" samples) as the last resort.
// going forward specifying banks shall always be required, even for elements that wouldn't require it on stable,
// which is why this is done locally here.
yield return hitSample.Name
```

peppy: Good catch – I moved the `getFallbackNames` out and forgot to remove this.

I'm fine without the way you have it, or just removing the nested call (not sure if yield is more efficient than `.Where`, that's the only reason I was using it).

smoogipoo: it's probably about the same either way.

bdach (approved):
seems correct to me :+1:

smoogipoo (approved):

PR #10620, opened 2020-10-30
Change shift-tab to cycle between available HUD visibility modes

### LINKED ISSUE #10570: Shift + Tab not working as expected
**Describe the bug:** 

After last additions of display modes to HUD overlay, shift + tab don't work as expected, it always moves between "Never" and "Hide during gameplay", no matter the option selected.

In the video below I do the next:

-Open menu search for HUD options and select "always" (in this case it was already selected)
-Play a beatmap in auto
-Press shift + tab 2 times (it changes to "Hide during gameplay")
-Press again shift + tab (now it changes to "Never")

I do this 4 or 5 times, and always it moves between this 2 options.

In auto it does not look bad, if you press shift + tab it hides the elements, and if you press again it shows them again, but if you are in gameplay, the elements are never shown.

https://drive.google.com/file/d/1qmdmTd8ciPgKO8ELsULhfBw3dNGVN67k/view?usp=sharing

**osu!lazer version:** 

2020.1020.0

**Logs:**

I don't think logs are needed, if they are please ask for them and I will attach them quickly.

peppy:
I ended up just making the key cycle. It feels good enough and keeps all options available and at a configuration level.

Closes https://github.com/ppy/osu/issues/10570

### DIFF
--- a/osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
+++ b/osu.Game.Tests/Visual/Gameplay/TestSceneHUDOverlay.cs
@@ -89,7 +89,7 @@ namespace osu.Game.Tests.Visual.Gameplay
         [Test]
         public void TestExternalHideDoesntAffectConfig()
         {
-            HUDVisibilityMode originalConfigValue = HUDVisibilityMode.HideDuringBreaks;
+            HUDVisibilityMode originalConfigValue = HUDVisibilityMode.HideDuringGameplay;
 
             createNew();
 
--- a/osu.Game/Configuration/HUDVisibilityMode.cs
+++ b/osu.Game/Configuration/HUDVisibilityMode.cs
@@ -12,9 +12,6 @@ namespace osu.Game.Configuration
         [Description("Hide during gameplay")]
         HideDuringGameplay,
 
-        [Description("Hide during breaks")]
-        HideDuringBreaks,
-
         Always
     }
 }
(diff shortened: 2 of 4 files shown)


smoogipoo:
What about someone that briefly wants to show the hud while they're spinning a spinner? This is more common than you'd think (I also do it). Are you expecting that person to either:
1. Press Shift+Tab another 3 times to hide the hud.
2. Press Shift+Tab 1 time to go to "during breaks", and then press it another 2 times to show it again when another spinner comes up?

Sounds like a bit of a mess. Are you against splitting it out into two settings, a hidden "override" one (visible/hidden) per-session, and a "default" user one (always/during gameplay/during breaks)?

I don't think users are going to be changing their default much/ever - I'd almost argue that "during gameplay" is something we don't want ever going by the discussions we've had in osu!stable regarding multiplayer (i.e. the hud should always be shown during a break).

peppy:
I'm fine with removing "hide during gameplay" and leaving it as a ternary. The idea here is to keep this simple (from an implementation and a UX perspective). Hopefully you are okay with it is there's only three options?

peppy:
Also I'd add that if the use case for shift-tab is to "quickly view hud temporarily", that should be a separate key binding (maybe shift or ctrl?) which only shows the HUD while held down.

smoogipoo:
People do that, yes. Personally I don't really use shift+tab, but I do that with tab - hold z/x and tab with my pinky. It's only longer spinners where an upcoming break can be felt by the progression in the track.

A binding to temporarily show the hud makes a lot of sense to me. That way the shift+tab setting becomes a never/always toggle, and doesn't even need a binding (can be changed only through settings).

peppy:
Are you against leaving this ternary toggle on shift-tab and adding a new key for the temporary display?

smoogipoo:
Well a ternary isn't needed if we remove "hide during breaks", right? The way I see it doing that would also mean "never" shouldn't exist.

As for it being a bound option, I'm not against it per-se, just thought it'd simplify things even further. But I guess for people coming from stable it's the more natural process...

smoogipoo (approved):

PR #10621, opened 2020-10-30
Add momentary HUD toggle

peppy:
Hold ctrl and get HUD.

Not sure if "Hold for HUD" is the best name for the binding, but it reads best to me. Open to alternatives.


enoslayd:
Temporarily show HUD?

peppy:
Sounds fine too, will wait for more opinions.

smoogipoo:
Yeah this feels really nice to me. I prefer the "hold for hud" naming and that can be adjusted later.

PR #10623, opened 2020-10-30
Fix "bounce" when metadata container text is empty in song select

### LINKED ISSUE #9335: Beatmap tags section bouncing when there is no beatmap source in song select

cadon0:
Closes #9335
Supersedes / closes #9338

Previously an empty `source` metadata container would push down the neighboring `tags` metadata container, before immediately initiating a fade out due to having no text. This made the `tags` container appear to "bounce"

The change can be verified with the ["all except source"](https://github.com/ppy/osu/blob/aa1daa0ad50205ea96da37b358e71820c04e41da/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapDetails.cs#L59) visual test. Also see the "all metrics" test for intended behaviour

Before:

After:

### DIFF
--- a/osu.Game/Screens/Select/BeatmapDetails.cs
+++ b/osu.Game/Screens/Select/BeatmapDetails.cs
@@ -300,6 +300,7 @@ namespace osu.Game.Screens.Select
 
             public MetadataSection(string title)
             {
+                Alpha = 0;
                 RelativeSizeAxes = Axes.X;
                 AutoSizeAxes = Axes.Y;
 


bdach:
This change is [basically what I suggested](https://github.com/ppy/osu/pull/9338#discussion_r444414609) in the previous attempt to fix this but I'll have to re-check again if it does behave ok.

bdach (approved):
I *think* this should be correct.

Joehuu (approved):
LGTM

smoogipoo (approved):

PR #10637, opened 2020-10-31
Fix potential null reference when loading background

peppy:
As seen in
https://discordapp.com/channels/188630481301012481/188630652340404224/772094427342569493.
Caused due to async load of the loader, which means it may not be ready
before Next() is called.


bdach:
As mentioned on discord, pretty big oversight on my part when I was fiddling the code. The long running load never made any sense.

I've taken the liberty of adjusting the test scene to remove the async load there as well as unnecessary.

bdach (approved):

PR #10640, opened 2020-10-31
Fix music being paused after retrying & exiting timeshift map

### LINKED ISSUE #10632: Music stop playing after retrying in timeshift lobby
**Describe the bug:** In timeshift lobby, when playing a map and then retrying, after exiting to the lobby, the music won't be played for any map.

**Screenshots or videos showing encountered issue:** https://youtu.be/YKUlBK2lp2Q

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5468947/runtime.log)

bdach:
Closes #10632.

# Summary

Regression of sorts from #10406. On restart, `Player` used `MusicController.Stop()` to stop the current track to avoid having it play while the retry/`PlayerLoader` sequence ran, but `MusicController.Stop()` had an implicit meaning of being a user-requested pause, which means that other methods like the beatmap changed callback or `EnsurePlayingSomething()` would not resume playback due to respecting what was seemingly the user's choice to pause playback.

Resolve [as proposed on discord](https://discord.com/channels/188630481301012481/188630652340404224/772107193309986836) by adding a parameter that allows to clarify intent.

# Remarks

Not sure on the param being default, or its default value - current variant is mostly backwards compatibility mode, but unsure if backwards compatibility is to be preserved.

I also replaced one direct `MusicController.CurrentTrack.Stop()` call in tests to use this new method.

### DIFF
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -187,7 +187,7 @@ namespace osu.Game.Overlays
         /// Specifying <c>true</c> will ensure that other methods like <see cref="EnsurePlayingSomething"/>
         /// will not resume music playback until the next explicit call to <see cref="Play"/>.
         /// </param>
-        public void Stop(bool requestedByUser = true)
+        public void Stop(bool requestedByUser = false)
         {
             IsUserPaused |= requestedByUser;
             if (CurrentTrack.IsRunning)
@@ -201,7 +201,7 @@ namespace osu.Game.Overlays
         public bool TogglePause()
         {
             if (CurrentTrack.IsRunning)
-                Stop();
+                Stop(true);
             else
                 Play();
 
--- a/osu.Game/Overlays/MusicController.cs
+++ b/osu.Game/Overlays/MusicController.cs
@@ -182,9 +182,14 @@ namespace osu.Game.Overlays
         /// <summary>
         /// Stop playing the current track and pause at the current position.
         /// </summary>
-        public void Stop()
+        /// <param name="requestedByUser">
+        /// Whether the request to stop was issued by the user rather than internally.
+        /// Specifying <c>true</c> will ensure that other methods like <see cref="EnsurePlayingSomething"/>
+        /// will not resume music playback until the next explicit call to <see cref="Play"/>.
+        /// </param>
+        public void Stop(bool requestedByUser = true)
         {
-            IsUserPaused = true;
+            IsUserPaused |= requestedByUser;
             if (CurrentTrack.IsRunning)
                 CurrentTrack.Stop();
         }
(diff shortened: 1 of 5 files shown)


osu.Game/Overlays/MusicController.cs
```diff
@@ -182,9 +182,14 @@ public bool Play(bool restart = false)
         /// <summary>
         /// Stop playing the current track and pause at the current position.
         /// </summary>
-        public void Stop()
+        /// <param name="requestedByUser">
+        /// Whether the request to stop was issued by the user rather than internally.
+        /// Specifying <c>true</c> will ensure that other methods like <see cref="EnsurePlayingSomething"/>
+        /// will not resume music playback until the next explicit call to <see cref="Play"/>.
+        /// </param>
+        public void Stop(bool requestedByUser = true)
```
peppy, line 190: I think I'd prefer the default to be false, personally. Also maybe we want a matching parameter on `Play`..

@smoogipoo thoughts?

peppy (changes requested):

osu.Game/Overlays/MusicController.cs
smoogipoo: `false` probably makes more sense in most usages, so I'd agree.

peppy (approved):

osu.Game/Overlays/MusicController.cs
```diff
@@ -166,10 +166,17 @@ public void EnsurePlayingSomething()
...
+        /// Specifying <c>true</c> will ensure that other methods like <see cref="EnsurePlayingSomething"/>
+        /// will resume music playback going forward.
+        /// </param>
         /// <returns>Whether the operation was successful.</returns>
-        public bool Play(bool restart = false)
+        public bool Play(bool restart = false, bool requestedByUser = false)
         {
-            IsUserPaused = false;
+            if (requestedByUser)
+                IsUserPaused = false;
```
smoogipoo, line 179: This is very weird to me. So if you `Play(.., false)`, the music controlled can potentially be left in `IsUserPaused = true` but the track still playing?

peppy: yep, for instance during gameplay. right now song select forcefully un(user)pauses on returning, but that may not be what we want at the end of the day, not sure.

smoogipoo: This variable needs to be renamed then. It can't remain as IsUser**Paused** when potentially it's not paused at all.

Needs to be UserPauseRequested, and appropriate documentation that says it doesn't necessarily mean the track is paused.

peppy: no issues with that, go for it

PR #10655, opened 2020-11-02
Downgrade SignalR package to avoid startup crashes on iOS

peppy:
Due to an [issue](https://github.com/dotnet/runtime/issues/35988) at xamarin's end (which *should* have been fixed via [this pr](https://github.com/xamarin/xamarin-macios/pull/8615) but still fails on latest preview release) we can't support this just yet.

### DIFF
--- a/osu.iOS.props
+++ b/osu.iOS.props
@@ -75,7 +75,7 @@
   </ItemGroup>
   <!-- See https://github.com/dotnet/runtime/issues/35988 (can be removed after Xamarin uses net5.0 / net6.0) -->
   <PropertyGroup>
-    <NoWarn>NU1605</NoWarn>
+    <NoWarn>$(NoWarn);NU1605</NoWarn>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.SignalR.Client" Version="3.0.3" />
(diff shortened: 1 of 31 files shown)


smoogipoo (dismissed):

peppy:
Don't merge just yet, looking to be too late in the method to actually stop the issue (possibly `static ctor` at net core's end..?)

peppy:
This looks to be the best solution for now. Leaving it as a local downgrade is a bit iffy as behaviour may differ as we move forward, but I think it's better than targeting the lower version at the `osu.Game` level?

I've tested this works on device and simulator (can spectate from device with no issues).

osu.iOS.props
```diff
@@ -73,6 +73,14 @@
     <PackageReference Include="ppy.osu.Framework.iOS" Version="2020.1029.1" />
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.1030.0" />
   </ItemGroup>
+  <!-- See https://github.com/dotnet/runtime/issues/35988 (can be removed after Xamarin uses net5.0 / net6.0) -->
+  <PropertyGroup>
+    <NoWarn>NU1605</NoWarn>
```
smoogipoo, line 78:
```suggestion
    <NoWarn>$(NoWarn);NU1605</NoWarn>
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #10658, opened 2020-11-02
Add very basic connection status logging for spectator streaming client

peppy:
Just enough to help with tracking down connection issues.


bdach (approved):
seems innocuous enough :+1:

PR #10659, opened 2020-11-02
Fix editor seek transform seeking too much

smoogipoo:
If the game lags, `time` can be much greater than `EndTime`, which would cause a seek very far into the future.

I haven't written a test for this since the same logic is done elsewhere, and this was only done this way as a measure for reducing implementation overhead at the time of writing.
Most noticeable with my hitobject pooling changes (https://github.com/smoogipoo/osu/tree/hitobject-pooling).

### DIFF
--- a/osu.Game/Screens/Edit/EditorClock.cs
+++ b/osu.Game/Screens/Edit/EditorClock.cs
@@ -266,8 +266,15 @@ namespace osu.Game.Screens.Edit
         {
             public override string TargetMember => nameof(currentTime);
 
-            protected override void Apply(EditorClock clock, double time) =>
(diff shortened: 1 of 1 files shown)


peppy:
Easily reproducible by breakpointing during a seek, for what it's worth. Has come up many times.

peppy:
Sure we don't want to change the underlying interpolation functions? Do we use them in an out-of-bounds way anywhere else?

smoogipoo:
If you want that to happen, make an issue for it. Because it's a framework change that has many of its own implications.

peppy (approved):

PR #10660, opened 2020-11-02
Fix looping mode not being set on first track after entering song select

### LINKED ISSUE #10656: Song select doesn't replay the first track opened with
Repro: https://drive.google.com/file/d/1rLhGFnriUWDZxcmXtmJTbHe1HRm7B0CC/view?usp=sharing

The first track song select is opened with doesn't restart playing, which causes a new map to get picked.

peppy:
Closes #10656.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -636,7 +636,10 @@ namespace osu.Game.Screens.Select
 
         private void endLooping()
         {
-            Debug.Assert(isHandlingLooping);
+            // may be called multiple times during screen exit process.
+            if (!isHandlingLooping)
+                return;
+
             music.CurrentTrack.Looping = isHandlingLooping = false;
 
             music.TrackChanged -= ensureTrackLooping;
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -619,12 +618,30 @@ public override bool OnExiting(IScreen next)
...
+            Debug.Assert(!isHandlingLooping);
+
+            music.CurrentTrack.Looping = isHandlingLooping = true;
+
+            music.TrackChanged += ensureTrackLooping;
+        }
+
+        private void endLooping()
+        {
+            Debug.Assert(isHandlingLooping);
```
bdach, line 639: As evidenced by the failing test, `endLooping()` can actually run twice on the same screen if it is suspended, and then exited from due to unwinding the screen stack to a screen lower down. Probably fine to remove this assert, or turn it into a guard statement? Not sure.

bdach (changes requested):

bdach:
Weird thing is, I can't seem to reproduce... I suppose this can't hurt either, though - hopefully there's no conceivable way looping can *not* get set anymore.

bdach (approved):

smoogipoo (approved):

PR #10661, opened 2020-11-02
Debounce calls to UpdateTernaryStates

peppy:
Just something I noticed in passing recently which may help with reducing performance overhead of some batch operations.


bdach (approved):
I can't see any harm in this, and it seems to have the desired effect :+1:

PR #10667, opened 2020-11-03
Fix editor clock not always remaining stopped when dragging timeline

### LINKED ISSUE #10461: Dragging upper timeline and playing song causes audio to freeze
Dragging the upper timeline while the song is paused and then playing the song causes audio to freeze. Somehow I've also been able to drag and start the audio at the same time, causing weird half freezing/stuttering behaviour. (Seen last in the video)

video: https://youtu.be/UMahcTCbTuc
osu!lazer version: 2020.1009.0

peppy:
Closes https://github.com/ppy/osu/issues/10461.

Not sure there's a better way to handle this. Optimally, the track would not be allowed to start during a drag, but this is hard to enforce from the timeline itself.

Also, while fixing this I managed to reproduce two other bugs (sample pause not succeeding / clock reporting paused while audio still playing) but these are not related to the fix here.


bdach (approved):
Seems fine.

For what it's worth I can't seem to reproduce either issue I saw on stream after abusing the timeline for a short while.

smoogipoo (approved):

PR #10673, opened 2020-11-03
Add opacity adjustment for waveform editor

### LINKED ISSUE #10666: Customize waveform opacity in editor
Feature mentioned on ppy's twitch by _**wouldwriteagain**_.

Could be similar to how you use a slider to change background dimming I suppose.

osu!_oNfLNTbHH1

peppy:
Good chance to start adding editor configuration persistence and fix up some issues with the menu display.

Closes #10666.


bdach (approved):
Not 100% about the usefulness of this, but it is pretty nice to have and I have no objections to the diff, so :+1:

PR #10674, opened 2020-11-03
Change editor timeline mouse wheel handling to scroll by default (and zoom with alt held)

### LINKED ISSUE #10670: Change scroll wheel timeline interaction in the editor
Using the scroll wheel on the timeline causes it to zoom in and out, which is less useful than having it scroll through the objects. Proposed change is identical to stable, which zooms in and out while holding alt + scrolling, and moves across the timeline without alt.

peppy:
The previous logic would zoom for non-precision devices, and scroll for precision. I've changed this to perform the same operations for both device types (for now).

Gesture support (as mentioned in the comment) is probably far enough away that we don't need to mention it for now.

Closes https://github.com/ppy/osu/issues/10670.

### DIFF
--- a/osu.Game.Tests/Visual/Editing/TestSceneZoomableScrollContainer.cs
+++ b/osu.Game.Tests/Visual/Editing/TestSceneZoomableScrollContainer.cs
@@ -7,13 +7,14 @@ using osu.Framework.Graphics.Colour;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shapes;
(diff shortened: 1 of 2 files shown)


bdach:
I've pushed two changes. 4b528e8 is cosmetic, but ddf0d75 is pretty important - the `AltPressed` branch wasn't early-returning `true`, therefore the on the editor timeline trying to zoom with <kbd>Alt</kbd> would actually both zoom and seek, therefore behaving pretty erratically.

Other than that, looks good to go as far as I can tell :+1:

bdach (approved):

PR #10676, opened 2020-11-03
Fix skin changes resulting in incorrectly applied transforms in MainCirclePiece

### LINKED ISSUE #6104: Visual problem with fading-out hitobjects when switching skin or skin settings during gameplay
**Describe the bug:** Switching between skins, you can see some weird behaviour with hitobjects that are still doing their "fade-out" animation. (You can reproduce this also changing skin in the settings overlay, but using the toggle in Visual Settings, if a map has its skin, is much easier)
Map used in the video: https://osu.ppy.sh/beatmapsets/208123#osu/490185

**Screenshots or videos showing encountered issue:** https://streamable.com/alpwb

**osu!lazer version:** 2019.913.0

### LINKED ISSUE #9923: Bloom/glow on the default skin's hitcircles sometimes not displayed
**Describe the bug:** 
When a user toggles the beatmap skin/beatmap hitsounds setting on the beatmap loading screen, the glow surrounding hitcircles will disappear. 
Restarting the beatmap without touching those settings fixes it, or if that doesn't happen, it'll randomly come back in the middle of gameplay.

https://streamable.com/9c5cyb

**osu!lazer version:** 
2020.820.0

**Logs:**

peppy:
- Closes https://github.com/ppy/osu/issues/6104
- Closes https://github.com/ppy/osu/issues/9923
- [x] Depends on https://github.com/ppy/osu/pull/10685.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/MainCirclePiece.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/MainCirclePiece.cs
@@ -42,10 +42,13 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
         private readonly IBindable<Color4> accentColour = new Bindable<Color4>();
         private readonly IBindable<int> indexInCurrentCombo = new Bindable<int>();
 
+        [Resolved]
+        private DrawableHitObject drawableObject { get; set; }
+
         [BackgroundDependencyLoader]
-        private void load(DrawableHitObject drawableObject)
+        private void load()
         {
-            OsuHitObject osuObject = (OsuHitObject)drawableObject.HitObject;
+            var osuObject = (OsuHitObject)drawableObject.HitObject;
 
             state.BindTo(drawableObject.State);
             state.BindValueChanged(updateState, true);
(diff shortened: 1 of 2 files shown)


bdach (dismissed):
I *think* this is right. Seems to resolve the issue, at the least.

Joehuu:
The same fix needs to be applied to https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs:

https://github.com/ppy/osu/blob/f3d5e7b15a24f41268fe8e4a9c05e62edcc3306f/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs#L142-L169

That will fix https://github.com/ppy/osu/issues/6104 (streamable video expired, but I saw it before).

bdach:
Good catch. I'll copy over in a sec.

bdach:
Actually, now that I look at it, other places might need doing the same things or similar... A good way to see that something is wrong is spamming the "Enable beatmap skins" setting over and over again. This resolution fixes normal hit circles, but sliders will still have circles flickering at their start position. It almost feels like this should be fixed at a higher level rather than peppering absolute sequences everywhere, but I'm not sure if that's even possible.

Dismissing approve pending discussion.

peppy:
I guess we could consider a new class `SkinnableGameplayDrawable` which takes the hitobject, and a method for applying transforms similar to `DrawableHitObject`'s `UpdateState` logic?

@smoogipoo thoughts on this in general?

smoogipoo:
I'm not sure if `SkinnableGameplayDrawable` is as good of an idea as exposing individual events instead of/in addition to the "overall" custom state function. My reasoning is that the same issue can occur with mods, which aren't drawables.

smoogipoo:
I think for now I'd prefer addding the delays on a per-case basis, and leave it to future considerations. I'd already done this with spinner components, iirc.

peppy:
a property inside `Result` maybe?

smoogipoo:
Yeah I think that's a fine idea.

bdach:
Seems good. Not approving as there are conflicts - they'll probably go away when the dependent PR is merged back here, but the approve will get eaten anyway, so not much point.

I suppose if we're going down the path of fixing locally, the other issues can be fixed in separate PRs.

smoogipoo (approved):

PR #10678, opened 2020-11-03
Fix LoadComponentAsync calls potentially occuring after beatmap wedge disposal

peppy:
As seen in https://ci.appveyor.com/project/peppy/osu/builds/36109658/tests.

Also adds cancellation logic for good measure.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapInfoWedge.cs
+++ b/osu.Game/Screens/Select/BeatmapInfoWedge.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Threading;
 using JetBrains.Annotations;
 using osuTK;
 using osuTK.Graphics;
@@ -85,6 +86,8 @@ namespace osu.Game.Screens.Select
 
         private WorkingBeatmap beatmap;
 
+        private CancellationTokenSource cancellationSource;
+
         public WorkingBeatmap Beatmap
         {
             get => beatmap;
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Select/BeatmapInfoWedge.cs
```diff
@@ -93,10 +96,12 @@ public WorkingBeatmap Beatmap
...
+                cancellationSource?.Cancel();
+                cancellationSource = new CancellationTokenSource();
 
                 beatmapDifficulty?.UnbindAll();
-                beatmapDifficulty = difficultyManager.GetBindableDifficulty(beatmap.BeatmapInfo);
-                beatmapDifficulty.BindValueChanged(_ => updateDisplay());
+                beatmapDifficulty = difficultyManager.GetBindableDifficulty(beatmap.BeatmapInfo, cancellationSource.Token);
+                beatmapDifficulty.BindValueChanged(_ => Schedule(updateDisplay));
 
                 updateDisplay();
```
bdach, line 106: Shouldn't this one also technically be scheduled for even better safety? Technically there's nothing stopping someone from setting `Beatmap` on a disposed wedge. There's also another `updateDisplay()` call registered on `ruleset.ValueChanged()` in BDL, so maybe best to wrap the entire body of `updateDisplay()` with a schedule so that it can't be ever called without one?

Probably highly theoretical, but if we're fixing anyways...

peppy: Yeah I did consider just wrapping the whole method (and using `Scheduler.AddOnce`). The only reason I left as-is is because it allows the `LoadComponentAsync` to potentially run earlier than otherwise (one frame?) by having it inline.

smoogipoo (approved):

PR #10679, opened 2020-11-03
Add the ability to delete slider control points using shift+right click

### LINKED ISSUE #10672: Allow deletion of control points on sliders using SHIFT right click when hovering on them
At the moment it deletes the entire slider, only way to delete control points is using the right click menu.

peppy:
Closes https://github.com/ppy/osu/issues/10672.

In two minds about how this should be implemented but went in this direction initially. The other way would be to add local handling of Shift-Right Click inside PathControlPointPiece (which is already doing mouse handling itself) but I convinced myself that handling this weird combination in one place is better.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
@@ -18,7 +18,6 @@ using osu.Game.Graphics.UserInterface;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Osu.Objects;
-using osuTK;
 using osuTK.Input;
 
 namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
@@ -115,9 +114,6 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
         public void OnReleased(PlatformAction action)
         {
         }
-
-        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) =>
-            Pieces.Any(p => p.ReceivePositionalInputAt(screenSpacePos));
 
         private void selectPiece(PathControlPointPiece piece, MouseButtonEvent e)
         {
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
@@ -18,6 +18,7 @@ using osu.Game.Graphics.UserInterface;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Osu.Objects;
+using osuTK;
 using osuTK.Input;
 
 namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
@@ -105,7 +106,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
             switch (action.ActionMethod)
             {
                 case PlatformActionMethod.Delete:
-                    return deleteSelected();
+                    return DeleteSelected();
             }
 
             return false;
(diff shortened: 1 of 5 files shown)


osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
```diff
@@ -115,6 +116,9 @@ public void OnReleased(PlatformAction action)
         {
         }
 
+        public override bool ReceivePositionalInputAt(Vector2 screenSpacePos) =>
+            Pieces.Any(p => p.ReceivePositionalInputAt(screenSpacePos));
```
bdach, line 120: Is this required? This override is what seems to specifically break tests (because of this the visualiser can't receive click events outside of any path piece, breaking deselection), and it seems to work fine without as far as I can tell.

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
```diff
@@ -219,18 +219,28 @@ internal void HandleDeselected(SelectionBlueprint blueprint)
         /// </summary>
         /// <param name="blueprint">The blueprint.</param>
         /// <param name="state">The input state at the point of selection.</param>
-        internal void HandleSelectionRequested(SelectionBlueprint blueprint, InputState state)
+        /// <returns>Whether right click was handled.</returns>
```
bdach, line 222: This is a little specific, maybe... I believe there could be a way to make this more generic.

As this is supposed to enable aborting the selection process due to some secondary action being performed, how about inverting this, saying that the return value of `true` indicates that the selection process should be aborted, and then doing `&=` instead of `|=` in `beginClickSelection`?

bdach (changes requested):
I feel like this is pretty alright in the current state, but maybe could be improved upon a bit still code quality-wise, and there's a regression evident in the tests.

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
peppy: Without this, points that don't overlap the slider body will not be deletable. Easy way to test is making a slider with four points, where the third doesn't follow a straight line.

bdach: Weird, I explicitly remember guessing it was for that and testing such a scenario - unless I screwed up, seemed to work.

osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs
peppy: Not 100% sure I follow this, are you saying this method should be returning false when quick deletion handles?

bdach: That's pretty much what I had in mind, yeah.

peppy: I feel that this change reads worse (because the main purpose here is to block the input event from traversing to the context menu), but I've changed it and don't mind either way.

peppy:
I looked at adding test coverage of this but it's... not simple.

- `SelectionBlueprintTestScene` doesn't have a `ComposerBlueprintContainer` / `SelectionHandler`. I tried to add these but that test scene is a touch archaic and each implementation uses it in a slightly different way (mania is especially hard to work with due to the `base.Content` override.
- Without the above classes, deletion is not possible as it relies on the input flow from these.

I'll leave the test here (which works with sufficient changes, but I broke too many other tests in the process).

```csharp
        [Test]
        public void TestControlPointDeletion()
        {
            moveMouseToControlPoint(0);
            AddStep("click", () => InputManager.Click(MouseButton.Left));
            checkControlPointSelected(0, true);
            AddStep("shift+right click", () =>
            {
                InputManager.PressKey(Key.ShiftLeft);
                InputManager.Click(MouseButton.Right);
                InputManager.ReleaseKey(Key.ShiftLeft);
            });
            checkControlPointCount(2);
        }
```

bdach (approved):
Tested manually, could not find anything funny. I'm fine with not having automated tests for this for now - that bridge can be crossed if this ever breaks. :+1:

smoogipoo (approved):

Issue #10668, opened 2020-11-03
Add commonly used editor shortcuts


UberFazz:
Flip shortcuts (ctrl+g, h, j) and hitsound shortcuts for samples/additions (ctrl/shift+q, w, e, r) aren't present.

camellirite:
j and k to shift objects by a tick too

UberFazz:
forgot to mention ctrl+left click on a white anchor to make it red as well

Joehuu:
If they are in the [Shortcut key reference wiki page](https://osu.ppy.sh/help/wiki/Shortcut_key_reference#beatmap-editor), they will probably be all implemented.

Stable shortcuts should be all documented in wiki anyway.

PR #10680, opened 2020-11-03
Add key bindings for flip and reverse patterns

peppy:
A start towards https://github.com/ppy/osu/issues/10668.


bdach:
`OnKeyDown()` wasn't checking whether it could perform the action it wanted to, which could actually lead to a crash if pressing the reverse hotkey while no object was selected. I've fixed this. (Briefly considered adding a test scene, but it'd be tedious and long and this shouldn't really ever get broken in its current state.)

Can't see much more fault with this, though. :+1:

bdach (approved):

peppy:
Hmm, I actually checked and the cases where `CanX` was false seemed to be noop. I guess that's not the case for the actual implementations.

PR #10685, opened 2020-11-04
Add new method handling hit state specifically

peppy:
Not sure whether the `StartTime` method should have `State` in its name, please advise.

# Breaking Changes

## DrawableHitObject.UpdateStateTransforms is now obsoleted.

Until now UpdateStateTransforms was applying results offsets to StartTime. This didn't cover the case of a HitObject with duration, where the call would be made with `StartTime + hitOffset` rather than `EndTime + hitOffset`.

To resolve this, a two new methods have been added to improve clarity on what each UpdateState call should be used for.

`UpdateStartTimeStateTransforms`: `hitObject.StartTime`, regardless of `ArmedState`
`UpdateHitStateTransforms`: `hitObject.EndTime` + `result.TimeOffset`, potentially dependent on `ArmedState`.

### To migrate

- For all `DrawableHitObjects`, move any non-state dependent transforms to `UpdateStartTimeStateTransforms`
- For objects with no duration, rename any remaining `UpdateStateTransforms` to `UpdateHitStateTransforms`.
- For objects with duration, remove any manually applied `BeginDelayedSequence(hitObject.Duration)` and use `UpdateHitStateTransforms()` instead of `UpdateStateTransforms()`

### DIFF
--- a/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
+++ b/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
@@ -6,9 +6,8 @@ using System.Collections.Generic;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
-using osu.Game.Rulesets.Objects;
-using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Catch.UI;
+using osu.Game.Rulesets.Objects.Drawables;
 using osuTK;
 using osuTK.Graphics;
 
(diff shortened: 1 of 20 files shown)


bdach:
Diff-wise I like this change a lot, definitely a code quality improvement, and all the changes seem correct as far as I can tell :+1:  I did have a bit of a hard time justifying the removal of the `Expire(true)` calls, but breakpointing in at those removing them doesn't seem to change the involved objects' start times.

Going to throw a tentative approve, but I don't feel comfortable merging without @smoogipoo's review as I haven't touched much of the ruleset transform logic yet.

bdach (approved):

smoogipoo (approved):
lgtm

PR #10689, opened 2020-11-05
Use asynchronous load instead of constructor for osu! DrawableHitObjects

smoogipoo:
Full set of changes can be found in https://github.com/ppy/osu/compare/master...smoogipoo:hitobject-pooling

The purpose of this PR is to move as much as possible out of the constructor for osu! DHOs. Pooling requires 0 implementation to remain in the ctor. There's a few places this can't be done, such as `DrawableSliderTail`, which will eventually receive an `ApplyParent(DrawableHitObject)` method to facilitate the remaining removals.

Also note that where bindables are used (stack height, position, scale), the `true` parameter can be removed after the pooling implementation but have to remain here for now.

Final questionable change is `LoadSamples` being moved from `DHO.load()` to `DHO.LoadAsyncComplete()`, which is required because `DrawableOsuHitObject` overrides `AddInternal` to forward it to another container. I don't see this as a huge change since it's still in the async process and only very slightly delayed. Consider it a temporary change as it'll be moved into the `Apply(HitObject)` method, which will still be called from `LoadAsyncComplete()`.

Will unblock after I've submitted a few more PRs.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
@@ -20,62 +20,50 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 {
     public class DrawableSlider : DrawableOsuHitObject, IDrawableHitObjectWithProxiedApproach
     {
+        public new Slider HitObject => (Slider)base.HitObject;
+
         public DrawableSliderHead HeadCircle => headContainer.Child;
         public DrawableSliderTail TailCircle => tailContainer.Child;
 
-        public readonly SliderBall Ball;
-        public readonly SkinnableDrawable Body;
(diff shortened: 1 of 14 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
```diff
@@ -20,62 +20,50 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
 {
     public class DrawableSlider : DrawableOsuHitObject, IDrawableHitObjectWithProxiedApproach
     {
+        public new Slider HitObject => (Slider)base.HitObject;
```
peppy, line 23: not 100% sure i feel about `public new` for this, but i guess it does clean things up quite a bit.

peppy (approved):

PR #10690, opened 2020-11-05
Remove stored hitobject references from osu! components

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10689

Full set of changes can be found in https://github.com/ppy/osu/compare/master...smoogipoo:hitobject-pooling

DHOs will be reused with different hitobject models, so the model shouldn't be stored in any drawable components.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
@@ -18,8 +18,6 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
     {
         private DrawableSpinner drawableSpinner;
 
-        private Spinner spinner;
-
         private const float initial_scale = 1.3f;
         private const float idle_alpha = 0.2f;
         private const float tracking_alpha = 0.4f;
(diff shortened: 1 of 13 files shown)


osu.Game.Rulesets.Osu/Skinning/LegacyNewStyleSpinner.cs
```diff
@@ -75,23 +75,21 @@ protected override void LoadComplete()
 
         private void updateStateTransforms(DrawableHitObject drawableHitObject, ArmedState state)
         {
-            if (!(drawableHitObject is DrawableSpinner))
+            if (!(drawableHitObject is DrawableSpinner spinner))
                 return;
 
-            var spinner = drawableSpinner.HitObject;
```
peppy, line 81: okay if i bring this back?

smoogipoo: yeah that's fine

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs
```diff
@@ -130,18 +127,18 @@ private void updateStateTransforms(DrawableHitObject drawableHitObject, ArmedSta
             if (!(drawableHitObject is DrawableSpinner))
                 return;
 
-            using (BeginAbsoluteSequence(spinner.StartTime - spinner.TimePreempt, true))
+            using (BeginAbsoluteSequence(drawableSpinner.HitObject.StartTime - drawableSpinner.HitObject.TimePreempt, true))
```
peppy, line 130: same here, would like to have a local variable

peppy (dismissed):

peppy (approved):

PR #10694, opened 2020-11-05
Fix paused samples potentially getting stuck in a playing state in rapid toggling

### LINKED ISSUE #10693: Pausing pausable samples can fail in a weird state
Found this while testing another issue the other day. Am currently investigating but leaving an issue open in case I can't figure out a solution.

`PausableSkinnableSound`'s state at the point of failure:

image

peppy:
Closes #10693. Should be obvious why.

### DIFF
--- a/osu.Game/Skinning/PausableSkinnableSound.cs
+++ b/osu.Game/Skinning/PausableSkinnableSound.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
+using osu.Framework.Threading;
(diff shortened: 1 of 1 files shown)


osu.Game/Skinning/PausableSkinnableSound.cs
```diff
@@ -66,8 +72,15 @@ public override void Play()
 
         public override void Stop()
         {
+            cancelPendingStart();
```
peppy, line 75: note that this isn't strictly required but feels good as a safety measure.

smoogipoo (approved):

PR #10700, opened 2020-11-05
Add support for spinner "spin" text in legacy skins

peppy:
- [x] Depends on #10697 for ease of merging.

To make this work, I've adjusted judgement logic to provide correct `TimeOffset` when hitwindows are empty. Let me know if this should be split out into its own PR (exists as a distinct commit for now).

Note that I've added a temporary offset to allow this to coexist with the SPM counter (which is not implemented in legacy skins yet):

With spins:

Without spins:

This is a screenshot showing the alignment without the offset:

image

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -546,7 +546,11 @@ namespace osu.Game.Rulesets.Objects.Drawables
             // Ensure that the judgement is given a valid time offset, because this may not get set by the caller
             var endTime = HitObject.GetEndTime();
 
-            Result.TimeOffset = Math.Min(HitObject.HitWindows.WindowFor(HitResult.Miss), Time.Current - endTime);
+            Result.TimeOffset = Time.Current - endTime;
+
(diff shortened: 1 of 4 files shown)


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -546,7 +546,11 @@ protected void ApplyResult(Action<JudgementResult> application)
             // Ensure that the judgement is given a valid time offset, because this may not get set by the caller
             var endTime = HitObject.GetEndTime();
 
-            Result.TimeOffset = Math.Min(HitObject.HitWindows.WindowFor(HitResult.Miss), Time.Current - endTime);
+            Result.TimeOffset = Time.Current - endTime;
+
+            double missWindow = HitObject.HitWindows.WindowFor(HitResult.Miss);
+            if (missWindow > 0)
+                Result.TimeOffset = Math.Min(Result.TimeOffset, missWindow);
```
bdach, line 553: A little uneasy about this change, it feels like it could break a few places - although I've checked a few off the top of my head and couldn't find anything so far, so maybe fine to go ahead with it.

smoogipoo (approved):

PR #10712, opened 2020-11-06
Move cache and lookup logic to base MemoryCachingComponent class

peppy:
Should simplify things in the corresponding two usages.

Wasn't sure if we should add a synchronous lookup pathway right now (`BeatmapDifficultyCache` uses this in a few places, but by keeping it local we can guarantee `GetAsync.Result` safety (unless the underlying cache implementation changes in the future)).

- [x] Depends on https://github.com/ppy/osu/pull/10711

### DIFF
--- a/osu.Game/Beatmaps/BeatmapDifficultyCache.cs
+++ b/osu.Game/Beatmaps/BeatmapDifficultyCache.cs
@@ -273,7 +273,8 @@ namespace osu.Game.Beatmaps
             public DifficultyCacheLookup([NotNull] BeatmapInfo beatmap, [NotNull] RulesetInfo ruleset, IEnumerable<Mod> mods)
             {
                 Beatmap = beatmap;
-                Ruleset = ruleset;
+                // In the case that the user hasn't given us a ruleset, use the beatmap's default ruleset.
+                Ruleset = ruleset ?? Beatmap.Ruleset;
                 OrderedMods = mods?.OrderBy(m => m.Acronym).ToArray() ?? Array.Empty<Mod>();
             }
 
--- a/osu.Game/Beatmaps/BeatmapDifficultyCache.cs
+++ b/osu.Game/Beatmaps/BeatmapDifficultyCache.cs
@@ -86,20 +86,30 @@ namespace osu.Game.Beatmaps
         /// <param name="mods">The <see cref="Mod"/>s to get the difficulty with.</param>
         /// <param name="cancellationToken">An optional <see cref="CancellationToken"/> which stops computing the star difficulty.</param>
         /// <returns>The <see cref="StarDifficulty"/>.</returns>
-        public async Task<StarDifficulty> GetDifficultyAsync([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo rulesetInfo = null, [CanBeNull] IEnumerable<Mod> mods = null,
-                                                             CancellationToken cancellationToken = default)
+        public Task<StarDifficulty> GetDifficultyAsync([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo rulesetInfo = null, [CanBeNull] IEnumerable<Mod> mods = null, CancellationToken cancellationToken = default)
         {
-            if (tryGetExisting(beatmapInfo, rulesetInfo, mods, out var existing, out var key))
-                return existing;
+            // In the case that the user hasn't given us a ruleset, use the beatmap's default ruleset.
+            rulesetInfo ??= beatmapInfo.Ruleset;
 
-            return await Task.Factory.StartNew(() =>
+            // Difficulty can only be computed if the beatmap and ruleset are locally available.
+            if (beatmapInfo.ID == 0 || rulesetInfo.ID == null)
             {
-                // Computation may have finished in a previous task.
-                if (tryGetExisting(beatmapInfo, rulesetInfo, mods, out existing, out _))
+                // If not, fall back to the existing star difficulty (e.g. from an online source).
+                return Task.FromResult(new StarDifficulty(beatmapInfo.StarDifficulty, beatmapInfo.MaxCombo ?? 0));
+            }
+
+            return GetAsync(new DifficultyCacheLookup(beatmapInfo, rulesetInfo, mods), cancellationToken);
+        }
+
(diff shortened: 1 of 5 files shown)


osu.Game/Database/MemoryCachingComponent.cs
```diff
@@ -12,6 +15,37 @@ namespace osu.Game.Database
...
+        protected virtual bool CacheNullValues => true;
+
+        /// <summary>
+        /// Retrieve the cached value for the given lookup.
+        /// </summary>
+        /// <param name="lookup">The lookup to retrieve.</param>
+        /// <param name="token">An optional <see cref="CancellationToken"/> to cancel the operation.</param>
+        protected async Task<TValue> GetAsync([NotNull] TLookup lookup, CancellationToken token = default)
+        {
+            if (cache.TryGetValue(lookup, out TValue performance))
```
smoogipoo, line 29: Maybe use `CheckExists()`?

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
```diff
@@ -302,29 +265,31 @@ protected override void Dispose(bool isDisposing)
 
         public readonly struct DifficultyCacheLookup : IEquatable<DifficultyCacheLookup>
         {
-            public readonly int BeatmapId;
-            public readonly int RulesetId;
-            public readonly Mod[] Mods;
+            public readonly BeatmapInfo Beatmap;
```
smoogipoo, line 268: I specifically didn't do this because I'm not sure we want to be storing a permanent game-wide reference to the `BeatmapInfo`. The same in a way with `RulesetInfo`, though I'm less concerned there.

It was already done this way (https://github.com/ppy/osu/commit/00e6217f60c9d1981e1eb16e1b21b39a70b844a9) before, and I don't remember if I actually noticed an issue before changing it to ints. Can this be changed back to storing the raw IDs, just to prevent unintentional regressions in an otherwise refactoring PR?

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
```diff
@@ -249,49 +244,17 @@ private StarDifficulty computeDifficulty(in DifficultyCacheLookup key, BeatmapIn
...
-                    return Cache[key] = new StarDifficulty();
+                    return new StarDifficulty();
                 }
 
-                // Check the cache first because this is now a different ruleset than the one previously guarded against.
-                if (tryGetExisting(beatmapInfo, beatmapInfo.Ruleset, Array.Empty<Mod>(), out var existingDefault, out var existingDefaultKey))
-                    return existingDefault;
-
-                return computeDifficulty(existingDefaultKey, beatmapInfo, beatmapInfo.Ruleset);
+                return GetAsync(new DifficultyCacheLookup(key.Beatmap, key.Beatmap.Ruleset, key.OrderedMods)).Result;
```
smoogipoo, line 250: Is this safe? We're already on the `updateScheduler` at this point, which has only one thread available to it. We can't be firing another async task.

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
```diff
@@ -86,35 +86,30 @@ public IBindable<StarDifficulty> GetBindableDifficulty([NotNull] BeatmapInfo bea
...
         /// <param name="cancellationToken">An optional <see cref="CancellationToken"/> which stops computing the star difficulty.</param>
         /// <returns>The <see cref="StarDifficulty"/>.</returns>
-        public async Task<StarDifficulty> GetDifficultyAsync([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo rulesetInfo = null, [CanBeNull] IEnumerable<Mod> mods = null,
-                                                             CancellationToken cancellationToken = default)
+        public Task<StarDifficulty> GetDifficultyAsync([NotNull] BeatmapInfo beatmapInfo, [CanBeNull] RulesetInfo rulesetInfo = null, [CanBeNull] IEnumerable<Mod> mods = null, CancellationToken cancellationToken = default)
         {
-            if (tryGetExisting(beatmapInfo, rulesetInfo, mods, out var existing, out var key))
-                return existing;
+            // In the case that the user hasn't given us a ruleset, use the beatmap's default ruleset.
+            rulesetInfo ??= beatmapInfo.Ruleset;
```
smoogipoo, line 92: This is only done here now? What about the bindable paths?

smoogipoo (changes requested):

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
peppy: I considered this, but I think it's fine. They are model types and shouldn't be an issue. Although I'm not sure how this will play in with realm.

Do you have an alternative? I added these in to provide a way to perform the construction/computation when required.

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
peppy: most scheduler implementations recognise this and don't spawn a new thread if already on the correct thread/pool.

image

It's implemented correctly in `ThreadedTaskScheduler` so shouldn't be an issue. Worth confirming this though.

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
peppy: Right. I had it in the base method before but thought it was safe after removing the synchronous flow. Will move back.

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
peppy: As mentioned in discord, one way around this would be to add an extra parameter to GetAsync which pipes through the `TInstance` (or `object` if we don't want more generics) – is this a direction you'd be happy with?

smoogipoo: I think I'm just going to cautiously accept this as it is...

smoogipoo (approved):

osu.Game/Beatmaps/BeatmapDifficultyCache.cs
peppy: Shouldn't be a huge deal if we find out this is too much of an overhead, at least.

PR #10713, opened 2020-11-06
Add user caching / lookup component and consume in the currently playing display

peppy:
- [x] Depends on https://github.com/ppy/osu/pull/10712

Tried to make the logic is understandable as possible, hopefully it makes sense. A bit of a hard scenario to make work elegantly.

### DIFF
--- a/osu.Game/Database/UserLookupCache.cs
+++ b/osu.Game/Database/UserLookupCache.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
@@ -16,7 +15,7 @@ namespace osu.Game.Database
 {
     public class UserLookupCache : MemoryCachingComponent<int, User>
     {
-        private readonly ConcurrentBag<int> nextTaskIDs = new ConcurrentBag<int>();
+        private readonly HashSet<int> nextTaskIDs = new HashSet<int>();
 
         [Resolved]
         private IAPIProvider api { get; set; }
--- a/osu.Game/Database/UserLookupCache.cs
+++ b/osu.Game/Database/UserLookupCache.cs
@@ -30,7 +30,7 @@ namespace osu.Game.Database
         /// </summary>
         private bool pendingRequestConsumedIDs;
 
-        public Task<User> GetUser(int userId, CancellationToken token = default) => GetAsync(userId, token);
+        public Task<User> GetUserAsync(int userId, CancellationToken token = default) => GetAsync(userId, token);
 
         protected override async Task<User> ComputeValueAsync(int lookup, CancellationToken token = default)
         {
--- a/osu.Game/Database/UserLookupCache.cs
+++ b/osu.Game/Database/UserLookupCache.cs
@@ -1,0 +1,118 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using osu.Framework.Allocation;
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests;
+using osu.Game.Users;
+
+namespace osu.Game.Database
+{
+    public class UserLookupCache : MemoryCachingComponent<int, User>
+    {
+        private readonly ConcurrentBag<int> nextTaskIDs = new ConcurrentBag<int>();
+
+        [Resolved]
+        private IAPIProvider api { get; set; }
+
+        private readonly object taskAssignmentLock = new object();
+
+        private Task<List<User>> pendingRequest;
+
+        /// <summary>
+        /// Whether <see cref="pendingRequest"/> has already grabbed its IDs.
+        /// </summary>
+        private bool pendingRequestConsumedIDs;
+
+        public Task<User> GetUser(int userId, CancellationToken token = default) => GetAsync(userId, token);
+
+        protected override async Task<User> ComputeValueAsync(int lookup, CancellationToken token = default)
+        {
+            var users = await getQueryTaskForUser(lookup);
+            return users.FirstOrDefault(u => u.Id == lookup);
+        }
+
+        /// <summary>
+        /// Return the task responsible for fetching the provided user.
+        /// This may be part of a larger batch lookup to reduce web requests.
+        /// </summary>
+        /// <param name="userId">The user to lookup.</param>
+        /// <returns>The task responsible for the lookup.</returns>
+        private Task<List<User>> getQueryTaskForUser(int userId)
+        {
+            lock (taskAssignmentLock)
+            {
+                nextTaskIDs.Add(userId);
+
+                // if there's a pending request which hasn't been started yet (and is not yet full), we can wait on it.
+                if (pendingRequest != null && !pendingRequestConsumedIDs && nextTaskIDs.Count < 50)
+                    return pendingRequest;
+
+                return queueNextTask(nextLookup);
+            }
+
+            List<User> nextLookup()
+            {
+                int[] lookupItems;
+
+                lock (taskAssignmentLock)
+                {
+                    pendingRequestConsumedIDs = true;
+                    lookupItems = nextTaskIDs.ToArray();
+                    nextTaskIDs.Clear();
+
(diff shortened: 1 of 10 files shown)


osu.Game/Database/UserLookupCache.cs
```diff
@@ -0,0 +1,118 @@
...
+using osu.Framework.Allocation;
+using osu.Game.Online.API;
+using osu.Game.Online.API.Requests;
+using osu.Game.Users;
+
+namespace osu.Game.Database
+{
+    public class UserLookupCache : MemoryCachingComponent<int, User>
+    {
+        private readonly ConcurrentBag<int> nextTaskIDs = new ConcurrentBag<int>();
```
bdach, line 19: I'm a bit confused as to why this is a concurrent collection, as all accesses to this happen only if `taskAssignmentLock` is taken...

osu.Game/Database/UserLookupCache.cs
```diff
@@ -0,0 +1,118 @@
...
+        /// Return the task responsible for fetching the provided user.
+        /// This may be part of a larger batch lookup to reduce web requests.
+        /// </summary>
+        /// <param name="userId">The user to lookup.</param>
+        /// <returns>The task responsible for the lookup.</returns>
+        private Task<List<User>> getQueryTaskForUser(int userId)
+        {
+            lock (taskAssignmentLock)
+            {
+                nextTaskIDs.Add(userId);
```
bdach, line 51: As a side note [the bag is not a set](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentbag-1?view=netcore-3.1#remarks), so duplicates can potentially get added in here.

osu.Game/Database/UserLookupCache.cs
```diff
@@ -0,0 +1,118 @@
...
+
+        /// <summary>
+        /// Queues new work at the end of the current work tasks.
+        /// Ensures the provided work is eventually run.
+        /// </summary>
+        /// <param name="work">The work to run. Can be null to signify the end of available work.</param>
+        /// <returns>The task tracking this work.</returns>
+        private Task<List<User>> queueNextTask(Func<List<User>> work)
+        {
+            lock (taskAssignmentLock)
```
bdach, line 94: Just to make sure, this lock is taken locally here (even if the caller has already taken it once) to avoid continuations executing this code without taking it, right?

osu.Game/Online/API/Requests/GetUsersRequest.cs
```diff
@@ -1,10 +1,25 @@
...
+
+        public GetUsersRequest(int[] userIds)
+        {
+            if (userIds.Length > max_ids_per_request)
+                throw new ArgumentException($"{nameof(GetUsersRequest)} calls only support up to {max_ids_per_request} IDs at once");
+
+            this.userIds = userIds;
+        }
+
+        protected override string Target => $@"users/?{userIds.Select(u => $"ids[]={u}&").Aggregate((a, b) => a + b)}";
```
bdach, line 23: Maybe not that huge of a deal but I'd probably write this as follows:

```suggestion
        protected override string Target => "users/?ids[]=" + string.Join("&ids[]=", userIds);
```

as `string.Join` uses a `StringBuilder` internally, avoiding the quadratic overhead of plain concatenation.

osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs
```diff
@@ -54,15 +58,20 @@ protected override void LoadComplete()
...
-                                if (playingUsers.Contains(user.Id))
-                                    userFlow.Add(createUserPanel(user));
+                                if (u.Result == null)
+                                    return;
+
+                                Schedule(() =>
+                                {
+                                    if (playingUsers.Contains(u.Result.Id))
+                                        userFlow.Add(createUserPanel(u.Result));
+                                });
```
bdach, line 73: I'd probably split a method somewhere here - the nesting is nearing on too much I'd say.

osu.Game/Database/UserLookupCache.cs
```diff
@@ -0,0 +1,118 @@
...
+        private readonly object taskAssignmentLock = new object();
+
+        private Task<List<User>> pendingRequest;
+
+        /// <summary>
+        /// Whether <see cref="pendingRequest"/> has already grabbed its IDs.
+        /// </summary>
+        private bool pendingRequestConsumedIDs;
+
+        public Task<User> GetUser(int userId, CancellationToken token = default) => GetAsync(userId, token);
```
bdach, line 33: I think our general convention is to have an `Async` suffix on async methods...?

osu.Game/Database/UserLookupCache.cs
peppy: Yeah I got this one confused.

osu.Game/Database/UserLookupCache.cs
peppy: Aware of that, left it due to the better API of the class (and because it's such a tiny use case that wouldn't affect anything). But based on it containing duplicates I'll change away.

osu.Game/Database/UserLookupCache.cs
peppy: It's just safety and readability yeah. Can be removed if that isn't seen as an improvement.

smoogipoo:
Crashes with:
```
Unhandled exception. System.AggregateException: A Task's exception(s) were not observed either by Waiting on the Task or accessing its Exception property. As a result, the unobserved exception was rethrown by the finalizer thread. (One or more errors occurred. (Value cannot be null. (Parameter 'source')))
 ---> System.AggregateException: One or more errors occurred. (Value cannot be null. (Parameter 'source'))
 ---> System.ArgumentNullException: Value cannot be null. (Parameter 'source')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at osu.Game.Database.UserLookupCache.ComputeValueAsync(Int32 lookup, CancellationToken token) in /home/smgi/Repos/osu/osu.Game/Database/UserLookupCache.cs:line 37
   at osu.Game.Database.MemoryCachingComponent`2.GetAsync(TLookup lookup, CancellationToken token) in /home/smgi/Repos/osu/osu.Game/Database/MemoryCachingComponent.cs:line 32
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at osu.Game.Overlays.Dashboard.CurrentlyPlayingDisplay.<>c__DisplayClass16_0.<onUsersChanged>b__1(Task`1 u) in /home/smgi/Repos/osu/osu.Game/Overlays/Dashboard/CurrentlyPlayingDisplay.cs:line 68
   at System.Threading.Tasks.ContinuationTaskFromResultTask`1.InnerInvoke()
```

Probably from:
```
[network] 2020-11-09 05:10:59 [verbose]: Processing response from https://osu.ppy.sh/api/v2/users/?ids[]=16496575&ids[]=8859767&ids[]=11536550&ids[]=5170001 failed with Newtonsoft.Json.JsonSerializationException: Cannot deserialize the current JSON array (e.g. [1,2,3]) into type 'osu.Game.Online.API.Requests.GetUsersResponse' because the type requires a JSON object (e.g. {"name":"value"}) to deserialize correctly.
[network] 2020-11-09 05:10:59 [verbose]: To fix this error either change the JSON to a JSON object (e.g. {"name":"value"}) or change the deserialized type to an array or a type that implements a collection interface (e.g. ICollection, IList) like List<T> that can be deserialized from a JSON array. JsonArrayAttribute can also be added to the type to force it to deserialize from a JSON array.
[network] 2020-11-09 05:10:59 [verbose]: Path '', line 1, position 1.
```
Seems to be a 404.

osu.Game/Database/UserLookupCache.cs
bdach: I don't mind it myself, was just making sure. Monitors are re-entrant anyway.

peppy:
This was due to a rollback of osu-web (fixing other issues). Am deploying again at the moment so give it another try in 5-10 minutes.

peppy:
@smoogipoo if you're going to test please do so within the next 5 minutes (rolling back osu-web again because fixed thing was not fixed).

smoogipoo:
It should still have a failsafe though. What happens if the request times out for example?

peppy:
You have a point. Are you in a state to add that or shall I? probably just try-catch.

smoogipoo (approved):

PR #10717, opened 2020-11-06
Implement HitObject application for DrawableHitObject

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu-framework/pull/3989

Full documentation will come later but the general process is:

* (sealed) DHO.Apply()
    * Invoked by the pooling implementation later on.
    * Invoked by rulesets for testing purposes (see the new tests as examples).
* (virtual) DHO.OnApply()
    * Invoked for consumers to implement.
    * For consumers to take on values from `HitObject`s.
    * In the osu! ruleset this is always going to be binding to hitobject bindables.
* (virtual) DHO.OnFree()
    * Invoked for consumers to implement.
    * For consumers to revert any values they previously took on in `OnApply()`.
    * In the osu! ruleset this is always going to be unbinding to the hitobject.

All rulesets can continue existing as they are currently made, however this will need to be done if they want to take advantage of pooling.

Unfortunately, all rulesets will break until recompile due to `DHO.HitObject` now becoming a property instead of a readonly field. I don't know how to improve this beyond exposing the new property under a different name (`AppliedObject`?) and deprecating the field.

Also I'm taking suggestions on naming - in my original changes I just had `Apply()`/`FreeAfterUse()` virtual, but at least `FreeAfterUse()` can't be virtual to deal with situations like catch which removes then adds hitobjects in other containers (triggering a `FreeAfterUse()` when not expected). I also like separating it out to prevent misuse of the DHO methods, but let me know if you agree.  
A potential alternative is to expose a property like `FreeOnRemovalWhenNotInPool` from `PoolableDrawable`.

# Breaking changes

## All rulesets need to be recompiled

`DrawableHitObject.HitObject` is now exposed as a property instead of a field, causing all rulesets to crash unless recompiled.

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -197,11 +197,12 @@ namespace osu.Game.Rulesets.Objects.Drawables
 
             HitObject.DefaultsApplied += onDefaultsApplied;
 
+            OnApply(hitObject);
+
             // If not loaded, the state update happens in LoadComplete(). Otherwise, the update is scheduled to allow for lifetime updates.
             if (IsLoaded)
                 Schedule(() => updateState(ArmedState.Idle, true));
 
-            OnApply(hitObject);
             hasHitObjectApplied = true;
         }
 
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -116,6 +116,11 @@ namespace osu.Game.Rulesets.Objects.Drawables
         public IBindable<ArmedState> State => state;
 
         /// <summary>
+        /// Whether <see cref="HitObject"/> is currently applied.
+        /// </summary>
+        private bool hasHitObjectApplied;
+
+        /// <summary>
         /// Creates a new <see cref="DrawableHitObject"/>.
         /// </summary>
         /// <param name="initialHitObject">
@@ -152,49 +157,15 @@ namespace osu.Game.Rulesets.Objects.Drawables
         }
 
         /// <summary>
-        /// Removes the <see cref="HitObject"/> currently applied to this <see cref="DrawableHitObject"/>,
-        /// </summary>
-        protected override void FreeAfterUse()
-        {
-            StartTimeBindable.UnbindFrom(HitObject.StartTimeBindable);
-            if (HitObject is IHasComboInformation combo)
-                comboIndexBindable.UnbindFrom(combo.ComboIndexBindable);
-
-            samplesBindable.UnbindFrom(HitObject.SamplesBindable);
-
-            // When a new hitobject is applied, the samples will be cleared before re-populating.
-            // In order to stop this needless update, the event is unbound and re-bound as late as possible in Apply().
-            samplesBindable.CollectionChanged -= onSamplesChanged;
-
-            if (nestedHitObjects.IsValueCreated)
-            {
-                foreach (var obj in nestedHitObjects.Value)
-                {
-                    obj.OnNewResult -= onNewResult;
-                    obj.OnRevertResult -= onRevertResult;
-                    obj.ApplyCustomUpdateState -= onApplyCustomUpdateState;
-                }
-
-                nestedHitObjects.Value.Clear();
-                ClearNestedHitObjects();
-            }
-
-            HitObject.DefaultsApplied -= onDefaultsApplied;
-            HitObject = null;
-
-            base.FreeAfterUse();
-        }
-
-        /// <summary>
         /// Applies a new <see cref="HitObject"/> to be represented by this <see cref="DrawableHitObject"/>.
         /// </summary>
-        /// <param name="hitObject"></param>
+        /// <param name="hitObject">The <see cref="HitObject"/> to apply.</param>
         public virtual void Apply(HitObject hitObject)
         {
-            if (HitObject != null)
-                FreeAfterUse();
+            if (hasHitObjectApplied)
+                Free();
 
-            HitObject = hitObject;
+            HitObject = hitObject ?? throw new InvalidOperationException($"Cannot apply a null {nameof(HitObject)}.");
 
             // Copy any existing result from the hitobject (required for rewind / judgement revert).
             Result = HitObject.Result;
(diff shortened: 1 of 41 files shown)


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -96,10 +100,10 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
         /// </remarks>
         protected virtual float SamplePlaybackPosition => 0.5f;
 
-        private BindableList<HitSampleInfo> samplesBindable;
-        private Bindable<double> startTimeBindable;
-        private Bindable<bool> userPositionalHitSounds;
-        private Bindable<int> comboIndexBindable;
+        public readonly Bindable<double> StartTimeBindable = new Bindable<double>();
```
bdach, line 103: This `public` doesn't have an explanation for now but I can see further changes will be using this field, so I'm purposefully ignoring this.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -111,50 +115,160 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
...
+            // Ensure the hitobject has a result.
+            HitObject.Result = Result;
+
+            foreach (var h in HitObject.NestedHitObjects)
+            {
+                var drawableNested = CreateNestedHitObject(h) ?? throw new InvalidOperationException($"{nameof(CreateNestedHitObject)} returned null for {h.GetType().ReadableName()}.");
+
+                drawableNested.OnNewResult += (d, r) => OnNewResult?.Invoke(d, r);
+                drawableNested.OnRevertResult += (d, r) => OnRevertResult?.Invoke(d, r);
+                drawableNested.ApplyCustomUpdateState += (d, j) => ApplyCustomUpdateState?.Invoke(d, j);
```
bdach, line 185: Shouldn't this be

```suggestion
                drawableNested.OnNewResult += onNewResult;
                drawableNested.OnRevertResult += onRevertResult;
                drawableNested.ApplyCustomUpdateState += onApplyCustomUpdateState;
```

It reads nicer for one, but secondly I think that the unsubscribes in `free()` won't have any effect as written as the delegates are not the same (at least that's what I'm seeing in the debugger).

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -111,50 +115,160 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
...
+            samplesBindable.BindTo(HitObject.SamplesBindable);
+            samplesBindable.BindCollectionChanged(onSamplesChanged, true);
 
             HitObject.DefaultsApplied += onDefaultsApplied;
 
-            startTimeBindable = HitObject.StartTimeBindable.GetBoundCopy();
-            startTimeBindable.BindValueChanged(_ => updateState(State.Value, true));
+            // If not loaded, the state update happens in LoadComplete(). Otherwise, the update is scheduled to allow for lifetime updates.
+            if (IsLoaded)
+                Schedule(() => updateState(ArmedState.Idle, true));
```
bdach, line 202: Maybe it should be ensured that this doesn't get called more than once? Not sure how realistic but I can see a scenario wherein this could get triggered multiple times but I don't think it ever really needs to.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: Good catch.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: Ah, this was due to a change I had in this PR (refactor of HOC) that was temporarily reverted. I'll change back to private.

In the full hitobject-pooling implementation I had all of these public just to conform to, e.g. all `DrawableOsuHitObject` ones being public (and required to be public).

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
smoogipoo: Would that ever be a problem? The rest of the code here is still going to run 1000s of times if it was called that many times within a single frame, of which sample reconstruction is by far the most expensive.

We decided a while back with the editor to keep the amount of debouncing to a minimum, and I don't think it's worth adding it here.

bdach: Fair point, was just asking. Don't recall being aware of such a decision being made.

osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -168,6 +169,11 @@ protected virtual void CreateNestedHitObjects()
         /// </summary>
         [NotNull]
         protected virtual HitWindows CreateHitWindows() => new HitWindows();
+
+        /// <summary>
+        /// The result this <see cref="HitObject"/> was judged with. Used internally for rewinding within <see cref="DrawableHitObject"/>.
+        /// </summary>
+        internal JudgementResult Result;
```
peppy, line 176: Hmm, sure about this one? This is the first time we are storing gameplay-related state inside the hitobject model.

bdach: I mean the result has to get stored *somewhere* if the drawable objects are to be recycled, and this seems to be the immediate and simplest solution.

peppy: I get that, but it also breaks the pattern we have been going for until now (`HitObject`s are never mutable by gameplay).

smoogipoo: I'll try to move it into the lifetime entries in my pooling implementation again. I tried that originally but not sure why it didn't work out.

peppy: Honestly not sure it's so bad where it is, but would definitely help for peace of mind. I can only imagine we'll run into issues down the track (editor, serialisation, etc.) if it's in `HitObject`.

Also doesn't set a great precedent (even though it's `internal`).

smoogipoo: Right, I remember what the issue is. Storing it inside the hitobject lifetime entries works for pooling top-level hitobjects, but doesn't work for pooling nested hitobjects. The alternative to this, I think, is to store a `Dictionary<HitObject, HitObjectLifetimeEntry>` in `DrawableRuleset`, and then either retrieve or construct dummy entries on-the-fly as drawable representations are requested.

I do agree that having this stored in `HitObject` doesn't set a good precedent.

smoogipoo: Tested and my idea does seem to work: https://github.com/smoogipoo/osu/commit/a6174a0e3199f6a0306aaddb2200fadedcc97de5

I've removed the storage in hitobject from this PR.

peppy (approved):

PR #10722, opened 2020-11-06
Change VSCode build task args to use - instead of /

### LINKED ISSUE #10721: VSCode Build Tasks don't work in git bash
**Describe the bug:** 
The build tasks in tasks.json don't work when using git bash (that comes with git for Windows) as a terminal because the arguments use / instead of -

**osu!lazer version:** 
Latest master.

**Logs:**
```
> Executing task: dotnet build --no-restore osu.Desktop /p:Configuration=Release /p:GenerateFullPaths=true /m /verbosity:m <

Microsoft (R) Build Engine version 16.4.0+e901037fe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.        

MSBUILD : error MSB1008: Only one project can be specified.
Switch: p:Configuration=Release

For switch syntax, type "MSBuild -help"
The terminal process "C:\Program Files\Git\bin\bash.exe '-c', 'dotnet build --no-restore osu.Desktop /p:Configuration=Release /p:GenerateFullPaths=true /m /verbosity:m'" terminated with exit code: 1.
```

nbvdkamp:
Resolves #10721 

I've tested this on Windows in cmd, powershell and git bash and on linux in regular bash.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -11,9 +11,9 @@
                 "build",
                 "--no-restore",
                 "osu.Desktop",
-                "/p:GenerateFullPaths=true",
-                "/m",
-                "/verbosity:m"
+                "-p:GenerateFullPaths=true",
+                "-m",
+                "-verbosity:m"
             ],
             "group": "build",
             "problemMatcher": "$msCompile"
@@ -26,10 +26,10 @@
                 "build",
                 "--no-restore",
                 "osu.Desktop",
-                "/p:Configuration=Release",
-                "/p:GenerateFullPaths=true",
-                "/m",
-                "/verbosity:m"
+                "-p:Configuration=Release",
+                "-p:GenerateFullPaths=true",
+                "-m",
+                "-verbosity:m"
             ],
             "group": "build",
             "problemMatcher": "$msCompile"
@@ -42,9 +42,9 @@
                 "build",
                 "--no-restore",
                 "osu.Game.Tests",
-                "/p:GenerateFullPaths=true",
-                "/m",
-                "/verbosity:m"
+                "-p:GenerateFullPaths=true",
+                "-m",
+                "-verbosity:m"
             ],
             "group": "build",
             "problemMatcher": "$msCompile"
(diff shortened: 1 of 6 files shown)


Joehuu:
Needs to be done on other `tasks.json`s in this repo (e.g. https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu.Tests/.vscode/tasks.json).

peppy:
Not 100% on this solution, but here's more information on the issue: https://stackoverflow.com/a/51171516

Provides two alternatives we may consider.

nbvdkamp:
Using two slashes works in git bash but gives
```
MSBUILD : error MSB1001: Unknown switch.
Switch: //p:GenerateFullPaths=true

For switch syntax, type "MSBuild -help"
The terminal process "C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -Command dotnet build --no-restore osu.Desktop //p:GenerateFullPaths=true //m //verbosity:m" terminated with exit code: 1.
```
in powershell. There's no spaces or quotes in any of the arguments so I'm not sure what other alternative you mean.

peppy:
One option is to fix this locally (see https://github.com/git-for-windows/build-extra/blob/a9090580ee37686a53166796dffda531041e16f8/ReleaseNotes.md)

I'm not sure we should be working around this behaviour locally. If we are to do so, this needs to also be applied on about 10 other repos we have. I'm not sure enough people using git-bash as their terminal in vscode to warrant this.

bdach:
Given that even [Microsoft docs](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?view=vs-2019) specify the posix-like `-p` switch syntax instead of the traditional windows-style `/p` one, I think I'd actually be in favour of this version as it looks to be more portable overall.

peppy:
If it's preferred then I'm definitely fine with swapping, just want to check it's the correct direction first (and apply to other projects).

peppy (approved):

PR #10729, opened 2020-11-07
Fix crash on undo in beatmap without objects

### LINKED ISSUE #10728: Crash when undoing change in a beatmap without objects
**Describe the crash:** 
A crash occurs when undoing a change in a beatmap with no hitobjects (adding a timing point and then pressing ctrl-z for example).

**osu!lazer version:** 
Latest master.

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5504819/runtime.log)

nbvdkamp:
Resolves #10728 

I am running into an odd stack overflow that I'm not 100% on the cause of but I've also gotten it to happen on master once. It has to do with undoing and redoing a few times but is probably unrelated.

### DIFF
--- a/osu.Game/Screens/Edit/LegacyEditorBeatmapPatcher.cs
+++ b/osu.Game/Screens/Edit/LegacyEditorBeatmapPatcher.cs
@@ -41,7 +41,8 @@ namespace osu.Game.Screens.Edit
             foreach (var block in result.DiffBlocks)
             {
                 // Removed hitobjects
-                if (oldHitObjectsIndex != -1)
+                if (oldHitObjectsIndex > 0)
+                {
                     for (int i = 0; i < block.DeleteCountA; i++)
                     {
                         int hoIndex = block.DeleteStartA + i - oldHitObjectsIndex - 1;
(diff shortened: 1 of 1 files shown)


bdach:
In the future if you're going to submit a fix for an issue you've found, just skip the issue and open the PR. It saves us having to guess whether you will fix the issue yourself or not.

nbvdkamp:
Will do, I thought the preferred procedure was opening issues for everything.

bdach:
Obviously not in the case wherein you're capable enough to fix them. That's just needless bureaucracy at this point.

osu.Game/Screens/Edit/LegacyEditorBeatmapPatcher.cs
```diff
@@ -41,26 +41,28 @@ public void Patch(byte[] currentState, byte[] newState)
             foreach (var block in result.DiffBlocks)
             {
                 // Removed hitobjects
-                for (int i = 0; i < block.DeleteCountA; i++)
-                {
-                    int hoIndex = block.DeleteStartA + i - oldHitObjectsIndex - 1;
+                if (oldHitObjectsIndex != -1)
```
bdach, line 44: Really should be `> 0`.

Also please add braces for this and the other statement. CI/code style analysis will complain about this 100%, and having a one-line condition followed by a multi-line loop is not readable.

bdach (changes requested):

bdach:
As a side note the history of this PR is really weird - some changes were added but then essentially reverted later. They can stay now to avoid force-pushing I suppose, but in the future please squash before PRing if such a case pops up.

peppy:
Should be >= 0 for the comparisons, but I've already got a PR which fixes this in a more amicable way. See #10730.

bdach:
You're right. Trying to go a little too fast to catch up today...

PR #10730, opened 2020-11-07
Fix legacy patcher crash on no objects present

### LINKED ISSUE #10728: Crash when undoing change in a beatmap without objects

peppy:
Fixed this by changing the way we write files, as I believe this is more expected behaviour. Note that the test doesn't actually cover the behaviour as it is quite hard to test with the existing test setup (which is guaranteeing pre and post-patch content is the same, but the failure relies on non-hitobject content changing, which isn't covered by the patcher right now).

- Closes #10728 
- Is alternative to and closes #10729

### DIFF
--- a/osu.Game.Tests/Editing/LegacyEditorBeatmapPatcherTest.cs
+++ b/osu.Game.Tests/Editing/LegacyEditorBeatmapPatcherTest.cs
@@ -38,6 +38,12 @@ namespace osu.Game.Tests.Editing
         }
 
         [Test]
+        public void TestPatchNoObjectChanges()
+        {
+            runTest(new OsuBeatmap());
+        }
+
+        [Test]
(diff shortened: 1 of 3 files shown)


bdach (commented):
looks fine to me, but guessing from the above that my approval is not really all that relevant here

peppy:
Yeah, want to make sure @smoogipoo is okay with the encoder change (not sure if there was a reason it was implemented as it is).

smoogipoo (approved):

PR #10735, opened 2020-11-07
Fix difficulty retrieval for online-sourced beatmaps

### LINKED ISSUE #10725: Online difficulty icon always Easy
**Describe the bug:** 
The online beatmap information(e.g. multiplayer lobby, beatmap details page) always displays as `Easy ` icon. However local beatmap list is fine
osu_2020-11-07_16-51-53
osu_2020-11-07_16-53-33
osu_2020-11-07_17-05-29
**osu!lazer version:** 
https://github.com/ppy/osu/compare/master...PercyDan54:custom
**Logs:**
[database.log](https://github.com/ppy/osu/files/5504508/database.log)
[network.log](https://github.com/ppy/osu/files/5504509/network.log)
[performance.log](https://github.com/ppy/osu/files/5504510/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5504511/runtime.log)

bdach:
Resolves #10725.

This *actually* broke in https://github.com/ppy/osu/pull/10712, and I was misled into thinking this was always broken due to the very subtle distinction between `GetAsync()` and `GetDifficultyAsync()`. I have no good idea how to rename those methods better to make the distinction less subtle, however (much less so seeing that the latter is public), so there's an inline comment in there instead.


peppy (approved):
oops! good catch

PR #10736, opened 2020-11-08
Fix hard crash on saving empty beatmap twice in a row

### LINKED ISSUE #10734: Crash on saving new beatmapset that is identical to an existing one
**Describe the crash:** 
If you open the editor from the main menu (to create a new beatmap), save the beatmap without doing anything and do this again an exception is thrown because the BeatmapSetInfo hashes are identical, violating a database uniqueness constraint.

**osu!lazer version:** 
Latest master.

**Logs:**
```
Exception has occurred: CLR/Microsoft.EntityFrameworkCore.DbUpdateException
An exception of type 'Microsoft.EntityFrameworkCore.DbUpdateException' occurred in osu.Game.dll but was not handled in user code: 'An error occurred while updating the entries. See the inner exception for details.'
 Inner exceptions found, see $exception in variables window for more details.
 Innermost exception 	 Microsoft.Data.Sqlite.SqliteException : SQLite Error 19: 'UNIQUE constraint failed: BeatmapSetInfo.Hash'.
   at Microsoft.Data.Sqlite.SqliteException.ThrowExceptionForRC(Int32 rc, sqlite3 db)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteReader(CommandBehavior behavior)
   at Microsoft.Data.Sqlite.SqliteCommand.ExecuteDbDataReader(CommandBehavior behavior)
   at System.Data.Common.DbCommand.ExecuteReader()
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.Execute(IRelationalConnection connection, DbCommandMethod executeMethod, IReadOnlyDictionary`2 parameterValues)
   at Microsoft.EntityFrameworkCore.Storage.Internal.RelationalCommand.ExecuteReader(IRelationalConnection connection, IReadOnlyDictionary`2 parameterValues)
   at Microsoft.EntityFrameworkCore.Update.ReaderModificationCommandBatch.Execute(IRelationalConnection connection)
```
[runtime.log](https://github.com/ppy/osu/files/5505542/runtime.log)

peppy:
Closes https://github.com/ppy/osu/issues/10734. This obviously isn't the perfect solution (will still show database errors) but at least it avoids the hard crash experienced otherwise.

Handling this in a better way isn't to simple right now.

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -307,8 +307,17 @@ namespace osu.Game.Screens.Edit
             // apply any set-level metadata changes.
             beatmapManager.Update(playableBeatmap.BeatmapInfo.BeatmapSet);
 
-            // save the loaded beatmap's data stream.
-            beatmapManager.Save(playableBeatmap.BeatmapInfo, editorBeatmap, editorBeatmap.BeatmapSkin);
+            try
+            {
+                // save the loaded beatmap's data stream.
+                beatmapManager.Save(playableBeatmap.BeatmapInfo, editorBeatmap, editorBeatmap.BeatmapSkin);
+            }
+            catch
+            {
+                // this is a temporary measure to avoid hard crashes on saving identical beatmaps.
+                // should be replaced in the future once we aren't relying on the isNewBeatmap flag.
+                beatmapManager.Delete(playableBeatmap.BeatmapInfo.BeatmapSet);
+            }
 
             updateLastSavedHash();
         }


osu.Game/Screens/Edit/Editor.cs
```diff
@@ -307,8 +307,17 @@ protected void Save()
...
+            try
+            {
+                // save the loaded beatmap's data stream.
+                beatmapManager.Save(playableBeatmap.BeatmapInfo, editorBeatmap, editorBeatmap.BeatmapSkin);
+            }
+            catch
+            {
+                // this is a temporary measure to avoid hard crashes on saving identical beatmaps.
+                // should be replaced in the future once we aren't relying on the isNewBeatmap flag.
+                beatmapManager.Delete(playableBeatmap.BeatmapInfo.BeatmapSet);
```
bdach, line 319: This can still crash, if you try to save once, get the exception notifications, and try to save again - it'll die on an `OptimisticConcurrencyException`. I'm looking at what's causing that to get thrown right now.

bdach (changes requested):

osu.Game/Screens/Edit/Editor.cs
peppy: Honestly maybe better to just leave this one not fixed at this point. It's not like it's going to negatively affect any normal user.

smoogipoo: Are you suggesting to not merge this PR? Or to merge it and leave the followup issue pointed out above unfixed?

peppy: Maybe best to just close this for now and deal with it in a better way when we can (via having a patcher/differ that supports the whole file, not just hitobjects).

PR #10740, opened 2020-11-08
Add missing user activities for Multi screens

Game4all:
# Summary

This adds "missing" user activities for the multi screens which are already live. 

I also took this opportunity to refactor the user activity handling logic in OsuScreen which was making it basically impossible to share the current activity to the screen stack (in the case of screens being subscreens and not actual screens). The logic now uses a bindable flow, very similarly to how it's done for OverlayActivationMode.

I'm hesitant to push the one-liner change to display the room name on the discord integration as it looks weird right now (the status and details are reversed)

### DIFF
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -137,7 +137,7 @@ namespace osu.Desktop
                     return edit.Beatmap.ToString();
 
                 case UserActivity.InLobby lobby:
-                    return lobby.Room.Name.ToString();
+                    return lobby.Room.Name.Value;
             }
 
             return string.Empty;
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -135,6 +135,9 @@ namespace osu.Desktop
 
                 case UserActivity.Editing edit:
                     return edit.Beatmap.ToString();
+
+                case UserActivity.InLobby lobby:
+                    return lobby.Room.Name.ToString();
             }
 
             return string.Empty;
--- a/osu.Game/Screens/Multi/Multiplayer.cs
+++ b/osu.Game/Screens/Multi/Multiplayer.cs
@@ -24,6 +24,7 @@ using osu.Game.Screens.Multi.Lounge;
 using osu.Game.Screens.Multi.Lounge.Components;
 using osu.Game.Screens.Multi.Match;
 using osu.Game.Screens.Multi.Match.Components;
+using osu.Game.Users;
 using osuTK;
 
 namespace osu.Game.Screens.Multi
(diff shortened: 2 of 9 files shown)


osu.Game/Screens/OsuScreen.cs
```diff
@@ -135,21 +124,22 @@ private void createDependencies(IReadOnlyDependencyContainer dependencies)
...
+            Activity = new Bindable<UserActivity>();
         }
 
         [BackgroundDependencyLoader(true)]
         private void load(OsuGame osu, AudioManager audio)
         {
             sampleExit = audio.Samples.Get(@"UI/screen-back");
+
+            if (Activity.Value == null)
+                Activity.Value = InitialActivity;
```
bdach, line 142: can use `??=`

bdach (changes requested):
i guess this seems mostly fine

bdach:
> I'm hesitant to push the one-liner change to display the room name on the discord integration as it looks weird right now (the status and details are reversed)

you're going to have to give more detail than that to get a meaningful answer to that question (i.e. why does it look weird)

Game4all:
Actually, after comparing with with other rich statuses it looks normal in fact ... So I think there's no harm in getting it into that branch then ...

osu.Desktop/DiscordRichPresence.cs
```diff
@@ -135,6 +135,9 @@ private string getDetails(UserActivity activity)
 
                 case UserActivity.Editing edit:
                     return edit.Beatmap.ToString();
+
+                case UserActivity.InLobby lobby:
+                    return lobby.Room.Name.ToString();
```
bdach, line 140: weird to call `.ToString()` on a... string - the rest have `.ToString()` calls for better formatting, but this one doesn't make much sense at all

osu.Game/Screens/OsuScreen.cs
```diff
@@ -135,21 +124,21 @@ private void createDependencies(IReadOnlyDependencyContainer dependencies)
...
-        [Resolved(canBeNull: true)]
-        private IAPIProvider api { get; set; }
-
         protected OsuScreen()
         {
             Anchor = Anchor.Centre;
             Origin = Anchor.Centre;
 
             OverlayActivationMode = new Bindable<OverlayActivation>(InitialOverlayActivationMode);
+            Activity = new Bindable<UserActivity>();
```
peppy, line 133: move to field itself

osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -337,6 +338,12 @@ private void subScreenChanged(IScreen newScreen)
                     break;
             }
 
+            if (lastScreen is IOsuScreen lastOsuScreen)
```
peppy, line 341: not hugely fond of this whole flow, but seems like about the best we can do with the current structure of multiplayer.

peppy (changes requested):

bdach (approved):

PR #10742, opened 2020-11-09
Fix ki* textures having an incorrect vertical position

Morilli:
This "broke" my skin in lazer whereas it looked fine in stable.
Should now match stable in positioning.

Here is a very basic example skin that shows the issue: [skin.zip](https://github.com/ppy/osu/files/5508234/testnew.zip)
Previously, the `scorebar-ki.png` would display below the visible part of the `scorebar-colour.png` texture.
With this change, this behavior is reversed and matches stable.

### DIFF
--- a/osu.Game/Skinning/LegacyHealthDisplay.cs
+++ b/osu.Game/Skinning/LegacyHealthDisplay.cs
@@ -80,7 +80,7 @@ namespace osu.Game.Skinning
                 Math.Clamp(Clock.ElapsedFrameTime, 0, 200),
                 fill.Width, (float)Current.Value * maxFillWidth, 0, 200, Easing.OutQuint);
 
-            marker.Position = fill.Position + new Vector2(fill.DrawWidth, 0);
+            marker.Position = fill.Position + new Vector2(fill.DrawWidth, isNewStyle ? fill.DrawHeight / 2 : 0);
         }
 
         public void Flash(JudgementResult result) => marker.Flash(result);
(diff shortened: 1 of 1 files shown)


smoogipoo:
There's more to this. It breaks the osu!classic skin and every other skin I have.

Also looking at stable, it does have Y=10 (not sure where the `/ 2` in lazer's code comes from.

peppy:
I used div2 because in my testing it looked correct (the Y=10 was just an approximation/hack based off the default skin, but i guess other legacy skins rely on it).

Morilli:
How about this?
```c#
marker.Position = fill.Position + new Vector2(fill.DrawWidth, marker is LegacyNewStyleMarker ? fill.DrawHeight / 2 : 0);
```
This differentiates between the oldstylemarker (which my fix was targeted at, I didn't even take the newstyle marker into account) and the new style marker, for which the current code apparently worked fine.

Looks fine to me for the classic skin and my skin with the ki* textures, does this code look passable?

bdach:
Ideally you'd put the adjustment in the old/new style markers themselves instead of jangling it based on a pattern match like done above.

Morilli:
I can see that, but how would I access `fill`'s drawheight in those classes, especially given that it is currently being modified in the code just before the marker position gets set?

bdach:
Reading the class again I guess it'd be fine if you used the `isNewStyle` field that is already there - but I don't know if `0` is ever a correct value for parity with stable. It might be that it just happens to work with your skin.

smoogipoo:
I'll leave it up to @peppy to review and merge this one, because stable has the ~10px offset in the new skin too, although all my skins seem to be working with these changes, and he's more likely to know why.

peppy:
Not 100% sure on this but seems okay with all the skins I have in my collection. Let's go with it for now until someone finds another issue.

peppy (approved):

PR #10746, opened 2020-11-09
Update framework

peppy:


smoogipoo (dismissed):

bdach:
There's one reference that's going to have to be renamed in `OsuGameDesktop` (`DesktopGameWindow` -> `OsuTKDesktopWindow`).

peppy:
Weird, my tests completed 100% without this failing.

PR #10758, opened 2020-11-10
Replace rankHistory in JsonProperty with rank_history for consistency

nbvdkamp:
As noted here
https://github.com/ppy/osu-web/blob/4818f49af0a16ee7dd170fc805a5c2074d44b6bd/app/Transformers/UserCompactTransformer.php#L66
the JSON property name used for rank history in lazer is inconsistent in naming convention.
Since this is already supported by web it seems to work fine.

Should I also PR to web to remove the TODO or is there something else that depends on 'rankHistory' existing there?

### DIFF
--- a/osu.Game/Users/User.cs
+++ b/osu.Game/Users/User.cs
@@ -186,7 +186,7 @@ namespace osu.Game.Users
             }
         }
 
-        [JsonProperty(@"rankHistory")]
+        [JsonProperty(@"rank_history")]
         private RankHistoryData rankHistory
         {
             set => statistics.RankHistory = value;


bdach:
> is there something else that depends on 'rankHistory' existing there? 

Your guess is as good as mine. As to whether it's fine to PR to remove - not sure what the practices are web side but in game/framework we generally give a 6 month transition period, and if going by that, that hasn't anywhere near passed yet (properly named property was added Sep 2: https://github.com/ppy/osu-web/pull/6580)

peppy:
I'd probably just leave the todo as it will likely get addressed with the second dated TODO.

peppy (approved):

PR #10760, opened 2020-11-10
Fix crash and edge cases of user profile graph implementation

### LINKED ISSUE #10753: Game crash after pressing too many F9s to find a player to spectate
**Describe the crash:** 
Game crash after pressing too many F9s to find a player to spectate
I was trying to find someone worth watching, lots of them are online 20days ago or too low rank.
Because osu doesnt have a back button, I have to press lots of F9 to go back to the online players page, and the game crashes.
**osu!lazer version:** 
2020.1109.0
**Logs:**
[performance.log](https://github.com/ppy/osu/files/5512687/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5512689/runtime.log)
**Computer Specifications:** 
i7-9700k
2080

peppy:
Closes #10753 by resetting index when the graph content changes.

Also forces an immediate update to the hover display if hovered when the content changes.

### DIFF
--- a/osu.Game/Overlays/Profile/Header/Components/RankGraph.cs
+++ b/osu.Game/Overlays/Profile/Header/Components/RankGraph.cs
@@ -30,7 +30,7 @@ namespace osu.Game.Overlays.Profile.Header.Components
         private readonly OsuSpriteText placeholder;
 
         private KeyValuePair<int, int>[] ranks;
-        private int dayIndex;
+        private int hoveredIndex = -1;
         public readonly Bindable<UserStatistics> Statistics = new Bindable<UserStatistics>();
 
         public RankGraph()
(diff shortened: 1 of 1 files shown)


bdach:
Looks ok. Wondering if this might also be worth clamping for safety:

https://github.com/ppy/osu/blob/c8917d12365b43f3d0f208ab24ef6956ff19d4f0/osu.Game/Overlays/Profile/Header/Components/RankGraph.cs#L190

peppy:
Not sure, if that part hasn't failed yet I'm not sure it will. Would require the mouse to somehow be handled outside of the valid `DrawWidth`?

smoogipoo:
Given that the calls to UpdateBallPosition() are external, I'd agree with clamping it.

bdach (approved):

PR #10768, opened 2020-11-10
Always store standardised score when populating ScoreInfo

### LINKED ISSUE #10752: Earn too much PP on Classic Score display mode in osu!mania Gamemode
If the Score display mode is set to "Classic". The player earns way to much PP on osu!mania.

I tested this with a few maps, and it seems to work reliably on all osu!mania maps (which i could beat aka below 3* and 4k only)!  osu!standard and Score display mode "Classic" is seemingly correct.

Map: flowers feat. nori by in love with a ghost [4k bloom]
The link to the map (if needed): https://osu.ppy.sh/beatmapsets/962550#mania/2016814

The scores:
Scores_flowers feat  nori - in love with a ghost by -MysticEyes  4k bloom

On Score display mode: Classic
Score_Classic_flowers feat  nori - in love with a ghost by -MysticEyes  4k bloom

On Score display mode: Standardised
Score_Standardised_flowers feat  nori - in love with a ghost by -MysticEyes  4k bloom

osu!lazer version: 2020.1109.0
OS: Linux Ubuntu 20.04

Logs:
[runtime.log](https://github.com/ppy/osu/files/5512283/runtime.log)

[database.log](https://github.com/ppy/osu/files/5512275/database.log)

[network.log](https://github.com/ppy/osu/files/5512277/network.log)

[performance.log](https://github.com/ppy/osu/files/5512279/performance.log)

bdach:
Resolves #10752 (albeit in a slightly roundabout way).

The rationale is that the mania difficulty calculation is partially based on the score, which it takes in from the `ScoreInfo` passed in. If the play was registered with classic scoring, it'd take that instead and potentially balloon out of control past the expected 1M mark, which after this PR won't be possible. Note that this will not retroactively "fix" "broken" scores.

I think this is a good thing to have going forward given that we already do dynamic conversion. Hesitant to also add a local recalculation in the mania diffcalc component as that might break backward compatibility, and is sorta generally wasteful, but I'll do it on request.

I also noticed that sometimes replaying a replay will end up with a different score at the end than stored, but that'll probably be best addressed separately anyway as it's not likely to be specific to this change. I'm investigating that right now.

### DIFF
--- a/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
+++ b/osu.Game/Rulesets/Scoring/ScoreProcessor.cs
@@ -290,7 +290,7 @@ namespace osu.Game.Rulesets.Scoring
         /// </summary>
         public virtual void PopulateScore(ScoreInfo score)
         {
-            score.TotalScore = (long)Math.Round(TotalScore.Value);
+            score.TotalScore = (long)Math.Round(GetStandardisedScore());
             score.Combo = Combo.Value;
             score.MaxCombo = HighestCombo.Value;
             score.Accuracy = Math.Round(Accuracy.Value, 4);


peppy (approved):
I think this makes sense unless there's a reason for `PopulateScore` to behave otherwise.

smoogipoo (approved):
Yeah I'm completely fine with this. As you said, we have dynamic conversion to handle the rest.

PR #10770, opened 2020-11-10
Fix crashes if animation sprite path does not match a file

### LINKED ISSUE #10748: Crashes due to null drawables in some storyboard animations
**Describe the crash:** 

Some storyboard animations result in crashes due to null drawables being created

It seems more common that sprites for storyboard animations take the filename declared for the animation in the `.osb` file
and append numbers, starting from 0. I have found two exceptions
 
1) The first (0th) sprite has no 0 appended to the animation filename, from [beatmap 192137](https://osu.ppy.sh/beatmapsets/192137)

    ```
    Animation,Background,Centre,"noise\noise_0000.jpg",320,240,10,1,LoopForever
     F,0,49813,,1
     M,0,49813,50313,320,240
     M,0,50313,,320,240
    ```
    Note the filename is `"noise\noise_0000.jpg"` with four zeroes

    Then note that the first sprite matches this exactly, and the rest append their index as expected

    image

    This crash will occur at any time the animation is expected to play, the first being just before the 50 second mark

1) Animations intentionally declared with more frames than sprites to cause a blinking effect, from [beatmap 5381](https://osu.ppy.sh/beatmapsets/5381)

    ```
    Animation,Foreground,TopLeft,"spr\play-skip.png",525,410,3,1000,LoopForever
     S,0,-44000,,0.8
     F,0,-44000,,1
     F,0,-1000,,0
     F,0,0,,1
     F,0,29900,,0
    ```
    Note the `,3,` in line 1, I assume this number represents the number of frames

    Then note that there are only two sprites

    image

    When playing this beatmap in osu!stable you are be able to see the custom skip button blinking in the corner. In current osu!lazer the third frame becomes a null drawable, causing an almost immediate crash (since the skip button is scheduled for the very start)

1) This is just an example of a storyboard animation without issues, also from [beatmap 5381](https://osu.ppy.sh/beatmapsets/5381):

    ```
    Animation,Background,BottomLeft,"bg\gavel.png",0,480,5,300,LoopOnce
     F,0,22543,23705,1
    ```
    image

**osu!lazer version:** 

Built from latest master

**Solution:**
I have a [work in progress fix](https://github.com/cadon0/osu/pull/1). While it does resolve both crashes:
1) I am not sure this is the best approach. Advice would be appreciated. A check to see if the file exists would be much nicer than waiting for `CreateSpriteFromResourcePath(...)` to return `null`, but the assets seem to be managed by stores. Is that a path worth going down?
2) It does not allow the blinking effect - this seems like it might be a deeper issue that should be tackled separately
3) There's no tests yet 🙂

cadon0:
Closes https://github.com/ppy/osu/issues/10748 (see issue for discussion)

Are drawables typically only covered by visual tests? Feel free to point out an example if I should get this covered by an automated test as well

This can be tested by loading [beatmap 5381](https://osu.ppy.sh/beatmapsets/5381) into the storyboard visual test. You should see the custom "Skip" flashing in the bottom right - may not be visible if window size is small. Without these changes it will crash after loading

### DIFF
--- a/osu.Game/Storyboards/Drawables/DrawableStoryboardAnimation.cs
+++ b/osu.Game/Storyboards/Drawables/DrawableStoryboardAnimation.cs
@@ -118,24 +118,8 @@ namespace osu.Game.Storyboards.Drawables
             for (int frame = 0; frame < Animation.FrameCount; frame++)
             {
                 string framePath = Animation.Path.Replace(".", frame + ".");
-                Drawable sprite = storyboard.CreateSpriteFromResourcePath(framePath, textureStore);
-
-                if (sprite != null)
-                {
-                    AddFrame(sprite, Animation.FrameDelay);
-                    continue;
-                }
-
-                framePath = Animation.Path.Replace("0.", frame + ".");
-                sprite = storyboard.CreateSpriteFromResourcePath(framePath, textureStore);
-
-                if (sprite != null)
-                {
-                    AddFrame(sprite, Animation.FrameDelay);
-                }
-
-                // todo: handle animation intentionally declared with more frames than sprites to cause a blinking effect
-                // e.g. beatmap 5381's "spr\play-skip.png"
+                Drawable sprite = storyboard.CreateSpriteFromResourcePath(framePath, textureStore) ?? Drawable.Empty();
+                AddFrame(sprite, Animation.FrameDelay);
             }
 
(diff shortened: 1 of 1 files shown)


bdach (approved):
Don't see much point in adding tests for this, it's just a nullref crash anyway.

peppy (approved):

PR #10774, opened 2020-11-10
Add error placeholders when not logged in

### LINKED ISSUE #933: Loading animation doesn't disappear when not logged in
The loading animation doesn't disappear in various areas when you're not logged in, for example the leaderboard details tab:

Or the social browser:

Furthermore what do we do in the cases where we're not connected for these areas? Should we be showing a message in the area or blocking access to it?

### LINKED ISSUE #4550: osu!direct when a user isn't logged in

ghost:
This PR gives the changes in the main menus (Chat, Dashboard, BeatmapListing, Ranking). Adds an error placeholder when the player is not logged in. 

Closes #4550
Closes https://github.com/ppy/osu/issues/933

### `CheckIsLoggedIn()` Method

In each page mentioned, the `checkIsLoggedIn` method checks if the user is logged in, and hides the menu by adding a placeholder if the user is not logged in. If the user logs in again, the method is called, and the menu reappears.

### Structure

The different overlays are now derived from `IOnlineComponent` to allow the use of the API.
The different overlays have been added to `APIAccess.components` by the `Register()` method so that the `APIStateChanged() `method is called when the user logs in/logs out
### Demo

.


bdach:
Looks like some files are missing (presumably whatever `IOnlineComponent` is supposed to be), and there will be code style issues even if that is alleviated. Abstaining from review until that is addressed.

bdach:
@Remenems are you going to be bringing this PR to a state in which it compiles and passes CI?

Remenems:
We are thinking about it, but it seems that our code compiles with the version of the project from 2 weeks ago, but not the current version. It may take longer than expected to fix the problem.

bdach:
If you do decide to continue please see #10579 for a path forward as to how to proceed after removal of `IOnlineComponent`. Either way please let us know whether you will be continuing this.

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -86,6 +98,11 @@ public RankingsOverlay()
         [BackgroundDependencyLoader]
         private void load()
         {
+            apiState.BindTo(api.State);
+            apiState.BindValueChanged(onlineStateChanged, true);
+
+            placeholderContainer.Child = errorPlaceholder;
```
bdach, line 104: This can't work.

* In this particular case, `errorPlaceholder` is not even initialised yet, hence the error (`null Drawables may not be added to CompositeDrawable`).
* Even if it *were* initialised, the flow you're trying to do isn't correct. Drawables need to be loaded if they are to be shown (either by being in the drawable tree by the end of `load()`, or, if loading is required at a later point, via `LoadComponentAsync()`). Otherwise if you try to put them in the hierarchy, you'll get a possibly slightly-confusing error message (`Disposed drawables cannot be added to the scene graph`).

There is not much reason to ever create the login placeholder more than once. Therefore I would suggest to add it into the placeholder container already in the constructor, but start with it hidden by setting `Alpha = 0` on `placeholderContainer`.

bdach (changes requested):

Remenems:
It works now, but AppVeyor gives formatting errors. I'll fix that a little bit later. (Sorry we are very busy at the moment, but we'll do it)

LittleEndu:
You can run the check that CI runs yourself with the Inspect Code ps script. You shouldn't rely on online CI

osu.Game/Overlays/BeatmapListingOverlay.cs
```diff
@@ -37,6 +41,14 @@ public class BeatmapListingOverlay : FullscreenOverlay<BeatmapListingHeader>
 
         private OverlayScrollContainer resultScrollContainer;
 
+        private readonly IBindable<APIState> apiState = new Bindable<APIState>();
+
+        [Resolved]
+        private IAPIProvider api { get; set; }
+
+        private Container placeholderContainer;
+        private readonly Placeholder errorPlaceholder = new LoginPlaceholder(@"Please sign in to view beatmap listing!");
```
bdach, line 50: This assignment should ideally be inlined in `load()` for readability in all cases.

osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -453,6 +487,39 @@ private void postMessage(TextBox textbox, bool newText)
...
+                currentChannelContainer.Hide();
+                placeholderContainer.Show();
+                textbox.Hide();
+            }
+            else
+            {
+                currentChannelContainer.Show();
+                placeholderContainer.Hide();
+                textbox.Show();
+            }
```
bdach, line 504: Isn't this essentially duplicating `OnlineStateChanged()`? Why not just keep one or the other?

osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -288,6 +318,7 @@ private void currentChannelChanged(ValueChangedEvent<Channel> e)
             // mark channel as read when channel switched
             if (e.NewValue.Messages.Any())
                 channelManager.MarkChannelAsRead(e.NewValue);
+            CheckIsLoggedIn();
```
bdach, line 321: Do these checks really need to be all over the place here? Isn't the API state change enough?

osu.Game/Overlays/DashboardOverlay.cs
```diff
@@ -141,6 +159,13 @@ private void onTabChanged(ValueChangedEvent<DashboardOverlayTabs> tab)
 
         private void onlineStateChanged(ValueChangedEvent<APIState> state) => Schedule(() =>
         {
+            if (state.NewValue == APIState.Online)
+                placeholderContainer.Hide();
+            else if (state.NewValue == APIState.Offline)
+            {
+                placeholderContainer.Show();
+                loading.Hide();
```
bdach, line 167: Kind of suspicious how the loading layer is hidden but then not shown again. Did you test that this behaves correctly after the user logs out and then back in again?

osu.Game/Overlays/RankingsOverlay.cs
```diff
@@ -239,5 +259,20 @@ protected override void Dispose(bool isDisposing)
...
+            if (state.NewValue == APIState.Online)
+            {
+                contentContainer.Show();
+                placeholderContainer.Hide();
+            }
+            else if (state.NewValue == APIState.Offline)
+            {
+                contentContainer.Hide();
+                placeholderContainer.Show();
+                loading.Hide();
```
bdach, line 274: Same as above.

bdach (changes requested):
In addition to CI inspections which have not been fully resolved yet:

osu.Game/Overlays/ChatOverlay.cs
Remenems: The problem is that the condition in OnlineStateChanged is not exactly the same. CheckIsLoggedIn() is called at "strategic" places, where we need to check if the user is logged in (Change of channel for example).

osu.Game/Overlays/ChatOverlay.cs
Remenems: In fact, if the checkIsLoggedIn() method is not called, the chat will be displayed when the user will change channel even if the user is not logged in, because there is no status change.

osu.Game/Overlays/DashboardOverlay.cs
Remenems: There is another place in the page code, where the loading spinner is displayed when needed.

Remenems:
@bdach I will try to take all your comments into consideration, although I'm very busy at the moment, I'm not sure I'll be able to do that right away.
Nevertheless, the CI inspection gives me strange errors: 
- It tells me to remove the `using System` that was already there before.
- It tells me to cross a line, where CodeFactor tells me to remove one.
I don't really know what to do :/

peppy:
@EVAST9919 may be worthwhile looking at this PR while you're working on #11470. They intertwine quite a bit.

As the author of this one is not around, it could be a good idea to get it up-to-date and see how it works in conjunction with your loading layer changes.

PR #10775, opened 2020-11-10
Disable kiai hit explosions on legacy skins

### LINKED ISSUE #10083: Taiko kiai glow should be disabled on legacy skins
When you reach a Kiai section in taiko, every time you hit a circle, a pink or blue vertical glow shows up, indicating that you hit a don or kat. Problem is: like several options in the game it confuses players, making it hard to see next notes and confuses you to clicking the opposite note instead.

opera_AlpsiSVRmj opera_nxZCLmAcmG

Would be amazing to have a feature to disable it, as no skinning element from stable nor any menu option is able to disable this, and specially with Spotlights coming up with a new season and players wanting to play it more and more. Having more accessibility would be great!

bdach:
Closes #10083.

Mostly mirrors `HitExplosion` in structure, although the layout is (and was) a little bit weirder, and the `EmptyExplosion` thing sticks out like a sore thumb - but I need to expire the `Empty()` to avoid accumulating them in the parent container.

Comparison of the draw hierarchy before/after, for reference:

| before | after |
| :-: | :-: |
| 2020-11-10-141416_1920x1060_scrot | 2020-11-10-141600_1920x1060_scrot |

### DIFF
--- a/osu.Game.Rulesets.Taiko/UI/KiaiHitExplosion.cs
+++ b/osu.Game.Rulesets.Taiko/UI/KiaiHitExplosion.cs
@@ -1,0 +1,47 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osu.Game.Rulesets.Objects.Drawables;
+using osu.Game.Rulesets.Taiko.Objects;
+using osu.Game.Skinning;
+using osuTK;
+
+namespace osu.Game.Rulesets.Taiko.UI
+{
+    public class KiaiHitExplosion : Container
+    {
+        public override bool RemoveWhenNotAlive => true;
+
+        [Cached(typeof(DrawableHitObject))]
+        public readonly DrawableHitObject JudgedObject;
+
+        private readonly HitType hitType;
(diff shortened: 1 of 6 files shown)


osu.Game.Rulesets.Taiko/UI/KiaiHitExplosion.cs
```diff
@@ -1,71 +1,53 @@
...
-            this.ScaleTo(new Vector2(1, 3f), 500, Easing.OutQuint);
-            this.FadeOut(250);
-
-            Expire(true);
-        }
+        /// <summary>
+        /// Helper function to use when an explosion is not desired.
+        /// Lifetime is set to avoid accumulating empty drawables in the parent container.
+        /// </summary>
+        public static Drawable EmptyExplosion() => Empty().With(d => d.LifetimeEnd = double.MinValue);
```
peppy, line 51: maybe just inline this at the point of usage? don't think there's much value in leaving it here unless it's potentially going to be reused somewhere?

also might look better as `new Drawable { Lifetime End = ...`

peppy (changes requested):

osu.Game.Rulesets.Taiko/UI/KiaiHitExplosion.cs
bdach: That's how I had it at the start but wasn't sure if it didn't look overly hideous. Will inline locally instead.

peppy (approved):

Issue #10512, opened 2020-10-15
Displaying the background on the results screen.


Moltanicaa:
**Describe the new feature:** 
1. Now, when viewing the results screen, the picture is very blurry. Could it be possible for the current background blur setting to influence this? So that you can turn off the blur and take a screenshot (displaying the background on the results screen is more important than you think).
2. Can I remove the forced screen dimming to almost 100% when opening extended results? It is practically useless. May be a disabled feature. The reason is the same - we love to share the results so that the background is visible.
**Proposal designs of the feature:** 
Can give personal settings for this section?

Issue #7376, opened 2019-12-27
Editor should have background dim/blur


recapitalverb:
**Describe the new feature:** The osu!editor should have background dim/blur, like in normal gameplay.

huoyaoyuan:
And also test playing in editor (if the concept doesn't change).

bdach:
> And also test playing in editor (if the concept doesn't change).

Seems out of scope for this particular issue, probably deserves a separate one or a mention over at #6686.

greevak:
Bump.

PR #10785, opened 2020-11-11
Make beatmap selection blur editable

GiantNuker:
Adds a slider to the options menu to change the menu blur (Personally I don't like it so its 0%)
It defaults to 80% (The slider is out of 25) which matches osu's blur beforehand.

[Default Blur](https://cdn.nuker.dev/khcpq5ii/2020-11-10T17:14:11.png) (Slider at 80% - blur int 20)
[30% Blur](https://cdn.nuker.dev/khcqd1yb/2020-11-10T17:31:51.png) (Slider at 30% - blur int 6
[Without Blur](https://cdn.nuker.dev/khcpw1vy/2020-11-10T17:17:48.png)

Personally I just don't really like the blur, so I have it disabled

### DIFF
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
@@ -372,7 +372,7 @@ namespace osu.Game.Tests.Visual.Background
 
             public void TriggerOnHover() => OnHover(new HoverEvent(new InputState()));
 
-            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BACKGROUND_BLUR);
+            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BackgroundBlur);
 
             protected override BackgroundScreen CreateBackground() => new FadeAccessibleBackground(Beatmap.Value);
         }
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
@@ -307,7 +307,7 @@ namespace osu.Game.Tests.Visual.Background
 
             public bool IsBackgroundVisible() => ((FadeAccessibleBackground)Background).CurrentAlpha == 1;
 
-            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BACKGROUND_BLUR);
+            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BackgroundBlur);
 
             /// <summary>
             /// Make sure every time a screen gets pushed, the background doesn't get replaced
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
@@ -306,7 +306,7 @@ namespace osu.Game.Tests.Visual.Background
 
             public bool IsBackgroundVisible() => ((FadeAccessibleBackground)Background).CurrentAlpha == 1;
 
-            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BACKGROUND_BLUR);
+            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BackgroundBlur);
 
             /// <summary>
             /// Make sure every time a screen gets pushed, the background doesn't get replaced
--- a/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
+++ b/osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
@@ -160,10 +160,10 @@ namespace osu.Game.Tests.Visual.Background
                 player.ReplacesBackground.Value = true;
                 player.StoryboardEnabled.Value = true;
             });
-            AddStep("Enable user dim", () => player.DimmableStoryboard.EnableUserDim.Value = true);
+            AddStep("Enable user dim", () => player.DimmableStoryboard.EnableGameplayDim.Value = true);
             AddStep("Set dim level to 1", () => songSelect.DimLevel.Value = 1f);
             AddUntilStep("Storyboard is invisible", () => !player.IsStoryboardVisible);
-            AddStep("Disable user dim", () => player.DimmableStoryboard.EnableUserDim.Value = false);
+            AddStep("Disable user dim", () => player.DimmableStoryboard.EnableGameplayDim.Value = false);
             AddUntilStep("Storyboard is visible", () => player.IsStoryboardVisible);
         }
 
@@ -277,7 +277,7 @@ namespace osu.Game.Tests.Visual.Background
             protected override BackgroundScreen CreateBackground()
             {
                 FadeAccessibleBackground background = new FadeAccessibleBackground(Beatmap.Value);
-                DimEnabled.BindTo(background.EnableUserDim);
+                DimEnabled.BindTo(background.EnableGameplayDim);
                 return background;
             }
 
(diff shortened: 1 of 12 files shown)


osu.Game/Screens/Play/PlayerLoader.cs
```diff
@@ -111,6 +111,12 @@ public PlayerLoader(Func<Player> createPlayer)
             this.createPlayer = createPlayer;
         }
 
+        public PlayerLoader(Func<Player> createPlayer, float backgroundBlur)
```
peppy, line 114: this shouldn't be required to be passed in here. `ScreenWithBeatmapBackground` should keep the "base" blur and then each screen should apply its multiplication to this.

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -45,9 +46,11 @@ public abstract class SongSelect : OsuScreen, IKeyBindingHandler<GlobalAction>
     {
         public static readonly float WEDGE_HEIGHT = 245;
 
-        protected const float BACKGROUND_BLUR = 20;
+        protected float BackgroundBlur = 20;
         private const float left_area_padding = 20;
 
+        private Bindable<float> backgroundBlurBindable { get; set; }
```
peppy, line 52: move this to `ScreenWithBeatmapBackground`

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -308,6 +311,26 @@ private void load(AudioManager audio, DialogOverlay dialog, OsuColour colours, S
...
+        {
+            base.LoadComplete();
+
+            backgroundBlurBindable.ValueChanged += _ => updateBackgroundBlur();
+            updateBackgroundBlur(); // for first time opening beatmap selection
+        }
+
+        private void updateBackgroundBlur()
+        {
+            BackgroundBlur = backgroundBlurBindable.Value * 25;
```
peppy, line 328: this 25 should still be a constant as it was. `BlurAmount` is already exposed from `ScreenWithBeatmapBackground` and can be used in each screen to apply different multiplications of it.

peppy (changes requested):

osu.Game/Screens/Select/SongSelect.cs
GiantNuker: Shouldn't it be moved to `BackgroundScreenBeatmap`?

peppy: That's the one, sorry for the confusion.

GiantNuker: Sorry if im missing something but it seems `userBlurLevel` is tied to `OsuSetting.BlurLevel` which is ingame blur right?

peppy: Yes, well you're going to have to figure out how to make this work. There's gameplay blur and UI blur now.

The class handles the case where user blur override can be enabled or not. The blur you're wanting to add is the case where "user blur" is "not applied". But now you're adding a new user blur kind.

It will probably involve renaming and refactoring the existing code to ensure it is understandable given the new complexities.

GiantNuker: Should it be added in `UserDimContainer` where `EnableUserDim` is, or in `BackgroundScreenBeatmap` as an additional boolean that overrides `EnableUserDim`?

peppy: `UserDimContainer` is specific to gameplay. It should be in `BackgroundScreenBeatmap` as above, in the case that "EnableUserDim" is not set. Although again, you need to rename "UserDim" to "GameplayDim" or something.

GiantNuker: `backgroundModeBeatmap.BlurAmount` is directly set in `SongSelect`, seems like adding the option to `BackgroundScreenBeatmap` won't actually work because its set to multiple different values (ex `20` in SongSelect` and `15` in `PlayerLoader`)

GiantNuker: Nevermind, looks like i'm just a bit dumb

GiantNuker:
Not sure how to fix the tests, can't be a constant value because the background blur isn't a constant anymore - should I pull the setting bindable in the test?

bdach:
Which test are we talking about there? Hard to tell with the review fixes presumably unpushed.

Joehuu:
This currently resolves half of https://github.com/ppy/osu/issues/10512 (results screen), which is blur. This doesn't seem to apply to editor right now, will also fix half of https://github.com/ppy/osu/issues/7376. Not sure if we should bundle dim here too.

As for bugs, going to song select for the second time doesn't apply blur correctly.

peppy:
it should not apply blur anywhere blue was not previously applied

peppy:
honestly tempted to say this should just be a toggle for simplicity.

GiantNuker:
> As for bugs, going to song select for the second time doesn't apply blur correctly.

@Joehuu when does this happen? I tried passing, failing, quitting while beatmap is loading, going back to the main menu, retrying and passing/failing, and the blur works fine for me. Everything is pushed except changing the blur amount in PlayerLoader to .75 (same blur as before pr)

Joehuu:
IcaUre5agF

GiantNuker:
Thanks, can reproduce that

osu.Game/Screens/Play/Player.cs
```diff
@@ -672,7 +672,8 @@ public override void OnEntering(IScreen last)
                 .Delay(250)
                 .FadeIn(250);
 
-            Background.EnableUserDim.Value = true;
+            Background.EnableGameplayDim.Value = true;
+            Background.EnableUIBlur.Value = false;
```
bdach, line 676: This feels all kinds of wrong. `Background` should probably expose a bindable of an enumeration type that's an actual exclusive choice, such as

```csharp
public enum DimMode
{
    Off,
    UI,
    Gameplay
}
```

and then adjust depending on that.

osu.Game/Screens/Backgrounds/BackgroundScreenBeatmap.cs
```diff
@@ -148,21 +165,25 @@ public override void Add(Drawable drawable)
...
             /// As an optimisation, we add the two blur portions to be applied rather than actually applying two separate blurs.
             /// </summary>
-            private Vector2 blurTarget => EnableUserDim.Value
-                ? new Vector2(BlurAmount.Value + (float)userBlurLevel.Value * USER_BLUR_FACTOR)
-                : new Vector2(BlurAmount.Value);
+            private Vector2 blurTarget => EnableGameplayDim.Value
+                ? new Vector2(BlurAmount.Value + (float)gameplayBlurLevel.Value * GAMEPLAY_BLUR_FACTOR)
+                : EnableUIBlur.Value
+                    ? new Vector2(BlurAmount.Value * (float)uiBlurLevel.Value * UI_BLUR_FACTOR)
+                    : new Vector2(BlurAmount.Value);
```
bdach, line 172: Nested ternaries are unreadable. Please split or make an actual method.

osu.Game/Screens/Backgrounds/BackgroundScreenBeatmap.cs
```diff
@@ -18,23 +18,33 @@ namespace osu.Game.Screens.Backgrounds
...
+        /// </summary>
+        public readonly Bindable<bool> EnableUIBlur = new Bindable<bool>();
 
         public readonly Bindable<bool> StoryboardReplacesBackground = new Bindable<bool>();
 
         /// <summary>
-        /// The amount of blur to be applied in addition to user-specified blur.
+        /// The amount of blur to be applied in addition to the game-specified or UI-specified blur.
         /// </summary>
         public readonly Bindable<float> BlurAmount = new BindableFloat();
```
bdach, line 49: This property is now intensely confusing. If gameplay dim is on, the amount is added onto the existing blur, and if UI dim is on, it *multiplies existing blur*. This is all way too complex to live in its current state.

osu.Game.Tests/Visual/Background/TestSceneUserDimBackgrounds.cs
```diff
@@ -371,7 +371,7 @@ public TestPlayerLoader(Player player)
 
             public void TriggerOnHover() => OnHover(new HoverEvent(new InputState()));
 
-            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BACKGROUND_BLUR);
+            public bool IsBlurCorrect() => ((FadeAccessibleBackground)Background).CurrentBlur == new Vector2(BackgroundBlur);
```
bdach, line 374: This doesn't compile. If the current user settings are to be fetched, resolve an `OsuConfigManager` (examples how to do so can be found in lots of existing test scenes).

bdach (changes requested):
Needs work & testing in the test scene.

GiantNuker:
I'll do that when i have some time

smoogipoo:
This will be done via the skin editor instead.

peppy:
Well actually, I think it'll be a setting (because it's something a user will want to apply over all skins they use), but closing this is fine as it would need some insane merge magic anyway.

PR #10787, opened 2020-11-11
Add the ability for the game OSD to display user bindings

peppy:
Adds binding display for mouse button toggle / HUD toggle keys.

- [x] Depends on #10786 for ease-of-merge

### DIFF
--- a/osu.Game/Configuration/OsuConfigManager.cs
+++ b/osu.Game/Configuration/OsuConfigManager.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
+using System.Diagnostics;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
 using osu.Framework.Configuration.Tracking;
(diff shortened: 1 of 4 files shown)


osu.Game/Configuration/OsuConfigManager.cs
```diff
@@ -172,12 +175,28 @@ public void Migrate()
...
-            new TrackedSetting<ScalingMode>(OsuSetting.Scaling, m => new SettingDescription(m, "scaling", m.GetDescription())),
-        };
+            // these need to be assigned in normal game startup scenarios.
+            Debug.Assert(LookupKeyBindings != null);
+            Debug.Assert(LookupSkinName != null);
+
+            return new TrackedSettings
+            {
+                new TrackedSetting<bool>(OsuSetting.MouseDisableButtons, v => new SettingDescription(!v, "gameplay mouse buttons", v ? "disabled" : "enabled", LookupKeyBindings(GlobalAction.ToggleGameplayMouseButtons))),
+                new TrackedSetting<HUDVisibilityMode>(OsuSetting.HUDVisibilityMode, m => new SettingDescription(m, "HUD Visibility", m.GetDescription(), $"cycle: shift-tab quick view: {LookupKeyBindings(GlobalAction.HoldForHUD)}")),
```
Joehuu, line 187: Seems like toggling hud is not a global action. I think it should be as holding for hud is.

peppy: correct. i'm against adding it as configurable unless someone actually wants that.

smoogipoo (approved):

PR #10789, opened 2020-11-11
Add hotkey to select a new random skin

peppy:
Defaults to ctrl-shift-r. Undecided if random should include defaults or only user skins, open to opinion.

- [x] Depends on #10787.
- [x] Depends on #3652.


smoogipoo:
Are you sure we really want/need a hotkey for this? Every global hotkey we add is potentially adding more conflicts... Are you sure we wouldn't want to use Ctrl+Shift+R in the future for the editor, for example...?

peppy:
I'm sure this has been a request in the past. I'd see this as a pretty common usage scenario? Will do a quick poll if you'd prefer.

smoogipoo (approved):

PR #10790, opened 2020-11-11
Fix frames potentially getting added to spectator replay in wrong format

### LINKED ISSUE #10777: Game crash after entering spectator mode
**Describe the crash:** 
i tried to spectate someone and then the game suddenly started to hang and crash.
this doesn't seem to repro all the time and im not sure how to do it.

**osu!lazer version:** 
2020.1109.0

**Logs:**
[logs.zip](https://github.com/ppy/osu/files/5518314/logs.zip)

**Computer Specifications:** 
OS: windows 10
CPU: i5-4440
iGPU: intel hd graphics 4600
RAM: 8 GB

peppy:
The way spectator currently works, the `Spectator` screen is responsible for adding new frames to the replay, even when it has a child (`SpectatorPlayer`) present.

There was a possibility that a new play had already started, and on returning to the Spectator screen (to initialise the new play) there would be a brief period where the Player instance is still reading from the replay, the `userBeganPlaying` call had not yet finished initialising the new target replay, and `userSentFrames` is run (asynchronously), writing frames to the previous replay using the incorrect ruleset instance).

To make this work, it doesn't `Schedule` frame addition to the replay (making things a bit unsafe). Changing this itself isn't such a simple one to do, so I instead opted to fix this via locking.

Closes https://github.com/ppy/osu/issues/10777.

### DIFF
--- a/osu.Game/Screens/Play/Spectator.cs
+++ b/osu.Game/Screens/Play/Spectator.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using JetBrains.Annotations;
 using osu.Framework.Allocation;
@@ -61,7 +62,9 @@ namespace osu.Game.Screens.Play
         [Resolved]
         private RulesetStore rulesets { get; set; }
 
-        private Replay replay;
+        private Score score;
+
+        private readonly object scoreLock = new object();
 
         private Container beatmapPanelContainer;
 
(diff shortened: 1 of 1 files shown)


smoogipoo:
So are you saying that there is no guarantee of synchronous execution between calls at SignalR level? That is, `userBeganPlaying` can have not finished while `userSentFrames` is called? Or is it that `userBeganPlaying` callbacks (both in `Spectator` and `SpectatorPlayer`) are scheduled?

This sounds like `Spectator` is serving too much of a purpose by adding frames to the replay. Is there a reason why it was done this way and not inside `SpectatorPlayer`? At the very least then you can have something like on `userFinishedPlaying` stop accepting frames and schedule the `Exit()`, and check that flag from `userSentFrames` before adding more frames. I think it doesn't matter that a few frames from the subsequent replay may potentially be discarded.

peppy:
I'm pretty sure signalr is single thread. That's not what's at play here. It's the local `Schedule` calls for `userBeganPlaying`

smoogipoo:
> allow receiving frames without being in player

So that means that the spectator player is always behind real-time by however long your download took? That's kind of weird in itself, because it breaks the notion of real time playback that people have grown to expect from #spectator and all that.
I think spectator is going to have to get as close to the user's play time as possible in any case, at which point discarding all earlier frames during download + load is not a big deal, except for rewind, at which point we probably want to have _full_ replay data anyway (not just from download + load).

peppy:
That's correct. The option will be given via a button to "skip to live", or can already be done via the seek bar.

smoogipoo:
My point is that "skip to live" should be default, it should be pressed once the screen is entered. Every other service handles streaming this way - youtube, twitch, even chat clients put you at the "live" point instead of however far back the messages go.

This is only to simplify this entire pre-buffering interaction, which as I said if you want to skip to non-live you need full replay data anyway not just download + load. Basically having this download + load behaviour sounds like a bandaid solution to me and is only overcomplicating this implementation.

peppy:
Sure, definitely. We can add the default skip to live later on.

The logic being at a higher level needs to remain because the server is going to be sending all frames up to current point of play eventually. Maybe we can consider moving it to `SpectatorStreamingClient` in the future?

smoogipoo (approved):
I'm only like 60% confident with this code, but seems ok.

smoogipoo:
I dunno... If I were to structure the requests I'd do it a different way altogether, by adding intermediate requests to retrieve metadata without also registering to receiving frame bundles, and then the frame bundle registration would occur inside `SpectatorPlayer` only.

That sort of structure would give us the frames exactly where we want them and only for as long as we want them, because it's all handled by `SpectatorPlayer`'s lifetime.

Issue #8918, opened 2020-04-30
Ignore preview points


enoslayd:
**Describe the new feature:** allow user to disable preview points (the one where a song starts from a certain point in song select). There will be several choices:

Default - default behavior
Only 1st preview - previews will use preview point only at the beginning of preview and after the song ends, will repeat the beginning of song
Disable - completely ignores preview point

**Proposal designs of the feature:** A dropdown box at the song select group in settings

peppy:
I see this as unnecessary customisation, personally. If you want to listen to full songs using the main menu is recommended.

enoslayd:
But maybe reconsider it? What if we want to repeat song?

peppy:
It's a very common concept to repeat from the preview point, and gives the song select screen a feel of "selecting a song" to it. Changing this feels wrong to me.

Joehuu:
> But maybe reconsider it? What if we want to repeat song? (Song select is the only way)

I think it should be implemented in the now playing overlay (with collections). Going with stable behavior, you can put one song in a collection and it'll repeat.

peppy:
I'm going to close this one for now. We may revisit this at a later date, but song select audio behaviour is going to match stable for the time being.

PR #10791, opened 2020-11-11
Option to disable beatmaps playing at the preview time in beatmap selection

GiantNuker:
Adds two toggles to settings, one to disable starting at the preview time when repeating a song, and one to disable preview time when selecting a new song.

By default starting from preview time when replaying is disabled.


peppy:
Strong "no" from me in favour of not adding settings for every last thing. The preview / looping behaviour here is considered core game mechanics to me.

Joehuu:
There's https://github.com/ppy/osu/issues/8918, closed though.

GiantNuker:
Alright, I guess that makes sense. 

> In addition, I'd recommend searching issues (and opening them if not existing) before PRing changes like this, just to save effort.
I would've done it anyway just to use myself. Will search prs/issues before pring again though

smoogipoo:
osu! is not a music player. The preview point is designed to get you right into the action and give you a taste of what to expect, which is imo crucial for enjoyment of a rhythm game.

Hard no from me too, sorry.

PR #10793, opened 2020-11-11
Refactor "perform from menu" logic to be aware of dialog prompts

### LINKED ISSUE #10761: Trying to spectate someone while being inside the editor beatmap setup screen spams popup messages
https://youtu.be/1j59i-SJWPo

**osu!lazer version:** 
2020.1109.0

peppy:
Until now, a request to perform an action from the main menu would keep trying until it was manually cancelled or reached its target screen. This change allows pausing when a dialog is being presented to the user.

- While a dialog is presented, parent screen navigation will stop
- If the dialog is dismissed and the screen changed (ie. user confirmed exiting a screen), navigation upwards will continue
- If the dialog was dismissed and the screen hasn't changed, the action is cancelled and the user is informed.

This is allowed to repeat each time an confirmation-causing-upwards-navigation occurs.

Closes https://github.com/ppy/osu/issues/10761

Also feels like a good improvement for code quality, moving this complex logic out of `OsuGame` itself.

### DIFF
--- a/osu.Game/PerformFromMenuRunner.cs
+++ b/osu.Game/PerformFromMenuRunner.cs
@@ -1,0 +1,145 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using osu.Framework.Allocation;
+using osu.Framework.Bindables;
+using osu.Framework.Graphics;
+using osu.Framework.Screens;
+using osu.Framework.Threading;
+using osu.Game.Beatmaps;
+using osu.Game.Overlays;
+using osu.Game.Overlays.Dialog;
+using osu.Game.Overlays.Notifications;
+using osu.Game.Screens.Menu;
+
+namespace osu.Game
+{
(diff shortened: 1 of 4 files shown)


osu.Game/PerformFromMenuRunner.cs
```diff
@@ -0,0 +1,145 @@
...
+            while (current != null)
+            {
+                if (validScreens.Contains(current.GetType()))
+                {
+                    current.MakeCurrent();
+                    break;
+                }
+
+                current = current.GetParentScreen();
+            }
```
bdach, line 97: This seems to also have been a theoretical problem earlier before, but what happens if none of the target screens to exit to is in the stack? If I'm reading correctly, looks like it'll spin every 200ms until canceled by the next perform.

peppy: MainMenu is always added and should always be in the stack (that was the intention).

peppy: Considered potentially checking the hierarchy for the presence of `MainMenu` and aborting if not there, but I don't think that's the behaviour we want. This `PresentFromMainMenu` may be called during the startup process where we are not at the menu yet, and it should still work in such cases.

bdach (approved):

PR #10797, opened 2020-11-11
Implement top-level hitobject pooling

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10796 (for the test to work)
- [x] https://github.com/ppy/osu/pull/10802

I won't be PRing the osu! changes until nested hitobject pooling + editor support is added as there would be quite a lot of allocations, but this branch has been implemented for osu! here: https://github.com/smoogipoo/osu/tree/osu-hitobject-pooling if you desire some early centipede-ing.

I've omitted a few things here from the full set of changes https://github.com/smoogipoo/osu/tree/hitobject-pooling, notable:
1. No lifetime extensions (removed in https://github.com/ppy/osu/commit/7d020181343f942159ae21df57b3ca8d07aad9f6)
2. No events for DHOs coming "in use" (or "current" in that branch above). (removed in https://github.com/ppy/osu/commit/7d020181343f942159ae21df57b3ca8d07aad9f6)
3. No editor support for keeping DHOs alive. (removed in https://github.com/ppy/osu/commit/7d020181343f942159ae21df57b3ca8d07aad9f6)
4. No editor/blueprint support whatsoever.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/MainCirclePiece.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/MainCirclePiece.cs
@@ -51,9 +51,15 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
             var drawableOsuObject = (DrawableOsuHitObject)drawableObject;
 
             state.BindTo(drawableObject.State);
-            state.BindValueChanged(updateState, true);
-
             accentColour.BindTo(drawableObject.AccentColour);
+            indexInCurrentCombo.BindTo(drawableOsuObject.IndexInCurrentComboBindable);
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            state.BindValueChanged(updateState, true);
             accentColour.BindValueChanged(colour =>
             {
                 explode.Colour = colour.NewValue;
@@ -61,7 +67,6 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
                 circle.Colour = colour.NewValue;
             }, true);
 
-            indexInCurrentCombo.BindTo(drawableOsuObject.IndexInCurrentComboBindable);
             indexInCurrentCombo.BindValueChanged(index => number.Text = (index.NewValue + 1).ToString(), true);
         }
 
--- a/osu.Game/Rulesets/UI/DrawableRuleset.cs
+++ b/osu.Game/Rulesets/UI/DrawableRuleset.cs
@@ -559,23 +559,22 @@ namespace osu.Game.Rulesets.UI
         protected void RegisterPool<TObject, TDrawable>(int initialSize, int? maximumSize = null)
             where TObject : HitObject
             where TDrawable : DrawableHitObject, new()
+            => RegisterPool<TObject, TDrawable>(new DrawablePool<TDrawable>(initialSize, maximumSize));
+
+        /// <summary>
+        /// Registers a <see cref="DrawableHitObject"/> pool with this <see cref="DrawableRuleset"/> which is to be used whenever
+        /// <see cref="DrawableHitObject"/> representations are requested for the given <typeparamref name="TObject"/> type (via <see cref="GetPooledDrawableRepresentation"/>).
+        /// </summary>
+        /// <param name="pool">The <see cref="DrawablePool{T}"/> to register.</param>
+        /// <typeparam name="TObject">The <see cref="HitObject"/> type.</typeparam>
+        /// <typeparam name="TDrawable">The <see cref="DrawableHitObject"/> receiver for <typeparamref name="TObject"/>s.</typeparam>
+        protected void RegisterPool<TObject, TDrawable>([NotNull] DrawablePool<TDrawable> pool)
+            where TObject : HitObject
+            where TDrawable : DrawableHitObject, new()
         {
-            var pool = CreatePool<TDrawable>(initialSize, maximumSize);
             pools[typeof(TObject)] = pool;
             AddInternal(pool);
         }
-
-        /// <summary>
-        /// Creates the <see cref="DrawablePool{T}"/> to retrieve <see cref="DrawableHitObject"/>s of the given type from.
-        /// </summary>
-        /// <param name="initialSize">The number of hitobject to be prepared for initial consumption.</param>
-        /// <param name="maximumSize">An optional maximum size after which the pool will no longer be expanded.</param>
-        /// <typeparam name="TDrawable">The type of <see cref="DrawableHitObject"/> retrievable from this pool.</typeparam>
-        /// <returns>The <see cref="DrawablePool{T}"/>.</returns>
-        [NotNull]
-        protected virtual DrawablePool<TDrawable> CreatePool<TDrawable>(int initialSize, int? maximumSize = null)
-            where TDrawable : DrawableHitObject, new()
-            => new DrawablePool<TDrawable>(initialSize, maximumSize);
 
         /// <summary>
         /// Attempts to retrieve the poolable <see cref="DrawableHitObject"/> representation of a <see cref="HitObject"/>.
--- a/osu.Game/Rulesets/UI/DrawableRuleset.cs
+++ b/osu.Game/Rulesets/UI/DrawableRuleset.cs
@@ -526,6 +526,9 @@ namespace osu.Game.Rulesets.UI
         /// </summary>
         public abstract void CancelResume();
 
+        private readonly Dictionary<Type, IDrawablePool> pools = new Dictionary<Type, IDrawablePool>();
+        private readonly Dictionary<HitObject, HitObjectLifetimeEntry> lifetimeEntries = new Dictionary<HitObject, HitObjectLifetimeEntry>();
+
         /// <summary>
         /// Whether this <see cref="DrawableRuleset"/> should retrieve pooled <see cref="DrawableHitObject"/>s.
         /// </summary>
@@ -534,8 +537,14 @@ namespace osu.Game.Rulesets.UI
         /// </remarks>
         protected virtual bool PoolHitObjects => false;
 
-        private readonly Dictionary<Type, IDrawablePool> pools = new Dictionary<Type, IDrawablePool>();
-
+        /// <summary>
+        /// Registers a <see cref="DrawableHitObject"/> pool with this <see cref="DrawableRuleset"/> which is to be used whenever
+        /// <see cref="DrawableHitObject"/> representations are requested for the given <typeparamref name="TObject"/> type (via <see cref="GetDrawableRepresentation"/>).
+        /// </summary>
+        /// <param name="initialSize">The number of drawables to be prepared for initial consumption.</param>
+        /// <param name="maximumSize">An optional maximum size after which the pool will no longer be expanded.</param>
+        /// <typeparam name="TObject">The <see cref="HitObject"/> type.</typeparam>
+        /// <typeparam name="TDrawable">The <see cref="DrawableHitObject"/> receiver for <typeparamref name="TObject"/>s.</typeparam>
         protected void RegisterPool<TObject, TDrawable>(int initialSize, int? maximumSize = null)
             where TObject : HitObject
             where TDrawable : DrawableHitObject, new()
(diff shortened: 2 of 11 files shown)


osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -38,9 +41,8 @@ namespace osu.Game.Rulesets.UI
     public abstract class DrawableRuleset<TObject> : DrawableRuleset, IProvideCursor, ICanAttachKeyCounter
         where TObject : HitObject
     {
-        public override event Action<JudgementResult> OnNewResult;
-
-        public override event Action<JudgementResult> OnRevertResult;
+        public override event Action<JudgementResult> NewResult;
+        public override event Action<JudgementResult> RevertResult;
```
smoogipoo, line 45: I can revert this if you'd like, it is a bit of a breaking change but is unlikely to be used by any rulesets.

The reason for the rename is to match `Playfield` + `DrawableHitObject` + `HitObjectContainer`.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/MainCirclePiece.cs
```diff
@@ -51,17 +51,22 @@ private void load()
             var drawableOsuObject = (DrawableOsuHitObject)drawableObject;
 
             state.BindTo(drawableObject.State);
-            state.BindValueChanged(updateState, true);
-
             accentColour.BindTo(drawableObject.AccentColour);
+            indexInCurrentCombo.BindTo(drawableOsuObject.IndexInCurrentComboBindable);
+        }
+
+        protected override void LoadComplete()
```
smoogipoo, line 58: This change should be pretty self-explanatory - `HitObject` may not be set during `load()`. These changes match what's done in `LegacyCirclePiece` and `LegacySpinner`, etc.

osu.Game/Rulesets/UI/HitObjectContainer.cs
```diff
@@ -1,78 +1,174 @@
...
+        public event Action<DrawableHitObject, JudgementResult> RevertResult;
+
+        private readonly Dictionary<DrawableHitObject, IBindable> startTimeMap = new Dictionary<DrawableHitObject, IBindable>();
+        private readonly Dictionary<HitObjectLifetimeEntry, DrawableHitObject> drawableMap = new Dictionary<HitObjectLifetimeEntry, DrawableHitObject>();
+        private readonly LifetimeEntryManager lifetimeManager = new LifetimeEntryManager();
+
+        [Resolved(CanBeNull = true)]
+        private DrawableRuleset drawableRuleset { get; set; }
 
         public HitObjectContainer()
```
smoogipoo, line 51: These changes look pretty rough, I know, but most of it is moving around.

The code in `#region Non-pooling support` should be pretty similar, as should the code in `#region Comparator + StartTime tracking` with the exception of resorting done via `SortInternal()` instead of remove-then-add (required for `PooledDrawables` to not be returned to the pool later on).

The rest of new code is inside `#region Pooling support`.

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -101,13 +117,68 @@ private void load()
...
+            OnHitObjectRemoved(h.HitObject);
+            return false;
+        }
+
+        private readonly Dictionary<HitObject, HitObjectLifetimeEntry> lifetimeEntryMap = new Dictionary<HitObject, HitObjectLifetimeEntry>();
+
+        /// <summary>
+        /// Adds a <see cref="HitObject"/> to this <see cref="Playfield"/>.
+        /// </summary>
+        /// <param name="entry">The <see cref="HitObjectLifetimeEntry"/> controlling the lifetime of the <see cref="HitObject"/>.</param>
```
bdach, line 148: Docs wise I'd highlight the distinction here between the pooled/HO and non-pooled/DHO paths.

osu.Game/Rulesets/UI/HitObjectContainer.cs
```diff
@@ -1,78 +1,174 @@
...
+
+        #region Non-pooling support
+
+        public virtual void Add(DrawableHitObject hitObject)
+        {
+            bindStartTime(hitObject);
             AddInternal(hitObject);
+
+            hitObject.OnNewResult += onNewResult;
+            hitObject.OnRevertResult += onRevertResult;
```
bdach, line 106: Probably relatively minor, but is the ordering difference compared to the pooling path meaningful in any capacity? (pooling binds first then adds, non-pooling adds first then binds). Probably would read better to unify one way or another if possible.

osu.Game/Rulesets/UI/HitObjectContainer.cs
```diff
@@ -1,78 +1,174 @@
...
         {
+            lifetimeManager.ClearEntries();
+
             ClearInternal(disposeChildren);
-            unbindStartTimeMap();
+            unbindAllStartTimes();
         }
 
-        private void unbindStartTimeMap()
+        protected override bool CheckChildrenLife() => base.CheckChildrenLife() | lifetimeManager.Update(Time.Current, Time.Current);
```
bdach, line 146: I don't really like code like this (would prefer to have a three-liner method instead, which I assume was the intention here) but not sure if it's just me.

bdach (commented):
I've read through this - I have some notes but they're all minor. The core of this looks solid as far as I can tell.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -196,15 +199,15 @@ public void RegenerateAutoplay()
         /// <summary>
         /// Creates and adds drawable representations of hit objects to the play field.
         /// </summary>
-        private void loadObjects(CancellationToken? cancellationToken)
+        private void loadObjects(CancellationToken cancellationToken)
```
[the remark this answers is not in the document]
smoogipoo: Just cleanup, yep.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -230,23 +233,36 @@ public override void CancelResume()
             ResumeOverlay?.Hide();
         }
 
-        /// <summary>
-        /// Creates and adds the visual representation of a <typeparamref name="TObject"/> to this <see cref="DrawableRuleset{TObject}"/>.
-        /// </summary>
-        /// <param name="hitObject">The <typeparamref name="TObject"/> to add the visual representation for.</param>
-        private void addHitObject(TObject hitObject)
+        public void AddHitObject(TObject hitObject)
```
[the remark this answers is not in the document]
smoogipoo: Ah yes, this will be how the editor will add hitobjects. Will xmldoc.

osu.Game/Rulesets/UI/HitObjectContainer.cs
smoogipoo: Not meaningful at all, but I'll move the `AddInternal()` to the bottom.

osu.Game/Rulesets/UI/HitObjectContainer.cs
smoogipoo: Just so I understand, this is what you're suggesting?:

image

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -284,12 +300,15 @@ public override void SetReplayScore(Score replayScore)
             }
         }
 
+        public sealed override DrawableHitObject GetDrawableRepresentation(HitObject hitObject)
+            => base.GetDrawableRepresentation(hitObject) ?? CreateDrawableRepresentation((TObject)hitObject);
```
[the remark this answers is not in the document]
smoogipoo: I thought about it a bit, and decided to remove the override. I was hoping to make that path used more than it was, but it's really not possible to do that when DHO exposes its own `CreateNestedHitObject`.

So I've renamed it to `GetPoolableDrawableRepresentation()`, and now it only handles retrieving from a pool.

osu.Game/Rulesets/UI/Playfield.cs
smoogipoo: Rewritten it a tiny bit, please check if it's fine now.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -500,6 +525,94 @@ public HitWindows FirstAvailableHitWindows
...
+
+        /// <summary>
+        /// Registers a <see cref="DrawableHitObject"/> pool with this <see cref="DrawableRuleset"/> which is to be used whenever
+        /// <see cref="DrawableHitObject"/> representations are requested for the given <typeparamref name="TObject"/> type (via <see cref="GetDrawableRepresentation"/>).
+        /// </summary>
+        /// <param name="initialSize">The number of drawables to be prepared for initial consumption.</param>
+        /// <param name="maximumSize">An optional maximum size after which the pool will no longer be expanded.</param>
+        /// <typeparam name="TObject">The <see cref="HitObject"/> type.</typeparam>
+        /// <typeparam name="TDrawable">The <see cref="DrawableHitObject"/> receiver for <typeparamref name="TObject"/>s.</typeparam>
+        protected void RegisterPool<TObject, TDrawable>(int initialSize, int? maximumSize = null)
```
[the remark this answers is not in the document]
smoogipoo: In my latest changes, I've removed `PoolHitObjects` altogether. The intention here is someone can incrementally implement pooling for their ruleset on a per-`HitObject` basis.

The new process is calling `CreateDrawableRepresentation(HitObject)`, and if that returns `null` pooling is assumed.

osu.Game/Rulesets/UI/HitObjectContainer.cs
bdach: Yeah, that's what I had in mind.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -500,6 +525,94 @@ public HitWindows FirstAvailableHitWindows
...
+        }
+
+        /// <summary>
+        /// Creates the <see cref="DrawablePool{T}"/> to retrieve <see cref="DrawableHitObject"/>s of the given type from.
+        /// </summary>
+        /// <param name="initialSize">The number of hitobject to be prepared for initial consumption.</param>
+        /// <param name="maximumSize">An optional maximum size after which the pool will no longer be expanded.</param>
+        /// <typeparam name="TDrawable">The type of <see cref="DrawableHitObject"/> retrievable from this pool.</typeparam>
+        /// <returns>The <see cref="DrawablePool{T}"/>.</returns>
+        protected virtual DrawablePool<TDrawable> CreatePool<TDrawable>(int initialSize, int? maximumSize = null)
```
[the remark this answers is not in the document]
smoogipoo: I do want the user to have full control over the pool, but I've removed this and replaced it with a `RegisterPool(DrawablePool)` overload which provides even more flexibility for the user.

A little more implementation required osu!-side to get it to work but that's ok (https://github.com/smoogipoo/osu/commit/5dc5b5d8f83ea8d8458c4fe167a9fd2799a8d749).

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -101,13 +117,75 @@ private void load()
...
+        public virtual bool Remove(DrawableHitObject h)
+        {
+            if (!HitObjectContainer.Remove(h))
+                return false;
+
+            OnHitObjectRemoved(h.HitObject);
+            return false;
+        }
+
+        private readonly Dictionary<HitObject, HitObjectLifetimeEntry> lifetimeEntryMap = new Dictionary<HitObject, HitObjectLifetimeEntry>();
```
peppy, line 139: what's the purpose of this? do we need to keep references for some reason?

peppy (changes requested):

smoogipoo:
Latest changes reduce the number of state updates a tiny bit, for a slight perf improvement.

From:

image
image

To:

image
image

osu.Game/Rulesets/UI/Playfield.cs
smoogipoo: The reason is for the editor, because there's no way to retrieve the entries post-addition. Perhaps it could be exposed from `LifetimeEntryManager` as an `IEnumerable<>`. Regardless, it's not necessary in this PR and I'll remove.

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -289,7 +328,7 @@ public override void SetReplayScore(Score replayScore)
         /// </summary>
         /// <param name="h">The HitObject to make drawable.</param>
         /// <returns>The DrawableHitObject.</returns>
-        public abstract DrawableHitObject<TObject> CreateDrawableRepresentation(TObject h);
+        public virtual DrawableHitObject<TObject> CreateDrawableRepresentation(TObject h) => null;
```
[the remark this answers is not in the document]
smoogipoo: I've improved the documentation a bit and made it abstract again.

peppy (approved):

PR #10799, opened 2020-11-11
Add test for accuracy heatmap

nbvdkamp:
Added a test for the AccuracyHeatmap where the HitEvents have kind of realistic positions (rather than none at all).

image

### DIFF
--- a/osu.Game.Tests/Visual/Ranking/TestSceneStatisticsPanel.cs
+++ b/osu.Game.Tests/Visual/Ranking/TestSceneStatisticsPanel.cs
@@ -69,10 +69,10 @@ namespace osu.Game.Tests.Visual.Ranking
 
             for (int i = 0; i < 500; i++)
             {
-                float angle = (float)random.NextDouble() * 2 * (float)Math.PI;
-                float radius = (float)random.NextDouble() * 0.5f * OsuHitObject.OBJECT_RADIUS;
+                double angle = random.NextDouble() * 2 * Math.PI;
+                double radius = random.NextDouble() * 0.5f * OsuHitObject.OBJECT_RADIUS;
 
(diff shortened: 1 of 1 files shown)


osu.Game.Tests/Visual/Ranking/TestSceneStatisticsPanel.cs
```diff
@@ -44,5 +60,24 @@ public void TestNullScore()
...
+        private static List<HitEvent> createPositionDistributedHitEvents()
+        {
+            var hitEvents = new List<HitEvent>();
+            // Use constant seed for reproducibility
+            var random = new Random(0);
+
+            for (int i = 0; i < 500; i++)
+            {
+                double angle = random.NextDouble() * 2 * Math.PI;
+                double radius = random.NextDouble() * 0.5f * OsuHitObject.OBJECT_RADIUS;
```
bdach, line 73: for the record, this is not how you properly do uniform sampling over a unit sphere in polar coordinates. to get actual uniform sampling, the square root of the random factor in the radius has to be taken, or otherwise the samples are biased towards points closer to the centre.

that said this doesn't matter at all in this context so i've left it in.

bdach (approved):

osu.Game.Tests/Visual/Ranking/TestSceneStatisticsPanel.cs
nbvdkamp: Yeah I noticed that but I thought it makes sense to have it biased towards the center as real hits will also be

bdach: that is also part of the reason why i left it as is.

PR #10803, opened 2020-11-12
Add nested hitobject pooling

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10797

`DrawableHitObject` looks up nested hitobjects via `DrawableRuleset.GetPooledDrawableRepresentation` first, before falling back to their own `CreateNestedHitObject` implementation. Note that it's still desirable to implement `CreateNestedHitObject` for testing purposes - in cases where there is no `DrawableRuleset` such as skinning tests, you would still want to have nested hitobjects.

When used, it also requires `ClearNestedHitObjects` to call `Clear(false)` instead of `Clear()` which is done in `DrawableSlider`/`DrawableSpinner`, etc - basically never dispose the poolable drawables. Maybe this is something we can address later framework-side...

Accepting suggestions for the `OnParentReceived` method... I don't really like the name `OnParentApplied` as an alternative since it can be confused with occurring _after_ a parent's `Apply()` but this is not the case. The order of calls is currently:
```
Parent.Apply()
Nested.Apply()
Nested.OnApply() (virtual)
Nested.OnParentReceived() (virtual)
Parent.OnApply() (virtual)
```

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -19,6 +19,7 @@ using osu.Game.Rulesets.Objects.Types;
 using osu.Game.Rulesets.Scoring;
 using osu.Game.Skinning;
 using osu.Game.Configuration;
+using osu.Game.Rulesets.UI;
 using osuTK.Graphics;
 
 namespace osu.Game.Rulesets.Objects.Drawables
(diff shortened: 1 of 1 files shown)


smoogipoo:
I've also updated https://github.com/smoogipoo/osu/tree/osu-hitobject-pooling to take advantage of this.

peppy (approved):
seems fine for now, including naming.

PR #10804, opened 2020-11-12
Change osu! ruleset to use pooled hitobjects

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/10803
- [x] https://github.com/ppy/osu/pull/10823


smoogipoo:
Tests should be fixed, but they require the editor changes to be merged (added as prereq).

PR #10810, opened 2020-11-12
Prevent SelectionBox handles from appearing for stacks

nbvdkamp:
Stacks were able to be scaled, resulting in division by 0 and a crash when the BlueprintPiece Position property is updated with NaN values. This prevents stacked circles from being scaled and also rotated because they're all at the same position anyway.

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
+++ b/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
@@ -21,12 +21,12 @@ namespace osu.Game.Rulesets.Osu.Edit
         {
             base.OnSelectionChanged();
 
-            bool canOperate = EditorBeatmap.SelectedHitObjects.Count > 1 || EditorBeatmap.SelectedHitObjects.Any(s => s is Slider);
+            Quad quad = selectedMovableObjects.Length > 0 ? getSurroundingQuad(selectedMovableObjects) : new Quad();
 
-            SelectionBox.CanRotate = canOperate;
-            SelectionBox.CanScaleX = canOperate;
-            SelectionBox.CanScaleY = canOperate;
-            SelectionBox.CanReverse = canOperate;
+            SelectionBox.CanRotate = quad.Width > 0 || quad.Height > 0;
+            SelectionBox.CanScaleX = quad.Width > 0;
+            SelectionBox.CanScaleY = quad.Height > 0;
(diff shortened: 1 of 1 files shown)


bdach:
Seems correct but I can't reproduce this crashing, even on an existing map with perfect stacks.

osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
```diff
@@ -21,12 +21,12 @@ protected override void OnSelectionChanged()
...
+            Quad quad = selectedMovableObjects.Length > 0 ? getSurroundingQuad(selectedMovableObjects) : new Quad();
 
-            SelectionBox.CanRotate = canOperate;
-            SelectionBox.CanScaleX = canOperate;
-            SelectionBox.CanScaleY = canOperate;
-            SelectionBox.CanReverse = canOperate;
+            SelectionBox.CanRotate = quad.Width > 0 || quad.Height > 0;
+            SelectionBox.CanScaleX = quad.Width > 0;
+            SelectionBox.CanScaleY = quad.Height > 0;
+            SelectionBox.CanReverse = EditorBeatmap.SelectedHitObjects.Count > 1 || EditorBeatmap.SelectedHitObjects.Any(s => s is Slider);
```
bdach, line 29: Although on the other side I'd argue this should also probably be disabled for stacks.

nbvdkamp: I did it this way because this also handles spinners, which are not movable but can be reversed in time. I just noticed that that doesn't work in lazer but it does in stable.

bdach: I guess you can *theoretically* map a perfect stack that isn't temporally symmetrical, something like

2020-11-12-235800_717x806_scrot

but it looks super off and also not sure if that's a thing that's even valid/reasonable mapping...

nbvdkamp: This doesn't only apply to perfect stacks, stacks with offset as well.
Also say it is valid, [dum surfer](https://osu.ppy.sh/beatmapsets/716193#osu/1513180) is an example that comes to mind, and probably a lot of old maps with stacks.
image

bdach (approved):
fair enough

peppy (approved):

PR #10818, opened 2020-11-13
Fix combo break sound not playing a second time after rewinding before it

peppy:
- [x] Depends on #10817 for ease of merge
- Moved to frame stable context so the `GameplayClock` matches the incoming events (from combo bindable).

Should probably have tests, can add is it's deemed required.


bdach (approved):
seems correct

bdach:
This also weakly depends on #10819 for being able to easily test manually I suppose.

PR #10819, opened 2020-11-13
Add back setting to toggle "always play first combo break"

peppy:
I removed this in https://github.com/ppy/osu/pull/8802/commits/956980ee9055738a73dd9791ca52ff1a3de2ab18 but not 100% on my reasoning for doing so. Best have it exposed in the current state for now, I think.

Optimally this would have description or tooltip text that explains this better, but that will come with a larger refactor of settings (we don't really support this kind of text yet).


bdach (approved):
never a fan of more settings but given that this one already exists...

Joehuu (approved):

PR #10820, opened 2020-11-13
Fix perform from menu not hiding overlays if already on target screen

peppy:
Regressed with the recent changes to this component.

To test manually, try jumping to a beatmap using the beatmap listing while already at song select.


bdach (approved):
CI failure surely not related...?

PR #10823, opened 2020-11-13
Add pooling support to the editor

smoogipoo:
A few changes required here and I'm not sure how to feel about them so let me know if there's any suggestions.

`BlueprintContainer` has to retrieve hitobjects in two situations:
1. Hitobjects that are available immediately (all for non-pooled, all currently visible for pooled rulesets)
2. Hitobjects that become available as they become visible (pooled rulesets only).

`BlueprintContainer` uses `Playfield.AllHitObjects` to retrieve all hitobjects available immediately (1), and then binds the additional events `OnHitObjectUsageBegan` and `OnHitObjectUsageFinished` to retrieve hitobjects that come visible (2).

This does require `BlueprintContainer` take in `HitObjectComposer` as a ctor param, which is a breaking change across all rulesets.

### DIFF
--- a/osu.Game/Rulesets/UI/Playfield.cs
+++ b/osu.Game/Rulesets/UI/Playfield.cs
@@ -188,6 +188,41 @@ namespace osu.Game.Rulesets.UI
         }
 
         /// <summary>
+        /// Sets whether to keep a given <see cref="HitObject"/> always alive within this or any nested <see cref="Playfield"/>.
+        /// </summary>
+        /// <param name="hitObject">The <see cref="HitObject"/> to set.</param>
+        /// <param name="keepAlive">Whether to keep <paramref name="hitObject"/> always alive.</param>
+        public void SetKeepAlive(HitObject hitObject, bool keepAlive)
+        {
+            if (lifetimeEntryMap.TryGetValue(hitObject, out var entry))
+            {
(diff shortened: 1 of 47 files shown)


osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -183,6 +205,77 @@ protected virtual void OnHitObjectRemoved(HitObject hitObject)
         {
         }
 
+        /// <summary>
+        /// Sets whether to keep a given <see cref="HitObject"/> always alive within this or any nested <see cref="Playfield"/>.
+        /// </summary>
+        /// <param name="hitObject">The <see cref="HitObject"/> to set.</param>
+        /// <param name="keepAlive">Whether to keep <paramref name="hitObject"/> always alive.</param>
+        public void SetKeepAlive(HitObject hitObject, bool keepAlive)
```
peppy, line 213: can these potentially be internal?

peppy (changes requested):

osu.Game/Rulesets/UI/Playfield.cs
smoogipoo: Yup!

peppy (approved):

PR #10829, opened 2020-11-13
Add music bindings to on screen display

### LINKED ISSUE #6798: "No key bound" on play/pause key bind pressed
**Describe the bug:** 
I have bound Play/pause in the main menu to C ( like stable ), and when I press it, it will do as expected but it states that no key is bound to it.
image
**osu!lazer version:** 
2019.1106.0

Joehuu:
Resolves https://github.com/ppy/osu/issues/6798.

### DIFF
--- a/osu.Game/Input/KeyBindingStore.cs
+++ b/osu.Game/Input/KeyBindingStore.cs
@@ -37,7 +37,7 @@ namespace osu.Game.Input
         /// </summary>
         /// <param name="globalAction">The action to lookup.</param>
         /// <returns>A set of display strings for all the user's key configuration for the action.</returns>
-        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction globalAction)
+        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction? globalAction)
         {
             foreach (var action in Query().Where(b => (GlobalAction)b.Action == globalAction))
             {
--- a/osu.Game/Input/KeyBindingStore.cs
+++ b/osu.Game/Input/KeyBindingStore.cs
@@ -37,7 +37,7 @@ namespace osu.Game.Input
         /// </summary>
         /// <param name="globalAction">The action to lookup.</param>
         /// <returns>A set of display strings for all the user's key configuration for the action.</returns>
-        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction? globalAction)
+        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction globalAction)
         {
             foreach (var action in Query().Where(b => (GlobalAction)b.Action == globalAction))
             {
--- a/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
+++ b/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
@@ -6,7 +6,6 @@ using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Input.Bindings;
 using osu.Game.Beatmaps;
-using osu.Game.Configuration;
 using osu.Game.Input.Bindings;
 using osu.Game.Overlays.OSD;
 
@@ -26,9 +25,6 @@ namespace osu.Game.Overlays.Music
         [Resolved(canBeNull: true)]
         private OnScreenDisplay onScreenDisplay { get; set; }
 
-        [Resolved]
-        private OsuConfigManager config { get; set; }
-
         public bool OnPressed(GlobalAction action)
         {
             if (beatmap.Disabled)
@@ -41,11 +37,11 @@ namespace osu.Game.Overlays.Music
                     bool wasPlaying = musicController.IsPlaying;
 
                     if (musicController.TogglePause())
-                        onScreenDisplay?.Display(new MusicActionToast(wasPlaying ? "Pause track" : "Play track", config.LookupKeyBindings(action)));
+                        onScreenDisplay?.Display(new MusicActionToast(wasPlaying ? "Pause track" : "Play track", action));
                     return true;
 
                 case GlobalAction.MusicNext:
-                    musicController.NextTrack(() => onScreenDisplay?.Display(new MusicActionToast("Next track", config.LookupKeyBindings(action))));
+                    musicController.NextTrack(() => onScreenDisplay?.Display(new MusicActionToast("Next track", action)));
 
                     return true;
 
@@ -55,11 +51,11 @@ namespace osu.Game.Overlays.Music
                         switch (res)
                         {
                             case PreviousTrackResult.Restart:
-                                onScreenDisplay?.Display(new MusicActionToast("Restart track", config.LookupKeyBindings(action)));
+                                onScreenDisplay?.Display(new MusicActionToast("Restart track", action));
                                 break;
 
                             case PreviousTrackResult.Previous:
-                                onScreenDisplay?.Display(new MusicActionToast("Previous track", config.LookupKeyBindings(action)));
+                                onScreenDisplay?.Display(new MusicActionToast("Previous track", action));
                                 break;
                         }
                     });
@@ -76,8 +72,8 @@ namespace osu.Game.Overlays.Music
 
         private class MusicActionToast : Toast
         {
-            public MusicActionToast(string action, string shortcut)
-                : base("Music Playback", action, shortcut)
+            public MusicActionToast(string value, GlobalAction action)
+                : base("Music Playback", value, action: action)
             {
             }
         }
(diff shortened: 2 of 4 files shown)


osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
```diff
@@ -25,6 +26,9 @@ public class MusicKeyBindingHandler : Component, IKeyBindingHandler<GlobalAction
         [Resolved(canBeNull: true)]
         private OnScreenDisplay onScreenDisplay { get; set; }
 
+        [Resolved]
+        private OsuConfigManager config { get; set; }
```
bdach, line 30: Not sure whether I agree with having the config resolved in the handler and passing down the bindings, instead of resolving it in the toast itself and only passing down the action. It would reduce duplication somewhat (and you could even get rid of the `string action` parameter in all cases but play/pause).

It requires some jiggling in `Toast` to allow overwriting the shortcut text post-creation but I think I'd prefer that version.

peppy: Can agree with this. Also probably needs null checking on `LookupKeyBindings`.

peppy (changes requested):
As proposed

osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
Joehuu: Not sure if I did it right, so I have a diff here.

```diff
diff --git a/osu.Game/Configuration/OsuConfigManager.cs b/osu.Game/Configuration/OsuConfigManager.cs
index 89a6ee8b0..44ac48d83 100644
--- a/osu.Game/Configuration/OsuConfigManager.cs
+++ b/osu.Game/Configuration/OsuConfigManager.cs
@@ -196,7 +196,7 @@ public override TrackedSettings CreateTrackedSettings()

         public Func<int, string> LookupSkinName { private get; set; }

-        public Func<GlobalAction, string> LookupKeyBindings { get; set; }
+        public Func<GlobalAction?, string> LookupKeyBindings { get; set; }
     }

     public enum OsuSetting
diff --git a/osu.Game/Input/KeyBindingStore.cs b/osu.Game/Input/KeyBindingStore.cs
index bc73d74d7..f4b7c873d 100644
--- a/osu.Game/Input/KeyBindingStore.cs
+++ b/osu.Game/Input/KeyBindingStore.cs
@@ -37,7 +37,7 @@ public KeyBindingStore(DatabaseContextFactory contextFactory, RulesetStore rules
         /// </summary>
         /// <param name="globalAction">The action to lookup.</param>
         /// <returns>A set of display strings for all the user's key configuration for the action.</returns>
-        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction globalAction)
+        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction? globalAction)
         {
             foreach (var action in Query().Where(b => (GlobalAction)b.Action == globalAction))
             {
diff --git a/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs b/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
index 0d6158d46..e4903b348 100644
--- a/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
+++ b/osu.Game/Overlays/Music/MusicKeyBindingHandler.cs
@@ -41,11 +41,11 @@ public bool OnPressed(GlobalAction action)
                     bool wasPlaying = musicController.IsPlaying;

                     if (musicController.TogglePause())
-                        onScreenDisplay?.Display(new MusicActionToast(wasPlaying ? "Pause track" : "Play track", config.LookupKeyBindings(action)));
+                        onScreenDisplay?.Display(new MusicActionToast(wasPlaying ? "Pause track" : "Play track", action));
                     return true;

                 case GlobalAction.MusicNext:
-                    musicController.NextTrack(() => onScreenDisplay?.Display(new MusicActionToast("Next track", config.LookupKeyBindings(action))));
+                    musicController.NextTrack(() => onScreenDisplay?.Display(new MusicActionToast("Next track", action)));

                     return true;

@@ -55,11 +55,11 @@ public bool OnPressed(GlobalAction action)
                         switch (res)
                         {
                             case PreviousTrackResult.Restart:
-                                onScreenDisplay?.Display(new MusicActionToast("Restart track", config.LookupKeyBindings(action)));
+                                onScreenDisplay?.Display(new MusicActionToast("Restart track", action));
                                 break;

                             case PreviousTrackResult.Previous:
-                                onScreenDisplay?.Display(new MusicActionToast("Previous track", config.LookupKeyBindings(action)));
+                                onScreenDisplay?.Display(new MusicActionToast("Previous track", action));
                                 break;
                         }
                     });
@@ -76,8 +76,8 @@ public void OnReleased(GlobalAction action)

         private class MusicActionToast : Toast
         {
-            public MusicActionToast(string action, string shortcut)
-                : base("Music Playback", action, shortcut)
+            public MusicActionToast(string value, GlobalAction action)
+                : base("Music Playback", value, action: action)
             {
             }
         }
diff --git a/osu.Game/Overlays/OSD/Toast.cs b/osu.Game/Overlays/OSD/Toast.cs
index 1497ca8fa..a15984a81 100644
--- a/osu.Game/Overlays/OSD/Toast.cs
+++ b/osu.Game/Overlays/OSD/Toast.cs
@@ -1,11 +1,14 @@
 <EF><BB><BF>// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.

+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
+using osu.Game.Configuration;
 using osu.Game.Graphics;
 using osu.Game.Graphics.Sprites;
+using osu.Game.Input.Bindings;
 using osuTK;
 using osuTK.Graphics;

@@ -20,7 +23,10 @@ public abstract class Toast : Container

         protected readonly OsuSpriteText ValueText;

-        protected Toast(string description, string value, string shortcut)
+        [Resolved]
+        private OsuConfigManager config { get; set; }
+
+        protected Toast(string description, string value, string shortcut = null, GlobalAction? action = null)
         {
             Anchor = Anchor.Centre;
             Origin = Anchor.Centre;
@@ -29,6 +35,8 @@ protected Toast(string description, string value, string shortcut)
             RelativeSizeAxes = Axes.Y;
             AutoSizeAxes = Axes.X;

+            OsuSpriteText spriteText;
+
             InternalChildren = new Drawable[]
             {
                 new Container // this container exists just to set a minimum width for the toast
@@ -68,7 +76,7 @@ protected Toast(string description, string value, string shortcut)
                     Origin = Anchor.Centre,
                     Text = value
                 },
-                new OsuSpriteText
+                spriteText = new OsuSpriteText
                 {
                     Anchor = Anchor.BottomCentre,
                     Origin = Anchor.BottomCentre,
@@ -76,9 +84,15 @@ protected Toast(string description, string value, string shortcut)
                     Alpha = 0.3f,
                     Margin = new MarginPadding { Bottom = 15 },
                     Font = OsuFont.GetFont(size: 12, weight: FontWeight.Bold),
-                    Text = string.IsNullOrEmpty(shortcut) ? "NO KEY BOUND" : shortcut.ToUpperInvariant()
                 },
             };
+
+            if (action != null)
+                spriteText.Text = config.LookupKeyBindings(action);
+            else if (!string.IsNullOrEmpty(shortcut))
+                spriteText.Text = shortcut.ToUpperInvariant();
+            else
+                spriteText.Text = "NO KEY BOUND";
         }
     }
 }
```

> and you could even get rid of the string action parameter in all cases but play/pause

I don't think so. All but one toggle (`MusicNext`) have two functionalities. `MusicPrev` can restart or go to previous track, and `MusicPlay` as you've said.

osu.Game/Input/KeyBindingStore.cs
```diff
@@ -37,7 +37,7 @@ public KeyBindingStore(DatabaseContextFactory contextFactory, RulesetStore rules
         /// </summary>
         /// <param name="globalAction">The action to lookup.</param>
         /// <returns>A set of display strings for all the user's key configuration for the action.</returns>
-        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction globalAction)
+        public IEnumerable<string> GetReadableKeyCombinationsFor(GlobalAction? globalAction)
```
peppy, line 40: I don't think you need to change this. Just only do the lookup when the `GlobalAction` is non-null (in `MusicKeyBindingHandler`).

peppy (dismissed):

smoogipoo (approved):

PR #10830, opened 2020-11-13
Make tournament strings more consistent

Joehuu:
Mostly for https://github.com/ppy/osu-wiki/pull/4061.

Intention of formatting (went with the format that was most used):
- Labels are sentence cased (`Current IPC source` and `Stream area resolution`, so changed `Current User`)
- Button texts are sentence cased (`Change source` and `Set height`, so changed `Change Login`)
- `Login` changed to `sign-in` (some wiki thing that still is relevant, https://github.com/ppy/osu-wiki/pull/1853#discussion_r238080203 and https://osu.ppy.sh/help/wiki/Help_Centre#sign-in)
- Descriptions end with a period (self-explanatory)
- Screen buttons are title cased (self-explanatory)
- Avoid using `sign-in` where possible (doesn't seem common format?) (`a login` -> `signing in`)


bdach:
If this is happening, it should probably happen over the entire tournament client at once, not just the setup screen - some other screens, like the round editor screen, have similar inconsistencies in label texts, for example.

[Here's a rough list with what should be most user-facing texts](https://gist.github.com/bdach/1ac0c57cf0f0faf1c56351150ebdbe2e). I grepped for `\".+\"` and trimmed everything that was either not user-facing or part of the design to the naked eye, but I didn't manually check every single instance.

peppy:
I think it's fine to do this incrementally, the changes in this PR look okay.

peppy (approved):

PR #10831, opened 2020-11-13
Add back fade effect to legacy hold notes

### LINKED ISSUE #8594: missed hold note doesnt gray out
Describe the bug: in stable, hold notes (when released early, missed or not hold), it grays out (idk if same applies to regular notes. 
screenshot051

in lazer however it doesnt gray out
when not hold:
osu_2020-04-03_15-48-10

when holded, then released early then re hold:
osu_2020-04-03_15-40-21

Skin:
https://github.com/ppy/osu/files/4425570/my.skin.zip

 

osu!lazer version: 2020.403.0

**Logs:**
not needed, i think

bdach:
Resolves #8594.

- [x] Depends on #10828 to work properly during rewind.

Duration and fade colour should be matching stable.

I had the core of this written hours ago, but the fickle logic of `DrawableHoldNote` during rewind held me up for way longer... I had to change the `HasBroken` flag to a nullable time instant to be able to revert effects after rewinding.


peppy:
I was about to ask why it needed to be stored (just use judgements!) but then realised this time value is not yet stored anywhere in judgements. That sounds like an oversight which will need to be fixed eventually (since we really want all gameplay interactions to be stored to judgements). Like maybe the hold note itself should become missed at that point..

bdach:
From what I recall not missing the hold note immediately on dropping was a conscious decision - if a tick isn't missed somewhere along the way, dropping a hold for a bit only causes the tail to be judged as MEH.

peppy (commented):
Looks sane but deferring judgement on this.

bdach:
We'll probably want to tear out all the judgement-related time instants from `DrawableHoldNote` following the direction from #10844 in preparation for pooling, but I suppose it's not really a hard requirement for this to be merged. Either way I can commit to making such a change sometime this week.

osu.Game.Rulesets.Mania/Skinning/LegacyBodyPiece.cs
```diff
@@ -92,11 +102,25 @@ private void load(ISkinSource skin, IScrollingInfo scrollingInfo, DrawableHitObj
...
+            missFadeTime.BindValueChanged(onMissFadeTimeChanged, true);
+
+            holdNote.ApplyCustomUpdateState += applyCustomUpdateState;
+            applyCustomUpdateState(holdNote, holdNote.State.Value);
+        }
+
+        private void applyCustomUpdateState(DrawableHitObject hitObject, ArmedState state)
+        {
+            if (state == ArmedState.Miss)
+                missFadeTime.Value ??= hitObject.StateUpdateTime;
```
smoogipoo, line 123: Shouldn't this be `HitStateUpdateTime`? Also what case does this code block handle - doesn't everything go through HoldNoteBroken?

bdach: Yeah, it should be the hit variant, my mistake - however this *is* required in case the hold note head is never hit, as such a note doesn't really get broken. Open to suggestions how this could be done better.

smoogipoo (approved):
Looks to work well :+1:

PR #10812, opened 2020-11-12
Allow spinners to be reversed

nbvdkamp:
To match stable.

PR #10832, opened 2020-11-13
WIP: Exclude non-movable hitobjects from selection boxes

nbvdkamp:
As @bdach mentioned in #10812 resizing selections with spinners in them results in very strange behavior and UI wise it doesn't work well since the resize handles are in the very corners of the playfield and the other selection box buttons are off screen.

This PR attempts to fix that by excluding non-movable hitobjects from selection boxes (but not from the selection itself) except when they are the only thing selected. This allows moving the rest of the object as if the spinner wasn't selected.

Since spinners can't be moved anyway I think this seems like a fairly intuitive/useful way of handling this.
I noticed after I got this working that the code I modified in BlueprintContainer also handles movement in time so currently this breaks moving spinners in time, any tips on how this should be fixed properly are appreciated.

https://streamable.com/vx48wk

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
+++ b/osu.Game.Rulesets.Osu/Edit/OsuSelectionHandler.cs
@@ -232,7 +232,7 @@ namespace osu.Game.Rulesets.Osu.Edit
         /// <param name="points">The points to calculate a quad for.</param>
         private Quad getSurroundingQuad(IEnumerable<Vector2> points)
         {
-            if (!EditorBeatmap.SelectedHitObjects.Any())
+            if (!points.Any())
                 return new Quad();
 
             Vector2 minPosition = new Vector2(float.MaxValue, float.MaxValue);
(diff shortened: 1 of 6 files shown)


bdach:
Out of the gate I can't say I like how this looks visually to be honest. So when the spinner is added to selection, it is shown as selected on the playfield and sort of on the timeline (by having the yellow outline added onto it), but is not part of either the playfield selection box, or the timeline selection box? It's half selected but half not? Super confusing to get a good grasp of.

I really would advocate for going the simpler way of just disabling the drag handles and only allowing reverse if any spinner is part of the selection.

nbvdkamp:
I just checked and stable also allows moving objects while spinners are selected, I think it would be good to have a mappers input on whether this is desirable or not (though I do agree it would probably be a lot easier to just disable it).

nbvdkamp:
I say that because currently dragging a selection that has a spinner as first element gives weird (extremely sensitive) behaviour too:
https://streamable.com/aorxnp

peppy:
In lazer we intend to allow spinners to be moved, so this is fine.

peppy:
@nbvdkamp just a heads up that it's appreciated if you provide a reason when closing PRs so we can ascertain the path forward.

nbvdkamp:
I thought you meant that the current behaviour is fine, since spinners will be movable in the future. Should this be handled in a different way for now?

peppy:
Aha, right. Maybe the current logic is fine for the time being then..

PR #10838, opened 2020-11-14
Fix crash on shift+right-click deleting objects

nbvdkamp:
This regressed in #10767

```
Exception has occurred: CLR/System.InvalidOperationException
An exception of type 'System.InvalidOperationException' occurred in System.Private.CoreLib.dll but was not handled in user code: 'Collection was modified; enumeration operation may not execute.'
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at osu.Game.Screens.Edit.Compose.Components.BlueprintContainer.beginClickSelection(MouseButtonEvent e) in E:\dev\osuFork\osu\osu.Game\Screens\Edit\Compose\Components\BlueprintContainer.cs:line 341
   at osu.Game.Screens.Edit.Compose.Components.BlueprintContainer.OnMouseDown(MouseDownEvent e) in E:\dev\osuFork\osu\osu.Game\Screens\Edit\Compose\Components\BlueprintContainer.cs:line 139
   at osu.Framework.Input.ButtonEventManager`1.<>c__DisplayClass14_0.<PropagateButtonEvent>b__0(Drawable target)
   at System.Linq.Enumerable.TryGetFirst[TSource](IEnumerable`1 source, Func`2 predicate, Boolean& found)
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at osu.Framework.Input.ButtonEventManager`1.PropagateButtonEvent(IEnumerable`1 drawables, UIEvent e)
   at osu.Framework.Input.MouseButtonEventManager.HandleButtonDown(InputState state, List`1 targets)
   at osu.Framework.Input.ButtonEventManager`1.handleButtonDown(InputState state)
   at osu.Framework.Input.ButtonEventManager`1.HandleButtonStateChange(InputState state, ButtonStateChangeKind kind)
   at osu.Framework.Input.InputManager.HandleMouseButtonStateChange(ButtonStateChangeEvent`1 e)
   at osu.Framework.Input.UserInputManager.HandleInputStateChange(InputStateChangeEvent inputStateChange)
   at osu.Framework.Input.StateChanges.ButtonInput`1.Apply(InputState state, IInputStateChangeHandler handler)
   at osu.Framework.Input.InputManager.Update()
   at osu.Framework.Input.PassThroughInputManager.Update()
   at osu.Framework.Graphics.Drawable.UpdateSubTree()
   at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
   at osu.Framework.Platform.GameHost.UpdateFrame()
   at osu.Framework.Threading.GameThread.ProcessFrame()
```

### DIFF
--- a/osu.Game.Tests/Visual/Editing/TestSceneEditorQuickDelete.cs
+++ b/osu.Game.Tests/Visual/Editing/TestSceneEditorQuickDelete.cs
@@ -28,6 +28,9 @@ namespace osu.Game.Tests.Visual.Editing
 
         protected override IBeatmap CreateBeatmap(RulesetInfo ruleset) => new TestBeatmap(ruleset, false);
 
+        private BlueprintContainer blueprintContainer
+            => Editor.ChildrenOfType<BlueprintContainer>().First();
+
         [Test]
         public void TestQuickDeleteRemovesObject()
         {
@@ -39,7 +42,7 @@ namespace osu.Game.Tests.Visual.Editing
 
             AddStep("move mouse to object", () =>
             {
-                var pos = getBlueprintContainer.SelectionBlueprints
+                var pos = blueprintContainer.SelectionBlueprints
                             .ChildrenOfType<HitCircleSelectionBlueprint>().First()
                             .ChildrenOfType<HitCirclePiece>().First().ScreenSpaceDrawQuad.Centre;
                 InputManager.MoveMouseTo(pos);
@@ -73,7 +76,7 @@ namespace osu.Game.Tests.Visual.Editing
 
             AddStep("move mouse to controlpoint", () =>
             {
-                var pos = getBlueprintContainer.SelectionBlueprints
+                var pos = blueprintContainer.SelectionBlueprints
                             .ChildrenOfType<SliderSelectionBlueprint>().First()
                             .ChildrenOfType<PathControlPointVisualiser>().First()
                             .ChildrenOfType<PathControlPointPiece>().ElementAt(1).ScreenSpaceDrawQuad.Centre;
@@ -85,9 +88,5 @@ namespace osu.Game.Tests.Visual.Editing
 
             AddAssert("slider has 2 points", () => slider.Path.ControlPoints.Count == 2);
         }
-
-        private BlueprintContainer getBlueprintContainer => Editor.ChildrenOfType<ComposeScreen>().First()
-                                                                  .ChildrenOfType<HitObjectComposer>().First()
-                                                                  .ChildrenOfType<BlueprintContainer>().First();
     }
 }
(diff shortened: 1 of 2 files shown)


bdach:
The logic in that PR very subtly changed. `beginClickSelection` would exit out of the loop at the first hovered blueprint, regardless of the value returned by `HandleSelectionRequested`. Any reason to not just revert that behaviour, instead of making a list copy?

Also I don't think we should be letting any fix in without test coverage. This has clearly proved too brittle to leave untested.

bdach (dismissed):
as above re: test coverage

nbvdkamp:
How should this be tested? I got it to 'work' in an EditorTestScene like this
```C#
        [Test]
        public void TestQuickDeleteRemovesObject()
        {
            var addedObject = new HitCircle { StartTime = 1000 };

            AddStep("add hitobject", () => EditorBeatmap.Add(addedObject));

            AddStep("select added object", () => EditorBeatmap.SelectedHitObjects.Add(addedObject));

            AddStep("move mouse to object", () => InputManager.MoveMouseTo(new Vector2(870,300)));
            AddStep("hold shift", () => InputManager.PressKey(Key.ShiftLeft));
            AddStep("rightclick", () => InputManager.Click(MouseButton.Right));
            AddStep("release shift", () => InputManager.ReleaseKey(Key.ShiftLeft));

            AddAssert("no hitobjects in beatmap", () => EditorBeatmap.HitObjects.Count == 0);
        }
```
But of course hardcoding screenspace positions is wrong (and breaks if the resolution changes).
The problem is that accessing the SelectionBlueprint for the circle is seemingly impossible without making a lot of things public that shouldn't be. (it's in a BlueprintContainer in a HitObjectComposer in the ComposeScreen in the Editor)

bdach:
You can use the helper method `.ChildrenOfType<T>()` to access a child drawable without exposing as `public`, and then use `ScreenSpaceDrawQuad` on it to get its center. There are many uses of this pattern across visual tests.

bdach:
I've taken the liberty of pushing in a few changes here to bring this to a hopefully mergeable state.

@nbvdkamp A few notes:

* Generally the convention is that properties are named like fields, and methods like methods - so it's either `blueprintContainer` as a property, or `getBlueprintContainer()` as a method - not `getBlueprintContainer` as sort of an in-between.
* `ChildrenOfType<T>` works recursively, so there is no need to traverse the hierarchy one level at a time. That is stated in the xmldoc, but I should have also mentioned this.
* I've also changed the fix logic a little bit - I might have misled by suggesting to "revert to the previous behaviour"; I meant just the part that returned on the first hovered drawable, not just straight up to the old logic, as it sort of didn't make sense (why return `true` from `beginClickSelection` if nothing is hovered?)

nbvdkamp:
Thanks for the fixes, I only just figured out why InspectCode wasn't working for me locally. (Turns out VS2019 being installed and its  MSBuild being used was the problem)

peppy:
Added one more check step just for safety/completionist sake.

peppy (approved):

PR #10840, opened 2020-11-14
Add null checks to unguarded resolved-as-possibly-null fields

bdach:
One case was [reported in discord](https://discord.com/channels/188630481301012481/188630652340404224/777156928765689866), but I figured I'd do a full sweep and caught a few more. A few are relatively recent, actually.

I would love for this to be enforced by CI, but it doesn't seem feasible without a roslyn analyser. I gave it a meek attempt at making framework also accept the `JetBrains.Annotations.CanBeNull` attribute as indication that nulls are permitted, but it's not really open to extensibility in that regard...

2020-11-14-151921_536x182_scrot

### DIFF
--- a/osu.Game.Tournament/Screens/Ladder/Components/DrawableTournamentMatch.cs
+++ b/osu.Game.Tournament/Screens/Ladder/Components/DrawableTournamentMatch.cs
@@ -144,9 +144,9 @@ namespace osu.Game.Tournament.Screens.Ladder.Components
                 if (selected)
                 {
                     selectionBox.Show();
-                    if (editor)
+                    if (editor && editorInfo != null)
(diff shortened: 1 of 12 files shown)


nbvdkamp:
Just encountered another one on this line:
https://github.com/ppy/osu/blob/aae59dc3cf0e1b518d1b6872de1162eb22457d74/osu.Game/Screens/Edit/Compose/Components/SelectionHandler.cs#L113
It crashes if you scale/rotate a selection in the test scenes for ComposeScreen and HitObjectComposer. 
The OnOperationEnded needs a check too.

bdach:
That's what happens when you try to check 100-odd cases manually... Thanks, will cover in a sec.

peppy (approved):

PR #10841, opened 2020-11-14
Add support for previewing tracks on spectator screen

### LINKED ISSUE #10836: Game crashes in the spectator mode screen when you press on the song preview button
https://youtu.be/IBgLr-hgOAA

**osu!lazer version:** 
2020.1114.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5540637/runtime.log)
[network.log](https://github.com/ppy/osu/files/5540638/network.log)
[performance.log](https://github.com/ppy/osu/files/5540639/performance.log)

**Computer Specifications:** 
OS: windows 10
CPU: i5-4440
iGPU: intel hd graphics 4600
RAM: 8 GB

bdach:
Resolves #10836.

Seems like a useful change on its own as players might not know absolutely every map they see anyway. There are no tests here because it's quite cumbersome to test (would have to rely on online preview track retrieval to succeed), but this can be semi-manually tested using the test scene and the right steps, and in the game, obviously. Will try to figure out an automated test on request.

I also removed one container clear in `showBeatmapPanel()` as it looks redundant (only call to method always unconditionally follows a `clearDisplay()`, which clears anyway).

### DIFF
--- a/osu.Game.Tests/Visual/Components/TestScenePreviewTrackManager.cs
+++ b/osu.Game.Tests/Visual/Components/TestScenePreviewTrackManager.cs
@@ -167,6 +167,21 @@ namespace osu.Game.Tests.Visual.Components
             AddAssert("game not muted", () => audio.Tracks.AggregateVolume.Value != 0);
         }
 
+        [Test]
+        public void TestOwnerNotRegistered()
+        {
+            PreviewTrack track = null;
+
+            AddStep("get track", () => Add(new TestTrackOwner(track = getTrack(), registerAsOwner: false)));
+            AddUntilStep("wait for loaded", () => track.IsLoaded);
+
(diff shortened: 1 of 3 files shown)


peppy:
Hmm, I guess this is fine to add locally but a bit of a hard one to predict failures for..

Would be nice if the error happened before the preview was actually attempted to be played.

Or just make it nullable to avoid hard crashes (might be preferable).

peppy:
Made some changes to avoid the crash in the first place. @bdach please do a quick check.

peppy (dismissed):

bdach:
Seems like a fair change, but given `PreviewTrackManager` has pretty good test coverage, I've added a test case for this new behaviour.

peppy (approved):

PR #10844, opened 2020-11-14
Fix rotation tracker state not being reset on drawable spinner re-use

### LINKED ISSUE #10843: Spinning a spinner till the end keeps the points for the next spinner
Describe the bug: Points you get for a spinner are kept for the next spinner if the spinner was spinned until it ends. (and probably continued spinning because the first spinner failed)

Screenshots or videos showing encountered issue: https://youtu.be/Adr9ge-ZRi4

osu!lazer version: 2020.1114.0

Logs: no logs

bdach:
Aims to resolve #10843.

This is a draft-ish, but working PR - draft-ish mostly because the solution in here has not been done anywhere else up to now. https://github.com/ppy/osu/pull/10831 has a similar problem, but works around it now, but if the direction of this PR is deemed correct, I will reapply it there.

The main reason why spinners broke was, of course, the re-use of drawables which had gameplay/scoring state inside of them, namely completion progress-related data in `RateAdjustedRotation` and `Complete`. The immediate solution would be to just create a new one on every application, but this is trickier than it might initially seem due to `Complete` being bindable and bound in places.

As per [this comment](https://github.com/ppy/osu/pull/10831#issuecomment-727117638), I figured I'd try the way of moving the important gameplay-related data out into the judgement results. There's *sort* of a precedent to adding data to results - hit circles store the hit position - and it felt like a better place than the judgement itself.

(As an aside, note that this is an extremely similar problem to the aforementioned PR. We can't rely on the actual judgement result being applied, as it is only applied at the end of the spinner - similarly to how breaking a hold note only has a tangible effect in the way of state changes when it actually reaches its end time.)

After moving the important bits out to the judgements, I had a bit of a toss-up what to do with the rest - for full safety things like `Rotation` and all of those other flags have to be cleared, lest the spinner behave weird for a frame after re-application or not be upright the next time around. At this point you could probably just as well justify recreating (the allocation cost sucks a bit, but spinners are far and few in between), but I just cleared manually. Can change on request.

A few side notes:

* I added a change that clears the DHO's result on `free()` in f8cabbd, as there was potential for the result to remain there and not be replaced on re-application. It wouldn't happen in actual gameplay, as if there's a `HitObjectLifetimeEntry` then the application will be unconditiional, but it did happen in tests. Not sure if that was an oversight or intentional; @smoogipoo - please advise!
* On another note, the `Update()`-backed bindables in the skinnable elements are a bit funky, but when I tried to make `TimeCompleted` a bindable in the result, it quickly became apparent that following actual bindable flow when pooling is involved quickly balloons out of control. The proper flow, as far as I can tell, would require to:

  * subscribe to `HitObjectApplied` and `HitObjectFreed` (the latter of which doesn't exist yet),
  * bind in the subscription to the first and unbind in the subscription to the second,
  * add unregisters in `Dispose()` to both,

  which is a hell of a lot of boilerplate code to have to do every time.

I reckon this PR might not go far enough in places (i.e. maybe `Progress` should be in the judgement too? maybe the skinnables should be refactored somehow?), but the intention was to get it out as an MVP of a fix to a pretty critical issue.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerApplication.cs
+++ b/osu.Game.Rulesets.Osu.Tests/TestSceneSpinnerApplication.cs
@@ -15,7 +15,7 @@ namespace osu.Game.Rulesets.Osu.Tests
     public class TestSceneSpinnerApplication : OsuTestScene
     {
         [Test]
-        public void TestApplyNewCircle()
+        public void TestApplyNewSpinner()
         {
             DrawableSpinner dho = null;
 
@@ -23,11 +23,14 @@ namespace osu.Game.Rulesets.Osu.Tests
             {
                 Position = new Vector2(256, 192),
                 IndexInCurrentCombo = 0,
-                Duration = 0,
+                Duration = 500,
             }))
             {
                 Clock = new FramedClock(new StopwatchClock())
             });
+
+            AddStep("rotate some", () => dho.RotationTracker.AddRotation(180));
+            AddAssert("rotation is set", () => dho.RotationTracker.RateAdjustedRotation == 180);
 
             AddStep("apply new spinner", () => dho.Apply(prepareObject(new Spinner
             {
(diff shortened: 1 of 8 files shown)


peppy:
I think this seems like a logical path forward, and looks sane on a code check.

peppy (approved):

smoogipoo (approved):
lgtm

PR #10846, opened 2020-11-15
Make EZ mod able to fail in Taiko

### LINKED ISSUE #10662: EZ makes it impossible to fail a map in taiko
**Describe the bug:** Upon failing a taiko map with EZ on (failing is only possible in the very end of the map) it instantly restores you full HP (as if you have an extra life) and counts it as a pass. EZ customization also has a redundant (for taiko) extra lives slider.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=lFMJMNoyn9Q

PercyDan54:
Attempts to resolve #10662


osu.Game.Rulesets.Taiko/Mods/TaikoModEasy.cs
```diff
@@ -1,12 +1,40 @@
...
+using System;
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Beatmaps;
+using osu.Game.Graphics;
 using osu.Game.Rulesets.Mods;
 
 namespace osu.Game.Rulesets.Taiko.Mods
 {
-    public class TaikoModEasy : ModEasy
+    public class TaikoModEasy : Mod, IApplicableToDifficulty, IApplicableFailOverride
```
bdach, line 12: Removing inheritance from `ModEasy` will break replay encoding (see `Ruleset.ConvertToLegacyMods`). I will be adding test coverage for that in another PR.

osu.Game.Rulesets.Taiko/Mods/TaikoModEasy.cs
```diff
@@ -1,12 +1,40 @@
...
     {
-        public override string Description => @"Beats move slower, less accuracy required, and three lives!";
+        public override string Name => "Easy";
+        public override string Acronym => "EZ";
+        public override string Description => @"Beats move slower, less accuracy required";
```
bdach, line 16: for consistency with the old one

```suggestion
        public override string Description => @"Beats move slower, and less accuracy required!";
```

bdach (dismissed):

bdach:
isn't the useless "life count" customisable mod setting going to come back if you just revert the inheritance as you did? i thought that was the entire point of splitting.

bdach:
@PercyDan54 duplicating the setting in three rulesets was not the correct way to do things, either. please see my latest commit for a better way to do this in the future.

osu.Game.Rulesets.Taiko/Mods/TaikoModEasy.cs
```diff
@@ -7,6 +7,6 @@ namespace osu.Game.Rulesets.Taiko.Mods
 {
     public class TaikoModEasy : ModEasy
     {
-        public override string Description => @"Beats move slower, less accuracy required, and three lives!";
+        public override string Description => @"Beats move slower, less accuracy required!";
```
Joehuu, line 10: Still no "and".
```suggestion
        public override string Description => @"Beats move slower, and less accuracy required!";
```

bdach: yeah, figured i'd leave it as is at that point, but sure

bdach (approved):

PR #10850, opened 2020-11-15
Clamp selection movement instead of refusing to move on playfield borders

nbvdkamp:
Prevents selection from getting 'stuck' on the borders of the playfield and lets it smoothly slide along them instead.
The old behaviour matches stable but I think this works much nicer UI wise.

https://streamable.com/qvw935


bdach (approved):
Seems fine. Can't find any way to break this, and in some respects it works even better than stable does (i.e. stable editor seems to block on the right and bottom sides of the playfield like the replaced lazer implementation used to)

peppy:
Feels much better

peppy (approved):

PR #10851, opened 2020-11-15
Fix transform mod using wrong position value to move to

### LINKED ISSUE #10848: Auto mod bugged when using Transform
Duplicate of #10847 because I forgot to add labels, but turns out I can’t set any.

**Describe the bug:** When playing back a beatmap with said mods, the cursor somehow clicks way out of place, and looks like if it’s doing some sort of “hard rock” (if you know what I mean) sometimes it hits the circles, sometimes it doesn’t. 

**Screenshots or videos showing encountered issue:** https://youtu.be/E6hD5l59IOE

**osu!lazer version:** 2020.1114.0 (TestFlight build on iPhone 12, iOS 14.2 [18B92])

**Logs:** no logs

PercyDan54:
Resolves #10848

### DIFF
--- a/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModTransform.cs
@@ -43,7 +43,7 @@ namespace osu.Game.Rulesets.Osu.Mods
 
                     float appearDistance = (float)(hitObject.TimePreempt - hitObject.TimeFadeIn) / 2;
 
-                    Vector2 originalPosition = drawable.Position;
+                    Vector2 originalPosition = hitObject.StackedPosition;
                     Vector2 appearOffset = new Vector2(MathF.Cos(theta), MathF.Sin(theta)) * appearDistance;
 
                     // the - 1 and + 1 prevents the hit objects to appear in the wrong position.


bdach:
I confirmed that the issue on `master` was caused by the mod reading old stale position values from pooled drawables, and as such this patch is the correct fix for that, as it sources from the applied hitobject instead of the not-yet-updated DHO.

bdach (dismissed):

bdach:
I take that back. Something really bad is happening with reads of `Position` and it might be indicative of a larger issue...

```
[runtime] 2020-11-15 17:20:16 [verbose]: position set to (113  65) on DHO #8344864
[runtime] 2020-11-15 17:20:16 [verbose]: mod read (113  65) from dho #8344864
[runtime] 2020-11-15 17:20:17 [verbose]: mod read (118  224) from dho #65711748
(...)
[runtime] 2020-11-15 17:20:18 [verbose]: position set to (203  49) on DHO #8344864
[runtime] 2020-11-15 17:20:18 [verbose]: mod read (113  65) from dho #8344864
[runtime] 2020-11-15 17:20:18 [debug]: Pressed (LeftButton) handled by DrawableHitCircle+HitReceptor.
[runtime] 2020-11-15 17:20:18 [verbose]: mod read (291  71) from dho #59914551
[runtime] 2020-11-15 17:20:18 [verbose]: position set to (231  133) on DHO #22634641
```

bdach:
This is caused by a larger issue in the drawable hitobject pooling mechanism and as such should not be fixed locally as done here. Sorry for the confusion.

PR #10859, opened 2020-11-16
Avoid updating score totals when at the results screen

### LINKED ISSUE #767: Do not read/send individual player score (text-based IPC) on the RankingScreen
**Describe the bug:** 
This could very well be a stable issue.  
Usually this happens on the very last map of the match when players sometimes quit immediately after the map resulting in score jumps (see the clip below) Potential fix would be to stop updating the score after the gameplay phase is over (lazer) or do not send the updated score of 0 if the player has left (stable)

https://clips.twitch.tv/JazzyHeartlessSardineBloodTrail

### LINKED ISSUE #10854: Do not read/send individual player score (text-based IPC) on the RankingScreen
**Describe the bug:** 
This could very well be a stable issue.  
Usually this happens on the very last map of the match when players sometimes quit immediately after the map resulting in score jumps (see the clip below) Potential fix would be to stop updating the score after the gameplay phase is over (lazer) or do not send the updated score of 0 if the player has left (stable)

https://clips.twitch.tv/JazzyHeartlessSardineBloodTrail

peppy:
Closes #10854.

### DIFF
--- a/osu.Game.Tournament.Tests/Components/TestSceneMatchScoreDisplay.cs
+++ b/osu.Game.Tournament.Tests/Components/TestSceneMatchScoreDisplay.cs
@@ -1,8 +1,11 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
+using System.Linq;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
+using osu.Framework.Testing;
 using osu.Framework.Utils;
 using osu.Game.Tournament.IPC;
 using osu.Game.Tournament.Screens.Gameplay.Components;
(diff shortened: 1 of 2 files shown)


smoogipoo:
This only works if all clients enter the ranking screen simultaneously, because `TourneyState.Ranking` is set as soon as one client enters the ranking screen (stable). Means in cases like users falling behind and the tourney client skipping them from realtime playback potentially won't get their scores updated after all other players reach the ranking screen.

Is this really what you want? Isn't it better for this to be a stable-side fix? Something like on a per-client level only update score if there's a user in the slot.

peppy:
Are you sure about that? That means the lazer is potentially marking the incorrect winner as one player finished but others are still going. We should definitely be *also* fixing this stable-side if that is the case.

peppy:
Closing as an incorrect solution. Will look at whether the original issue can be resolved at stable's end.

PR #10860, opened 2020-11-16
Use `dotnet tool` for InspectCode build script

ekrctb:
Same as ppy/osu-framework#4002 but fortunately no additional warnings are generated by the version difference of ReSharper.

I wonder if `cake` is necessary if the script is this simple.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -13,6 +13,24 @@
       "commands": [
         "dotnet-format"
       ]
+    },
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):
We have discussed removing cake in the past, and just having a code inspection script to run CFS + `dotnet format` + inspectcode.

peppy:
Definitely agree with dropping cake if that's something we can do. I thought we had an issue open for that even.

PR #10861, opened 2020-11-16
Rewrite user lookup request logic to be safer

### LINKED ISSUE #10852: Crash after entering the "currently playing" tab
https://youtu.be/gL_3urGVWlI

**osu!lazer version:** 
2020.1114.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/5542552/database.log)
[network.log](https://github.com/ppy/osu/files/5542553/network.log)
[performance.log](https://github.com/ppy/osu/files/5542554/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5542555/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5542556/updater.log)

**Computer Specifications:** 
OS: windows 10
CPU: i5-4440
iGPU: intel hd graphics 4600
RAM: 8 GB

peppy:
Hopefully this makes the whole process a lot easier to mentally parse.  Also fixes the case where users may have been added to the pending lookup list *after* the length check (due to the lock not being held over the task execution previously).

Closes https://github.com/ppy/osu/issues/10852.


osu.Game/Database/UserLookupCache.cs
```diff
@@ -47,70 +40,109 @@ private Task<List<User>> getQueryTaskForUser(int userId)
...
-                    lookupItems = nextTaskIDs.ToArray();
-                    nextTaskIDs.Clear();
-
-                    if (lookupItems.Length == 0)
-                    {
-                        queueNextTask(null);
-                        return new List<User>();
-                    }
-                }
+                    var dequeued = tasks.Dequeue();
```
smoogipoo, line 54: This code probably needs to be locked? This continuation is going to run on the global threadpool is it not?

smoogipoo (changes requested):

osu.Game/Database/UserLookupCache.cs
peppy: oh good point, i didn't even consider the queue's thread safety here (although i believe it won't fail on a dequeue since it's guaranteed to have an item).

do you prefer locking or `ConcurrentQueue`?

smoogipoo: Are you still going to have to lock the rest of the code? If so then I'd prefer locking to keep it simple and one-pattern. Probably extract the callback into a separate method while you're at it.

smoogipoo:
This seems very complicated, and I'm a little scared to be merging this when I'm imagining potential issues like ... If a user batch comes in during before `ContinueWith` does its thing, there's potentially going to be that many requests firing off because of the `if (wasRun) return false;` check.

Would you be against/do you see anything wrong with simplifying it the following way (i.e. is there a case I'm missing?):

```csharp
protected override async Task<User> ComputeValueAsync(int lookup, CancellationToken token = default)
    => await queryUser(lookup);

private readonly List<(int id, TaskCompletionSource<User>)> pendingUserTasks = new List<(int, TaskCompletionSource<User>)>();
private Task pendingRequestTask;
private readonly object taskAssignmentLock = new object();

private Task<User> queryUser(int userId)
{
    lock (taskAssignmentLock)
    {
        var tcs = new TaskCompletionSource<User>();

        // Add to the queue.
        pendingUserTasks.Add((userId, tcs));

        // Create a request task if there's not already one.
        if (pendingRequestTask == null)
            createNewTask();

        return tcs.Task;
    }
}

private void performLookup()
{
    var userTasks = new List<(int id, TaskCompletionSource<User> task)>();

    // Grab at most 50 users from the queue.
    lock (taskAssignmentLock)
    {
        while (pendingUserTasks.Count > 0 && userTasks.Count < 50)
        {
            userTasks.Add(pendingUserTasks[^1]);
            pendingUserTasks.RemoveAt(pendingUserTasks.Count - 1);
        }
    }

    // Query the users.
    var request = new GetUsersRequest(userTasks.Select(t => t.Item1).ToArray());

    // rather than queueing, we maintain our own single-threaded request stream.
    api.Perform(request);

    // Create a new request task if there's still more users to query.
    lock (taskAssignmentLock)
    {
        pendingRequestTask = null;
        if (pendingUserTasks.Count > 0)
            createNewTask();
    }

    // Notify of completion.
    foreach (var (id, task) in userTasks)
        task.SetResult(request.Result?.Users?.FirstOrDefault(u => u.Id == id));
}

private void createNewTask() => pendingRequestTask = Task.Run(performLookup);
```

peppy:
That simplification will potentially lookup the same user multiple times, not sure about that.

I'm not sure I follow your logic about the "before continuewith" thing. I can't see an issue on re-checking multiple times. Are you maybe missing the fact that new users are added to the last task in the queue?

smoogipoo:
> That simplification will potentially lookup the same user multiple times

I think adding a `CheckExists()` call inside the 50 user retrieval would fix that.

```csharp
// Grab at most 50 users from the queue.
lock (taskAssignmentLock)
{
    while (pendingUserTasks.Count > 0 && userTasks.Count < 50)
    {
        (int id, TaskCompletionSource<User> task) next = pendingUserTasks[^1];

        pendingUserTasks.RemoveAt(pendingUserTasks.Count - 1);

        // Perform a secondary check for existence, in case the user was queried in a previous batch.
        if (CheckExists(next.id, out var existing))
            next.task.SetResult(existing);
        else
            userTasks.Add(next);
    }
}
```

But your code wouldn't handle inter-request deduping either, right? 

> Are you maybe missing the fact that new users are added to the last task in the queue

Yeah, I was missing that. It should be fine in that case, but still speaks towards the complexity of this.

peppy:
PR your way if you feel strongly about it and we can compare. It might work out better but I'd need to check both next to each other. I had it written in a similar way while working on this PR but ended up with what I had because it felt easier to read.

PR #10867, opened 2020-11-16
Refactor user caching to improve readability and fix threadsafety issues

### LINKED ISSUE #10852: Crash after entering the "currently playing" tab

smoogipoo:
Closes https://github.com/ppy/osu/issues/10852
Supersedes and closes https://github.com/ppy/osu/pull/10861

With the superseded PR, I found it very hard to wrap my head around:
1. Request queues and nested user queues.
2. Task continuations.
3. The nested locking.

So this PR attempts to simplify this logic - it stores one queue of `(userId, TaskCompletionSource)` which can grow indefinitely even if there's duplicates, and the `TaskCompletionSource` for each is what's awaited on by the cache. Requests grab as many as 50 users from this queue that have not been fulfilled/are not duplicates.

### DIFF
--- a/osu.Game/Database/UserLookupCache.cs
+++ b/osu.Game/Database/UserLookupCache.cs
@@ -1,7 +1,6 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


peppy:
Seems pretty good readability wise.

Why not make `pendingUserTasks` a queue? It would save the manual `RemoveAt` etc.

Also the way you have it you are running the newest queries first, rather than the oldest (also would be fixed by using standard queue operations).

peppy (changes requested):
as proposed

osu.Game/Database/UserLookupCache.cs
```diff
@@ -15,103 +14,73 @@ namespace osu.Game.Database
...
-                    // append the new request on to the last to be executed.
-                    pendingRequest = pendingRequest.ContinueWith(_ => work());
-                    pendingRequestConsumedIDs = false;
+                    (int id, TaskCompletionSource<User> task) next = pendingUserTasks.Dequeue();
+
+                    // Perform a secondary check for existence, in case the user was queried in a previous batch.
+                    if (CheckExists(next.id, out var existing))
+                        next.task.SetResult(existing);
+                    else
+                        userTasks.Add(next);
```
peppy, line 61: Not 100% on the necessity of the `CheckExisting` call you have, but there's also of requesting the same user in the lookup twice (which was handled on my PR).

To fix this, we could either:

- Store the IDs in a separate `HashSet` alongside `userTasks` (ensures always sending a request for 50 users).
- call Distinct in the `userTasks.Select(t => t.id)` below (simpler to implement, but requests may be short of 50. probably doesn't matter)

leaning towards the second option.

peppy (changes requested):

osu.Game/Database/UserLookupCache.cs
smoogipoo: At the point where we're calling `Distinct`, we may as well just store a temporary hashset along with the main list and keep the up-to-50 users. Fine with that?

peppy: ok

bdach:
I like this version a lot as written (forgot `TaskCompletionSource` was a thing, which is a big shame, as it's really useful...) About the only thing I'm not 100% about is this potentially-quadratic double loop:

https://github.com/ppy/osu/blob/85b0f714670215fc85eec8215f393674ccfd062a/osu.Game/Database/UserLookupCache.cs#L85-L87

although given that the user count is capped to 50 maybe not really a huge deal (2.5k accesses to elements in the worst case, barring pathological cases with multiple duplicates).

peppy:
yeah, considered suggesting a dictionary to avoid that, but it comes at added complexity for a relatively uncommon execution path.

peppy:
Have implemented using dictionary, @smoogipoo cherry pick if you agree with it i guess.

https://github.com/peppy/osu/commit/009d666241fb7adfd1c1f11d7603478dbd9166b0

peppy (approved):

PR #10872, opened 2020-11-16
Fix triangles disappearing after a while

### LINKED ISSUE #10858: Little to no triangles displayed on osu! hit objects after a while
**Describe the bug:** Can happen on spinners too, so linking https://osu.ppy.sh/beatmapsets/90935#osu/260177 (Tewi Spinners) as the spinner circle is stuck in the center. Happens in gameplay and replay after a certain time during the map (after 27 seconds on this map). Doesn't seem to happen on taiko's objects, so I think it's something with pooling.

Ref: https://osu.ppy.sh/comments/1314650

C1A4cWzvkY

Last version:
AeoKkhsYx0

**osu!lazer version:** 2020.1114.0

**Logs:** not needed

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10858

Apart from spinners, it also occurs with normal hitobjects where their explosion's triangles stop showing up after a little while.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/CirclePiece.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/CirclePiece.cs
@@ -13,6 +13,11 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 {
     public class CirclePiece : CompositeDrawable
     {
+        [Resolved]
+        private DrawableHitObject drawableObject { get; set; }
+
+        private TrianglesPiece triangles;
+
         public CirclePiece()
         {
             Size = new Vector2(OsuHitObject.OBJECT_RADIUS * 2);
@@ -26,7 +31,7 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
         }
 
         [BackgroundDependencyLoader]
-        private void load(TextureStore textures, DrawableHitObject drawableHitObject)
+        private void load(TextureStore textures)
         {
             InternalChildren = new Drawable[]
             {
@@ -36,13 +41,32 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
                     Origin = Anchor.Centre,
                     Texture = textures.Get(@"Gameplay/osu/disc"),
                 },
-                new TrianglesPiece(drawableHitObject.GetHashCode())
+                triangles = new TrianglesPiece
                 {
                     RelativeSizeAxes = Axes.Both,
                     Blending = BlendingParameters.Additive,
                     Alpha = 0.5f,
                 }
             };
+
+            drawableObject.HitObjectApplied += onHitObjectApplied;
+            onHitObjectApplied(drawableObject);
+        }
+
+        private void onHitObjectApplied(DrawableHitObject obj)
+        {
+            if (obj.HitObject == null)
+                return;
+
+            triangles.Reset((int)obj.HitObject.StartTime);
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+
+            if (drawableObject != null)
+                drawableObject.HitObjectApplied -= onHitObjectApplied;
         }
     }
 }
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ExplodePiece.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ExplodePiece.cs
@@ -1,14 +1,21 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Game.Rulesets.Objects.Drawables;
 using osuTK;
 
 namespace osu.Game.Rulesets.Osu.Objects.Drawables.Pieces
 {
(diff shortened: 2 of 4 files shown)


bdach:
`ExpireOffScreenTriangles` no longer even does anything since 97e57178a77af3fee0c8cd448df623f042115846.

As for `CreateNewTriangles`, it's been true as long as the component has lived (since 12e265a3e2d0cac98b49aff080f6396000df8970). Given that `TrianglesPiece` won't update if it's not present I don't think that's necessarily a problem but I haven't done basically any benchmarking. Makes sense that they would get disappeared as the movement logic is based off of `Time.Elapsed`.

smoogipoo:
> Might as well remove it here?

I believe there's other rulesets using `Triangles`, so I've deprecated it instead.

peppy:
This doesn't seem to resolve the issue for me. After several seeks things still have no triangles.

peppy:
I'd argue the correct solution here is either to reconstruct the drawable or add a `Triangles.Reset()` method, since `CreateNewTriangles => true` is probably not what we want (I can't begin to cover the edge cases this could cause with pooling, depending on when in time the triangles drawables are reused).

Keep in mind there's still remaining edge cases with `Triangles`, such as seeking backwards until they are in view (ie. in the editor) displaying completely differently to seeking forwards into view (the display is based on the current time at the point of `LoadComplete`, mostly). Not sure whether that's something to consider while fixing this one.

smoogipoo:
It would be exceedingly difficulty to fix the rewinding case, and it's really not that noticeable/important imo.

> This doesn't seem to resolve the issue for me. After several seeks things still have no triangles.

I can't replicate. Make sure to rebuild?

> cover the edge cases this could cause with pooling

Can you name one? Just so I know what to look out for? The one thing jarring thing that I've noticed is new triangles popping in that I didn't notice originally, but is there more to it that you've noticed?

peppy:
As discerned in discord, explosions are working but triangles inside hitcircles are not.

smoogipoo:
I've changed it to use a new `Triangles.Reset()`, which adds back the `CreateNewTriangles => false` override. I've also synced up the circle and explosion pieces as part of this.

>  explosions are working but triangles inside hitcircles are not.

Seems to somehow fix this also.

bdach:
From what I can tell, the original solution didn't work because of the directionality of `Time.Elapsed`:

https://github.com/ppy/osu/blob/851c7d524fe6f3ce6b638718ec4abf454c58d2df/osu.Game/Graphics/Backgrounds/Triangles.cs#L157

With a large negative `Time.Elapsed` the large triangles would move up quite a ways and only return after quite a bit of forwards playback. This is also corroborated by the fact that sometimes I could see the rare small triangle still in bounds.

Since the new method uses an absolute time instant in `Reset()`, I believe it should work well and it seems to do so in my testing (scrolled back and forth tens of times).

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ExplodePiece.cs
```diff
@@ -18,13 +25,36 @@ public ExplodePiece()
...
             };
+
+            drawableObject.HitObjectApplied += onHitObjectApplied;
+            onHitObjectApplied(drawableObject);
+        }
+
+        private void onHitObjectApplied(DrawableHitObject obj)
+        {
+            if (obj.HitObject == null)
+                return;
```
bdach, line 47: Are these null-checks really required? From what I can tell DHO cannot invoke `HitObjectApplied` with a null `HitObject`, or else they'd throw here in the first place:

https://github.com/ppy/osu/blob/851c7d524fe6f3ce6b638718ec4abf454c58d2df/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs#L191

Or is this for safety in case that the object is freed in the meantime? Can't imagine how that'd happen, though...

smoogipoo: Shouldn't be required - not sure why I left this here.

smoogipoo: Ah no - I added it for safety because of the immediate invocation in `load()`, at which point DHOs can't be assumed to have a hitobject (although they might in tests, which is why that invocation exists in the first place).

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/CirclePiece.cs
```diff
@@ -36,13 +41,32 @@ private void load(TextureStore textures, DrawableHitObject drawableHitObject)
...
                 }
             };
+
+            drawableObject.HitObjectApplied += onHitObjectApplied;
+            onHitObjectApplied(drawableObject);
+        }
+
+        private void onHitObjectApplied(DrawableHitObject obj)
+        {
+            if (obj.HitObject == null)
```
peppy, line 58: how can this happen? seems a bit dodge.

peppy: never mind, duplicate.

osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/ExplodePiece.cs
peppy: Could this potentially work?

image

smoogipoo: No because of the second care I mentioned in my edit - if the pool is empty, the new drawable is loaded after `Apply()`.

bdach (approved):
I think I'm ok with this

peppy (approved):

PR #10876, opened 2020-11-16
Fix slider control point connections not being updated

nbvdkamp:
The indices stored in `PathControlPointConnectionPiece`s weren't being updated, leading to incorrect behaviour when adding points to or removing points from the middle of a slider, for example:
image

### DIFF
--- a/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointConnectionPiece.cs
+++ b/osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointConnectionPiece.cs
@@ -20,17 +20,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
 
         private readonly Path path;
         private readonly Slider slider;
-        private int controlPointIndex;
-
-        public int ControlPointIndex
-        {
-            get => controlPointIndex;
-            set
-            {
-                controlPointIndex = value;
-                updateConnectingPath();
-            }
-        }
+        public int ControlPointIndex { get; set; }
 
         private IBindable<Vector2> sliderPosition;
         private IBindable<int> pathVersion;
@@ -38,7 +28,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
         public PathControlPointConnectionPiece(Slider slider, int controlPointIndex)
         {
             this.slider = slider;
-            this.controlPointIndex = controlPointIndex;
+            ControlPointIndex = controlPointIndex;
 
             Origin = Anchor.Centre;
             AutoSizeAxes = Axes.Both;
@@ -74,7 +64,7 @@ namespace osu.Game.Rulesets.Osu.Edit.Blueprints.Sliders.Components
 
             path.ClearVertices();
 
-            int nextIndex = controlPointIndex + 1;
+            int nextIndex = ControlPointIndex + 1;
             if (nextIndex == 0 || nextIndex >= slider.Path.ControlPoints.Count)
                 return;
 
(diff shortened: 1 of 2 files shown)


osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointConnectionPiece.cs
```diff
@@ -20,7 +20,17 @@ public class PathControlPointConnectionPiece : CompositeDrawable
...
+
+        public int ControlPointIndex
+        {
+            get => controlPointIndex;
+            set
+            {
+                controlPointIndex = value;
+                updateConnectingPath();
+            }
+        }
```
bdach, line 33: maybe make this a bindable to match `sliderPosition`/`pathVersion` at this point?

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
```diff
@@ -66,6 +66,17 @@ private void onControlPointsChanged(object sender, NotifyCollectionChangedEventA
             switch (e.Action)
             {
                 case NotifyCollectionChangedAction.Add:
+                    // If inserting in the the path (not appending),
```
bdach, line 69:
```suggestion
                    // If inserting in the path (not appending),
```

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
```diff
@@ -82,12 +93,25 @@ private void onControlPointsChanged(object sender, NotifyCollectionChangedEventA
                     break;
 
                 case NotifyCollectionChangedAction.Remove:
+                    int oldSize = Pieces.Count;
```
bdach, line 96: I'd just inline this into the condition at this point.

bdach (commented):
seems correct, just a few stylistic suggestions

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointConnectionPiece.cs
nbvdkamp: Turns out the `updateConnectingPath` call wasn't necessary there since the pathVersion value changes right after so I made it an auto-implemented property instead.

osu.Game.Rulesets.Osu/Edit/Blueprints/Sliders/Components/PathControlPointVisualiser.cs
```diff
@@ -103,7 +101,7 @@ private void onControlPointsChanged(object sender, NotifyCollectionChangedEventA
 
                     // If removing before the end of the path,
                     // update indices of connections after remove location
-                    if (e.OldStartingIndex + e.OldItems.Count < oldSize)
+                    if (e.OldStartingIndex + e.OldItems.Count < Pieces.Count + e.OldItems.Count)
```
bdach, line 104: hmm, i missed the fact that `Pieces` was being removed from in between... this reads a bit strange now. subtracting `e.OldItems.Count` from both sides of the inequality is functionally equivalent, but not sure if legible.

nbvdkamp: I guess it still makes sense without the addition, if the index where we started removing is less than the number of items remaining in the path after removal that must mean we removed a section in the middle of the path.

peppy (approved):

PR #10877, opened 2020-11-17
Fix hold note judgements displaying incorrectly

smoogipoo:
Two overlapping issues here:
1. Hold note judgement explosions were always displayed (they always get `IgnoreHit`), so `judgedObject.DisplayResult` needs to be used.
2. By applying (1), `HoldNoteTick`s would not display judgement explosions, so I've removed their override and added a case specifically for them in the `Stage` to stop the text from displaying.


bdach (approved):
I also noticed this during work on the hold note fade PR, but forgot to circle back to it / got distracted along the way...

peppy (approved):

PR #10878, opened 2020-11-17
Add ability for skins to have more control over judgement transforms

### LINKED ISSUE #10195: For legacy skins, judgements should use animations that match stable

peppy:
Closes #10195.

Can be tested with existing test scene (`osu.Game.Rulesets.Osu.Tests.TestSceneDrawableJudgement`) – special skin is animated and should no longer show a scale transform on misses *or* hits. All others should remain as previous.

Ended up implementing this in a non-breaking way (I believe).

### DIFF
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -77,7 +77,7 @@ namespace osu.Game.Rulesets.Judgements
             this.Delay(FadeOutDelay).FadeOut(400);
         }
 
-        public virtual void Apply([NotNull] JudgementResult result, [CanBeNull] DrawableHitObject judgedObject)
+        public void Apply([NotNull] JudgementResult result, [CanBeNull] DrawableHitObject judgedObject)
         {
             Result = result;
             JudgedObject = judgedObject;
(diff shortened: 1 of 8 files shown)


bdach (approved):
pushed two menial changes, otherwise looks good to me :+1:

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -91,33 +109,31 @@ protected override void PrepareForUse()
...
                     break;
 
                 default:
                     ApplyHitAnimations();
                     break;
             }
 
+            if (skinnableJudgement.Drawable is IAnimatableJudgement animatable)
+            {
+                using (BeginAbsoluteSequence(Result.TimeAbsolute))
```
smoogipoo, line 133: This won't have any effect because it doesn't have `recurse = true`.

peppy: should be fixed in the follow-up PR already

smoogipoo: Indeed it's removed ':D

peppy: not removed, just much further out in scope.

PR #10881, opened 2020-11-17
Fix TestSceneFruitObjects not displaying anything

### LINKED ISSUE #10880: Catch.Tests/TestSceneFruitObjects regression
The test scene should show fruits in several skins, but doesn't show any fruits.
Bisected to <https://github.com/ppy/osu/commit/cead67d51209a1f3a1869448d15345b9d78577f5> this commit but I'm not sure how to fix the issue.

ekrctb:
Fix #10880.
Are additional assertions (to prevent regression) necessary? Not sure how to assert this in a clean way.


peppy (dismissed):

smoogipoo (approved):
I think this is fine. Unfortunate this broke but it's very edgecase and I don't expect it to ever be used outside of tests.

PR #10882, opened 2020-11-17
Add File Import screen

PercyDan54:
The code is from [a custom osu build](https://gitee.com/matrix-feather/osu)
It's useful for Android users who can't import files because they can't open multiple windows.
Gif showcase (Large file):


bdach:
First of all, @MATRIX-feather, are you aware that this has been PRed on your behalf? Did you agree to this?

@PercyDan54, you cannot PR changes that are not your own without declaring whether you have consent to do so first.

Secondly, I am 99% sure this cannot be upstreamed in its current form. If (and that's a big if) this screen is to be added, it should only be visible to Android users. I don't believe we should be adding this at all, however, as it's a major workaround of a major shortcoming of the current Android build.

MATRIX-feather:
> First of all, @MATRIX-feather, are you aware that this has been PRed on your behalf? Did you agree to this?
> 
> @PercyDan54, you cannot PR changes that are not your own without declaring whether you have consent to do so first.
> 
> Secondly, I am 99% sure this cannot be upstreamed in its current form. If (and that's a big if) this screen is to be added, it should only be visible to Android users. I don't believe we should be adding this at all, however, as it's a major workaround of a major shortcoming of the current Android build.

Yes, I agree

peppy:
Rather than this, how about adding proper file import support? Surely this can be done much easier without this weird UX.

bdach:
Please clarify what part you agree with as it's not clear to me. Also, are you sure you don't want at least authorship credit for the commit? The change has been committed in @PercyDan54's name and by their own admission it's not their code.

PercyDan54:
> it should only be visible to Android users.

Maybe add `RuntimeInfo.OS == RuntimeInfo.Platform.Android` ?

peppy:
See my previous comment. As is the case in iOS, files should be pulled in from outside the app via associations, not vice versa.

LittleEndu:
If the user gives permission for the app to access files on their phone, then on android it makes at least a little sense to have it inside the app. iirc that's not the case on iOS because apps  don't have file level access to anything, unlike android where it's managed by the permission.

MATRIX-feather:
@bdach I agreed with @PercyDan54  to submit my code upstream, commit message is not very necessary to me tho.

peppy:
@LittleEndu iOS can also do this. but again, it's not the UX i am going for. there's no reason for it to be done this way, and the file selector wasn't made with mobile in mind. this may come at a later point but it should be preceded with support for importing files the opposite direction.

there is zero chance of this PR getting merged unless it is a button under the debug settings or something completely isolated and temporary like that.

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+                                                },
+                                            }
+                                        }
+                                    }
+                                }
+                            },
+                        }
+                    }
+                }
+            };
```
peppy, line 211: this just can't exist. you probably want to split some of this out into its own component or re-think structure in general.

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+                                                    RelativeSizeAxes = Axes.Both
+                                                },
+                                                new FillFlowContainer
+                                                {
+                                                    RelativeSizeAxes = Axes.X,
+                                                    AutoSizeAxes = Axes.Y,
+                                                    Spacing = new Vector2(10),
+                                                    Children = new Drawable[]
+                                                    {
+                                                        new SettingsEnumDropdown<FileFilterType>
```
peppy, line 148: Rather than having this, why not just always show every filetype? I don't see much benefit to being able to filter it.

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+                                                        {
+                                                            Anchor = Anchor.BottomCentre,
+                                                            Origin = Anchor.BottomCentre,
+                                                            LabelText = "File Type",
+                                                            Current = filterType,
+                                                            Margin = new MarginPadding { Bottom = 15 }
+                                                        },
+                                                        new GridContainer
```
peppy, line 156: as an example of restructuring, this should not require a `GridContainer` at all.

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+                                                                new Drawable[]
+                                                                {
+                                                                    new TriangleButton
+                                                                    {
+                                                                        Anchor = Anchor.BottomCentre,
+                                                                        Origin = Anchor.BottomCentre,
+                                                                        RelativeSizeAxes = Axes.X,
+                                                                        Height = 50,
+                                                                        Width = 0.9f,
+                                                                        Text = "Refresh",
```
peppy, line 178: I'd argue this just shouldn't exist. If we're adding a refresh button it would be part of the file chooser itself.

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+            Task.Factory.StartNew(() => gameBase.Import(paths), TaskCreationOptions.LongRunning);
+        }
+
+        public enum FileFilterType
+        {
+            Skin,
+            Beatmap,
+            Replay,
+            All
+        }
```
peppy, line 318: remove this to simplify implementation

osu.Game/Screens/Import/FileImportScreen.cs
```diff
@@ -0,0 +1,320 @@
...
+            fileSelector = new FileSelector(initialPath: directory, validFileExtensions: fileExtensions)
+            {
+                RelativeSizeAxes = Axes.Both
+            };
+
+            currentDirectory.BindTo(fileSelector.CurrentPath);
+            currentFile.BindTo(fileSelector.CurrentFile);
+
+            fileSelectContainer.Add(fileSelector);
+        }
```
peppy, line 267: remove this to simplify implementation

peppy (changes requested):

peppy:
Curious why you can't use the existing [MigrationSelectScreen](https://github.com/ppy/osu/blob/master/osu.Game/Overlays/Settings/Sections/Maintenance/MigrationSelectScreen.cs) rather than this custom design?

peppy:
Were the above actions intentional?

PercyDan54:
> Were the above actions intentional?

I accidentally did that

peppy:
I've tidied this up into a mergeable state. Not going to dwell on implementation for now because I'm not sure how permanent this will even be.

A second pass would see it sharing dialog logic with `ManageCollectionsDialog` (which requires some refactoring to handle non-global dialogs like this). But that's for another day.

peppy (approved):

PR #10884, opened 2020-11-17
Tighten osu! ruleset lifetime expiry for past hitobjects

### LINKED ISSUE #10857: After seeking far into the future in non-frame-stable contexts, previous hitobjects remain alive
**Describe the bug:** 
In the editor, you can sometimes select previous objects that are a very long time back (no longer on screen + multiple combos ago). I have no idea what causes this.

https://streamable.com/cqepx3

**osu!lazer version:** 
2020.1114.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/5544183/database.log)
[network.log](https://github.com/ppy/osu/files/5544184/network.log)
[performance.log](https://github.com/ppy/osu/files/5544185/performance.log)
[performance-audio.log](https://github.com/ppy/osu/files/5544186/performance-audio.log)
[performance-draw.log](https://github.com/ppy/osu/files/5544187/performance-draw.log)
[performance-update.log](https://github.com/ppy/osu/files/5544188/performance-update.log)
[runtime.log](https://github.com/ppy/osu/files/5544190/runtime.log)

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10857

Doing this local to the osu! ruleset for now to see how it feels, if any problems come up, and how others feel about this with the expectation that this is to be moved into `DrawableHitObject` in the future.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
@@ -61,6 +61,13 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             PositionBindable.BindTo(HitObject.PositionBindable);
             StackHeightBindable.BindTo(HitObject.StackHeightBindable);
             ScaleBindable.BindTo(HitObject.ScaleBindable);
+
+            // Manually set to reduce the number of future alive objects to a bare minimum.
+            LifetimeStart = HitObject.StartTime - HitObject.TimePreempt;
+
+            // Arbitrary lifetime end to prevent past objects in idle states remaining alive in non-frame-stable contexts.
+            // An extra 1000ms is added to always overestimate the true lifetime, and a more exact value is set by hit transforms and the following expiry.
+            LifetimeEnd = HitObject.GetEndTime() + HitObject.HitWindows.WindowFor(HitResult.Miss) + 1000;
         }
 
         protected override void OnFree(HitObject hitObject)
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
```diff
@@ -60,6 +61,13 @@ protected override void OnApply(HitObject hitObject)
             PositionBindable.BindTo(HitObject.PositionBindable);
             StackHeightBindable.BindTo(HitObject.StackHeightBindable);
             ScaleBindable.BindTo(HitObject.ScaleBindable);
+
+            // Manually set to reduce the number of future alive objects to a bare minimum.
+            LifetimeStart = HitObject.StartTime - HitObject.TimePreempt;
+
+            // Arbitrary lifetime end to prevent past objects in idle states remaining alive in non-frame-stable contexts.
+            // An extra 1000ms is added to always overestimate the true lifetime, and a more exact value is set by hit transforms and the following expiry.
+            LifetimeEnd = HitObject.GetEndTime() + HitObject.HitWindows.WindowFor(HitResult.Miss) + 1000;
```
peppy, line 70: we already do very similar to this in scrolling containers right?

smoogipoo: Nah, scrolling containers also rely on the hitobjects to expire, because it is/used to be hard for them to tell when the appropriate expiry time should be.
They’ll need something similar though.

bdach:
I'm fine with this source and this is definitely an improvement, but something is not 100% right with nested elements still on rewind - [see video](https://drive.google.com/file/d/13Rx2pnDyuLO_HJPPuD7q5Sk3i_vs49Ee/view?usp=sharing). Sometimes the nested elements (sliderhead circles, repeats, etc.) come back and then play correctly forwards again, but sometimes not. Sliderheads are definitely the worst of this as they sometimes require to go back tens of beats before they play forwards correctly again.

I haven't looked at what might the cause be - I'm not saying this is a problem with this PR, but I figured I'd point this out.

peppy:
@bdach i think this is the same thing i mentioned recently in discord. will open an issue to cover it, because it is apparently a different issue.

peppy (approved):

peppy:
This definitely needs to be moved to a non-ruleset specific location is we decide to do this, but let's see how things go.

PR #10886, opened 2020-11-18
Add HUD Visibility dropdown to Visual Settings panel

InfinityGhost:
Added a dropdown for the HUD visibility setting in the visual settings panel in the pre-gameplay screen.
It is included in the same settings group in the regular settings panel and figured it would work quite well here as well.

### DIFF
--- a/osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
+++ b/osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
--- a/osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
+++ b/osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.UserInterface;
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            public PlayerSettingsDropdownHeader()
+            {
+                Height = 25;
+                Foreground.Padding = new MarginPadding { Top = 4, Bottom = 4, Left = 8, Right = 8 };
+            }
+        }
+
+        private class PlayerSettingsDropdownMenu : OsuDropdownMenu
+        {
```
smoogipoo, line 36: I think you meant to override OnHover here and set to `false`.

osu.Game/Screens/Play/PlayerSettings/PlayerSettingsDropdown.cs
```diff
@@ -0,0 +1,39 @@
...
+        where T : struct, Enum
+    {
+        protected override DropdownHeader CreateHeader() => new PlayerSettingsDropdownHeader();
+
+        protected override DropdownMenu CreateMenu() => new PlayerSettingsDropdownMenu();
+
+        [BackgroundDependencyLoader]
+        private void load(OsuColour colours)
+        {
+            AccentColour = colours.Yellow;
```
smoogipoo, line 23: The text doesn't look all that great. I think the text should change to black it it's active.

Can be done by deriving `OsuDropdownMenuItem` and setting `ForegroundSelected = Color4.Black` I believe.

smoogipoo (changes requested):

peppy:
before reviewing, is this really something we want? the toggles on the loading screen are supposed to be things people change on a per beatmap basis. i strongly disagree this is a candidate for that.

smoogipoo:
I thought the same thing, and it only has an effect for people that use `Never`, but if someone does use that then I can see why they may want to enable it for replays only.

With that being said the other way around becomes kinda weird, and as you said, effectively a per-beatmap toggle which is not what we want. So yeah I'd also tend to not merging this if that's the case.

smoogipoo:
Actually this can already be toggled via Shift+Tab, so yeah I'd deem it unnecessary.

InfinityGhost:
> Actually this can already be toggled via Shift+Tab, so yeah I'd deem it unnecessary.

While unnecessary, I still think it provides some user accessibility for those who don't know the keybinds or play on a touchscreen and wish to change it prior to gameplay.

> the toggles on the loading screen are supposed to be things people change on a per beatmap basis.

I do agree on this point of view, however I could see a few people actually changing this type of thing on a per beatmap basis as well.

peppy:
I don't want these kinds of options in the load screen. They should be specifically about the gameplay / beatmap content.

PR #10890, opened 2020-11-18
Adjust judgement animations to match stable

### LINKED ISSUE #10172: Changing skin settings during gameplay causes some judgements to show a wrong animation
**Describe the bug:** If you change any skin setting (_beatmap hitsounds_ for example) during gameplay, you can notice some judgements with a wrong animation. It seems to happen only with the default skin, but I'm not 100% sure.

Judgement animation

Very weird, but the wrong animation looks like the osu!classic's one (is it just a coincidence?)

### LINKED ISSUE #10195: For legacy skins, judgements should use animations that match stable

peppy:
This adds complete coverage of new and old style legacy skin judgements. Until now, we were applying a single set of transforms (based on lazer's default skin) across the board, as it was a bit difficulty to correctly insert the transforms in place (resolved in https://github.com/ppy/osu/pull/10878).

Particles are still not implemented; that will come in a follow-up PR.

- [x] Depends on #10878 
- [x] Depends on #10889
- Closes #10172
- Closes #10195

### DIFF
--- a/osu.Game.Rulesets.Mania/UI/DrawableManiaJudgement.cs
+++ b/osu.Game.Rulesets.Mania/UI/DrawableManiaJudgement.cs
@@ -5,6 +5,7 @@ using osu.Framework.Graphics;
 using osu.Game.Rulesets.Judgements;
 using osu.Game.Rulesets.Objects.Drawables;
 using osu.Game.Rulesets.Scoring;
+using osuTK;
 
 namespace osu.Game.Rulesets.Mania.UI
 {
@@ -19,6 +20,27 @@ namespace osu.Game.Rulesets.Mania.UI
         {
         }
 
+        protected override void ApplyMissAnimations()
+        {
+            if (!(JudgementBody.Drawable is DefaultManiaJudgementPiece))
+            {
+                // this is temporary logic until mania's skin transformer returns IAnimatableJudgements
+                JudgementBody.ScaleTo(1.6f);
+                JudgementBody.ScaleTo(1, 100, Easing.In);
+
+                JudgementBody.MoveTo(Vector2.Zero);
+                JudgementBody.MoveToOffset(new Vector2(0, 100), 800, Easing.InQuint);
+
+                JudgementBody.RotateTo(0);
+                JudgementBody.RotateTo(40, 800, Easing.InQuint);
+                JudgementBody.FadeOutFromOne(800);
+
+                LifetimeEnd = JudgementBody.LatestTransformEndTime;
+            }
+
+            base.ApplyMissAnimations();
+        }
+
         protected override void ApplyHitAnimations()
         {
             JudgementBody.ScaleTo(0.8f);
@@ -29,11 +51,11 @@ namespace osu.Game.Rulesets.Mania.UI
                          .FadeOut(200);
         }
 
-        protected override Drawable CreateDefaultJudgement(HitResult result) => new ManiaJudgementPiece(result);
+        protected override Drawable CreateDefaultJudgement(HitResult result) => new DefaultManiaJudgementPiece(result);
 
-        private class ManiaJudgementPiece : DefaultJudgementPiece
+        private class DefaultManiaJudgementPiece : DefaultJudgementPiece
         {
-            public ManiaJudgementPiece(HitResult result)
+            public DefaultManiaJudgementPiece(HitResult result)
                 : base(result)
             {
             }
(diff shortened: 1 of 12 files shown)


osu.Game.Rulesets.Mania/UI/DrawableManiaJudgement.cs
```diff
@@ -19,22 +20,42 @@ public DrawableManiaJudgement()
         {
         }
 
-        protected override double FadeInDuration => 50;
+        protected override void ApplyMissAnimations()
+        {
+            if (!(JudgementBody.Drawable is DefaultManiaJudgementPiece))
+            {
+                // this is temporary logic until mania's skin transformer returns IAnimatableJudgements
```
bdach, line 27: I suppose that's going to be a separate pull?

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -62,32 +67,49 @@ private void load()
...
+            // we need to trigger a reinitialisation to make things right.
+            currentDrawableType = null;
+
+            PrepareForUse();
+        }
+
+        protected override void Dispose(bool isDisposing)
+        {
+            base.Dispose(isDisposing);
+            skinSource.SourceChanged -= onSkinChanged;
```
bdach, line 88: probably needs a null-check of `skinSource` because finaliser things

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
```diff
@@ -109,32 +131,46 @@ protected override void PrepareForUse()
...
+                        break;
+
+                    default:
+                        ApplyHitAnimations();
+                        break;
+                }
+
+                if (JudgementBody.Drawable is IAnimatableJudgement animatable)
+                {
+                    var drawableAnimation = (Drawable)animatable;
```
bdach, line 163: I'd propose adding lifetime members to `IDrawable` in framework to avoid this casting, but not sure if worth just for this / in line with intentions for that interface...

osu.Game.Rulesets.Osu.Tests/TestSceneDrawableJudgement.cs
```diff
@@ -43,10 +43,8 @@ public void TestHitLightingDisabled()
             showResult(HitResult.Great);
 
             AddUntilStep("judgements shown", () => this.ChildrenOfType<TestDrawableOsuJudgement>().Any());
-            AddAssert("judgement body immediately visible",
-                () => this.ChildrenOfType<TestDrawableOsuJudgement>().All(judgement => judgement.JudgementBody.Alpha == 1));
             AddAssert("hit lighting hidden",
-                () => this.ChildrenOfType<TestDrawableOsuJudgement>().All(judgement => judgement.Lighting.Alpha == 0));
+                () => this.ChildrenOfType<TestDrawableOsuJudgement>().All(judgement => !judgement.Lighting.Transforms.Any()));
```
bdach, line 47: this assertion change is a little far-reaching I'd say - I'd probably check `TargetMember` as well for a bit more robustness

osu.Game.Rulesets.Mania/UI/DrawableManiaJudgement.cs
peppy: at some point, not imminently planned though. i'm just fixing osu! rule set for now.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
peppy: yeah, was disappointed it wasn't there. @smoogipoo any reason these properties don't exists in the interface already?

osu.Game.Rulesets.Osu.Tests/TestSceneDrawableJudgement.cs
peppy: can actually be changed back if you'd prefer. the important part ended up being the All -> Any.

bdach:
Looks good to go in for now as far as I can tell, although I'm still a little curious if the `IDrawable` stuff can be simplified/de-cast.

osu.Game/Rulesets/Judgements/DrawableJudgement.cs
smoogipoo: I'd be fine with that, but I don't think it blocks this PR from merge.

smoogipoo (approved):

PR #10892, opened 2020-11-18
Make audio latency configurable

ArniDagur:
- [ ] Depends on https://github.com/ppy/osu-framework/pull/4008.

Result of discussion from https://github.com/ppy/osu-framework/pull/3832.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
@@ -19,8 +19,22 @@ namespace osu.Game.Overlays.Settings.Sections.Audio
             {
                 new SettingsSlider<int>
                 {
-                    LabelText = "Device Period",
-                    Current = audio.DevicePeriod,
+                    LabelText = "Device update period",
+                    Current = audio.DeviceUpdatePeriod,
+                    KeyboardStep = 1,
+                    TransferValueOnCommit = true
+                },
+                new SettingsSlider<int>
+                {
+                    LabelText = "Device buffer size",
+                    Current = audio.DeviceBufferSize,
+                    KeyboardStep = 1,
+                    TransferValueOnCommit = true
+                },
+                new SettingsSlider<int>
+                {
+                    LabelText = "Playback buffer size",
+                    Current = audio.PlaybackBufferSize,
                     KeyboardStep = 1,
                     TransferValueOnCommit = true
                 }
--- a/osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
@@ -1,0 +1,30 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Audio;
+using osu.Framework.Graphics;
+using osu.Game.Configuration;
+
+namespace osu.Game.Overlays.Settings.Sections.Audio
+{
+    public class LatencySettings : SettingsSubsection
+    {
+        protected override string Header => "Latency";
+
+        [BackgroundDependencyLoader]
+        private void load(AudioManager audio, OsuConfigManager config)
+        {
+            Children = new Drawable[]
+            {
+                new SettingsSlider<int>
+                {
+                    LabelText = "Device Period",
+                    Current = audio.DevicePeriod,
+                    KeyboardStep = 1,
+                    TransferValueOnCommit = true
+                }
+            };
+        }
+    }
+}
--- a/osu.Game/Overlays/Settings/Sections/AudioSection.cs
+++ b/osu.Game/Overlays/Settings/Sections/AudioSection.cs
@@ -2,6 +2,7 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Runtime.InteropServices;
 using System.Linq;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Sprites;
@@ -22,13 +23,18 @@ namespace osu.Game.Overlays.Settings.Sections
 
         public AudioSection()
         {
-            Children = new Drawable[]
+            List<Drawable> children = new List<Drawable>();
+            children.Add(new AudioDevicesSettings());
+            children.Add(new VolumeSettings());
+            children.Add(new OffsetSettings());
+            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX)))
             {
-                new AudioDevicesSettings(),
-                new VolumeSettings(),
-                new OffsetSettings(),
-                new MainMenuSettings()
-            };
+                // Adjusting latency is neither required nor supported on
+                // Windows and Mac.
+                children.Add(new LatencySettings());
+            }
+            children.Add(new MainMenuSettings());
+            Children = children;
         }
     }
 }
--- a/osu.Game/Overlays/Settings/Sections/AudioSection.cs
+++ b/osu.Game/Overlays/Settings/Sections/AudioSection.cs
@@ -21,20 +21,22 @@ namespace osu.Game.Overlays.Settings.Sections
 
         public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "sound" });
 
+        private IEnumerable<Drawable> getSections()
+        {
+            yield return new AudioDevicesSettings();
+            yield return new VolumeSettings();
+            yield return new OffsetSettings();
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) {
+                // Adjusting latency is neither required nor supported on
+                // other platforms.
+                yield return new LatencySettings();
+            }
+            yield return new MainMenuSettings();
+        }
+
         public AudioSection()
         {
-            List<Drawable> children = new List<Drawable>();
-            children.Add(new AudioDevicesSettings());
-            children.Add(new VolumeSettings());
-            children.Add(new OffsetSettings());
-            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX)))
-            {
-                // Adjusting latency is neither required nor supported on
-                // Windows and Mac.
-                children.Add(new LatencySettings());
-            }
-            children.Add(new MainMenuSettings());
-            Children = children;
+            ChildrenEnumerable = getSections();
         }
     }
 }


osu.Game/Overlays/Settings/Sections/AudioSection.cs
```diff
@@ -22,13 +23,18 @@ public class AudioSection : SettingsSection
 
         public AudioSection()
         {
-            Children = new Drawable[]
+            List<Drawable> children = new List<Drawable>();
+            children.Add(new AudioDevicesSettings());
+            children.Add(new VolumeSettings());
+            children.Add(new OffsetSettings());
+            if (!(RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX)))
```
bdach, line 30: * use `osu.Framework.RuntimeInfo` instead.
* this should probably be `RuntimeInfo.OS == RuntimeInfo.Platform.Linux`. i don't believe this should show on mobile platforms, either.

osu.Game/Overlays/Settings/Sections/AudioSection.cs
```diff
@@ -22,13 +23,18 @@ public class AudioSection : SettingsSection
...
-                new AudioDevicesSettings(),
-                new VolumeSettings(),
-                new OffsetSettings(),
-                new MainMenuSettings()
-            };
+                // Adjusting latency is neither required nor supported on
+                // Windows and Mac.
+                children.Add(new LatencySettings());
+            }
+            children.Add(new MainMenuSettings());
```
bdach, line 36: would probably rewrite the section generation to use `yield return` as it looks pretty bad as done here.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
```diff
@@ -0,0 +1,30 @@
...
+        protected override string Header => "Latency";
+
+        [BackgroundDependencyLoader]
+        private void load(AudioManager audio, OsuConfigManager config)
+        {
+            Children = new Drawable[]
+            {
+                new SettingsSlider<int>
+                {
+                    LabelText = "Device Period",
```
bdach, line 22:
```suggestion
                    LabelText = "Device period",
```
setting casing is all over the place right now but this seems to be the more prevalent convention

bdach (changes requested):

osu.Game/Overlays/Settings/Sections/AudioSection.cs
peppy: not wanted on macOS either. this is linux specific.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
peppy: is "device period" actually a technical term, or just an internal variable name in bass? if it's not a well-known term (with a wiki page or similar) a better name would be "buffer length", since that is actually what it's doing.

peppy: note that i'd also rename this framework-side as well, since "device period" means crap all to me without referencing the bass documentation.

bdach: windows's [`IAudioClient`](https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-getdeviceperiod) and [ALSA](https://users.suse.com/~mana/alsa090_howto.html) both reference "periods", but either way it's such a niche technical term it might as well be called "frobulator size". really I am not a fan of this existing in any capacity on any OS either if it can be helped at all, but, alas.

peppy: Hmm, I see..

When exposed to a user, it's most commonly referred to as buffer length or buffer size (search ASIO buffer), at least historically. This is also referenced in the bass documentation as the outcome of setting the variable, so it matches.

I'd still go with *buffer something* over *something period* just to be user (and dev?) friendly.

ArniDagur: > really I am not a fan of this existing in any capacity on any OS either if it can be helped at all, but, alas.

If you don't like exposing such "low level" settings, I can think of two options as well:

1. Don't anything to the settings UI. Users will still be able to configure the device period through the `framework.ini` file, similar to how users can edit the config files on osu!stable to get smooth cursor movement in the main menu:
```
FrameSync = CompeletelyUnlimited
FrameSyncGl = CompletelyUnlimited
CustomFrameLimit = 999
```
2. Hide the latency tab (which is already Linux specific), and any other "advanced settings" behind an "Advanced User: I know what I'm doing" checkbox somewhere on the bottom of the settings UI.

bdach: `DebugSection` already exists - it might be a better place for such obscure settings (we already have bypass front-to-back and cache purge there), although then again it's not strictly a debug setting...

bdach: don't get me wrong however, my general distaste for exposing this to the user notwithstanding, I'd be able to look away and let this be merged even it's in the game UI. just very disappointing that we have to show this to the user at all in any way or circumstance.

osu.Game/Overlays/Settings/Sections/AudioSection.cs
ArniDagur: Sorry, this is my first time writing any C# code. I looked at https://www.kenneth-truyers.net/2016/05/12/yield-return-in-c/ and didn't see any example that's similar to the code we had before. Could you write out the pattern you had in mind?

bdach: what I intended to suggest is to have a function that returns the sections in order like so:

```csharp
public IEnumerable<Drawable> getSections()
{
    yield return new AudioDevicesSettings();
    yield return new VolumeSettings();
    yield return new OffsetSettings();

    if (RuntimeInfo.OS == RuntimeInfo.Platform.Linux)
        yield return new LatencySettings();

    yield return new MainMenuSettings();
}
```

and then do `ChildrenEnumerable = getSections()`.

ArniDagur: Neat. Thanks.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
peppy: it's fine to be exactly where it is with a better name, until such point that it's not necessary to play the game comfortably.

the naming changes have to happen though.

ArniDagur: I changed it to `DeviceUpdatePeriod`, since it's the time in milliseconds how often BASS _periodically_ _updates_ the _device's_ buffer. I also made is so the buffer sizes can be modified.

peppy: That isn't what I want though. In the interest of keeping things simple I would hope to follow my original naming suggestion. And there's no reason to expose the buffer if it is not needed to fix the issue.

ArniDagur: >  In the interest of keeping things simple I would hope to follow my original naming suggestion.

The thing is that the following (quoting from above)

> When exposed to a user, it's most commonly referred to as buffer length or buffer size (search ASIO buffer), at least historically. This is also referenced in the bass documentation as the outcome of setting the variable, so it matches.

is a misunderstanding, both on my part and yours. What you're referring to is the buffer size (or length), which I just now made configurable. The `DevicePeriod` setting controls how long BASS sleeps before putting more data into the buffer.

Thus, the name `LinuxAudioBufferLength` would be a misnomer. If `DeviceUpdatePeriod` is undescriptive, I suggest one of:
* `DeviceBufferUpdateInterval`
* `DeviceBufferUpdatePeriod`

I personally dislike the `Linux` prefix because BASS doesn't state that the settings only apply on Linux, and other games using the osu!framework may want to configure them on other platforms. Ultimately it's your choice however.

> And there's no reason to expose the buffer size if it is not needed to fix the issue.

Audio latency is not binary (either an issue or not). Speaking from experience talking to people on the `#osu-linux` channel on ThePoon's Discord, it seems that e.g. PulseAudio and ALSA settings that work well on some people's systems may introduce crackling on other's. Personally, I'm using a playback buffer size of 22 milliseconds in osu!lazer without any audio bugs (compared to 100ms which is the default now). I'm very much in favor of giving people as much control as possible over audio settings, to adjust them to their hardware, software, and willingness to tolerate audio crackling. I'm not the only one that thinks that way, most people on `#osu-linux` run special patched versions of Wine to give more control over audio latency. I'm hoping people won't feel the need to run special patched versions of osu!lazer as well, by introducing first party support.

I suggest adding all three settings, and if--after people have experimented with them--the conclusion is that they're not useful to reduce latency in general, removing them from the UI while retaining support in the framework.

Relevant BASS documentation:
* https://www.un4seen.com/doc/#bass/BASS_CONFIG_DEV_PERIOD.html
* https://www.un4seen.com/doc/#bass/BASS_CONFIG_DEV_BUFFER.html
* https://www.un4seen.com/doc/#bass/BASS_CONFIG_BUFFER.html

peppy: Very strongly disagree with exposing more than we need to, but I'm just repeating myself here. We already set the device buffer length and it shouldn't be adjusted. CONFIG_BUFFER has zero effect since it's only relevant to dsp/realtime audio.

How about `AudioBufferUpdatePeriod` for the name

peppy: i don't think you understand regarding the prefix.

it NEEDS TO EXIST. else users are going to start theory crafting about it affecting windows etc.

please understand we are looking to make a consumable framework here which removes unknowns. not expose every last setting from the backends we are using.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
```diff
@@ -0,0 +1,44 @@
...
+        protected override string Header => "Latency";
+
+        [BackgroundDependencyLoader]
+        private void load(AudioManager audio, OsuConfigManager config)
+        {
+            Children = new Drawable[]
+            {
+                new SettingsSlider<int>
+                {
+                    LabelText = "Device update period",
```
peppy, line 22: `Audio buffer length` should work fine and be understandable to users. Underlying configuration can be `LinuxAudioBufferUpdatePeriod `

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
```diff
@@ -0,0 +1,44 @@
...
+                    KeyboardStep = 1,
+                    TransferValueOnCommit = true
+                },
+                new SettingsSlider<int>
+                {
+                    LabelText = "Playback buffer size",
+                    Current = audio.PlaybackBufferSize,
+                    KeyboardStep = 1,
+                    TransferValueOnCommit = true
+                }
```
peppy, line 40: remove these

osu.Game/Overlays/Settings/Sections/AudioSection.cs
```diff
@@ -20,15 +21,22 @@ public class AudioSection : SettingsSection
 
         public override IEnumerable<string> FilterTerms => base.FilterTerms.Concat(new[] { "sound" });
 
+        private IEnumerable<Drawable> getSections()
+        {
+            yield return new AudioDevicesSettings();
+            yield return new VolumeSettings();
+            yield return new OffsetSettings();
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux)) {
```
peppy, line 29: incorrect braces formatting

peppy (changes requested):

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
ArniDagur: > Very strongly disagree with exposing more than we need to

> please understand we are looking to make a consumable framework here which removes unknowns. not expose every last setting from the backends we are using.

A minimal API is a good goal for sure. Keep in mind however that in a rhythm game like osu!, if there's any area where more complexity in the settings UI (or framework API for that matter) is justified, it's to allow for lower audio latency. Players are very sensitive to latency, playing with too much _feels off_, and it directly correlates with in-game accuracy.

> We already set the device buffer length and it shouldn't be adjusted. 

We set it to 100ms, but why shouldn't be adjusted if a user finds that it results in lower latency on their system? Besides the aforementioned added complexity to the settings UI and framework API.

> CONFIG_BUFFER has zero effect since it's only relevant to dsp/realtime audio.

I don't think that's true. This setting controls the playback buffer size and I can very much tell that the audio becomes glitchy when I set it too low. You can try it yourself if you don't believe me.

> i don't think you understand regarding the prefix.
> 
> it NEEDS TO EXIST. else users are going to start theory crafting about it affecting windows etc.

All right, I understand.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
ArniDagur: > `Audio buffer length` should work fine 

As I said before, this setting doesn't control the buffer length, does it?

> Underlying configuration can be `LinuxAudioBufferUpdatePeriod`

Okay.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
bdach: The reason we're pushing back so strongly against exposing internals more than required is that 99.999999% of users plainly don't have the knowledge and understanding required to grasp such settings. There is a lot of existing folksy advice already out there that's supposed to "improve performance" that is either bogus or just completely wrong (ie. setting realtime process priority or thinking unlimited draw FPS is unconditionally better than 2x limited). We don't want to go through the same spiel with audio, believe me. This is even less so the case seeing that 2/3 major platforms "just work", and Linux, as Linux tends to do, seems to suffer from ecosystem fragmentation badly on the audio front.

Snowmander:
Apologies in advance, but would exposing two more internal settings impact the support burden? I believe that allowing device buffer length to be modified can be helpful to some people with specific hardware, and bringing down every single ms of latency in general. I understand that the idea of exposing potentially unnecessary settings is quite disgusting from a developer pov, but I feel like they’re much needed settings and it’s better to go all out and let people tinker with everything latency related however they seem fit. Especially since that 100ms device period is just an arbitrary number that seems to work for everyone.

No technical support needs to be provided for people who install Linux for this and are tinkering this deep. If the settings are too “protruded” and too accessible to rude newcomers, burying it in a submenu or a config file can be explored in the future.

I can experiment to find if those two settings make a difference in my hardware (which I’m almost positive it does) and make a case for them, but unfortunately my setup is slightly busted right now and I can’t afford to mess with it right now. If I do so in the future and can make a valid case for them, would the developers be open to the possibility of exposing those settings?

peppy:
thanks for your comment. we will expose them if and when that is required.

it is not happening in this PR and also not happening without satisfactory documentation as to why it should be exposed.

osu.Game/Overlays/Settings/Sections/Audio/LatencySettings.cs
peppy: as long as there is one setting the text i specified works best to describe *to a user* what it is affecting.

we will add descriptive text later when setting supports it.

2str8:
So i wanna ask, is this could be implemented when Windows version gonna switch to WASAPI? Can we manually adjust latency on both Operating Systems in future, if this is not happening right now, as i can see? But still, if it possible to apply this PR - that would be awesome, especially for many of those, who plays this game on linux because of configurable latency. Thanks.

peppy:
Windows already uses WASAPI. We will be looking at using a more direct transfer method in the future, and, if necessary we will add configuration for it. The general rule is that we only add configuration where actually required to ensure stability at the performance levels we expect.

This PR is already implementing in a way which can be extended to such use cases in the future (another reason i am proposing using generic and understandable terms).

smoogipoo:
Closing due to inactivity. Still need to discuss how to do this since we all have different ideas/expectations.

PR #10900, opened 2020-11-18
Fix possible NullReferenceException when disposing SliderSelectionBlueprint

OliBomby:
I noticed frequent NullReferenceExceptions when scrolling through any map in the editor with my local debug build. I found that the exception came from the 'controlPoints' getter in SliderSelectionBlueprint when HitObject is null. It seems this getter gets called by Drawable.UnbindAllUnbindables during disposing and that's when the null reference can happen.

image

I suggest this simple fix that makes 'controlPoints' an IList so it gets ignored by the unbinding code. The code that uses 'controlPoints' only uses its IList properties, so nothing is compromised here I believe.

I wasn't able to recreate the issue with unit tests, so I didn't make any extra unit tests.


bdach:
First off, PRing changes from `master` back to your fork is not the proper way to update it as it leads to broken history. Please see [this guide](https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/syncing-a-fork) on how to do it properly.

Secondly, this fix is reallt a very unsafe bypass. To describe the nullref scenario you mention in detail, the `UnbindAllBindables()` in question is coming from the finaliser path, which means that [there are basically no guarantees about the state of the object being finalised](https://ericlippert.com/2015/05/18/when-everything-you-know-is-wrong-part-one/). In the case in question, `HitObject` is null.

The fix proposed in this PR circumvents the framework-side automatic unbind logic. While right now it is *probably maybe sort of* safe because as the bindable comes from another object, it will *possibly maybe* be unbound by that other object, in general it isn't safe as it might lead to bindable subscriptions not being cleaned up and result in leaks.

I believe a more correct fix to apply here would be to null-propagate `HitObject`, as the framework-side code [does not try to automatically unbind bindable-type properties if they are `null`](https://github.com/ppy/osu-framework/blob/e00a7512e3798f6dda1728dd952226962a31cb72/osu.Framework/Graphics/Drawable.cs#L154-L157):

```csharp
private BindableList<PathControlPoint> controlPoints => HitObject?.Path.ControlPoints;
```

although it is still risky by the virtue of all of the surrounding code expecting it not to be `null`, so all of that code might need to be rewritten/reworked somewhat.

bdach (changes requested):
as above

bdach:
Also, in the future please PR from a topic branch [as described in the contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md).

smoogipoo:
I don't get it, the exception should be caught and handled by the unbind code:

bdach:
Yeah I'm pretty sure it is handled. This PR is proposing a way to not have it thrown in the first place.

smoogipoo:
Yeah idk, it's super weird to have a null-check here to fix essentially a framework bug, if it can even be called that (the exception can be ignored and execution resumed).

Imo the real fix is to make framework not unbind properties - I think it was only left doing so as a "what could go wrong" measure, because auto-properties would come under the fields unbind anyway. I don't think it was made for the purpose of unbinding external bindables (that sounds dodgy as heck).

Or, with taking that into consideration, another fix is to bind a local list to the hitobject's control points.

OliBomby:
> Yeah idk, it's super weird to have a null-check here to fix essentially a framework bug, if it can even be called that (the exception can be ignored and execution resumed).
> 
> Imo the real fix is to make framework not unbind properties - I think it was only left doing so as a "what could go wrong" measure, because auto-properties would be covered under the fields unbind anyway. I don't think it was made for the purpose of unbinding external bindables (that sounds dodgy as heck).
> 
> Or, with taking that into consideration, another fix is to bind a local list to the hitobject's control points.

The exception halts the local debug program repeatedly while playing a map and when the exception is ignored it still causes a lag spike while my IDE is doing some kind of exception overhead. That's why it's still an issue.

I too think that the real fix would be to make the framework not unbind properties. The unbind should happen on the field already. That's why my idea was to cast the property as an IList, so it gets ignored by the framework unbind code. It kinda does the same thing as the real fix but only for this property.

Null-propagating `HitObject` also works to prevent the exception, but I like this less since it adds more logic and it makes a reader question whether `HitObject` can be null in this class (which it can't under normal circumstances).

bdach:
Local bindable list or just nuking automatic unbind are both vastly better options, I agree.

PR #10903, opened 2020-11-19
Add ParticleExplosion and use to give legacy skins particle explosions

### LINKED ISSUE #6149: Particles are not implemented
**Describe the bug:** 
My skin "tranquil sublimity" isn't getting rendered properly. My "100" and "50" images aren't appearing and I don't get the aesthetic of my skin from stable.

*Stable:*
screenshot083
*Lazer:*
osu_2019-09-17_19-07-12

**Skin File:**
[tranquil sublimity.zip](https://github.com/ppy/osu/files/3623994/tranquil.sublimity.zip)

**osu!lazer version:** Latest

**Logs:** (not required)

peppy:
- [x] Depends on #10890
- Closes #6149

### DIFF
--- a/osu.Game/Graphics/ParticleExplosion.cs
+++ b/osu.Game/Graphics/ParticleExplosion.cs
@@ -60,7 +60,7 @@ namespace osu.Game.Graphics
                 double direction = RNG.NextSingle(0, MathF.PI * 2);
 
                 this.MoveTo(new Vector2(0.5f));
-                this.MoveTo(new Vector2(0.5f) + positionForOffset(0.5f), duration);
+                this.MoveTo(new Vector2(0.5f) + positionForOffset(RNG.NextSingle(0.5f)), duration);
 
                 this.FadeOutFromOne(duration);
                 Expire();
--- a/osu.Game/Graphics/ParticleExplosion.cs
+++ b/osu.Game/Graphics/ParticleExplosion.cs
@@ -20,7 +20,7 @@ namespace osu.Game.Graphics
             {
                 double rDuration = RNG.NextDouble(duration / 3, duration);
 
-                AddInternal(new Particle(rDuration, RNG.NextSingle(0, MathF.PI * 2))
+                AddInternal(new Particle(rDuration)
                 {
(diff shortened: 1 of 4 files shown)


osu.Game/Graphics/ParticleExplosion.cs
```diff
@@ -0,0 +1,144 @@
...
+                        new Vector2(InflationAmount.X / DrawRectangle.Width, InflationAmount.Y / DrawRectangle.Height),
+                        null, TextureCoords);
+                }
+            }
+        }
+
+        private readonly struct ParticlePart
+        {
+            private readonly double duration;
+            private readonly double direction;
```
turbedi, line 123: This can be changed to a `float` if I didn't miss something
 Then line 138 can be changed to
`
            return new Vector2(0.5f) + travelledDistance * new Vector2(MathF.Sin(direction), MathF.Cos(direction));
`

osu.Game/Graphics/ParticleExplosion.cs
```diff
@@ -0,0 +1,144 @@
...
+
+            parts.Clear();
+            for (int i = 0; i < particleCount; i++)
+                parts.Add(new ParticlePart(duration));
+        }
+
+        protected override void Update()
+        {
+            base.Update();
+            Invalidate(Invalidation.DrawNode);
```
bdach, line 58: maybe could add a guard here to not invalidate if `Time.Current <= startTime + duration`? not sure how much of a perf influence that will have, though.

bdach (commented):
looks pretty good to me, although i don't have much experience with `DrawNode`s yet

osu.Game/Graphics/ParticleExplosion.cs
peppy: I think I'm going to leave this as is just to be cautious. I'm thinking of cases where time jumps in large portions and causes an invalidation to be skipped.

Lifetimes of the explosion should be enough to avoid any performance issues.

smoogipoo (approved):

PR #10905, opened 2020-11-19
Implement hitsample pooling

smoogipoo:
This is mostly a performance benefit, which reduces `DrawableHitObject.Apply()` from 4.5% of total time to 0.9%, as result of not constantly loading and clearing samples.

Before:
Screenshot_2020-11-19_22-09-45

After:
Screenshot_2020-11-19_22-09-51

It also reduces allocations during gameplay, but they're pretty tiny anyway.

I've chosen to not limit the size of the pool since benchmarks don't reveal samples as too costly - they're absolutely dwarfed by everything else - followpoints, their associated drawnodes, imagesharp arraypools, etc. Samples account for maybe 10MB of the entire process across several maps I've tested. In particular of note:
- https://osu.ppy.sh/beatmapsets/150945#osu/372245 (performance - >100 samples from one single pool)
- https://osu.ppy.sh/beatmapsets/177187#osu/426550 (memory - fully keysounded)

A lot of the implementation of `SkinnableSound` has been extracted into `PoolableSkinnableSample`, which is the pooled component. `SkinnableSound` is given a new `Samples` property which is to be set by the `DrawableHitObject` when new samples should be loaded.  
tl;dr: the process is:
1. `DrawableHitObject` sets `SkinnableSound.Samples`.
2. `SkinnableSamples` clears all existing samples, looks up a respective `PoolableSkinnableSample` from the pool first (if any), and falls back to constructing a `PoolableSkinnableSample` itself if that failed.
3. `PoolableSkinnableSample` does the lookup for the `SampleChannel` only once between skin changes.


osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -34,21 +34,68 @@ public class SkinnableSound : SkinReloadableDrawable, IAdjustableAudioComponent
...
         protected bool PlayWhenZeroVolume => Looping;
 
-        protected readonly AudioContainer<DrawableSample> SamplesContainer;
+        protected readonly AudioContainer<PoolableSkinnableSample> SamplesContainer;
+
+        [Resolved]
+        private ISampleStore sampleStore { get; set; }
+
+        [Resolved(CanBeNull = true)]
+        private IPooledSampleProvider pooledProvider { get; set; }
```
peppy, line 43: maybe rename to `samplePool`? current name feels a bit weird.

peppy (changes requested):

osu.Game/Skinning/IPooledSampleProvider.cs
```diff
@@ -0,0 +1,14 @@
...
+
+using JetBrains.Annotations;
+using osu.Game.Audio;
+
+namespace osu.Game.Skinning
+{
+    public interface IPooledSampleProvider
+    {
+        [CanBeNull]
+        PoolableSkinnableSample GetPooledSample(ISampleInfo sampleInfo);
```
bdach, line 12: is this realistically public API that will need xmldoc?

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -336,6 +343,29 @@ DrawableHitObject IPooledHitObjectProvider.GetPooledDrawableRepresentation(HitOb
             });
         }
 
+        private readonly Dictionary<ISampleInfo, DrawablePool<PoolableSkinnableSample>> samplePools = new Dictionary<ISampleInfo, DrawablePool<PoolableSkinnableSample>>();
+
+        public PoolableSkinnableSample GetPooledSample(ISampleInfo sampleInfo)
+        {
+            if (!samplePools.TryGetValue(sampleInfo, out var existingPool))
+                AddInternal(samplePools[sampleInfo] = existingPool = new DrawableSamplePool(sampleInfo, 1));
```
bdach, line 351: slightly worried about the thread safety of this. sample changes after initial load i believe *should* be relatively safe because they should be occurring on the update thread, but how about the initial load? i believe the flow will be something like

```
DrawableHitObject.LoadCompleteAsync()
DrawableHitObject.Apply(HitObject, HitObjectLifetimeEntry)
DrawableHitObject.onSamplesChanged(object, NotifyCollectionChangedEventArgs)
DrawableHitObject.LoadSamples()
SkinnableSound.Samples.set
SkinnableSound.updateSamples()
Playfield.GetPooledSample(ISampleInfo) (via interface)
```

which i *think* is outside of the update loop. what i'm worried about is silently creating two pools due to lack of thread safety of the dictionary, but i'm not sure whether i'm being paranoid or not.

osu.Game/Skinning/PoolableSkinnableSample.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        public void RemoveAllAdjustments(AdjustableProperty type) => sampleContainer.RemoveAllAdjustments(type);
+
+        public IBindable<double> AggregateVolume => sampleContainer.AggregateVolume;
+
+        public IBindable<double> AggregateBalance => sampleContainer.AggregateBalance;
+
+        public IBindable<double> AggregateFrequency => sampleContainer.AggregateFrequency;
+
+        public IBindable<double> AggregateTempo => sampleContainer.AggregateTempo;
```
bdach, line 161: maybe `#region` this as is tradition?

osu.Game/Skinning/SkinnableSound.cs
```diff
@@ -128,6 +175,9 @@ public void RemoveAdjustment(AdjustableProperty type, BindableNumber<double> adj
         public void RemoveAllAdjustments(AdjustableProperty type)
             => SamplesContainer.RemoveAllAdjustments(type);
 
+        /// <summary>
+        /// Whether any samples currently playing.
```
bdach, line 179:
```suggestion
        /// Whether any samples are currently playing.
```

osu.Game/Storyboards/Drawables/DrawableStoryboardSample.cs
```diff
@@ -38,7 +38,7 @@ protected override void SkinChanged(ISkinSource skin, bool allowFallback)
             foreach (var mod in mods.Value.OfType<IApplicableToSample>())
             {
                 foreach (var sample in SamplesContainer)
-                    mod.ApplyToSample(sample);
+                    mod.ApplyToSample(sample.Sample);
```
bdach, line 41: seems slightly suspect as `Sample` is annotated as `CanBeNull` (can this realistically fail in a gameplay scenario? if not, should have assertions maybe?)

osu.Game/Audio/SampleInfo.cs
```diff
@@ -20,5 +22,16 @@ public SampleInfo(params string[] sampleNames)
         public IEnumerable<string> LookupNames => sampleNames;
 
         public int Volume { get; } = 100;
+
+        public override int GetHashCode()
+        {
+            return HashCode.Combine(sampleNames, Volume);
```
bdach, line 28: this seems slightly dodge as `Equals` checks sequence equality while `GetHashCode()` will will do a reference-based hashcode. i think what might be better here is `StructuralComparisons.StructuralEqualityComparer.GetHashCode()` ([it exists!](https://docs.microsoft.com/en-us/dotnet/api/system.collections.structuralcomparisons.structuralequalitycomparer?view=net-5.0))

osu.Game/Skinning/IPooledSampleProvider.cs
peppy: this is the kind of interface i'd probably argue to be internal as i don't think rulesets should be providing their own pooling backend for samples.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
```diff
@@ -40,7 +41,7 @@ public class DrawableSlider : DrawableOsuHitObject, IDrawableHitObjectWithProxie
         private Container<DrawableSliderTail> tailContainer;
         private Container<DrawableSliderTick> tickContainer;
         private Container<DrawableSliderRepeat> repeatContainer;
-        private Container<PausableSkinnableSound> samplesContainer;
+        private PausableSkinnableSound slidingSample;
```
peppy, line 44: didn't this just become a container last week? ;)

osu.Game/Rulesets/UI/Playfield.cs
peppy: I don't believe we ever load `DrawableHitObjects` asynchronously (correct me if i'm wrong on this one), so maybe a non-issue right now, but still something worth considering.

At very least, may want to add a `ThreadSafety.EnsureUpdateThread` call here.

peppy (commented):
Looks mostly fine, just waiting for existing review to be applied.

osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSlider.cs
smoogipoo: It did yeah, but it doesn't need to be disposed or removed anymore.

peppy:
Is this ready for review now? Noticed you said you may be making further changes.

smoogipoo:
Nope, still working through the hitsample stuff.

osu.Game/Rulesets/UI/Playfield.cs
peppy: I think this is fine until it comes up as an issue (crash). The scenario explained by @bdach is a non-issue because at the point of initial load the player hierarchy has not been committed to a root, so these operations are allowed. Also, it should be running the full load sequence on a single thread.

peppy:
While I think this code feels fine, it's another example of something which *really* requires a wiki page documenting how to use for ruleset creators. I don't find it 100% obvious that setting the `.Samples` on one of these instances is enough to make use of pooling.

Nor is the requirement of nulling the sample after use that obvious. Arguable, adding a `Free()` method may be more friendly? Not sure..

peppy (approved):

peppy:
On a second pass I think it's fine. Definitely need wiki docs but that goes for the rest of the ruleset API too. Consumers can still use the non-pooled method for custom samples and can read the internal code if they're interested in using the new pooling setup.

PR #10918, opened 2020-11-20
Remove default allowRetry parameter value from ResultsScreen

peppy:
There were a couple of cases (fixed in this PR) which incorrectly used the default value for this parameter where they shouldn't have been. Another case where having a default is likely not a good idea as it can be easily overlooked and provide incorrect behaviour.


smoogipoo (dismissed):

smoogipoo:
Real test failure here.

bdach:
Resolving the test failure seems simple enough (specifying `true` instead of `false` in `TestSceneResultsScreen.UnrankedSoloResultsScreen` resolves), but the `allowRetry` default is still in the base `ResultsScreen`:

https://github.com/ppy/osu/blob/2db42f8e6780eb2fd67854b54413ce8506a2895a/osu.Game/Screens/Ranking/ResultsScreen.cs#L60-L66

I assume it was not intended to remain?

bdach:
I've pushed the changes described above. @peppy check if you're OK with them whenever you can.

peppy:
Looks good, thanks!

bdach (approved):

PR #10923, opened 2020-11-20
Implement follow point pooling

smoogipoo:
This is basically a complete rewrite of `FollowPointRenderer` to delegate instantiation through `FollowPointLifetimeEntry`, in a similar way as done for `DrawableHitObject`s.

The pool limits are fairly lax, but I think that's alright and at maximum capacity followpoints + connections only account for 10MB of memory usage overall. Centipede gets pretty close to/exceeds the maximum capacities I've set.

Top memory usage now after a run of Centipede (sitting at on result screen):
Screenshot_2020-11-20_17-08-17


smoogipoo:
image

Ordered by Bytes instead.

peppy (approved):
Looks and seems to work as expected 👍 .

PR #10924, opened 2020-11-20
Add method of initialising a `DrawableHitObject` from `Playfield` regardless of pooling

ekrctb:
When existing `Playfield` implementations want to apply something to every `DrawableHitObject`, they had to either override `Add` or using a custom `DrawablePool`, depending on whether the hit objects are pooled.
However, `Add` is not called for nested `DrawableHitObject`s while for pooled hit objects nested objects are treated equally.

This PR aims to simplify code by introducing new virtual method to `Playfield`.

The custom pools in the standard ruleset and its editor are now unneeded and removed.
I also used the new event in `CatchPlayfield`. But I didn't have enough confidence to use the new event in other rulesets.


peppy (dismissed):
Seems like a pretty good code quality change.

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -91,6 +99,7 @@ protected Playfield()
             {
                 h.NewResult += (d, r) => NewResult?.Invoke(d, r);
                 h.RevertResult += (d, r) => RevertResult?.Invoke(d, r);
+                h.DrawableHitObjectAdded += d => DrawableHitObjectAdded?.Invoke(d);
```
smoogipoo, line 102: I'm not sure about this, is there a case you can imagine for this to be used? I'd advise this to only be used for setup purposes and so it should correspond to the Playfield the DHO was added to.

osu.Game/Rulesets/UI/HitObjectContainer.cs
```diff
@@ -147,6 +157,16 @@ public virtual void Add(DrawableHitObject hitObject)
...
             AddInternal(hitObject);
+
+            onDrawableHitObjectAddedRecursive(hitObject);
+        }
+
+        private void onDrawableHitObjectAddedRecursive(DrawableHitObject hitObject)
+        {
+            DrawableHitObjectAdded?.Invoke(hitObject);
+
+            foreach (var nested in hitObject.NestedHitObjects)
```
smoogipoo, line 168: This is weird, because it's not going to trigger on new nested DHOs, e.g. after sliders paths are changed to add more ticks. I expect this to be more important in catch than osu! with the changes made in this PR, so it deserves attention.

I think the source of even here should be from inside the DHO itself.

smoogipoo (changes requested):

ekrctb:
I reworked the implementation and now the event has more concrete semantics.
The event is called once for each unique `DrawableHitObject` and it can be used for initialization.

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -325,6 +340,8 @@ DrawableHitObject IPooledHitObjectProvider.GetPooledDrawableRepresentation(HitOb
                 // This is done before Apply() so that the state is updated once when the hitobject is applied.
                 if (!dho.IsLoaded)
```
ekrctb, line 341: So, if this comment is misleading as if the DHO is added and removed from the `Playfield` in one frame then this `if` clause can run multiple times.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
```diff
@@ -55,21 +55,18 @@ public CatchPlayfield(BeatmapDifficulty difficulty, Func<CatchHitObject, Drawabl
...
-            h.OnRevertResult += onRevertResult;
-
-            base.Add(h);
-
-            var fruit = (DrawableCatchHitObject)h;
-            fruit.CheckPosition = CheckIfWeCanCatch;
+            ((DrawableCatchHitObject)d).CheckPosition = CheckIfWeCanCatch;
         }
 
+        public bool CheckIfWeCanCatch(CatchHitObject obj) => CatcherArea.AttemptCatch(obj);
```
bdach, line 68: This doesn't need to be `public` anymore.

osu.Game.Rulesets.Osu/Edit/DrawableOsuEditRuleset.cs
```diff
@@ -26,8 +29,47 @@ private class OsuEditPlayfield : OsuPlayfield
...
             protected override GameplayCursorContainer CreateCursor() => null;
 
-            protected override DrawablePool<TDrawable> CreatePool<TDrawable>(int initialSize, int? maximumSize = null)
-                => new DrawableOsuEditPool<TDrawable>(CheckHittable, OnHitObjectLoaded, initialSize, maximumSize);
+            protected override void OnNewDrawableHitObject(DrawableHitObject d)
+            {
+                d.ApplyCustomUpdateState += updateState;
+            }
+
+            private const double editor_hit_object_fade_out_extension = 700;
```
bdach, line 37: I'd move over the xmldoc that was removed from `DrawableOsuEditPool` here.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -68,6 +69,29 @@ public OsuPlayfield()
             NewResult += onNewResult;
         }
 
+        protected override void OnNewDrawableHitObject(DrawableHitObject drawable)
+        {
+            ((DrawableOsuHitObject)drawable).CheckHittable = CheckHittable;
```
bdach, line 74: `OsuPlayfield.CheckHittable` also has no reason to be `public` anymore.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -114,27 +129,6 @@ protected override void OnHitObjectRemoved(HitObject hitObject)
...
-                case DrawableSliderTick _:
-                case DrawableSliderRepeat _:
-                case DrawableSpinnerTick _:
-                    break;
-
-                case DrawableSpinner _:
-                    spinnerProxies.Add(drawable.CreateProxy());
-                    break;
-
-                case IDrawableHitObjectWithProxiedApproach approach:
```
bdach, line 132: `IDrawableHitObjectWithProxiedApproach` effectively has no uses anymore with these changes.

osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -141,6 +146,11 @@ public abstract class DrawableHitObject : SkinReloadableDrawable
 
         private Container<PausableSkinnableSound> samplesContainer;
 
+        /// <summary>
+        /// Whether the initialization logic in <see cref="Playfield" /> has applied.
+        /// </summary>
+        internal bool HasInitialized;
```
bdach, line 152: nit: would probably rename to `IsInitialised`

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -124,6 +135,11 @@ private void load()
         /// <param name="h">The DrawableHitObject to add.</param>
         public virtual void Add(DrawableHitObject h)
         {
+            if (h.HasInitialized)
+                throw new InvalidOperationException($"{nameof(Playfield.Add)} doesn't support {nameof(DrawableHitObject)} reuse. Use pooling instead.");
```
bdach, line 139: class name qualification is redundant

```suggestion
                throw new InvalidOperationException($"{nameof(Add)} doesn't support {nameof(DrawableHitObject)} reuse. Use pooling instead.");
```

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -113,6 +114,16 @@ private void load()
             }
         }
 
+        private void onNewDrawableHitObject(DrawableHitObject d)
+        {
+            d.OnNestedDrawableCreated += onNewDrawableHitObject;
+
+            OnNewDrawableHitObject(d);
+
+            Debug.Assert(!d.HasInitialized);
```
bdach, line 123: shouldn't this assert be at the start of the method to avoid double-subscribing to `OnNestedDrawableCreated`?

osu.Game/Rulesets/UI/Playfield.cs
```diff
@@ -157,6 +173,17 @@ protected virtual void OnHitObjectRemoved(HitObject hitObject)
         {
         }
 
+        /// <summary>
+        /// Invoked before a new <see cref="DrawableHitObject"/> is added to this <see cref="Playfield"/>.
+        /// It is invoked only once even the drawable is pooled and used multiple times for different <see cref="HitObject"/>s.
```
bdach, line 178:
```suggestion
        /// It is invoked only once even if the drawable is pooled and used multiple times for different <see cref="HitObject"/>s.
```

bdach (dismissed):
Core change seems like a code quality improvement but in my view this PR doesn't go far enough and leaves some loose ends untied.

osu.Game/Rulesets/UI/Playfield.cs
ekrctb: If the assertion triggers that is a bug of the code (not meant to be a recoverable error) and such a double subscription shouldn't be matter. Then, I think it is good to group the check and the set together to eliminate thoughts like what if `HasInitialized` is set in `OnNewDrawableHitObject` call above.

bdach: I don't think anybody other than `Playfield` should really ever be mutating `IsInitialised`, though. That's another reason it makes more sense to me to guard at the start but I guess it doesn't matter much.

peppy (approved):

smoogipoo (approved):
lgtm

PR #10925, opened 2020-11-20
Fix skinnable sounds playing more than once on skin change

peppy:
I thought I already fixed this one recently...

- [x] Depends on #10905 for ease of merging


smoogipoo:
Keep in mind that a significant amount of this logic, including restart logic, was moved inside `PoolableSkinnableSample`. Is that implementation still fine? Or does it also need the `!Looping` check?

peppy:
I didn't double check this after moving it across (only on master), probably needs to be there too.

bdach:
FYI I'm holding off for a bit until the dependent PR gets merged before I review this.

bdach:
Do you recall what the repro scenario for this is / whether this was in response to an open issue? I'm guessing it has to do with looping samples, although I'm not sure I understand why those shouldn't be restarted on skin change...

peppy:
just smash the random skin hotkey while watching a replay. if it's still broken, the same sounds will play on already hit objects more than once.

bdach:
Oh, I read the condition backwards, my bad. Will double-check master and this.

osu.Game/Skinning/PoolableSkinnableSample.cs
```diff
@@ -102,7 +102,7 @@ private void updateSample()
             sampleContainer.Add(Sample = new DrawableSample(ch) { Looping = Looping });
 
             // Start playback internally for the new sample if the previous one was playing beforehand.
-            if (wasPlaying)
+            if (wasPlaying && !Looping)
```
bdach, line 105: Upon actually checking, I actually don't think I originally misread and this should probably be

```suggestion
            if (wasPlaying && Looping)
```

right? The looping samples need to be restarted to continue playing, but the non-looping ones don't. It seems this logic can actually result in losing looping samples if I switch from skin A to B and then back to A again (i.e. during a spinner).

peppy: It does indeed look back to front.

bdach: In fact this actually might need to even be `wasPlaying || Looping` as I'm hitting some skins here that have `wasPlaying` set to apparently false despite being looping samples. I'm not yet 100% on what the cause of that would be, though.

bdach: Never mind, the case I'm hitting is caused by some frivolous audio handling framework-side without error checking.

peppy:
Have double checked and the new behaviour sounds correct in gameplay.

smoogipoo (approved):

PR #10928, opened 2020-11-20
Fix silent NRE in slider selection blueprint

smoogipoo:
Supersedes and closes https://github.com/ppy/osu/pull/10900

As mentioned in the issue, this is hidden by `Drawable`'s try-catch around the unbind process. Whether the unbind process should consider properties (probably not) is a framework-side issue (https://github.com/ppy/osu-framework/issues/4014).


bdach (approved):
seems fine - CI failure surely not related

PR #10932, opened 2020-11-21
Split out profile sub-section component

EVAST9919:
Split from https://github.com/ppy/osu/pull/10842

### DIFF
--- a/osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
+++ b/osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
@@ -5,8 +5,6 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
-using osu.Game.Graphics;
-using osu.Game.Graphics.Sprites;
 using osu.Game.Users;
 using JetBrains.Annotations;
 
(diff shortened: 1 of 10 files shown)


osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
```diff
@@ -0,0 +1,78 @@
...
+using osu.Game.Users;
+using JetBrains.Annotations;
+
+namespace osu.Game.Overlays.Profile.Sections
+{
+    public abstract class ProfileSubsection : FillFlowContainer
+    {
+        protected readonly Bindable<User> User = new Bindable<User>();
+
+        protected RulesetStore Rulesets { get; private set; }
```
bdach, line 20: could remove this param from BDL and just annotate this field with `[Resolved]`

osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        protected OsuSpriteText Missing { get; private set; }
```
bdach, line 22: is there a reason why this is exposed in the base as protected? can't this just live as private in `PaginatedProfileSubsection`? this class does nothing with this sprite text.

if this is moved back there, the `missingText` constructor parameter should be, as well.

osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
```diff
@@ -0,0 +1,78 @@
...
+                },
+            };
+
+            Rulesets = rulesets;
+        }
+
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+            User.BindValueChanged(OnUserChanged, true);
```
bdach, line 66: really weird how both the bindable and `OnUserChanged` are both exposed as protected. only one or the other should be, really.

(ideally the user should be resolved via DI, actually, but i'm not going to insist on that.)

bdach (changes requested):

osu.Game/Overlays/Profile/Sections/ProfileSubsection.cs
EVAST9919: We need User protected because it's ID used for request creation (for example in `PaginatedScoreContainer`)

bdach: then remove `OnUserChanged` and just `BindValueChanged` locally wherever needed?

bdach (approved):
seems reasonable

PR #10935, opened 2020-11-21
Rewind judgement transforms before clearing

### LINKED ISSUE #10933: disappearing judgements on mania
**Describe the bug:** judgements on mania seems to disappear and reappear from time to time. seems to happen on default and custom skin. not tested on other rulesets  tho

**Screenshots or videos showing encountered issue:** https://drive.google.com/file/d/1c-UBbj467lkhnJ56Dj5iYAcXMlub3XKT/view?usp=sharing

**osu!lazer version:** 2020.1121.0

**Logs:** ithink not needed

bdach:
Resolves #10933.

Regressed in https://github.com/ppy/osu/pull/10890 (unsurprisingly). The immediate cause of the bug was that `DrawableManiaJudgement` used a `FadeOut` instead of a `FadeOutFromOne`, therefore polluting a pooled judgement's state after the first time the animation plays out:

https://github.com/ppy/osu/blob/194f2e621b75003ed1a5027a33b49eeb7c679e7f/osu.Game.Rulesets.Mania/UI/DrawableManiaJudgement.cs#L49-L51

but after some thought it occurred to me that fixing locally is definitely too low a level, as this is bound to happen again with different properties of other pooled judgements. Therefore the proposed resolution is to leverage the fact that judgement transforms are now rewindable since the linked PR and rewind to a sane state before proceeding with clearing, in a similar way that `DrawableHitObject` does.

This is turning out to be quite a popular type of bug, so maybe even this resolution is on too low a level and we should be looking into embedding it into `DrawablePool` framework-side. The type of resolution that is done here would only be applicable for drawables with rewindable transforms, however, so it might be a bit awkward to move this behaviour to framework as it'll be a little edge-casey.

Not sure what sort of automated testing I could do here, but I could include some if requested.

### DIFF
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -149,8 +149,10 @@ namespace osu.Game.Rulesets.Judgements
 
         private void runAnimation()
         {
+            // undo any transforms applies in ApplyMissAnimations/ApplyHitAnimations to get a sane initial state.
             ApplyTransformsAt(double.MinValue, true);
             ClearTransforms(true);
+
             LifetimeStart = Result.TimeAbsolute;
 
             using (BeginAbsoluteSequence(Result.TimeAbsolute, true))
--- a/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
+++ b/osu.Game/Rulesets/Judgements/DrawableJudgement.cs
@@ -149,6 +149,7 @@ namespace osu.Game.Rulesets.Judgements
 
         private void runAnimation()
         {
+            ApplyTransformsAt(double.MinValue, true);
             ClearTransforms(true);
             LifetimeStart = Result.TimeAbsolute;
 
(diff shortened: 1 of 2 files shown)


peppy:
Yeah I was aware of this being a potential failure vector (in that PR i already fixed multiple instances of the same thing). I did consider fixing via rewind but wasn't sure it was the correct direction (and still am not). Feels like for cases where state isn't changing (and reuse is common) the implementation should just be responsible to returning things to a correct state?

Like as you say, maybe this is an issue with the pool implementation itself (some kind of reset logic should be present in `PrepareForUse`?)

Have to have a  think about this one.

bdach:
> Feels like for cases where state isn't changing (and reuse is common) the implementation should just be responsible to returning things to a correct state?

I'm not sure what do you mean by "state" in that sentence. If state isn't changing, then there's nothing to restore, and transforms *do* actually change state by the virtue of how they're implemented (mutating underlying properties of a transformable), especially in the non-rewindable case - which is why I think rewind is vastly the better option from a usabillity standpoint. What would the argument for manually undoing be in this case - performance?

peppy:
I mean in the case or DrawableHitObjects, the rewind makes sense because it is rewinding the transforms to the point in time where are hit or miss is decided (potentially only to that point), meaning the initial animation portion may remain.

In the case of judgements, their state is decided at the point they are retrieved from the pool and doesn't change until they are freed. Rewinding to an arbitrary point in time just feels a bit weird from that angle.

peppy:
After further investigation, it seems like `DrawableHitObject` is actually doing exactly the same thing as this. I think it's fine to leave as-is for now, although we may need to reconsider when we get to doing a transform optimisation pass on gameplay (this is quite a hot path and the LINQ incurred by these calls is non-negligible).

peppy (approved):

PR #10936, opened 2020-11-21
Fix legacy follow point animations not playing after a while

### LINKED ISSUE #10930: Follow points don't behave properly
**Describe the bug:** 
I notice that the follow points have some type of delay or they don't appear at all, in this version especially 
https://www.youtube.com/watch?v=bk8LDO8HbcE&feature=youtu.be
**Skins**
http://www.mediafire.com/file/qg02ky12906g8f8/skins.zip/file
**osu!lazer version:** 
2020.1121.0

bdach:
Resolves #10930.

"Regressed" in #10923, although the underlying structure was not really helping spot this earlier. In short, the following set failed for pooled follow points:

https://github.com/ppy/osu/blob/f77ca4ac3accf22e6c313d7b2e5302e0665db969/osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs#L83

because the value of `PlaybackPosition` is only ever read in `LoadComplete()` in current `master`:

https://github.com/ppy/osu/blob/194f2e621b75003ed1a5027a33b49eeb7c679e7f/osu.Game/Skinning/LegacySkinExtensions.cs#L78-L87

To resolve, `IAnimationTimeReference.AnimationStartTime` is made bindable and `PlaybackPosition` is updated on every change of that bindable. I also briefly considered running in `Update()` and think this solution is slightly better, but if you have differing opinions, I can also go for that instead.

### DIFF
--- a/osu.Game/Skinning/LegacySkinExtensions.cs
+++ b/osu.Game/Skinning/LegacySkinExtensions.cs
@@ -4,6 +4,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Allocation;
+using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Animations;
 using osu.Framework.Graphics.OpenGL.Textures;
@@ -70,6 +71,8 @@ namespace osu.Game.Skinning
             [Resolved(canBeNull: true)]
             private IAnimationTimeReference timeReference { get; set; }
 
+            private readonly Bindable<double> animationStartTime = new BindableDouble();
+
             public SkinnableTextureAnimation(bool startAtCurrentTime = true)
                 : base(startAtCurrentTime)
             {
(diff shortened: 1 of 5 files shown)


osu.Game/Skinning/LegacySkinExtensions.cs
```diff
@@ -82,8 +85,18 @@ protected override void LoadComplete()
                 if (timeReference != null)
                 {
                     Clock = timeReference.Clock;
-                    PlaybackPosition = timeReference.Clock.CurrentTime - timeReference.AnimationStartTime;
+                    ((IBindable<double>)animationStartTime).BindTo(timeReference.AnimationStartTime);
```
peppy, line 88: Might be best to expose as a `Bindable` instead (in the interface) if this is meant to be a mutable bindable.

peppy (changes requested):

osu.Game/Skinning/LegacySkinExtensions.cs
bdach: The intention was to allow mutability in whoever is controlling the animation and disallow mutability in the animation itself by scoping to the interface, but I guess I can go either way.

peppy: I guess it's neither here nor there. `Bindable`/`IBindable` still remaining a strange pattern in all usages

peppy (approved):

PR #10937, opened 2020-11-21
Don't scroll NowPlayingOverlay background while it's fading out

### LINKED ISSUE #8393: Now playing overlay sometimes has to scroll to correct background when opened

LittleEndu:
Closes #8393
Supersedes and closes #8397

I think the real issue here is that I was hitting the next song in song select before the overlay was fully faded out, and not ``audioEquals`` as peppy thought [here](https://github.com/ppy/osu/issues/8393#issuecomment-602091026). I debugged ``audioEquals`` a little bit just in case and it seems to work correctly.

Testing the PR out here: [youtube link](https://youtu.be/tTPfdWaOJ_o). Double clicking on the random bottom, first click closes the overlay, second picks next song while the overlay is (hopefully, the transition is 800ms) still fading out.

Can create tests if deemed necessary by adapting from #8397 to fit this behaviour instead.

### DIFF
--- a/osu.Game/Overlays/NowPlayingOverlay.cs
+++ b/osu.Game/Overlays/NowPlayingOverlay.cs
@@ -212,11 +212,14 @@ namespace osu.Game.Overlays
             dragContainer.ScaleTo(1, transition_length, Easing.OutElastic);
         }
 
+        private bool fadingOut;
+
         protected override void PopOut()
         {
+            fadingOut = true;
             base.PopOut();
 
-            this.FadeOut(transition_length, Easing.OutQuint);
+            this.FadeOut(transition_length, Easing.OutQuint).OnComplete(_ => fadingOut = false);
             dragContainer.ScaleTo(0.9f, transition_length, Easing.OutQuint);
         }
 
@@ -232,7 +235,7 @@ namespace osu.Game.Overlays
         {
             base.Update();
 
-            if (pendingBeatmapSwitch != null)
+            if (pendingBeatmapSwitch != null && !fadingOut)
             {
                 pendingBeatmapSwitch();
                 pendingBeatmapSwitch = null;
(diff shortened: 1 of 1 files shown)


bdach:
i'm not sure what was achieved here off of the attached video? the background still scrolls.

LittleEndu:
> The movement is intentional.

Quote from peppy from here: https://github.com/ppy/osu/issues/8393#issuecomment-602091026

bdach:
i'm incredibly confused then. the pr title states "don't scroll", the video shows scrolling, and you're claiming that this is intended. to put it bluntly: what is the exact difference between master and this PR? where is it visible on the video? and if it isn't visible on the video, what was the point of including it if it shows behaviour already in master?

LittleEndu:
> what is the exact difference between master and this PR?

Background is supposed to scroll when you open the overlay if you have changed the song at any point while the overlay was closed. In master, the background would scroll while the overlay is fading out, and if the user opened the overlay then after, the background would have already scrolled. This behaviour made me think that the scrolling was the unintentional part (to me it looks weird when the overlay both scrolls the background and fades in at the same time). 

The easing of the fade is ``Easing.OutQuint`` so to the user the overlay already looks like it was closed even though it might still be ever so slightly visible. I don't remember how I first hit this but the apparent inconsistency of this behaviour instantly weirded me out, that's the only reason I want this fixed.

peppy:
I'm not sure the animation behaviour should be completely disabled during the fade out (feels like the UI is just not responding to changes at that point?).

Not really sure this change is required at all..

peppy:
Closing this for reasons mentioned in last comment.

LittleEndu:
> the UI is just not responding to changes at that point

UI most certainly isn't, but keyboard shortcuts still make it respond and animate the overlay while it's fading out

PR #10941, opened 2020-11-22
Improve LogoVisualisation looks when MenuGlow colors are used

### LINKED ISSUE #10875: Visualizer bars around the osu! logo have visible Moiré patterns
**Describe the bug:** 
On skins that have ``MenuGlow`` set to high luminosity values, visualizer bars around the osu! logo have visible [Moiré patterns](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern). These patterns change slightly depending on song intensity (I think kiai time changes the amount of bars and makes the patter more visible) and view parallax.

``MenuGlow: 255, 255, 255`` on these screenshots (what I have set for my skin)

**osu!lazer version:** 
2020.1114.0

LumpBloom7:
Resolves #10875

With this change, skin colours won't be as bright as they are in the current release, and it avoid the problem of the moire effects mentioned in the linked issue.

Before(left) and after:
tmp

Default on latest release as baseline:
Screenshot (477)

### DIFF
--- a/osu.Game/Screens/Menu/LogoVisualisation.cs
+++ b/osu.Game/Screens/Menu/LogoVisualisation.cs
@@ -20,6 +20,7 @@ using osu.Framework.Audio;
 using osu.Framework.Audio.Track;
 using osu.Framework.Bindables;
 using osu.Framework.Utils;
+using osu.Framework.Extensions.Color4Extensions;
 
 namespace osu.Game.Screens.Menu
 {
@@ -67,7 +68,7 @@ namespace osu.Game.Screens.Menu
 
         private int indexOffset;
 
-        public Color4 AccentColour { get; set; }
+        public Color4 AccentColour { get; set; } = Color4.White.Opacity(.2f);
 
         /// <summary>
         /// The relative movement of bars based on input amplification. Defaults to 1.
--- a/osu.Game/Screens/Menu/LogoVisualisation.cs
+++ b/osu.Game/Screens/Menu/LogoVisualisation.cs
@@ -174,7 +174,7 @@ namespace osu.Game.Screens.Menu
             // Assuming the logo is a circle, we don't need a second dimension.
             private float size;
 
-            private readonly Color4 colour = Color4.White.Opacity(.2f);
+            private readonly Color4 colour = Color4.White.Opacity(0.2f);
 
             private float[] audioData;
 
--- a/osu.Game/Screens/Menu/LogoVisualisation.cs
+++ b/osu.Game/Screens/Menu/LogoVisualisation.cs
@@ -11,7 +11,6 @@ using osu.Framework.Graphics.Primitives;
 using osu.Framework.Graphics.Shaders;
 using osu.Framework.Graphics.Textures;
 using osu.Game.Beatmaps;
-using osu.Game.Graphics;
 using System;
 using System.Collections.Generic;
 using JetBrains.Annotations;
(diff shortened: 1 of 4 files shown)


LumpBloom7:
Some more comparisons.
Before: 
Screenshot (479)

After:
Screenshot (480)

osu.Game/Screens/Menu/LogoVisualisation.cs
```diff
@@ -176,7 +174,8 @@ private class VisualisationDrawNode : DrawNode
             // Assuming the logo is a circle, we don't need a second dimension.
             private float size;
 
-            private Color4 colour;
+            private readonly Color4 colour = Color4.White.Opacity(.2f);
```
peppy, line 177:
```suggestion
            private readonly Color4 colour = Color4.White.Opacity(0.2f);
```

osu.Game/Screens/Menu/OsuLogo.cs
```diff
@@ -144,7 +144,6 @@ public OsuLogo()
                                                     RelativeSizeAxes = Axes.Both,
                                                     Origin = Anchor.Centre,
                                                     Anchor = Anchor.Centre,
-                                                    Alpha = 0.5f,
```
peppy, line 147: what's the deal with these? were they never applying due to missing draw node logic?

LumpBloom7: They were applying correctly, but is less obvious when used with brighter colours. 

When using `#ff0064` (a rather dark colour), this applied on top of the 0.2 opacity of the actual bars, the final result was way too transparent. I chose to just remove it, since the relative difference when using brighter colours is negligible.

peppy: that doesn't make sense though. going to require further investigation.

LumpBloom7: The reasoning I had was to make lower luminosity colours a bit more visible with this change, and it's pretty much just that. But after a bit of staring, I'm kinda feeling that the blame should instead lie with the colors used, because trying to accommodate every single colour would be endless.

RBRat3: I still question the merits of this being an issue, unless there are plans to expose an alpha value to users in the config but as it currently stands users have access to the top end of that additive sprite and adding more alpha robs users of that especially if they're using a custom texture. If they needed more transparency they could have lowered the value like in the original issue to something like 127 across the board (black = transparency). That background also wasn't doing them any favors adding white against a light blue sky so the fix here might not even help. Stable has the same moire issue and alpha just hides the problem but it's not what I consider a fix, whole visualizer probably needs to be built in a buffer before an aa pass to properly get rid of it but that's beyond my skillset and probably outside the current scope of things.

peppy (dismissed):

peppy:
I've adjusted the opacity back to what it was on master (I think it looks better). Other than that all seems okay.

ColdVolcano:
isn't this: https://github.com/ppy/osu/blob/ae609b9d48c50ae6f7ad291f191ee76309fda123/osu.Game/Screens/Menu/OsuLogo.cs#L303-L306 going to make the alpha back to 0.5 like it was before, making it seemingly impossible to see the visualiser now?

peppy:
Good catch. Completely breaks that behaviour. I think for now it might be fine to just remove the flashing logic, since i've never really noticed it actually taking effect.

peppy:
I've restored the flashing behaviour because it seems like something we probably want.

peppy (approved):

PR #10943, opened 2020-11-23
Scroll editor setup screen to file selector on display

### LINKED ISSUE #10942: Make song setup screen in editor scroll to track select when clicked
**Describe the new feature:** 
It's a little weird trying to click the track select button and then scrolling down to see all the folders. What I propose is making the setup screen scroll down to the track select when clicking on it.

When clicking this:
image
Make the setup screen scroll to the track select:
image

peppy:
Previously the file selector would potentially display off-screen, making for confusing UX.

Closes #10942.

### DIFF
--- a/osu.Game/Screens/Edit/Setup/FileChooserLabelledTextBox.cs
+++ b/osu.Game/Screens/Edit/Setup/FileChooserLabelledTextBox.cs
@@ -61,7 +61,7 @@ namespace osu.Game.Screens.Edit.Setup
                 CurrentFile = { BindTarget = currentFile }
             };
 
-            sectionsContainer?.ScrollTo(fileSelector);
+            sectionsContainer.ScrollTo(fileSelector);
         }
 
         internal class FileChooserOsuTextBox : OsuTextBox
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Edit/Setup/FileChooserLabelledTextBox.cs
```diff
@@ -47,14 +52,16 @@ private void onFileSelected(ValueChangedEvent<FileInfo> file)
...
+            Target.Child = fileSelector = new FileSelector(validFileExtensions: ResourcesSection.AudioExtensions)
             {
                 RelativeSizeAxes = Axes.X,
                 Height = 400,
-                Anchor = Anchor.Centre,
-                Origin = Anchor.Centre,
                 CurrentFile = { BindTarget = currentFile }
             };
+
+            sectionsContainer?.ScrollTo(fileSelector);
```
bdach, line 64: what's up with this nullcheck? the resolved property does not allow nulls, and surely this can't fire before BDL/injection?

peppy: ￼oh i meant to make it allow null, just in case this component is used in such a way.

smoogipoo (approved):

PR #10950, opened 2020-11-24
Support hit object pooling in ScrollingPlayfield

ekrctb:
`ScrollingHitObjectContainer` is currently incompatible to hit object pooling as it relies on `Add` and `Remove` overrides.
This PR aims to add hit object pooling support to `ScrollingHitObjectContainer` in a way that both pooled and non-pooled hit objects use common code as much as possible, and simplify the logic.

After several tries to write the logic, I believe I now have a relatively simple way to manage hit objects that don't break layout computation delaying optimization #5951.
I modified `HitObjectUsageBegan` and `HitObjectUsageFinished` to use it. It is a breaking change.


ekrctb:
I'm not very sure about changes on `HitObjectUsageBegan` and `HitObjectUsageFinished`. And it is not invoked for non-pooled drawables?
I'm confused what event to subscribe and what is invoked for pooled drawables and non-pooled drawables.

smoogipoo:
They're internal events specifically for the editor to know when a pooled DHO has become part of the hierarchy to attach a `SelectionBlueprint` for it.

See `ComposeBlueprintContainer.CreateBlueprintFor()` - it enumerates all DHOs to find the relevant one through `Playfield.AllHitObjects`.

smoogipoo:
This seems to break moving hold notes in the editor - the tail becomes positioned on top of the head.

I'm actually not sure how it's even working in the editor, to be honest... The code in `ScrollingPlayfield` looks insufficient to handle the editor - `DefaultsApplied` should be used instead but it also doesn't recurse through nested DHOs. On debugging I can see `updateLayoutRecursive()` being hit, but I don't know what causes `layoutComputed` to get cleared - neither `Clear()` nor `InvalidateHitObject`, nor the clear from inside `Update()` are being hit. Could you guide me as to how this is working?

ekrctb:
@smoogipoo Indeed lifetime is not updated in editor.
The only layout invalidation in mania editor is by <https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Mania/Edit/ManiaBeatSnapGrid.cs#L104> (I didn't know "grid lines" is treated as hit objects).

ekrctb:
I finally figured out why it didn't work in mania.
Mania uses nested `Column`s.
`Column` overrides `Add`, but it didn't call `base.Add`. `Playfield.Add` contains logic of `OnNewDrawableHitObject`.
Because I relied on `OnNewDrawableHitObject` to register invalidation events, the hit objects were never invalidated in mania.

smoogipoo:
That's so dodge :/ I really want `Add` to be sealed and even originally had it that way for the HOLE path, but mania stuck a thorn in it too...

osu.Game/Rulesets/UI/HitObjectContainer.cs
```diff
@@ -178,6 +182,26 @@ protected override void OnChildLifetimeBoundaryCrossed(LifetimeBoundaryCrossedEv
...
+        /// </remarks>
+        protected virtual void OnAdd(DrawableHitObject drawableHitObject)
+        {
+        }
+
+        /// <summary>
+        /// Invoked when a <see cref="DrawableHitObject"/> is removed from this container.
+        /// </summary>
+        /// <remarks>
+        /// This method is not invoked for nested <see cref="DrawableHitObject"/>s.
```
smoogipoo, line 199: I think this may become an issue with the editor, but let's go with this for now and reconsider if that comes up.

smoogipoo (dismissed):
lgtm

peppy:
image

This `HashSet` looks to grow indefinitely in the case of non-pooled rulesets. Is that intended? In the case of pooled rulesets I'm guessing this is not an issue?

ekrctb:
@peppy The set can contain all `DHO`s, which is intended. But if the set contains already removed objects, then that is a memory leak and it is a bug, as the removed hit objects should be removed from the set as well, by this code <https://github.com/ppy/osu/pull/10950/files#diff-f6c2da17badc7994f7f260aa742e786feb77b03eec74a940d890910b486c3f16R179>.
Could you clarify what situation you are referring to?

peppy:
I was just testing with mania in a test scene and noticing it keeps references. I guess it's not a big deal since in non-pooled rulesets that's already the case in countless other places.

smoogipoo (approved):

PR #10952, opened 2020-11-24
Add ability to pause/resume replay playback

### LINKED ISSUE #5490: Pause replay playback
**Feature Request**
Allow pausing replay playback. Since we can already scrub through the replay timeline, it would be nice to be able to pause the replay at the current time. 
One use case for this is being able to watch the entire replay at a certain speed. Currently, you have to adjust the playback speed then drag the playback head back to the beginning.
I suggest using the spacebar as the pause key to keep it consistent with the editor.

peppy:
Doesn't work when watching autoplay, which arguably is something we may want in the future (not so easy to implement, though).

Closes https://github.com/ppy/osu/issues/5490.


osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -196,5 +197,8 @@ public enum GlobalAction
 
         [Description("Random Skin")]
         RandomSkin,
+
+        [Description("Pause Replay")]
+        PauseReplay,
```
Joehuu, line 202: Maybe better wording as it can resume playback. Also descriptions were once sentence cased, the recent ones somehow stayed title case.

```suggestion
        [Description("Toggle replay playback")]
        ToggleReplayPlayback,
```

osu.Game/Screens/Play/ReplayPlayer.cs
```diff
@@ -35,5 +37,24 @@ protected override ScoreInfo CreateScore()
 
             return Score.ScoreInfo;
         }
+
+        public bool OnPressed(GlobalAction action)
+        {
+            switch (action)
+            {
+                case GlobalAction.PauseReplay:
```
Joehuu, line 45:
```suggestion
                case GlobalAction.ToggleReplayPlayback:
```

osu.Game/Input/Bindings/GlobalActionContainer.cs
bdach: I think I'd propose `Pause / Resume replay` or similar rather than "toggle", the same way we have `Play / Pause` on `GlobalAction.MusicPlay`. I feel the latter is a little overly verbose but */shrug*.

bdach (commented):
No problems with this in general as far as I can see.

smoogipoo (dismissed):

PR #10954, opened 2020-11-24
Make all objects in selection candidates for spatial snapping

### LINKED ISSUE #10898: Editor object snapping only snaps position to first object in selection
**Describe the bug:** 
Selecting a group of objects in the editor only snaps position to the first object relative to time in the selection. 

https://streamable.com/icn2j3

**osu!lazer version:** 
development build @ ecd6d096002c8e64f3a33e792f5b5b5554c48c8c

**Logs:**

[database.log](https://github.com/ppy/osu/files/5561773/database.log)
[network.log](https://github.com/ppy/osu/files/5561777/network.log)
[performance.log](https://github.com/ppy/osu/files/5561775/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5561776/runtime.log)

peppy:
Closes #10898.


bdach:
This is a bit wonky when a stack is part of the selection, as you can sort of nudge between the two stacked objects and sometimes snap correctly getting stacking, and sometimes snap wrongly such that one of the objects get actually positioned at `Position + StackOffset` instead of being counted as part of the stack ([see video](https://drive.google.com/file/d/1qUpLnhQmtf__NYW1jR323Al1AftfFtNJ/view?usp=sharing)). Unsure if a huge problem.

I also got it to break once (snapped to opposite object from selected pair) but I have no idea how to reproduce and didn't capture a video....

peppy:
For case 1: How do you think it should behave for the case of snapping? Always snap based on the non-stacked position? I think making this happen is quite non-trivial, due to the fact that all the snapping code right now is done using screen space coordinates. Sounds like we probably want to be snapping based on the underlying hitobject's pre-stacked location, potentially?

For case 2: this is due to using ElementAt on a differently ordered lookup array (`SelectedObjects` vs the time storted list). Will address this.

bdach:
Yeah, re: case 1, that's pretty much the solution I also had in mind. But I'm not sure it's necessarily something to be done in this PR, might be best left for a separate one

peppy:
Definitely will handle that one separately

smoogipoo (approved):
Works fine, I'm not going to put much work into thinking about this

PR #10955, opened 2020-11-24
Improve scroll tracking of selected item

### LINKED ISSUE #10934: Beatmap carousel always scrolls from top when filtering
**Describe the bug:** 
Because all beatmap Y positions move closer to the top when applying a more narrow filter, all scrolling seems to scroll from the top even thought the beatmap might have already been on screen below the selected beatmap. Same vice-versa - when expanding/removing the filter, all scrolling seems to happen from bottom.

https://imgur.com/a/y2vBAl6 (gif would have been over 10MB)

**osu!lazer version:** 
Unpushed code at #8384 (it's difficult to see in actual releases because the selection after filtering is random)
master is at 09f905355d2052a6f15c4c52178fec0e4f206775

edit: Example video from 2020.1121.0: https://imgur.com/a/gVqtzNz

peppy:
As pointed out in #10934, during searching/filtering, the scroll position can get out of sync with the current item, causing massive visual noise while it catches up.

This PR changes the way we track the active item, by keeping a reference to the active panel rather than its position, allowing correctly tracking the vertical transform as panels move around each other.

Note that this doesn't track 1:1 (likely due to an off-by-one update?) but this is actually for the best, and leaves a bit of play in the animation. I tried making it track perfectly and it doesn't look or feel as good.

Also does a good job cleaning up the tracking code.

master:

this PR:

Closes #10934.

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -114,7 +114,6 @@ namespace osu.Game.Screens.Select
 
             ScrollableContent.Clear(false);
             itemsCache.Invalidate();
-            scrollPositionCache.Invalidate();
 
             // apply any pending filter operation that may have been delayed (see applyActiveCriteria's scheduling behaviour when BeatmapSetsLoaded is false).
             FlushPendingFilterOperations();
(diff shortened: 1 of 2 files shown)


peppy:
Actually this PR doesn't really do what the description says (since it's still using the final location), and on visual inspection the improvement isn't as much as I'd have hoped for. I think the code is a bit nicer though.

Will see if I can improve further.

LittleEndu:
Absolute scrolling / scrolling with right click is broken here (only works after you have scrolled once with scrollwheel). 

Also this doesn't really address #10934, in the issue the selection changes because of the filter.

peppy:
See the previous comment

peppy:
Closing because it doesn't do what it set out to do. See #10973 instead.

PR #10956, opened 2020-11-24
Move some members from `CatchHitObject` to `PalpableCatchHitObject` to make it more specific

ekrctb:
### Motivation of this change

This is not a necessary change, but
- More type safety.
- When pooling is used, we have to manage several `Bindable`s. This change makes easier to do that, I believe.

### Discussion

Maybe rename `PalpableCatchHitObject` -> `PalpableHitObject` and `DrawablePalpableCatchHitObject` -> `DrawablePalpableHitObject`, or anything else? The class name is somewhat long.


osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapProcessor.cs
```diff
@@ -192,24 +192,24 @@ private static void applyOffset(ref float position, float amount)
...
+                    palpableObjects.Add(fruitObject);
 
                 if (currentObject is JuiceStream)
                 {
-                    foreach (var currentJuiceElement in currentObject.NestedHitObjects)
+                    foreach (var juice in currentObject.NestedHitObjects)
                     {
-                        if (!(currentJuiceElement is TinyDroplet))
-                            objectWithDroplets.Add((CatchHitObject)currentJuiceElement);
+                        if (juice is PalpableCatchHitObject palpableObject && !(juice is TinyDroplet))
```
bdach, line 206: Not a huge fan of the new pattern match to `PalpableCatchHitObject` added here - juice streams should implicitly only have palpable nested objects and anything else is an invariant violation. I'd probably just straight-cast or make `JuiceStream`/`BananaShower` re-expose `NestedHitObjects` with the more-specific type:

```csharp
public new IReadOnlyList<PalpableCatchHitObject> NestedHitObjects => base.NestedHitObjects.Cast<PalpableCatchHitObject>();
```

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
```diff
@@ -8,8 +8,10 @@
 
 namespace osu.Game.Rulesets.Catch.Objects.Drawables
 {
-    public class DrawableFruit : PalpableDrawableCatchHitObject
+    public class DrawableFruit : DrawablePalpableCatchHitObject
     {
+        public override bool StaysOnPlate => true;
```
bdach, line 13: This override is redundant (restates the `true` from `base`). If we want this to be explicit in each inheritor then the base should be abstract and not virtual.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
```diff
@@ -0,0 +1,42 @@
...
+using osu.Framework.Graphics;
+using osu.Framework.Graphics.Containers;
+using osuTK;
+using osuTK.Graphics;
+
+namespace osu.Game.Rulesets.Catch.Objects.Drawables
+{
+    public abstract class DrawablePalpableCatchHitObject : DrawableCatchHitObject
+    {
+        public virtual bool StaysOnPlate => true;
```
bdach, line 15: This is overriden, but never actually used anywhere. Should just stick with the property on the HO, probably.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
```diff
@@ -2,55 +2,16 @@
...
-    public abstract class PalpableDrawableCatchHitObject : DrawableCatchHitObject
-    {
-        protected Container ScaleContainer { get; private set; }
-
-        protected PalpableDrawableCatchHitObject(CatchHitObject hitObject)
-            : base(hitObject)
-        {
-            Origin = Anchor.Centre;
-            Size = new Vector2(CatchHitObject.OBJECT_RADIUS * 2);
-            Masking = false;
```
bdach, line 25: Why wasn't this masking specification moved?

osu.Game.Rulesets.Catch/Objects/Drawables/DropletPiece.cs
```diff
@@ -23,7 +23,7 @@ public DropletPiece()
         private void load(DrawableHitObject drawableObject)
         {
             DrawableCatchHitObject drawableCatchObject = (DrawableCatchHitObject)drawableObject;
-            var hitObject = drawableCatchObject.HitObject;
+            var hitObject = (PalpableCatchHitObject)drawableCatchObject.HitObject;
```
bdach, line 26: I'd do this slightly differently, namely re-expose the HO as the more-specific type in the DHO:

```csharp
public abstract class DrawablePalpableCatchHitObject : DrawableCatchHitObject
{
    public new PalpableCatchHitObject HitObject => (PalpableCatchHitObject)base.HitObject;
    // ...
}
```

which then can be used more directly here to cast once instead of twice:

```csharp
var drawablePalpable = (DrawablePalpableHitObject)drawableObject;
var hitObject = drawablePalpable.HitObject;
```

osu.Game.Rulesets.Catch/Objects/Drawables/FruitPiece.cs
```diff
@@ -31,7 +31,7 @@ public FruitPiece()
         private void load(DrawableHitObject drawableObject)
         {
             DrawableCatchHitObject drawableCatchObject = (DrawableCatchHitObject)drawableObject;
-            hitObject = drawableCatchObject.HitObject;
+            hitObject = (PalpableCatchHitObject)drawableCatchObject.HitObject;
```
bdach, line 34: Same thing here as in `DropletPiece`.

osu.Game.Rulesets.Catch/Objects/PalpableCatchHitObject.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Rulesets.Catch.Objects
+{
+    /// <summary>
+    /// Represents a single object that can be caught by the catcher.
+    /// This includes normal fruits, droplets, and bananas but excludes objects that acts only as a container of nested hit objects.
```
bdach, line 8: nit
```suggestion
    /// This includes normal fruits, droplets, and bananas but excludes objects that act only as a container of nested hit objects.
```

osu.Game.Rulesets.Catch/Objects/PalpableCatchHitObject.cs
```diff
@@ -0,0 +1,31 @@
...
+        /// Whether this fruit can initiate a hyperdash.
+        /// </summary>
+        public bool HyperDash => HyperDashTarget != null;
+
+        /// <summary>
+        /// The target fruit if we are to initiate a hyperdash.
+        /// </summary>
+        public CatchHitObject HyperDashTarget;
+
+        public virtual bool StaysOnPlate => true;
```
bdach, line 29: Needs xmldoc as a public member.

osu.Game.Rulesets.Catch/Skinning/LegacyFruitPiece.cs
```diff
@@ -51,7 +52,7 @@ private void load(DrawableHitObject drawableObject, ISkinSource skin)
                 },
             };
 
-            if (drawableCatchObject.HitObject.HyperDash)
+            if (((PalpableCatchHitObject)drawableCatchObject.HitObject).HyperDash)
```
bdach, line 55: osu! has this pattern where resolved DHOs are cast to the more-specific type apt to the drawable piece at play and stored to fields with that type which I quite like and would make this a bit less of an eyesore:

https://github.com/ppy/osu/blob/7f3c8ad744df80079bc377b63c570ca6243426f9/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/DefaultSpinnerDisc.cs#L51-L54

bdach (changes requested):
Can generally agree with the spirit of this change.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
ekrctb: Why this change got into this commit is just a cherry picking oversight.
Why `Masking = false` is not needed is because it actually is no-op because `Masking` always `false` here.

osu.Game.Rulesets.Catch/Objects/Drawables/DropletPiece.cs
ekrctb: (To all comments about `DrawablePalpableHitObject` casts in `load`)
I agree with suggestions, but why I left ugly casts there is because those codes will be rewritten when the hit objects are pooled.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
ekrctb: I thought it should belong to `Drawable*` because it is a "visual only" property. It won't affect autoplay generation or difficulty calculation for example.

osu.Game.Rulesets.Catch/Beatmaps/CatchBeatmapProcessor.cs
ekrctb: I'm not sure about this because casting doesn't work for `IReadOnlyList` and I have to make it `IEnumerable<PalpableCatchHitObject> NestedHitObjects => ...`.

peppy (approved):
Looks pretty logical

bdach (approved):
looks good :+1:

PR #10961, opened 2020-11-25
Rewind pooled DHOs into better states

### LINKED ISSUE #10893: Rewinding can cause already-hit hitobjects to appear idle
Seems to be a regression of pooling. Reproducible in both editor and gameplay by rewinding ~2 seconds back in time.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10893

Should be tested with https://github.com/ppy/osu-framework/pull/4035

The MainCirclePiece/LegacyMainCirclePiece changes are as discussed in Discord. Not sure on adding the new armed state just yet, and this is standard across all other skinning usages/rulesets.

### DIFF
--- a/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
+++ b/osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
@@ -38,7 +38,6 @@ namespace osu.Game.Rulesets.Osu.Skinning
 
         private SkinnableSpriteText hitCircleText;
 
-        private readonly IBindable<ArmedState> state = new Bindable<ArmedState>();
         private readonly Bindable<Color4> accentColour = new Bindable<Color4>();
         private readonly IBindable<int> indexInCurrentCombo = new Bindable<int>();
 
@@ -113,7 +112,6 @@ namespace osu.Game.Rulesets.Osu.Skinning
             if (overlayAboveNumber)
                 AddInternal(hitCircleOverlay.CreateProxy());
 
-            state.BindTo(drawableObject.State);
             accentColour.BindTo(drawableObject.AccentColour);
             indexInCurrentCombo.BindTo(drawableOsuObject.IndexInCurrentComboBindable);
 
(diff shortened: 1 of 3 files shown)


osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
```diff
@@ -259,7 +259,17 @@ public void Apply([NotNull] HitObject hitObject, [CanBeNull] HitObjectLifetimeEn
...
-                Schedule(() => updateState(ArmedState.Idle, true));
+            {
+                Scheduler.Add(() =>
+                {
+                    if (Result.IsHit)
+                        updateState(ArmedState.Hit, true);
+                    else if (Result.HasResult)
+                        updateState(ArmedState.Miss, true);
+                    else
+                        updateState(ArmedState.Idle, true);
```
bdach, line 270: wondering whether a null-check here or a cancel of this schedule wouldn't be a good idea - in a theoretical scenario in which the DHO is freed before the schedule executes (e.g. if the DHO never becomes alive) this might fire at some wrong time and die on a nullref, but i'm not sure how theoretical that is

smoogipoo: Being a part of the hierarchy in a freed state? That sounds like something that should be asserted instead...

osu.Game.Rulesets.Osu/Skinning/LegacyMainCirclePiece.cs
```diff
@@ -137,19 +135,21 @@ protected override void LoadComplete()
         {
             base.LoadComplete();
 
-            state.BindValueChanged(updateState, true);
             accentColour.BindValueChanged(colour => hitCircleSprite.Colour = LegacyColourCompatibility.DisallowZeroAlpha(colour.NewValue), true);
             if (hasNumber)
                 indexInCurrentCombo.BindValueChanged(index => hitCircleText.Text = (index.NewValue + 1).ToString(), true);
+
+            drawableObject.ApplyCustomUpdateState += updateState;
```
peppy, line 142: Not sure about this direction, feels like we are going to be bitten in the ass by using the bindable in similar cases in the future...

smoogipoo: Are you able to forego this for the time being since it's a common implementation, and consider an `ArmedState.None` at a later point? There's a lot of things to consider such as how will every ruleset (mods/editor custom states/etc) respond when they receive an `ArmedState.None` event upon `FreeAfterUse()`.

There's a lot of safety checks I want to do when that happens, but I also kinda want to stop working on DHOs and stop breaking changes for a while soon.

peppy: yeah, but we should probably add a note somewhere about the behaviour (maybe on State's xmldoc? not sure)

smoogipoo: I've added an xmldoc remark on State for now.

peppy (approved):
let's let this go for now

Issue #10909, opened 2020-11-19
Sliderheads behave in multiple ways in editor


Pennek:
Sometimes they follow the sliderbody and other times they fade in place after being clicked.

video: https://youtu.be/64hl5qPO9HM
osu!lazer version: 2020.1114.0

bdach:
This might fall under scope of #10895, I think. Not 100% sure.

peppy:
Very weird. The autoplay differences are expected, but the case where the head circle is "hit" but doesn't follow the snaking is definitely odd.

bdach:
I looked into this for a fair bit and the cause here is pretty much down to non-frame-stable/instant seeks. Normally, in frame-stable contexts, you can generally expect that when rewinding gameplay, the gameplay has actually played out once in the past in the forward direction. In particular, the objects will have had their `HitObjectLifetimeEntries` created and populated with the result which happened in the forward run.

In editor this is not the case. Therefore you can skip far forward instantly via timeline, and all objects during that skipped time have no entry, therefore they have no result, therefore appear as idle/don't animate on rewind.

As to fixing that, however, that is the tricky part. You *could* initialise all of the entries eagerly, so that they have the max result applied in their entry. But, because `CreateResult()` is declared on the drawable object, rather than the normal hitobject or the entry, you *can't actually do that reliably preemptively until the object comes alive*.

Not sure where to take this forward. I may come back to this later but I figure I may as well leave this comment here.

PR #10962, opened 2020-11-25
Fix judgement offsets in non-framestable contexts

### LINKED ISSUE #10895: Artifacts in hit objects when scrolling backwards in the editor
**Describe the bug:** 
When scrolling backwards in the osu! standard editor all sorts of weird stuff happens with the hit objects. All the examples I've observed are:
- Hit circles fading in too early.
- Slider repeats facing the wrong direction.
- Slider repeats appearing in the wrong location before the slider appears.
- Sliders without any sliderhead. (only saw this one in local debug version)
- Ghost version of every slider. (if you skipped forwards in the timeline)

It's recreatable by going into the editor with any map, scrolling or skipping forwards a bit, and then slowly scrolling backwards in time.
Might be related to #10893 .

image
image
image
image
image
image

**osu!lazer version:** 
2020.1114.0

**Logs:**

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10895
Recommend testing with https://github.com/ppy/osu/pull/10961 for best results.

* `Drawable.MaximumJudgementOffset` added, to control the offset.
    * The offset is applied to both `ApplyResult` (as before).
    * Doesn't _really_ need to be overridden, but doing so can fix certain cases like hold notes/drumroll ticks not rewinding correctly because they don't use standard hit windows.
* `TimeOffset` is now always clamped, which fixes judgements occurring far into the future in non-framestable contexts.

As for the osu! hitobject changes:
* The `LifetimeStart` set was a no-op since it's already done by `HitObjectLifetimeEntry` and is never reset by the DHO.
* The `LifetimeEnd` movement into `OsuHitObjectLifetimeEntry` improves performance in the editor by skipping over intermediate hitobjects, at the cost of putting some DHOs into an `Idle` state upon rewind. I believe this can be improved in the editor itself by applying judgements directly upon seeking.


bdach:
The diff reads logically and correct (I think, but I'm barely conscious at this point, so buyer beware), but from a visual inspection I don't think https://github.com/ppy/osu/issues/10909 is fully resolved with this PR yet. Scrolling backwards still sometimes yields an unfaded sliderhead where the slider ball should be (it fades out properly after another rewind or two, but it's still there while it shouldn't be the first time after rewinding).

smoogipoo:
You're right about that, I've removed it from the description. The cause there is the hit state not being set because this is not in a framestable context. As per the description, the way around this is probably to apply judgements upon seek in the editor.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
```diff
@@ -176,6 +176,8 @@ private class OsuHitObjectLifetimeEntry : HitObjectLifetimeEntry
             public OsuHitObjectLifetimeEntry(HitObject hitObject)
                 : base(hitObject)
             {
+                // Arbitrary lifetime end to prevent past objects in idle states remaining alive in non-frame-stable contexts.
```
peppy, line 179: Does this comment still make sense? I don't believe the `+1000` is stopping objects from remaining alive, as much as giving them a chance to be judged or missed?

peppy (dismissed):
Seems good, just need clarification on comment.

osu.Game.Rulesets.Osu/UI/OsuPlayfield.cs
smoogipoo: It looks like the `+1000` isn't necessary. DHOs set `LifetimeEnd = double.MaxValue` on state update (i.e. when they come alive), and a more appropriate lifetime is set from there. Will remove.

peppy (approved):

PR #10966, opened 2020-11-26
Workaround TestSceneCatchModRelax flaky failure

ekrctb:
This test fails depending on the frame time.
It is easy to reproduce local visual testing on master by setting speed to 200%.
Example: <https://ci.appveyor.com/project/peppy/osu/builds/36518240/tests> I'm pretty sure the changes of that PR doesn't affect this test.


smoogipoo:
Why is it behaving so erratically in the first place? Even at 50% speed it's doing weird things like jittering on the spot, or moving to the hdash for one frame before any hitobjects are visible.

Tested back to before https://github.com/ppy/osu/pull/10605 and it was still a problem :/

smoogipoo (approved):
Given that the underlying issues still occur after this PR, I'll get this in for now as it should reduce the frequency of test failures.

PR #10973, opened 2020-11-26
Improve scroll tracking of selected item

### LINKED ISSUE #10934: Beatmap carousel always scrolls from top when filtering

peppy:
Fixes many of the animation edge cases / glitches of song select along the way.

I've described each change in detail in the commit messages, so suggest you read those during review to better understand reasoning (although there is also inline commenting which hopefully covers most of what is going on).

Closes #10934.
Supersedes and closes #10955. I went in a different direction to that version when implementing this.

master:

this PR:

Previous worst case scenario (last beatmap in list, meaning largest Y value for panel):

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
@@ -917,7 +917,7 @@ namespace osu.Game.Tests.Visual.SongSelect
             {
                 get
                 {
-                    foreach (var item in ScrollableContent)
+                    foreach (var item in Scroll.Children)
                     {
                         yield return item;
 
--- a/osu.Game/Graphics/Containers/OsuScrollContainer.cs
+++ b/osu.Game/Graphics/Containers/OsuScrollContainer.cs
@@ -12,7 +12,19 @@ using osuTK.Input;
 
 namespace osu.Game.Graphics.Containers
 {
-    public class OsuScrollContainer : ScrollContainer<Drawable>
+    public class OsuScrollContainer : OsuScrollContainer<Drawable>
+    {
+        public OsuScrollContainer()
+        {
+        }
+
+        public OsuScrollContainer(Direction direction)
+            : base(direction)
+        {
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


LittleEndu:
It would be awesome if scrolling on random could be restored. Scrolling seems to happen if you <kbd>Shift</kbd>+<kbd>F2</kbd> but only once, there's definitely something inconsistent going on. 

Filtering + randomizing seems to act the most inconsistent: https://youtu.be/q2xxIyd8kN8

peppy:
@LittleEndu should be better now, please give it a test.

LittleEndu:
I can still get some weird behaviour when I force the selection to change. Sometimes the selected beatmap needs to scroll onto screen, sometimes there are panels flying across other panels that were already on screen.

https://youtu.be/8jPv7vycOBU is with #8384 applied such so you know what's going to be selected.

https://youtu.be/Gm2UQZaOd3U and https://youtu.be/Pbp2c7TbW4s are just this PR. Still can get some weird behaviour if the filter results in just a few beatmaps to remain. 

Also found #10987 while testing this, but that isn't caused by this PR. Although this means you can't really tell how ruleset changing looks like. Only other thing I can think to check is ``presentBeatmap`` and that seems to work just fine with a satisfying seek just like random does now.

peppy:
that behaviour is expected. panels which were not newly retrieved from pool will be animated.

further work could be done to add animation back to pooled panels but that's not happening in this PR.

smoogipoo (approved):
Seems to work as expected and the changes look sane :+1:

PR #10977, opened 2020-11-27
Remove hitObject argument from DHO.OnApply and OnFree

ekrctb:
### Motivation of this change

- A DHO can use `HitObject` property instead.
- The argument is unused in all existing instances.
- `HitObject` property has more specific `TObj` type parameter type instead of `HitObject` type.
- Often `osu.Game.Rulesets.Objects` namespace has to be imported just because of this `HitObject` type.

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableOsuHitObject.cs
@@ -53,9 +53,9 @@ namespace osu.Game.Rulesets.Osu.Objects.Drawables
             });
         }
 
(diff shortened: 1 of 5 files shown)


smoogipoo (approved):
Generally I don't like doing this because it's a breaking change, but seeing as pooling is very new and other rulesets haven't finished updating, I think it's fine.

smoogipoo:
@LumpBloom7 just poking you because the pooling PR in your ruleset has already been merged (though no release yet)

LumpBloom7:
Thanks for the notice, I haven't yet merged my pr because I'm still waiting on a few improvements osu-side myself, so it's all fine for me.

PR #10979, opened 2020-11-27
Add bindables for osu!catch hit object states

ekrctb:
### Motivation of this change

A step to support hit object pooling in osu!catch.

### Known issues

It is currently allocating new `Drawable`s (`FruitPiece` and `DropletPiece`) when state changes, and thus it is not maximally efficient if pooling is used. This will be addressed in future PRs.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
@@ -30,13 +30,13 @@ namespace osu.Game.Rulesets.Catch.Tests
         }
 
         private Drawable createDrawableFruit(FruitVisualRepresentation rep, bool hyperdash = false) =>
-            setProperties(new TestDrawableFruit(new Fruit(), rep), hyperdash);
+            SetProperties(new TestDrawableFruit(new Fruit(), rep), hyperdash);
 
-        private Drawable createDrawableDroplet(bool hyperdash = false) => setProperties(new DrawableDroplet(new Droplet()), hyperdash);
+        private Drawable createDrawableDroplet(bool hyperdash = false) => SetProperties(new DrawableDroplet(new Droplet()), hyperdash);
 
-        private Drawable createDrawableTinyDroplet() => setProperties(new DrawableTinyDroplet(new TinyDroplet()));
+        private Drawable createDrawableTinyDroplet() => SetProperties(new DrawableTinyDroplet(new TinyDroplet()));
 
-        private DrawableCatchHitObject setProperties(DrawableCatchHitObject d, bool hyperdash = false)
+        protected virtual DrawableCatchHitObject SetProperties(DrawableCatchHitObject d, bool hyperdash = false)
         {
             var hitObject = d.HitObject;
             hitObject.StartTime = 1000000000000;
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
@@ -3,6 +3,8 @@
 
 using NUnit.Framework;
 using osu.Framework.Graphics;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.ControlPoints;
 using osu.Game.Rulesets.Catch.Objects;
 using osu.Game.Rulesets.Catch.Objects.Drawables;
 using osuTK;
@@ -55,9 +57,9 @@ namespace osu.Game.Rulesets.Catch.Tests
         protected virtual DrawableCatchHitObject SetProperties(DrawableCatchHitObject d)
         {
             var hitObject = d.HitObject;
+            hitObject.ApplyDefaults(new ControlPointInfo(), new BeatmapDifficulty { CircleSize = 0 });
             hitObject.StartTime = 1000000000000;
             hitObject.Scale = 1.5f;
-            hitObject.Samples.Clear(); // otherwise crash due to samples not loaded
 
             d.Anchor = Anchor.Centre;
             d.RelativePositionAxes = Axes.None;
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
@@ -30,7 +30,7 @@ namespace osu.Game.Rulesets.Catch.Tests
         }
 
         private Drawable createDrawableFruit(FruitVisualRepresentation rep, bool hyperdash = false) =>
-            setProperties(new DrawableFruit(new TestCatchFruit(rep)), hyperdash);
+            setProperties(new TestDrawableFruit(new Fruit(), rep), hyperdash);
 
         private Drawable createDrawableDroplet(bool hyperdash = false) => setProperties(new DrawableDroplet(new Droplet()), hyperdash);
 
@@ -56,14 +56,17 @@ namespace osu.Game.Rulesets.Catch.Tests
             return d;
         }
 
-        public class TestCatchFruit : Fruit
+        public class TestDrawableFruit : DrawableFruit
         {
-            public TestCatchFruit(FruitVisualRepresentation rep)
-            {
-                VisualRepresentation = rep;
-            }
+            private readonly FruitVisualRepresentation visualRepresentation;
 
-            public override FruitVisualRepresentation VisualRepresentation { get; }
+            protected override FruitVisualRepresentation GetVisualRepresentation(int indexInBeatmap) => visualRepresentation;
+
+            public TestDrawableFruit(Fruit fruit, FruitVisualRepresentation rep)
+                : base(fruit)
+            {
+                visualRepresentation = rep;
+            }
         }
     }
 }
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using NUnit.Framework;
 using osu.Framework.Graphics;
 using osu.Game.Rulesets.Catch.Objects;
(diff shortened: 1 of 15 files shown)


ekrctb:
Known issue addition:
Random rotation is set only once at DHO load. If a DHO is reused multiple times, rotations are the same.
I think we want to use some kind of seeded pseudorandom function instead of global `RNG`, to make fruit not change rotation when a replay is rewound when pooling is used. I'm not sure how to create the seed (maybe just using `IndexInBeatmap` is fine?).

peppy:
In other cases we use `StartTime` for seeding, probably best match suit if possible
https://github.com/ppy/osu/blob/60b6b56c045d77284a968a984d7ecab7f8e9169d/osu.Game.Rulesets.Osu/Objects/Drawables/Pieces/CirclePiece.cs#L61

ekrctb:
Rather than using `Random` I think there should be a stateless deterministic function
`static uint Hash(uint seed, uint series = 0)`.
(will be a separate PR).

ekrctb:
To make it more clear: the known issues above only affect after hit object pooling is implemented in osu!catch.
This PR is should be able to merged and no issues occur in master.

osu.Game.Rulesets.Catch.Tests/TestSceneFruitObjects.cs
```diff
@@ -17,33 +16,48 @@ protected override void LoadComplete()
...
+
+        protected virtual DrawableCatchHitObject SetProperties(DrawableCatchHitObject d)
         {
             var hitObject = d.HitObject;
             hitObject.StartTime = 1000000000000;
             hitObject.Scale = 1.5f;
-
-            if (hyperdash)
-                ((PalpableCatchHitObject)hitObject).HyperDashTarget = new Banana();
+            hitObject.Samples.Clear(); // otherwise crash due to samples not loaded
```
bdach, line 60: Can't this be something like

```csharp
hitObject.ApplyDefaults(new ControlPointInfo(), new BeatmapDifficulty { CircleSize = 0 });
```

like the error message indicates?

osu.Game.Rulesets.Catch.Tests/TestSceneFruitVisualChange.cs
```diff
@@ -0,0 +1,32 @@
...
+                HyperDashBindable = { BindTarget = hyperDash },
+            }))));
+
+            AddStep("droplet changes hyper", () => SetContents(() => SetProperties(new DrawableDroplet(new Droplet
+            {
+                HyperDashBindable = { BindTarget = hyperDash },
+            }))));
+
+            Scheduler.AddDelayed(() => indexInBeatmap.Value++, 250, true);
+            Scheduler.AddDelayed(() => hyperDash.Value = !hyperDash.Value, 1000, true);
```
bdach, line 29: what's the point of these? why aren't they normal test steps?

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
```diff
@@ -16,27 +16,47 @@ public abstract class CatchHitObject : HitObject, IHasXPosition, IHasComboInform
...
+        private float originalX;
 
         /// <summary>
         /// The horizontal position of the fruit between 0 and <see cref="CatchPlayfield.WIDTH"/>.
         /// </summary>
         public float X
         {
-            get => x + XOffset;
-            set => x = value;
+            // TODO: I don't like this asymmetry.
```
bdach, line 30: ...and rightly so, this can't live in the current state. if you ask me, there should be an `XBindable`, an `XOffsetBindable`, and an `AggregateBindable` that sums the two into an `XWithOffsetBindable`. names are temporary, would actually advise changing to `PositionBindable` etc.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
```diff
@@ -19,10 +22,26 @@ public abstract class DrawableCatchHitObject : DrawableHitObject<CatchHitObject>
         protected DrawableCatchHitObject(CatchHitObject hitObject)
```
bdach, line 22: should be annotated while we're at it

```suggestion
        protected DrawableCatchHitObject([CanBeNull] CatchHitObject hitObject)
```

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
```diff
@@ -12,6 +13,17 @@ public abstract class DrawablePalpableCatchHitObject : DrawableCatchHitObject
     {
         public new PalpableCatchHitObject HitObject => (PalpableCatchHitObject)base.HitObject;
 
+        public Bindable<bool> HyperDash { get; } = new Bindable<bool>();
+
+        public Bindable<float> ScaleBindable { get; } = new Bindable<float>(1);
+
+        public readonly Bindable<int> IndexInBeatmap = new Bindable<int>();
```
bdach, line 20: `BindableBool` and `BindableFloat` exist (the latter is actually much preferred over `Bindable<float>` as the latter does not have precision checking built-in).

also, is there a reason why two of these are get-only auto-properties and the third is a readonly field?

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
```diff
@@ -31,12 +43,42 @@ protected DrawablePalpableCatchHitObject(CatchHitObject h)
...
+            IndexInBeatmap.Value = h.IndexInBeatmap;
         }
 
         [BackgroundDependencyLoader]
         private void load()
         {
-            ScaleContainer.Scale = new Vector2(HitObject.Scale);
+            XBindable.BindValueChanged(x =>
+            {
+                if (!IsOnPlate) X = x.NewValue;
```
bdach, line 55: * why is this `X` set in here instead of the base `DrawableCatchHitObject`?
* what's with the `IsOnPlate` check? that should surely be an assert - objects on plate should never be having their `X` changed, surely? would probably even go as far as to say that `XBindable` should be disabled for plated objects

bdach (dismissed):

osu.Game.Rulesets.Catch/Objects/Drawables/Pieces/FruitPiece.cs
```diff
@@ -16,38 +15,31 @@ internal class FruitPiece : CompositeDrawable
         /// </summary>
         public const float RADIUS_ADJUST = 1.1f;
 
-        private BorderPiece border;
-        private PalpableCatchHitObject hitObject;
+        public readonly Bindable<FruitVisualRepresentation> VisualRepresentation = new Bindable<FruitVisualRepresentation>();
+        public readonly Bindable<bool> HyperDash = new Bindable<bool>();
+
+        public BorderPiece Border;
```
smoogipoo, line 21:
```suggestion
        public BorderPiece Border { get; private set; }
```

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
smoogipoo: I think the goal is to eventually have the plated objects be their own thing and not DHOs? Which would mean that this check is temporary while that's still the case.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
```diff
@@ -31,12 +43,42 @@ protected DrawablePalpableCatchHitObject(CatchHitObject h)
                 Origin = Anchor.Centre,
                 Anchor = Anchor.Centre,
             });
+
+            IndexInBeatmap.Value = h.IndexInBeatmap;
```
smoogipoo, line 47: This is a bit weird, why is this done here instead of relying on the binding in `OnApply`?

osu.Game.Rulesets.Catch.Tests/TestSceneFruitVisualChange.cs
ekrctb: I mimicked <https://github.com/ppy/osu/blob/master/osu.Game.Rulesets.Osu.Tests/TestSceneHitCircleComboChange.cs>.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
```diff
@@ -11,18 +12,48 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
...
+                });
+        }
 
-            ScaleContainer.Rotation = (float)(RNG.NextDouble() - 0.5f) * 40;
+        protected override void Update()
+        {
+            base.Update();
+
+            if (fruitPiece != null)
+                fruitPiece.Border.Alpha = (float)Math.Clamp((StartTimeBindable.Value - Time.Current) / 500, 0, 1);
```
smoogipoo, line 56: I don't like skinnable members' properties being set from the parent DHO. Would prefer moving this back into `FruitPiece` and instead of `hitObject = drawableObject.HitObject` as it originally did, it could directly do `border.Alpha = ... drawable.HitObject.StartTime - Time.Current ...`

Unless there's a bigger reason for doing it this way?

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
```diff
@@ -19,10 +22,26 @@ public abstract class DrawableCatchHitObject : DrawableHitObject<CatchHitObject>
...
             : base(hitObject)
         {
-            X = hitObject.X;
+            if (hitObject != null)
+                XBindable.Value = hitObject.X;
```
smoogipoo, line 26: Same here, seems weird to not reply on `OnApply`. Can this just be removed?

smoogipoo (changes requested):

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
ekrctb: I don't think those types are preferred here. I think the features (number step snapping, custom ToString etc.) of `BindableX` types are desired in user setting feature (`Bindable` type was originally in `Configuration` namespace, and later used for other purposes, and thus it has number of configuration use-case specific features remained).
> auto-properties
it is just an oversight. I'll make it readonly fields.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
ekrctb: `X` is only set for `Palpable` objects because non-palpable hit objects (as currently implemented) require to have `X` position of the `DHO` to be 0, independent of `X` of its hit object, due to relative positioning.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
ekrctb: It can be removed. It was for testing (when DHO is not placed in a playfield and not receiving `OnApply`) but it is not needed for testing anymore.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
ekrctb: Yes, one reason is to make pieces usable from non-DHO fruits on the plate.
Another reason having a `HitObject` on `FruitPiece` is not good (I think) is it has to react when the hit object is changed (because it will be recycled through pooling), but because it is not a DHO, `OnAppy` and `OnFree` doesn't exist and we have to manually make a method similar to it.
Then, another way to implement the logic is having a `Bindable` for `StartTime`. But I think the current way works better for the fruits on the plate.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
smoogipoo: Yeah it should be removed then imo, just for consistency. The passing of properties is going to be a part of the eventual wiki on pooling so there should be a clear direction for how to do this stuff (e.g. only via `OnApply()`).

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
smoogipoo: On what level will pooling happen for catch? Is it going to be:
1. A piece pool which _both_ DHOs and the plated objects use, or:
2. A piece pool for just the plated objects, and the DHOs being assigned pieces statically at construction time?
If the second, pieces can DI the DHO, which would be `null` for plated ones.

osu.Game.Rulesets.Catch.Tests/TestSceneFruitVisualChange.cs
ekrctb: I think it is a good way to see everything is correct quickly, in a manual visual testing.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
ekrctb: Objects on the plate doesn't change X after initialized, but it sets `X` once in the initialization there, and it overwrites the X value previously set by the catcher code.
And yes this code should be removed when caught objects are implemented differently.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
ekrctb: I'm thinking of the option 2.
Oh, so the suggestion is that storing DHO, not HO in a `FruitPiece`, right? (so the hit object change is not needed as I described, as it will be accessed through the DHO each update).
I think it works.

smoogipoo: Yep, that's exactly it. Already done that way for other skinnable pieces osu!-side.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableCatchHitObject.cs
smoogipoo: Let's overlook this one for now :) I'm sure pooling will clean this up a lot.

smoogipoo (approved):
lgtm!

osu.Game.Rulesets.Catch/Objects/CatchHitObject.cs
smoogipoo: I think we all agree this is bad, but in the interest of keeping the amount of possible breaking changes in this PR minimal, let's keep it as it is for now. Changes in this area have to be scrutinised hard with the possibility of breaking diffcalc.

PR #10983, opened 2020-11-27
Add `StatelessRNG` to be used for randomly-looking visual elements

ekrctb:
### Motivation for this change

Existing call to global `RNG` like <https://github.com/ppy/osu/blob/60b6b56c045d77284a968a984d7ecab7f8e9169d/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs#L25> is

- Not consistent across different gameplay, which might not be desired for testing purposes, or to the player (the player may be distracted from having different randomness for each replay).
- When pooling is used, the same object will have different when a replay is rewound or in the editor.

There is also code like <https://github.com/ppy/osu/blob/c0c197501e3f794e6582c0744270a268c1a21163/osu.Game.Rulesets.Catch/Objects/Banana.cs#L37> to get stability.

Another way is to use seeded `Random` but:

- Only one integer argument is used. It leads to a situation of different visual element "correlates" each other if the same seed is used. Solving it requires some ad-hoc code.
- It is a small but non-zero allocation.
- The stateful nature requires some consideration of when to create a new `Random` and when the random values can be generated.

This PR introduces `StatelessRNG` as an abstract function of "scrambling" input seed number. The desired property of this function is that it "looks random" for humans, as it is for visual elements, and no cryptographic security is needed. Also, the good performance of the function is nice-to-have.

`mix` function is adopted from [MurmurHash3](https://github.com/aappleby/smhasher/blob/61a0530f28277f2e850bfc39600ce61d02b518de/src/MurmurHash3.cpp#L81-L89) (public domain declared).

### DIFF
--- a/osu.Game/Utils/StatelessRNG.cs
+++ b/osu.Game/Utils/StatelessRNG.cs
@@ -1,0 +1,44 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Utils
+{
+    /// Provides a fast stateless function that can be used in randomly-looking visual elements.
+    public static class StatelessRNG
+    {
+        private static ulong mix(ulong x)
+        {
+            unchecked
+            {
+                x ^= x >> 33;
+                x *= 0xff51afd7ed558ccd;
+                x ^= x >> 33;
+                x *= 0xc4ceb9fe1a85ec53;
+                x ^= x >> 33;
+                return x;
+            }
+        }
+
+        /// Compute an integer from given seed and series number.
+        /// <param name="seed">
+        /// The seed value of this random number generator.
+        /// </param>
+        /// <param name="series">
+        /// The series number.
+        /// Different values are computed for the same seed in different series.
+        /// </param>
+        public static ulong Get(int seed, int series = 0) =>
+            unchecked(mix(((ulong)(uint)series << 32) | ((uint)seed ^ 0x12345678)));
+
+        /// Compute a floating point value between 0 and 1 (excluding 1) from given seed and series number.
+        /// <param name="seed">
+        /// The seed value of this random number generator.
+        /// </param>
+        /// <param name="series">
+        /// The series number.
+        /// Different values are computed for the same seed in different series.
+        /// </param>
+        public static float GetSingle(int seed, int series = 0) =>
+            (float)(Get(seed, series) & ((1 << 24) - 1)) / (1 << 24); // float has 24-bit precision
+    }
+}
--- a/osu.Game/Utils/StatelessRNG.cs
+++ b/osu.Game/Utils/StatelessRNG.cs
@@ -1,6 +1,8 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System;
+
 namespace osu.Game.Utils
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


osu.Game/Utils/StatelessRNG.cs
```diff
@@ -0,0 +1,44 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+namespace osu.Game.Utils
+{
+    /// Provides a fast stateless function that can be used in randomly-looking visual elements.
```
bdach, line 6: These general descriptions should be in `<summary></summary>` tags (goes for both the class and the public methods).

Also, would rename public members to match `RNG`:

* `Get()` -> `Next()`
* `GetSingle()` -> `NextSingle()`

osu.Game/Utils/StatelessRNG.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// Compute an integer from given seed and series number.
+        /// <param name="seed">
+        /// The seed value of this random number generator.
+        /// </param>
+        /// <param name="series">
+        /// The series number.
+        /// Different values are computed for the same seed in different series.
+        /// </param>
+        public static ulong Get(int seed, int series = 0) =>
```
bdach, line 30: I'd be strongly for returning `long`. Unsigned types are super awkward to use and always require a cast when usage is needed in signed cases (which is by far the more common case)

osu.Game/Utils/StatelessRNG.cs
```diff
@@ -0,0 +1,44 @@
...
+            unchecked(mix(((ulong)(uint)series << 32) | ((uint)seed ^ 0x12345678)));
```
bdach, line 31: Is this double `(ulong)(uint)` cast necessary?

osu.Game/Utils/StatelessRNG.cs
ekrctb: If we return a negative value we cannot use it like `var face =  StatelessRNG.Get(seed) % 6`.

Remark: While taking a remainder of a RNG to interval is often discouraged due to

1. often RNG has more quality in higher-bits than lower-bits.
2. It is slightly non-uniform (getting 0 has ~ `1/(2^64/6)` higher chance than 5).

But the method is fine in this case because 1. it is opposite: the particular mix function mixes slightly mixes lower-order bits more. 2. the uniformity is maximal due to 64-bit seed, and we cannot do better.

osu.Game/Utils/StatelessRNG.cs
ekrctb: It is necessary, because `(ulong)-1` is `0xffffffffffffffff`, while `(ulong)(uint)-1` is `0x00000000ffffffff`.

osu.Game/Utils/StatelessRNG.cs
ekrctb: Maybe I should rename `ulong` version to `GetRaw` or something and make a new `GetInt(range, seed, series)`. Requiring a lengthy explanation to use is a symptom of a bad design.

osu.Game/Utils/StatelessRNG.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// </param>
+        /// <param name="series">
+        /// The series number.
+        /// Different values are computed for the same seed in different series.
+        /// </param>
+        public static ulong NextUlong(int seed, int series = 0)
+        {
+            unchecked
+            {
+                //
```
bdach, line 40: empty comment?

bdach (dismissed):
seems fine, at least code-wise

smoogipoo (approved):
Looks okay

PR #10986, opened 2020-11-27
Fix state application always checking newest state for early abort, rather than current

peppy:
Applying a change, undoing that change, then applying the same change would result in the second state change not being saved. Oversight in lookup index.

### DIFF
--- a/osu.Game.Tests/Editing/EditorChangeHandlerTest.cs
+++ b/osu.Game.Tests/Editing/EditorChangeHandlerTest.cs
@@ -62,7 +62,7 @@ namespace osu.Game.Tests.Editing
 
             string hash = handler.CurrentStateHash;
 
-            // save a save without making any changes
+            // undo a change without saving
             handler.RestoreState(-1);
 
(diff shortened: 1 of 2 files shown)


osu.Game.Tests/Editing/EditorChangeHandlerTest.cs
```diff
@@ -44,6 +43,36 @@ public void TestSaveRestoreState()
...
+            addArbitraryChange(beatmap);
+            handler.SaveState();
+
+            Assert.That(handler.CanUndo.Value, Is.True);
+            Assert.That(handler.CanRedo.Value, Is.False);
+            Assert.That(stateChangedFired, Is.EqualTo(1));
+
+            string hash = handler.CurrentStateHash;
+
+            // save a save without making any changes
```
bdach, line 65: can't seem to decipher this statement - did you mean something like:

```suggestion
            // undo a change without saving
```

smoogipoo (approved):

Issue #9594, opened 2020-07-17
 Section highligh should cover the whole visible range


Flutterish:
**Describe the new feature:** Section highligh should cover the whole visible range

**Proposal designs of the feature:** The top of the highligh should be located at the top of the top-most visible section and the bottom of the highligh should be located at the bottom of the bottom-most visible section. I am unsure how sections are implemented so it might require moving the highligh from individual sections to the section container. Additionaly all visible sections should light up as well.

![osu](https://user-images.githubusercontent.com/40297338/87785671-dc6bd680-c838-11ea-93d8-a30f2400169f.png)

recognitionmark:
wh- what do you mean?

Alchyr:
In the example, the orange highlight on the left showing which section is currently viewed is expanded to cover all currently visible sections, rather than just the "primary" one.

peppy:
Nah, should be the current section. Other sections should be dimmed like stable, though.

Flutterish:
Which section is the current section, though? Given the current solution, ( in the example ) having the second and third last sections be active is not even possible. The active section jumps from last to fourth last when I scroll. ( Possibly related #9196  )

peppy:
Empty space should likely be added to the bottom to make it possible, or the algorithm should be improved to consider display height.l

PR #10990, opened 2020-11-27
Set settings section as selected if user presses the sidebar button

### LINKED ISSUE #9196: Settings menu button click not highlighted correctly
osu_2020-06-03_10-56-32

` 

LittleEndu:
Closes #9196

settings

~~I assume the section we scroll to has to reach the search box to be considered selected but currently it will never reach that (at least not high enough to always satisfy ``<= 0``)~~

~~Considering the margin when deciding what section is selected sounds most reasonable, but maybe this should be even more than ``51`` because users would consider the section reached well before.~~

### DIFF
--- a/osu.Game/Overlays/UserProfileOverlay.cs
+++ b/osu.Game/Overlays/UserProfileOverlay.cs
@@ -202,7 +202,7 @@ namespace osu.Game.Overlays
                 RelativeSizeAxes = Axes.Both;
             }
 
-            protected override OsuScrollContainer CreateScrollContainer() => new OverlayScrollContainer();
+            protected override UserTrackingScrollContainer CreateScrollContainer() => new OverlayScrollContainer();
 
             protected override FlowContainer<ProfileSection> CreateScrollContentContainer() => new FillFlowContainer<ProfileSection>
             {
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -889,27 +889,9 @@ namespace osu.Game.Screens.Select
             }
         }
 
-        private class CarouselScrollContainer : OsuScrollContainer
+        private class CarouselScrollContainer : UserTrackingScrollContainer
         {
             private bool rightMouseScrollBlocked;
-
-            /// <summary>
-            /// Whether the last scroll event was user triggered, directly on the scroll container.
-            /// </summary>
-            public bool UserScrolling { get; private set; }
-
-            // ReSharper disable once OptionalParameterHierarchyMismatch 2020.3 EAP4 bug. (https://youtrack.jetbrains.com/issue/RSRP-481535?p=RIDER-51910)
-            protected override void OnUserScroll(float value, bool animated = true, double? distanceDecay = default)
-            {
-                UserScrolling = true;
-                base.OnUserScroll(value, animated, distanceDecay);
-            }
-
-            public new void ScrollTo(float value, bool animated = true, double? distanceDecay = null)
-            {
-                UserScrolling = false;
-                base.ScrollTo(value, animated, distanceDecay);
-            }
 
             protected override bool OnMouseDown(MouseDownEvent e)
             {
--- a/osu.Game/Graphics/Containers/SectionsContainer.cs
+++ b/osu.Game/Graphics/Containers/SectionsContainer.cs
@@ -95,7 +95,7 @@ namespace osu.Game.Graphics.Containers
 
         protected override Container<T> Content => scrollContentContainer;
 
-        private readonly OsuScrollContainer scrollContainer;
+        private readonly UserTrackingScrollContainer scrollContainer;
         private readonly Container headerBackgroundContainer;
         private readonly MarginPadding originalSectionsMargin;
         private Drawable expandableHeader, fixedHeader, footer, headerBackground;
@@ -139,7 +139,7 @@ namespace osu.Game.Graphics.Containers
         public void ScrollToTop() => scrollContainer.ScrollTo(0);
 
         [NotNull]
-        protected virtual OsuScrollContainer CreateScrollContainer() => new OsuScrollContainer();
+        protected virtual UserTrackingScrollContainer CreateScrollContainer() => new UserTrackingScrollContainer();
 
         [NotNull]
         protected virtual FlowContainer<T> CreateScrollContentContainer() =>
@@ -201,9 +201,9 @@ namespace osu.Game.Graphics.Containers
                 {
                     SelectedSection.Value = Children.LastOrDefault();
                 }
-                else
+                else if (scrollContainer.UserScrolling)
                 {
-                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= section.Margin.Top).LastOrDefault()
+                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 1).LastOrDefault()
                                             ?? Children.FirstOrDefault();
                 }
             }
--- a/osu.Game/Graphics/Containers/SectionsContainer.cs
+++ b/osu.Game/Graphics/Containers/SectionsContainer.cs
@@ -203,7 +203,7 @@ namespace osu.Game.Graphics.Containers
                 }
                 else
                 {
-                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 0).LastOrDefault()
+                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= section.Margin.Top).LastOrDefault()
                                             ?? Children.FirstOrDefault();
                 }
             }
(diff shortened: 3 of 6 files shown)


bdach:
I don't think this is a correct solution. This is actually an issue with the comparison-to-zero which suffers from floating-point precision issues. I'm claiming that based on the following log excerpt I acquired when I logged what the `diff` function returns for each section when navigating between them:

```
[runtime] 2020-11-29 18:08:01 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:01 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:01 [debug]: MouseClick handled by SidebarButton.
[runtime] 2020-11-29 18:08:01 [verbose]: diffs: 120.26445, 476.26434, 1325.2622, 2577.262, 3420.262, 3782.2617, 4199.2617, 4386.2617, 4960.2617
[runtime] 2020-11-29 18:08:01 [verbose]: diffs: 58.10704, 414.10693, 1263.1049, 2515.1045, 3358.1045, 3720.1042, 4137.1045, 4324.1045, 4898.104
[runtime] 2020-11-29 18:08:01 [verbose]: diffs: 53.97528, 409.97516, 1258.973, 2510.973, 3353.973, 3715.9727, 4132.9727, 4319.9727, 4893.9727
(...)
[runtime] 2020-11-29 18:08:03 [verbose]: diffs: -355.9988, 0.0010986328, 848.999, 2100.9988, 2943.9988, 3305.9988, 3722.9988, 3909.9988, 4483.998
[runtime] 2020-11-29 18:08:03 [verbose]: diffs: -355.9988, 0.0010681152, 848.999, 2100.9988, 2943.9988, 3305.9988, 3722.9988, 3909.9988, 4483.9985
[runtime] 2020-11-29 18:08:03 [verbose]: diffs: -355.99988, 3.0517578E-05, 848.9979, 2100.9976, 2943.9978, 3305.9976, 3722.9978, 3909.9978, 4483.9976
[runtime] 2020-11-29 18:08:12 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:13 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:13 [debug]: MouseClick handled by SidebarButton.
[runtime] 2020-11-29 18:08:13 [verbose]: diffs: -355.99988, 6.1035156E-05, 848.9994, 2100.999, 2943.9985, 3305.9985, 3722.9985, 3909.999, 4483.999
[runtime] 2020-11-29 18:08:13 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:14 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:14 [debug]: MouseClick handled by SidebarButton.
[runtime] 2020-11-29 18:08:14 [verbose]: diffs: -355.99997, -6.1035156E-05, 848.9992, 2100.999, 2943.9988, 3305.9983, 3722.998, 3909.9985, 4483.9985
[runtime] 2020-11-29 18:08:14 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:14 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:14 [debug]: MouseClick handled by SidebarButton.
[runtime] 2020-11-29 18:08:14 [verbose]: diffs: -355.99988, 6.1035156E-05, 848.9994, 2100.999, 2943.9985, 3305.9985, 3722.9985, 3909.999, 4483.999
[runtime] 2020-11-29 18:08:17 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:17 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:17 [debug]: MouseClick handled by SidebarButton.
[runtime] 2020-11-29 18:08:17 [verbose]: diffs: -355.99997, -6.1035156E-05, 848.9992, 2100.999, 2943.9988, 3305.9983, 3722.998, 3909.9985, 4483.9985
[runtime] 2020-11-29 18:08:17 [debug]: MouseDownEvent(Left) handled by Sidebar+SidebarScrollContainer.
[runtime] 2020-11-29 18:08:17 [debug]: ClickEvent(Left) handled by SidebarButton.
[runtime] 2020-11-29 18:08:17 [debug]: MouseClick handled by SidebarButton.
```

Note that at the first click, the diff value never satisfies `<= 0`, and then begins to oscillate on subsequent clicks. Each button actually behaves slightly differently (some of them will reach 0, but only after an inordinately long wait).

LittleEndu:
I get that it's a floating-point precision issue, but I also think the section should be considered reached well before the diff satisfies ``<= 0``. ``Margin.Top`` (or ``51`` which it equals to) starts to sound more and more like an arbitrary number in my head, maybe something more like 1/4th the overlay height would be a better number to use.

bdach:
all of these sound more arbitrary to me than 0 if I'm honest.

LittleEndu:
I'm suggesting something way bigger than 0 because I think users would consider the "Skin" section reached on this screenshot and would find it weird that the button isn't highlighted.
osu_2020-11-29_22-14-42

I don't think something like some % of container height is arbitrary for this.

bdach:
then why isn't "input" reached? if we're going that far, then surely all sections which have a part of them visible on the screen should be highlighted, no?

(side note, just noticed that the positioning of the "seasonal backgrounds" options makes zero sense. it's in audio settings.)

LittleEndu:
> then surely all sections which have a part of them visible on the screen should be highlighted

#9594 ?

bdach:
well if that was rejected, then i'd say it should either be reaching the search box as is currently, or an entirely different measure of determining selection should be used (i.e. the section which has most of it visible on screen at the current moment is selected). that's way less arbitrary than saying "the section is selected when you cross the top 25% of the screen", because that arbitrary percentage has no visible reference point.

osu.Game/Overlays/UserProfileOverlay.cs
```diff
@@ -202,7 +202,7 @@ public ProfileSectionsContainer()
                 RelativeSizeAxes = Axes.Both;
             }
 
-            protected override OsuScrollContainer CreateScrollContainer() => new OverlayScrollContainer();
+            protected override UserTrackingScrollContainer CreateScrollContainer() => new OverlayScrollContainer();
```
bdach, line 205: this change seems unnecessary

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -889,28 +889,10 @@ protected override void PerformSelection()
...
+        private class CarouselScrollContainer : UserTrackingScrollContainer
         {
             private bool rightMouseScrollBlocked;
 
-            /// <summary>
-            /// Whether the last scroll event was user triggered, directly on the scroll container.
-            /// </summary>
-            public bool UserScrolling { get; private set; }
-
-            // ReSharper disable once OptionalParameterHierarchyMismatch 2020.3 EAP4 bug. (https://youtrack.jetbrains.com/issue/RSRP-481535?p=RIDER-51910)
```
bdach, line 901: why was this removed? have you checked if this is resolved in rider EAP?

LittleEndu: yes it was. and it didn't trigger after removing so I guess the tracker is correct

osu.Game/Overlays/UserProfileOverlay.cs
LittleEndu: Can change that and cast in ``SectionsContainer`` ctor instead if that's preferred
```cs
scrollContainer = (UserTrackingScrollContainer)CreateScrollContainer().With(s =>
                {
                    s.RelativeSizeAxes = Axes.Both;
                    s.Masking = true;
                    s.ScrollbarVisible = false;
                    s.Child = scrollContentContainer = CreateScrollContentContainer();
                })
```

bdach: no, that's much worse - never mind, thought that the `UserProfileOverlay` change was inheriting from a different type

LittleEndu: hmmm, but wouldn't something like this make more sense:
```cs
                else if (!(scrollContainer is UserTrackingScrollContainer user) || user.UserScrolling)
                {
                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 1).LastOrDefault()
                                            ?? Children.FirstOrDefault();
                }
```

bdach: I think that's unnecessary complication.

peppy:
If you're going to go in this direction and remove the animation during scroll (i think it feels okay without it so i'm neutral), why not fix it in a permanent way - match on the last clicked section rather than Y coordinates.

Your change doesn't fix all cases, see this scenario:

LittleEndu:
I think I'll try to implement what bdach suggested, as in have the section which takes up the most of visible area consider as selected, and see how that looks. Something like that should look better, fix the issue, keep the animation, and would not introduce a new class.

PR #10999, opened 2020-11-29
Create "User Interface" settings section

bdach:
Noticed this thing during discussion in #10990. In the existing setup we had:

- an *User Interface* section in *Graphics*,
- a *Main Menu* section in *Audio* (most of settings in which weren't even audio-related),
- a *Song Select* section in *Gameplay* (sorta makes sense but not really).

This PR proposes to unify them all under a new *User Interface* section which should hopefully make a bit more sense, with the exception of the hold-to-confirm setting which remained in *Gameplay* as it seems more apt there.

Icon is slightly random but had no better ideas.

2020-11-29-223818_274x103_scrot

2020-11-29-223924_515x889_scrot

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
@@ -5,7 +5,6 @@ using osu.Framework;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Game.Configuration;
-using osu.Game.Graphics.UserInterface;
 using osu.Game.Rulesets.Scoring;
 
 namespace osu.Game.Overlays.Settings.Sections.Gameplay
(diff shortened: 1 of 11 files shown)


osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
```diff
@@ -63,6 +64,12 @@ private void load(OsuConfigManager config)
                     LabelText = "Always show key overlay",
                     Current = config.GetBindable<bool>(OsuSetting.KeyOverlay)
                 },
+                new SettingsSlider<float, TimeSlider>
```
smoogipoo, line 67: Not sure I agree with this one being in here... With that being said I also don't agree with the gameplay one being tied to the game-wide/exit one, but that's something I've argued in the past.

smoogipoo (dismissed):
Seems a little more organised to me. @peppy would also like your opinion.

osu.Game/Overlays/Settings/Sections/Gameplay/GeneralSettings.cs
peppy: It's not gameplay specific so agree it shouldn't be here (this also affects the main menu exit timer, for instance).

peppy:
Made a few touch-ups and moved some elements around. I think this is a good step in the right direction.

peppy (approved):

smoogipoo (approved):

PR #11004, opened 2020-11-30
Add ability to customise key binding for in-game overlay toggle

### LINKED ISSUE #10887: change binding for in-game interface toggle.
There should be a binding setting for "toggle in-game interface" it would be gladly appreciated by me and people who's part of the keyboard dosen't work.
osu!_cu9fdPFwiq

zoogies:
closes #10887 
creates a keybind for toggling the in-game overlay as opposed to it only being shift-tab, keybind defaults to "i"

let me know if anything is wrong with my code, this is my first PR and I'm not very familiar yet 😄

### DIFF
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -68,6 +68,7 @@ namespace osu.Game.Input.Bindings
             new KeyBinding(new[] { InputKey.Control, InputKey.Tilde }, GlobalAction.QuickExit),
             new KeyBinding(new[] { InputKey.Control, InputKey.Plus }, GlobalAction.IncreaseScrollSpeed),
             new KeyBinding(new[] { InputKey.Control, InputKey.Minus }, GlobalAction.DecreaseScrollSpeed),
+            new KeyBinding(InputKey.I, GlobalAction.ToggleInGameInterface),
             new KeyBinding(InputKey.MouseMiddle, GlobalAction.PauseGameplay),
             new KeyBinding(InputKey.Space, GlobalAction.TogglePauseReplay),
             new KeyBinding(InputKey.Control, GlobalAction.HoldForHUD),
@@ -144,6 +145,9 @@ namespace osu.Game.Input.Bindings
 
         [Description("Decrease scroll speed")]
         DecreaseScrollSpeed,
+
+        [Description("Toggle in-game interface")]
+        ToggleInGameInterface,
 
         [Description("Select")]
         Select,
--- a/osu.Game/Input/Bindings/GlobalActionContainer.cs
+++ b/osu.Game/Input/Bindings/GlobalActionContainer.cs
@@ -68,7 +68,7 @@ namespace osu.Game.Input.Bindings
             new KeyBinding(new[] { InputKey.Control, InputKey.Tilde }, GlobalAction.QuickExit),
             new KeyBinding(new[] { InputKey.Control, InputKey.Plus }, GlobalAction.IncreaseScrollSpeed),
             new KeyBinding(new[] { InputKey.Control, InputKey.Minus }, GlobalAction.DecreaseScrollSpeed),
-            new KeyBinding(InputKey.I, GlobalAction.ToggleInGameInterface),
+            new KeyBinding(new[] { InputKey.Shift, InputKey.Tab }, GlobalAction.ToggleInGameInterface),
             new KeyBinding(InputKey.MouseMiddle, GlobalAction.PauseGameplay),
             new KeyBinding(InputKey.Space, GlobalAction.TogglePauseReplay),
             new KeyBinding(InputKey.Control, GlobalAction.HoldForHUD),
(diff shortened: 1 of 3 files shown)


osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -145,6 +146,9 @@ public enum GlobalAction
         [Description("Decrease scroll speed")]
         DecreaseScrollSpeed,
 
+        [Description("Toggle in-game interface")]
+        ToggleInGameInterface,
```
smoogipoo, line 150: This needs to be at the very bottom of the enum, otherwise it offsets all other values and breaks existing keybinds. It's a known limitation.

osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -68,6 +68,7 @@ public GlobalActionContainer(OsuGameBase game)
             new KeyBinding(new[] { InputKey.Control, InputKey.Tilde }, GlobalAction.QuickExit),
             new KeyBinding(new[] { InputKey.Control, InputKey.Plus }, GlobalAction.IncreaseScrollSpeed),
             new KeyBinding(new[] { InputKey.Control, InputKey.Minus }, GlobalAction.DecreaseScrollSpeed),
+            new KeyBinding(InputKey.I, GlobalAction.ToggleInGameInterface),
```
smoogipoo, line 71: Can't be `i`. Make it `Shift+Tab` by default.

smoogipoo (dismissed):

osu.Game/Input/Bindings/GlobalActionContainer.cs
Yoyolick: makes sense, sorry ill fix

Yoyolick:
Just to update, I've fixed these two things but discovered a notification that needs to be changed to list the user's keybind instead of only "shift-tab". Once I finish fixing that ill push those changes.

Joehuu:
> Just to update, I've fixed these two things but discovered a notification that needs to be changed to list the user's keybind instead of only "shift-tab". Once I finish fixing that ill push those changes.

Not sure why @smoogipoo wants to do a structural change by moving `LookupKeyBindings` to `KeyBindingStore` in this PR. `HUDOverlay` already DIs `OsuConfigManager`, so you can just add `{config.LookupKeyBindings(GlobalAction.ToggleInGameInterface)}` to the notification string.

I'd wait for him to reply though.

Yoyolick:
> > Just to update, I've fixed these two things but discovered a notification that needs to be changed to list the user's keybind instead of only "shift-tab". Once I finish fixing that ill push those changes.
> 
> 
> 
> Not sure why @smoogipoo wants to do a structural change by moving `LookupKeyBindings` to `KeyBindingStore` in this PR. `HUDOverlay` already DIs `OsuConfigManager`, so you can just add `{config.LookupKeyBindings(GlobalAction.ToggleInGameInterface)}` to the notification string.
> 
> 
> 
> I'd wait for him to reply though.

Yeah I actually realized that I could just do config.LookupKeyBindings and ended up going with that for the time being, I figured once I push it we could revisit the idea of the structural change and he could request changes to make it clearer

smoogipoo:
DI-ing the config to lookup the keybinding text sounds quite weird to me. If you wanted to query the bindings otherwise you'd use the `KeyBindingStore`.

Yoyolick:
This should be finished, the notification as well as the popup all read the new keybinding now.

settings example
popup example
notification

osu.Game/Screens/Play/HUDOverlay.cs
```diff
@@ -377,6 +346,24 @@ public bool OnPressed(GlobalAction action)
...
+
+                        case HUDVisibilityMode.HideDuringGameplay:
+                            configVisibilityMode.Value = HUDVisibilityMode.Always;
+                            break;
+
+                        case HUDVisibilityMode.Always:
+                            configVisibilityMode.Value = HUDVisibilityMode.Never;
+                            break;
+                    }
+                    updateVisibility();
```
peppy, line 365: this is automatically called via bindable flow and didn't need to be added here.

peppy:
I've fixed the remaining issues with this PR. In the future please make sure to name your branch and not create PRs from `master`. See https://github.com/ppy/osu/blob/master/CONTRIBUTING.md

peppy (approved):

bdach (approved):

osu.Game/Input/Bindings/GlobalActionContainer.cs
```diff
@@ -200,5 +201,8 @@ public enum GlobalAction
 
         [Description("Pause / resume replay")]
         TogglePauseReplay,
+
+        [Description("Toggle in-game interface")]
```
Joehuu, line 205: `In-game interface` is another word to the `HUD` terminology, the other being `score overlay` from the notification. I think we should just use `HUD` for all user facing strings, like `Hold for HUD`. Also `in-game` is redundant, since it's already in the in game keybinding section.

bdach: There's bound to be more places with inconsistencies like this, so I think this should be a separate pull.

Joehuu (approved):

smoogipoo (approved):

PR #11006, opened 2020-11-30
Fix osu!catch combo counter not showing after 1 combo

### LINKED ISSUE #10929: Dissapearing combo counter in catch
The combo counter in catch dissappears after 1 combo. It re-appears after a miss, but will dissappear after 1 combo once again. Seems to happen on all maps and even in replays, as seen in the video. It only seems to happen on skins with the combo above the catcher, since the combo stays on the osu!lazer skin (which has the combo counter next to the score counter) but dissappears on the osu!classic skin (which has the combo counter above the catcher). Can't fully test this since osu!lazer is the only skin I know of that doesn't have the combo counter above the catcher.

Video of the bug: https://icebearbestbear.s-ul.eu/OsN8KxyW (only contains dissappearing combo counter while playing and watching replays)

version 2020.1120.0

peppy:
Closes #10929. May be considered a temporary fix until we figure out a permanent solution to this.

### DIFF
--- a/osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
+++ b/osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
@@ -55,6 +55,11 @@ namespace osu.Game.Rulesets.Catch.UI
                 HitObjectContainer,
                 CatcherArea,
             };
+        }
+
+        protected override void LoadComplete()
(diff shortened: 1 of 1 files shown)


smoogipoo (approved):
Temporary solution for sure

PR #11007, opened 2020-11-30
Pool `Pulp` piece for catch fruit objects

ekrctb:
- [x] Depends on #10979

Draft because pooled pulps seems not be returned (the count in the statistics only increases).
@smoogipoo How to properly return a drawable to the pool?


smoogipoo:
They need to be removed from the hierarchy, either via:
1. `Clear(false)` (`false` is important so they don't dispose, see [this o!f issue](https://github.com/ppy/osu-framework/issues/4038))
2. `Remove(poolable)`
3. `Expire()`
After which they'll be returned automatically.

ekrctb:
I was confused a bit but the number of drawables increasing in the gameplay is normal because hit object pooling is not implemented yet (all drawables won't be removed in a gameplay).
However, the number of drawables also increases in `TestSceneFruitVisualChange` (from PR #10979), which is not expected, as only one drawable should be present at one time.
I don't know why pooled drawables are not returned in this case.

smoogipoo:
Interesting edge case. It looks like that's because:
1. The removal happens at the parent level - `DrawableFruit` and the poolable objects themselves are not directly removed from the hierarchy (they're still in the sub-hierarchy that had its parent removed from the hierarchy).
2. `ScaleContainer.Child` effectively does a disposal-clear...

bdach:
Merging `master` back here to eliminate prereq commits from #10979.

@ekrctb would be appreciated if you tagged prereqs using checkbox list items (`- [ ]`) in the PR description.

ekrctb:
Closing to investigate other possible approaches.

PR #11008, opened 2020-11-30
Remove semi-transparent backgrounds from settings and notifications overlays

peppy:
I tried also updating the colours to the "new" versions from designs but they don't match due to colour profile differences (so I'm not yet sure if they are correct or not) and also don't look great without all the UI elements also being updated.

### DIFF
--- a/osu.Game/Overlays/SettingsPanel.cs
+++ b/osu.Game/Overlays/SettingsPanel.cs
@@ -12,6 +12,7 @@ using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
 using osu.Framework.Graphics.Shapes;
 using osu.Framework.Input.Events;
+using osu.Game.Graphics;
 using osu.Game.Graphics.Containers;
 using osu.Game.Graphics.UserInterface;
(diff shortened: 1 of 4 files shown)


osu.Game/Overlays/SettingsPanel.cs
```diff
@@ -214,7 +215,7 @@ protected override void UpdateAfterChildren()
                 base.UpdateAfterChildren();
 
                 // no null check because the usage of this class is strict
-                HeaderBackground.Alpha = -ExpandableHeader.Y / ExpandableHeader.LayoutSize.Y * 0.5f;
+                HeaderBackground.Alpha = -ExpandableHeader.Y / ExpandableHeader.LayoutSize.Y * 1;
```
smoogipoo, line 218:
```suggestion
                HeaderBackground.Alpha = -ExpandableHeader.Y / ExpandableHeader.LayoutSize.Y;
```

smoogipoo (changes requested):

smoogipoo (approved):

PR #11009, opened 2020-11-30
Attempt to reduce settings visual clutter with better paddings

peppy:
Before (left) vs After (right)
image

- [x] Depends on #11008


smoogipoo (approved):

bdach (approved):
I do like these numbers better :+1:

PR #11011, opened 2020-11-30
Initial implementation of hit object pooling in osu!catch ruleset

ekrctb:
This is the initial implementation of hit object pooling in osu!catch. All hit objects, `Palpable` or not, nested or not, are pooled through the pool in the `CatchPlayfield`, and recycled for multiple hit objects.

The idea of hit object pooling is because only a small number of `DrawableHitObject`s (DHO for short) are alive (i.e. on-screen) compared to the total number of `HitObject` (HO for short) in a beatmap, recycling a DHO for multiple HO will reduce allocation and initialization time.

The loading time and memory usage is greatly reduced by this PR. I roughly tested on this beatmap <https://osu.ppy.sh/beatmapsets/1023806#fruits/2141652>, and I got:

- Before: 5s loading time, 1.3G memory.
- After: 1s loading time, 675M memory.

The difficulty of the hit object pooling is a DHO has to react to changes of HO. This reactivity is usually implemented using `Bindable`s. In-place changes are preferred to reduce allocation.

### What is not implemented?

- A `DrawableFruit` currently recreates either `FruitPiece` or `LegacyFruitPiece` when the DHO is recycled to a different hit object. To remove this allocation,
  - For `FruitPice`: I have a draft of pooling `Pulp` #11007. But the PR has still unsolved issues.
  - For `LegacyFruitPiece`: I think we should replace `Texture` in-place, but where to store the textures?
- Currently, when a hit object is caught, a new DHO is created from the hit object and placed on the plate.
  - A plan is to use non-DHO `Drawable`s for the objects on the plate (using `FruitPiece` etc.).
  - Then, the caught objects should be pooled separately.

`TestSceneCatchPlayerLegacySkin` is added because legacy skin takes a different code path and I had a bug in `LegacyFruitPiece`.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
@@ -2,39 +2,13 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using NUnit.Framework;
-using osu.Framework.Allocation;
-using osu.Framework.Audio;
-using osu.Framework.IO.Stores;
-using osu.Game.Skinning;
 using osu.Game.Tests.Visual;
 
 namespace osu.Game.Rulesets.Catch.Tests
 {
     [TestFixture]
-    public class TestSceneCatchPlayerLegacySkin : PlayerTestScene
+    public class TestSceneCatchPlayerLegacySkin : LegacySkinPlayerTestScene
     {
-        private ISkinSource legacySkinSource;
-
-        protected override TestPlayer CreatePlayer(Ruleset ruleset) => new SkinProvidingPlayer(legacySkinSource);
-
-        [BackgroundDependencyLoader]
-        private void load(AudioManager audio, OsuGameBase game)
-        {
-            var legacySkin = new DefaultLegacySkin(new NamespacedResourceStore<byte[]>(game.Resources, "Skins/Legacy"), audio);
-            legacySkinSource = new SkinProvidingContainer(legacySkin);
-        }
-
         protected override Ruleset CreatePlayerRuleset() => new CatchRuleset();
-
-        public class SkinProvidingPlayer : TestPlayer
-        {
-            [Cached(typeof(ISkinSource))]
-            private readonly ISkinSource skinSource;
-
-            public SkinProvidingPlayer(ISkinSource skinSource)
-            {
-                this.skinSource = skinSource;
-            }
-        }
     }
 }
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
@@ -28,18 +28,12 @@ namespace osu.Game.Rulesets.Catch.Tests
 
         public class SkinProvidingPlayer : TestPlayer
         {
+            [Cached(typeof(ISkinSource))]
             private readonly ISkinSource skinSource;
 
             public SkinProvidingPlayer(ISkinSource skinSource)
             {
                 this.skinSource = skinSource;
-            }
-
-            protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
-            {
-                var dependencies = new DependencyContainer(base.CreateChildDependencies(parent));
-                dependencies.CacheAs(skinSource);
-                return dependencies;
             }
         }
     }
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
@@ -1,0 +1,46 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Audio;
+using osu.Framework.IO.Stores;
+using osu.Game.Skinning;
+using osu.Game.Tests.Visual;
+
+namespace osu.Game.Rulesets.Catch.Tests
+{
+    [TestFixture]
+    public class TestSceneCatchPlayerLegacySkin : PlayerTestScene
+    {
+        private ISkinSource legacySkinSource;
+
+        protected override TestPlayer CreatePlayer(Ruleset ruleset) => new SkinProvidingPlayer(legacySkinSource);
+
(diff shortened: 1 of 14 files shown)


osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
```diff
@@ -0,0 +1,46 @@
...
+using osu.Framework.Allocation;
+using osu.Framework.Audio;
+using osu.Framework.IO.Stores;
+using osu.Game.Skinning;
+using osu.Game.Tests.Visual;
+
+namespace osu.Game.Rulesets.Catch.Tests
+{
+    [TestFixture]
+    public class TestSceneCatchPlayerLegacySkin : PlayerTestScene
```
bdach, line 14: Seeing that you've gone through the trouble of adding a legacy skin `PlayerTestScene`, mind making it more abstract (`LegacySkinPlayerTestScene`) so that it can be re-used for other rulesets?

osu.Game.Rulesets.Catch.Tests/TestSceneCatchPlayerLegacySkin.cs
```diff
@@ -0,0 +1,46 @@
...
+            {
+                this.skinSource = skinSource;
+            }
+
+            protected override IReadOnlyDependencyContainer CreateChildDependencies(IReadOnlyDependencyContainer parent)
+            {
+                var dependencies = new DependencyContainer(base.CreateChildDependencies(parent));
+                dependencies.CacheAs(skinSource);
+                return dependencies;
+            }
```
bdach, line 43: could just as well stick a `[Cached(typeof(ISkinSource))]` on the field for a bit more brevity

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
```diff
@@ -57,6 +57,17 @@ public CatchPlayfield(BeatmapDifficulty difficulty, Func<CatchHitObject, Drawabl
...
 
+        [BackgroundDependencyLoader]
+        private void load()
+        {
+            RegisterPool<Droplet, DrawableDroplet>(1);
+            RegisterPool<TinyDroplet, DrawableTinyDroplet>(1);
+            RegisterPool<Fruit, DrawableFruit>(1);
+            RegisterPool<Banana, DrawableBanana>(1);
+            RegisterPool<JuiceStream, DrawableJuiceStream>(1);
+            RegisterPool<BananaShower, DrawableBananaShower>(1);
```
bdach, line 68: Only one of everything at the start is a little... stringent, don't you think? At least I'd allow a couple of fruits up front, and definitely a lot more droplets and bananas? Not sure what the overhead is for creating-on-the-fly when pool limit is exceeded, but it shouldn't be a large memory footprint cost to increase these a bit and eliminate that overhead entirely.

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
```diff
@@ -127,5 +124,17 @@ protected override void UpdateAfterChildren()
...
+        private DrawableCatchHitObject createCaughtFruit(DrawablePalpableCatchHitObject hitObject)
+        {
+            return hitObject.HitObject switch
+            {
+                Banana banana => new DrawableBanana(banana),
+                Fruit fruit => new DrawableFruit(fruit),
+                TinyDroplet tiny => new DrawableTinyDroplet(tiny),
+                Droplet droplet => new DrawableDroplet(droplet),
+                _ => null
+            };
```
bdach, line 137: use a standard `switch` here

bdach (changes requested):

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
ekrctb: What is the reason the switch statement is preferred than the switch expression?

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
ekrctb: I could have guessed numbers without any clue but I doubt setting initial capacity will make any noticeable impact. Something like this tends to be over-allocated.

osu.Game.Rulesets.Catch/UI/CatcherArea.cs
smoogipoo: Some are opposed to the `x switch` syntax.

osu.Game.Rulesets.Catch/UI/CatchPlayfield.cs
smoogipoo: It's always going to be a guess, but in general I prefer expending a little bit of memory to try to prevent allocations during gameplay that could lead to stutters.

In my simple test, instantiating + adding a `DrawableFruit` takes 0.08ms, but this on a PC with top-end IPC and lots of multi-core headroom. Instantiating just 10 fruits which I'd say is a lower minimum, could be very expensive on lower end/mobile platforms.

Definitely want to see a little bit larger initial capacity here. Somewhere around 10-20 for the fruits and 5-10 for juicestreams, as an example.

peppy: i think the best way is to take a heavy beatmap (insane/expert/aspire) and judge requirements. pools should aim to never be exceeded (with obvious exceptions being things like centipede).

ekrctb: I tested more.
If I'm understanding correctly, the DHO instantiation time will not be noticeable as long as multiple DHOs are not instantiated in one frame.

But currently all nested DHOs are instantiated at once when the parent DHO became alive, and will likely be noticeable as a `BananaShower` can have hundreds of bananas (it currently noticeably stutters when the banana shower became alive).
I think this should be fixed to make nested drawables instantiated incrementally.
I measured some difficult maps and the maximum number of hit objects varies greatly between maps.

- `Fruit` A stream map had more than 100, but others are 30~50.
- `Droplet` 20~50.
- `TinyDroplet` typically ~20, .
- `Banana` typically ~70, one map with a long spinner had 257, but I'm pretty sure more long spinners exist in other ranked maps.

peppy: I'd say the upper numbers you state there (and maybe 100 for banana) sound reasonable.

ekrctb:
As I didn't notice any bad impact either by increasing the capacity, I'm okay with this capacity. Maybe it can be revisit when other allocation issues are solved.

Also I'm remarking here that the loading of pooled nested DHOs should be addressed separately.

bdach (approved):
looks like a good step forward to me :+1:

smoogipoo (approved):

PR #11017, opened 2020-11-30
Fix current beatmap temporarily becoming empty during ruleset change

### LINKED ISSUE #10987: Changing ruleset causes no beatmap to be selected for a brief moment
**Describe the bug:** 
When a mapset that has difficulties from multiple rulesets is selected, when you switch to a different ruleset, for a brief moment no map is selected. This causes a reset for the background and the audio track.

https://youtu.be/lX2yYTA7uYo

**osu!lazer version:** 
current master at c0c197501e3f794e6582c0744270a268c1a21163

peppy:
When changing the ruleset at song select, there was a scenario where it would be set to default (empty) for one debounce length where this was not actually required. This occurs when the currently selected beatmap set has other difficulties which match the target ruleset, specifically. 

Closes https://github.com/ppy/osu/issues/10987

### DIFF
--- a/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
+++ b/osu.Game.Tests/Visual/SongSelect/TestScenePlaySongSelect.cs
@@ -644,6 +644,54 @@ namespace osu.Game.Tests.Visual.SongSelect
         }
 
         [Test]
+        public void TestChangingRulesetOnMultiRulesetBeatmap()
+        {
+            int changeCount = 0;
+
+            AddStep("change convert setting", () => config.Set(OsuSetting.ShowConvertedBeatmaps, false));
+            AddStep("bind beatmap changed", () =>
+            {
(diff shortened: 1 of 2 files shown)


bdach:
Gonna be honest, I'm not really feeling comfortable approving or merging this as presented. I'm not sure I see where the nested operation would be coming from.

I guess there are tests which definitely gives more confidence, but ideally I'd still rather want to understand what the corner case is.

peppy:
It's the case already described in the inline comments (changing ruleset -> transferRuleset -> refilter -> eager select from carousel beatmap)

Should be pretty easy to follow along with - set a breakpoint on the scheduler add/debounce call and note that it is fired twice when running the test, or changing rulesets on a map with more than one ruleset. the second fire will run 200ms after the first, leaving a period of 200ms where the selection is in a non-determined/incorrect state.

smoogipoo (approved):

PR #11019, opened 2020-11-30
Consider most visible SectionsContainer section as selected one

### LINKED ISSUE #9196: Settings menu button click not highlighted correctly

LittleEndu:
Idea from here: https://github.com/ppy/osu/pull/10990#issuecomment-735449612 with a slight modification. If the most visible section has a section before it that's 100% visible, consider that one selected instead.

Opening PR as draft because I honestly couldn't explain how better or not this is from what we have right now. This would fix #9196 though which was the goal (recording lagged a bit, sorry): 
settings-new


osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -196,16 +198,46 @@ protected override void UpdateAfterChildren()
...
+                /*if (scrollContainer.IsScrolledToEnd())
                 {
                     SelectedSection.Value = Children.LastOrDefault();
                 }
                 else
                 {
                     SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 0).LastOrDefault()
                                             ?? Children.FirstOrDefault();
-                }
+                }*/
```
bdach, line 240: Opening as draft is no excuse to leave behind dead commented code.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -196,16 +198,46 @@ protected override void UpdateAfterChildren()
 
                 float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
                 Func<T, float> diff = section => scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
+                Func<T, float> sectionPos = section => scrollContainer.GetChildPosInContent(section);
+                Func<T, float> visibleSize = section => Math.Min(
+                    sectionPos(section) + section.Height, currentScroll + scrollContainer.DisplayableContent) - Math.Max(sectionPos(section), currentScroll
```
bdach, line 203: `sectionPos` is needlessly calculated twice

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -196,16 +198,46 @@ protected override void UpdateAfterChildren()
...
+                );
+                var sortedByVisible = new List<T>(Children);
+                sortedByVisible.Sort((a, b) => visibleSize(b).CompareTo(visibleSize(a)));
+                T mostVisible = sortedByVisible.FirstOrDefault();
+
+                if (mostVisible != null)
```
bdach, line 209: invert and early-return to reduce nesting

bdach (changes requested):

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: I thought this the easiest way to test master vs this PR. Also the reason why ``diff`` is still there and caused CI failure.

bdach: Dynamic recompilation works with `git checkout` as far as I know, so actually the easiest way to check would be to detached-head-checkout d957679 and then ac94b2d.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -195,16 +201,35 @@ protected override void UpdateAfterChildren()
...
+                sortedByVisible.Sort((a, b) => visibleSize(b).CompareTo(visibleSize(a)));
+                T mostVisible = sortedByVisible.FirstOrDefault();
+
+                if (mostVisible == null)
+                    return;
+
+                int mostVisibleIndex = IndexOf(mostVisible);
+                T ch;
+
+                while (mostVisibleIndex != 0 && Precision.AlmostEquals(visibleSize(ch = Children[mostVisibleIndex - 1]), ch.Height, 1))
```
LittleEndu, line 223: Probably want something like this so scrolling upwards using buttons feels as fast as scrolling downwards

```suggestion
                while (mostVisibleIndex != 0 && Precision.AlmostBigger(visibleSize(
                        ch = Children[mostVisibleIndex - 1]), ch.Height * (externallyScrolledTo as T == ch ? 0.8 : 1), 1
                ))
```

bdach: I can't understand what this code snippet is proposing. What is `0.8` and why is it there?

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -195,16 +201,35 @@ protected override void UpdateAfterChildren()
...
                 }
-                else
+
+                SelectedSection.Value = Children[mostVisibleIndex];
+
+                if (Precision.AlmostBigger(currentScroll, scrollContainer.ScrollableExtent))
                 {
-                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 0).LastOrDefault()
-                                            ?? Children.FirstOrDefault();
+                    SelectedSection.Value = (scrollContainer.UserScrolling ? null : externallyScrolledTo as T) ?? Children.LastOrDefault();
```
bdach, line 232: I also don't understand this. Can't this just be

```suggestion
                    SelectedSection.Value = externallyScrolledTo as T ?? Children.LastOrDefault();
```

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -20,6 +22,7 @@ public class SectionsContainer<T> : Container<T>
         where T : Drawable
     {
         public Bindable<T> SelectedSection { get; } = new Bindable<T>();
+        private Drawable externallyScrolledTo;
```
bdach, line 25: I think this needs a better name. Every occurrence of this I have to think about what this actually is. Maybe `lastClickedSection`?

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: I guessed we want the very last section to be selected when there is a user scroll (via mouse scroll wheel). There's actually a test case for this. This line is written like that so that happens. Another option would be to always reset ``externallyScrolledTo``/``lastClickedSection`` when user starts to scroll.

bdach: I think I would prefer resetting myself. I have a feeling that not doing so complicates all of this code a lot.

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: If the roundabout way ``ScrollTo(section)`` is triggered in user profile overlay is by clicking the tab title, then sure. Otherwise naming it ``lastClickedSection`` would be wrong, because it can be triggered by means other than clicking buttons.

bdach: I'm even more confused now, admittedly. What exactly causes an "external scroll"?

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: When scrolling downwards the section comes visible sooner. That's because our target position for the section is the top of the screen. When scrolling upwards, the section stays behind the header and isn't considered selected until it is 100% visible. Just thought that maybe we should make an exception for the section user clicked for and consider it selected when only 80% is visible. 80% was chosen arbitrarily, felt like good number.

Another fix for it would be to scroll the section to the middle of the screen. More specifically, instead of scrolling top of the section to top of the screen, scroll the middle of the section (or at minimum half the screen height if section takes up more than screen worth of space, just so section header stays on screen) to the middle of the screen.

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: > What exactly causes an "external scroll"?

Usages of ``ScrollTo()``. 
2020-12-05 16_34_27-Window

For settings, it is clicking the sidebar button, but for user profile it's some complicated change in value, which is probably triggered by clicking on the tab titles.
osu_2020-12-05_16-37-05

osu.Game/Graphics/Containers/SectionsContainer.cs
bdach: What about the exemption where it's 1 instead of 0.8? Also this assignment to `ch` within the loop is confusing me intensely. I'd like this loop to be simple, stupid code even, if possible.

osu.Game/Graphics/Containers/SectionsContainer.cs
bdach: Do you mean the carousel? User profile's "external scroll" logic is dead simple as far as I can see.

LittleEndu: https://github.com/ppy/osu/blob/a40a16955c319a052c2d604088682cbb01aa0bc3/osu.Game/Overlays/UserProfileOverlay.cs#L114

It's specifically this line that triggers the ``ScrollTo`` when the current tab changes. But I don't know much about how ``TabControl`` classes work and if current tab can only be changed by clicking on the tab title.

bdach: Yeah, that's just a click on a tab.

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: > What about the exemption where it's 1 instead of 0.8?

It's actually not an exemption, it's the default for every section except the one user wanted to see.

I would suggest just scrolling the section further than just he top of the screen, but then it becomes the question about how much further. Main issue would be for small sections that would become unselected if scroll the the middle, and like a mentioned sections that take up more space than screen height.

> I'd like this loop to be simple, stupid code even, if possible.

Isn't accessing an index O(1)? I agree that I should have just written ``Children[mostVisibleIndex - 1]`` more than once.

bdach: > It's actually not an exemption, it's the default for every section except the one user wanted to see.

I'd say that should be accurately commented, then. It's not immediately obvious what that condition does. Maybe the other changes around this logic will improve this, not sure.

> Isn't accessing an index O(1)? I agree that I should have just written `Children[mostVisibleIndex - 1]` more than once.

Computational complexity is not what I'm concerned about here - code complexity is. This one-liner while condition has a test in it, followed by another test, that has a nested assignment and a function call. It's not easy reading. That can all be in the body and be several lines.

LittleEndu:
I think this is ready for review. Can't find anything else to change besides the typo in comment here (should be large sections): https://github.com/ppy/osu/blob/7ff3ffaaf17285f508dc545804fce196ce86c9e3/osu.Game/Graphics/Containers/SectionsContainer.cs#L241-L243

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -195,16 +204,49 @@ protected override void UpdateAfterChildren()
...
+                var sortedByVisible = new List<T>(Children);
+                sortedByVisible.Sort((a, b) => visibleHeight(b).CompareTo(visibleHeight(a)));
+                T mostVisible = sortedByVisible.FirstOrDefault();
+
+                if (mostVisible == null)
+                    return;
+
+                int mostVisibleIndex = IndexOf(mostVisible);
+
+                bool usePreviousSection()
```
peppy, line 225: You can't just drop inline functions in the middle of a huge block like this. It makes for really bad readability. I'd split this out into its own private method if possible, including documentation on what "usePreviousSection" actually means.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -195,16 +204,49 @@ protected override void UpdateAfterChildren()
...
+
+                    bool moreVisible = Precision.AlmostBigger(visibleHeight(sectionBefore), visibleHeight(Children[mostVisibleIndex]), 1);
+                    bool completelyVisible = Precision.AlmostBigger(visibleHeight(sectionBefore), sectionHeight, 1);
+
+                    return moreVisible || completelyVisible;
                 }
-                else
+
+                // we want to use 100% visible previous sections instead because otherwise small sections sandwiched between large section would never get selected
+                while (usePreviousSection())
```
peppy, line 242: i think the function name here is misleading/confusing. "while use previous section is true" means nothing to me if i read it out loud.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -195,16 +204,49 @@ protected override void UpdateAfterChildren()
...
+                    mostVisibleIndex--;
```
peppy, line 243: what is this? i'm honestly not willing to read this whole method due to how badly it mentally parses for me.

move all this logic out of `UpdateAfterChildren` and provide xmldoc explanation of what you're trying to achieve here, so we can begin the review process.

peppy (changes requested):

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -183,72 +183,82 @@ protected override void UpdateAfterChildren()
...
-                    if (mostVisibleIndex == 0)
-                        return false;
+            SelectedSection.Value = Children[mostVisibleIndex];
+        }
 
-                    T sectionBefore = Children[mostVisibleIndex - 1];
-                    // consider the section user wants to see smaller than it actually is to lower the visibility requirement
-                    // this makes the section get selected sooner when scrolling upwards, to match it already getting selected sooner when scrolling downwards
-                    float sectionHeight = sectionBefore.Height * (lastClickedSection as T == sectionBefore ? 0.8f : 1.0f);
+        /// <returns>true if previous section is more visible than one at <paramref name="sectionIndex"/></returns>
```
LittleEndu, line 236:
``` suggestion
        /// <returns>true if section before one at <paramref name="sectionIndex"/> is more visible than one at <paramref name="sectionIndex"/></returns>
```

Something like this probably reads better

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -183,72 +183,82 @@ protected override void UpdateAfterChildren()
...
+            T mostVisible = sortedByVisibility.FirstOrDefault();
 
-                if (mostVisible == null)
-                    return;
+            if (mostVisible == null)
+                return;
 
-                int mostVisibleIndex = IndexOf(mostVisible);
+            int mostVisibleIndex = IndexOf(mostVisible);
+            while (isPreviousMoreVisible(mostVisibleIndex))
```
peppy, line 230: i think this may read better if you make it into a `for` loop with a `break` condition.

LittleEndu: I think the best thing to do is actually rewrite the sort. Sort on this "more visible or 100% visible" notion to begin with. 

Got this idea while driving to work, I'll implement it when I get home.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -177,36 +183,90 @@ protected override void UpdateAfterChildren()
...
+            //
+            // if this is ever not the case, this comparison should be rewritten
+            float aIndex = IndexOf(a);
+            float bIndex = IndexOf(b);
+            float aVisibleHeight = visibleHeight(a);
+            float bVisibleHeight = visibleHeight(b);
+            // visibility requirement is lower (only 80%) for the section user clicked for
+            // this makes the section get selected sooner when scrolling upwards, to match it already getting selected sooner when scrolling downwards
+            bool aAllVisible = Precision.AlmostBigger(aVisibleHeight / a.Height, lastClickedSection as T == a ? 0.8f : 1.0f);
+            bool bAllVisible = Precision.AlmostBigger(bVisibleHeight / b.Height, lastClickedSection as T == b ? 0.8f : 1.0f);
```
bdach, line 240: I still don't like this corner case and would ideally see it removed but /shrug.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -177,36 +183,90 @@ protected override void UpdateAfterChildren()
...
+                // this is so we don't select a small section as soon as it gets on screen (only unwanted while scrolling downwards)
+                if (aAllVisible && aIndex < bIndex)
+                    return -1;
+                if (bAllVisible && bIndex < aIndex)
+                    return 1;
+            }
+
+            // if both are all visible - prioritize higher sections
+            if (aAllVisible && bAllVisible)
+                return IndexOf(a).CompareTo(IndexOf(b));
```
bdach, line 255: those indices are already computed once

```suggestion
                return aIndex.CompareTo(bIndex);
```

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: Like it says in the first comment, this differentiation between scroll directions is necessary because our scroll target is top of the screen. And it makes sense for it to be the top of the screen because most sections we scroll to have a title.

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -177,36 +183,90 @@ protected override void UpdateAfterChildren()
...
             }
+
+            var sortedByVisibility = new List<T>(Children);
+            sortedByVisibility.Sort(compareVisibility);
+            SelectedSection.Value = sortedByVisibility.FirstOrDefault();
+        }
+
+        private int compareVisibility(T a, T b)
+        {
+            // note: the decision to prioritize higher sections and differentiate scroll direction
```
peppy, line 229: honestly this feels wrong, and doesn't match stable. why aren't we centering the target instead? will that simplify things?

peppy:
This is an especially bad case that remains:

https://user-images.githubusercontent.com/191335/102574930-57615200-4135-11eb-85ad-f629026331e9.mov

You click a button then drag one pixel and are now in a new section.

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: It won't really be more simple, it will just be different. There will be other edge cases you would need to consider if the sections are scrolled to middle

LittleEndu: Actually, if we were to scroll to middle, then an algorithm other than visibility is probably the best. You would probably want to select the section intersecting with the middle line, which might be easier to do changing the old code. The edge case would come from big sections that won't fit on screen and would still need to be scrolled to the top, and from the bottom of the screen where there might be several sections that can never reach the middle but should still be highlighted if clicked on.

peppy: Yeah, it would probably need to scroll the start of the section to have the start of the section sitting at around 25% of the vertical height of the settings container (rather than the precise centre of the section). I'm not sure if stable already does this or not.

LittleEndu:
Should be all fine now that scroll target is 20%

https://user-images.githubusercontent.com/11320483/102636369-3a407980-415d-11eb-81a1-185f2e3cecb2.mp4

peppy:
This still doesn't work. Make a tall window and click buttons and it doesn't go to expected section.

LittleEndu:
> Make a tall window and click buttons and it doesn't go to expected section.

Seems to work for me:

https://user-images.githubusercontent.com/11320483/102896921-db864300-446f-11eb-8896-5154ff21c4e6.mp4

peppy:
It's great that it works for you, but it doesn't for me, which is why I reported it. My window was roughly 1280x1440 to reproduce issues.

LittleEndu:
There must be something else wrong for you, there is nothing in code that would make a difference on different window sizes (at least not anything that I can think of)

peppy:
Just going to close this PR for now then. Let me know if you have interest in fixing it up. I think it will be faster for me to write a fix for this from scratch as a separate PR.

LittleEndu:
I wanna fix this but I can't reproduce the issue with the steps you provided...

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -177,36 +189,104 @@ protected override void UpdateAfterChildren()
...
+            // set top cutoff such that if the smallest section is above our target then the section at our target is more visible in every situation
+            // that is to say, top cutoff should be just above scroll target but not so much above that the smallest section could be 100% visible,
+            // hiding, or being more visible than, our actual target section when above it
+            float topCutoff = Math.Max(currentScroll, currentScroll + scrollContainer.DisplayableContent * scroll_target_multiplier - smallestSection.Height * 0.9f);
+            float bottomCutoff = currentScroll + scrollContainer.DisplayableContent * bottom_visibility_cutoff_multiplier;
+
+            float top = Math.Max(sectionPosition, topCutoff);
+            float bottom = Math.Min(sectionPosition + section.Height, bottomCutoff);
+            return bottom - top;
         }
```
peppy, line 290: if you do continue work on this, remove this complete code. this should be possible with <10 loc. i cannot understand any of this.

PR #11031, opened 2020-12-01
Fix editor beat snapping not working correctly when starting with a new beatmap

### LINKED ISSUE #11018: osu! Editor not snapping to Timeline ticks after Timing Point has Changed
**Describe the bug:** 
When setting timing for the first time in a Map and going back to the Compose tab, When trying to scroll in the Timeline you'd expect the Cursor to Snap to the Individual Timeline ticks, but in this case It doesn't.

Steps to recreate (hopefully):
1. Start mapping a new Song
2. Create a new Timing Point
3. Set it to some BPM
3. Go to Compose Tab and Scroll for a bit
4. Try Scrolling a bit
5. Change the BPM of the Timing Point to something lower/higher
6. Go Back and Try Scrolling, It won't snap to Timeline ticks anymore after 

https://streamable.com/o31u9t

**osu!lazer version:** 
2020.1128.0

**Logs:**
[performance.log](https://github.com/ppy/osu/files/5617455/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5617456/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5617458/updater.log)
[database.log](https://github.com/ppy/osu/files/5617459/database.log)
[network.log](https://github.com/ppy/osu/files/5617460/network.log)

I hope my recreation steps arent too rough and actually work. I hope i can help with this somewhat.
Ill be happy to help incase any more information is needed!

peppy:
Closes https://github.com/ppy/osu/issues/11018

Also fixes various other issues with seeking and timing (as can be expected by the fix...)

### DIFF
--- a/osu.Game/Screens/Edit/Editor.cs
+++ b/osu.Game/Screens/Edit/Editor.cs
@@ -106,14 +106,14 @@ namespace osu.Game.Screens.Edit
         [BackgroundDependencyLoader]
         private void load(OsuColour colours, GameHost host, OsuConfigManager config)
         {
-            beatDivisor.Value = Beatmap.Value.BeatmapInfo.BeatDivisor;
-            beatDivisor.BindValueChanged(divisor => Beatmap.Value.BeatmapInfo.BeatDivisor = divisor.NewValue);
-
             if (Beatmap.Value is DummyWorkingBeatmap)
             {
                 isNewBeatmap = true;
                 Beatmap.Value = beatmapManager.CreateNew(Ruleset.Value, api.LocalUser.Value);
             }
+
+            beatDivisor.Value = Beatmap.Value.BeatmapInfo.BeatDivisor;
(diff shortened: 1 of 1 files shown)


osu.Game/Screens/Edit/Editor.cs
```diff
@@ -109,6 +109,12 @@ private void load(OsuColour colours, GameHost host, OsuConfigManager config)
             beatDivisor.Value = Beatmap.Value.BeatmapInfo.BeatDivisor;
             beatDivisor.BindValueChanged(divisor => Beatmap.Value.BeatmapInfo.BeatDivisor = divisor.NewValue);
```
bdach, line 110: this also accesses `Beatmap.Value`, which I assume was the whole root cause of the issue (clock was being created using dummy beatmap instead of new one) - maybe the beatmap creation code should be at the very start of `load()` for complete sanity?

peppy: hmm, good catch. this just means that the divisor won't be saved/restored between sessions, but still worth fixing.

smoogipoo (approved):

PR #11032, opened 2020-12-01
Fix timeline not updating ticks correctly after arbitrary timing control point changes

peppy:
Done using change handler as we don't yet have an easy way of monitoring `ControlPointInfo` for all changes (without bindings on three or four levels).


bdach:
Pushed in a change aiming to resolve the actual test failures. Other than that, seems good :+1:

bdach (approved):

smoogipoo (approved):

PR #11033, opened 2020-12-01
Make HitSampleInfo immutable

smoogipoo:
I'm a little bit hesitant to PR this, but I want to hear your thoughts.

In https://github.com/ppy/osu/pull/10905, I had to add inspection disables to `GetHashCode()` implementations because they reference mutable properties (`Name`, `Bank`,  etc). With the editor, I think having mutable samples is also pretty unsafe as it means we can't respond to changes unless we make their properties into bindables which in turn blows the complexity of hitsamples out of proportion.

Both of the above point towards making `HitSampleInfo` (and its derivatives) immutable. Ideally I'd also like to make them structs, but that's not possible due to `LookupNames` overrides. The usage then becomes similar to that of `FontUsage` - ctor params + `With()` method to return a new instance with any overridden properties.

One distinction between this and `FontUsage` is the usage of the new `Optional` struct, because I think we may want to do something like `sample.With(bank: null)` in the future. I believe we may want this in o!f itself for places like `FontUsage`, so if the direction of this PR is okay, I can move that struct to o!f.

I think when we update to C# 9 in the future, [records](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/records) may be usable to simplify the implementation, which have their own `with` syntax, although this syntax doesn't seem well supported in Rider EAP either.

# Breaking Changes

## `HitSampleInfo` has been made immutable

* Use constructor arguments instead of object initialisers.
* To change a property, use `sample.With(...)` (returns a new instance).
* If overridden, make sure to implement `IEquatable` and also override `With()`.

### DIFF
--- a/osu.Game/Audio/HitSampleInfo.cs
+++ b/osu.Game/Audio/HitSampleInfo.cs
@@ -83,7 +83,7 @@ namespace osu.Game.Audio
             => other != null && Name == other.Name && Bank == other.Bank && Suffix == other.Suffix;
 
         public override bool Equals(object? obj)
-            => Equals((HitSampleInfo?)obj);
+            => obj is HitSampleInfo other && Equals(other);
 
         public override int GetHashCode() => HashCode.Combine(Name, Bank, Suffix);
     }
--- a/osu.Game/Audio/HitSampleInfo.cs
+++ b/osu.Game/Audio/HitSampleInfo.cs
@@ -72,13 +72,13 @@ namespace osu.Game.Audio
         /// <summary>
         /// Creates a new <see cref="HitSampleInfo"/> with overridden values.
         /// </summary>
-        /// <param name="name">An optional new sample name.</param>
-        /// <param name="bank">An optional new sample bank.</param>
-        /// <param name="suffix">An optional new lookup suffix.</param>
-        /// <param name="volume">An optional new volume.</param>
+        /// <param name="newName">An optional new sample name.</param>
+        /// <param name="newBank">An optional new sample bank.</param>
+        /// <param name="newSuffix">An optional new lookup suffix.</param>
+        /// <param name="newVolume">An optional new volume.</param>
         /// <returns>The new <see cref="HitSampleInfo"/>.</returns>
-        public virtual HitSampleInfo With(Optional<string> name = default, Optional<string?> bank = default, Optional<string?> suffix = default, Optional<int> volume = default)
-            => new HitSampleInfo(name.GetOr(Name), bank.GetOr(Bank), suffix.GetOr(Suffix), volume.GetOr(Volume));
+        public virtual HitSampleInfo With(Optional<string> newName = default, Optional<string?> newBank = default, Optional<string?> newSuffix = default, Optional<int> newVolume = default)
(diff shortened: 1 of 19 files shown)


osu.Game/Audio/HitSampleInfo.cs
```diff
@@ -56,6 +69,23 @@ public virtual IEnumerable<string> LookupNames
...
-        public HitSampleInfo Clone() => (HitSampleInfo)MemberwiseClone();
+        /// <summary>
+        /// Creates a new <see cref="HitSampleInfo"/> with overridden values.
+        /// </summary>
+        /// <param name="name">An optional new sample name.</param>
+        /// <param name="bank">An optional new sample bank.</param>
+        /// <param name="suffix">An optional new lookup suffix.</param>
+        /// <param name="volume">An optional new volume.</param>
+        /// <returns>The new <see cref="HitSampleInfo"/>.</returns>
+        public virtual HitSampleInfo With(Optional<string> name = default, Optional<string?> bank = default, Optional<string?> suffix = default, Optional<int> volume = default)
```
bdach, line 80: Nitpicky but I'd maybe change the params to be `newName`, `newBank`, `newSuffix`, `newVolume`, to read better.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
```diff
@@ -523,25 +477,61 @@ public int CustomSampleBank
...
+
+            public LegacyHitSampleInfo(string name, string? bank = null, int volume = 0, int customSampleBank = 0, bool isLayered = false)
+                : base(name, bank, customSampleBank >= 2 ? customSampleBank.ToString() : null, volume)
+            {
+                CustomSampleBank = customSampleBank;
+                IsLayered = isLayered;
+            }
+
+            public override HitSampleInfo With(Optional<string> name = default, Optional<string?> bank = default, Optional<string?> suffix = default, Optional<int> volume = default)
+                => With(name, bank, volume);
```
bdach, line 490: This is probably the shadiest corner of this PR, what with `suffix` getting outright dropped if somebody uses this override by accident - not seeing a good way to avoid this, though... I'd probably outright throw and advise not to use this instead.

bdach (commented):
I honestly don't hate this. Sure it's awkward around a few places and there's clearly a lack of support from the language for this style of programming at this stage but I see the changes at the usage sites as non-negligible code quality improvements. Am definitely curious of @peppy's opinion, though.

osu.Game/Rulesets/Objects/Legacy/ConvertHitObjectParser.cs
smoogipoo: I'm fine with throwing if suffix is non-default in this case, however what should I do about `FileHitSampleInfo`? It discards all except volume, and it's going to hit all of those via `SampleControlPoint.ApplyTo(sample)` that DHOs call.

peppy (approved):

peppy:
no issues with this, feels better IMO.

Issue #10580, opened 2020-10-22
Display the resultant difficulty settings after speed/rate changing


vernonlim:
Currently, only the star rating display is changed with DT/speed changes, with no indication that other settings like AR or OD are affected. In stable, a small arrow is put next to the difficulty setting to indicate that it is raised (or lowered with HT), and the millisecond values displayed when hovered over are changed. 

However, this isn't the terminology players use. In scoreposts and conversation, players use the AR/OD values calculated with formulas when discussing rate changed maps. An example is referring to AR9 + DT as AR10.3/AR10.33 and OD10 + DT as OD11/OD11.11 (decimals are usually omitted for convenience). It would be much less confusing for players if these values were displayed directly in the menu. 

Instead of just SR, the other difficulty settings could be made red and increased to their values with the speed change (EDIT: in this case DT).
![image](https://user-images.githubusercontent.com/25032854/96833992-d6a71200-1473-11eb-8272-ee1c9924d29f.png)

An example of this already implemented in the third party client McOsu, though without SR calculation due to it not being implemented. This implementation actually displays values rounded to 1 decimal place in the menu, and values rounded to 2 in the mod menu (in lazer, this should probably be all made 2 to be consistent with SR)
![image](https://user-images.githubusercontent.com/25032854/96834255-474e2e80-1474-11eb-93bf-b65e38778fcb.png)
![image](https://user-images.githubusercontent.com/25032854/96834371-7b295400-1474-11eb-8369-d1a1a631c771.png)

For ease of gameplay customization, this should somehow be applied to the 'Difficulty Adjust' mod as well (like it is in McOsu).
![image](https://user-images.githubusercontent.com/25032854/96834850-479af980-1475-11eb-9b87-1ba9e1191cb4.png)

First, the values could be displayed permanently somewhere. Having to hover over a slider in order to see what value it is at is very inconvenient. (Could be another issue, but is required for this)

Then values after speed changing could be displayed next to the original values when DT/HT is selected. This way, players can see what the original value they are setting is, and what value they are going to get when playing with the selected speed. 

The display of both original and after-speed-changing values is important since otherwise it would get confusing when working with complicated combinations of difficulty settings and speed changes (this is a flaw with McOsu's implementation from my own experience).

GameBoyYeet:
> First, the values could be displayed permanently somewhere. Having to hover over a slider in order to see what value it is at is very inconvenient. (Could be another issue, but is required for this)

This is especially true for mobile.

[...]

PR #11035, opened 2020-12-01
Add playlist length to match settings overlay

### LINKED ISSUE #11002: Multiplayer playlist can be longer than lobby duration
**osu!lazer version:** 2020.1128.0

smoogipoo:
Fixes https://github.com/ppy/osu/issues/11002 ?

Various displays at different lengths:
Screenshot_2020-12-01_19-59-08

Screenshot_2020-12-01_19-59-20

Screenshot_2020-12-01_19-59-34

### DIFF
--- a/osu.Game/Online/API/Requests/Responses/APIBeatmap.cs
+++ b/osu.Game/Online/API/Requests/Responses/APIBeatmap.cs
@@ -75,6 +75,7 @@ namespace osu.Game.Online.API.Requests.Responses
                 StarDifficulty = starDifficulty,
                 OnlineBeatmapID = OnlineBeatmapID,
                 Version = version,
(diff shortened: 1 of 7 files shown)


bdach:
Rate-changing mods don't affect the runtime while they probably should, but that's an issue that we have all over the place so I wouldn't necessarily hold up this PR for that as having this is better than nothing. Adjusting the runtime to mods can happen as part of #10580 or something.

bdach (dismissed):

peppy (dismissed):

peppy:
@smoogipoo I've added this display to the room screen as well. This required a slight change to the osu-web transformer, which should be deployed today. Please sanity check code changes.

peppy:
Is deployed now

image

smoogipoo:
Can also be added to the room inspector in the listing. I'll do it.

peppy (dismissed):

peppy (approved):

PR #11047, opened 2020-12-01
Add "Start cinema mod with Ctrl+Shift+Enter"

### LINKED ISSUE #9221: Cinema mod shortcut not implemented

orozso:
Add hotkey to "cinema" mod as in osu!stable.

Closes https://github.com/ppy/osu/issues/9221.

### DIFF
--- a/osu.Game/Screens/Select/PlaySongSelect.cs
+++ b/osu.Game/Screens/Select/PlaySongSelect.cs
@@ -81,7 +81,7 @@ namespace osu.Game.Screens.Select
             // Ctrl+Enter should start map with autoplay enabled.
             if (GetContainingInputManager().CurrentState?.Keyboard.ControlPressed == true)
             {
-                var auto = Ruleset.Value.CreateInstance().GetAutoplayMod();
+                var auto = (GetContainingInputManager().CurrentState?.Keyboard.ShiftPressed == true) ? Ruleset.Value.CreateInstance().GetCinemaMod() : Ruleset.Value.CreateInstance().GetAutoplayMod();
                 var autoType = auto?.GetType();
 
                 var mods = Mods.Value;
@@ -90,7 +90,7 @@ namespace osu.Game.Screens.Select
                 {
                     notifications?.Post(new SimpleNotification
                     {
-                        Text = "The current ruleset doesn't have an autoplay mod avalaible!"
+                        Text = "The current ruleset doesn't have an autoplay/cinema mod avalaible!"
                     });
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


bdach:
This has been attempted once before in #9224, and as far as I can see, [this implementation suffers from the exact same bug that prevented that one from being merged](https://github.com/ppy/osu/pull/9224#issuecomment-640112999).

bdach (changes requested):
as above - please add automated test coverage for the aforementioned edge case if you decide to resolve

peppy:
I'm not even sure we want this. Cinema is such a rare use case that I feel it probably doesn't need its own dedicated (hidden) hotkey..

osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -108,5 +128,13 @@ protected override bool OnStart()
 
             return true;
         }
+
+        private Mod getActiveAutoMod()
+        {
+            if (Mods.Value.Any(m => m.GetType() == Ruleset.Value.CreateInstance().GetAutoplayMod()?.GetType())) return Ruleset.Value.CreateInstance().GetAutoplayMod();
+            else if (Mods.Value.Any(m => m.GetType() == Ruleset.Value.CreateInstance().GetCinemaMod()?.GetType())) return Ruleset.Value.CreateInstance().GetCinemaMod();
```
bdach, line 135: I don't agree with this being handled locally as done here. The higher-level logic of mod selection/deselection should be deselecting one when the other is selected, via `IncompatibleMods` or similar. (see #7155) Otherwise we're going to end up with local fixes every time an issue like this arises.

That was also another problem the previous attempt at this had.

osu.Game/Screens/Select/PlaySongSelect.cs
```diff
@@ -81,23 +92,32 @@ protected override bool OnStart()
...
 
-                var mods = Mods.Value;
-
-                if (autoType == null)
+                if (modType == null)
                 {
                     notifications?.Post(new SimpleNotification
                     {
-                        Text = "The current ruleset doesn't have an autoplay mod avalaible!"
+                        Text = "The current ruleset doesn't have an autoplay/cinema mod avalaible!"
```
bdach, line 102: this text should depend on the hotkey pressed, not be an either/or catch-all

bdach (changes requested):

PR #11055, opened 2020-12-02
Use stateless RNG for osu!catch hit objects

ekrctb:
Motivation: It makes the randomness consistent when a replay is rewound, or a replay is played multiple times.

- Fix random rotation not updated when a DHO is reused <https://github.com/ppy/osu/pull/10979#issuecomment-734626713>.

### DIFF
--- a/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
+++ b/osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
@@ -3,7 +3,6 @@
 
 using JetBrains.Annotations;
 using osu.Framework.Graphics;
-using osu.Framework.Utils;
 
 namespace osu.Game.Rulesets.Catch.Objects.Drawables
 {
@@ -21,6 +20,17 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
         {
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            RandomSeed.BindValueChanged(_ =>
+            {
+                UpdateComboColour();
+                UpdateInitialTransforms();
+            });
+        }
+
         protected override void UpdateInitialTransforms()
         {
             base.UpdateInitialTransforms();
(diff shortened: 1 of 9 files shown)


osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
```diff
@@ -21,21 +20,32 @@ public DrawableBanana([CanBeNull] Banana h)
...
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            RandomSeed.BindValueChanged(_ =>
+            {
+                UpdateComboColour();
+                UpdateInitialTransforms();
```
peppy, line 30: Not sure how safe this is, since this method was only intended to be called by `DrawableHitObject` itself.

Is it feasible to just note make this see a bindable and assert that it is only changed on construction or pool retrieval instead?

peppy (changes requested):

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
ekrctb: I'm not sure what your expected code. Is it better now?

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
```diff
@@ -20,12 +21,19 @@ public abstract class DrawableCatchHitObject : DrawableHitObject<CatchHitObject>
...
         protected DrawableCatchHitObject([CanBeNull] CatchHitObject hitObject)
             : base(hitObject)
         {
             Anchor = Anchor.BottomLeft;
         }
 
+        /// <summary>
+        /// Get a random number in range [0,1) based on seed <see cref="RandomSeed"/>.
+        /// </summary>
+        public float RandomSingle(int series) => StatelessRNG.NextSingle(RandomSeed, series);
```
bdach, line 35: Is there a particular reason why this and `RandomSeed` are public here? Would rather prefer `protected float RandomSingle(int)` and `private int randomSeed` if at all possible.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableFruit.cs
```diff
@@ -41,6 +38,13 @@ private void load()
             HyperDash.BindValueChanged(_ => updatePiece(), true);
         }
 
+        protected override void UpdateInitialTransforms()
+        {
+            base.UpdateInitialTransforms();
+
+            ScaleContainer.Rotation = (RandomSingle(1) - 0.5f) * 40;
```
bdach, line 45: Not sure about having a plain property set in `UpdateInitialTransforms()`. Would probably make this a

```suggestion
            ScaleContainer.RotateTo((RandomSingle(1) - 0.5f) * 40);
```

for consistency at least I guess? There's no other usage that just does plain sets on drawable properties as far as I can tell (there are some that set lifetime, but that's not a transformable property anyway), but I'm not sure if that has any functional difference (it shouldn't from what I know of how transforms work).

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCatchHitObject.cs
ekrctb: I originally wanted to use it here <https://github.com/ppy/osu/blob/a16b265090942f7c5fa06234978005776586ff5b/osu.Game.Rulesets.Catch/UI/Catcher.cs#L211-L213> (not included in this PR).
I also imagine a use case where mod want to apply custom random transformation to a hit object.

bdach (approved):
seems fine to me :+1:

peppy (approved):
Seems good

PR #11060, opened 2020-12-02
Fix MouseHandler not being ignored when Raw Input is enabled

FamousPig:
Fixes [this issue discussed on discord](https://discord.com/channels/188630481301012481/589331078574112768/783394725264228362). Basically raw input is behaving incorrectly with way too high sensitivity and some sort of negative acceleration.

"Inspired" by [this comment](https://github.com/ppy/osu-framework/pull/3978#discussion_r519170109)

I admit this is a rather dirty fix but I personally don't think this code is much worse off than before. I think the real fix involves changes I am not capable of implementing properly myself. In that way this is more of a proposal than usual.

Also seems to fix part of https://github.com/ppy/osu-framework/issues/4050 on my machine.The cursor no longer gets stuck, though I can replicate [this  teleporting behavior](https://github.com/ppy/osu-framework/issues/4050#issuecomment-737060233).

No testing done on tablet.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
@@ -6,6 +6,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Configuration;
 using osu.Framework.Graphics;
+using osu.Framework.Input.Handlers.Mouse;
 using osu.Game.Configuration;
 using osu.Game.Graphics.UserInterface;
 using osu.Game.Input;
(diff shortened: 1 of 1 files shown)


bdach:
Right, I forgot about this. After enabling raw input on my linux box I can reproduce the cursor getting stuck in top-left corner.

This solution is probably good enoough to work with right now, although I think the fact that `MouseHandler` is public in framework is a mistake (it should not be), so it should probably be a hard-coded string in either case.

FamousPig:
Is this what you had in mind. My concern is that the issue will return if the name of MouseHandler is changed(to SDLMouseHandler or something).

bdach:
It was actually the plan to change it to that at some point. And yeah, it will return if that's missed again, but the entire API for using raw input is really bad right now. Exposing mouse handlers just for that purpose is kind of just as bad, though.

@peppy think this is fine for now? maybe we should have at least a temporary raw input bindable framework-side or something?

smoogipoo (approved):
This looks fine for now - it's what framework does in its test too. Wouldn't be against a framework-side raw input toggle but that can happen as a further step.

peppy (approved):
Good first step.

PR #11066, opened 2020-12-03
Fix DrawableHitObjects potentially expiring before their hit samples finish playback

### LINKED ISSUE #11057: osu!catch hitsounds are cut off
Regressed by #11011

It sounds like some hit sound is stopped after very short time while playing.

ekrctb:
Fix #11057

Not sure how this is related to the previous similar issue fix #10373.

### DIFF
--- a/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
+++ b/osu.Game/Rulesets/Objects/Drawables/DrawableHitObject.cs
@@ -453,7 +453,7 @@ namespace osu.Game.Rulesets.Objects.Drawables
             state.Value = newState;
 
             if (LifetimeEnd == double.MaxValue && (state.Value != ArmedState.Idle || HitObject.HitWindows == null))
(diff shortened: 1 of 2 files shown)


bdach:
The issue this aims to fix appears to have disappeared or at least been made less frequent on `master` after merge of the sample pooling PRs (https://github.com/ppy/osu/pull/10905 + https://github.com/ppy/osu/pull/11033), although there are still singular hitsounds that sound wrong. Worse yet, it seems that some hitsounds can apparently be freed by their DHOs while still playing, and the lifetime extension in this PR does not seem to alleviate that, either.

Not sure about what the best course of action here is, or why the behaviour changed exactly. Will circle back to this if I manage to get through today's PR queue quickly enough.

peppy (approved):

peppy (approved):

peppy:
@smoogipoo can you double-check this one?

PR #11070, opened 2020-12-03
Fix song select panels not loading if partially offscreen

### LINKED ISSUE #10745: Beatmap panels don't get loaded early enough
Example: https://drive.google.com/file/d/142MAeVA1dVTh8slhJwZ7Y6tAPlpAeje4/view?usp=sharing

The top-most beatmap set panel only gets loaded after it fully crosses into the undimmed area.

smoogipoo:
Fixes https://github.com/ppy/osu/issues/10745

The DLW starts with 0 size because the construction of the content is delayed until the DLW starts loading. This gives it an initial size (`RelativeSizeAxes = Both`).


peppy:
This has come up in the past too hasn't it.. almost feels like we want to have an explicit `bool` specifying whether `DelayLoadWrapper`s should start relatively sized..

peppy (approved):

PR #11071, opened 2020-12-03
Refactor fruit catching logic

ekrctb:
To make the future implementation of caught object pooling easier.

- Move fruit catching logic to `Catcher`.
  - I think it is more logical.
- Simplify fruit catching code.
  - No longer wait for DHO loading until the fruit is dropped.
- Found a possibly-unintended behavior in fruit stacking logic
  - `f.DrawSize.X / 2 * f.Scale.X` was always equal to constant `CatchHitObject.OBJECT_RADIUS / 2`, because fruit scaling is done in the child `ScalingContainer`.
- Simplify fruit dropping code.
  - There was a repeated `Remove` call which has a quadratic time complexity, it should be near-linear time now.
- `CatcherArea` tests are moved to `Catcher` test scene, as the relevant code is moved.
- Add more `Catcher` tests.
  - Found another possibly-unintended behavior: circle size scaling is applied twice to the fruits on the plate.
    - This can be changed when a caught object is no longer a DHO.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
@@ -15,7 +15,7 @@ namespace osu.Game.Rulesets.Catch.Tests
         [BackgroundDependencyLoader]
         private void load()
         {
-            SetContents(() => new Catcher(new Container())
+            SetContents(() => new Catcher(new Container(), new Container())
             {
                 RelativePositionAxes = Axes.None,
                 Anchor = Anchor.Centre,
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
@@ -1,26 +1,192 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
 using osu.Framework.Allocation;
 using osu.Game.Rulesets.Catch.UI;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
+using osu.Framework.Testing;
+using osu.Game.Beatmaps;
+using osu.Game.Beatmaps.ControlPoints;
+using osu.Game.Configuration;
+using osu.Game.Rulesets.Catch.Objects;
+using osu.Game.Rulesets.Catch.Objects.Drawables;
+using osu.Game.Tests.Visual;
 
 namespace osu.Game.Rulesets.Catch.Tests
 {
     [TestFixture]
-    public class TestSceneCatcher : CatchSkinnableTestScene
+    public class TestSceneCatcher : OsuTestScene
     {
-        [BackgroundDependencyLoader]
-        private void load()
+        [Resolved]
+        private OsuConfigManager config { get; set; }
+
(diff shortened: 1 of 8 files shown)


osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -479,33 +454,109 @@ private void updateState(CatcherAnimationState state)
...
+        private void drop(Drawable d, DroppedObjectAnimation animation)
+        {
+            var originalX = d.X * Scale.X;
 
-            fruit.ApplyCustomUpdateState += onFruitOnApplyCustomUpdateState;
-            onFruitOnApplyCustomUpdateState(fruit, fruit.State.Value);
+            d.Anchor = Anchor.TopLeft;
+            d.Position = caughtFruitContainer.ToSpaceOfOtherDrawable(d.DrawPosition, droppedObjectTarget);
+
+            animate(d, animation, originalX);
```
peppy, line 527: I believe you're going to need to re-add these animations after a skin change (since an `updateState` is forced by `DrawableHitObject` on skin change):




You can test this easily using the "random skin" hotkey (Ctrl+Shift+R)

peppy (changes requested):
Code looks a *lot* better than it used to.

bdach (changes requested):
Generally like this, but some things can still be improved upon.

osu.Game.Rulesets.Catch.Tests/TestSceneCatcher.cs
```diff
@@ -1,26 +1,192 @@
...
+        }
+
+        [Test]
+        public void TestCatcherStateBanana()
+        {
+            AddStep("catch hyper kiai fruit", () => attemptCatch(new TestKiaiFruit
+            {
+                HyperDashTarget = new Fruit { X = 100 }
+            }));
+            AddStep("miss banana", () => attemptCatch(new Banana()));
```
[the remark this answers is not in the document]
ekrctb: Nice catch. It was meant to be `TestBananaMissPreservesCatcherState`.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -479,33 +453,109 @@ private void updateState(CatcherAnimationState state)
...
-            fruit.ApplyCustomUpdateState += onFruitOnApplyCustomUpdateState;
-            onFruitOnApplyCustomUpdateState(fruit, fruit.State.Value);
+            d.Anchor = Anchor.TopLeft;
+            d.Position = caughtFruitContainer.ToSpaceOfOtherDrawable(d.DrawPosition, droppedObjectTarget);
 
-            void onFruitOnApplyCustomUpdateState(DrawableHitObject o, ArmedState state)
+            // we cannot just apply the transforms because DHO clears transforms when state is updated
+            d.ApplyCustomUpdateState += (o, state) => animate(o, animation, originalX, startTime);
+            if (d.IsLoaded)
+                animate(d, animation, originalX, startTime);
```
peppy, line 530: another unfortunate case where it would have been great if we could use the `UpdateState` bindable, but i'm guessing we can't due to state changes not always being fired when we expect them to be.

peppy:
@ekrctb I made some code quality changes, can you make sure you're okay with them? Specific attention to the public methods I made private (in case you had intention to use them in the future for testing or otherwise).

ekrctb:
@peppy The changes look good to me.

peppy (approved):

peppy:
@bdach want to give this a quick re-review before merge?

bdach:
Seems ok from a rapid read-through.

bdach (approved):

PR #11073, opened 2020-12-03
Remove OnParentReceived() from DrawableHitObject

smoogipoo:
`nested.OnParentReceived()` would be called after `nested.Apply()`, which was a weird ordering - `OsuModHidden` depends on knowing the parenting DHO by the time `ApplyCustomUpdateState` is invoked.

Now, the new `ParentHitObject` property is assigned pre-`Apply()` such that the DHO has access to both the local object (via `HitObject`) and the parent object (via `ParentHitObject`) inside `OnApply()`. Any code that previously existed inside `OnParentReceived()` can now be safely moved into `OnApply()`.

The order of method calls is now the following:

```
Parent.Apply()
    Nested.Apply()
        Nested.OnApply()
        Nested.updateState()*
    Parent.OnApply()
    Parent.updateState()*

* The order of updateState() can vary depending on the IsLoaded state.
  If not loaded, then updateState() happens at a later point in time (first update) and in hierarchical order.
```

### DIFF
--- a/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderHead.cs
+++ b/osu.Game.Rulesets.Osu/Objects/Drawables/DrawableSliderHead.cs
@@ -4,20 +4,19 @@
 using System;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
-using osu.Game.Rulesets.Objects.Drawables;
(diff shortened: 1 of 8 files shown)


bdach (approved):
seems correct as far as I can tell :+1:  would appreciate a double-check though

peppy (approved):

PR #11083, opened 2020-12-03
Select recommended beatmap if last selection is filtered

### LINKED ISSUE #8757: Carousel should select recommended beatmap on ruleset change
**Describe the bug:** 
Right now when you change ruleset it is ``CarouselGroupEagerSelect`` that makes the selection (based on the index, so if you go to a lower ID ruleset, the highest difficulty is selected, if you go to a higher ID ruleset, lowest difficulty is selected). ``CarouselBeatmapSet`` should overwrite its selection strategy better so that on ruleset change recommended difficulty from that ruleset is selected.

**osu!lazer version:** 
master is at ecd40072c441b42606982656d38c1e4a908952ef

LittleEndu:
Closes #8757

Tests are passing so I guess this isn't a major change to how beatmap carousel works. Only unintuitive thing I can imagine that could happen is if you somehow manage to filter out only the selected difficulty from others, the recommended difficulty is selected instead of next.

Can write tests if this feature is deemed important enough.

Edit: Or can PR something like https://github.com/ppy/osu/compare/master...LittleEndu:ruleset-change-recommended instead if this should only run on ruleset change

### DIFF
--- a/osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
+++ b/osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
@@ -45,7 +45,7 @@ namespace osu.Game.Screens.Select.Carousel
 
         protected override CarouselItem GetNextToSelect()
         {
-            if (LastSelected == null)
+            if (LastSelected == null || LastSelected.Filtered.Value)
             {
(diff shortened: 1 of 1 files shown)


peppy:
I think this is fine for now. Rather than over-thinking the case where selection is filtered, let's wait for someone to actually complain about it, as it will add further complexity to an already complex system. This change is quite trivial so I'm happy for it to go ahead as-is.

peppy (approved):

PR #11069, opened 2020-12-03
Fix hidden mod crashing in osu! ruleset

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/11073

Fixes https://github.com/ppy/osu/issues/11015

This became easier to repro after https://github.com/ppy/osu/pull/10985 hence the test I added.

This PR reworks `OsuModHidden` to look up the slider head rather than storing locally. It also moves the initial fade adjustment into `ApplyToBeatmap`- not sure why this wasn't done from the get-go.

PR #11084, opened 2020-12-03
Fix spinner ticks getting increased visibility state in hidden mod

bdach:
- [x] Depends on #11073 (first in chain).
- [x] Depends on #11069 (for ease of mergeability).

Noticed while reviewing the aforementioned PR(s).

# Summary

The increased visibility logic for osu! hidden has regressed in #10696. The old `IsFirstHideableObject()` method did not
consider nested hitobjects, while its replacement - `IsFirstAdjustableObject()` - did. Therefore, spinner ticks could be
considered first adjustable objects, breaking the old logic.

To resolve, exclude ticks from being considered for increased vislibility.

# Remarks

There is no need to match over `SpinnerBonusTick`, as it inherits from `SpinnerTick`.

This now has actual automated test coverage, as this is the second time this has regressed. Previous coverage was purely visual.

As for the test code,`GameplayBeatmap` has to be used instead of the normal bindable `Beatmap`, beecause the former uses osu!-specific hitobjects, while the latter returns convert objects (i.e. `ConvertSlider`s).

Similarly, the mod has to be fetched from the player instead of the global bindable, as `Player` has its own cloned instance of the mod, to which the beatmap is applied. The global bindable instance does not have `FirstObject` set.

### DIFF
--- a/osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
+++ b/osu.Game.Rulesets.Osu.Tests/Mods/TestSceneOsuModHidden.cs
@@ -2,12 +2,15 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System.Collections.Generic;
+using System.Linq;
 using NUnit.Framework;
+using osu.Framework.Testing;
 using osu.Game.Beatmaps;
 using osu.Game.Rulesets.Objects;
 using osu.Game.Rulesets.Objects.Types;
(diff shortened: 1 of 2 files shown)


smoogipoo (approved):

peppy (approved):
I feel it's a bit weird that the lookup now includes nested (as the parent object should probably be responsible for handling this kind of state declaration) but it is what it is.

PR #11091, opened 2020-12-04
Move `Piece*` classes under `Skinning` namespace

ekrctb:
### Motivation

`Skinning.Legacy*` classes and `Objects.Drawables.Pieces.*Piece` classes serve similar purpose but the namespaces are far apart.

This PR moves:

- `Objects.Drawables.Pieces.*` -> `Skinning.Default` new namespace
- `Skinning.Legacy*Piece` -> `Skinning.Legacy` new namespace.

Opinions welcome (another namespace suggestions etc.).


bdach:
I don't have any objections to this change in general, albeit I'm not sure why this PR is a draft.

peppy:
Hmm.. I'm not 100% sure on this one. The idea behind legacy namespaces was that the new components (pieces) weren't to be considered "default" as much as "the future", including support for skins going forwards. But maybe this makes more sense? Not sure.

May also need to consider that this is potentially a breaking change, since the `piece` classes are exposed publicly.

@smoogipoo thoughts?

smoogipoo:
I'm approving of this PR because I've also been confused about where the skinnable components reside.

> May also need to consider that this is potentially a breaking change, since the piece classes are exposed publicly.

This is just for the rulesets though, right? There's currently no ruleset that references any of the default ones, and that's not possible via the nuget package anyway as the rulesets aren't included in ppy.osu.Game.

Or am I missing some files?

Another way to look at this - when we implement the new default skin, how will things be laid out? I think this structure makes the path very obvious.

smoogipoo (approved):

peppy:
Ah, right. They are in ruleset projects already. I guess that's okay then.

peppy (approved):
Let's go with this for now and see how it plays out.

PR #5265, opened 2019-07-05
Add android support for importing osz files

EdrowsLuo:
Download osz then open it with lazer.
(reopen using a new branch)

PR #11099, opened 2020-12-05
Add support for importing files on Android

### LINKED ISSUE #6525: Import skins by opening the osk (Android)
I don't know if it's already implemented but:
The possibility to import maps and skins by openning respective files in the app (Click-to-import)

Method: going to file manager, clicking the file, selecting it to open with osu! and import the skin/beatmap

Game4all:
Supersedes #5265
Closes #6525

- [x] Depends on #11115

# Summary

This adds the ability to import files on Android, by adding a way to import files from a Stream and a filename in ArchiveModelManager as starting from 7.0, apps don't expose anymore paths to files but content URIs, which can at best be used to get a stream to the file and the filename. 

Turns out passing the filename around is required because some model managers (namely SkinManager) are using the archive reader name to populate the skin info. 

This adds ability to import beatmap / skins by clicking them in downloads (ie: in a web brower) or by sharing the file to osu! through the share button in a file explorer.

## Testing

I haven't added any automated tests for importing but guess I could add one if prompted to and if it's deemed necessary.
Tested this on three of my android devices which are unfortunately all above > 9.0 so, if anyone owns an actual device running under < 7.0, testing this works would be appreciated.


Game4all:
Actually this isn't ready yet for review and is missing robustness in retrieving filenames correctly so I'll close this and reopen when deemed ready.

Game4all:
I guess this is ready for review. `osu://` URL handling will come in a separate PR.

osu.Android/OsuGameActivity.cs
```diff
@@ -23,8 +30,37 @@ protected override void OnCreate(Bundle savedInstanceState)
...
+
             Window.AddFlags(WindowManagerFlags.Fullscreen);
             Window.AddFlags(WindowManagerFlags.KeepScreenOn);
         }
+
+        protected override void OnNewIntent(Intent intent)
+        {
+            if (intent.Action == Intent.ActionDefault)
+            {
+                if (intent.Scheme == ContentResolver.SchemeContent)
```
bdach, line 43: Would prefer to have these conditions merged.

osu.Android/OsuGameActivity.cs
```diff
@@ -23,8 +30,37 @@ protected override void OnCreate(Bundle savedInstanceState)
...
+            if (intent.Action == Intent.ActionSend)
+            {
+                var content = intent.ClipData.GetItemAt(0);
+                handleImportFromUri(content.Uri);
+            }
+        }
+
+        private void handleImportFromUri(Uri uri)
+        {
+            var cursor = ContentResolver.Query(uri, null, null, null);
```
bdach, line 56: [To quote docs](https://developer.android.com/reference/android/content/ContentResolver#query(android.net.Uri,%20java.lang.String[],%20android.os.Bundle,%20android.os.CancellationSignal)), the second argument is:

> A list of which columns to return. Passing null will return all columns, which is inefficient.

Maybe this should just fetch the filename only?

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -212,6 +212,47 @@ public async Task<TModel> Import(string path, CancellationToken cancellationToke
             return import;
         }
 
+        /// <summary>
+        /// Import one <typeparamref name="TModel"/> from a <see cref="Stream"/>.
+        /// </summary>
+        /// <param name="stream">The stream to import files from.</param>
+        /// <param name="filename">The filename of the archive being imported.</param>
+        public async Task Import(Stream stream, string filename)
```
bdach, line 220: Honestly I'd probably look into making

```csharp
public Task<TModel> Import(ArchiveReader archive, CancellationToken cancellationToken = default)
```

public, create the archive one level higher and just use that overload instead. There's a difference in that the initialisation notification won't fire right now, but maybe that logic should be moved there anyway. Silent imports aren't the best UX.

osu.Game/OsuGame.cs
```diff
@@ -483,13 +483,13 @@ public void PerformFromScreen(Action<IScreen> action, IEnumerable<Type> validScr
         /// <param name="retrieveInstance">A function to retrieve a (potentially not-yet-constructed) target instance.</param>
         /// <param name="action">The action to perform on the instance when load is confirmed.</param>
         /// <typeparam name="T">The type of the target instance.</typeparam>
-        private void waitForReady<T>(Func<T> retrieveInstance, Action<T> action)
+        public void WaitForReady<T>(Func<T> retrieveInstance, Action<T> action)
```
bdach, line 486: Not sure about exposing this as public, feels a bit like making `Schedule()` public. Would probably make `protected` and add a specialised method in `OsuGameAndroid` just for that import path (i.e. move the
```csharp
game.WaitForReady(() => game, _ => Task.Run(() => game.Import(stream, cursor.GetString(filename_column))));
```
thing to `OsuGameAndroid` as a public method and give it some reasonable name).

bdach (dismissed):

osu.Game/Database/ArchiveModelManager.cs
Game4all: I'm failing to understand the intent behind using this overload instead, and looking at the places which use this overload, not sure about moving import status notification logic there  (not sure, maybe move that logic one level higher aswell?). 

As a matter of fact I closed my game a few times while testing skin importing thinking that something went wrong whereas the skin was just taking a bit of time to import, and I hadn't added any import status notification yet,  so signaling the user the import is running is definitely needed, especially on mobile devices which may take a bit more of time to process files.

bdach: Mostly I don't like the fact that the notification logic is partially duplicated inside the new overload, and that the stream+filename combo is passed down sort of unnecessarily. If you find a better way to address any of these then go ahead.

Game4all: ah, I see.  I that case , I think the common notification initialization code could be moved to a protected method , something like `CreateImportNotification()` and then the overload you mentionned earlier could accept optionally a `ProgressNotification` that it would update according to the import state.

osu.Game/Database/ICanAcceptFiles.cs
```diff
@@ -17,6 +18,14 @@ public interface ICanAcceptFiles
         /// <param name="paths">The files which should be imported.</param>
         Task Import(params string[] paths);
 
+        /// <summary>
+        /// Import the specified stream.
+        /// </summary>
+        /// <param name="stream">The stream to import files from.</param>
+        /// <param name="filename">The filename of the archive being imported.</param>
+        /// <returns></returns>
```
peppy, line 26:
```suggestion
        /// <summary>
        /// Import a single file from a stream.
        /// </summary>
        /// <param name="stream">The stream to import the file from. This should be in a supported archive format.</param>
        /// <param name="filename">The filename of the archive being imported.</param>
```

osu.Android/OsuGameAndroid.cs
```diff
@@ -65,6 +67,13 @@ public override Version AssemblyVersion
             }
         }
 
+        /// <summary>
+        /// Schedules a file to be imported once the game is loaded.
+        /// </summary>
+        /// <param name="stream">A stream to the file to import.</param>
+        /// <param name="filename">The filename of the file to import.</param>
+        public void ScheduleImport(Stream stream, string filename) => WaitForReady(() => this, _ => Task.Run(() => Import(stream, filename)));
```
peppy, line 75: Is there a reason we can't just make this standard for the import logic? ie. making `OsuGameBase.Import` virtual and overriding with the `WaitForReady` addition in `OsuGame`.

peppy (changes requested):

osu.Game/Database/ICanAcceptFiles.cs
peppy: Outdated review, see #11115

bdach:
I guess I'll mark this as dependent on #11115 until that's merged.

osu.Game/Database/ArchiveModelManager.cs
Game4all: marking this as resolved, as #11115 handles this in a nice way.

osu.Android/OsuGameAndroid.cs
peppy: This one is still standing. Was there a reason for not changing the way this is done?

Game4all:
Test failure is unrelated an seems to stem from this place : https://github.com/ppy/osu/blob/6da854e37c3eb3b8f426904895f6ece374bea498/osu.Game.Tests/Visual/Gameplay/TestSceneFailJudgement.cs#L26-L33

bdach:
It's a known (albeit rare and not completely explained) failure.

osu.Game/OsuGame.cs
```diff
@@ -426,6 +427,12 @@ public void PresentScore(ScoreInfo score, ScorePresentType presentType = ScorePr
             }, validScreens: new[] { typeof(PlaySongSelect) });
         }
 
+        public override Task Import(Stream stream, string filename)
+        {
+            WaitForReady(() => this, _ => Task.Run(() => base.Import(stream, filename)));
+            return Task.CompletedTask;
+        }
```
bdach, line 434: This override looks unsafe to await on. If I'm reading this correct, the base `Import(Stream, string)` method, if awaited on, won't return control to the caller until the import sequence has actually completed, while this variant is essentially a fire-and-forget: `WaitForReady` may execute the action immediately, but it might also schedule it; worse yet, the action is wrapped in a `Task.Run` anyway. Therefore I imagine this will always return to the caller immediately after the task is started on/assigned to a pool thread.

I think what you might actually want here is to assign the inner task to a variable, pass it to `WaitForReady` so that it can be executed on another thread by `Task.Run`, and then return that stored-but-unstarted task so that its state changes can be observed by callers.

Game4all: Are we ever gonna await on this though? Looking at the other place we are are calling the `Import(string[])`  overload,  we are just firing the import and forgetting about it since the UI logic / error logic is contained further down in the call hierarchy.

Game4all: Actuall, nevermind , the approach given above sounds more logical given we're scheduling the task to be long running.

osu.Android/OsuGameActivity.cs
```diff
@@ -23,8 +30,46 @@ protected override void OnCreate(Bundle savedInstanceState)
 
             base.OnCreate(savedInstanceState);
 
+            OnNewIntent(Intent);
```
peppy, line 33: this seems a bit suspicious. would want to test if it's required, and if so, leave an inline comment explaining why.

peppy (changes requested):

peppy:
I've tidied this up, but haven't tested on device. @bdach are you able to give it a quick run and maybe touch on my remaining review point?

I think it looks okay for a code perspective otherwise, so feel free to comment/merge as required.

bdach:
Yep, I'll fire up windows today and give this a spin. Was on my tentative to-do list anyway.

bdach:
On testing this, I'm getting the following:

```
2020-12-14 19:52:35.601 15937-15937/sh.ppy.osulazer E/AndroidRuntime: FATAL EXCEPTION: main
    Process: sh.ppy.osulazer, PID: 15937
    java.lang.NoSuchMethodError: no non-static method "Landroid/content/ContentResolver;.query(Landroid/net/Uri;[Ljava/lang/String;Landroid/os/Bundle;Landroid/os/CancellationSignal;)Landroid/database/Cursor;"
        at crc64f3e94a995e5d9e96.OsuGameActivity.n_onCreate(Native Method)
        at crc64f3e94a995e5d9e96.OsuGameActivity.onCreate(OsuGameActivity.java:30)
        at android.app.Activity.performCreate(Activity.java:6956)
```

Indeed this overload has only been added in [API level 26.](https://developer.android.com/reference/android/content/ContentResolver#query(android.net.Uri,%20java.lang.String[],%20android.os.Bundle,%20android.os.CancellationSignal)) Guess I'll revert to the old version that's from API 1...

bdach:
Even after that, map imports are still failing here...

```
2020-12-14 19:02:38 [error]: Could not import (1130730 Au5 feat. Trove - The Journey)
2020-12-14 19:02:38 [error]: System.ArgumentException: Archive streams must be Readable and Seekable
2020-12-14 19:02:38 [error]: at SharpCompress.Archives.AbstractArchive`2[TEntry,TVolume].CheckStreams (System.IO.Stream stream) [0x00010] in <84e094a6785f4014b2dc1f205d7b9fa9>:0
2020-12-14 19:02:38 [error]: at System.Linq.
```

@Game4all have you ever seen any of these?

bdach:
I've pushed changes that make this functional to me. They resolve the issues I pointed out above. I additionally reverted import of `.osr` files, as upon trying to import one, it didn't work, as `.osr`s are archives. Adding support for that extension is a bit more involved (need a `ZipUtils.IsZipArchive` analogue for streams, or to restructure the file type handling somewhat), so that can come in a separate pull.

Game4all:
Seing that you ran into some pretty interesting issues while testing this, could you please specify on what hardware did you test this on ? ( OS version + the app used to trigger the import intent).

During the testing I have done on 3 real devices all running Android > 9.0, I never ran into any of the issues aforementioned even though I tested this works correctly with a few file explorers / firefox / chrome / downloads.

The only thing I haven't tested is replay files import (that one is actually entirely my fault, never bothered to check that replay importing was working, else would've noticed it wasn't functional)

bdach:
Tested on Samsung Galaxy J7 (2016), Android version 7.0. Explains the issue with `ContentResolver` (that one's partially on me too as I suggested it off API docs not having looked at API versions properly), but the stream seekability, not so much.

From what I can see the `InputStream` returned by the content resolver [has no members for seeking](https://developer.android.com/reference/java/io/InputStream), so this likely was depending on underlying implementation details on higher API versions.

bdach (dismissed):

osu.Android/OsuGameActivity.cs
```diff
@@ -23,8 +32,65 @@ protected override void OnCreate(Bundle savedInstanceState)
...
+            var filenameColumn = cursor.GetColumnIndex(OpenableColumns.DisplayName);
+
+            var stream = ContentResolver.OpenInputStream(uri);
+            string filename = cursor.GetString(filenameColumn);
+
+            if (stream != null)
+                Task.Factory.StartNew(() => runImport(stream, filename), TaskCreationOptions.LongRunning);
+        }
+
+        private Task runImport(Stream stream, string filename)
```
peppy, line 85: thinking we want to dispose the original stream here (after copying it out)? probably safest to.

peppy (changes requested):

osu.Android/OsuGameActivity.cs
bdach: Yeah that's a fair point. I'll grab that later.

bdach: I think I've resolved this, but not in the most immediate way - apparently [it's inadvisable to do `ContentProvider` stuff on the main UI thread](https://developer.android.com/guide/topics/providers/content-provider-basics#SimpleQuery) so the whole rigamarole is on a background task now.

peppy (approved):

PR #11101, opened 2020-12-05
Hide confine mode dropdown when full-screen

### LINKED ISSUE #11097: Confine mode setting dropdown has no effect when in exclusive fullscreen
**Describe the bug:** 

Whatever option I choose, the mouse cursor is always confined both in game and menu. This did not happen on yesterday's release.

Quick demo: https://wala.s-ul.eu/IczCzITf

Monitor setup:

20201204T203432-1837237546

**osu!lazer version:** 
Latest release 2020.1204.0 / 9064ca9

**Logs:**

[logs.zip](https://github.com/ppy/osu/files/5645406/logs.zip)

bdach:
Closes #11097.

# Summary

After hard-locking the mouse confine mode to `Always` in full-screen to prevent confine issues from popping up, the confine mode dropdown in settings had confusing UX due to seemingly having no effect when full-screen.

This change aims to block modification to said dropdown when (and only when) the game is in full-screen.

# Remarks

The method of achieving the underlying functionality has changed slightly (framework-side `Always` will be set instead of `Fullscreen` while full-screen), but I think leasing was a good pattern to use here.

Additionally added test coverage to this, due to growing levels of complexity around confine logic.

### DIFF
--- a/osu.Game/Input/ConfineMouseTracker.cs
+++ b/osu.Game/Input/ConfineMouseTracker.cs
@@ -28,37 +28,26 @@ namespace osu.Game.Input
         {
             frameworkConfineMode = frameworkConfigManager.GetBindable<ConfineMouseMode>(FrameworkSetting.ConfineMouseMode);
             frameworkWindowMode = frameworkConfigManager.GetBindable<WindowMode>(FrameworkSetting.WindowMode);
-            frameworkWindowMode.BindValueChanged(_ => updateGameConfineMode());
+            frameworkWindowMode.BindValueChanged(_ => updateConfineMode());
 
             osuConfineMode = osuConfigManager.GetBindable<OsuConfineMouseMode>(OsuSetting.ConfineMouseMode);
             localUserPlaying = game.LocalUserPlaying.GetBoundCopy();
 
-            osuConfineMode.ValueChanged += _ => updateFrameworkConfineMode();
-            localUserPlaying.BindValueChanged(_ => updateFrameworkConfineMode(), true);
+            osuConfineMode.ValueChanged += _ => updateConfineMode();
+            localUserPlaying.BindValueChanged(_ => updateConfineMode(), true);
         }
 
-        private LeasedBindable<OsuConfineMouseMode> leasedOsuConfineMode;
-
-        private void updateGameConfineMode()
-        {
-            if (frameworkWindowMode.Value == WindowMode.Fullscreen && leasedOsuConfineMode == null)
-            {
-                leasedOsuConfineMode = osuConfineMode.BeginLease(true);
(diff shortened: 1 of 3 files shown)


osu.Game/Input/ConfineMouseTracker.cs
```diff
@@ -28,26 +28,37 @@ private void load(OsuGame game, FrameworkConfigManager frameworkConfigManager, O
...
+        private LeasedBindable<OsuConfineMouseMode> leasedOsuConfineMode;
+
+        private void updateGameConfineMode()
         {
-            // confine mode is unavailable on some platforms
-            if (frameworkConfineMode.Disabled)
-                return;
+            if (frameworkWindowMode.Value == WindowMode.Fullscreen && leasedOsuConfineMode == null)
+            {
+                leasedOsuConfineMode = osuConfineMode.BeginLease(true);
```
peppy, line 46: not sure the lease is necessary - is there a specific reason for this being done? just comparing to other places in settings we do this kind of disabling and they don't seem to be leasing, not sure it's achieving much.

bdach: It's mostly for visual reasons. This way, when entering fullscreen the dropdown will gray out and its text will say "Always" until the window mode is changed to windowed/borderless again, at which point the old value will be restored. Just felt nice, mostly.

peppy: Oh, didn't realise that was the intention. I'm not sure then, since the setting is "confine mouse cursor to *window*".. is there even a window when running fullscreen?

bdach: Hiding is also a viable way, sure. I think I slightly prefer disabling since it gives the user indication as to what's happening, but I'm not going to defend it much.

bdach: I've reverted the leasing logic and made the dropdown hidden in full-screen, but I did keep it in the input section since I think it's still more apt there, considering the other options in that section (i.e. absolute input mapping toggle, among others).

peppy:
I do think disabling will be better if we can show a tooltip or message below it explaining why it is disabled, but not sure about using leasing to do so, rather than just setting disabled directly.

bdach:
Looking at this again and testing some more, dropdowns don't really have good support for being disabled (they look as if they're interactable normally, and you can actually provoke a crash by using keyboard controls to select a value on a dropdown with disabled `Current`). So this current iteration might be the best stop-gap solution for now.

Let me know if you'd like me to pursue framework-side changes to dropdowns such that they can support disabled `Current` better.

peppy (approved):
I think this is okay for now. Definitely not the final UX we want.

PR #11107, opened 2020-12-07
Nerf PP awarded for AR11 on short maps in the osu! ruleset

Xexxar:
Hello All. This is part of a series of small changes I am suggesting to be implemented to keep the current competitive scene alive until a more robust solution is introduced, or until I get fed up enough to start attempting to work on DiffCalc. The goal for these changes is to create a series of easily implementable changes that will help to improve the competitive scene at all levels of play. 

**The Problem**
In PerformanceCalc, there exists a scaling buff for all maps with approach rate (AR) exceeding 10.33. As there has been an increase in top level play regarding AR11, I believe it is appropriate to re-examine this buff and adjust its usage. For the sake of keeping the differences easier to quantify, I'll just be referring to the amount of points that will be rewarded with AR11, but know all of this applies in a linearly decreasing fashion for AR's 11 through 10.33.

Currently, the buff is simply a 20% buff to speedPP and aimPP. The reason I am addressing this as problematic is that a large portion of the skill involved with reading AR11 is the length and complexity of the song. If we look at top scores on AR11 maps, they are almost exclusively on short maps. 

**The Solution**
My belief is that AR11 is very similar to FL, as it is a skill that is significantly more difficulty the longer a map is. Hence, I have introduced a buff and a nerf for AR11 maps. The details of the code can be viewed in the changes, but in summary, the change simply makes the AR buff scale from 0% at 0 objects, to 26.66% at 1000 objects for AR11, thus putting a 20% buff on 750 objects. These numbers are based on my own personal observations of map difficulties and length in several cases, as well as communicating with several members of the community. I also decided to buff higher than 750 objects as there are very few AR11 plays that have that many objects, but this can theoretically be reverted.

**Top Player Results**
Whitecat: http://puu.sh/GUW1A/ed6c31c828.txt
badeu: http://puu.sh/GUW3l/a0772ef216.txt
idke: http://puu.sh/GUW42/c02beaf4e4.txt
chocomint: http://puu.sh/GUW1L/846fbb2118.txt
fgsky: http://puu.sh/GUW4j/6dc9896053.txt

These are sort of hard to read as puush didnt like the unicode, but you can still figure out what it showing... albeit its not ideal. Run them yourselves if you want to see it in the proper format.

For those who are wondering, I made the change to the lengthBonus in the speedPP calculation to match the format of aimPP. If its preferred it can be refactored in aimPP to be the format of speedPP.

### DIFF
--- a/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
+++ b/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
@@ -101,9 +101,9 @@ namespace osu.Game.Rulesets.Osu.Difficulty
 
             double approachRateFactor = 0.0;
             if (Attributes.ApproachRate > 10.33)
-              approachRateFactor += 0.4 * (Attributes.ApproachRate - 10.33);
+                approachRateFactor += 0.4 * (Attributes.ApproachRate - 10.33);
             else if (Attributes.ApproachRate < 8.0)
-              approachRateFactor += 0.1 * (8.0 - Attributes.ApproachRate);
+                approachRateFactor += 0.1 * (8.0 - Attributes.ApproachRate);
 
             aimValue *= 1.0 + Math.Min(approachRateFactor, approachRateFactor * (totalHits / 1000.0));
 
--- a/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
+++ b/osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
@@ -99,16 +99,13 @@ namespace osu.Game.Rulesets.Osu.Difficulty
             if (Attributes.MaxCombo > 0)
                 aimValue *= Math.Min(Math.Pow(scoreMaxCombo, 0.8) / Math.Pow(Attributes.MaxCombo, 0.8), 1.0);
 
-            double approachRateFactor = 1.0;
-
+            double approachRateFactor = 0.0;
             if (Attributes.ApproachRate > 10.33)
-                approachRateFactor += 0.3 * (Attributes.ApproachRate - 10.33);
+              approachRateFactor += 0.4 * (Attributes.ApproachRate - 10.33);
             else if (Attributes.ApproachRate < 8.0)
-            {
-                approachRateFactor += 0.01 * (8.0 - Attributes.ApproachRate);
-            }
+              approachRateFactor += 0.1 * (8.0 - Attributes.ApproachRate);
 
-            aimValue *= approachRateFactor;
+            aimValue *= 1.0 + Math.Min(approachRateFactor, approachRateFactor * (totalHits / 1000.0));
 
             // We want to give more reward for lower AR when it comes to aim and HD. This nerfs high AR and buffs lower AR.
             if (mods.Any(h => h is OsuModHidden))
@@ -137,8 +134,9 @@ namespace osu.Game.Rulesets.Osu.Difficulty
             double speedValue = Math.Pow(5.0 * Math.Max(1.0, Attributes.SpeedStrain / 0.0675) - 4.0, 3.0) / 100000.0;
 
             // Longer maps are worth more
-            speedValue *= 0.95 + 0.4 * Math.Min(1.0, totalHits / 2000.0) +
-                          (totalHits > 2000 ? Math.Log10(totalHits / 2000.0) * 0.5 : 0.0);
+            double lengthBonus = 0.95 + 0.4 * Math.Min(1.0, totalHits / 2000.0) +
+                                 (totalHits > 2000 ? Math.Log10(totalHits / 2000.0) * 0.5 : 0.0);
+            speedValue *= lengthBonus;
 
             // Penalize misses exponentially. This mainly fixes tag4 maps and the likes until a per-hitobject solution is available
             speedValue *= Math.Pow(0.97, countMiss);
@@ -147,11 +145,11 @@ namespace osu.Game.Rulesets.Osu.Difficulty
             if (Attributes.MaxCombo > 0)
                 speedValue *= Math.Min(Math.Pow(scoreMaxCombo, 0.8) / Math.Pow(Attributes.MaxCombo, 0.8), 1.0);
 
-            double approachRateFactor = 1.0;
+            double approachRateFactor = 0.0;
             if (Attributes.ApproachRate > 10.33)
-                approachRateFactor += 0.3 * (Attributes.ApproachRate - 10.33);
+                approachRateFactor += 0.4 * (Attributes.ApproachRate - 10.33);
 
-            speedValue *= approachRateFactor;
+            speedValue *= 1.0 + Math.Min(approachRateFactor, approachRateFactor * (totalHits / 1000.0));
 
             if (mods.Any(m => m is OsuModHidden))
                 speedValue *= 1.0 + 0.04 * (12.0 - Attributes.ApproachRate);


bdach:
Without delving into details of what this aims to achieve (which I still think could use elaboration - I really don't like looking at any of these sorts of changes without seeing a before/after graph or other visual representation of what the functions involved in the calculation look like), I don't know if this development model of "many small adjustments" is sustainable time-wise.

As far as I know each change in diffcalc or ppcalc requires [porting to stable](https://osu.ppy.sh/home/changelog/cuttingedge/20200916), and then rolling out recalculations across the board. I'm not sure whether a time commitment necessary to do things this way can be made.

Xexxar:
My main purpose with splitting the changes up is to make them easier to read, however I certainly did not want them to be implemented until all are completed in aggregate. As far as visual graphs for these, I can provide some rudimentary ones.

From AR10.33 to AR11, the current AR Curve is: (where a is our AR)
https://www.wolframalpha.com/input/?i=max%280%2C+.3+*+%28a-10.33%29%29%2C+a+from+10.1+to+11

The Proposed amendment is: (where a is our AR)
https://www.wolframalpha.com/input/?i=max%280%2C+.4+*+%28a-10.33%29%29%2C+a+from+10.1+to+11

With the object count based nerf shown here: (where b is our buff, which is .26 for AR11)
https://www.wolframalpha.com/input/?i=1+%2B+min%28b%2C+b+*+%28x+%2F+1000%29%29%2C+x+from+0+to+1250

If these aren't enough, I'm sure a desmos graph can be set up.

peppy:
I can't speak on the merit of this change, but for changes like this (to the performance part only) there is no stable code changes required. Also to answer @bdach's concerns, we definitely welcome small changes like this. Infra is setup to deploy such changes on a regular basis, as this was the intended path forward (to keep PP updated with changes more often than we are currently used to).

smoogipoo:
Has discussion about this taken place anywhere public? If not I'll let this one simmer a bit.

My initial thought is that this buff should be removed altogether. Personally, with a 144Hz screen, at least ARs up to 10.6 are pretty simple in that anyone can comfortably adjust themselves to that AR without feeling like the game is running away from them. Above that it's harder to adjust to but I'd say it's feasible (definitely up to 10.8), with the big problem being to then find a map that is AR11-able without exceeding BPM limits. 

In-fact most players generally play better at higher ARs - their timing tightens up and density is reduced resulting in more to-the-point aim. Here's a test: play an AR9 map then add HR (AR10), is your UR lower or higher after adding HR? For me it's _always_ lower, even though there's nothing preventing it from being as low.
Conversely, if you're one of the players that consistently plays high AR, _lower_ AR is going to be much harder for you. From what I've seen, AR8 maps (esp +HD) in tournament pools are actually the most challenging for most players, as AR>9 is generally standard.

Really, the big factor in all of this is: does the player have a >60Hz monitor. 10.3+ might be hard/impossible on a 60Hz monitor, but 144Hz greatly simplifies it and 240Hz probably even makes AR11 simple as cake.

Basically, I'm not sure if you can qualify the skill of players or difficulty based on AR.

riesha:
> Really, the limiting factor in all of this is: does the player have a >60Hz monitor. 10.3+ might be hard/impossible on a 60Hz monitor, but 144Hz greatly simplifies it and 240Hz probably even makes AR11 simple as cake.

Some of the best ar11 sightreaders play on 60hz without feeling handicapped at all... 144hz makes it relatively easier since it makes every single game feel better in general + most players start playing more ar11 when they get one anyways so naturally they improve and 240hz and above doesn't come close to making it easy as cake any ar11 player will tell you that

smoogipoo:
Sure there's going to be players able to do AR11 at 60Hz without feeling handicapped, but that wasn't the point of my post. My point was: does it make it easier solely by having a 144Hz+ monitor? As someone who can sightread AR10.6 on 144Hz as if it feels like AR10 on 60Hz without _any practice whatsoever_ (that's the important part), I can say that's definitely the case.  
The difficulty increase, imo, comes solely from the increase in BPM.

Kastellan:
> In-fact most players generally play better at higher ARs - their timing tightens up and density is reduced resulting in more to-the-point aim. Here's a test: play an AR9 map then add HR (AR10), is your UR lower or higher after adding HR? For me it's _always_ lower, even though there's nothing preventing it from being as low.
> Conversely, if you're one of the players that consistently plays high AR, _lower_ AR is going to be much harder for you. From what I've seen, AR8 maps (esp +HD) in tournament pools are actually the most challenging for most players, as AR>9 is generally standard.

It's true that most top players are the most comfortable at AR between 9.5 and 10.5. However just because you get better UR on AR10 than AR9 (since AR10 is more comfortable) doesn't mean that this trend continues as the AR gets higher and higher. Eventually you will reach a point where you aren't skilled enough to read the notes in time and the game gets much harder.

> Really, the limiting factor in all of this is: does the player have a >60Hz monitor. 10.3+ might be hard/impossible on a 60Hz monitor, but 144Hz greatly simplifies it and 240Hz probably even makes AR11 simple as cake.

Speaking as someone who plays with a 240hz and has played a fair amount of ar11 (both before and after getting a 240hz), I think I can say this assumption is wrong. 240hz makes everything feel smooth and overall the game feels nicer, but it doesn't specifically make ar11 easier. Plus there are plenty of prolific AR11 players that use 60hz.
It would be more accurate to say that the limiting factor is whether or not the player can adjust their gamma settings, but even that's a stretch.

cl8n:
the argument abt whether high AR is *really* an element of difficulty doesn't seem important to me cuz you can look at players' performances and see that they're having an easier time with shorter maps on high AR. this change is a good response to how the meta for these plays has changed over the years and I don't think more justification is necessary—it's pp's job to give a rating that *feels* appropriate right now. sometime in the future this will probably need to be tweaked again, but I think that's fine and how an adaptive system is supposed to work

also @smoogipoo your monitor example just seems wrong or at least conflicting with everyone else's experience. a frame time difference of ~10ms doesn't make you able to read 100ms faster... my guess is that your setup's total perceived latency using the 144hz monitor happened to be most comfortable for you compared to the 60hz one. and like kastellan said, if we're gonna go down the "hardware can make it easier" route, gamma changes have a much larger effect

Xexxar:
As far as the statements regarding bdach and ppy, if I am interpreting correctly then changes to PP/ppCalc are quickly implemented and are fine to be done in isolation, but changes to SR/diffCalc should only be proposed when complete and require a more extensive deployment process.

If this is the case then my proposal style should conform with above.

With all that in mind, I do have additional plans for things that I'd like to propose to amend the ppCalc, but I am wondering what implementation would look like in the event that this proposal was approved. Is there an ideal frequency of changes you're looking at @peppy? Would this simply be merged to the lazer client, but not rolled out onto stable until a later agreed upon date being deployed with other changes? I only ask as you said you have the pipelines for quick rollouts for changes like these, but I am wondering what that entails in more detail.



As far as other comments regarding whether or not AR11 is worthy of the buff etc, I think its clear that the existing buff is not really properly representing the difficulty of the maps that people are farming. For that reason, either removing the buff altogether or introducing a scaling component with length is probably an acceptable proposal. But at the end of the day, I agree with cl8n in that the change here more accurately represents the competitive scene at the moment. While removing it outright can produce some good results, it also produces some negative results that are probably better avoided by using a scaling buff.

peppy:
- Merge changes into lazer
- Backport diffcalc changes to stable (annoying)
- Update server-side diffcalc (1-2 days)
- Update server-side performance (background process 1-2 days, plus on next score submission for each user)

Rollout of just performance-side changes bypasses all but the last step. It still have a computational overhead but the *deployment* process is a lot simpler because there's fewer moving parts.

smoogipoo (approved):
Let's go with this for now then

Xexxar:
I see. Well, I certainly think this change is a step in the right direction, albeit some of my other planned rebalancing plans related to accuracy potentially warranted an outright removal of the AR11 buff in favor of an OD11 buff. Certainly this change by itself is better than the existing live values, and I think this small change will disrupt a minimum amount of people almost exclusively at the top ranks, so it might be a good change to prototype with this faster deployment method. 

My only concern is the frequency of change. Obviously I don't think you all want the system to change more than one time per month, correct? If we were to move this towards implementation, would it be best to wait until all changes are ready, or do you not see this as an issue.

peppy:
I'm fine with more often than once per month. Let's see how things go. We actually have an (unused) system which let us show changes like this on the performance graphs so people are aware of why their graph suddenly jumped (i believe it isn't implemented osu-web end, but that can be done if people start to complain).

FlashoFoSho2266:
I think the difficulty of high AR is a bit excessively undermined by saying it's mainly memorization. I don't agree and believe that the larger portion of high AR players practiced for hundreds of hours to hone their reading and reaction times, and that memorization was instead a minor factor in their success.
    I believe this proposal still has potential though with more tweaking and balancing. The nerf should be focused on shorter maps more so, in the 0-350(400?) object range. This would hit all maps that are shorter than tv-size and many tv-sizes that have simple rhythms that consist mainly of PP jumps while avoiding nerfing tv-size maps with complex rhythm. Ultimately putting the original 20% scaled buff at roughly that 350-400 object range. 
    Lastly, objects above that benchmark deserve a scaled buff as well, starting from 400 objects at 20% and ending with 1000 objects at a max buff of around 30%. Certainly this all needs more testing and feedback from experienced players along with a feasible formula, the general idea is the main thing that I wanted to convey.

PR #11110, opened 2020-12-07
Implement in-place update of catch DrawableHitObject

ekrctb:
In previous PR #11011, my pooling implementation was incomplete in a way there are still allocations when a DHO is reused.
In this PR, I implemented the remaining code so no repeated `Drawable` allocations are made when a DHO is reused.

There are "default" and "legacy" pieces.

For default piece `PulpFormation`, `Pulp`s are lazily allocated but never deallocated for a `DrawableFruit`. Previously allocated `Pulp`s are reused.

For legacy pieces, it sets `Texture`.
An issue was skin fallback logic. The default skin is used when the legacy skin doesn't have a texture, but what if the skin has texture for one fruit type (`FruitVisualRepresentation`) and not for other types?
I made it so only one type (arbitrary `fruit-pear`) is checked for the fallback logic.
The reasoning is that a skin with only some types of fruit textures is very rare (I don't know any instance) because that will be visually inconsistent. It is a fair assumption that a skin provides either all four fruit types or none.

For both, I made some refactoring to make `CatchHitObjectPiece` and `LegacyCatchHitObjectPiece` base classes to remove code duplication.
For both, a hyper dash border piece is always allocated and made it show/hide by setting `Alpha`.

I removed `DrawableBanana : DrawableFruit` inheritance. A banana cannot be hyper, and is not logically similar enough.
Code reuse is done through new `CatchHitObjectPiece` class.
Also, it is common for a skin to provide four types of fruits, but not the banana. Thus, I needed to distinguish four types of fruits and the banana.


osu.Game.Rulesets.Catch/Skinning/LegacyCatchHitObjectPiece.cs
```diff
@@ -0,0 +1,100 @@
...
+        [Resolved]
+        private ISkinSource skin { get; set; }
+
+        protected ISkinSource Skin => skin;
+
+        [Resolved(canBeNull: true)]
+        private DrawableHitObject drawableHitObject { get; set; }
+
+        [CanBeNull]
+        protected DrawablePalpableCatchHitObject DrawableHitObject => (DrawablePalpableCatchHitObject)drawableHitObject;
```
bdach, line 38: What's the reason for this null-permissiveness? Removing the `CanBeNull` annotations doesn't seem to fail any tests, and I'd surely like to think any of these pieces should really rely on a DHO existing above them (osu! DHO pieces make these sorts of assumptions already).

osu.Game.Rulesets.Catch/Skinning/LegacyCatchHitObjectPiece.cs
```diff
@@ -0,0 +1,100 @@
...
+        public readonly Bindable<bool> HyperDash = new Bindable<bool>();
+
+        private readonly Sprite colouredSprite;
+        private readonly Sprite overlaySprite;
+        private readonly Sprite hyperSprite;
+
+        [Resolved]
+        private ISkinSource skin { get; set; }
+
+        protected ISkinSource Skin => skin;
```
bdach, line 32: can be merged
```suggestion
        [Resolved]
        protected ISkinSource skin { get; private set; }
```

osu.Game.Rulesets.Catch/Skinning/Default/PulpFormation.cs
```diff
@@ -34,11 +33,23 @@ protected PulpFormation()
...
+        {
+            for (; numPulps > 0; numPulps--)
+                InternalChildren[numPulps - 1].Alpha = 0;
+        }
+
+        protected void Add(Vector2 position, Vector2 size)
         {
-            DrawableCatchHitObject drawableCatchObject = (DrawableCatchHitObject)drawableObject;
-            AccentColour.BindTo(drawableCatchObject.AccentColour);
+            if (numPulps == InternalChildren.Count)
```
bdach, line 44: Not sure if it's just me but I got thoroughly confused by this check and how it can ever be false. I *think* the intent would be conveyed better if `numPulps` was renamed to `pulpsInUse`.

osu.Game.Rulesets.Catch/Skinning/Default/PulpFormation.cs
```diff
@@ -34,11 +33,23 @@ protected PulpFormation()
             distance * MathF.Sin(angle * MathF.PI / 180),
             distance * MathF.Cos(angle * MathF.PI / 180));
 
-        [BackgroundDependencyLoader]
-        private void load(DrawableHitObject drawableObject)
+        protected void Clear()
+        {
+            for (; numPulps > 0; numPulps--)
+                InternalChildren[numPulps - 1].Alpha = 0;
```
bdach, line 39: This is also way too cleverly written for my liking. I assume the intention here is to set `Alpha` and roll back `numPulps` to 0 at the same time, but I'd honestly just prefer a plain `for` loop and value reset to 0 at the end.

osu.Game.Rulesets.Catch/Skinning/Default/PulpFormation.cs
```diff
@@ -34,11 +33,23 @@ protected PulpFormation()
...
+        }
+
+        protected void Add(Vector2 position, Vector2 size)
```
bdach, line 42: Would rename to `AddPulp` to convey intention better (looks really weird at call sites, like the generic `Drawable.Add()` but without a drawable).

osu.Game.Rulesets.Catch/Skinning/LegacyCatchHitObjectPiece.cs
ekrctb: The nullability is not necessary for now.
The plain is don't use DHO for objects on the plate (see `Catcher.createCaughtObject`).
And the piece classes will be used in both DHOs and caught objects like `CaughtFruit { FruitPiece }` and `DrawableFruit { FruitPiece }`.
For a caught object, this `DrawableHitObject` will be null.

peppy (dismissed):

osu.Game.Rulesets.Catch/Skinning/Default/CatchHitObjectPiece.cs
```diff
@@ -0,0 +1,55 @@
...
+    {
+        public readonly Bindable<Color4> AccentColour = new Bindable<Color4>();
+        public readonly Bindable<bool> HyperDash = new Bindable<bool>();
+
+        [Resolved(canBeNull: true)]
+        [CanBeNull]
+        protected DrawableHitObject DrawableHitObject { get; private set; }
+
+        [CanBeNull]
+        protected BorderPiece BorderPiece;
```
smoogipoo, line 25: Imo this is pretty weird and we've avoided doing this elsewhere in the game, but I'll overlook it here. At least it's properly annotated.

@peppy are you fine with this?

osu.Game.Rulesets.Catch/Skinning/Legacy/LegacyFruitPiece.cs
```diff
@@ -1,83 +1,39 @@
...
     {
-        private readonly string lookupName;
+        public readonly Bindable<FruitVisualRepresentation> VisualRepresentation = new Bindable<FruitVisualRepresentation>();
 
-        private readonly Bindable<Color4> accentColour = new Bindable<Color4>();
-        private readonly Bindable<bool> hyperDash = new Bindable<bool>();
-        private Sprite colouredSprite;
-
-        public LegacyFruitPiece(string lookupName)
+        private readonly string[] lookupNames =
```
smoogipoo, line 14:
```suggestion
        private static readonly string[] lookupNames =
```
However, is there a reason you did it this way instead of a switch?

smoogipoo (changes requested):

osu.Game.Rulesets.Catch/Skinning/Default/CatchHitObjectPiece.cs
peppy: As in having this at a base class even though it's never instantiated here? I guess we'd usually have a `CreateBorderPiece` method, could probably add that in?

osu.Game.Rulesets.Catch/Skinning/Legacy/LegacyFruitPiece.cs
ekrctb: I'll make it a switch

osu.Game.Rulesets.Catch/Skinning/Default/CatchHitObjectPiece.cs
ekrctb: I agree it is kind of weird.
But I couldn't do `CreateBorderPiece` because I wanted to make set all `InternalChildren` on the derived class' constructor.

smoogipoo: I imagined it was something like that... Your recent changes improve it enough for me, so I'll resolve this.

smoogipoo (approved):

PR #11111, opened 2020-12-07
Fix possible null reference in MouseSettings

ekrctb:
When bindable change event is invoked before variable is initialized, it will be a null reference.
Found in CI <https://ci.appveyor.com/project/peppy/osu/builds/36690846/tests>.

### DIFF
--- a/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
+++ b/osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
@@ -68,7 +68,7 @@ namespace osu.Game.Overlays.Settings.Sections.Input
             };
 
             windowMode = config.GetBindable<WindowMode>(FrameworkSetting.WindowMode);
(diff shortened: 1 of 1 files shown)


osu.Game/Overlays/Settings/Sections/Input/MouseSettings.cs
```diff
@@ -70,6 +67,9 @@ private void load(OsuConfigManager osuConfig, FrameworkConfigManager config)
                 },
             };
 
+            windowMode = config.GetBindable<WindowMode>(FrameworkSetting.WindowMode);
+            windowMode.BindValueChanged(mode => confineMouseModeSetting.Alpha = mode.NewValue == WindowMode.Fullscreen ? 0 : 1);
```
peppy, line 71: while we're here, this probably requires a `true`, else the initial state may be incorrect.

```suggestion
            windowMode.BindValueChanged(mode => confineMouseModeSetting.Alpha = mode.NewValue == WindowMode.Fullscreen ? 0 : 1, true);
```

peppy (changes requested):

peppy (approved):

PR #11114, opened 2020-12-07
Pool catcher trail sprite

ekrctb:
20 catcher trails are created per second when the catcher is dashing. I guess pooling it makes sense.
Unfortunately I had to wrap `Sprite` to a composite drawable to make it `PoolableDrawable`.


smoogipoo (approved):
lgtm

PR #11115, opened 2020-12-07
Add the ability to import into ArchiveModelManagers from a stream

peppy:
Addresses issued that arose in the android import pr (see https://github.com/ppy/osu/pull/11099/files#r536876462).

@Game4all please test this and make sure it works for your use case.


Game4all (dismissed):
Cherry-picked this on top of the Android import changes in a temporary branch. Works well 👍.

osu.Game/Database/ImportTask.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        /// <summary>
+        /// The path to the file (or filename in the case a stream is provided).
+        /// </summary>
+        public string Path { get; }
+
+        /// <summary>
+        /// An optional stream which provides the file content.
+        /// </summary>
+        public Stream Stream { get; }
```
bdach, line 24: Maybe worth chucking a `[CanBeNull]` onto this or even doing `#nullable enable` and making this a `Stream?` for additional compiler-level safety?

I would do this myself but not 100% sure what the preferred method would be.

osu.Game/Database/ArchiveModelManager.cs
```diff
@@ -183,16 +190,17 @@ protected async Task<IEnumerable<TModel>> Import(ProgressNotification notificati
 
         /// <summary>
         /// Import one <typeparamref name="TModel"/> from the filesystem and delete the file on success.
+        /// Note that this bypasses the UI flow and should only be used for special cases or testing.
         /// </summary>
-        /// <param name="path">The archive location on disk.</param>
+        /// <param name="task">The archive location on disk.</param>
```
bdach, line 195: xmldoc is slightly stale with this change, although at this point might as well just cull it maybe as it's internal API now - but if not, how about
```suggestion
        /// <param name="task">The <see cref="ImportTask"/> containing data about the <typeparamref name="TModel"/> to import.</param>
```

bdach (commented):
No issues here with the core of the change either, only a few small notes:

smoogipoo (approved):

PR #11116, opened 2020-12-07
Added a Spectate button in the UserProfileOverlay

mohniaz-com:
I found that a spectate button on the UserProfileOverlay would be very useful because it can allow players to easily spectate their friends. I created a new SpectateSection class that has a PurpleTriangleButton class (the same button used in the CurrentlyPlayingDisplay). The PurpleTriangleButton will switch the game to the Spectator view while passing the User object from the UserProfileOverlay.

P.S. This is my first pull request to a public repository and so what you see in this pull request might be awful.

### DIFF
--- a/osu.Game/Overlays/Profile/Sections/SpectateSection.cs
+++ b/osu.Game/Overlays/Profile/Sections/SpectateSection.cs
@@ -1,0 +1,35 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Allocation;
+using osu.Framework.Graphics;
+using osu.Framework.Screens;
+using osu.Game.Screens.Multi.Match.Components;
+using osu.Game.Screens.Play;
+
+namespace osu.Game.Overlays.Profile.Sections
+{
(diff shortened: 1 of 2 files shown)


smoogipoo:
This looks very bad. I'm sure @arflyte can come up with a better way to spectate friends than opening their profile and having a massive horizontal button in a section.

For reference, this PR adds the following button to the user profile page:
image

The current dashboard overlay lists the currently playing users in a different section to friends:
image

What we want is probably two things:
1. A better UI/UX for spectating users other than a button underneath their panel.
2. Filtering in the dashboard overlay. As far as I'm aware there is no design for this (@arflyte?)

peppy:
Also this should not even be considered until we can query whether a player is actually playing, and not show the button if that's not the case.

smoogipoo:
Isn't that querying what we already have? Also, I think you'd want to spectate players that aren't playing too (in preparation for them to start playing).

peppy:
Currently we are sending every user playing so I guess that can be used, but it's not something which will scale (as per inline comments it's a very temporary setup).

And sure, but it should still be based on the user being online. The button shouldn't just be there on every user regardless of online state, that's really weird UX.

smoogipoo:
Let's close this for now since it's not the direction either of us expect.

PR #11118, opened 2020-12-07
Add the ability to switch tournaments to the SetupScreen

MiraiSubject:
This adds an extra entry to the setup screen that allows the user to switch the tournaments in the client instead of using the `tournament.ini`. 

Currently requires a restart to apply the changes.

image


peppy:
If it requires a restart, please trigger the restart.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -13,12 +15,15 @@ public class TournamentStorage : MigratableStorage
     {
         private const string default_tournament = "default";
         private readonly Storage storage;
+        private readonly Storage allTournaments;
```
bdach, line 18: I'd make this a get-only delegating property:

```csharp
private Storage allTournamentsStorage => UnderlyingStorage;
```

The aliasing is nice for readability but the assignment is unnecessarily troublesome.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -29,9 +34,22 @@ public TournamentStorage(Storage storage)
             else
                 Migrate(UnderlyingStorage.GetStorageForDirectory(default_tournament));
 
+            CurrentTournament = new Bindable<string>(storageConfig.Get<string>(StorageConfig.CurrentTournament));
```
bdach, line 37: This pattern is super confusing to me and risky (will be wrong if a restart is ever not needed). Maybe just do

```suggestion
            CurrentTournament = storageConfig.GetBindable<string>(StorageConfig.CurrentTournament);
```

Pretty sure that'd spare you the `Set` call in `updateTournament`, too.

osu.Game.Tournament/IO/TournamentStorage.cs
```diff
@@ -29,9 +34,22 @@ public TournamentStorage(Storage storage)
...
             Logger.Log("Using tournament storage: " + GetFullPath(string.Empty));
+
+            CurrentTournament.BindValueChanged(updateTournament, false);
```
bdach, line 40: `false` is the default so it doesn't really need to be specified.

osu.Game.Tournament/Screens/SetupScreen.cs
```diff
@@ -40,6 +42,9 @@ public class SetupScreen : TournamentScreen, IProvideVideo
         [Resolved]
         private RulesetStore rulesets { get; set; }
 
+        [Resolved]
+        private Storage storage { get; set; }
```
bdach, line 46: I'd probably prefer this to be done as such:

```csharp
private TournamentStorage storage;

// in BDL, before reload()
private void load(FrameworkConfigManager frameworkConfig, Storage storage)
{
    // ...
    this.storage = (TournamentStorage)storage;
    reload();
}

// can then remove the access null-checks
```

I think that's much clearer when it comes to the intent (the incoming storage needs to be a tournament storage, and is guaranteed non-null if the hard cast succeeds).

bdach (commented):
Generally this will probably require UX considerations as to what happens if you choose another tournament. Getting booted from the game outright probably won't be the best option.

MiraiSubject:
> Generally this will probably require UX considerations as to what happens if you choose another tournament. Getting booted from the game outright probably won't be the best option.

Yeah I agree. I'll be looking into something that will at worst reload the scenes to refresh the videos. Reading the bracket on switch seems a bit more complicated than I had anticipated. 

For now I have addressed the review comments

osu.Game.Tournament/IO/TournamentStorage.cs
MiraiSubject: Unfortunately doing that displays a different behaviour.

Currently when it's in the constructor as-is `UnderlyingStorage` is points to `tournaments/` because `ChangeTargetStorage` hasn't ran yet. 

Because the storage gets changed in the constructor, the suggested change lists the folders of `tournaments/default` rather than `tournaments`

Before: 

image

Suggested change: 
image

bdach: Oh yikes, I missed the fact it's migrating itself inside the ctor... I guess this is more acceptable then, but I'd probably change all other accesses to `UnderlyingStorage` in the ctor to use `allTournaments` instead. Should hopefully make it clearer what's going on...?

MiraiSubject: Yeah that will likely make it clearer. I will do that then instead

peppy:
At very least this should have a temporary exit procedure (even if it doesn't automatically restart the game). You can probably check how `SquirrelUpdateManager` does things and copy where applicable.

MiraiSubject:
I got the panel so far changed so it now also contains a close button and it uses `GracefullyExit()` that's being used in the UpdateManager. I moved that to `OsuGameBase` so the tournament game can also use it. 

As for restarting, I'm not quite sure how to go about that.

I tried adding the `BindValueChanged` of the `Current` bindable to the component creation and add the button once it changes. But I realised it already counts as a change when it gets populated initially. A solution would be to check if the old value is the same as the new value and then hide accordingly, but there might be a better way that I'm not aware of.

image

peppy (dismissed):
Should do for now.

peppy (approved):

PR #11119, opened 2020-12-07
Refactor multiplayer filter control 

smoogipoo:
Just a little bit of refactoring to make room for realtime multiplayer. The only visual change is the change in brightness of the control to make it look better - currently the textbox blends into the background and looks ugly.

A few things to point out:
1. I'm using the prefix "Timeshift" for now, and will do a pass on all components later on. Eventually "timeshift" will be renamed to "playlists" and be split out accordingly.
2. Eventually there'll be a `MultiplayerFilterControl`, with the only implementation being to set `criteria.Category = "realtime"`.
3. As per the above, the request takes a category string rather than an enum now.
4. The `LoungeSubScreen` implementation will have an abstract method to create this control.
5. As a result of this refactoring, a few miscellaneous controls that were only used for this and older overlays have been removed.


bdach (approved):
can't see any issues with this :+1:

peppy (approved):

PR #11123, opened 2020-12-08
Move catcher state changing logic to OnNewResult and OnRevertResult

ekrctb:
`Catcher.AttemptCatch` was not pure, as it is changing the catcher state internally.
I made `Catcher.CanCatch` a pure method and moved all state-changing logic to `OnNewResult`.
Then, I implemented `OnRevertResult` to restore some catcher states.

- Fix catcher state is not restored when a replay is rewound.
- Fix hit explosion and caught objects are duplicated when a replay is rewound.


smoogipoo (dismissed):
Good stuff

smoogipoo (approved):

PR #11126, opened 2020-12-08
Clear all transforms of catcher trail sprite before returned to pool

### LINKED ISSUE #11124: Catch hyperdash sometimes flashes in an incorrect position
https://drive.google.com/file/d/10V786jNKgq-jfJLhl1vRgf_f15gAyrre/view?usp=sharing It also flashes at the previous hyperdash location if you move the catcher around (hyperdash -> turn off hyperdash -> move catcher -> hyperdash).

Looks to be something like reusing a sprite without setting an updated position, from the recent pooling changes.

ekrctb:
Fix #11124.


smoogipoo (approved):
Oof, this is going to be an annoying one to keep in the back of the mind. Maybe something to consider at `PoolableDrawable` level in the future...

bdach:
Yeah this is like the third time this pops up. Thought doing that at DHO level was going to be enough but clearly not. Could probably use a flag framework-side or something.

PR #11127, opened 2020-12-08
Add client-side models and interfaces required for multiplayer

peppy:
These are shared models and interfaces that are required by the SignalR server (and future client) implementations to allow communication.

Open to discussion on the namespace (and probably force push to synchronise with the client implementation).


osu.Game/Online/RealtimeMultiplayer/IMultiplayerClient.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <param name="state">The new state of the user.</param>
+        Task UserStateChanged(long userId, MultiplayerUserState state);
+
+        /// <summary>
+        /// Signals that a match is to be started. This will *only* be sent to clients which are to begin loading at this point.
+        /// </summary>
+        Task LoadRequested();
+
+        /// <summary>
+        /// Signals that a match has started. All user in the <see cref="MultiplayerUserState.Loaded"/> state should begin gameplay as soon as possible.
```
bdach, line 56:
```suggestion
        /// Signals that a match has started. All users in the <see cref="MultiplayerUserState.Loaded"/> state should begin gameplay as soon as possible.
```

osu.Game/Online/RealtimeMultiplayer/IMultiplayerServer.cs
```diff
@@ -0,0 +1,48 @@
...
+        /// </summary>
+        /// <param name="userId">The new user which is to become host.</param>
+        /// <exception cref="NotHostException">A user other than the current host is attempting to transfer host.</exception>
+        Task TransferHost(long userId);
+
+        /// <summary>
+        /// As the host, update the settings of the currently joined room.
+        /// </summary>
+        /// <param name="settings">The new settings to apply.</param>
+        Task ChangeSettings(MultiplayerRoomSettings settings);
```
bdach, line 33: I imagine this could also potentially throw `NotHostException`.

osu.Game/Online/RealtimeMultiplayer/InvalidStateChange.cs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Game.Online.RealtimeMultiplayer
+{
+    public class InvalidStateChange : Exception
```
bdach, line 8: Exception class naming in this PR is sort of half and half - there are:

* `InvalidStateChange`
* `InvalidStateException`
* `NotHostException`
* `NotJoinedRoomException`
* `UserAlreadyInMultiplayerRoom`

Would probably unify by doing `InvalidStateChange` -> `InvalidStateChangeException` and `UserAlreadyInMultiplayerRoom` -> `AlreadyInRoomException`.

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomSettings.cs
```diff
@@ -0,0 +1,30 @@
...
+        public int? RulesetID { get; set; }
+
+        [NotNull]
+        public IEnumerable<APIMod> Mods { get; set; } = Enumerable.Empty<APIMod>();
+
+        public bool Equals(MultiplayerRoomSettings other) => BeatmapID == other.BeatmapID && Mods.SequenceEqual(other.Mods) && RulesetID == other.RulesetID;
+
+        public override string ToString() => $"Beatmap:{BeatmapID} Mods:{string.Join(',', Mods)} Ruleset:{RulesetID}";
+
+        public static MultiplayerRoomSettings Empty() => new MultiplayerRoomSettings();
```
bdach, line 28: A bit curious why this is a thing. I'd probably either remove this or hide the parameterless ctor by making it `private`, so that there's no ambiguity whether the ctor and this static factory method differ in any capacity.

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomSettings.cs
```diff
@@ -0,0 +1,30 @@
...
+    public class MultiplayerRoomSettings : IEquatable<MultiplayerRoomSettings>
+    {
+        public int? BeatmapID { get; set; }
+
+        public int? RulesetID { get; set; }
+
+        [NotNull]
+        public IEnumerable<APIMod> Mods { get; set; } = Enumerable.Empty<APIMod>();
+
+        public bool Equals(MultiplayerRoomSettings other) => BeatmapID == other.BeatmapID && Mods.SequenceEqual(other.Mods) && RulesetID == other.RulesetID;
```
bdach, line 24: Going to have to be careful about order of `Mods` with this comparer. Maybe the setter of `Mods` should ensure sorting?

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomUser.cs
```diff
@@ -0,0 +1,42 @@
...
+    [Serializable]
+    public class MultiplayerRoomUser : IEquatable<MultiplayerRoomUser>
+    {
+        public readonly long UserID;
+
+        public MultiplayerUserState State { get; set; } = MultiplayerUserState.Idle;
+
+        public User? User { get; set; }
+
+        public MultiplayerRoomUser(in int userId)
```
bdach, line 20: Incoming param is `int` but local field is `long` - not sure if intended.

bdach (commented):
Super nitpicky review but there's not much to go on here. Haven't considered any bigger implications of the API design or read the server-side PR, just read the structures as they are. I'd probably miss stuff - I imagine kinks in this design can be pointed out as progress goes along.

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomSettings.cs
peppy: Hmm... I think it's safe enough for now as worst-case it's just going to trigger an update of settings to the same settings. I copied this from the existing implementation in `SpectatorState` for what that's worth.

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomUser.cs
peppy: `int` is intended - i don't think we're going to hit `long` user IDs for quite a while longer.

osu.Game/Online/RealtimeMultiplayer/MultiplayerRoomSettings.cs
peppy: parameterless needs to be there for serialization i belied. will just remove this since it doesn't add much clarity.

bdach (dismissed):
Throwing this a tentative approve as I can't see any issues with the models myself at this point, and maybe an approval will be useful to get this merged and continue work over at the spectator server repository :+1:

Namespace is fine I think. I guess something like `osu.Game.Multiplayer.Realtime` might also work, but not sure if it's a big deal.

smoogipoo:
Removed `IStatefulMultiplayerClient` - interface proved to be useless with my mp changes.

Also split out interfaces so that my client doesn't have to implement `JoinRoom(long roomId)`. My hierarchy becomes something like the following now:
```
class StatefulMultiplayerClient : IMultiplayerClient, IMultiplayerRoomServer
{
    public Task<MultiplayerRoom> JoinRoom(Room room) // Note Room object, NOT long roomId
    {
        return Room = await connection.InvokeAsync<MultiplayerRoom>(nameof(IMultiplayerLoungeServer.JoinRoom), room.RoomId.Value);
    }
}
```
(as discussed)

peppy:
Seems fine.

smoogipoo (approved):

PR #11129, opened 2020-12-08
Port StanR's dynamic SO pp changes

smoogipoo:
Port of https://github.com/ppy/osu-performance/pull/110

Seems reasonable.


bdach (approved):
looks to be equivalent :+1:

PR #11131, opened 2020-12-08
Pool objects caught by the catcher

ekrctb:
Addressing the second homework mentioned in my previous PR #11011.

Before this PR, when a hit object is caught by the catcher, a DHO was created and placed on the catcher plate. Caught objects are constantly created and destroyed, and the allocation overhead is significant. Pooling caught objects is desired to reduce stuttering during the gameplay.

However, several DHO logics that are unneeded for a static caught object interfere with the desired pooling logic. For example, we had to use `ApplyCustomState` to workaround an issue of transforms cleared by DHO.

Thus, I decided to create a new class `CaughtObject` for a caught object, not a derived class of DHO.
To reduce code duplication, I extracted common properties of `DrawablePalpableCatchHitObject` and `CaughtObject` to `IHasCatchObjectState` interface.

Visual properties such as `AccentColour` are copied from the DHO to `CaughtObject`, not bound.
A known issue is if the skin is changed then caught objects will keep old skin's combo colors.
 The issue is properties such as `AccentColour` are only present in DHO, not in HOs. I'm currently not sure how to solve this issue. We cannot bind to a DHO because the DHO will be reused for another hit object while the `CaughtObject` is still alive.

I refactored some of the object catching logic of the catcher.
<s>I removed `ScaleContainer` from catch DHO to make the implementation simpler.</s>
Edit: I added it back.

### DIFF
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
@@ -73,7 +73,7 @@ namespace osu.Game.Rulesets.Catch.Tests
 
             SetContents(() =>
             {
-                var droppedObjectContainer = new Container
+                var droppedObjectContainer = new Container<CaughtObject>
                 {
                     RelativeSizeAxes = Axes.Both
                 };
@@ -102,7 +102,7 @@ namespace osu.Game.Rulesets.Catch.Tests
 
         private class TestCatcherArea : CatcherArea
         {
-            public TestCatcherArea(Container droppedObjectContainer, BeatmapDifficulty beatmapDifficulty)
+            public TestCatcherArea(Container<CaughtObject> droppedObjectContainer, BeatmapDifficulty beatmapDifficulty)
                 : base(droppedObjectContainer, beatmapDifficulty)
             {
             }
--- a/osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
+++ b/osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
@@ -73,7 +73,10 @@ namespace osu.Game.Rulesets.Catch.Tests
 
             SetContents(() =>
             {
-                var droppedObjectContainer = new Container();
+                var droppedObjectContainer = new Container
+                {
+                    RelativeSizeAxes = Axes.Both
+                };
 
                 return new CatchInputManager(catchRuleset)
                 {
--- a/osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
+++ b/osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
@@ -15,7 +15,7 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
     [Cached(typeof(IHasCatchObjectState))]
     public abstract class CaughtObject : SkinnableDrawable, IHasCatchObjectState
     {
-        public CatchHitObject HitObject { get; private set; }
+        public PalpableCatchHitObject HitObject { get; private set; }
         public Bindable<Color4> AccentColour { get; } = new Bindable<Color4>();
         public Bindable<bool> HyperDash { get; } = new Bindable<bool>();
 
@@ -29,7 +29,6 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
         protected CaughtObject(CatchSkinComponents skinComponent, Func<ISkinComponent, Drawable> defaultImplementation)
             : base(new CatchSkinComponent(skinComponent), defaultImplementation)
         {
-            Anchor = Anchor.TopCentre;
             Origin = Anchor.Centre;
 
             RelativeSizeAxes = Axes.None;
@@ -43,6 +42,17 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
             Rotation = objectState.Rotation;
             AccentColour.Value = objectState.AccentColour.Value;
             HyperDash.Value = objectState.HyperDash.Value;
+        }
+
+        protected override void FreeAfterUse()
+        {
+            ClearTransforms();
+
+            Alpha = 1;
+            LifetimeStart = double.MinValue;
+            LifetimeEnd = double.MaxValue;
+
+            base.FreeAfterUse();
         }
     }
 
--- a/osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
+++ b/osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
@@ -21,6 +21,8 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
         public Bindable<Color4> AccentColour { get; } = new Bindable<Color4>();
         public Bindable<bool> HyperDash { get; } = new Bindable<bool>();
 
+        float IHasCatchObjectState.Scale => Scale.X;
+
         /// <summary>
         /// Whether this hit object should stay on the catcher plate when the object is caught by the catcher.
         /// </summary>
@@ -43,7 +45,7 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
         public virtual void CopyStateFrom(IHasCatchObjectState objectState)
         {
             HitObject = objectState.HitObject;
-            Scale = objectState.Scale;
+            Scale = new Vector2(objectState.Scale);
             Rotation = objectState.Rotation;
             AccentColour.Value = objectState.AccentColour.Value;
             HyperDash.Value = objectState.HyperDash.Value;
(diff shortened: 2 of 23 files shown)


osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            Alpha = 1;
+            LifetimeStart = double.MinValue;
+            LifetimeEnd = double.MaxValue;
+
+            base.FreeAfterUse();
+        }
+    }
+
+    public class CaughtFruit : CaughtObject, IHasFruitState
```
peppy, line 65: Probably should move these to their own files (even for small classes we generally do this as a rule).

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -441,138 +458,123 @@ private void updateState(CatcherAnimationState state)
...
+            return droppedObject;
         }
 
-        private void removeFromPlate(DrawablePalpableCatchHitObject caughtObject, DroppedObjectAnimation animation)
+        private void clearPlate(DroppedObjectAnimation animation)
         {
-            if (!caughtFruitContainer.Remove(caughtObject))
-                throw new InvalidOperationException("Can only drop a caught object on the plate");
+            var caughtObjects = caughtFruitContainer.Children.ToArray();
+            var droppedObjects = caughtObjects.Select(getDroppedObject).ToArray();
```
bdach, line 538: Not sure about this double copy. Is there a reason to not just immediately `.Select(getDroppedObject)`, therefore allocating just one array?

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -441,138 +458,123 @@ private void updateState(CatcherAnimationState state)
...
+        private void removeFromPlate(CaughtObject caughtObject, DroppedObjectAnimation animation)
         {
-            var originalX = d.X * Scale.X;
-            var startTime = Clock.CurrentTime;
+            var droppedObject = getDroppedObject(caughtObject);
 
-            d.Anchor = Anchor.TopLeft;
-            d.Position = caughtFruitContainer.ToSpaceOfOtherDrawable(d.DrawPosition, droppedObjectTarget);
+            if (!caughtFruitContainer.Remove(caughtObject))
+                throw new InvalidOperationException("Can only drop a caught object on the plate");
```
bdach, line 553: Not sure if I'm understanding the intent right, would maybe reword to
```suggestion
                throw new InvalidOperationException("Can only drop objects that were previously caught on the plate");
```

osu.Game.Rulesets.Catch.Tests/TestSceneCatcherArea.cs
```diff
@@ -73,7 +73,10 @@ private void createCatcher(float size)
 
             SetContents(() =>
             {
-                var droppedObjectContainer = new Container();
+                var droppedObjectContainer = new Container
+                {
+                    RelativeSizeAxes = Axes.Both
+                };
```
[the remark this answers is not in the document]
ekrctb: I had a bug of not setting `Anchor` of `CaughtObject` when the object is dropped. The bug was not apparent in this visual test when the container is size 0 (not setting `RelativeSizeAxes`) so I added it for the sake of robustness of the visual test.

osu.Game.Rulesets.Catch/Objects/Drawables/DrawablePalpableHasCatchHitObject.cs
```diff
@@ -5,60 +5,49 @@
...
 using osu.Framework.Graphics;
-using osu.Framework.Graphics.Containers;
 using osuTK;
+using osuTK.Graphics;
 
 namespace osu.Game.Rulesets.Catch.Objects.Drawables
 {
-    public abstract class DrawablePalpableCatchHitObject : DrawableCatchHitObject
+    [Cached(typeof(IHasCatchObjectState))]
+    public abstract class DrawablePalpableHasCatchHitObject : DrawableCatchHitObject, IHasCatchObjectState
```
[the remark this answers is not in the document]
ekrctb: This is an error from Rider renaming feature. I'll fix it.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
```diff
@@ -0,0 +1,100 @@
...
+            HyperDash.Value = objectState.HyperDash.Value;
+        }
+
+        protected override void FreeAfterUse()
+        {
+            ClearTransforms();
+
+            Alpha = 1;
+            LifetimeStart = double.MinValue;
+            LifetimeEnd = double.MaxValue;
```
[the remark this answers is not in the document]
ekrctb: Lifetime is set by `Expire` when a caught object is dropped. If lifetime is not reset issues will occur.
I removed assignment of `LifetimeStart` because that was not doing anything (I thought the object will be removed when a replay is rewound but it is not, and I added object removing logic to `OnRevertResult`)

osu.Game.Rulesets.Catch/Skinning/Default/CatchHitObjectPiece.cs
```diff
@@ -54,8 +47,8 @@ protected override void LoadComplete()
 
         protected override void Update()
         {
-            if (BorderPiece != null && DrawableHitObject?.HitObject != null)
-                BorderPiece.Alpha = (float)Math.Clamp((DrawableHitObject.HitObject.StartTime - Time.Current) / 500, 0, 1);
+            if (BorderPiece != null && ObjectState?.HitObject != null)
```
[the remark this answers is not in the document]
ekrctb: For a DHO, `HitObject` can be null here for some reason.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObject.cs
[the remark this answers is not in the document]
ekrctb: I somehow thought lifetime had to reset but apparently it is reset by `DrawablePool` and the lines above is not necessary. I think I misunderstood a different bug (that I solved).

osu.Game.Rulesets.Catch/UI/Catcher.cs
ekrctb: Actually this check is not needed anymore because this method was public but it is now private and the exception shouldn't occur.

bdach (dismissed):
Looks good to go as far as I can tell :+1:

bdach:
I think the accent colour issue can be resolved in a separate pull. An option would be to have `AccentColour` in the hitobjects themselves and associate `CaughtObject`s with them, but I imagine managing that state could turn ugly...

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableBanana.cs
```diff
@@ -44,12 +44,12 @@ protected override void UpdateInitialTransforms()
             const float end_scale = 0.6f;
             const float random_scale_range = 1.6f;
 
-            ScaleContainer.ScaleTo(HitObject.Scale * (end_scale + random_scale_range * RandomSingle(3)))
-                          .Then().ScaleTo(HitObject.Scale * end_scale, HitObject.TimePreempt);
+            this.ScaleTo(HitObject.Scale * (end_scale + random_scale_range * RandomSingle(3)))
```
peppy, line 47: This is probably not what we want to be doing. The rationale behind not adjusting one's own scale is that mods may want to do so as well, causing potentially loss (or otherwise incorrect) transforms.

Can you add back the local scale container and instead forward the interface implementation to the `HitObject` model's scale (can be done implicitly to avoid conflict with the underlying `Drawable.Scale`)?

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableDroplet.cs
```diff
@@ -26,9 +24,9 @@ public DrawableDroplet([CanBeNull] CatchHitObject h)
         [BackgroundDependencyLoader]
         private void load()
         {
-            ScaleContainer.Child = new SkinnableDrawable(
+            AddInternal(new SkinnableDrawable(
```
peppy, line 27: As above.

peppy (changes requested):

ekrctb:
My commit above changed the radius of banana considered for the caught object stacking. It is now more correct and matches visual (previously scaling transform was ignored for the radius).

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -124,7 +134,11 @@ public Catcher([NotNull] Container trailsTarget, [NotNull] Container droppedObje
             InternalChildren = new Drawable[]
             {
                 hitExplosionPool = new DrawablePool<HitExplosion>(10),
-                caughtFruitContainer = new Container<DrawablePalpableCatchHitObject>
+                caughtFruitPool = new DrawablePool<CaughtFruit>(50),
```
peppy, line 137: As a safety to avoid asynchronous loading, we'd usually put these inside the `[BackgroundDependencyLoader]` method, but I've checked this case is fine here (async happens from `CreateDrawableRuleset` and includes all `ctor`s).

peppy (approved):
Looks good here!

PR #11133, opened 2020-12-09
Allow ModDifficultyAdjustment to extend beyond the sane limits of the game

### LINKED ISSUE #7378: Difficulty adjust mod should allow AR to be set up to 11
I think the mod should allow up to AR 11 at the very least since this is actually an achievable AR by applying DT to AR 10.

Also I can see many players who would want to practice AR 10.3 on certain maps where it would other wise be much higher (10.4-11).

smoogipoo:
Just something I think would be pretty cool to have, for practice purposes or otherwise. Allows extending AR, OD, HP, and CS to 11, but can be extended further if desired.

Closes https://github.com/ppy/osu/issues/7378.


Joehuu:
Linked a related issue.

bdach (dismissed):
I see no issues with this at the code level. Can't ever see myself using it but I guess it won't hurt to have it be a thing.

osu.Game.Rulesets.Osu/Objects/Drawables/Connections/FollowPointConnection.cs
```diff
@@ -110,8 +110,11 @@ public static void GetFadeTimes(OsuHitObject start, OsuHitObject end, float frac
             double startTime = start.GetEndTime();
             double duration = end.StartTime - startTime;
 
+            // For now, adjust the pre-empt for approach rates > 10.
```
peppy, line 113: Is this to match how it behaves in stable?

smoogipoo: Stable's also static at 800 at all times, with the only thing changing being the animation speed.

I did intend to make it match but after thinking about it again this is actually incorrect as it should be a constant 33% reduction rather than based on preempt time, both here and with the change in `OsuHitObject`.

peppy: This likely deserves a better inline explanation I think. I'm still not really sure how this works, even with your comment above (and if I'm not sure, it's probably going to be equally hard for others to figure out).

Joehuu:
I can reproduce 100% of the time (if you are patient enough of getting second step right):
- load same map ^, or any epilepsy map
- when map is done loading, hover visual settings until it goes to gameplay
- when in player, press quick retry

kzN4xX0XQk

smoogipoo:
Give the new inline descriptions another read through - they should hopefully better explain the situation.

peppy (approved):

PR #11134, opened 2020-12-09
Change accuracy curve for speed pp awarded in the osu! ruleset

Xexxar:
**refactored 12/10/2020**

Hello all, this change is part of my goal of revitalizing the existing pp system in an effort to keep the competitive scene alive. In addition to this goal, this change is specifically addressed towards encouraging the "rhythm" part of a "rhythm game." While I wish this change was adjusting diffcalc to incorporate rhythm as a skill being assessed in the maps difficulty, I believe there is still merit in this change, as I will try to demonstrate.

**I am by no means saying that this approach is the best solution, nor that its preferable to a rework that would better balance accuracy with the difficulty of the map, etc. I simply am trying to produce a more favorable result in the short term that will fix this glaring oversight in the existing functions.**

**The Problem**
In the current system, PP comes from three sources: aim, speed, and accuracy. In times past, accuracy played a large part of a plays total PP value, as the largest scores were only receiving a total of 500-700pp, and accuracyPP was usually equal to about 150-200pp. With this, we have about 20 to 40% of pp coming from accuracy. As scores have changed and evolved over time, the prevalence of scores where accuracy accounts for less than 20% of pp has exploded, and because of this the role of accuracy has diminished as well, with some players being rewarded over 1000 pp for an 87% full combo play (https://osu.ppy.sh/scores/osu/3338602354)

Currently, this play is worth 695.15 speed pp, which is reduced from 798.89 speed pp for an SS based off the existing accuracy multiplication found here

***
            // Scale the speed value with accuracy _slightly_
            speedValue *= 0.02 + accuracy;
***

While this reduction in PP is good, the question I have is whether or not this reduction is enough. Lets take a player as an example.

While a rather extreme example, this player (found here: https://osu.ppy.sh/users/8811966) has a global accuracy ranking of 80%, and several plays that are low 70%s. While this player is certainly able to play some impressive aim maps, his low accuracy begs the question as to why such a player is nearly 4 digits. 

**Purpose of the Proposal**
Accuracy is an incredibly important part of any game that is a rhythm game. While I am not denying that many of these plays are impressive and should be worth something, the high pp values that are being awarded for low accuracy plays is damaging to the playerbase. New players are incentivized to reach for maps outside of their skill range for the hope of miraculously pulling off a low accuracy FC, rather than focusing on growing as a player. While it made sense to scale accuracy _slightly_ before, the evolution of the game has brought maps specifically designed to harness the rewards for speed and aim, and as such, players are turning to these maps rather than playing things in their real comfort zone. By punishing low acc plays, players are incentivized to turn towards playing the game the way it was meant to be played, as a rhythm game (not as an aim trainer or a keyboard mashing simulator).

**Proposal**
The change takes the existing curve for speed and modifies it to be the new curve shown below.

Current Speed Curve: https://www.desmos.com/calculator/wazbl4t1ji

Proposed New Speed Curve: https://www.desmos.com/calculator/grhau486u4

Difference Curve: https://www.desmos.com/calculator/grmcwpraol

All three at once: https://www.desmos.com/calculator/8qfjfnto4r

These functions give a sharp curve for most OD ranges, while not being too extreme at lower OD ranges (by capping curve sloping at OD8). This means that players who are able to get 90% FCs on OD11 are not punished as hard as players who are getting 90% FCs on OD 9 or 8.

In addition to this curve, I also added an existing system in delta_t's calculation to punish double tapping. This code is found here.

```
// Scale the speed value with # of 50s to punish doubletapping.
            speedValue *= Math.Pow(0.98, countMeh < totalHits / 500.0 ? 0.5 * countMeh : countMeh - totalHits / 500.0 * 0.5);
```

Graph: https://www.desmos.com/calculator/tzbx9fghnl

**Dissent**
Now I'm sure there will some (or a lot) of pushback on this change. For this reason, I felt it would be a good idea to just bring up a couple of those now.

- The change affects the 95-100% region, which isn't the intention of this change

The amount of change in this range is typically only 10~ for ranges between 95-100 for most standard 300-500pp. Top players can receive accuracy buffs of about 20PP at the max ranges. I think these minor buffs encourage pursuing near SS accuracy on maps, which is good.

- What about maps where difficulty is not consistent? This won't address that.

Due to the fact that maps can vary wildly in difficulty throughout (think Hidamari no Uta), if only a small portion of the map encapsulates 90% of the difficulty, and you only perform bad accuracy during that part, you can still easily have about 95%. This change unfortunately cannot address that. The point of this change is to generally incentivize higher accuracy. Though it does it in an arbitrary way, so does the existing scaling.

- Shouldn't this be done using the accuracy skill?

Probably. But reworking accuracy is a tall order in itself. There are tr3's accuracy changes that certainly have some merit, perhaps this can be revisited at a later time in the future.

I'm sure there are more concerns but I wanted to just start the discussion.

Some notable profiles with these changes: **(NOTE THESE SCORES CONTAIN AR11 REWORK AS WELL)**

From the post:
Lokikaos: https://cdn.discordapp.com/attachments/745401507553476759/786411973159419934/lokikaos.txt
cute_catboy: https://cdn.discordapp.com/attachments/745401507553476759/786615013547704401/cute_catboy.txt

Some others:
Whitecat: https://cdn.discordapp.com/attachments/745401507553476759/786614783075156019/whitecat.txt
Aetrna: https://cdn.discordapp.com/attachments/745401507553476759/786614604959580160/aetrna.txt
mrekk: https://cdn.discordapp.com/attachments/745401507553476759/786615337176662056/mrekk.txt
chocomint: https://cdn.discordapp.com/attachments/745401507553476759/786615528369160212/chocomint.txt (fdfd 900 pp)
unko: https://cdn.discordapp.com/attachments/745401507553476759/786619083843567666/unko.txt
umbre: https://cdn.discordapp.com/attachments/745401507553476759/786619116785631302/umbre.txt
matrix: https://cdn.discordapp.com/attachments/745401507553476759/786619880949809163/matrix.txt
idke: https://cdn.discordapp.com/attachments/745401507553476759/786624476536111165/idke.txt
ming3012: https://cdn.discordapp.com/attachments/745401507553476759/786630236350709800/ming.txt
vaxei: https://cdn.discordapp.com/attachments/745401507553476759/786629908352466994/vaxei.txt
ryuk: https://cdn.discordapp.com/attachments/745401507553476759/786627755416485938/ryuk.txt
xexxar (me!): https://cdn.discordapp.com/attachments/745401507553476759/786619507808272384/xexxar.txt

### DIFF
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -52,6 +52,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.1202.0" />
-    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.1203.0" />
+    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.1212.0" />
   </ItemGroup>
 </Project>
--- a/osu.Game/osu.Game.csproj
+++ b/osu.Game/osu.Game.csproj
@@ -26,7 +26,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.2.6" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="2.2.6" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
-    <PackageReference Include="ppy.osu.Framework" Version="2020.1203.0" />
+    <PackageReference Include="ppy.osu.Framework" Version="2020.1212.0" />
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.1202.0" />
     <PackageReference Include="Sentry" Version="2.1.8" />
     <PackageReference Include="SharpCompress" Version="0.26.0" />
--- a/osu.iOS.props
+++ b/osu.iOS.props
@@ -70,7 +70,7 @@
     <Reference Include="System.Net.Http" />
   </ItemGroup>
   <ItemGroup Label="Package References">
-    <PackageReference Include="ppy.osu.Framework.iOS" Version="2020.1203.0" />
+    <PackageReference Include="ppy.osu.Framework.iOS" Version="2020.1212.0" />
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.1202.0" />
   </ItemGroup>
   <!-- See https://github.com/dotnet/runtime/issues/35988 (can be removed after Xamarin uses net5.0 / net6.0) -->
@@ -88,7 +88,7 @@
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite" Version="2.2.6" />
     <PackageReference Include="Microsoft.EntityFrameworkCore.Sqlite.Core" Version="2.2.6" />
     <PackageReference Include="Newtonsoft.Json" Version="12.0.3" />
-    <PackageReference Include="ppy.osu.Framework" Version="2020.1203.0" />
+    <PackageReference Include="ppy.osu.Framework" Version="2020.1212.0" />
     <PackageReference Include="SharpCompress" Version="0.26.0" />
     <PackageReference Include="NUnit" Version="3.12.0" />
     <PackageReference Include="SharpRaven" Version="2.4.0" />


bdach:
We use the tab width of 4 in this repository, not 2 like originally used in this PR. I've fixed it up for you this time (along with a pair of typos), but in the future I would ask you to be more vigilant about code style. As outlined in the [contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md) you can use the `InspectCode.ps1` script to run code analysis locally before pushing.

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -121,10 +121,13 @@ private double computeAimValue()
                                 : 0.0);
             }
 
-            // Scale the aim value with accuracy _slightly_
-            aimValue *= 0.5 + accuracy / 2.0;
+            // Scale the aim value with accuracy _a lot_
+            if (accuracy > .8)
+                aimValue *= 0.5 + Math.Pow(Math.Sin(2.5 * (accuracy - .8) * Math.PI), 2) / 2;
```
bdach, line 126: I'm not sure I can meaningfully comment on the general validity of this curve change, but I find the actual shape/plot a little... arbitrary? Is there a particular reason why the piece at and above 0.8 is a sin<sup>2</sup> instead of, say, another linear, or quadratic piece?

If it's to keep smoothness, then that's not really happening, as the function is not smooth (first derivative is not continuous as far as I can tell - it goes from 5/8ths on the left side of 0.8 to 0 on the right side).

Not entirely sure why small changes in accuracy in the middle part of that sine curve part should result in larger changes than at its ends.

02Naitsirk:
The accuracy curve really needs to scale with OD and it's a bad idea to arbitrarily nerf below 95% without accounting for OD. aetrna's 3mod score on Ascension to Heaven is 85% and worth 750, which is extremely low. When the OD is lowered to match the original, the accuracy increases to 95%. So, the score should be worth more than the DT play because the circle size is lower and it is AR 11. You have to create a function where 85% OD 11 is equivalent to 95% OD 9.6, and so on.

Xexxar:
Rewrote the curve. Will update main post later.

bdach:
Please don't open PRs if you're going to just entirely nuke the original change like that. That wasted like 15 minutes of my time. Get it ready and only then PR.

Xexxar:
Will close for now then. My mistake.

Xexxar:
Reopened this change after working through the problem with several other members of the community. Main points to take away:

Only the speed curve is affected by this change now.
Curve scales with OD to not overly nerf low accuracy but high OD plays.

bdach:
> In addition to this curve, I also added an existing system in delta_t's calculation to punish double tapping.

Would you mind elaborating a little as to what "double tapping" is referring to in this context? Intuitively I'm imagining some cheese technique that exploits near-overlapping hit windows at high BPM to secure non-misses while not actually following the beat?

Xexxar:
> > In addition to this curve, I also added an existing system in delta_t's calculation to punish double tapping.
> 
> Would you mind elaborating a little as to what "double tapping" is referring to in this context? Intuitively I'm imagining some cheese technique that exploits near-overlapping hit windows at high BPM to secure non-misses while not actually following the beat?

Yeah that's basically exactly correct. The play mentioned in the post by Lokikaos features 330 bpm streams which he is able to do by hitting both keys at once. This gives him a lot of 50s in his play. Using a proper technique would be significantly more difficult, and would rarely produce 50s. Thus, checking for 50s and nerfing plays based on the # is just a good way to nerf this technique, while not penalizing proper playstyles.

smoogipoo:
The "proposed" graph is different from the one in "all three at once", I'm pretty sure you want the "all three at once" one.

Also would be nice to graph the second part which nerfs double tapping, something like this maybe? https://www.desmos.com/calculator/tzbx9fghnl

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -154,10 +154,10 @@ private double computeSpeedValue()
             if (mods.Any(m => m is OsuModHidden))
                 speedValue *= 1.0 + 0.04 * (12.0 - Attributes.ApproachRate);
 
-            // Scale the speed value with accuracy _slightly_
-            speedValue *= 0.02 + accuracy;
-            // It is important to also consider accuracy difficulty when doing that
-            speedValue *= 0.96 + Math.Pow(Attributes.OverallDifficulty, 2) / 1600;
+            // Scale the speed value with accuracy and OD
+            speedValue *= (.95 + Math.Pow(Attributes.OverallDifficulty, 2) / 750) * Math.Pow(accuracy, (14.5 - Math.Max(Attributes.OverallDifficulty, 8)) / 2);
```
smoogipoo, line 158:
```suggestion
            speedValue *= (0.95 + Math.Pow(Attributes.OverallDifficulty, 2) / 750) * Math.Pow(accuracy, (14.5 - Math.Max(Attributes.OverallDifficulty, 8)) / 2);
```

smoogipoo (changes requested):
Just a code style nit

Xexxar:
fixed the original curves mentioned in smoogi's post as well as the code style

smoogipoo (dismissed):

Xexxar:
closing for now cause i torched this branch when i was working on something else on accident

peppy:
That's not really how we use git/github. By doing so you are hiding the history of this change. Can you instead let me know which commits you didn't intend to be here and we'll just revert things back?

Xexxar:
All commits from 12/12. I'm not used to working with different remotes and when I pulled in Joz's SR calculation I was having difficulties and accidently merged the changes into this branch by mistake. Specifically commit ecaf4de18c47d126a3bbc6559c7e457be51a2c1f. I tried to revert the change, but it doesn't seem like that went through. I guess at the end of the day this MR can easily be squashed to make the commit cleaner, but it just bothered me. Sorry for any confusion I caused, and thanks for the help.

peppy:
I've just force pushed this back since there are no attached code reviews (shouldn't break anything).

Please make sure you know what you are doing before you commit to the branch again (you can check your local `git log` to make sure you are happy with new commits before `push`ing).

Xexxar:
as the suggestion of spaceman_atlas, this thread is "a train wreck" so im just reopening the MR at https://github.com/ppy/osu/pull/11175. :^)

PR #11138, opened 2020-12-10
Port StanR's NF multiplier changes

smoogipoo:
Port of https://github.com/ppy/osu-performance/pull/127


bdach (approved):
seems functionally equivalent and it's already been merged `osu-performance`-side, so throwing an approval just in case one's needed

peppy (approved):
guess we are going with this for now

PR #11139, opened 2020-12-10
Fix single threaded seeking not working due to unnecessary seek call

### LINKED ISSUE #11122: Seeking gameplay doesn't work in single-threaded execution
**Describe the bug:** 

Seems like the update of the clock isn't happening as instantaneously as we'd like from `GameplayClockContainer.Seek(time)`, and the following call to `GameplayClockContainer.Start()` seeks back to the original time.

https://drive.google.com/file/d/1bktoK_Ac1EGFwSGQ1Num9KQ-XqbR4pI6/view?usp=sharing

peppy:
Seeking to `GameplayClock.Time` when calling start - even if the clock is already started - would cause an incorrect second seek, cancelling the user's proposed seek.

I also moved the pause logic here because it seems harmless and feels better.

Closes #11122.


osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -95,6 +95,16 @@ public GameplayClockContainer(WorkingBeatmap beatmap, double gameplayStartTime,
             localGameplayClock = new LocalGameplayClock(userOffsetClock);
 
             GameplayClock.IsPaused.BindTo(IsPaused);
+
+            IsPaused.BindValueChanged(onPaused);
+        }
+
+        private void onPaused(ValueChangedEvent<bool> isPaused)
```
bdach, line 102: Sort of a nitpick but I'd rename to `onPausedChanged`.

osu.Game/Screens/Play/GameplayClockContainer.cs
```diff
@@ -154,13 +164,16 @@ public void Restart()
...
-            // Seeking the decoupled clock to its current time ensures that its source clock will be seeked to the same time
-            // This accounts for the audio clock source potentially taking time to enter a completely stopped state
-            Seek(GameplayClock.CurrentTime);
-            adjustableClock.Start();
-            IsPaused.Value = false;
+            if (!adjustableClock.IsRunning)
+            {
+                // Seeking the decoupled clock to its current time ensures that its source clock will be seeked to the same time
+                // This accounts for the audio clock source potentially taking time to enter a completely stopped state
+                Seek(GameplayClock.CurrentTime);
```
bdach, line 171: Just to clarify that I understand the mode of failure correct: this "worked" when multi-threaded but failed single-threaded because in multi-threaded mode enqueued seeks could execute while the update thread was doing its things, making this a race in the multi-threaded case:

```
Update thread						Audio thread

enqueues seek to requested time
									executes enqueued 1st seek
enqueues seek to current time
									executes enqueued 2nd seek
```

and making the first seek a no-op in the single-threaded case (as the second seek reverted it entirely), right?

```
Frame N

update loop enqueues seek to requested time
update loop enqueues seek to current time

Frame N+1

audio thread executes enqueued 1st seek
audio thread executes enqueued 2nd seek <-- reverting 1st seek
```

peppy: That's correct, yeah. The audio thread was running so often that it was not an issue in multi-threaded execution.

bdach (approved):

PR #11150, opened 2020-12-11
Update framework

peppy:
Includes SDL fixes and better pooling display in GlobalStatistics.


bdach:
Think https://github.com/ppy/osu-framework/pull/4090 is worth holding this bump off for?

bdach (approved):

PR #11159, opened 2020-12-12
Refactor and restore taiko hits test scene to a working state

bdach:
Currently on `master` taiko's `TestSceneHits` is in a pretty busted-up state from a collection of relatively recent mutually unrelated changes. This PR aims to restore things to a working state, without considering larger refactorings yet.

Taking it commit by commit:

* b56e932 - This wasn't actually broken; the commit is actually a preparatory change for pooling and sort of the pretext for this PR. Because at one point nested objects will be pooled, it gets appropriately harder to jam in the parent hitobject like the test scene does now (because with the pooling changes I'm replacing taiko drawables' `MainObject` with the now-existing `ParentHitObject`). Yields a net line decrease, too, so I think it's objectively a change for the better.
* d7551d9 - Broke with some test scene refactorings. Defaults were applied twice (once to the HO itself in the test scene, once in `DrawableTestHit`).
* f49d908 - Broke with judgement pooling/animation changes due to using an absolute sequence starting at `Result.TimeAbsolute`.
* 49768d4 - Broke with editor changes to support two-way sample and colour/strong state synchronisation. This is by far the wonkiest change and I dislike it right now, but I'm not necessarily aiming to address that just yet (and risk breaking all of taiko). I might address this at a later point after I'm done with the pooling series; my main thought for now is to move out the synchronisation logic from DHOs to HOs, although I haven't yet even looked towards checking if that is a viable direction wrt legacy decoding logic.
* e0ba5a4 - Not entirely sure where this broke, but I do recall it working as it does after the change.

### DIFF
--- a/osu.Game.Rulesets.Taiko.Tests/DrawableTaikoRulesetTestScene.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/DrawableTaikoRulesetTestScene.cs
@@ -16,6 +16,8 @@ namespace osu.Game.Rulesets.Taiko.Tests
 {
     public abstract class DrawableTaikoRulesetTestScene : OsuTestScene
     {
+        protected const int DEFAULT_PLAYFIELD_CONTAINER_HEIGHT = 768;
+
         protected DrawableTaikoRuleset DrawableRuleset { get; private set; }
         protected Container PlayfieldContainer { get; private set; }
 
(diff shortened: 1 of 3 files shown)


peppy (approved):
looks like a sane starting point, i'm sure it will change during pooling impl. so not going to dwell on it too much.

PR #11160, opened 2020-12-13
Make carousel scrolling drawable independent 

### LINKED ISSUE #11156: Beatmap panels should have consistent scrolling animations
**Describe the bug:** 
> further work could be done to add animation back to pooled panels but that's not happening in this PR.
https://github.com/ppy/osu/pull/10973#issuecomment-735027676

When scrolling to a new location, beatmap panels should have consistent animations. Animations should be done in a way that no panel will overlap during an operation like filtering where the order of panels isn't changed. Also no panel should be left with no animations (see 2nd video for that, sorting looks cool but that effect is lost if the panel isn't in pool already)

Unexpected behaviour (filtrering does not change the order of beatmaps so scrolling should not overlap)
https://youtu.be/ZvcV6Ncudvs

Somewhat expected behaviour (sorting is an operation that *does* change the order of beatmaps)
https://youtu.be/d4UvLCO0Nuw

**osu!lazer version:**
2020.1204.0

LittleEndu:
* [x] Closes #11156
* [ ] Figure out why panels weren't updating their position while being filtered (I can't reproduce the behaviour mentioned in the comment)

## Summary
This PR would remove anything that changes drawable Y position from beatmap carousel, moving that logic to underlying models instead. This makes all scrolling look consistent as the underlying models are never unloaded.

## Changes
* Calculate Y position in ``CarouselBeatmapSet`` - Call that calculation from ``BeatmapCarousel``
* Use calculated Y position in ``DrawableCarouselBeatmapSet``
* Correctly initialize ``DrawableCarouselBeatmapSet`` Y position when it's gotten from pool
* Find target Y / ``CarouselYPosition`` even for filtered sets
  * This is to add more consistency when filtering items. Newly visible items get a good position right away.
* Fix bug in scrolling off extremes 
  * Squeezing it into this PR, is a 2 line change (``visibleHalfHeight - BleedTop/Bottom`` change)
* Update drawable Y when scrolling to avoid off-by-one frame issues
  * Issues like drawables appearing in wrong places because they didn't get unloaded correctly by some scrolling that would have placed them outside visible area

### DIFF
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -629,8 +629,11 @@ namespace osu.Game.Screens.Select
             }
 
             // Update Y positions for the non-drawable models
-            foreach (var set in beatmapSets)
-                set.UpdateYPosition(Time.Elapsed);
+            if (Clock != null)
+            {
+                foreach (var set in beatmapSets)
+                    set.UpdateYPosition(Time.Elapsed);
+            }
 
             // Update externally controlled state of currently visible items (e.g. x-offset and opacity).
             // This is a per-frame update on all drawable panels.
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -567,6 +567,9 @@ namespace osu.Game.Screens.Select
 
         #endregion
 
+        private double timeSinceLastModelUpdate;
+        private const int model_updates_per_second = 10;
+
         protected override void Update()
         {
             base.Update();
@@ -631,8 +634,23 @@ namespace osu.Game.Screens.Select
             // Update Y positions for the non-drawable models
             if (Clock != null)
             {
-                foreach (var set in beatmapSets)
-                    set.UpdateYPosition(Time.Elapsed);
+                if (timeSinceLastModelUpdate + Time.Elapsed > 1000.0 / model_updates_per_second)
+                {
+                    foreach (var set in beatmapSets)
+                        set.UpdateYPosition(timeSinceLastModelUpdate + Time.Elapsed, true);
+
+                    timeSinceLastModelUpdate = 0;
+                }
+                else
+                {
+                    foreach (var d in Scroll.Children)
+                    {
+                        var set = d.Item as CarouselBeatmapSet;
+                        set?.UpdateYPosition(Time.Elapsed, false);
+                    }
+
+                    timeSinceLastModelUpdate += Time.Elapsed;
+                }
             }
 
             // Update externally controlled state of currently visible items (e.g. x-offset and opacity).
--- a/osu.Game/Screens/Select/BeatmapCarousel.cs
+++ b/osu.Game/Screens/Select/BeatmapCarousel.cs
@@ -579,7 +579,6 @@ namespace osu.Game.Screens.Select
                 updateYPositions();
 
             // if there is a pending scroll action we apply it without animation and transfer the difference in position to the panels.
-            // this is intentionally applied before updating the visible range below, to avoid animating new items (sourced from pool) from locations off-screen, as it looks bad.
             if (pendingScrollOperation != PendingScrollOperation.None)
                 updateScrollPosition();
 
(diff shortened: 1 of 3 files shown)


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -618,27 +617,32 @@ protected override void Update()
...
-                        panel.Y = item.CarouselYPosition;
+                        panel.UpdateY();
 
                         Scroll.Add(panel);
                     }
                 }
             }
 
+            // Update Y positions for the non-drawable models
+            if (Clock != null)
```
peppy, line 632: How can clock ever be null? (hint: it can't)

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -618,27 +617,32 @@ protected override void Update()
...
+            {
+                foreach (var set in beatmapSets)
```
peppy, line 634: is this not updated *every* panel ever? this is going to be a performance issue (the very reason everything else in this whole class is structured to specifically avoid this).

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -729,21 +733,21 @@ private void updateYPositions()
...
-                if (item.Filtered.Value)
-                    continue;
-
                 switch (item)
                 {
                     case CarouselBeatmapSet set:
                     {
-                        visibleItems.Add(set);
                         set.CarouselYPosition = currentY;
+                        if (item.Filtered.Value)
```
peppy, line 747: this is SUPER weird to be here, with no comment and only in this part of the `switch`

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
```diff
@@ -43,6 +44,26 @@ public CarouselBeatmapSet(BeatmapSetInfo beatmapSet)
                       .ForEach(AddChild);
         }
 
+        public float YPosition;
+
+        public void UpdateYPosition(double elapsed)
```
peppy, line 49: models storing drawable specific values is a huge regression. this shouldn't need to exist.

peppy (changes requested):
Don't like this one bit.

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: It can in testing. The testing only function somehow calls update before the clock is initialized. Could try to remove that function instead.

https://github.com/ppy/osu/blob/941ccb6939bea605132e9818b532212fb86ce6e6/osu.Game/Screens/Select/BeatmapCarousel.cs#L98-L103

bdach: That just exposes the root of the problem which is coupling of models to their drawable representation, which [I warned about in discord yesterday already](https://discord.com/channels/188630481301012481/188630652340404224/787445202310529055).

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: This part of the switch is the only part. In fact I think carousel root only hold beatmap sets.

But this is here so that we set the target Y / ``CarouselYPosition`` for even the filtered sets. They would appear at or fade into the beatmap set above them

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: I have 5392 beatmaps and see no difference in performance compared to master.

bdach: How did you measure that?

LittleEndu: Just looking at FPS. I get ~100 FPS update when scrolling in both this PR and current release

bdach: A more accurate (and believable) way to do that would be to either profile or run a benchmark.

osu.Game/Screens/Select/Carousel/CarouselBeatmapSet.cs
LittleEndu: I don't think there's any way to both have drawable pooling and have all this logic still be calculated. If the drawable panel is off screen and released back into pool, there just isn't any other place that could calculate the Y position. 

I would argue this logic has nothing to do with "drawable specific values". If ``YPosition`` sounds too related to drawables, renaming it to ``ScrollPosition`` might be less related.

bdach: It's data directly used to drawing on screen, therefore it's drawable specific.

An alternative that wouldn't be drawable specific would be some sort of index on the carousel.

LittleEndu: Like move that logic to ``CarouselRoot : CarouselGroupEagerSelect`` ?? Calculate it all at the same time?

bdach: Maybe, not necessarily, not sure. A viewmodel-like structure wrapping the model might be an answer. You'd have to try and see how you go.

peppy: At *very least*, such a value can not be updated every frame as you have changed it to be. This is the incorrect approach at this whole thing.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -567,6 +567,9 @@ private void endRepeatSelection(object source = null)
 
         #endregion
 
+        private double timeSinceLastModelUpdate;
+        private const int model_updates_per_second = 10;
```
bdach, line 571: I don't know how else to respond to this than "just no". This is on multiple levels of wrong and makes the concerns of violating lack of separation between data and drawing logic miles worse.

bdach (changes requested):

LittleEndu:
I'll look to rework this to be in line with model - view - binder structure.

PR #11162, opened 2020-12-13
improve selection box rotation UX

Grrum:
When I try to rotate a selection from the selection box, the rotation will spazz out if I rotate too much. It's not a great UX, so I made a better one that will always follow the mouse as you rotate.

The problem seemed a bit small to create an issue on github, especially when the fix is straightforward.

Not sure how to make and submit a gif that would demonstrate the before and after behavior, so if someone could show me how that would be much appreciated!

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionBox.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionBox.cs
@@ -191,7 +191,7 @@ namespace osu.Game.Screens.Edit.Compose.Components
                 {
                     Anchor = Anchor.TopCentre,
                     Y = -separation,
-                    HandleDrag = e => OnRotation?.Invoke(e.Delta.X),
+                    HandleDrag = e => OnRotation?.Invoke(convertDragEventToAngleOfRotation(e)),
                     OperationStarted = operationStarted,
                     OperationEnded = operationEnded
                 }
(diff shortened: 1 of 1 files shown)


bdach:
Why was the previous PR closed?

bdach:
So I checked this out and while I can see where this pull is coming from, I'm not sure I agree this is an improvement to the UX. [Here's a video to demonstrate my point.](https://drive.google.com/drive/u/0/folders/13odkQbP7ogN7J8nxrId_a5Fb6IuZ57CY)

To summarise what I think is wrong with the above clip - I really don't like how the rotate handle stays static while the mouse is being moved around when held down. I attached a clip from Google Slides for comparison - the rotate functionality in most office suites will have rotation working as you have it, but it will *also* rotate the entire bounding box of the selection, instead of keeping it as just an upright box as this PR has it.

Grrum:
I think this is an improvement to the current behavior. I should have provided a gif to show the comparison, and I hope it's clear why the the old rotation is unintuitive and should be changed:
ezgif com-gif-maker
These rotations do not make sense given what the mapper is trying to do. 

It would be nice to move the rotation handle as well. I think it's important that the box stays horizontal after the rotation is complete in order to keep the flip/scaling transformations from being too complicated. 

While the PR doesn't polish off this feature completely, I think this can be broken up into steps. Step one is to make rotations make sense. Step two is to rotate the handle in some way to look nice. I can create an issue for that.

Old PR was closed because it was corrupted in some unimportant way.

peppy:
Yeah, I can agree that what is here could be considered as an intermediary improvement over what we have in master. Definitely also agree about the bounding box, but that's a change for another day.

peppy (approved):

PR #11163, opened 2020-12-13
Refactor taiko's strong state management in hitobject hierarchy

bdach:
This is another seemingly inconsequential pull, brought up in part in preparation for larger pooling changes. Reading the changes commit-by-commit might be slightly easier than the entire diff at once.

Overall strong state is in a somewhat awkward spot in current `master` - we have a flag that determines whether a HO can be strong or not, and a few places that need to work around this. This pull aims to clean up that area a bit by using a more OOP-like paradigm - the three object types that need being strong (hits, drum rolls and drum roll ticks) inherit `TaikoStrongHitObject`, and all strong logic is removed from the base `TaikoHitObject` (f74567e). This change in approach is mirrored in the DHOs themselves (b1635ec).

Some existing nested HO references get a `Nested` in their name in this pull too, to unify with DHOs and be able to distinguish what is what easier (1ddc896, 4d444df). Naming is one of the more awkward pieces of this PR and the new approach might take time to adjust to. Very much willing to listen to suggestions of alternative naming.

61c488c addresses my original problem that needs to be solved wrt pooling, which is that taiko DHOs that can be strong have separate nested DHO classes for their strong hit portions, which are all coupled to the old `StrongHitObject` model. Our pooling model assumes a 1-1 correspondence between the model and the DHO, so something had to give. My elected choice was to split fully, in having each HO type define its own nested HO type, which then can be tied to a drawable pool further down in the series. I imagine this might not be entirely agreeable, which is why I'm PRing this before further changes.

Finally, 091b08b leverages the above for a bit more type safety, adding an explicit nested strong object type parameter specification to strong DHOs.

Overall I feel this is a net improvement in code quality as this structure makes a few kinks go away, but unsure if that sentiment will be shared by anyone else.


smoogipoo (dismissed):
Looks pretty sane to me

peppy:
Not a huge fan of the `TaikoStrongHitObject` naming, is that just me? Feels like it needs to have something in the name to state that it's not *always* strong. When reading through code it might be easy to mistake this.

smoogipoo:
Something like `TaikoRhythmicHitObject` maybe? Line of thinking is a hitobject which has to be hit at a specific time rather than be hit at whatever arbitrary pace like swells.
Not sure, otherwise.

peppy:
I'd be okay with "rhythmic" if there was more to it than just adding the strong support..

@smoogipoo are you against `Strongable`? It sounds a bit weird but it is the only one I can think of.

smoogipoo:
I’m not against it

smoogipoo (approved):

peppy (approved):

PR #11166, opened 2020-12-13
Implement taiko bar line pooling

bdach:
While I'm waiting for opinions on https://github.com/ppy/osu/pull/11163 to come in before proceeding with the "main course" so to speak, there's still plenty of pooling to be done around in other areas. Bar lines are relatively isolated and easy to make work with pooling, so here's a basic pull with that.

Most of the changes should be relatively inoffensive; the biggest ones by far are probably:

* Unifying `DrawableBarLine` and `DrawableMajorBarLine`: That seemed like a smarter idea than splitting `BarLine` into that and `MajorBarLine`, but if you think otherwise, that is also a valid avenue to go down.
* Purposefully not adding the bar line playfield as a nested playfield to avoid interference in the editor: Not sure how to resolve this better. This will probably have to be returned to when work on taiko editor resumes, as the current way of generating bar lines is probably not going to play with it too well (currently they're being generated once ever on drawable ruleset load).

Initial count of 15 is of course ballpark and should be way above what's ever necessary.

osu_2020-12-13_22-39-36


smoogipoo (dismissed):
lgtm

smoogipoo (approved):

peppy (approved):

PR #11167, opened 2020-12-14
Adjust mania judgement position in legacy skins according to their ScorePosition setting

### LINKED ISSUE #8884: ScorePosition in skin.ini is not implemented yet
**Describe the bug:** stable displays mania hit bursts according to skin.ini scorePosition value. lazer defaults to upper part of mania stage

*Lazer*
image

*stable*
screenshot063

**osu!lazer version:** 2020.421.0

**Logs:** not needed

Firmatorenio:
Attempts to resolve #8884.

Seeing as `Stage`'s `JudgementContainer` is already offset, cut off at the top, not very window scaling friendly and doesn't extend all the way down below the judgement line on 16:9 as it is, it seemed fitting (not to mention easier) to just move the container itself rather than the judgements in it. This could probably be somehow bound to a screen ratio (so that when you change the window's `Width` the judgements' alignment on the playfield didn't change), but as Stable doesn't allow resizing the window all the way into portrait mode to begin with, there isn't much parity to speak of there.

This quite easily supports changing skins on the fly and allows other `ScrollingPlayfield`s to use the skin to fetch their own parameters if needed. Unfortunately, I lack the understanding of how the tests work to create an automatic one of my own, but I have tested it manually with a bunch of edited skins.

Stable (ScorePosition at default/100/200/300):
image

Before, any value on legacy skins resulted in this:
image

After (Lazer/Classic/100/200/300):
image


smoogipoo:
We're keeping such legacy things inside legacy skinning components, except for things like hit position/column spacing that are too hard to do right now. Mania didn't have such a component, but it needed one anyway so I've added one and moved the requisite code in there, in-turn reverting a few changes made in this PR.

smoogipoo (dismissed):

peppy:
Looks to have broken (at least) miss transforms:

peppy (changes requested):
As mentioned

peppy (approved):

PR #11169, opened 2020-12-14
Split `X` property of `CatchHitObject` to `OriginalX` and `EffectiveX`

ekrctb:
Previously, `X` was asymmetric in the getter (gets effective X) and the setter (sets original X), and it could lead to confusion.
This PR splits this property to explicitly named two properties:

- `OriginalX`: original value specified in the beatmap.
- `EffectiveX`: final value used for the gameplay after beatmap processing.

I reviewed each usage of `X` to decide which new property should be used in that context. `OriginalX` is used when working on a beatmap (before or during processing) and `EffectiveX` should be used for gameplay.

`IHasXPosititon` interface is explicitly implemented to prevent accidental use of the property.

I left only setter of `X` (no getter) to make get-only property `OriginalX` describe what it is, by their name (it is the original `X` value). And it makes the diff smaller.


smoogipoo:
For those reviewing this: I've tested this doesn't break diffcalc.

peppy (approved):

smoogipoo (approved):

PR #11171, opened 2020-12-14
Add summary scoring data to replay FrameBundle headers

peppy:
Opening this PR early to get feedback on what is included. Should probably not be merged until we start to consume this (for the in-game scoreboard, for instance).

I've tested that it is sent and received correctly, and it includes the basic details required to display user scores. Not 100% on the structure for getting and populating the score.


smoogipoo:
Looks ok structure wise, will approve and can be merged when proven to be sufficient via a consumption.

smoogipoo (dismissed):

peppy:
@smoogipoo this should be good to go. i've confirmed it contains most everything required to make this work for leaderboard display purposes.

osu.Game/Screens/Play/Player.cs
```diff
@@ -157,14 +157,16 @@ protected override void LoadComplete()
...
-        private Replay recordingReplay;
+        private Score recordingScore;
 
         /// <summary>
         /// Run any recording / playback setup for replays.
         /// </summary>
         protected virtual void PrepareReplay()
         {
-            DrawableRuleset.SetRecordTarget(recordingReplay = new Replay());
+            DrawableRuleset.SetRecordTarget(recordingScore = new Score { Replay = new Replay() });
```
bdach, line 167: I think this could just as well be

```suggestion
            DrawableRuleset.SetRecordTarget(recordingScore = new Score());
```

as `Score` creates its own anyway (spotted this after being confused why the next line doesn't seem to check for null score info, but this one creates the replay explicitly).

osu.Game/Screens/Play/Player.cs
```diff
@@ -157,14 +157,16 @@ protected override void LoadComplete()
                 PrepareReplay();
         }
 
-        private Replay recordingReplay;
+        private Score recordingScore;
```
bdach, line 160: Judging from the fact that this is null-checked lower down in a frame count check I'd probably either `Score?` or `[CanBeNull]` this.

bdach (commented):
Mostly looks good. A few minor notes.

osu.Game/Screens/Play/Player.cs
peppy: Can't use `Score?` on non-nullable classes, so let's go with the attrib version for now.

bdach (approved):
seems fine to me :+1:

PR #11175, opened 2020-12-14
Change accuracy curve for speed pp awarded in the osu! ruleset

Xexxar:
**Continuation of https://github.com/ppy/osu/pull/11134 as I am a bumbling monkey trying to use git. (reopened 12/14).**

Hello all, this change is part of my goal of revitalizing the existing pp system in an effort to keep the competitive scene alive. In addition to this goal, this change is specifically addressed towards encouraging the "rhythm" part of a "rhythm game." While I wish this change was adjusting diffcalc to incorporate rhythm as a skill being assessed in the maps difficulty, I believe there is still merit in this change, as I will try to demonstrate.

**I am by no means saying that this approach is the best solution, nor that its preferable to a rework that would better balance accuracy with the difficulty of the map, etc. I simply am trying to produce a more favorable result in the short term that will fix this glaring oversight in the existing functions.**

**The Problem**
In the current system, PP comes from three sources: aim, speed, and accuracy. In times past, accuracy played a large part of a plays total PP value, as the largest scores were only receiving a total of 500-700pp, and accuracyPP was usually equal to about 150-200pp. With this, we have about 20 to 40% of pp coming from accuracy. As scores have changed and evolved over time, the prevalence of scores where accuracy accounts for less than 20% of pp has exploded, and because of this the role of accuracy has diminished as well, with some players being rewarded over 1000 pp for an 87% full combo play (https://osu.ppy.sh/scores/osu/3338602354)

Currently, this play is worth 695.15 speed pp, which is reduced from 798.89 speed pp for an SS based off the existing accuracy multiplication found here

***
            // Scale the speed value with accuracy _slightly_
            speedValue *= 0.02 + accuracy;
***

While this reduction in PP is good, the question I have is whether or not this reduction is enough. Lets take a player as an example.

While a rather extreme example, this player (found here: https://osu.ppy.sh/users/8811966) has a global accuracy ranking of 80%, and several plays that are low 70%s. While this player is certainly able to play some impressive aim maps, his low accuracy begs the question as to why such a player is nearly 4 digits. 

**Purpose of the Proposal**
Accuracy is an incredibly important part of any game that is a rhythm game. While I am not denying that many of these plays are impressive and should be worth something, the high pp values that are being awarded for low accuracy plays is damaging to the playerbase. New players are incentivized to reach for maps outside of their skill range for the hope of miraculously pulling off a low accuracy FC, rather than focusing on growing as a player. While it made sense to scale accuracy _slightly_ before, the evolution of the game has brought maps specifically designed to harness the rewards for speed and aim, and as such, players are turning to these maps rather than playing things in their real comfort zone. By punishing low acc plays, players are incentivized to turn towards playing the game the way it was meant to be played, as a rhythm game (not as an aim trainer or a keyboard mashing simulator).

**Proposal**
The change takes the existing curve for speed and modifies it to be the new curve shown below.

Current Speed Curve: https://www.desmos.com/calculator/wazbl4t1ji

Proposed New Speed Curve: https://www.desmos.com/calculator/grhau486u4

Difference Curve: https://www.desmos.com/calculator/grmcwpraol

All three at once: https://www.desmos.com/calculator/8qfjfnto4r

These functions give a sharp curve for most OD ranges, while not being too extreme at lower OD ranges (by capping curve sloping at OD8). This means that players who are able to get 90% FCs on OD11 are not punished as hard as players who are getting 90% FCs on OD 9 or 8.

In addition to this curve, I also added an existing system in delta_t's calculation to punish double tapping. This code is found here.

```
// Scale the speed value with # of 50s to punish doubletapping.
            speedValue *= Math.Pow(0.98, countMeh < totalHits / 500.0 ? 0.5 * countMeh : countMeh - totalHits / 500.0 * 0.5);
```

Graph: https://www.desmos.com/calculator/tzbx9fghnl

**Dissent**
Now I'm sure there will some (or a lot) of pushback on this change. For this reason, I felt it would be a good idea to just bring up a couple of those now.

- The change affects the 95-100% region, which isn't the intention of this change

The amount of change in this range is typically only 10~ for ranges between 95-100 for most standard 300-500pp. Top players can receive accuracy buffs of about 20PP at the max ranges. I think these minor buffs encourage pursuing near SS accuracy on maps, which is good.

- What about maps where difficulty is not consistent? This won't address that.

Due to the fact that maps can vary wildly in difficulty throughout (think Hidamari no Uta), if only a small portion of the map encapsulates 90% of the difficulty, and you only perform bad accuracy during that part, you can still easily have about 95%. This change unfortunately cannot address that. The point of this change is to generally incentivize higher accuracy. Though it does it in an arbitrary way, so does the existing scaling.

- Shouldn't this be done using the accuracy skill?

Probably. But reworking accuracy is a tall order in itself. There are tr3's accuracy changes that certainly have some merit, perhaps this can be revisited at a later time in the future.

I'm sure there are more concerns but I wanted to just start the discussion.

Some notable profiles with these changes: **(NOTE THESE SCORES CONTAIN AR11 REWORK AS WELL)**

From the post:
Lokikaos: https://cdn.discordapp.com/attachments/745401507553476759/786411973159419934/lokikaos.txt
cute_catboy: https://cdn.discordapp.com/attachments/745401507553476759/786615013547704401/cute_catboy.txt

Some others:
Whitecat: https://cdn.discordapp.com/attachments/745401507553476759/786614783075156019/whitecat.txt
Aetrna: https://cdn.discordapp.com/attachments/745401507553476759/786614604959580160/aetrna.txt
mrekk: https://cdn.discordapp.com/attachments/745401507553476759/786615337176662056/mrekk.txt
chocomint: https://cdn.discordapp.com/attachments/745401507553476759/786615528369160212/chocomint.txt (fdfd 900 pp)
unko: https://cdn.discordapp.com/attachments/745401507553476759/786619083843567666/unko.txt
umbre: https://cdn.discordapp.com/attachments/745401507553476759/786619116785631302/umbre.txt
matrix: https://cdn.discordapp.com/attachments/745401507553476759/786619880949809163/matrix.txt
idke: https://cdn.discordapp.com/attachments/745401507553476759/786624476536111165/idke.txt
ming3012: https://cdn.discordapp.com/attachments/745401507553476759/786630236350709800/ming.txt
vaxei: https://cdn.discordapp.com/attachments/745401507553476759/786629908352466994/vaxei.txt
ryuk: https://cdn.discordapp.com/attachments/745401507553476759/786627755416485938/ryuk.txt
xexxar (me!): https://cdn.discordapp.com/attachments/745401507553476759/786619507808272384/xexxar.txt


Xexxar:
i noticed an error with the 50s function that i was given. stan must have sent the wrong one from delta's stuff. it penalizes low 50s count which it shouldnt so Im replacing the existing red curve with the blue curve.

https://www.desmos.com/calculator/tzbx9fghnl

smoogipoo (dismissed):
Sounds well reasoned

osu.Game.Rulesets.Osu/Difficulty/OsuPerformanceCalculator.cs
```diff
@@ -156,10 +156,10 @@ private double computeSpeedValue()
             if (mods.Any(m => m is OsuModHidden))
                 speedValue *= 1.0 + 0.04 * (12.0 - Attributes.ApproachRate);
 
-            // Scale the speed value with accuracy _slightly_
-            speedValue *= 0.02 + accuracy;
-            // It is important to also consider accuracy difficulty when doing that
-            speedValue *= 0.96 + Math.Pow(Attributes.OverallDifficulty, 2) / 1600;
+            // Scale the speed value with accuracy and OD
+            speedValue *= (.95 + Math.Pow(Attributes.OverallDifficulty, 2) / 750) * Math.Pow(accuracy, (14.5 - Math.Max(Attributes.OverallDifficulty, 8)) / 2);
```
peppy, line 160:
```suggestion
            speedValue *= (0.95 + Math.Pow(Attributes.OverallDifficulty, 2) / 750) * Math.Pow(accuracy, (14.5 - Math.Max(Attributes.OverallDifficulty, 8)) / 2);
```

peppy (dismissed):

smoogipoo (approved):

ginotec:
In high-aim maps, you have a small "hit time" to not break combo.
I think that acc should be afected in a negative way with aim because of that...
(what im tying to say is that a fullcombo on aim based maps is a little more easier to acc well if the map timing is correct)

PR #2289, opened 2018-03-23
Fix BeatmapCarousel's flush not correctly applying selection changes

peppy:
They may have been delayed until the next Update, which is too late in this case.

PR #11176, opened 2020-12-14
Track expected Y position of all beatmapset panels

### LINKED ISSUE #11156: Beatmap panels should have consistent scrolling animations

LittleEndu:
Closes #11156
Supersedes #11160

~~I'll write a real description if this approach is more correct/acceptable than my last one.~~

## Changes
* Calculate panel Y positions using new helper class
* Store current Y position of all beatmapset panels when filtering
  * This is done in ``updateYPositions()`` which is called on filter changes or changes to beatmap library like adding new maps
* Use stored Y position when getting new panels from the pool
  * This is the part that closes the issue as now even fresh from the pool panels have the same animation as already old ones
* Find target Y even for filtered sets
  * This is to add more consistency when filtering items. Newly visible items get a good position right away.
* Update drawable beatmapsets even if they are filtered
  * Now that they have a good target Y to travel towards this safeguard would actually make some animations, like panels leaving when they got filtered, look worse
* Fix bug in scrolling off extremes 
  * Squeezing it into this PR, is a 2 line change (``visibleHalfHeight - BleedTop/Bottom`` change)
* Remove unnecessary safeguard call to ``Update()`` in flush
  * Causes test failures otherwise because Clock won't be initialized
* Iterate over ``beatmapSets`` in ``updateYPositions()`` 
  * Carousel root already only stores ``CarouselBeatmapSets`` so the switch statement in ``updateYPositions()`` is unnecessary


osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -814,6 +821,9 @@ private void updateScrollPosition()
 
                         Scroll.ScrollTo(scrollTarget.Value, false);
 
+                        foreach (var i in root.Children)
```
peppy, line 824: this is still updating the y position of every item, right..?

LittleEndu: Yes, but this is immediate scroll only. I'll have to check, but isn't immediate scroll done only on filter changes?

peppy: Looks to be the case for now. At very least, this needs an inline comment mentioning this is the case (because it's not obvious when reading through the code).

osu.Game.Tests/Visual/SongSelect/TestSceneBeatmapCarousel.cs
```diff
@@ -748,6 +748,7 @@ private void createCarousel(Action<BeatmapCarousel> carouselAdjust = null, Conta
                 carousel = new TestBeatmapCarousel
                 {
                     RelativeSizeAxes = Axes.Both,
+                    Clock = Clock
```
peppy, line 751: This seems really weird. Are you instead able to remove whatever is dependent on this into BDL or LoadComplete?

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -729,22 +733,28 @@ private void updateYPositions()
...
-                if (item.Filtered.Value)
-                    continue;
-
                 switch (item)
                 {
                     case CarouselBeatmapSet set:
                     {
-                        visibleItems.Add(set);
+                        // find the current Y position even for filtered sets. This looks better for when the set becomes visible again as it then needs a good direction to travel from
+                        set.ExpectedYPosition = timeSinceLastYUpdate == default ? currentY : DrawableCarouselBeatmapSet.YLerp(set.CarouselYPosition, set.ExpectedYPosition, elapsed);
```
peppy, line 750: what is `default`? specify an actual number instead

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -729,22 +733,28 @@ private void updateYPositions()
...
+            double elapsed = Clock.CurrentTime - timeSinceLastYUpdate;
+
             foreach (CarouselItem item in root.Children)
             {
-                if (item.Filtered.Value)
-                    continue;
-
                 switch (item)
                 {
                     case CarouselBeatmapSet set:
```
peppy, line 747: Since you've moved the `item.Filtered` early exit inside the `switch`, i'd propose changing this to a direct cast or typed-`if` to avoid future slip-ups if other `case`s were added:

```csharp
if (!(item is CarouselBeatmapSet set)) continue;
```

or, if that isn't required:

```csharp
CarouselBetamapSet set = (CarouselBeatmapSet)item;
```

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
```diff
@@ -22,6 +22,18 @@ public class DrawableCarouselBeatmapSet : DrawableCarouselItem, IHasContextMenu
     {
         public const float HEIGHT = MAX_HEIGHT;
 
+        public static float YLerp(float targetY, float currentY, double elapsed)
```
peppy, line 25: Not a huge fan of having this static and called from outside. I think it's be more elegant if it could be non-static and update the panel itself, rather than return a value for various different use cases.

ie. it seems both use cases in `BeatmapCarousel` are operating  on the panel itself, so potentially all of that logic could be moved internal to this class (or the corresponding model) instead?

peppy (changes requested):

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: Right now carousel root only holds beatmap sets so checking this isn't required. I would change the foreach to happen over beatmapSets, avoided that because it would unnest this entire switch causing unnecessary whitespace changes in the diff view.

peppy: Whitespace changes are the least of our concerns ever.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
LittleEndu: > ie. it seems both use cases in ``BeatmapCarousel`` are operating on the panel itself

That turns out to actually not be the case. In ``Update()`` it's giving an initial position to the drawable, but in ``updateYPositions()`` it's operating on the ``CarouselBeatmapSet`` and not the drawable. 

The reason that I think this should be one function is exactly that items and their drawable representations need to be in-sync so that no weird edge cases can happen. If a copy of this function is moved both to ``CarouselBeatmapSet`` and ``DrawableCarouselBeatmapSet`` there's no guarantee somebody won't change one function and forget to change the other.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -426,7 +426,8 @@ public void FlushPendingFilterOperations()
             if (PendingFilter?.Completed == false)
             {
                 applyActiveCriteria(false);
-                Update();
+                if (IsLoaded) // Update requires Clock to be initialized
+                    Update();
```
LittleEndu, line 430: Actually, this code is 3 years old (from https://github.com/ppy/osu/pull/2289)

There's a test (``TestScenePlaySongSelect.TestChangeBeatmapBeforeEnter``) that's testing that exact behaviour and it doesn't fail when I remove this line... so I think it's safe to remove it.

osu.Game/Screens/Select/Carousel/CarouselItem.cs
```diff
@@ -15,6 +15,8 @@ public abstract class CarouselItem : IComparable<CarouselItem>
         /// </summary>
         public float CarouselYPosition;
 
+        public float ExpectedYPosition;
```
bdach, line 18: what is `ExpectedYPosition` and how does it differ from `CarouselYPosition`? would like an xmldoc on the former and maybe a rename of the latter that explains the difference better.

osu.Game/Screens/Select/BeatmapCarousel.cs
bdach: it might be 3 years old, but it has never depended on an actual readout of the clock being correct before.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
bdach: if so, it should be in a separate helper class somewhere, or in the model.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -729,51 +734,51 @@ private void updateYPositions()
...
             visibleItems.Clear();
 
-            float currentY = visibleHalfHeight;
+            // visibleHalfHeight includes BleedTop, but Scroll screenspace doesn't, so we remove it here
+            float currentY = visibleHalfHeight - BleedTop;
 
             scrollTarget = null;
 
-            foreach (CarouselItem item in root.Children)
+            double elapsed = Clock.CurrentTime - timeSinceLastYUpdate;
```
bdach, line 742: this reads bogus. if this intends to be a time delta, it should be `timeOfLastYUpdate`, not `timeSinceLastYUpdate`.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -729,51 +734,51 @@ private void updateYPositions()
...
+
+            foreach (CarouselBeatmapSet set in beatmapSets)
             {
-                if (item.Filtered.Value)
+                // find the current Y position even for filtered sets. This looks better for when the set becomes visible again as it then needs a good direction to travel from
+                set.ExpectedYPosition = timeSinceLastYUpdate == 0 ? currentY : DrawableCarouselBeatmapSet.YLerp(set.CarouselYPosition, set.ExpectedYPosition, elapsed);
```
bdach, line 747: what is this 0 check for? is it for checking if `timeSinceLastUpdate` has been set before? if so, then use a `double?` instead of representing that information in-band like this.

osu.Game/Screens/Select/BeatmapCarousel.cs
```diff
@@ -621,7 +624,8 @@ protected override void Update()
                         var panel = setPool.Get(p => p.Item = item);
 
                         panel.Depth = item.CarouselYPosition;
-                        panel.Y = item.CarouselYPosition;
+                        // place the panel at current Y by accounting for the time the panel has spent not updated yet
+                        panel.Y = DrawableCarouselBeatmapSet.YLerp(item.CarouselYPosition, item.ExpectedYPosition, elapsed);
```
bdach, line 628: why is nothing happening with `CarouselYPosition` here? isn't it getting out-of-sync?

bdach (changes requested):
utterly struggling to understand this diff and there's still no proper description

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: Renaming it to ``TargetYPosition`` might be the best.

We have already calculated our target Y in ``updateYPositions`` and this won't change until another filtering has happened. This is also just the initial placement for the panel, the drawable itself will keep moving towards the target Y until it is reached.

bdach: so now it's `Target` and `Expected`? very confusing...

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: Yeah, but I'm thinking that removing this should be fine.

osu.Game/Screens/Select/BeatmapCarousel.cs
LittleEndu: ``Target`` and ``Current`` make most since, but ``Current`` is only update on filter so it's not really current.

bdach: then maybe go descriptive? `YPositionOnLastFilter`? mind you that doesn't help this diff *that* much in terms of readability because the logic is half-split between that field and the actual drawable's `Y` anyway...

the overall point is to be able to tell from a glance on the names which value is the stale one and which one are we going towards.

osu.Game/Screens/Select/BeatmapCarousel.cs
bdach: possibly? not sure. enough stuff has changed that it's plausible that guard is not needed anymore. tried a few times to use the terse description of the repro scenario in [this comment](https://github.com/ppy/osu/pull/2289#issuecomment-375599378) but saw no discernible breakage.

osu.Game/Screens/Select/Carousel/DrawableCarouselBeatmapSet.cs
LittleEndu: Can the function stay static in the helper class? What would even be a good name for such a class, ``CarouselHelper`` or ``CarouselUtils`` or something?

bdach: it should be even probably a static class entirely. as for the name, either works for me, really.

LittleEndu:
Added a description to the PR. Also recorded some videos (github should support these now 🎉) of behaviour to look out for


https://user-images.githubusercontent.com/11320483/102536205-f6dd0100-40b1-11eb-907c-b7c20b7f2e19.mp4


https://user-images.githubusercontent.com/11320483/102536338-268c0900-40b2-11eb-8c35-96e807e437f5.mp4


https://user-images.githubusercontent.com/11320483/102536461-520ef380-40b2-11eb-8089-c3cb967f1b6b.mp4


https://user-images.githubusercontent.com/11320483/102536482-5804d480-40b2-11eb-9846-dd37a242f937.mp4

peppy:
Not actually sure all of these are better than what we have on master.. the cases where there is now a huge delay waiting for repositioning, rather than getting instant updates particularly. Will need to do a bit of testing to decide if this is the direction we want to go.

LittleEndu:
> Will need to do a bit of testing to decide [...]

I'm not sure what testing you meant but I would like to ask if this PR should be closed and a discussions thread around carousel animations be opened instead.

> the cases where there is now a huge delay waiting [...]

This is probably because I have ~5400 beatmaps loaded. And even if it wasn't the case, it should be the animation that gets reworked/sped up, not the direction the animation is played from which I'm trying to make consistent in this PR.

Also of note if you merge master into this PR, you also get the horrible performance carousel has on net5 (when many beatmaps are loaded)

peppy:
I'd recommend closing for now as I don't believe this is going to get merged any time soon. A good starting point would be to agree on how behaviour should be improved first, and then make a PR afterwards (to avoid upkeep as the game changes around it).

LittleEndu:
Closing and moving discussion to #13220

PR #11182, opened 2020-12-16
Add browser-like tab management hotkeys to chat overlay

### LINKED ISSUE #10148: make some browser tab shortcuts work for osu chat tabs

angelaz1:
Fixes #10148

- [x] Depends on #13314.

Adding in web browser-like keyboard shortcuts to the chat overlay:
- Closing a single chat tab with Ctrl + w 
- Pulling up the channel selector overlay with Ctrl + t to open a new tab
- Reopening the most recently closed tab that has not already been reopened Ctrl+Shift+t

Co-Authors: @jessmeng @Joseph-Ramos-CMU


osu.Game.Tests/Visual/Online/TestSceneChatOverlay.cs
```diff
@@ -208,6 +208,60 @@ public void TestChannelCloseButton()
             AddAssert("All channels closed", () => !channelManager.JoinedChannels.Any());
         }
 
+        private void pressControlW()
```
peppy, line 211: private methods should go beneath public ones.

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -33,6 +33,14 @@ public class ChannelManager : PollingComponent, IChannelPostTarget
         private readonly BindableList<Channel> availableChannels = new BindableList<Channel>();
         private readonly BindableList<Channel> joinedChannels = new BindableList<Channel>();
 
+        // Keeps a list of closed channels. More recently closed channels appear at higher indeces
+        private readonly BindableList<Channel> closedChannels = new BindableList<Channel>();
```
peppy, line 37: This should be only storing the name or ID, not the full channel object. This is going to be keeping messages, users, etc. in memory forever.

peppy (changes requested):

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -33,6 +33,14 @@ public class ChannelManager : PollingComponent, IChannelPostTarget
...
+
+        // For efficiency purposes, this constant bounds the number of closed channels we store.
+        // This number is somewhat arbitrary; future developers are free to modify it.
+        // Must be a positive number.
+        private const int closed_channels_max_size = 50;
```
bdach, line 42: We actually have an existing utility collection that does this in this repository (see `LimitedCapacityQueue`). It would have to be moved out from its current namespace first, though, so unsure if worth - and even then maybe best left for a separate pull.

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -408,13 +416,51 @@ public void LeaveChannel(Channel channel)
...
+
+            Channel lastClosedChannel = closedChannels.Last();
+
+            closedChannels.Remove(lastClosedChannel);
+
+            // If the user already joined the channel, try the next
+            // channel in the list
+            if (joinedChannels.IndexOf(lastClosedChannel) >= 0)
+            {
+                JoinLastClosedChannel();
```
bdach, line 456: I'm not really comfortable with using recursion here, especially seeing that the worst case depth is 50 calls, and an iterative solution would work just fine.

As an aside, you could avoid this edge case whatsoever by checking if any new channel joined had been previously closed, and removing it from the list in that case. That would probably not play well with `LimitedCapacityQueue`, however, as it lacks a random access removal operation.

osu.Game/Overlays/ChatOverlay.cs
```diff
@@ -357,6 +357,32 @@ protected override bool OnKeyDown(KeyDownEvent e)
                 }
             }
 
+            if (e.ControlPressed)
+            {
+                if (e.ShiftPressed)
+                {
+                    switch (e.Key)
+                    {
+                        case Key.T:
```
bdach, line 366: We have a framework-side mechanism called platform actions that might be good to use here, as tab operations are pretty well defined UI interactions overall. I can prepare a pull adding those.

bdach (changes requested):
One big issue (the recursion usage); rest is mostly code quality stuff.

osu.Game/Online/Chat/ChannelManager.cs
Joseph-Ramos-CMU: Thanks for the feedback. I'll fix it to get rid of the recursion. As for avoiding the edge case, I thought about adding some deduplicate code. However, this code would have to be run whenever the user closed a tab, which could make closing a tab slow should the closedChannel list get large. I intentionally saved the checks for when the user actually used the Ctrl+Shift+t shortcut, this way the new feature would have minimal impact on the performance of any other part of the code.

bdach: I mean I don't think it would be *that* much worse than the current code. Let's denote the number of previously-closed channels with `n`, and the number of currently-joined channels with `m`.

Current code:

* An item is added to `closedChannels` on every channel close, which is an amortised `O(1)` operation - amortised because the list has to be expanded in `O(n)` time every now and again, but as far as I know .NET uses the proper exponential reallocation scheme to achieve amortised `O(1)`. (As an aside, `closedhannels` is not currently checked for duplicates when added to, which seems unintended.)
* Each attempt to restore a channel is an `O(m)` operation (worst case for the `IndexOf()` to check if the channel we're trying to restore is open), and there can be up to `n` attempts, yielding a total of `O(n * m)`.

What I'm proposing:

* Adding to `closedChannels` would be the same as in the current code. (Preferably with duplicate checks.)
* When joining a channel, we'd have to check closed channels to make sure we clean them up, which is an `O(n)` operation. Joining `m` channels would therefore be `O(n * m)` again (but in `m` different operations instead of one).
* Restoring the last closed channel would be `O(1)` again, as the only thing to be done is to rejoin the last closed-but-not-reopened channel from `closedChannels`.

I guess the argument could be made that joining channels is a more frequent operation than restoring the last closed one, but I'm not sure `m` and `n` will ever get large enough where that will be a problem. That said I don't insist on this solution - I'm fine with just removing the recursion.

bdach:
On reading this diff again I noticed that the <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>T</kbd> hotkey is not tested. It would be nice to have coverage for this as well.

Joseph-Ramos-CMU:
I believed we've addressed all the critical changes you requested. As for test coverage for Ctrl+Shift+t, we tried adding tests but were having a lot of issues with Appveyor. The commit history should show some of the tests we tried. Pretty much followed the same format as the Ctrl+W and Ctrl+T tests, but it wasn't working for some reason. It might be because the Ctrl+Shift+t shortcut only works if the user has the chat tabs selected, but even after updating the tests to account for this we still had issues.

Let us know if there are any other must-do fixes.

osu.Game/Online/Chat/ChannelManager.cs
```diff
@@ -408,13 +419,72 @@ public void LeaveChannel(Channel channel)
...
+                    if (lastClosedChannel != null)
+                    {
+                        CurrentChannel.Value = JoinChannel(lastClosedChannel);
+                    }
+                    else
+                    {
+                        // Try to get User to open PM chat
+                        var req = new GetUserRequest(lastClosedChannelId);
+                        req.Success += user => CurrentChannel.Value = JoinChannel(new Channel(user));
+                        api.Queue(req);
```
bdach, line 480: This will unnecessarily re-fetch the user every time. The proper solution for this is to use `UserLookupCache` which will 100% have the user locally fetched already.

bdach (changes requested):

osu.Game/Overlays/ChatOverlay.cs
peppy: @bdach Shall we get this done framework side and try and merge this PR?

Should probably be added for the "close tab" key below as well (ctrl-w on windows; cmd-q on macOS).

bdach: I already have added those actions, just never got round to updating this pull with them as it kept falling down the backlog.

peppy: I've made the required changes here. @bdach can you give this a once-over whenever you have a chance?

bdach:
There's still some weirdness left in this PR that I hadn't seen or had time to address properly yet. Going to try and clean this up more myself now.

bdach:
Have pushed my changes, mostly cosmetic. I did spot one more issue, namely that when the channel selector overlay is open, the hotkeys stop working - I think it's because the channel selector is taking  focus away from the chat overlay, so I'm not sure that's easily fixable. Other than that seems passable.

@peppy feel free take a look at the latest changes if you're so inclined :)

bdach (approved):

PR #11183, opened 2020-12-16
Add button to close all open tabs

### LINKED ISSUE #10540: Opening chat opens all previous conversations
Opening chat by pressing `F8` automatically opens all previous conversations. Each conversation has to be deleted manually, as there is no button in the context menu to close all of them at once. When you've been playing the game for a while this can easily mean hundreds of open tabs. (or are a tournament host sending messages to dozens of people)

video: https://youtu.be/GU6-UCGT5XQ
osu!lazer version: 2020.1017.0

angelaz1:
Fixes #10540

Adds a "Close All" button to the chat overlay, allowing the user to close all open chat tabs


peppy:
Before reviewing this, have you tested this and how it behaves with the API? Sounds like it's potentially going to be firing tens or hundreds of requests which may create a large backlog (ie. we may want an API endpoint to close all chat conversations).

angelaz1:
That's a good point. We mostly worked on this with some small test cases (mostly testing if the functionality works), but I agree that it's currently quite inefficient as it essentially just sends out requests to close each channel one at a time.

peppy:
I'd say we definitely want to figure out the correct UX for this before implementing it. The current button is *way* too visible for something like this (especially since there's no confirmation - this is a destructive operation that would be seldom used so it definitely shouldn't be shown in the tab bar itself).

bdach:
Browsers/tabbed editors usually have expanded tabbing operation options under the right click menu. Going off of that, gonna throw out a loose suggestion - how about something like this (or an analogue of this but done on the tab headers themselves)?

2020-12-16-194859_232x141_scrot

Very janky mock-up - probably needs better discoverability as in my quick-and-dirty version there's zero (icon is not interactable otherwise).

peppy:
Or just including it at the bottom of the overflow dropdown? There's generally no reason to use a "close all tabs" button if the tab bar isn't even full.

frenzibyte:
Given the conflicts and the date of this PR, it's probably for the best closing this and starting fresh after https://github.com/ppy/osu-web/issues/7007 is resolved and the new redesigned chat overlay is in place.

PR #11185, opened 2020-12-16
Implement a spectator data driven leaderboard

peppy:
This is a simple (dynamic) leaderboard implementation which is backed by spectator data. To be used in the real-time multiplayer implementation.

I've tested this with live data and it works well enough (albeit existing users are not sending header information yet, so scores are not correctly handled). Will probably need some adjustments when it is integrated with the multiplayer system (as discussed, it may accept `User` instances as its input, rather than raw IDs).

- [x] Depends on #11171
- [x] Depends on https://github.com/ppy/osu/pull/11191


osu.Game/Screens/Play/HUD/MultiplayerGameplayLeaderboard.cs
```diff
@@ -0,0 +1,125 @@
...
+        private void load(OsuConfigManager config)
+        {
+            streamingClient.OnNewFrames += handleIncomingFrames;
+
+            foreach (var user in userIds)
+            {
+                streamingClient.WatchUser(user);
+
+                // probably won't be required in the final implementation.
+                var resolvedUser = userLookupCache.GetUserAsync(user).Result;
```
bdach, line 55: Slightly alarmed by the `.Result`. Not saying this will necessarily deadlock (and I haven't thought very hard if it can), but may be wise to at least stick a `[LongRunningLoad]` on this for the time being if we're going with this simplification for now?

osu.Game.Tests/Visual/Gameplay/TestSceneMultiplayerGameplayLeaderboard.cs
```diff
@@ -0,0 +1,136 @@
...
+        public void SetUp() => Schedule(() =>
+        {
+            OsuScoreProcessor scoreProcessor;
+
+            streamingClient.Start(Beatmap.Value.BeatmapInfo.OnlineBeatmapID ?? 0);
+
+            Children = new Drawable[]
+            {
+                streamingClient,
+                lookupCache,
```
bdach, line 42: If `SetUp()` runs more than once, these are getting disposed the second time this `Children` assignment runs (as it's a shorthand for `Clear()` followed by an `AddRange()`), causing the other test failure.

bdach (changes requested):

bdach:
As an aside, the design for the leaderboard is not the final one, right? It's pretty janky as-is - no rolling on the counters, text is aligned to the left which looks quite off if the score magnitudes differ (as seen in the gif in the OP; I'd probably say it should be right-aligned for that reason).

osu.Game/Screens/Play/HUD/MultiplayerGameplayLeaderboard.cs
peppy: Only this way with the comment above attached to it. As discussed with smoogi, the users will likely be provided in the ctor rather than raw IDs, at which point this code will be removed.

peppy:
The design is the "current" design. If we implement the new one it wouldn't match the rest of the UI (as flyte is 8 iterations ahead as always). Lack of rolling is intentional for now (stable also does this) to reduce visual distraction. We may add more animations going forwards but the idea here is to get a leaderboard in game before the deadline is hit for getting multiplayer live, not having the ultimate visual experience.

osu.Game/Rulesets/Scoring/ScoreProcessor.cs
```diff
@@ -227,6 +232,34 @@ public double GetScore(ScoringMode mode, int maxCombo, double accuracyRatio, dou
...
         }
 
+        /// <summary>
+        /// Given a minimal set of inputs, return the computed score and accuracy for the tracked beatmap / mods combination.
+        /// </summary>
+        /// <param name="mode">The <see cref="ScoringMode"/> to compute the total score in.</param>
+        /// <param name="maxCombo">The maximum combo achievable in the beatmap.</param>
+        /// <param name="statistics">Statistics to be used for calculating accuracy, bonus score, etc.</param>
+        /// <returns>The computed score and accuracy for provided inputs.</returns>
+        public (double score, double accuracy) GetScoreAndAccuracy(ScoringMode mode, int maxCombo, Dictionary<HitResult, int> statistics)
```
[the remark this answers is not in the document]
peppy: Hmm, that's weird (sounds like an oversight). It is required for calculating score.

osu.Game/Rulesets/Scoring/ScoreProcessor.cs
```diff
@@ -68,7 +68,12 @@ public class ScoreProcessor : JudgementProcessor
         private readonly double comboPortion;
 
         private int maxAchievableCombo;
-        private double maxBaseScore;
+
+        /// <summary>
+        /// The maximum achievable base score.
+        /// </summary>
+        public double MaxBaseScore { get; private set; }
```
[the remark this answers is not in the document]
peppy: No, this can be reverted. Was from a previous implementation.

bdach:
test case from this PR exposes a shortcoming of the new design wrt text truncation:

2020-12-19-184039_402x161_scrot

tempted to just fix locally instead of making a separate pull for this...

osu.Game.Tests/Visual/Gameplay/TestSceneMultiplayerGameplayLeaderboard.cs
```diff
@@ -0,0 +1,155 @@
...
+            }
+
+            private readonly Dictionary<int, FrameHeader> lastHeaders = new Dictionary<int, FrameHeader>();
+
+            public void RandomlyUpdateState()
+            {
+                foreach (var userId in PlayingUsers)
+                {
+                    if (RNG.Next(0, 1) == 1)
+                        continue;
```
bdach, line 113: This is actually always true right now, because `RNG.Next` is max-exclusive. I'm going to assume this meant to be a 50-50 split and change this to a `RNG.NextBool()` as I have a few other changes queued up anyway.

bdach:
I've applied a couple of minor fix-ups - looks good to me now.

@peppy would appreciate if you looked through my changes and make sure you're OK wth them!

bdach (approved):

osu.Game.Tests/Visual/Gameplay/TestSceneMultiplayerGameplayLeaderboard.cs
peppy: i was wondering why something felt off with the test display!

PR #11190, opened 2020-12-17
Add global friends list to API providers

frenzibyte:
Exposes a `BindableList` of the current user's friends, this allows components to determine whether a certain user is followed by the local user.

### DIFF
--- a/osu.Game/Online/API/APIAccess.cs
+++ b/osu.Game/Online/API/APIAccess.cs
@@ -41,6 +41,8 @@ namespace osu.Game.Online.API
 
         public Bindable<User> LocalUser { get; } = new Bindable<User>(createGuestUser());
 
+        public BindableList<User> Friends { get; } = new BindableList<User>();
+
         public Bindable<UserActivity> Activity { get; } = new Bindable<UserActivity>();
 
         protected bool HasLogin => authentication.Token.Value != null || (!string.IsNullOrEmpty(ProvidedUsername) && !string.IsNullOrEmpty(password));
@@ -142,6 +144,10 @@ namespace osu.Game.Online.API
                             LocalUser.Value.Status.Value = new UserStatusOnline();
 
                             failureCount = 0;
+
+                            var friendsReq = new GetFriendsRequest();
+                            friendsReq.Success += f => Friends.AddRange(f);
+                            handleRequest(friendsReq);
 
                             //we're connected!
                             state.Value = APIState.Online;
@@ -352,8 +358,12 @@ namespace osu.Game.Online.API
             password = null;
             authentication.Clear();
 
-            // Scheduled prior to state change such that the state changed event is invoked with the correct user present
-            Schedule(() => LocalUser.Value = createGuestUser());
+            // Scheduled prior to state change such that the state changed event is invoked with the correct user and their friends present
+            Schedule(() =>
+            {
+                LocalUser.Value = createGuestUser();
+                Friends.Clear();
+            });
 
             state.Value = APIState.Offline;
         }
--- a/osu.Game/Online/API/APIAccess.cs
+++ b/osu.Game/Online/API/APIAccess.cs
@@ -140,6 +140,7 @@ namespace osu.Game.Online.API
                         }
 
                         var userReq = new GetUserRequest();
+
                         userReq.Success += u =>
                         {
                             localUser.Value = u;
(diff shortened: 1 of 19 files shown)


osu.Game/Online/API/APIAccess.cs
```diff
@@ -143,8 +145,16 @@ private void run()
...
 
-                            //we're connected!
-                            state.Value = APIState.Online;
+                            var friendsReq = new GetFriendsRequest();
+                            friendsReq.Success += f =>
+                            {
+                                Friends.AddRange(f);
+
+                                //we're connected!
+                                state.Value = APIState.Online;
```
bdach, line 154: so if the user request succeeds but the friends request fails, the user is no longer going to go online? i don't think that's a good idea.

osu.Game/Overlays/OverlayView.cs
```diff
@@ -42,25 +42,29 @@ private void load()
         /// <summary>
         /// Create the API request for fetching data.
         /// </summary>
-        protected abstract APIRequest<T> CreateRequest();
+        protected virtual APIRequest<T> CreateRequest() => null;
```
bdach, line 45: this is actively a regression in code quality of this component. now it's easy for implementors of `OverlayView` to forget this has to be overridden.

i think the way to go forward is to remove `FriendDisplay`'s inheritance of `OverlayView` altogether. it does not fit the overall structure with this change at all, and all that's left after removing `CreateRequest()` and all associated logic is just an `APIStateChanged` subscription. it's so little code that it might as well be "duplicated" in that class locally.

bdach (changes requested):

osu.Game/Online/API/APIAccess.cs
frenzibyte: Hmm, not entirely sure how fine it is for a user to not have its friends list retrieved successfully though, that wouldn't make the user fully online to the game?

bdach: switching to `Failing` sounds like a better solution for sure.

on that note i'd probably split the friends request to a method or something, as nested success callbacks make me uneasy. personal opinion though.

osu.Game/Online/API/APIAccess.cs
```diff
@@ -255,6 +253,19 @@ public RegistrationRequest.RegistrationRequestErrors CreateAccount(string email,
             return null;
         }
 
+        private void fetchFriends(Action onSuccess, Action onFail)
```
bdach, line 256: these callbacks seem sort of overdone to be honest, i'd probably just inline.

frenzibyte: setting `State` to `Failing` inside this method? don't think that'll be a good idea given that the method is allowed to be called anywhere from the whole class.

could just revert this extraction and call it a day.

bdach: arguing that a private method has to be safely callable from any point in its class does not make a lot of sense to me.

i'm just saying it's the only usage for now, no reason to overcomplicate matters. if someone doesn't look at what methods they're calling and if they can call them in the first place without breaking stuff, they probably shouldn't be programming in the first place.

bdach: arguing that a private method has to be safely callable from any point in its class does not make a lot of sense to me.

i'm just saying it's the only usage for now, no reason to overcomplicate matters. if someone doesn't look at what methods they're calling and if they can call them in the first place without breaking stuff, they probably shouldn't be programming in the first place.

peppy: I've moved this in locally and un-nested for better flow.

osu.Game/Online/API/DummyAPIAccess.cs
```diff
@@ -18,6 +19,12 @@ public class DummyAPIAccess : Component, IAPIProvider
             Id = 1001,
         });
 
+        public BindableList<User> Friends { get; } = new BindableList<User>(new User
```
peppy, line 22: Where is this visible in game? Or is it here for future use? Feels a bit weird.

peppy:
I've made all the remaining non-`IBindables` into them. I've also modified the flow of the connection process slightly.

@smoogipoo would appreciate a quick review of this.

peppy (dismissed):

smoogipoo (dismissed):

PR #11191, opened 2020-12-17
Update gameplay leaderboard scores in-line with the new design

frenzibyte:
- [x] Depends on #11190 

New design look with the multiplayer leaderboard's test scene from #11185 (upper half):


osu.Game/Screens/Play/HUD/GameplayLeaderboard.cs
```diff
@@ -29,32 +28,44 @@ public GameplayLeaderboard()
...
-        /// <param name="currentScore">The bindable current score of the player.</param>
+        /// <param name="totalScore">A bindable of the player's total score.</param>
+        /// <param name="accuracy">A bindable of the player's accuracy.</param>
+        /// <param name="combo">A bindable of the player's current combo.</param>
         /// <param name="user">The player.</param>
-        public void AddPlayer([NotNull] BindableDouble currentScore, [NotNull] User user)
+        public void AddPlayer([NotNull] IBindableNumber<double> totalScore,
+                              [NotNull] IBindableNumber<double> accuracy,
+                              [NotNull] IBindableNumber<int> combo,
+                              [NotNull] User user)
```
bdach, line 38: i'd just make `GameplayLeaderboardScore` a non-nested public model class at this point and just pass that in instead of jamming all 4 of its constituent members through 2 levels of calls.

could also make a ctor that just takes the user in, and initialises all other bindables to sane initial values. would clean up the test helpers.

osu.Game/Screens/Play/HUD/GameplayLeaderboard.cs
```diff
@@ -29,32 +28,44 @@ public GameplayLeaderboard()
...
+        /// </summary>
+        /// <param name="totalScore">A bindable of the player's total score.</param>
+        /// <param name="accuracy">A bindable of the player's accuracy.</param>
+        /// <param name="combo">A bindable of the player's current combo.</param>
+        /// <param name="user">The player.</param>
+        /// <param name="localUser">Whether the provided <paramref name="user"/> is the local user.</param>
+        protected void AddPlayer([NotNull] IBindableNumber<double> totalScore,
+                                 [NotNull] IBindableNumber<double> accuracy,
+                                 [NotNull] IBindableNumber<int> combo,
+                                 [NotNull] User user, bool localUser) => Schedule(() =>
```
bdach, line 54: does `localUser` have to get dragged through all these calls? `GameplayLeaderboardScore` already BDLs in an `IAPIProvider` - any reason not to use that and do an equality check there against `LocalUser`?

bdach (changes requested):

osu.Game/Screens/Play/HUD/GameplayLeaderboard.cs
frenzibyte: Yeah, that was a misleading name, it's not `localUser`, but more of the `localPlayer`, as in the case of autoplay mods, the local player would be the autoplay's score "user".

osu.Game/Screens/Play/HUD/GameplayLeaderboard.cs
frenzibyte: Sure that'll work.

osu.Game/Screens/Play/HUD/GameplayLeaderboard.cs
bdach: i don't understand how a `localUser` is different from a `localPlayer`. if it's related to autoplay (and only autoplay) then i'd just resolve the mods bindable and just check that.

frenzibyte: there's also the case of spectating, the player the user is spectating would be the `localPlayer` (though that makes the param name even more wrong, what do you think?)

bdach: does that also apply to replays? if so i'd just go for `replayPlayer`.

frenzibyte: I've renamed it to `localOrReplayPlayer` for counting the local user in.

peppy:
This is no longer dependent on #11190 but I'm just going to leave it in there as it's too much of a pain to decouple the commits now.

I've done a fair bit of refactoring based on missing design information (ie. what colour represents what). @smoogipoo requesting a re-review here.

peppy (dismissed):

peppy:
Don't review/merge yet, there's a huge regression in data flow.

peppy:
Should be good now.

peppy (dismissed):

frenzibyte:
Featuring user avatar now ;)

peppy (dismissed):
Avatar implementation looks okay.

bdach:
I've piled some of my own changes. They're all minor and (hopefully) uncontroversial, and I've checked that none of them should conflict with the dependent PR.

Seems good to me otherwise.

bdach (approved):

bdach:
Seems like info about the appveyor build pass (https://ci.appveyor.com/project/peppy/osu/builds/36920793) didn't reach github somehow, so can't merge on my own.

PR #11196, opened 2020-12-18
Align selection box circles to center of border

Grrum:
… center of the border

Just a visual improvement in the editor. Before vs after:
circleBefore
circleAfter

### DIFF
--- a/osu.Game/Screens/Edit/Compose/Components/SelectionBox.cs
+++ b/osu.Game/Screens/Edit/Compose/Components/SelectionBox.cs
@@ -2,7 +2,6 @@
 // See the LICENCE file in the repository root for full licence text.
 
 using System;
-using System.Collections.Generic;
 using osu.Framework.Allocation;
 using osu.Framework.Graphics;
 using osu.Framework.Graphics.Containers;
@@ -93,6 +92,7 @@ namespace osu.Game.Screens.Edit.Compose.Components
             }
         }
 
+        private Container dragHandles;
         private FillFlowContainer buttons;
 
         public const float BORDER_RADIUS = 3;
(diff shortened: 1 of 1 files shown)


peppy (dismissed):
I think this is fine.

peppy:
@smoogipoo leaving this for you to double-check, in case you have a better idea for how it should be implemented (I can't think of one off the top of my head though).

peppy (dismissed):

Grrum:
I thought of a better way. Use a dictionary.

bdach:
I've come up with yet another solution to this that I consider the cleanest.

osu.Game/Screens/Edit/Compose/Components/SelectionBox.cs
```diff
@@ -235,13 +236,27 @@ private void addButton(IconUsage icon, string tooltip, Action action)
...
+            Y = dragCircleAdjustments[anchor].Y,
             HandleDrag = e => OnScale?.Invoke(e.Delta, anchor),
             OperationStarted = operationStarted,
             OperationEnded = operationEnded
         });
 
+        /// <summary>
+        /// Adjust Drag circle to be centered on the center of the border instead of on the edge.
+        /// </summary>
+        private Dictionary<Anchor, Vector2> dragCircleAdjustments = new Dictionary<Anchor, Vector2>(){
```
peppy: i honestly think the last version you had was better. would recommend a revert.

bdach: To clarify: do you think the dictionary was better than the padding trick I used? Not 100% sure how to interpret this comment.

peppy: Oh, somehow missed your commit. That is indeed much better.

peppy (approved):

PR #11199, opened 2020-12-18
Asyncify player score creation and submission

smoogipoo:
The main purpose of this PR is to asyncify the score submission process and have the completion progress delegate wait for submission to complete before continuing to the results screen.

In realtime multiplayer, score submission includes setting the `FinishedPlay` state and waiting for a `ResultsReady` event from the server.

I've added waiting for the timeshift request to complete in this PR.

### DIFF
--- a/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
+++ b/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
@@ -107,25 +107,23 @@ namespace osu.Game.Screens.Multi.Play
         {
             Debug.Assert(token != null);
 
-            bool completed = false;
+            var tcs = new TaskCompletionSource<bool>();
             var request = new SubmitRoomScoreRequest(token.Value, roomId.Value ?? 0, playlistItem.ID, score.ScoreInfo);
 
             request.Success += s =>
             {
                 score.ScoreInfo.OnlineScoreID = s.ID;
-                completed = true;
+                tcs.SetResult(true);
             };
 
             request.Failure += e =>
             {
                 Logger.Error(e, "Failed to submit score");
-                completed = true;
+                tcs.SetResult(false);
             };
 
             api.Queue(request);
-
-            while (!completed)
-                await Task.Delay(100);
+            await tcs.Task;
 
             return await base.SubmitScore(score);
         }
--- a/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
+++ b/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
@@ -103,8 +103,10 @@ namespace osu.Game.Screens.Multi.Play
             return score;
         }
 
-        protected override async Task<ScoreInfo> SubmitScore(Score score)
+        protected override async Task SubmitScore(Score score)
         {
+            await base.SubmitScore(score);
+
             Debug.Assert(token != null);
 
             var tcs = new TaskCompletionSource<bool>();
@@ -124,8 +126,6 @@ namespace osu.Game.Screens.Multi.Play
 
             api.Queue(request);
             await tcs.Task;
-
-            return await base.SubmitScore(score);
         }
 
         protected override void Dispose(bool isDisposing)
--- a/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
+++ b/osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
@@ -105,16 +105,29 @@ namespace osu.Game.Screens.Multi.Play
 
         protected override async Task<ScoreInfo> SubmitScore(Score score)
         {
-            await base.SubmitScore(score);
-
             Debug.Assert(token != null);
 
+            bool completed = false;
             var request = new SubmitRoomScoreRequest(token.Value, roomId.Value ?? 0, playlistItem.ID, score.ScoreInfo);
-            request.Success += s => score.ScoreInfo.OnlineScoreID = s.ID;
-            request.Failure += e => Logger.Error(e, "Failed to submit score");
+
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
```diff
@@ -95,19 +96,38 @@ protected override ResultsScreen CreateResults(ScoreInfo score)
...
+            request.Failure += e =>
+            {
+                Logger.Error(e, "Failed to submit score");
+                completed = true;
+            };
+
             api.Queue(request);
 
-            return score;
+            while (!completed)
```
peppy, line 127: is there a reason for using `Task.Delay` rather than a manual reset event?

osu.Game/Screens/Play/Player.cs
```diff
@@ -748,39 +759,67 @@ public override bool OnExiting(IScreen next)
...
 
-            if (recordingScore?.Replay.Frames.Count > 0)
-            {
-                score.Replay = recordingScore.Replay;
+        /// <summary>
+        /// Submits the player's <see cref="Score"/>.
+        /// </summary>
+        /// <param name="score">The <see cref="Score"/> to submit.</param>
+        /// <returns>The submitted score.</returns>
+        protected virtual async Task<ScoreInfo> SubmitScore(Score score)
```
peppy, line 799: Bit confused by this being called `SubmitScore` but instead running import code. Might be worth moving the replay import code to its own method and having the base method empty here?

peppy (changes requested):

osu.Game/Screens/Play/Player.cs
smoogipoo: Yeah I knew this was going to come up. I struggled with this too. So you're suggesting to split it out into `ImportScore` and `SubmitScore`?

osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
smoogipoo: I don't want to be completely blocking the threadpool thread via an MRE, but I've used a TaskCompletionSource instead since it's awaitable.

osu.Game/Screens/Play/Player.cs
smoogipoo: I've split it out into two methods. I like this structure more if only to not worry about ordering of `base` calls - `TimeshiftPlayer` previously awaited `base.SubmitScore()` at the very end to order the import process such that it has an online score id.

bdach (commented):
Looks mostly fine to me, just a few low-impact notes.

osu.Game/Screens/Play/Player.cs
```diff
@@ -527,33 +531,50 @@ private void updateCompletionState(ValueChangedEvent<bool> completionState)
...
-                Mods = Mods.Value.ToArray(),
-            };
-
-            if (DrawableRuleset.ReplayScore != null)
-                score.User = DrawableRuleset.ReplayScore.ScoreInfo?.User ?? new GuestUser();
-            else
-                score.User = api.LocalUser.Value;
-
-            ScoreProcessor.PopulateScore(score);
+                scheduleCompletion();
```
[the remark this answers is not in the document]
peppy: I personally preferred the schedule call where it was in the original code (is a bit hidden here if you're not looking carefully).

Also, just for clarity what's the state where the completion state can go from true to false? Is that actually something we need/want to be covering?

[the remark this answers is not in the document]
bdach: The way I interpreted this was that this schedule just repeats every frame until the underlying task completes. I don't think it can feasibly ever go from true to false, it'll just be false until (in the timeshift case) the underlying task completion source's result is set.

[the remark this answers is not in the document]
peppy: Specifically wonder about this logic:

https://github.com/ppy/osu/pull/11199/files#diff-2711e888291c0df6e400a5c169b90fabe525f3aa8ffe88734eeea89504abb7f2L511-L516

and what it's there for. But based off there being tests covering this it seems like something we want (if a user exits during the completion procedure?)

image

osu.Game/Screens/Multi/Play/TimeshiftPlayer.cs
```diff
@@ -95,19 +96,36 @@ protected override ResultsScreen CreateResults(ScoreInfo score)
...
 
-            var request = new SubmitRoomScoreRequest(token.Value, roomId.Value ?? 0, playlistItem.ID, score);
-            request.Success += s => score.OnlineScoreID = s.ID;
-            request.Failure += e => Logger.Error(e, "Failed to submit score");
-            api.Queue(request);
+            var tcs = new TaskCompletionSource<bool>();
+            var request = new SubmitRoomScoreRequest(token.Value, roomId.Value ?? 0, playlistItem.ID, score.ScoreInfo);
 
-            return score;
+            request.Success += s =>
```
peppy, line 115: we could really do with a better way of asyncing this in the future, if this becomes a more common flow (`api.Perform` should return this awaitable task).

peppy (approved):

peppy:
In the interest of moving things forward I think this is in an amicable stage to merge. Can iterate on code quality around these new flows as issues arise.

osu.Game/Screens/Play/Player.cs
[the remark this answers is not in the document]
smoogipoo: Yeah those errors are the only reason why the schedule exists. They're also the reason why the submission/import task has a `??=` attached to it.

I didn't feel like breaking it for the time being.

PR #11200, opened 2020-12-18
Make Sudden Death & Perfect mod auto restart toggleable

PercyDan54:


bdach:
Seems to be working and I can't see any harm in adding this, but two considerations:

* From what I gather, in the current form `osu-web` changes would have to happen for this to get in, otherwise score submission will start failing with these mods (see https://github.com/ppy/osu-web/pull/6180 for an example of this). However, I'm not 100% sure we should be submitting this setting's value to web, either. Looking for more opinions on that one.

* Given that perfect on stable restarts by default, I'd probably throw something like the following into this diff:

```csharp
protected ModPerfect()
{
    Restart.Value = Restart.Default = true;
}
```

Joehuu:
> However, I'm not 100% sure we should be submitting this setting's value to web, either. Looking for more opinions on that one.

It doesn't change anything in-game. I think it should be a local setting like HD's visibility on first hit object, if anything, or people will be forced into playing with that setting in timeshift.

bdach:
Yeah that's my primary concern, but I really wouldn't want to see that setting moved to actual game settings either. A middle ground (wherein we could exclude attributes from submission, probably via `SettingSourceAttribute`) would be the best way, I think.

peppy:
I think this is still fine to get in, but the proposed default change has not been applied.

PercyDan54:
> I think this is still fine to get in, but the proposed default change has not been applied.

I'm a bit confused.
`ModFailCondition` has `RestartOnFail` default to true, so shall I  move the setting to `ModFailCondition` instead?

bdach:
Doing so will also make Perfect auto-restart toggleable. So it depends on what you want to achieve really. I can't say I care much either way on its own, but people may be used to stable behaviour.

PercyDan54:
Asking for a proposal of wether to make PF auto-restart toggleable or not

bdach:
I think there's no reason not to add the setting to both Perfect and Sudden Death if it's being added to one of them.

bdach:
@PercyDan54 while reviewing the git log i have noticed that one of the merge commits in this pull (4706dcf) has come from a service called `fastgit.org`, and has been signed with a GPG key that isn't github's. do you mind elaborating on what this service is, exactly? because from my brief research it looks quite shady. i would *much* prefer that you only merge in code from github *directly*.

PercyDan54:
> @PercyDan54 while reviewing the git log i have noticed that one of the merge commits in this pull ([4706dcf](https://github.com/ppy/osu/commit/4706dcf5256fe089111483920bc03f40c8210147)) has come from a service called `fastgit.org`, and has been signed with a GPG key that isn't github's. do you mind elaborating on what this service is, exactly? because from my brief research it looks quite shady. i would _much_ prefer that you only merge in code from github _directly_.

It's a third party mirror of github I'm using, because GFW blocks github and makes it works slow.
I know mirror sites are somehow risky so I'm merging from github whenever possible.
The signed GPG key is mine.

bdach:
I've added testing for this. Should be good to go now, theoretically.

bdach (approved):

smoogipoo (approved):

PR #11204, opened 2020-12-18
Abstractify Multiplayer + RoomManager and componentise polling components

smoogipoo:
In preparation for having a `RealtimeRoomManager : RoomManager` which has much of the same polling requirements but slightly different implementations.

This is a good time to let you know of the structure of things until realtime multiplayer is fully implemented. All new timeshift components will go in `Visual/Multi/Timeshift`, all new realtime multiplayer components will go in `Visual/Multi/RealtimeMultiplayer` to match the `Online/RealtimeMultiplayer` namespace.  
Shared components will live in `Visual/Multi/Components`, and as such I've moved `RoomManager` into there.


osu.Game/Screens/Multi/Timeshift/TimeshiftMultiplayer.cs
```diff
@@ -0,0 +1,48 @@
...
+using osu.Game.Screens.Multi.Lounge;
+using osu.Game.Screens.Multi.Match;
+
+namespace osu.Game.Screens.Multi.Timeshift
+{
+    public class TimeshiftMultiplayer : Multiplayer
+    {
+        protected override void UpdatePollingRate(bool isIdle)
+        {
+            var timeshiftManager = (TimeshiftRoomManager)RoomManager;
```
bdach, line 15: This makes me think that maybe `Multiplayer` should be taken a step further and it should be a generic class, something of this sort, possibly?

```csharp
public abstract class Multiplayer<TRoomManager> : OsuScreen
	where TRoomManager : RoomManager
```

Would then allow to expose `RoomManager` as `TRoomManager` and avoid this cast, but unsure if too enterprise:tm:...

smoogipoo: Unfortunately `Multiplayer` is DI'd and used elsewhere in code. It's probably not the correct way of doing things but making `Multiplayer` generic feels like too much abstraction for very little gain.

osu.Game/Screens/Multi/Components/RoomPollingComponent.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        public IBindable<bool> InitialRoomsReceived => initialRoomsReceived;
+        private readonly Bindable<bool> initialRoomsReceived = new Bindable<bool>();
+
+        [Resolved]
+        protected IAPIProvider API { get; private set; }
+
+        protected RoomPollingComponent()
+        {
+            TimeBetweenPolls.BindValueChanged(time => base.TimeBetweenPolls = time.NewValue);
```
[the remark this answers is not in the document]
peppy: @smoogipoo Is there a reason this wasn't the way it was done? just for simplicity?

osu.Game/Screens/Multi/Timeshift/TimeshiftMultiplayer.cs
peppy: I think this is fine for now. Less generic has generally worked better for us until otherwise necessary (at the expense of an occasional explicit cast).

osu.Game/Screens/Multi/Components/RoomPollingComponent.cs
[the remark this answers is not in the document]
smoogipoo: I've made it into a bindable now.

osu.Game/Screens/Multi/Components/RoomPollingComponent.cs
```diff
@@ -0,0 +1,29 @@
...
+using osu.Game.Online.API;
+using osu.Game.Online.Multiplayer;
+
+namespace osu.Game.Screens.Multi.Components
+{
+    public abstract class RoomPollingComponent : PollingComponent
+    {
+        public Action<List<Room>> RoomsReceived;
+
+        public readonly Bindable<bool> InitialRoomsReceived = new Bindable<bool>();
```
peppy, line 18: What's the point of this one? I've checked follow up PRs and can only find actual usage in tests (and it feels a bit flaky because it's getting reset by a different class). Thinking maybe an event action may work better if that's the only usage?

peppy (changes requested):

osu.Game/Screens/Multi/Components/RoomPollingComponent.cs
bdach: `RoomManager`'s correspondent bindable binds to these ones I'm pretty sure, and higher up `InitialRoomsReceived` controls the lounge loading layer showing up. I do agree that I'm not 100% on this structure either, given that `RoomPollingComponent` does

```csharp
InitialRoomsReceived.Value = true;
RoomsReceived?.Invoke(rooms);
```

Feels like two members doing kind of the same thing.

peppy: Aha, right. Found it in `LoungeSubScreen`.

Yeah still not 100% sold on the bindable being exposed (would prefer an `IBindable` + `ResetRooms()` method or similar?). It looks like the rationale for doing it this way was to pass the bindable down one or two levels, so not sure how much overhead that will add without trying to implement it.

smoogipoo: I did it this way to avoid making unnecessary changes into what is already a large changeset. But sure.

peppy (approved):
seems ok

bdach (approved):
looks good!

PR #11206, opened 2020-12-18
Add a base class for all realtime multiplayer composites

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/11205

Extends `MultiplayerComposite` by also providing the client and a `OnRoomChanged()` method.


bdach:
The single non-prereq commit looks sane, but not approving yet since it's pretty deep in the chain.

peppy (approved):

bdach (approved):

PR #11207, opened 2020-12-18
Implement the realtime multiplayer participants list

smoogipoo:
- [x] https://github.com/ppy/osu/pull/11206

It's a bit of an older design, but should do for now.

image

### DIFF
--- a/osu.Game/Screens/Multi/RealtimeMultiplayer/Participants/ParticipantPanel.cs
+++ b/osu.Game/Screens/Multi/RealtimeMultiplayer/Participants/ParticipantPanel.cs
@@ -1,0 +1,187 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Diagnostics;
+using osu.Framework.Allocation;
+using osu.Framework.Extensions.Color4Extensions;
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Multi/RealtimeMultiplayer/Participants/ParticipantPanel.cs
```diff
@@ -0,0 +1,187 @@
...
+
+        protected override void OnRoomChanged()
+        {
+            base.OnRoomChanged();
+
+            if (Room == null)
+                return;
+
+            if (User.State == MultiplayerUserState.Ready)
+                readyMark.FadeIn(50);
```
bdach, line 146: nit: would extract this transition duration to a const

bdach (commented):
Not too much to say about this one either, save for one nit.

osu.Game/Screens/Multi/RealtimeMultiplayer/Participants/ParticipantPanel.cs
smoogipoo: Sure. I've made it a const local to this method. I'm not going to move it outside because this is also super temporary - I'd implement this differently like giving the ready mark a y-movement, but I'm not spending too much time on designs at this point.

bdach (approved):

peppy (approved):

PR #11208, opened 2020-12-18
Implement the realtime multiplayer ready button

smoogipoo:
Prereqs:
- [x] https://github.com/ppy/osu/pull/11206

https://user-images.githubusercontent.com/1329837/102645877-7fd56480-41a6-11eb-8566-1255eb8a600e.mp4

### DIFF
--- a/osu.Game.Tests/Visual/RealtimeMultiplayer/TestSceneRealtimeReadyButton.cs
+++ b/osu.Game.Tests/Visual/RealtimeMultiplayer/TestSceneRealtimeReadyButton.cs
@@ -97,6 +97,12 @@ namespace osu.Game.Tests.Visual.RealtimeMultiplayer
         [Test]
         public void TestBecomeHostWhileReady()
         {
+            AddStep("add host", () =>
+            {
+                Client.AddUser(new User { Id = 2, Username = "Another user" });
+                Client.TransferHost(2);
+            });
+
             addClickButtonStep();
             AddStep("make user host", () => Client.TransferHost(Client.Room?.Users[0].UserID ?? 0));
 
--- a/osu.Game.Tests/Visual/RealtimeMultiplayer/TestSceneRealtimeReadyButton.cs
+++ b/osu.Game.Tests/Visual/RealtimeMultiplayer/TestSceneRealtimeReadyButton.cs
@@ -1,0 +1,129 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.Linq;
+using NUnit.Framework;
+using osu.Framework.Allocation;
+using osu.Framework.Audio;
+using osu.Framework.Graphics;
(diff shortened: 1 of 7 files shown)


osu.Game/Screens/Multi/Timeshift/TimeshiftReadyButton.cs
```diff
@@ -0,0 +1,38 @@
...
+            BackgroundColour = colours.Green;
+            Triangles.ColourDark = colours.Green;
+            Triangles.ColourLight = colours.GreenLight;
+        }
+
+        protected override void Update()
+        {
+            base.Update();
+
+            Enabled.Value = endDate.Value == null || DateTimeOffset.UtcNow.AddSeconds(30).AddMilliseconds(GameBeatmap.Value.Track.Length) < endDate.Value;
```
bdach, line 35: This `endDate` null check wasn't there before - are there eventually going to be timeshift lobbies without an end date?

osu.Game.Tests/Visual/RealtimeMultiplayer/TestSceneRealtimeReadyButton.cs
```diff
@@ -0,0 +1,129 @@
...
+
+            addClickButtonStep();
+            AddAssert("match started", () => Client.Room?.Users[0].State == MultiplayerUserState.WaitingForLoad);
+        }
+
+        [Test]
+        public void TestBecomeHostWhileReady()
+        {
+            addClickButtonStep();
+            AddStep("make user host", () => Client.TransferHost(Client.Room?.Users[0].UserID ?? 0));
```
bdach, line 101: There's actually no second user in this room, so this step has no effect and this test doesn't test what it says it does. The current user already has host.

bdach (changes requested):

bdach:
Not entirely sure about the UX of this. The one particular case I have in mind is readying up when you have host.

* If everyone else has readied up, only the button text changes, the colour doesn't. So the click sorta doesn't seem to have any effect unless you spot the changing copy.
* If there's someone that hasn't readied up, the button changes colour and text.

My main issue is that the button change is too easy to miss. What I would propose instead:

* When not ready, the button is a neutral colour, not green (let's say gray); text is "Ready"
* When readying up:
	* If not host, then the button goes green; text is "Waiting for host..."
	* If host and not all players have readied up, the button goes yellow as currently; not sure about the text, maybe something like "Start game (ready players only)"
	* If host and all players have readied up, the button goes green, with text "All ready!"

peppy:
I discussed this in voice with @smoogipoo, but optimally either on the button text or next to it, it would show `Start game (1 / 9 ready)` or similar.

osu.Game/Screens/Multi/Timeshift/TimeshiftReadyButton.cs
smoogipoo: This is actually an error. Good catch.

peppy (approved):
👍🏻

PR #11212, opened 2020-12-19
Add ability to import multiple files at once on android

Game4all:
Basically a follow-up to #11099

# Summary

This adds the ability to import multiple files at once on android from a file explorer by selecting them and sharing them to osu!.

The previously added method to import files from a stream have been refactored to directly take in an array of import tasks and the provided list of import tasks is being directly created from inside the intent handling code.

## Testing
Tested this on my 3 devices running android versions above 9.0. However, you shouldn't run into any problems this time while testing this on earlier android versions.

It is noteworthy that importing more than 20+ beatmaps at once will extend the game loading time by a dozen of seconds, presumably because the beatmap import tasks are running before the tasks loading asynchronously drawable components, and I guess this may depends of the device specs (number of avalaible cores for concurrent tasks), I'm not sure at which measure this is a problem (it is however definitely a problem for very large imports).


osu.Android/OsuGameActivity.cs
```diff
@@ -48,41 +51,64 @@ private void handleIntent(Intent intent)
...
+        private void handleImportFromUris(params Uri[] uris) => Task.Factory.StartNew(async () =>
         {
-            // there are more performant overloads of this method, but this one is the most backwards-compatible
-            // (dates back to API 1).
-            var cursor = ContentResolver?.Query(uri, null, null, null, null);
+            var tasks = new List<ImportTask>();
 
-            if (cursor == null)
-                return;
+            await Task.WhenAll(uris.Select(async uri =>
```
bdach, line 84: yeah, no wonder this is holding up the game launch. i find this sort of dodgy, you're likely starving the TPL thread pool with that sort of code.

did you try creating these tasks serially via continuations (ie. one by one instead of a free-for-all like this)?

osu.Game/OsuGameBase.cs
```diff
@@ -395,14 +395,14 @@ public async Task Import(params string[] paths)
             }
         }
 
-        public virtual async Task Import(Stream stream, string filename)
+        public virtual async Task Import(params ImportTask[] tasks)
         {
-            var extension = Path.GetExtension(filename)?.ToLowerInvariant();
+            var extension = Path.GetExtension(tasks.First().Path).ToLowerInvariant();
```
bdach, line 400: silently assuming that all imported files will be of the same extension is not safe to do here and will result in import failures if the user doesn't happen to uphold this assumption. each file should be either handled separately, or they should be grouped by extension.

bdach (changes requested):

osu.Android/OsuGameActivity.cs
Game4all: Actually I was suspecting that one to be retarding the game launch, but upon inspection, resolving all filenames + opening the file streams and copying them to a memory stream takes about 1.5s for 20 beatmaps. Will still look into what you proposed out of curiosity

osu.Game/OsuGameBase.cs
peppy: That's actually how we do it on `master` at the moment, for best or worst (see method above):

https://github.com/ppy/osu/blob/28674cedd6849ebe4bee8417cad3ed8fe385b2ac/osu.Game/OsuGameBase.cs#L389

Game4all: Guess it's a good opportunity to fix both import paths in this PR

peppy: You can try, but I'm pretty sure it was done that way with reason. Also if you're changing the other case I'd prefer it in a second PR.

bdach: From what I can tell this `.First()` call originates from https://github.com/ppy/osu/commit/e0d28564d0d69b4132f6dea94f1e2a162d18c2e4. Before it was checking (arguably more properly) if all extensions were the same before proceeding. Can't see any discussion of that change in the associated PR.

Game4all:
Also, now that we've got an import overload taking directly a list of `ImportTasks` as parameter, I guess that the import system could be refactored to exclusively use `ImportTask`s everywhere and eventually drop the string-based import system ... but that's work for another PR .

bdach:
yep, definitely out of scope for this one.

osu.Android/OsuGameActivity.cs
Game4all: I still get results lying in a  similar time frame when running them sequentially one by one so I believe the problem lies somewhere else. Upon looking at `ArchiveModelManager` :https://github.com/ppy/osu/blob/4fba0c8e6a3279216294a5a75fa2f79687d0a0c1/osu.Game/Database/ArchiveModelManager.cs#L136 this line is  spawning a task for each import task which may deplete the TPL pool and we're limiting the actual number of imports running to one concurrent import further down, so the problem may lie there.

Game4all: After replacing that exact line to runs tasks sequentially I can affirm the line aforementionned causes the game loading to be longer when importing a lot of beatmaps. Opinions welcome on how to handle that elegantly (as I don't think putting a platform check to chose whether to run tasks sequentially or in parallel in the middle of the code is the best idea)

bdach: hmm, so the diagnosis was correct, just not the place... a way to change this *might* be to have a dedicated separate pool for imports, although I'm not sure whether spawning even more threads on mobile is a good idea. maybe the tasks should run sequentially during launch and then switch to fully parallelised while the game is ready to go? not sure.

I probably wouldn't base it on any platform check since this will likely behave similarly everywhere, it's just it's been a rarely-checked scenario until now (albeit possible, as the game accepts filenames to import via command line).

Game4all: I don't think either spawning one more thread per archive manager is a good idea, so alternatively I guess the game could maybe use a custom threaded task scheduler for loading menu components asynchronously instead .

Honestly I'm tempted to just make import tasks running sequentially but it is easily foreseeable this will reduce the import performance when running large imports so .. dunno .

peppy: Something like this?

https://github.com/ppy/osu/blob/3e3d5b2e37431d5e6074fbcb84ca58a7adae0cbe/osu.Game/Database/ArchiveModelManager.cs#L48

Game4all: Yet another alternative would be to schedule all imports tasks on `import_scheduler` but not sure how much it would tank performance again ...

peppy: This issue also seems out of scope. I'd open a separate issue covering this (since it looks like it would probably fail on desktop import as well?) and continue with getting this PR merged as-is.

Game4all: This is now tracked by #11251. As such, marking this as resolved.

bdach:
@Game4all This needs conflict resolution (and probably updating the new intent filters to adhere to what we know about how they work now).

smoogipoo (dismissed):

smoogipoo (commented):
This looks ok to me from just a code viewpoint.

bdach:
Made a few adjustments, tested, seems to work.

bdach (approved):

PR #11216, opened 2020-12-20
Fix selected item not being highlighted on some setting dropdowns

### LINKED ISSUE #8912: Some dropdowns don't show selected option
**Describe the bug:** Some dropdowns don't show the selected option in the list until you change it manually (but the issue occurs again when re-launching the game).

These dropdowns seem to have this issue:
- Intro Sequence
intro sequence
- Background Source
background source

Expected Behaviour:
- Intro Sequence
intro sequence
- Background Source
background source

Joehuu:
Fixes https://github.com/ppy/osu/issues/8912.


osu.Game/Overlays/Settings/Sections/Graphics/LayoutSettings.cs
```diff
@@ -59,11 +59,10 @@ private void load(FrameworkConfigManager config, OsuConfigManager osuConfig, Gam
 
             Children = new Drawable[]
             {
-                windowModeDropdown = new SettingsDropdown<WindowMode>
+                windowModeDropdown = new SettingsEnumDropdown<WindowMode>
                 {
                     LabelText = "Screen mode",
                     Current = config.GetBindable<WindowMode>(FrameworkSetting.WindowMode),
-                    ItemSource = windowModes,
```
frenzibyte, line 66: Does not respect `host.Window.SupportedWindowModes` anymore.

frenzibyte (changes requested):

frenzibyte:
This technically is hiding the issue instead of fixing it, I believe the underlying issue is about `Dropdown.Current` set to a value that does not have a corresponding item yet (because `Items` is set **after** `Current` is), therefore selecting a temporarily-created item.

This can be solved framework-side by updating selection to the real item when `Items` is set.

peppy:
@frenzibyte sure, but the dropdowns which can be changed to use the enum type still should be (code quality improvement). only issue is `windowModeDropdown`, as you pointed out, which has changed behaviour.

peppy (approved):

PR #11222, opened 2020-12-20
Enable pooling for taiko DHOs

### LINKED ISSUE #6123: Beatmap for osu!taiko uses 3GB of RAM
**Describe the crash:** 
If you trying to load this (https://osu.ppy.sh/beatmapsets/537048#osu/1138154) beatmap (taiko) on the latest difficult , osu!lazer uses 3GB of RAM, and freezes if you trying to go to the menu.
https://streamable.com/li5ze
**osu!lazer version:** 
Lastest.
**Logs:** 
[updater.log](https://github.com/ppy/osu/files/3613713/updater.log)
[network.log](https://github.com/ppy/osu/files/3613714/network.log)
[performance.log](https://github.com/ppy/osu/files/3613715/performance.log)
[runtime.log](https://github.com/ppy/osu/files/3613716/runtime.log)
**Computer Specifications:** 
G4560,
RX 470 Nitro+,
16GB DDR4-2400

bdach:
- (Mostly) closes #6123.
- [x] Depends on #11221.

In terms of pooling, this will be most likely the single biggest noticeable change memory-usage-wise. It's especially visible on long maps; for a data point, I ran a rough single-run test on [this marathon map](https://osu.ppy.sh/beatmapsets/447342#taiko/960444) and the results are as follows:

| | load time | memory usage at song select | memory usage at ~1:00 |
| :-: | -: | -: | -: |
| `master` | 1min 6s | 847k | 1.450G |
| this PR | 3s | 960k | 1.059G |

Allocations are not entirely gone, mind you - there's still work to be done in that respect. Individual DHOs will still recreate their child drawables on application and cause GC churn, and flying hits and explosions are yet to be pooled.

Pool counts are ballpark and based on the densest maps I had locally imported; hits rarely go above high 20s/low 30s and the rest is for safety, and drum rolls and swells get on average 20 nested objects per main object instance.

In my testing I couldn't see any breakage in gameplay, but editor seemed to have some issues, perhaps predictably so, given that this is the first scrolling ruleset with both pooling and editor. For instance, frame-stable seeks, as usual, break stuff and take out too many drawables from the pool, but I'm not yet sure what's the cause of that (and they do get returned when playback resumes). Let me know if you want me to address that before merging this, or if we'll proceed with this and fix up editor later as was the case with osu! efforts.

### DIFF
--- a/osu.Game.Rulesets.Taiko.Tests/TestSceneSampleOutput.cs
+++ b/osu.Game.Rulesets.Taiko.Tests/TestSceneSampleOutput.cs
@@ -1,6 +1,7 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Collections.Generic;
 using System.Linq;
 using osu.Framework.Testing;
 using osu.Game.Audio;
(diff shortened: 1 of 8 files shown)


peppy (approved):
nice and clean.

peppy:
Let's fix editor issues as follow-up.

PR #11226, opened 2020-12-21
Fix incorrectly provided texture loader store to skins

peppy:
This fixes the remaining cases of local `TextureLoaderStore` construction (bypassing host implementation). In the process, I attempted to clean up the way information is passed down to `WorkingBetamap` and `LegacySkin` classes, as the constructors / data feeding for both of these was getting pretty ugly.


osu.Game.Tests/Gameplay/TestSceneStoryboardSamples.cs
```diff
@@ -176,5 +178,9 @@ public TestDrawableStoryboardSample(StoryboardSampleInfo sampleInfo)
             {
             }
         }
+
+        public AudioManager AudioManager => Audio;
+        public IResourceStore<byte[]> Files => null;
+        public IResourceStore<TextureUpload> CreateTextureLoaderStore(IResourceStore<byte[]> underlyingStore) => null;
```
bdach, line 184: These implementations are all in pretty big classes, so I'd `#region` them, at least in order to be able to visually separate them out.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -68,9 +69,11 @@ public partial class BeatmapManager : DownloadableArchiveModelManager<BeatmapSet
         private readonly RulesetStore rulesets;
         private readonly BeatmapStore beatmaps;
         private readonly AudioManager audioManager;
-        private readonly TextureStore textureStore;
+        private readonly LargeTextureStore largeTextureStore;
         private readonly ITrackStore trackStore;
 
+        private readonly GameHost host;
```
bdach, line 75: As per default ctor argument, this is a potentially

```suggestion
		[CanBeNull]
        private readonly GameHost host;
```

field.

osu.Game/Beatmaps/BeatmapManager.cs
```diff
@@ -492,6 +496,12 @@ public void Dispose()
             onlineLookupQueue?.Dispose();
         }
 
+        TextureStore IBeatmapResourceProvider.LargeTextureStore => largeTextureStore;
+        ITrackStore IBeatmapResourceProvider.Tracks => trackStore;
+        AudioManager IStorageResourceProvider.AudioManager => audioManager;
+        IResourceStore<byte[]> IStorageResourceProvider.Files => Files.Store;
+        IResourceStore<TextureUpload> IStorageResourceProvider.CreateTextureLoaderStore(IResourceStore<byte[]> underlyingStore) => host.CreateTextureLoaderStore(underlyingStore);
```
bdach, line 503: Above suggestion shows that this has an unchecked access to possibly-null host. Only caller (`LegacySkin`) is null-tolerant, so I'd null-check here and call it a day.

```suggestion
        IResourceStore<TextureUpload> IStorageResourceProvider.CreateTextureLoaderStore(IResourceStore<byte[]> underlyingStore) => host?.CreateTextureLoaderStore(underlyingStore);
```

osu.Game/Beatmaps/BeatmapManager_WorkingBeatmap.cs
```diff
@@ -21,16 +19,12 @@ public partial class BeatmapManager
...
         {
-            private readonly IResourceStore<byte[]> store;
-            private readonly TextureStore textureStore;
-            private readonly ITrackStore trackStore;
+            private readonly IBeatmapResourceProvider resources;
 
-            public BeatmapManagerWorkingBeatmap(IResourceStore<byte[]> store, TextureStore textureStore, ITrackStore trackStore, BeatmapInfo beatmapInfo, AudioManager audioManager)
-                : base(beatmapInfo, audioManager)
+            public BeatmapManagerWorkingBeatmap(BeatmapInfo beatmapInfo, IBeatmapResourceProvider resources)
+                : base(beatmapInfo, resources?.AudioManager)
```
bdach, line 25: Sort of confused by the null-check on `resources` there, especially seeing that the rest of this class assumes the converse, and the only caller of this ctor passes `this`. Would remove this null-check (and maybe annotate param as `[NonNull]`).

bdach (changes requested):
Not feeling 100% on this pull... The new pattern feels solid for normal code (apart from the comments below), but sort of flaky when tests become involved. Stuff is sometimes null and sometimes non-null, which is maybe fine, but puts me at some unease.

I guess the interface members could get nullability annotations - the biggest obstacle to that is `SkinnableTestScene`'s implementation of `Files`, which is the only one that can return nulls. I don't know, maybe fine to have test code not feel completely safe.

osu.Game/Skinning/LegacySkin.cs
```diff
@@ -54,12 +53,12 @@ public new LegacySkinConfiguration Configuration
...
-            : this(skin, new LegacySkinResourceStore<SkinFileInfo>(skin, storage), audioManager, "skin.ini")
+        public LegacySkin(SkinInfo skin, IStorageResourceProvider resources)
+            : this(skin, new LegacySkinResourceStore<SkinFileInfo>(skin, resources.Files), resources, "skin.ini")
         {
         }
 
-        protected LegacySkin(SkinInfo skin, IResourceStore<byte[]> storage, AudioManager audioManager, string filename)
+        protected LegacySkin(SkinInfo skin, IResourceStore<byte[]> storage, IStorageResourceProvider resources, string filename)
             : base(skin)
         {
```
bdach, line 63: These are also a tiny bit confusing - public ctor can assume non-null `resources`, but protected can already have null `storage` and `resources`.

peppy:
@bdach regarding your comment on testing: is that important? if there's a null ref, it's going to fail in the test anyway. if you look at the previous code, we were already throwing around nulls (sometimes) and generally only providing implementations / instances where actually required by the test.

bdach:
Yeah as I said at the end there it's probably fine, the current state seems better than it was previously anyway.

bdach (approved):

peppy:
@smoogipoo can you take a quick look at this and make sure you're okay with the general structure?

peppy:
In discussion this is fine to go in for now.

PR #11227, opened 2020-12-21
Fix incorrect joinedroom null checks

smoogipoo:
Would never fail because it's nullchecking the bindable rather than the bindable's value.


peppy:
weird the r# can't pick up on that one

peppy (approved):

PR #11233, opened 2020-12-21
Fix potential cross-thread operation during chat channel load

### LINKED ISSUE #11231: (Rare?) error after going through some overlays such as news, changelog and ranking
**Describe the crash:** 
i just cycled through some overlays such as news, changelog and ranking and suddenly an exception was thrown. i dont know how to reproduce this and cant get it again.

**osu!lazer version:** 
2020.1212.0

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5722852/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5722853/updater.log)
[database.log](https://github.com/ppy/osu/files/5722854/database.log)
[network.log](https://github.com/ppy/osu/files/5722855/network.log)
[performance.log](https://github.com/ppy/osu/files/5722856/performance.log)

peppy:
The callbacks are scheduled to the API thread, but hooked up in BDL load. This causes a potential case of cross-thread collection enumeration.

I've tested and it seems like the schedule logic should be fine for short term. Longer term, we probably want to re-think how this works so background operations aren't performed on the `DrawableChannel` in the first place (chat shouldn't have an overhead like this when not visible).

Closes #11231.

### DIFF
--- a/osu.Game/Overlays/Chat/DrawableChannel.cs
+++ b/osu.Game/Overlays/Chat/DrawableChannel.cs
@@ -103,7 +103,7 @@ namespace osu.Game.Overlays.Chat
             Colour = colours.ChatBlue.Lighten(0.7f),
         };
 
-        private void newMessagesArrived(IEnumerable<Message> newMessages)
+        private void newMessagesArrived(IEnumerable<Message> newMessages) => Schedule(() =>
         {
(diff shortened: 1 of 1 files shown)


bdach (approved):
seems logical :+1:

peppy:
Going to get this one in. We'll probably not notice any serious issues with this until it hits user testing (the only potential problems will come from scenarios where osu! is running for a long time).

PR #11234, opened 2020-12-21
Fix room category being serialised as int

smoogipoo:


osu.Game/Online/Multiplayer/Room.cs
```diff
@@ -35,9 +35,22 @@ public class Room
         public readonly Bindable<int> ChannelId = new Bindable<int>();
 
         [Cached]
-        [JsonProperty("category")]
+        [JsonIgnore]
         public readonly Bindable<RoomCategory> Category = new Bindable<RoomCategory>();
 
+        // Todo: osu-framework bug (https://github.com/ppy/osu-framework/issues/4106)
+        [JsonProperty("category")]
+        private string categoryString
```
peppy, line 43: can this potentially use `[JsonConverter(typeof(StringEnumConverter))]` with an enum rather than local string logic here?

smoogipoo: Good point.

smoogipoo: Looks like osu-web is case-sensitive so it's still a little ugly but not too bad.

peppy (approved):

PR #11236, opened 2020-12-21
Update osu!mania legacy skin's judgement animation to match stable

### LINKED ISSUE #11213: Mania hits don't match stable size
**Describe the bug:** 

I imported my main skin to lazer, I use the default 300g mania hit taken from the default skin from [this topic](https://osu.ppy.sh/community/forums/topics/129191), i quickly noticed a difference in the size of the mania hits, they look way bigger in lazer than in stable.

Here's [the skin](https://drive.google.com/file/d/1TRbfUsyBzAsxtnOdJ5Ye2bQxZ5W8bKtG/view?usp=sharing)

Stable: 
stable

Lazer: 
lazer

I cropped the hits for comparison:
Comparacion

**osu!lazer version:** 

2020.1212.0

peppy:
Closes #11213.

stable:

https://user-images.githubusercontent.com/191335/102754749-c1396000-43b0-11eb-8d2d-313a0b42ffc8.mov

lazer:

https://user-images.githubusercontent.com/191335/102754840-e6c66980-43b0-11eb-901a-b35021bcd1b8.mov

### DIFF
--- a/osu.Game.Rulesets.Mania/Skinning/Legacy/LegacyManiaJudgementPiece.cs
+++ b/osu.Game.Rulesets.Mania/Skinning/Legacy/LegacyManiaJudgementPiece.cs
@@ -69,7 +69,7 @@ namespace osu.Game.Rulesets.Mania.Skinning.Legacy
                     animation.ScaleTo(1.2f).Then().ScaleTo(1, 100, Easing.Out);
 
                     animation.RotateTo(0);
-                    animation.RotateTo(RNG.NextSingle(-18, 18), 100, Easing.Out);
+                    animation.RotateTo(RNG.NextSingle(-5.73f, 5.73f), 100, Easing.Out);
                     break;
 
                 default:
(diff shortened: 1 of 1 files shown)


osu.Game.Rulesets.Mania/Skinning/Legacy/LegacyManiaJudgementPiece.cs
```diff
@@ -56,31 +56,30 @@ public void PlayAnimation()
...
-
-                    animation.MoveTo(Vector2.Zero);
-                    animation.MoveToOffset(new Vector2(0, 100), 800, Easing.InQuint);
+                    animation.ScaleTo(1.2f).Then().ScaleTo(1, 100, Easing.Out);
 
                     animation.RotateTo(0);
-                    animation.RotateTo(40, 800, Easing.InQuint);
-
-                    this.FadeOutFromOne(800);
+                    animation.RotateTo(RNG.NextSingle(-18, 18), 100, Easing.Out);
```
bdach, line 72: This seems like way too much rotation - it's almost a 40 degree random spread, stable definitely doesn't go that wide. Feels like this should be about a third of the current range at most.

bdach (changes requested):

osu.Game.Rulesets.Mania/Skinning/Legacy/LegacyManiaJudgementPiece.cs
peppy: i converted wrong, should be 5.73 instead (stable range is -0.1 to 0.1 radians)

bdach (approved):

PR #11238, opened 2020-12-21
Improve drag and selection UX on editor timeline

### LINKED ISSUE #10486: Timeline requires too much precision to select circles
I think making the selection area bigger would cause less headache

https://puu.sh/GBWWL/5c4cebff14.png

Since I personally don't see many people actually dragging the timeline using a mouse, this change would be welcomed

peppy:
Closes #10486 (and potentially #10487?) by increasing the drag area, the blueprint circle size, and also highlighting the draggable area when hovering for better visibility.

before:

after:


bdach (approved):
Yeah I'd say this is an improvement :+1:

PR #11239, opened 2020-12-21
Implement realtime multiplayer

smoogipoo:
Implementing this as a one-pass thing since it's pretty self-contained.


peppy (changes requested):

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
```diff
@@ -0,0 +1,81 @@
...
+using osu.Framework.Allocation;
+using osu.Game.Online.Multiplayer;
+using osu.Game.Online.RealtimeMultiplayer;
+using osu.Game.Scoring;
+using osu.Game.Screens.Multi.Play;
+using osu.Game.Screens.Ranking;
+
+namespace osu.Game.Screens.Multi.RealtimeMultiplayer
+{
+    public class RealtimePlayer : TimeshiftPlayer
```
bdach, line 17: This inheritance is maybe a bit ehh, but abstracting out further may be too much of muchness... Probably fine to leave be for now until it's a problem.

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            if (Token == null)
+                return; // Todo: Somehow handle token retrieval failure.
+
+            client.MatchStarted += onMatchStarted;
+            client.ResultsReady += onResultsReady;
+            client.ChangeState(MultiplayerUserState.Loaded);
+
+            while (!started)
+                Thread.Sleep(100);
```
bdach, line 46: slight teeth pain at this but I suppose also maybe not super important for now - can this cause problems? maybe TODO it at least?

osu.Game/Online/Multiplayer/PlaylistItem.cs
```diff
@@ -64,8 +64,8 @@ public PlaylistItem()
 
         public void MapObjects(BeatmapManager beatmaps, RulesetStore rulesets)
         {
-            Beatmap.Value = apiBeatmap.ToBeatmap(rulesets);
-            Ruleset.Value = rulesets.GetRuleset(RulesetID);
+            Beatmap.Value ??= apiBeatmap.ToBeatmap(rulesets);
+            Ruleset.Value ??= rulesets.GetRuleset(RulesetID);
```
bdach, line 68: So `MapObjects()` won't ever overwrite a non-null beatmap anymore? The commit that introduced these (536df07) has a `(for testing)` annotation in the message - is this meant to be final? This affects the abstract `RoomManager`, and both multiplayer modes as a consequence.

osu.Game/Online/RealtimeMultiplayer/RealtimeMultiplayerClient.cs
```diff
@@ -0,0 +1,171 @@
...
+        {
+            if (connection != null)
+                return;
+
+            connection = new HubConnectionBuilder()
+                         .WithUrl(endpoint, options =>
+                         {
+                             options.Headers.Add("Authorization", $"Bearer {api.AccessToken}");
+                         })
+                         .AddNewtonsoftJsonProtocol(options => { options.PayloadSerializerSettings.ReferenceLoopHandling = ReferenceLoopHandling.Ignore; })
```
bdach, line 66: Out of curiosity more than anything, does this serialiser setting actually matter? There are actual reference loops?

bdach (commented):
Pretty hard (=large) diff to go through. Skimmed the new components without paying much attention - focused more on actual logic and changes to existing components.

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
frenzibyte: worth a TODO comment at least IMO.

osu.Game/Online/RealtimeMultiplayer/RealtimeMultiplayerClient.cs
peppy: there was for spectator at least (beatmap -> beatmapset -> beatmaps). i think it's just a safe default at this point, either way.

osu.Game/Online/Multiplayer/PlaylistItem.cs
smoogipoo: Yes this is meant to be final. It's here for tests where the beatmap is set directly on `PlaylistItem`s.

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
smoogipoo: You've brought up an interesting point though, which is that this should've had a timeout. I've changed it to using an MRE.

peppy:
Trying to join an ended room gets the lobby into a permanently broken state:

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeRoomManager.cs
```diff
@@ -38,10 +38,10 @@ protected override void LoadComplete()
         }
 
         public override void CreateRoom(Room room, Action<Room> onSuccess = null, Action<string> onError = null)
-            => base.CreateRoom(room, r => joinMultiplayerRoom(r, onSuccess), onError);
+            => base.CreateRoom(room, r => joinMultiplayerRoom(r, onSuccess, onError), onError);
```
peppy, line 41: this is really hard to read :/

peppy (dismissed):

peppy:
Mostly happy with the state of changes (with a focus on changes to existing classes). For the new stuff, there's still many issues, a lot of which we are already tracking on a todo list. I think merging this ASAP and following up with PRs is the correct way forward from here on.

PR #11252, opened 2020-12-22
Add simple state display for participants

peppy:
Better visibility over what users in the room are currently doing (especially when there is an ongoing game and a new user joins the room)


bdach:
I've applied minor adjustments to this, but otherwise this looks pretty good to me :+1:

bdach (approved):

PR #11253, opened 2020-12-22
Add leaderboard display to realtime player

peppy:
Done in the simplest way possible for now. Eventually we'll probably want it to exist in a derived version of the HUD display. but this should do fine for now.


bdach:
I've pushed in a fix for a potential crash when the submission token is not issued correctly (found it when trying to bypass that logic to test how it looks in-game, but didn't disable all of the submission logic).

As an aside, now that the leaderboard is being populated with real data - the accuracy numbers don't match up with the HUD. This is because the accuracy portion calculation uses the max score over the entire beatmap instead of max score up to the given point in gameplay. Unsure if this is something we want to leave be, fix now, or fix later, so not merging this one just yet pending that discussion.

peppy:
Nice catch. I've fixed this in a "mostly correct" way. The values may be slightly off what they should be based on latency to server (individual players will not necessarily be at the same point in gameplay as the local `ScoreProcessor`).

A proper fix would be to either send the raw accuracy value in the header packet, or send the rolling base score. Can go in this direction if you think that's a more amicable solution (but will need nuget/server updates, so maybe it should be applied as a separate fix when we have the time/throughput for it?).

bdach:
I think that's an ok fix for now. The biggest thing to me was to make sure the user's own accuracy didn't look completely out of whack in comparison to HUD, and this version achieves that. Sending raw accuracy can happen a little later.

bdach (approved):

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
```diff
@@ -55,6 +62,31 @@ private void load()
                     this.Exit();
                 });
             }
+
+            Debug.Assert(client.Room != null);
+
+            int[] userIds = client.Room.Users.Where(u => u.State >= MultiplayerUserState.WaitingForLoad).Select(u => u.UserID).ToArray();
```
smoogipoo, line 68: Something to keep in mind is that this is unsafe given it's being done on the BDL thread while there are potentially users joining/leaving scheduled to the update thread. It could lead to collection modified during enumeration exceptions. I think the best way to do this is to pass in the userIds via ctor, which is guaranteed to be on the update thread.

peppy: yeah, wasn't 100% on this, will move it.

PR #11254, opened 2020-12-22
Fix beatmap track not looping on multiplayer song selects

frenzibyte:
This was occurring due to `Multiplayer.updateTrack()`'s looping logic overriding any other one (`SongSelect`'s looping logic in this case) and assuming it should always be set to `false`.

Fixed it by moving the looping logic inside `MatchSubScreen` such that when it's not the current screen, looping can be changed freely.

### DIFF
--- a/osu.Game/Screens/Multi/Match/RoomSubScreen.cs
+++ b/osu.Game/Screens/Multi/Match/RoomSubScreen.cs
@@ -9,6 +9,7 @@ using osu.Framework.Screens;
 using osu.Game.Audio;
 using osu.Game.Beatmaps;
 using osu.Game.Online.Multiplayer;
+using osu.Game.Overlays;
 using osu.Game.Rulesets.Mods;
 
 namespace osu.Game.Screens.Multi.Match
@@ -22,6 +23,9 @@ namespace osu.Game.Screens.Multi.Match
 
         [Resolved(typeof(Room), nameof(Room.Playlist))]
         protected BindableList<PlaylistItem> Playlist { get; private set; }
+
+        [Resolved]
+        private MusicController music { get; set; }
 
         [Resolved]
         private BeatmapManager beatmapManager { get; set; }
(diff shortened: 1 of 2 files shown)


bdach:
Yeah, this seems like an amicable solution to me. The current way of doing this was problematic at best (what ended up calling `Multiplayer`'s looping disable logic was a beatmap change inside the `MatchSongSelect` itself...)

bdach (dismissed):

osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -322,43 +302,10 @@ private void subScreenChanged(IScreen lastScreen, IScreen newScreen)
...
-            if (screenStack.CurrentScreen is RoomSubScreen)
-            {
-                var track = Beatmap.Value?.Track;
-
-                if (track != null)
-                {
-                    track.RestartPoint = Beatmap.Value.Metadata.PreviewTime;
-                    track.Looping = true;
-
-                    music?.EnsurePlayingSomething();
```
peppy, line 342: previously, this was run on each beatmap change, ensuring that music starts playing after a download completes, or the multiplayer screen is entered. your changes have removed this call so now it is silent.

peppy (changes requested):

osu.Game/Screens/Multi/Multiplayer.cs
frenzibyte: ah, thought tracks start automatically on beatmap changes, changed.

osu.Game/Screens/Multi/Match/RoomSubScreen.cs
```diff
@@ -37,6 +41,37 @@ protected override void LoadComplete()
 
             managerUpdated = beatmapManager.ItemUpdated.GetBoundCopy();
             managerUpdated.BindValueChanged(beatmapUpdated);
+
+            if (music != null)
+                music.TrackChanged += applyToTrack;
```
peppy, line 46: why are you doing this differently to the original code? why not just keep things simple and copy that code across?

by doing this you've called an event leak. there is no unbind of this event binding.

peppy (changes requested):

osu.Game/Screens/Multi/Match/RoomSubScreen.cs
frenzibyte: Ah, I forgot to unbind..., I intended to rewrite this such that I don't end up with extra useless code (because I changed the place the code was in), but looks like that's a bad practice, will copy exactly how the code was in `Multiplayer`, my bad.

bdach: I believe I've resolved this. A double check would be appreciated.

peppy (approved):

PR #11255, opened 2020-12-22
Fix caught objects disappearing when a replay is rewound

ekrctb:
Before this change, when a replay is rewound, caught objects were missing because caught objects are permanently removed after dropped from the catcher plate.
Also, objects that exploded from the plate were not properly removed and results in buggy floating objects.
Those issues are more noticeable in a catch editor (my WIP version). I thought solving this issue is important for the catch editor.

To manage caught objects efficiently, I used a `LifetimeEntryManager`. Each time an object is caught, a `CaughtObjectEntry` is created and added to `CaughtObjectContainer`. `CaughtObjectContainer` instantiates pooled `DrawableCaughtObject` from `CaughtObjectEntry` each time it became alive and remove when dead.
The logic is similar to the way `HitObjectContainer` uses `HitObjectLifetimeEntry`.

`CaughtObjectEntry` times number of hit objects are retained in memory for every gameplay, but the memory usage is not significant as I tested.


osu.Game.Rulesets.Catch/Objects/Drawables/IHasCatchObjectState.cs
```diff
@@ -12,7 +12,9 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
     /// </summary>
     public interface IHasCatchObjectState
     {
-        PalpableCatchHitObject HitObject { get; }
+        CatchObjectType ObjectType { get; }
+
+        double StartTime { get; }
```
bdach, line 17: Immediately I'm not sure I agree with this premise of having that data added to all catch DHOs and copied to the caught object model. Why isn't this just the hitobject itself? Hitobjects (the models) are not pooled after all. Can't the dropped objects just take a reference to the HOs?

The main reason that I'm opposed to this is that catch object types are encoded already, in the hitobject types themselves. Adding this `ObjectType` enum to boot seems like unnecessary duplication - this sort of stuff should really have a singular source of ground truth.

I'm not sure if it's worth to review the rest of this since this is a pretty central piece of this entire change, so abstaining with the rest for now.

ekrctb: The main motivation for removing `HitObject` reference was that its properties could change after the object is caught in the editor.
It turns out, however, in the current state of PR, `HitObject` properties being changed after caught doesn't cause issues. [The way](https://github.com/ppy/osu/blob/fa94703b9b663e443896a3971bfc2c2102350b93/osu.Game.Rulesets.Catch/Skinning/Default/CatchHitObjectPiece.cs#L51)`StartTime` currently used doesn't matter `StartTime` as long as it is after the current time. And the type of `HitObject` cannot be changed.
I'll make it references to a `HitObject` for now to make the PR simpler.

bdach: A bit hard to me to respond to that without knowing exactly in which circumstances the properties could change. Let's just cross that bridge when we get to it, but I'd probably honestly rather prefer cloning hitobjects to this myself.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+
+        private readonly DrawablePool<DrawableCaughtFruit> caughtFruitPool;
+        private readonly DrawablePool<DrawableCaughtBanana> caughtBananaPool;
+        private readonly DrawablePool<DrawableCaughtDroplet> caughtDropletPool;
+
+        /// <summary>
+        /// The randomness used to compute position in stack.
+        /// It is incremented for each <see cref="Add"/> call and decremented for each <see cref="Remove"/> call to make a replay consistent.
+        /// </summary>
+        private int randomSeed = 1;
```
bdach, line 38: Might as well seed by index in beatmap at this point to save on tracking this.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+        {
+            randomSeed++;
+
+            lifetimeManager.AddEntry(entry);
+
+            if (entry.State != CaughtObjectState.Stacked) return;
+
+            // `DropStackedObjects` may be called before lifetime update.
+            if (entry.LifetimeStart <= Time.Current)
+                aliveStackedObjects.Add(entry);
```
bdach, line 67: This seems slightly suspicious and potentially buggy to me. Can whatever this is fixing be fixed in any other way, like (wild guess) scheduling stacked object drop after children, for instance?

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+            removeDrawable(entry);
+
+            aliveStackedObjects.Remove(entry);
+
+            if (!dropEntryMap.TryGetValue(entry, out var dropEntry)) return;
+
+            dropEntryMap.Remove(entry);
+
+            lifetimeManager.RemoveEntry(dropEntry);
+            removeDrawable(dropEntry);
```
bdach, line 93: All of these fields are really hard to keep track of (what was added where and what needs to be cleaned up from where).

Would probably factor out four symmetrical methods, something like:

```csharp
public void addImmediateEntry(CaughtObjectEntry entry)
{
	lifetimeManager.AddEntry(entry);
	if (entry.State == CaughtObjectState.Stacked)
		addDelayedDropEntry(entry);
}

public void removeImmediateEntry(CaughtObjectEntry entry)
{
	lifetimeManager.RemoveEntry(entry);
	if (entry.State == CaughtObjectState.Stacked)
		removeDelayedDropEntry(entry);
}

public void addDelayedDropEntry(CaughtObjectEntry entry)
{
	var dropEntry = new CaughtObjectEntry(CaughtObjectState.Dropped, entry.PositionInStack, entry)
	{
		LifetimeStart = double.PositiveInfinity,
		LifetimeEnd = double.PositiveInfinity
	};

	lifetimeManager.AddEntry(dropEntry);
	dropEntryMap[entry] = dropEntry;
}

public void removeDelayedDropEntry(CaughtObjectEntry entry)
{
	dropEntryMap.Remove(entry, out var dropEntry);
	lifetimeManager.RemoveEntry(dropEntry);
}
```

I would hope that something can be done to eliminate the `aliveStackedObjects` adds/deletes here, such that that logic can be entirely located in `entryBecame{Alive,Dead}`.

Also not sure why `Remove()` is cleaning up drop entry drawables but `Add()` never adds them. If that is indeed required for some reason, then I'd leave a comment in `addDelayedDropEntry` as to why.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+            const float radius_div_2 = CatchHitObject.OBJECT_RADIUS / 2;
+            const float allowance = 10;
+
+            int iteration = 0;
+
+            while (aliveStackedObjects.Any(f => Vector2Extensions.Distance(f.PositionInStack, position) < (displayRadius + radius_div_2) / (allowance / 2)))
+            {
+                float diff = (displayRadius + radius_div_2) / allowance;
+
+                position.X += (StatelessRNG.NextSingle(randomSeed, iteration + 2) - 0.5f) * diff * 2;
```
bdach, line 107: guessing sequential values were intended here

```suggestion
                position.X += (StatelessRNG.NextSingle(randomSeed, iteration * 2) - 0.5f) * diff * 2;
```

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+            removeDrawable(entry);
+        }
+
+        private void dropStackedObjects(double time, Action<DrawableCaughtObject> applyTransforms, int mirrorDirection)
+        {
+            foreach (var entry in aliveStackedObjects)
+            {
+                entry.LifetimeEnd = time;
+
+                if (!dropEntryMap.TryGetValue(entry, out var dropEntry)) continue;
```
bdach, line 164: This seems like it should be an assertion rather than a soft check. Can this realistically happen and not be a bug?

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,232 @@
...
+                }
+            }
+
+            drawableMap[entry] = drawable;
+        }
+
+        private void removeDrawable(CaughtObjectEntry entry)
+        {
+            if (!drawableMap.TryGetValue(entry, out DrawableCaughtObject drawable))
+                return;
```
bdach, line 204: Can use `.Remove(entry, out var drawable)`, but again, I'd really be expecting this to never fail.

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -464,127 +469,62 @@ private void updateState(CatcherAnimationState state)
...
-
-                default:
-                    return null;
+            if (hitObject.HitObject is Droplet)
+            {
+                // droplet explodes immediately
+                entry = new CaughtObjectEntry(CaughtObjectState.Dropped, positionInStack, hitObject)
+                {
+                    ApplyTransforms = applyExplodeTransforms,
+                    DropPosition = CaughtObjectContainer.GetCurrentDropPosition(positionInStack),
```
bdach, line 492: Seems weird for the catcher to have to call the container's own method like this. Should probably be `CaughtObjectContainer`'s responsibility to calculate this (which should hopefully allow `GetCurrentDropPosition` to become private).

bdach (changes requested):

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
ekrctb: Multiple objects can be caught in one frame. `aliveStackObjects` affect stack positioning and we cannot just schedule the explosion to remove this code.
To update `aliveStackObjects` via `entryBecameAlive`, we need to call `lifetimeManager.Update` but this results in quadratic time complexity and is awkward (`LifetimeEntryManager` delays new entry to become alive for reasons).
My current stance is this way is the best.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
ekrctb: Because testing code doesn't set IndexInBeatmap, using it makes stacking ugly in testing.
It is fine to modify testing code, but I think tracking this value in `CaughtObjectContainer` (or maybe in `Catcher`) is better as it requires less preparation to use from outside.

ekrctb:
I moved all caught object creation logic to `CaughtObjectContainer`.
`CaughtObjectEntry` is now abstract and `CaughtObjectContainer` has `StackedObjectEntry` and `DroppedObjectEntry` private classes to store all internal data.
`CaughtObjectEntry` doesn't reference `DrawableCaughtObject` and it should apply transforms to a `Drawable` without knowing specifics of it being `DrawableCaughtObject`.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectEntry.cs
```diff
@@ -0,0 +1,72 @@
...
+
+        public Bindable<Color4> AccentColour { get; } = new Bindable<Color4>();
+
+        public Bindable<bool> HyperDash { get; } = new Bindable<bool>();
+
+        public Vector2 DisplaySize { get; }
+
+        public float DisplayRotation { get; }
+
+        public Bindable<FruitVisualRepresentation> VisualRepresentation { get; } = new Bindable<FruitVisualRepresentation>();
```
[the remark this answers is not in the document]
ekrctb: It is indeed possible to remove `VisualRepresentation` Bindable <https://github.com/ekrctb/osu/commit/e5ef1aefc0faeea03fcb4cb962042c559de4c834>.
However, it requires `IndexInBeatmap` bindable in `IHasCatchObjectState`.
Because `HitObject` referenced from `IHasObjectState` can change at anytime, a consumer cannot bind to `ObjectState.HitObject.SomeBindable` because it will become stale as soon as `ObjectState.HitObject` changes.

This is kind of bad design and I don't really like `IHasCatchObjectState` as it is very ad-hoc and there are still code duplication. Probably a better solution could be found in the future. But I don't touch it in this PR.
Also I made changes to made `CaughtObjectEntry` not implement `IHasCatchObjectState` so it is less relevant now.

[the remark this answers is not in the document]
bdach: Yeah, my original comment was more of a passing remark, not something to be addressed in this PR.

osu.Game.Rulesets.Catch/Objects/Drawables/IHasCatchObjectState.cs
```diff
@@ -12,7 +12,7 @@ namespace osu.Game.Rulesets.Catch.Objects.Drawables
     /// </summary>
     public interface IHasCatchObjectState
     {
-        PalpableCatchHitObject HitObject { get; }
+        CatchHitObject HitObject { get; }
```
bdach, line 15: Not sure where this generalisation is coming from - I think this should remain as `Palpable...`? This surely doesn't apply to banana showers or juice streams, right?

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCaughtObject.cs
```diff
@@ -41,23 +34,26 @@ protected CaughtObject(CatchSkinComponents skinComponent, Func<ISkinComponent, D
...
-            Scale = Vector2.Divide(objectState.DisplaySize, Size);
-            Rotation = objectState.DisplayRotation;
-            AccentColour.Value = objectState.AccentColour.Value;
-            HyperDash.Value = objectState.HyperDash.Value;
+            HitObject = entry.HitObject;
+
+            AccentColour.Value = entry.AccentColour;
+            HyperDash.Value = entry.HyperDash;
+
+            Scale = Vector2.Divide(entry.DisplaySize, Size) * 0.5f;
```
bdach, line 44: Where is this `0.5f` coming from? I'm not sure I saw it be there before.

osu.Game.Rulesets.Catch/Judgements/CatchJudgementResult.cs
```diff
@@ -20,6 +21,8 @@ public class CatchJudgementResult : JudgementResult
         /// </summary>
         public bool CatcherHyperDash;
 
+        public CaughtObjectEntry CaughtObjectEntry;
```
bdach, line 24: Not 100% happy with the layering/separation of concerns here... Should the judgement result really be aware of when/how it was caught via the entry? Maybe this would be better suited to the HO itself? Or be tracked somewhere external entirely, as I think was the case (dictionaries in `CaughtObjectContainer`)?

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,334 @@
...
+
+                case Droplet _:
+                    return caughtDropletPool.Get();
+
+                default:
+                    return null;
+            }
+        }
+
+        private class StackedObjectEntry : CaughtObjectEntry
```
bdach, line 259: Would split these nested classes/enums to separate files just to keep line count of this file down.

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,334 @@
...
+        /// </summary>
+        public CaughtObjectEntry AddStackObject(IHasCatchObjectState source, Vector2 positionInStack)
+        {
+            var delayedDropEntry = new DroppedObjectEntry(positionInStack, source);
+            var stackEntry = new StackedObjectEntry(positionInStack, delayedDropEntry, source);
+
+            addImmediateEntry(stackEntry);
+
+            // `DropStackedObjects` may be called before lifetime update.
+            if (stackEntry.LifetimeStart <= Clock.CurrentTime)
```
bdach, line 68: I don't know if it's just me but I'd prefer this conditional was flipped, as it feels more natural for me to have the instant being checked on the left side of the comparison against a start time of an interval (was part of the reason why I misread earlier and thought this was buggy).

osu.Game.Rulesets.Catch/Objects/Drawables/CaughtObjectContainer.cs
```diff
@@ -0,0 +1,334 @@
...
+
+                iteration++;
+            }
+
+            position.X = Math.Clamp(position.X, -CatcherArea.CATCHER_SIZE / 2, CatcherArea.CATCHER_SIZE / 2);
+
+            return position;
+        }
+
+        public void DropStackedObjects(DroppedObjectAnimation animation, int mirrorDirection)
```
bdach, line 127: nit: would unify up naming here (there's `AddStackObject` and `AddDropObject`, but `DropStackedObject`).

osu.Game.Rulesets.Catch/UI/Catcher.cs
```diff
@@ -464,127 +461,29 @@ private void updateState(CatcherAnimationState state)
...
-
-            if (caughtObject == null) return;
-
-            caughtObject.CopyStateFrom(drawableObject);
-            caughtObject.Anchor = Anchor.TopCentre;
-            caughtObject.Position = position;
-            caughtObject.Scale /= 2;
+            var entry = source.HitObject is Droplet
+                ? CaughtObjectContainer.AddDropObject(source, positionInStack, DroppedObjectAnimation.Explode, stackMirrorDirection)
+                : CaughtObjectContainer.AddStackObject(source, positionInStack);
```
bdach, line 468: This logic should probably be in `CaughtObjectContainer` (it's even duplicated in tests). Seems like there can possibly be only one method that accepts a caught HO, maybe.

bdach (changes requested):

osu.Game.Rulesets.Catch/Objects/Drawables/DrawableCaughtObject.cs
ekrctb: It was `caughtObject.Scale /= 2` <https://github.com/ppy/osu/pull/11255/files#diff-c3ab09080a53b7f21d7ceffeaf2e5bfd4ab7ce5c5262696ea8d725197f87c640L476>.

osu.Game.Rulesets.Catch/Judgements/CatchJudgementResult.cs
ekrctb: A caught object is a result of a judgment (catching) so maintaining the caught object entry makes sense to me.
I cannot be a property of the HO because a hit object can be associated with multiple judgment results (not currently supported in `Playfield`, but used in tests, and I think it makes sense conceptually).
A caught object could be associated with `HitObjectLifetimeEntry` but there is no code path to modify this object before #11262.

bdach: Do you have a preview or other loose branch of how this would look with that mentioned pull? That might provide better rationale for having that merged.

ekrctb: I wrote a prototype of associating `CaughtObjectEntry` with `HitObjectLifetimeEntry` <https://github.com/ekrctb/osu/commit/392d048fe2e73236262427dc5bf0667b9b0ad31a>.
I think this way is not great. It need to expose `LifetimeEntry` in DHO to make it accessible.

ekrctb:
It will be rewritten.

PR #11256, opened 2020-12-22
Disallow joining ended rooms in realtime multiplayer

frenzibyte:


osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
```diff
@@ -115,7 +115,9 @@ private void addRooms(IEnumerable<Room> rooms)
                     {
                         if (room == selectedRoom.Value)
                         {
-                            joinSelected();
+                            if (roomManager.CanJoinRoom(room))
+                                joinSelected();
```
bdach, line 119: I'm not convinced by the way this is done at all. This logic is all very far from `RoomManager`, wherein the request is being fired - it should be in `JoinRoom()` in my opinion.

There are a few ways to implement that, and I think that `CanJoinRoom` doesn't even maybe need to exist - can probably just override `RealtimeRoomManager`'s impl, like so:

```csharp
public override void JoinRoom(Room room, Action<Room> onSuccess = null, Action<string> onError = null)
{
    if (room.Status.Value is RoomStatusEnded)
    {
        onError?.Invoke("Cannot join room as it has already ended.");
        return;
    }

    base.JoinRoom(room, r => joinMultiplayerRoom(r, onSuccess, onError), onError);
}
```

bdach (changes requested):

osu.Game/Screens/Multi/Lounge/Components/RoomsContainer.cs
frenzibyte: Ah, was thinking `onError` is not the right way for this as this case is more of guaranteed while other cases of invoking should not be possible to happen, will change back...

peppy:
I don't get it. What is this doing? The server already disallows this from happening right?

frenzibyte:
Yeah but now it early checks for that and returns, without logging as an error and otherwise, thought this would be more correct than before now?

peppy:
I don't see the point.

frenzibyte:
Previously I made a method that checks logically whether it should be fine to join the room or not, to avoid making the join request at all otherwise, but https://github.com/ppy/osu/pull/11256#discussion_r547531356 tells that it's not really the way here.

peppy:
Well it's a bit unfortunate how the flow of clicking rooms happens in the lounge right now (hard to handle in a more appropriate place). Maybe this is fine for now.

peppy (approved):

PR #11259, opened 2020-12-22
Change vertical alignment of scroll target when clicking settings category buttons

### LINKED ISSUE #9196: Settings menu button click not highlighted correctly

LittleEndu:
Closes #9196
Supersedes #11019  

## Changes

 * Adds test coverage
 * Scroll to 20% lower
   * When scrolled all the way to top is a special case now
   * Manually check if scrolled all the way to bottom because current code doesn't animate the highlighting
 * Consider section intersecting at 25% as selected
   * This is so that scrolling upwards just 1 pixel doesn't instantly unselect the section
 * Prioritize the section user clicked on when scrolled to extremes
   * For this we need to know if the container scrolled to the section or did the user, that's done in ``UserTrackingScrollContainer``
      * Code for this already exists in ``BeatmapCarousel`` so it's extracted from there


osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -194,18 +212,27 @@ protected override void UpdateAfterChildren()
                 headerBackgroundContainer.Height = (ExpandableHeader?.LayoutSize.Y ?? 0) + (FixedHeader?.LayoutSize.Y ?? 0);
                 headerBackgroundContainer.Y = ExpandableHeader?.Y ?? 0;
 
-                float scrollOffset = FixedHeader?.LayoutSize.Y ?? 0;
+                // scroll offset is our fixed header height if we have it plus 20% of content height
+                // plus 5% to fix floating point errors and to not have a section instantly unselect when scrolling upwards
+                // but the 5% can't be bigger than our smallest section height, otherwise it won't get selected correctly
+                float sectionOrContent = Math.Min(smallestSection?.Height / 2.0f ?? 0, scrollContainer.DisplayableContent * 0.05f);
+                float scrollOffset = (FixedHeader?.LayoutSize.Y ?? 0) + scrollContainer.DisplayableContent * scroll_target_multiplier + sectionOrContent;
```
bdach, line 219: This `(FixedHeader?.LayoutSize.Y ?? 0) + scrollContainer.DisplayableContent * scroll_target_multiplier` factor shows up in at least 2 places as far as I can see. It really should be factored out to its own property (`scrollTarget`?)

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -194,18 +212,27 @@ protected override void UpdateAfterChildren()
...
                 Func<T, float> diff = section => scrollContainer.GetChildPosInContent(section) - currentScroll - scrollOffset;
 
-                if (scrollContainer.IsScrolledToEnd())
+                if (Precision.AlmostBigger(0, scrollContainer.Current))
```
bdach, line 222: `scrollContainer.IsScrolledToStart()` should really be a thing at this point (`ScorePanelList` has a similar local variant).

osu.Game/Graphics/Containers/SectionsContainer.cs
```diff
@@ -194,18 +212,27 @@ protected override void UpdateAfterChildren()
...
                 }
-                else
+
+                if (Precision.AlmostBigger(scrollContainer.Current, scrollContainer.ScrollableExtent))
                 {
-                    SelectedSection.Value = Children.TakeWhile(section => diff(section) <= 0).LastOrDefault()
-                                            ?? Children.FirstOrDefault();
+                    SelectedSection.Value = lastClickedSection as T ?? Children.LastOrDefault();
+                    return;
                 }
```
bdach, line 232: why is this a thing and why isn't it using `IsScrolledToEnd()`? is this related to

> Manually check if scrolled all the way to bottom because current code doesn't animate the highlighting

and if so, mind expounding on that? because I don't understand the above remark

bdach (dismissed):

osu.Game/Graphics/Containers/SectionsContainer.cs
LittleEndu: ``IsScrolledToEnd()`` uses the target scroll position instead of the current one. So if you scroll from first section to middle section, the highlighting will follow the movement, but if you scroll from first to last, the highlighting will instantly jump to last section because ``IsScrolledToEnd()`` returns true.

peppy:
I've refactored this to a point I can understand what's going on. Seems to be good enough to get this in.

peppy (dismissed):

bdach:
The "20% lower" figure seems dicey at best seeing that situations like this are apparently possible:

2021-01-21-233455_1920x1138_scrot

The section is pretty much halfway down the scrollbar - works decent enough for user scrolling, but it's a bit weird when clicking if you ask me. I think `scroll_y_centre = 0.1f` works a bit better but not sure. Feel free to merge past me if you disagree.

Source looks saner to me, so going to dismiss my review on that basis.

peppy:
Sure let's go with 10%. I think it feels fine (and matches closer to stable, visually).

peppy (dismissed):

bdach (approved):
let's hopefully get this over with

PR #11262, opened 2020-12-23
Remove nested hit object lifetime entries from playfield when parent hit object is removed or edited

### LINKED ISSUE #11261: Nested hit objects are leaked when parent hit object is edited or deleted
1. Inspect `Playfield.lifetimeEntryMap.Count` while editing a slider shape.
2. See the number is increasing rapidly.

Nested hit objects should be removed when the parent hit object is removed or when the nested hit objects are recreated by `ApplyDefaults`.

It is a visual issue for osu!catch editor because new nested hit objects will be stacked every frame while the parent object is edited, resulting in a massive amount of stacked objects.

ekrctb:
Fix #11261.

I added `OnHitObjectLifetimeEntryAdded` and `OnHitObjectLifetimeEntryRemoved`. It will be used in #11255 (caught object should be removed when hit object is removed).


osu.Game/Rulesets/Objects/HitObject.cs
```diff
@@ -35,6 +35,11 @@ public class HitObject
         /// </summary>
         public event Action<HitObject> DefaultsApplied;
 
+        /// <summary>
+        /// Invoked before <see cref="NestedHitObjects"/> are cleared.
+        /// </summary>
+        internal event Action<HitObject> RemoveNestedHitObjects;
```
bdach, line 41: I'm not a huge fan of this event either. Honestly wondering whether just subscribing to `DefaultsApplied` and cleaning up all of the old entries manually wouldn't be better.

Or, heck, making `NestedHitObjects` a `BindableList`, but that sounds like it might not be very pleasant at all...

smoogipoo: Yeah, I really don't like this either. There has to be a better way...

smoogipoo:
Does anything break if we simply don't add entries to the map in `Playfield.GetPooledDrawableRepresentation()`?

The only think I can see is maybe `KeepAllAlive()`, but I'm not sure this should be handled by that anyway...

ekrctb:
If nested hit objects entries are not added to the map, `Result` associated to the nested hit objects will be lost when a replay is rewound.

(I don't like this PR either, I'm trying to find other ways)

smoogipoo:
As this is all very internal stuff, maybe it's fine to bring it in as a hack to fix the larger issue for the time being? I'd rather an `internal` hack that requires no implementation from ruleset authors than leaking memory...

Maybe with a `//Todo: This is a hack for nested poolable hitobjects to have their entries cleaned up from the Playfield`.

ekrctb:
I have an another working solution in <https://github.com/ppy/osu/compare/master...ekrctb:hitobject-entry-nesting?expand=1>.
The idea is maintaining all parent-child hit object relationship at `Playfield`, mirroring `HitObject.NestedObjects` tree structure.
However, I'm not really sure such duplication is needed, and performance implications. It *might* be a direction towards nested hit object loading lag issue #12659, though.
Also it is not tested against non-pooled cases yet.

peppy:
This seems like something we still want to consider.

peppy:
@smoogipoo wondering if you have any better ideas for addressing this. I haven't looked into it in detail yet (and haven't check on the proposal in https://github.com/ppy/osu/pull/11262#issuecomment-875531279) but the version in this PR still reads pretty bad to me.

smoogipoo:
I don't have any better ideas at this time.

peppy:
Okay, I'll revisit this today and see where it leads.

peppy:
Looking at this with fresh eyes, the solution provided in https://github.com/ppy/osu/compare/master...ekrctb:hitobject-entry-nesting?expand=1 does seem better from a readability perspective. I'm also unsure about performance implications (mainly the event binding and potential for recycling lifetime mappings if `ApplyDefaults` is called after add to playfield in a non-editor scenario), but will do some testing.

peppy:
https://github.com/ppy/osu/pull/20303

PR #11263, opened 2020-12-23
Fix flashlight not updating its position during replay rewinding

### LINKED ISSUE #11260: Flashlight position in osu!standard doesn't update when rewinding replay

https://user-images.githubusercontent.com/35318437/102945025-6fa4da00-4471-11eb-856c-2b0bb34c51cc.mp4

**osu!lazer version:** since implementation of smoothing (https://github.com/ppy/osu/pull/4771)

peppy:
Closes #11260.

### DIFF
--- a/osu.Game.Rulesets.Osu/Mods/OsuModFlashlight.cs
+++ b/osu.Game.Rulesets.Osu/Mods/OsuModFlashlight.cs
@@ -56,7 +56,7 @@ namespace osu.Game.Rulesets.Osu.Mods
                 var destination = e.MousePosition;
 
                 FlashlightPosition = Interpolation.ValueAt(
-                    Math.Clamp(Clock.ElapsedFrameTime, 0, follow_delay), position, destination, 0, follow_delay, Easing.Out);
+                    Math.Clamp(Math.Abs(Clock.ElapsedFrameTime), 0, follow_delay), position, destination, 0, follow_delay, Easing.Out);
 
                 return base.OnMouseMove(e);
             }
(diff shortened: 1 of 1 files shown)


osu.Game.Rulesets.Osu/Mods/OsuModFlashlight.cs
```diff
@@ -56,7 +56,7 @@ protected override bool OnMouseMove(MouseMoveEvent e)
                 var destination = e.MousePosition;
 
                 FlashlightPosition = Interpolation.ValueAt(
-                    Math.Clamp(Clock.ElapsedFrameTime, 0, follow_delay), position, destination, 0, follow_delay, Easing.Out);
+                    Math.Clamp(Math.Abs(Clock.ElapsedFrameTime), 0, follow_delay), position, destination, 0, follow_delay, Easing.Out);
```
Joehuu, line 59:
```suggestion
                    Math.Min(Math.Abs(Clock.ElapsedFrameTime), follow_delay), position, destination, 0, follow_delay, Easing.Out);
```

bdach: Hm, I guess that's a correct optimisation to make, although it's a bit of a clever one and clever code isn't always the best code... @peppy thoughts?

peppy: i'm fine either way. makes sense that the lower clamp isn't required due to abs.

Issue #3530, opened 2018-10-03
Auto play on CTB has bugs


lolita-is-godddd:
With seeking, acc is not 100%, score doesn't decrease, and game doesn't erase sprite (fruit).
Build is "2018.928.0".
![osu_2018-10-03_21-51-39](https://user-images.githubusercontent.com/25249461/46411469-de919d80-c756-11e8-8315-2b89b43b724d.png)
[for debug.log](https://github.com/ppy/osu/files/2442043/debug.log)

Hanamuke:
I've already opened PR #3518 to fix, autoplay accuracy. It still miss some fruits on 8*+ maps (1 miss on speed of link for instance), and the fruits are still buggy on rewind, but those are rendering issues.

Hanamuke:
I'm expect there will still be problems on maps with stacked objects, i may right some further fixing, but that will be significantly more difficult to fix

lolita-is-godddd:
thanks for work

lolita-is-godddd:
also, speed up makes miss even catch correctly

Joehuu:
Fixed in https://github.com/ppy/osu/pull/6032, not sure about the linked PR now.

Issue #6150, opened 2019-09-18
FramedReplayHandler doesn't respect direction for the current and next frame


smoogipoo:
From the middle of two frames - 0 -> 1, during rewind the "current" frame remains 0 and the "next" frame remains 1, although they should be reversed.

There's a few other edge cases to consider, such as rewinding from before the first frame, playing back normally from before the first frame, and vice versa for the last frame.

I've already done work on this in https://github.com/ppy/osu/compare/master...smoogipoo:fix-framedreplayhandler which should already be mostly working, but the test cases need a lot of work.

PR #11264, opened 2020-12-23
Fix CurrentTime is not always between CurrentFrame.Time and NextFrame.Time

ekrctb:
It should be always:
- `CurrentFrame.Time <= CurrentTime <= NextFrame.Time` when direction == 1
- `CurrentFrame.Time >= CurrentTime >= NextFrame.Time` when direction == -1
but the invariant was broken when the direction changed (a replay is rewound).

The issue is most visible in osu!catch ruleset, and it can be seen as the catcher goes outside of the playfield when an autoplay is rewound, due to the low number of autoplay frames, but the issue is present on all rulesets.

Also, removed assumption of `CompareTo` always returns {-1,0,1}. The document says only sign matters, and it can be any integer.

- <s>Fixes/ #3530</s> (There are multiple issues reported there, but the main issue "acc is not 100%" is solved here).
  - It seems like there are more ways a replay misses hit objects.

### DIFF
--- a/osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
+++ b/osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
@@ -123,9 +123,7 @@ namespace osu.Game.Rulesets.Replays
             // if we have a next frame, check if it is before or at the current time in playback, and advance time to it if so.
             if (next != null)
             {
-                int compare = time.CompareTo(next.Time);
-
-                if (compare == 0 || compare == currentDirection)
+                if (currentDirection == 1 ? next.Time <= time : time <= next.Time)
                 {
                     currentFrameIndex = clampedNextFrameIndex;
                     return CurrentTime = CurrentFrame.Time;
@@ -164,12 +162,15 @@ namespace osu.Game.Rulesets.Replays
             if (!CurrentTime.HasValue)
             {
                 currentDirection = 1;
+                return;
             }
-            else
-            {
-                currentDirection = time.CompareTo(CurrentTime);
-                if (currentDirection == 0) currentDirection = 1;
-            }
+
+            var newDirection = CurrentTime <= time ? 1 : -1;
+
+            if (currentDirection == newDirection) return;
+
+            currentFrameIndex = clampedNextFrameIndex;
+            currentDirection = newDirection;
         }
     }
 }


ekrctb:
So tests are failing.

- `System.ArgumentException : '0' cannot be greater than -1.`
This can be just solved by boundary checking.

- `FramedReplayInputHandlerTest` and `StreamingFramedReplayInputHandlerTest`: I missed this tests. I'm now not sure about my change.

- `TestSceneAutoplay`: the key counter is broken. All four ruleset's `FramedReplayInputHandler` use `CurrentFrame` to populate `Actions`, but I'm not sure why this worked. Shouldn't it use `NextFrame` when it is playing backward? When `next.Time < current.Time`, the frame should represent time interval `[next.Time, current.Time]` (right?).

I'm not sure what is the intended semantics of `FramedReplayInputHandler`. It is kind of complicated with many nullable values. And I have no idea about "important section" at all...

bdach:
It might be easier to understand what this is aiming to do if you gave some larger context behind why you decided to make this change in particular. Does this show in normal replay playback, or editor? I'd like to think frame stability should prevent this from popping up in replays.

frenzibyte:
Bringing https://github.com/ppy/osu/issues/6150 up for visibility as it sounds related to what is being achieved here.

ekrctb:
Indeed the #6150 is the issue (I should have searched).
The issue occur in replay and editor.

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
```diff
@@ -123,9 +123,7 @@ private bool inImportantSection
             // if we have a next frame, check if it is before or at the current time in playback, and advance time to it if so.
             if (next != null)
             {
-                int compare = time.CompareTo(next.Time);
-
-                if (compare == 0 || compare == currentDirection)
+                if (currentDirection == 1 ? next.Time <= time : time <= next.Time)
```
bdach, line 126: Honestly I'd prefer to revert this and stick `Math.Sign` on the comparison result if anything. The way this is written makes my head hurt worse than the original (which I've read a few times already and it's already quite a headache).

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
```diff
@@ -164,12 +162,15 @@ private void updateDirection(double time)
...
-            {
-                currentDirection = time.CompareTo(CurrentTime);
-                if (currentDirection == 0) currentDirection = 1;
-            }
+
+            var newDirection = CurrentTime <= time ? 1 : -1;
+
+            if (currentDirection == newDirection) return;
+
+            currentFrameIndex = clampedNextFrameIndex;
```
bdach, line 172: So as far as I can tell, this current frame index change is the only actual functional change in this PR, correct? Can you write a test case that you would expect to fail without this and maybe we can proceed from there?

This replay stuff is super fragile and touching it without adding tests is just asking for trouble further down the line.

bdach (changes requested):

ekrctb:
It turned out the issue needs more subtle code change to fix.

osu.Game/Rulesets/Replays/FramedReplayInputHandler.cs
ekrctb: I agree and the PR was rushed and I didn't even understand the original code.
I'll probably PR more substantial rewrite in the next year.

PR #11266, opened 2020-12-23
Fix multiple cases of unobserved async errors being thrown in fire-and-forget usages

peppy:
This should (hopefully) cover all the cases where a SignalR exception could arrive as an unobserved exception due to there being no `await` or `ContinueWith` on an async call.

- [x] Depends on #11265 for ease of merging.

### DIFF
--- a/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMatchSongSelect.cs
+++ b/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMatchSongSelect.cs
@@ -58,7 +58,7 @@ namespace osu.Game.Screens.Multi.RealtimeMultiplayer
 
                 client.ChangeSettings(item: item).ContinueWith(t =>
                 {
-                    return Schedule(() =>
+                    Schedule(() =>
                     {
                         loadingLayer.Hide();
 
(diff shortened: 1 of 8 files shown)


osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMatchSongSelect.cs
```diff
@@ -58,7 +58,7 @@ protected override bool OnStart()
 
                 client.ChangeSettings(item: item).ContinueWith(t =>
                 {
-                    return Schedule(() =>
+                    Schedule(() =>
```
bdach, line 61: Is there any semantic difference with this change, other than that the underlying task changes type (from `Task<ScheduledDelegate>` to `Task`)? Not really bothered either way, just curious.

bdach:
I've also resolved conflicts here. Was a bit less straightforward but still not too bad, but double check would be appreciated.

osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMatchSongSelect.cs
peppy: just couldn't see a reason fort he returning of the schedule.

bdach (dismissed):

bdach (approved):

peppy:
Post-merge looks fine.

PR #11267, opened 2020-12-23
Move player constructor configuration to dedicated class; add AllowRestart parameter

peppy:


osu.Game/Screens/Play/Player.cs
```diff
@@ -317,81 +313,90 @@ private Drawable createGameplayComponents(WorkingBeatmap working, IBeatmap playa
...
-                    OnRetry = Restart,
-                    OnQuit = performUserRequestedExit,
-                },
-                new HotkeyRetryOverlay
-                {
-                    Action = () =>
+                    FailOverlay = new FailOverlay
                     {
-                        if (!this.IsCurrentScreen()) return;
+                        OnRetry = Restart,
```
bdach, line 360: I know both fail and retry are disabled in realtime right now so this should never be the case, but the retry button shouldn't probably even appear on the fail overlay in this case - maybe leave as TODO for later?

osu.Game/Screens/Ranking/ResultsScreen.cs
```diff
@@ -164,15 +164,18 @@ private void load()
             {
                 buttons.Add(new RetryButton { Width = 300 });
```
bdach, line 165: So I was a little surprised that this doesn't need its own treatment compared to `HotkeyRetryOverlay`, and as it turns out `ResultsScreen` has its own `allowRetry` parameter... It should probably be used instead of reading player's? Not 100% sure, it's a little messy.

osu.Game/Screens/Play/Player.cs
peppy: I'm not sure failing is ever going to be a thing in multiplayer. Probably fine to leave until we add such a flow.

bdach (approved):
looks correct :+1:

PR #11268, opened 2020-12-23
Fix realtime multiplayer song select getting stuck after selecting invalid beatmap

bdach:
- [x] Depends on #11266 for the error handling.


peppy:
Your fix seems fine. I've pushed some outstanding changes I had which are *mostly* just code quality improvements to the slightly related method. Check you're okay with it.

bdach:
More than fine, I was actually thinking about doing the same but concluded the race is unlikely and left it to be for now. Definitely agreeing with doing that, though.

peppy (approved):

Issue #10423, opened 2020-10-08
Add support for opening osu:// scheme links


Moltanicaa:
**Describe the new feature:** 

If the implementation isn't difficult, why not?

It's about this button
![IMG_20201008_161321.jpg](https://user-images.githubusercontent.com/59089206/95463243-389a4e80-0981-11eb-925e-b5bdd34e7265.jpg)

rawr-x3:
I can’t speak on Android cause it does the same on iOS, but for computer, you may have chosen osu!lazer as the directory for direct downloads, not stable.

Gabixel:
It doesn't really work like that: the osu!direct button launches osu! with some special arguments that tell him to open the in-game osu!direct with that specific map. It doesn't have much sense to move it to lazer (for now) since it's still under development and users probably want to launch the stable version with that. I guess this is a low priority for now.

ThatsNiceGuy:
`osu://` links already have worked on iOS for a long time. (Open new tab in safari, type in `osu://` into address bar, press enter, Safari will prompt you to open link in the game)

Joehuu:
Yeah, reason why I didn't put an iOS tag on this issue.

jamesnicolas:
I found this recent-ish forum post
https://osu.ppy.sh/community/forums/topics/1936976?n=1
which kind of implies osu!direct links work currenntly. Is that true? I can't seem to get osu:// links to work on linux. I also tried `xdg-open osu://`. Just wondering if there's a setting or configuration I'm missing or it just hasn't been implemented yet.

smallketchup82:
`osu://` links are set up by default on windows. but i cannot speak for macos or linux.

i've tried it myself, setting up mime types and everything, but also can't seem to get it to work. not entirely sure what the issue is but my guess is that lazer isnt handling the url protocol on platforms other than windows. i'll give it a look and see where the problem lies.

smallketchup82:
found the fix. it has to do with your osu.desktop file. for reference, here's mine:

```
[Desktop Entry]
Name=osu!(lazer)
Exec=gamemoderun /home/smallketchup/Desktop/osu.AppImage %u
Type=Application
Icon=lutris_osu
Comment=osu!(lazer)
Categories=Game
Terminal=false
StartupNotify=true
Hidden=false
NoDisplay=false
MimeType=x-scheme-handler/osu
```

[...]

PR #11269, opened 2020-12-23
Handle osu! URIs on Android

Game4all:
This should resolve the Android part of #10423 and should put Android on par with iOS in terms of integration.

# Summary
This PR adds the ability to handle URI links with the `osu://` scheme on Android. This also sets the `LaunchMode` of the osu! game activity to `SingleInstance` to spawn the activity in a separate task.

# Testing
Done on two devices running android  > 9.0, shouldn't have any problems with versions <= 7.0.
Can easily be tested by pasting some `<a>` tags to an html file with `href` set to an osu! URI.

I've put two links to test behaviour in a html file to be copied directly to the device, zipped under this line because github doesn't accept directly HTML files.
[testing_osu_links.zip](https://github.com/ppy/osu/files/5735689/testing_osu_links.zip)

### DIFF
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -14,7 +14,7 @@ using osu.Framework.Android;
 
 namespace osu.Android
 {
-    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false)]
+    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false, LaunchMode = LaunchMode.SingleInstance)]
     [IntentFilter(new[] { Intent.ActionDefault, Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataPathPatterns = new[] { ".*\\.osz", ".*\\.osk" }, DataMimeType = "application/*")]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryBrowsable, Intent.CategoryDefault }, DataScheme = osu_url_scheme)]
     public class OsuGameActivity : AndroidGameActivity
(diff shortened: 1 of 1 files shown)


osu.Android/OsuGameActivity.cs
```diff
@@ -14,10 +14,13 @@
 
 namespace osu.Android
 {
-    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false)]
+    [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false, LaunchMode = LaunchMode.SingleInstance)]
     [IntentFilter(new[] { Intent.ActionDefault, Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataPathPatterns = new[] { ".*\\.osz", ".*\\.osk" }, DataMimeType = "application/*")]
+    [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryBrowsable, Intent.CategoryDefault }, DataScheme = osu_url_scheme)]
```
bdach, line 19: iOS [also routes `osump://` scheme links](https://github.com/ppy/osu/blob/3087aba5bc359c5a23075882d89578c3f2fd45aa/osu.iOS/Info.plist#L120-L123). While the game [can't handle them yet](https://github.com/ppy/osu/blob/fa94703b9b663e443896a3971bfc2c2102350b93/osu.Game/OsuGame.cs#L270-L278), might as well add them while we're here for full parity.

osu.Android/OsuGameActivity.cs
```diff
@@ -49,6 +52,8 @@ private void handleIntent(Intent intent)
                 case Intent.ActionDefault:
                     if (intent.Scheme == ContentResolver.SchemeContent)
                         handleImportFromUri(intent.Data);
+                    else if (intent.Scheme == osu_url_scheme)
+                        Task.Run(() => game.HandleLink(intent.DataString));
```
bdach, line 56: I'm not sure whether it's really all that required to `Task.Run` a single schedule, but iOS sort of does the same thing for some reason... Was there a reason for this?

osu.Android/OsuGameActivity.cs
Game4all: whoops missed that one, nice catch

osu.Android/OsuGameActivity.cs
Game4all: I looked up commit history in the iOS project as well as the related PRs to find a clue why the link handling was wrapped in a task but couldn't find any information on it , and so I decided to wrap the call, in default of not having any clue why it had been done so in iOS, just in case. But yeah not sure either it is required since it is anyway scheduled further down in HandleLink() method itself so guess it'll cause no harm to directly call HandleLink().

peppy:
Let's get this in. Should be harmless enough.

PR #11271, opened 2020-12-23
Prefer connecting to dev server when running in DEBUG

peppy:
Minimum implementation, meant for visibility. Needs further discussion on how to provide this configuration (probably not via #if and probably all in one place).

This server will require registering a new account (from within lazer). It will allow playing / testing multiplayer and everything else, but some features like beatmap download will fail due to a limited deployment.

### DIFF
--- a/osu.Game/Online/API/Requests/Responses/APIChangelogEntry.cs
+++ b/osu.Game/Online/API/Requests/Responses/APIChangelogEntry.cs
@@ -48,6 +48,7 @@ namespace osu.Game.Online.API.Requests.Responses
     public enum ChangelogEntryType
     {
         Add,
-        Fix
+        Fix,
+        Misc
     }
 }
--- a/osu.Game/Overlays/Changelog/ChangelogBuild.cs
+++ b/osu.Game/Overlays/Changelog/ChangelogBuild.cs
@@ -131,33 +131,36 @@ namespace osu.Game.Overlays.Changelog
                         t.Padding = new MarginPadding { Left = 10 };
                     });
 
-                    if (entry.GithubUser.UserId != null)
+                    if (entry.GithubUser != null)
                     {
-                        title.AddUserLink(new User
+                        if (entry.GithubUser.UserId != null)
                         {
-                            Username = entry.GithubUser.OsuUsername,
-                            Id = entry.GithubUser.UserId.Value
-                        }, t =>
+                            title.AddUserLink(new User
+                            {
+                                Username = entry.GithubUser.OsuUsername,
+                                Id = entry.GithubUser.UserId.Value
+                            }, t =>
+                            {
+                                t.Font = fontMedium;
+                                t.Colour = entryColour;
+                            });
+                        }
+                        else if (entry.GithubUser.GithubUrl != null)
                         {
-                            t.Font = fontMedium;
-                            t.Colour = entryColour;
-                        });
-                    }
-                    else if (entry.GithubUser.GithubUrl != null)
-                    {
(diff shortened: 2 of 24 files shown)


bdach:
I'm somewhat certain that the typical .NET-Core-ish way to do this is an `appsettings.json`. Would probably be something like:

```json
{
	"api": {
		"endpoint": "https://dev.ppy.sh",
		"client_secret": "3LP2mhUrV89xxzD1YKNndXHEhWWCRLPNKioZ9ymT",
		"client_id": 5
	},
	"spectator": {
		"endpoint": "https://dev.ppy.sh/spectator"
	},
	"multiplayer": {
		"endpoint": "https://dev.ppy.sh/multiplayer"
	}
}
```

Of course there would be a corresponding `appsettings.Release.json`. Not sure how to jam that into our setup yet as we don't have the typical DI ASP.NET Core does, but throwing this out as a format suggestion first and foremost before digging further in that direction.

bdach:
On checking there is a Microsoft-endorsed way to use that format for console apps, but it [somewhat hijacks `Program.Main`](https://docs.microsoft.com/en-us/dotnet/core/extensions/configuration)... Not sure whether we want to go for that in that case?

bdach:
Here's a rough draft branch: https://github.com/ppy/osu/compare/master...bdach:dev-server-config-working

Can be improved in places, generally meant as a MVP. Seems to launch fine in both configs; API doesn't work in debug for what I presume to be obvious reasons (different user set?)

ThePooN:
Yes, it's a completely isolated environment. You need to create a seperate account via lazer.

bdach:
After some discussion in discord I think I'd agree with the sentiment expressed therein of not forcing this always for debug, but instead having alternative run configs for dev/production osu-web instances, just so you can switch on a dime. I wouldn't mind all that much if the two instances were full parity, but they aren't, and i.e. debugging potential osu!direct download issues on release builds could end up being a pain.

Both this PR in its current state and my draft thing can be adapted to that model by just choosing a different arbitrary compilation symbol to switch to, I guess.

peppy:
The intended direction is to always use this for debug. Anything missing can be added.

osu.Game/Online/API/Requests/Responses/APIChangelogEntry.cs
```diff
@@ -48,6 +48,7 @@ public class APIChangelogEntry
     public enum ChangelogEntryType
     {
         Add,
-        Fix
+        Fix,
+        Misc
```
bdach, line 52: sort of a drive-by change, but i'll look away for that one

peppy: This is actually a valid type, just haven't used it in years.

bdach: Yeah I checked web, just saying that it's not hugely related to the main purpose of the PR.

osu.Game/Overlays/Changelog/ChangelogBuild.cs
```diff
@@ -131,33 +131,36 @@ private void load(OsuColour colours, OverlayColourProvider colourProvider)
...
-                            t.Font = fontMedium;
-                            t.Colour = entryColour;
-                        });
+                            title.AddText(entry.GithubUser.DisplayName, t =>
+                            {
+                                t.Font = fontMedium;
+                                t.Colour = entryColour;
+                            });
+                        }
                     }
```
bdach, line 164: This is super ugly but it wasn't pretty before, and refactoring this isn't in scope of this PR, so again, probably going to look away but open an issue for this and the disabled test.

bdach (approved):

PR #11273, opened 2020-12-23
Improve reliability of handling user joined messages

bdach:
Fixes intermittent test failures as seen in https://ci.appveyor.com/project/peppy/osu/builds/36971822/tests (one instance of many).

`TestSceneRealtimeReadyButton` was manually adding `API.LocalUser`, which wasn't actually needed. The base `RealtimeMultiplayerTestScene` by default creates a new room as `API.LocalUser`, therefore automatically adding that user to the room - and as such there is no need to add them manually unless the `joinRoom` ctor param is specified as `false`.

While those test failures were a shortcoming of the test (which is fixed with the first commit here), they exposed a potential vulnerable point of the multiplayer client logic. In case of unreliable message delivery it is not unreasonable that duplicate messages might arrive, in which case the same scenario that failed in the tests could crash the game.

To ensure that is not the case, explicitly screen each new joined user against the room user list, to ensure that duplicates do not show up. `UserLeft` is already tolerant in that respect (if a user is requested to be removed twice by the server, the second removal just won't do anything).

I've also added test coverage for such edge-case scenarios, mostly for posterity.

### DIFF
--- a/osu.Game/Tests/Visual/RealtimeMultiplayer/TestRealtimeMultiplayerClient.cs
+++ b/osu.Game/Tests/Visual/RealtimeMultiplayer/TestRealtimeMultiplayerClient.cs
@@ -31,7 +31,7 @@ namespace osu.Game.Tests.Visual.RealtimeMultiplayer
         public void RemoveUser(User user)
         {
             Debug.Assert(Room != null);
-            ((IMultiplayerClient)this).UserLeft(Room.Users.Single(u => u.User == user));
+            ((IMultiplayerClient)this).UserLeft(new MultiplayerRoomUser(user.Id));
 
             Schedule(() =>
             {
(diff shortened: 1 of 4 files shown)


osu.Game/Tests/Visual/RealtimeMultiplayer/TestRealtimeMultiplayerClient.cs
```diff
@@ -31,7 +31,7 @@ public class TestRealtimeMultiplayerClient : StatefulMultiplayerClient
         public void RemoveUser(User user)
         {
             Debug.Assert(Room != null);
-            ((IMultiplayerClient)this).UserLeft(Room.Users.Single(u => u.User == user));
+            ((IMultiplayerClient)this).UserLeft(new MultiplayerRoomUser(user.Id));
```
peppy, line 34: it this change required? seems like the previous logic is more correct (obviously this is a test so it doesn't matter so much, but this will potentially be providing unfilled details and break reference comparisons).

bdach: It is for the removal test to work. The model implements `IEquatable` and I believe that is what's going to be used?

peppy: aha. guess it's ok

peppy (approved):

Issue #11250, opened 2020-12-22
Tracking issue for multiplayer issues


smoogipoo:
# Required before release
- [x] Disconnection should part the room (#11265)
- [x] Retrying should be disabled (also investigate incorrect player state on retry - [gets stuck in playing server-side](https://gist.github.com/peppy/a01dc1d637642fedf4915dfe626818d7)) (https://github.com/ppy/osu/pull/11267)
- [x] Add "playing" state to the user panel (#11252)
- [x] Add leaderboard to multiplayer player (#11253)
- [x] Multiplayer song select isn't applying looping to hold selection (#11254)
- [x] Joining ended games shouldn't be possible (shouldn't send a request in the first place) (#11256)
- [x] Game crashes on exceptions from signalr. Make errors show as notifications. (#11266)
- [x] Changing ruleset after opening a room breaks everything (#11281)

# Nice to have

- [x] Don't send loaded state while notification steals focus (move loading/loaded states to PlayerLoader) (#11279)
- [x] Disable skipping, or make all users skip simultaneously. (#11282)
- [x] Add sounds for state changes and room start (https://github.com/ppy/osu/pull/11283 / https://github.com/ppy/osu/pull/11284)
- [x] Frame bundle should include enough data to show accuracy... accurately (either rolling max score or calculated acc) (https://github.com/ppy/osu/pull/11278)
- [x] Add "Waiting for players" / "Waiting for results" to player/playerloader. (#11279)
- [ ] Don't accept any user inputs in player until gameplay has actually started.
- [x] Server should check whether beatmap is valid in ChangeSettings calls (https://github.com/ppy/osu-server-spectator/pull/16)
- [x] Disable buttons in match subscreen to prevent double actions. (https://github.com/ppy/osu/pull/11353)
- [x] Change "create room" button to "create match", disable while multiplayer client is not connected. (#11274)
- [x] Add states for when the user doesn't have the map / is downloading the map. (https://github.com/ppy/osu/pull/11410)
- [x] If an error occurs at the change settings dialog, all info will be cleared from it rather than reset or left in current state. (https://github.com/ppy/osu/issues/11416)
- [x] If a room is created with a beatmap with mods chosen, they will get discarded after the room is created (need to select the mods twice for it to succeed). (https://github.com/ppy/osu-server-spectator/pull/19)
- [x] Investigate chat being gray in some cases (when using F8 chat) (see existing issue #9508)
[...]

PR #11274, opened 2020-12-23
Split create game buttons for timeshift and realtime multiplayer

bdach:
In order to change the text and add disable-on-client-disconnection logic to the latter (a low-hanging item from #11250).

### DIFF
--- a/osu.Game/Screens/Multi/Multiplayer.cs
+++ b/osu.Game/Screens/Multi/Multiplayer.cs
@@ -22,6 +22,7 @@ using osu.Game.Screens.Multi.Lounge;
 using osu.Game.Screens.Multi.Lounge.Components;
 using osu.Game.Screens.Multi.Match;
 using osu.Game.Users;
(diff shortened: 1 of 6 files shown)


osu.Game/Screens/Multi/Multiplayer.cs
```diff
@@ -129,12 +128,18 @@ protected Multiplayer()
...
                     },
                     new Header(screenStack),
-                    createButton = new CreateRoomButton
+                    createButton = CreateNewMultiplayerGameButton().With(button =>
                     {
-                        Anchor = Anchor.TopRight,
-                        Origin = Anchor.TopRight,
-                        Action = () => OpenNewRoom()
-                    },
+                        button.Anchor = Anchor.TopRight;
```
peppy, line 133: not 100% sure about having this much layout specific code here rather than in a base class, but let's go with it for now.

peppy (approved):

PR #11276, opened 2020-12-24
Fix spectator client sending no data if there is no local user input

peppy:
This half fixes the issue of idle input looking the same as the user being paused to someone spectating a user. It is mainly intended to fix the multiplayer scoreboard, as this fix only covers the cases where `FrameBundleHeader` data is user, not where time / frame information is required.

A further fix will be required to keep player moving forward during such periods, likely by sending the last input frame with an updated timestamp. This requires the `SpectatorStreamingClient` to know about game time which adds further complexity, so I've left it as a follow up task.

Note that this continues to send updates while paused or at a retry screen. These scenarios should also be handled better in the future.

### DIFF
--- a/osu.Game/Online/Spectator/SpectatorStreamingClient.cs
+++ b/osu.Game/Online/Spectator/SpectatorStreamingClient.cs
@@ -29,9 +29,15 @@ namespace osu.Game.Online.Spectator
     public class SpectatorStreamingClient : Component, ISpectatorClient
     {
         /// <summary>
-        /// The maximum milliseconds between frame bundle sends.
+        /// The minimum milliseconds between frame bundle sends.
         /// </summary>
-        public const double TIME_BETWEEN_SENDS = 200;
+        public const double MIN_TIME_BETWEEN_SENDS = 200;
+
+        /// <summary>
+        /// The maximum milliseconds between frame bundle sends.
+        /// A frame bundle will always be sent at least once after this duration elapsees.
+        /// </summary>
+        public const double MAX_TIME_BETWEEN_SENDS = 1000;
 
         private HubConnection connection;
 
@@ -292,7 +298,10 @@ namespace osu.Game.Online.Spectator
         {
             base.Update();
 
-            if (pendingFrames.Count > 0 && Time.Current - lastSendTime > TIME_BETWEEN_SENDS)
+            double timeSinceLastSend = Time.Current - lastSendTime;
+
+            if ((pendingFrames.Count > 0 && timeSinceLastSend > MIN_TIME_BETWEEN_SENDS)
+                || (currentScore != null && timeSinceLastSend > MAX_TIME_BETWEEN_SENDS))
                 purgePendingFrames();
         }
 
--- a/osu.Game/Rulesets/UI/DrawableRuleset.cs
+++ b/osu.Game/Rulesets/UI/DrawableRuleset.cs
@@ -268,6 +268,12 @@ namespace osu.Game.Rulesets.UI
             return false;
         }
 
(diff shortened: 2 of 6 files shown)


osu.Game/Rulesets/UI/RulesetInputManager.cs
```diff
@@ -30,12 +30,21 @@ public ReplayRecorder Recorder
...
+                    recorder.Dispose();
+                }
+                else
+                {
+                    if (recorder != null)
+                        throw new InvalidOperationException("Cannot attach more than one recorder");
+
+                    recorder = value;
+                    KeyBindingContainer.Add(recorder);
+                }
```
bdach, line 47: I'm not a 100% fan of this "set null to stop recording" logic, but having this as a property that can be initialised with the rest of the hierarchy works nice for test code, so maybe fine to have this...

bdach:
Upon testing this, it is somehow possible for the client to continue sending frames after hard-exiting player. Not sure what that is all about, but I'm guessing that wasn't intended. Going to look into that now.

osu.Game/Rulesets/UI/RulesetInputManager.cs
peppy: Yeah, not 100% on it. Although you could think of it less as "to stop recording" and more just "disposing the old instance".. not sure.

peppy:
Are you sure that's not also a thing on master? I wouldn't be surprised if it's not being disposed on exiting player (and never was) under certain conditions...

peppy:
Blocking pending further consideration.

osu.Game/Online/Spectator/SpectatorStreamingClient.cs
```diff
@@ -292,7 +298,10 @@ protected override void Update()
         {
             base.Update();
 
-            if (pendingFrames.Count > 0 && Time.Current - lastSendTime > TIME_BETWEEN_SENDS)
+            double timeSinceLastSend = Time.Current - lastSendTime;
+
+            if ((pendingFrames.Count > 0 && timeSinceLastSend > MIN_TIME_BETWEEN_SENDS)
+                || (currentScore != null && timeSinceLastSend > MAX_TIME_BETWEEN_SENDS))
```
bdach, line 304: On revisiting this one, this check is causing empty frames to be sent outside of gameplay. `currentScore` is not actually cleared in `endPlaying` - and doing that can actually trip an assertion in `purgePendingFrames()`.

The following seems to work okay:

```suggestion
                || (isPlaying && timeSinceLastSend > MAX_TIME_BETWEEN_SENDS))
```

osu.Game/Rulesets/UI/DrawableRuleset.cs
```diff
@@ -268,6 +268,12 @@ public bool RemoveHitObject(TObject hitObject)
             return false;
         }
 
+        public override void EndRecording()
+        {
+            if ((KeyBindingInputManager is IHasRecordingHandler recordingInputManager))
```
bdach, line 273: one set of parens too many

```suggestion
            if (KeyBindingInputManager is IHasRecordingHandler recordingInputManager)
```

bdach (changes requested):

peppy:
@smoogipoo i believe you already addressed the concept looking to be addressed by this PR recently. can you confirm?

smoogipoo:
Yeah, well now it always sends frames at 60fps.

I think the future solution is to make spectator continue playback even though no frames have come through, and pausing on the next hitobject? Or as done here I guess - increase the send interval.

But probably safe to close for now.

PR #11279, opened 2020-12-24
Don't begin gameplay until all users are in a completely prepared state

peppy:
Before this change, players' clients would signal ready when the bulk of the beatmap processing completed, but while the user still had the ability to interrupt the gameplay start process (by, for instance, hovering the settings on the `PlayerLoader` screen). A client may also show a notification during this period indefinitely blocking gameplay start.

To ensure that all players start in sync, I've moved the "is loaded" state change to now be when the gameplay clock is ready to start. This is at a point where the user has no ability to interrupt the start of gameplay.

This also adds a loading overlay which signifies that the game is waiting on other players. Note that this does not have any explanatory text, but I intend to show status updates on the leaderboard as to which players we are waiting on, which should give enough visual cues for the time being.

### DIFF
--- a/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
+++ b/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimePlayer.cs
@@ -3,12 +3,12 @@
 
 using System;
 using System.Diagnostics;
-using System.Threading;
 using System.Threading.Tasks;
 using JetBrains.Annotations;
 using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Logging;
+using osu.Game.Graphics.UserInterface;
 using osu.Game.Online.Multiplayer;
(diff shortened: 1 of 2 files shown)


peppy:
Note that I experimented with implementing this at a derived `PlayerLoader` level but it was convoluted to implement and likely not able to be as reliable.

I also removed the arbitrary 30s abort because users can already choose to abort via the exit button. Can add it back if it's seen as something we want. (likewise, I increased the results wait to 60s for the same reasons).

bdach:
I've fixed up an outdated comment. Seems OK otherwise.

Note that while the player is waiting for permission to start gameplay it is still accepting keyboard inputs (best tested with mania), but I don't think that's worth blocking this PR for. That can always come in a separate one and there are bigger fish to fry (such as the track looping issue we found today, which is easily reproducible with this change by commenting out the body of `onResultsReady()`).

bdach (approved):

PR #11286, opened 2020-12-24
Rename "timeshift" to "playlists"

peppy:
This only covers the user-facing instances. Code and class name changes will happen once things have calmed down. Note that there's still quite a few references (especially in the `Multiplayer` class) to "room" where it should really say "playlist". These will take a bit more work an abstraction.

This is just important to set the branding apart of "realtime" multiplayer and "timeshift" multiplayer, so we stop referring to multiplayer with the silly "realtime" prefix. Multiplayer *is* real-time.

### DIFF
--- a/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMultiplayer.cs
+++ b/osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMultiplayer.cs
@@ -62,6 +62,8 @@ namespace osu.Game.Screens.Multi.RealtimeMultiplayer
             return room;
         }
 
+        protected override string ScreenTitle => "Multiplayer";
+
         protected override RoomManager CreateRoomManager() => new RealtimeRoomManager();
 
         protected override LoungeSubScreen CreateLounge() => new RealtimeLoungeSubScreen();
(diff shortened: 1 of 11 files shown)


osu.Game/Screens/Multi/RealtimeMultiplayer/RealtimeMultiplayer.cs
```diff
@@ -57,7 +57,7 @@ protected override void UpdatePollingRate(bool isIdle)
 
         protected override Room CreateNewRoom()
         {
-            var room = base.CreateNewRoom();
+            var room = new Room { Name = { Value = $"{API.LocalUser}'s awesome room" } };
```
bdach, line 60: Looks semi-awkward on the match creation screen with "match" everywhere but the textbox:

2020-12-24-155901_892x252_scrot

Thoughts on

```suggestion
            var room = new Room { Name = { Value = $"{API.LocalUser}'s awesome match" } };
```

?

peppy: sure. we're going to really need to standardise terms at some point.

bdach: As discussed in discord, "realtime" has been reverted back to "room", while playlists have been changed to playlists.

bdach (dismissed):

PR #11289, opened 2020-12-24
Add "beatmap not downloaded / downloading" user states to enum

frenzibyte:
Dependency for server-side support.

### DIFF
--- a/osu.Game/Online/RealtimeMultiplayer/MultiplayerUserState.cs
+++ b/osu.Game/Online/RealtimeMultiplayer/MultiplayerUserState.cs
@@ -6,16 +6,6 @@ namespace osu.Game.Online.RealtimeMultiplayer
     public enum MultiplayerUserState
     {
         /// <summary>
-        /// The user has not downloaded the beatmap yet.
-        /// </summary>
-        BeatmapNotDownloaded,
-
-        /// <summary>
-        /// The user is downloading the beatmap.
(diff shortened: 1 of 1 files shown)


osu.Game/Online/RealtimeMultiplayer/MultiplayerUserState.cs
```diff
@@ -5,6 +5,16 @@ namespace osu.Game.Online.RealtimeMultiplayer
 {
     public enum MultiplayerUserState
     {
+        /// <summary>
+        /// The user has not downloaded the beatmap yet.
+        /// </summary>
+        BeatmapNotDownloaded,
```
peppy, line 11: have you looked how this is serialized, and whether inserting at the *start* of the enum will break everything?

frenzibyte: oh..., my bad.

peppy:
Have you confirmed either way? I was asking because I was unsure - it may actually be okay depending on how the serialization is done.

frenzibyte:
Leaving it the way I had has broken up everything due to different enum values between this version and the server's version (server's `Idle` becomes `BeatmapNotDownloaded` here, etc...), I think the states can be moved back to top when the server-side changes are merged.

frenzibyte:
Closing as I'm going in another direction.

PR #11290, opened 2020-12-24
Fix multiplayer gameplay potentially looping audio after reaching end

bdach:
Fixes the track potentially looping in player, therefore preventing users from progressing to results. For 100% reproduction of that issue, comment out the body of `onResultsReady()` in `RealtimePlayer`.

Must have been an issue all along, but probably masked by the fact that player exited too fast to begin looping.

Not a 100% sure this is proper but it seems to behave properly, pass multiplayer tests and makes sense to me code-wise.


peppy:
Not 100% on this one but can't find any regressions.

peppy (approved):

PR #11300, opened 2020-12-25
Rename multiplayer/timeshift namespaces to match their new terms

### LINKED ISSUE #11291: Clean up in-source naming of multiplayer variants
Follow-up/continuation of https://github.com/ppy/osu/pull/11286, but for the code references.

peppy:
This may require a couple of passes to catch weird edge cases.

simple ones:
`Timeshift` -> `Playlists`
`realtime` -> ``
`RealtimeMultiplayer` -> `Multiplayer`

a bit more annoying:
`Multiplayer` (the base `abstract` screen) -> `MultiplayerScreen` - should this maybe be `RoomsScreen`?
`MultiplayerTestScene` (the base `abstract` test) -> `RoomTestScene` to avoid conflicts with a second `MultiplayerTestScene`

 I'm also not sure about when we should be including prefixes (when the prefix is already in the namespace):

image

Closes https://github.com/ppy/osu/issues/11291


osu.Game/Online/Rooms/MultiplayerScore.cs
```diff
@@ -13,7 +13,7 @@
 using osu.Game.Scoring;
 using osu.Game.Users;
 
-namespace osu.Game.Online.Multiplayer
+namespace osu.Game.Online.Rooms
```
bdach, line 16: Would propose renaming the class to `RoomScore` too, along with its derivatives (`IndexedMultiplayerScores`, `MultiplayerScore`, `MultiplayerScores`, `MultiplayerScoresAround`) - or maybe moving to the `Multiplayer` namespace? Not sure but something doesn't fit too well here.

Not to mention that we have an `IndexPlaylistScoresRequest` that returns `IndexedMultiplayerScores`... maybe that should be left as a follow up, then? idk

osu.Game/Online/Rooms/RoomCategory.cs
```diff
@@ -1,12 +1,12 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-namespace osu.Game.Online.Multiplayer
+namespace osu.Game.Online.Rooms
 {
     public enum RoomCategory
     {
         Normal,
```
bdach, line 8: Not sure what the `Normal` category is supposed to be now - maybe it should be `Playlist`, `SpotlightPlaylist`, `Multiplayer`?

osu.Game/Screens/Multi/Multiplayer/Multiplayer.cs
```diff
@@ -28,46 +28,46 @@ public override void OnResuming(IScreen last)
 
         protected override void UpdatePollingRate(bool isIdle)
         {
-            var timeshiftManager = (RealtimeRoomManager)RoomManager;
+            var playlistsManager = (MultiplayerRoomManager)RoomManager;
```
bdach, line 31: Variable name doesn't match the type. (It was borked before, but now that we're fixing...)

osu.Game/Screens/Multi/Play/PlaylistsPlayer.cs
```diff
@@ -20,7 +20,7 @@
 
 namespace osu.Game.Screens.Multi.Play
```
bdach, line 21: Should probably make its way over to
```suggestion
namespace osu.Game.Screens.Multi.Playlists
```

osu.Game/Screens/Multi/Playlists/PlaylistsMultiplayer.cs
```diff
@@ -4,46 +4,46 @@
...
+using osu.Game.Online.Rooms;
 using osu.Game.Screens.Multi.Components;
 using osu.Game.Screens.Multi.Lounge;
 using osu.Game.Screens.Multi.Match;
 
-namespace osu.Game.Screens.Multi.Timeshift
+namespace osu.Game.Screens.Multi.Playlists
 {
-    public class TimeshiftMultiplayer : Multiplayer
+    public class PlaylistsMultiplayer : MultiplayerScreen
```
bdach, line 14: I'd probably drop the `Multiplayer` here and just name it `Playlists`.

osu.Game/Screens/Multi/Ranking/PlaylistsResultsScreen.cs
```diff
@@ -11,13 +11,13 @@
 using osu.Framework.Graphics.Containers;
 using osu.Game.Graphics.UserInterface;
 using osu.Game.Online.API;
-using osu.Game.Online.Multiplayer;
+using osu.Game.Online.Rooms;
 using osu.Game.Scoring;
 using osu.Game.Screens.Ranking;
 
 namespace osu.Game.Screens.Multi.Ranking
```
bdach, line 18: Similarly
```suggestion
namespace osu.Game.Screens.Multi.Playlists
```

bdach:
Seems like a good start. Follow-up items I'd suggest:

* Clean up the room vernacular for multiplayer/playlists ("match/room/playlist" and all that)
* Move entire `osu.Game.Screens.Multi` namespace to `osu.Game.Screens.OnlinePlay`, and possibly its most direct descendant classes to match (i.e. `MultiplayerScreen` -> `OnlinePlayScreen`, etc.)

bdach:
As for the test failures:

* `TestSceneMatchLeaderboard` should inherit `RoomTestScene`, but that test scene is actually entirely broken (uses a local request and data structure instead of `APILeaderboard`, and silences errors, while the underlying API contract changed - and it uses a hardcoded ID anyway, so it really needs to be removed or rewritten entirely).
* Same inheritance error applies to `TestSceneMultiplayerRoomManager`.

I'll push the changes in a sec.

osu.Game/Online/Rooms/RoomCategory.cs
peppy: actually you make a good point - we can't change this without extra handling for osu-web api type mapping. will probably need to revert this one back.

bdach:
Let's get this in as a first step. The rest can come in later.

bdach (approved):

PR #11316, opened 2020-12-26
Fix join room failures on large lobbies

### LINKED ISSUE #11294: Failed to join multiplayer room error
**Describe the crash:** 
i just tried to join a multiplayer room and then this error appeared. im not able to reproduce this again.

**osu!lazer version:** 
2020.1225.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/5740314/database.log)
[network.log](https://github.com/ppy/osu/files/5740315/network.log)
[performance.log](https://github.com/ppy/osu/files/5740316/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5740317/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5740318/updater.log)

peppy:
Closes https://github.com/ppy/osu/issues/11294

I looked at potentially changing this to use the locking flow I'm using server side but due to the number of usages (and the majority already correctly scheduled) I went in this direction instead. We can reassess at a future point in time if locking is deemed what we want.

This also fixes user population running single file and bypassing the `UserLookupCache` batching.

### DIFF
--- a/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
+++ b/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
@@ -108,8 +108,7 @@ namespace osu.Game.Online.Multiplayer
 
             Debug.Assert(Room != null);
 
-            foreach (var user in Room.Users)
-                await PopulateUser(user);
+            await Task.WhenAll(Room.Users.Select(PopulateUser));
 
             updateLocalRoomSettings(Room.Settings);
         }
--- a/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
+++ b/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
@@ -126,7 +126,7 @@ namespace osu.Game.Online.Multiplayer
 
         public virtual Task LeaveRoom()
         {
-            Schedule(() =>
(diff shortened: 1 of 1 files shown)


bdach:
Looking at the failing tests here - for what it's worth, the first one I'm looking at, I actually have no idea how it ever worked.

bdach:
Never mind, most likely all the test failures are due to that `getRoomUsers()` logic. The schedule doesn't manage to run in time before the reset event times out, joining the room fails because a null user list is return, at which point the client leaves the room and nulls it out, leaving behind a quite broken state that tests definitely cannot run in.

peppy:
Shall we just increase the timeout and call it a day? It was arbitrary anyway.

bdach:
I kind of want to see why the schedule runs so late first.

bdach:
Pushed out changes that should hopefully allow the tests to pass (at least they did locally, and I checked a few times). Make sure you're OK with them, but there should be no surprises.

bdach:
Well, changes certainly seem to have helped.

Not a 100% on this general method, locking would be preferable as mentioned, but let's take it step by step for now...

bdach (approved):

peppy:
Changes look fine.

PR #11318, opened 2020-12-26
Fix crash on local user not being present when updating the ready button display

### LINKED ISSUE #11292: Multiplayer crash after losing connection inside the lobby and rejoining
**Describe the crash:** 
i completed a map in multiplayer, returned to the lobby and waited for a new map and after waiting for a few seconds i suddenly lost the connection. i tried to reconnect but it didnt work so i tried it again until i could join the lobby again and then the game crashed. 

**osu!lazer version:** 
2020.1225.0

**Logs:**
[database.log](https://github.com/ppy/osu/files/5740196/database.log)
[network.log](https://github.com/ppy/osu/files/5740197/network.log)
[performance.log](https://github.com/ppy/osu/files/5740198/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5740199/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5740200/updater.log)

peppy:
Also renames a confusing event (wasn't sure if it was fired when the room instance changed.. hopefully renaming from Changed to Updated helps with understanding this).

- [x] Depends on #11316
- Closes #11292.

### DIFF
--- a/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
+++ b/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
@@ -33,7 +33,7 @@ namespace osu.Game.Online.Multiplayer
         /// <summary>
         /// Invoked when any change occurs to the multiplayer room.
         /// </summary>
-        public event Action? RoomChanged;
+        public event Action? RoomUpdated;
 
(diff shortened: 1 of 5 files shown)


bdach (approved):
looks obvious

PR #11319, opened 2020-12-26
Fix copying a multiplayer room attempting to create a playlist room instead

### LINKED ISSUE #11310: Server asks for the "ends_at" value when attempting to copy a Multi room
**Describe the bug:** Right-clicking on a room to create a copy, the server returns an error asking for the "ends_at" value (which I believe is from Timeshift). ~My guess is that the server is trying to create a Timeshift lobby somehow.~

steps
error

**Logs:**
from [network.log](https://github.com/ppy/osu/files/5742205/network.log):
```
2020-12-25 16:51:36 [verbose]: Performing request osu.Game.Online.Multiplayer.CreateRoomRequest
2020-12-25 16:51:36 [verbose]: Request to https://osu.ppy.sh/api/v2/rooms failed with System.Net.WebException: UnprocessableEntity.
2020-12-25 16:51:36 [verbose]: Response was: {"error":"'ends_at' is required"}
2020-12-25 16:51:36 [verbose]: Failing request osu.Game.Online.Multiplayer.CreateRoomRequest (osu.Game.Online.API.APIException: 'ends_at' is required
2020-12-25 16:51:36 [verbose]: ---> System.Net.WebException: UnprocessableEntity
2020-12-25 16:51:36 [verbose]: --- End of inner exception stack trace ---)
```

sh30801:
Fixes #11310 by properly copying the `Category` value, as described in the issue.

### DIFF
--- a/osu.Game/Online/Rooms/Room.cs
+++ b/osu.Game/Online/Rooms/Room.cs
@@ -130,6 +130,7 @@ namespace osu.Game.Online.Rooms
         {
             RoomID.Value = other.RoomID.Value;
             Name.Value = other.Name.Value;
+            Category.Value = other.Category.Value;
 
             if (other.Host.Value != null && Host.Value?.Id != other.Host.Value.Id)
                 Host.Value = other.Host.Value;
--- a/osu.Game/Online/Rooms/Room.cs
+++ b/osu.Game/Online/Rooms/Room.cs
@@ -130,7 +130,9 @@ namespace osu.Game.Online.Rooms
         {
             RoomID.Value = other.RoomID.Value;
             Name.Value = other.Name.Value;
-            Category.Value = other.Category.Value;
+
+            if (other.Category.Value != RoomCategory.Spotlight)
+                Category.Value = other.Category.Value;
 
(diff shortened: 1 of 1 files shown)


peppy:
Not sure how useful copying a multiplayer room is, but let's fix this for now.

peppy (dismissed):

peppy:
Actually, I think this may have been left out to avoid room creation with spotlight category. We may want to add a safe-guard against this.

bdach:
Yeah that was what I was slightly worried about which is not why I hadn't done this myself; `CopyFrom` is called both from the UI operation and a few internal places. Might be safer to do this somewhere closer to the origin of the former.

bdach (approved):
looks to be working correctly

bdach:
@Neuheit For future reference we don't really use that (js-like?) commit message format you did. Just a normal message is fine. The diff speaks by itself as to where the change was, and somewhat what the fix was; elaborating can be always done after the commit subject, after a blank line, as is the standard.

Neuheit:
Understood, thank you!

PR #11320, opened 2020-12-26
Fix leaderboard showing score updates for quit users

### LINKED ISSUE #11308: Players are not removed from the leaderboard when leaving the match
**Describe the bug:**

When a player leaves mid match (or during loading), he is not removed from the ingame leaderboard. In addition, if they start playing in singleplayer, score information keeps being updated in the match.

Video shows player "fmpgdev" leaving the lobby right after load and starting a beatmap in singleplayer twice.

https://user-images.githubusercontent.com/22643729/103137835-f0134700-46cc-11eb-85af-eea6ffee3940.mp4

**osu!lazer version:** master 4e21bd0

**Logs:**

fmpgdev2(staying):
[database.log](https://github.com/ppy/osu/files/5742134/database.log)
[network.log](https://github.com/ppy/osu/files/5742135/network.log)
[performance.log](https://github.com/ppy/osu/files/5742136/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5742137/runtime.log)
[updater.log](https://github.com/ppy/osu/files/5742138/updater.log)

fmpgdev(leaving):
[database.log](https://github.com/ppy/osu/files/5742140/database.log)
[network.log](https://github.com/ppy/osu/files/5742141/network.log)
[performance.log](https://github.com/ppy/osu/files/5742142/performance.log)
[runtime.log](https://github.com/ppy/osu/files/5742143/runtime.log)

peppy:
The spectator data feed was not unbound if a user quit the room, meaning if they continued to join another room (or play solo) their now incorrect score was still showing up on the leaderboard.

This change unsubscribes and also shows quit users visually (via a tint of their panel to gray).

Closes https://github.com/ppy/osu/issues/11308.

### DIFF
--- a/osu.Game/Screens/Play/HUD/GameplayLeaderboardScore.cs
+++ b/osu.Game/Screens/Play/HUD/GameplayLeaderboardScore.cs
@@ -52,7 +52,7 @@ namespace osu.Game.Screens.Play.HUD
                     positionText.Text = $"#{scorePosition.Value.FormatRank()}";
 
                 positionText.FadeTo(scorePosition.HasValue ? 1 : 0);
-                updateColour();
+                updateState();
             }
         }
 
@@ -231,20 +231,20 @@ namespace osu.Game.Screens.Play.HUD
             TotalScore.BindValueChanged(v => scoreText.Text = v.NewValue.ToString("N0"), true);
             Accuracy.BindValueChanged(v => accuracyText.Text = v.NewValue.FormatAccuracy(), true);
             Combo.BindValueChanged(v => comboText.Text = $"{v.NewValue}x", true);
-            HasQuit.BindValueChanged(v => updateColour());
+            HasQuit.BindValueChanged(v => updateState());
         }
 
         protected override void LoadComplete()
         {
             base.LoadComplete();
 
-            updateColour();
+            updateState();
             FinishTransforms(true);
         }
 
         private const double panel_transition_duration = 500;
 
-        private void updateColour()
+        private void updateState()
         {
             if (HasQuit.Value)
             {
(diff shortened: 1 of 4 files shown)


osu.Game/Screens/Play/HUD/GameplayLeaderboardScore.cs
```diff
@@ -244,6 +254,9 @@ protected override void LoadComplete()
 
         private void updateColour()
         {
+            if (HasQuit.Value)
+                return;
```
bdach, line 258: Doing it this way has a quirk - if the user that quit was first on the leaderboard, their panel will still have a wider size, even as the user drops down in rankings naturally. I'd probably jam the "has quit" case in here (and rename the method, as it clearly does more than just update colour).

osu.Game/Screens/Play/HUD/MultiplayerGameplayLeaderboard.cs
```diff
@@ -33,43 +48,68 @@ public MultiplayerGameplayLeaderboard(ScoreProcessor scoreProcessor, int[] userI
...
 
             scoringMode = config.GetBindable<ScoringMode>(OsuSetting.ScoreDisplayMode);
             scoringMode.BindValueChanged(updateAllScores, true);
         }
 
+        protected override void LoadComplete()
+        {
+            base.LoadComplete();
+
+            // BindableList handles binding in a really bad way (Clear then AddRange) so we need to do this manually..
```
bdach, line 85: side note - `BindableList.Clear()` should maybe be raising a `NotifyCollectionChangedAction.Reset` on clear, although that's debatable. sort of unwieldy indeed.

bdach (changes requested):

peppy:
The asssert failing can be fixed by changing the `.Expire` i added to a forced `Remove` then `Dispose` (as below) but this is pretty ugly.

image


I don't think there's actually much point in having this assert (and it could also fail of the scoring mode was changed, see `updateAllScores`) so i've opted to just remove it.

bdach (approved):

PR #11321, opened 2020-12-26
Fix catcher position update delayed one frame in osu!catch replay

ekrctb:
Because catcher position is checked for each hit object, and `HitObjectContainer` came before `CatcherArea` in the children ordering, the replay frame processed in `CatchInputManager` was not applied until one frame later.
That was one reason why the catch autoplay misses hit objects (especially when fast-forwarded). I determined more causes of the autoplay issue, but those are not included during this PR.

<del>
Because all other replay state changes are notified from `InputManager` through `InputQueue`, like mouse position in standard, I think the logic of catcher position change should be written in the input manager. I think the way `Catcher` class is directly referenced by the input manager (rather than interfaces used in `KeyBindingContainer`) is against the principle but I don't think abstracting it improves the code.
</del>


osu.Game.Rulesets.Catch/CatchInputManager.cs
```diff
@@ -13,6 +18,26 @@ public CatchInputManager(RulesetInfo ruleset)
...
+
+                if (lastState.CatcherX != null)
+                {
+                    foreach (Drawable drawable in NonPositionalInputQueue)
+                    {
+                        if (drawable is Catcher catcher)
+                            catcher.UpdatePosition(lastState.CatcherX.Value);
+                    }
+                }
+            }
```
bdach, line 37: Yeah I don't know about this. This is all very ad-hoc. I can't see osu! (the ruleset) doing anything similar anywhere, and this looks pretty bad all things considered.

I feel this should be solved in a more reusable manner for the sake of the overall ruleset API instead of scraping all over the current structure. This'll probably break immediately if we have to change anything.

I'd be in favour of more abstraction here, definitely. Have you tried anything in that direction? If so, do you have any branches to show?

bdach (changes requested):

osu.Game.Rulesets.Catch/CatchInputManager.cs
ekrctb: That I said standard ruleset doing a similar thing was my confusion involved, but the example is `DrawableOsuHitObject` uses `IsHovered` (set by `InputManager` after mouse move input) (it is similar in the sense that it is a data flow from `InputManager` to descendants other than via `KeyBindingContainer`).

As for abstraction, I don't have an idea of any good abstraction currently. I also find the whole `RulesetInputManagerInputState`, `ReplayState`, etc. quite confusing, and probably reconsidering it is desired too.

I can start thinking about how to improve the API. However, **the issue solved here is quite a blatant correctness issue**, and I think solving it should be prioritized over such an API design thinking.

bdach: I'm gonna push back on that last part a little - if this is a problem for catch, it may well be a problem for another custom ruleset, and as such I'm of the opinion that we should be looking to get a "nice" solution in as not to set a bad precedent.

Regardless I'll have a more detailed look at if this can be improved in any way. Not a fan of delaying code quality matters for "later", because "later" more often than not becomes "never".

ekrctb:
Just to note alternative solution: `Catcher` use `inputManager.CatcherX ?? X` in the hit detection logic.
It works (I tried before this PR solution), and it is more similar to the current logic used in `CatcherArea`.
However, I don't like this way because the direction of the data flow (descendant request property of `InputManager`) is opposite of the other replay actions (`InputManager` triggers descendant event handler) and I feel the later way should be used consistently.

bdach:
On the other side, as you said, osu! already uses `IsHovered`, already violating that principle, right? I'm not sure I see it as such big of a deal myself. Is there another reason other than consistency to not do that?

ekrctb:
I don't like `IsHovered` either, but `IsHovered` is different because `IsHovered` is a property of `Drawable` and not `InputManager` (so a `Drawable` can get its `IsHovered` without knowledge of `InputManager`), and it is more like event handlers than `InputManager.CurrentState`.

> Is there another reason other than consistency to not do that?
Stateful operations (changing catcher direction based on position difference) require maintaining the previous position and require a bit more code.
(I don't like `Catcher` uses `Drawable.X` (publically settable) for its logic but I think it is orthogonal to the issue).

peppy:
One thing I've noticed as that with this change, the catcher still has a change to update the catcher's position (used for local input) which would now happen *after* the replay input does its update. This could potentially cause minor glitches in movement I think?

https://github.com/ppy/osu/blob/96f3e0dad8243d221592d99ed39949488d829e6f/osu.Game.Rulesets.Catch/UI/Catcher.cs#L424

I'm also a bit hesitant to approve this solution. A good starting point is to accept that we have both "local" input and "overriding" replay frames - quite unlike any other mode. Whether this will be a permanent thing or not is unsure, but it allows for floating-point accurate movement that would potentially be hard to achieve with just down/up events (the chance of introducing compounding errors gets higher).

I think at some point in the hierarchy we need a property to provide the ground truth for the catcher's X position. Maybe `DrawableCatchRuleset` would be a good place (we do slightly different things in other rulesets as well). The implementation would be something like

`public float CatcherXPosition => inputManager.ReplayOverride.X ?? playfield.Catcher.X`

This property would be used for updating the position, moving the logic back to where it was previously. The "local" movement would still be updated but would be ignored as long as replay state was available.

@ekrctb how does this sound? i think it's first in the alternate direction you proposed.

ekrctb:
I revised the PR to adopt the suggested approach. Now, `Catcher` is pulling the override positions from `InputManager`, not the other way around.

I introduced `LogicalX` in `Catcher` separated from `X` to stress it is used for fruit catching logic, not the visual draw position.
The property is currently not strictly necessary as it is mirroring `X`, but I wonder if it makes sense to have a separate model object of `Catcher`, by making it `DrawableCatcher`, like `HitObject` vs `DrawableHitObject`.
The intent is `LogicalX` will be a property of `Catcher` (the model), and `X` will be a property of `DrawableCatcher` if such distinction can be made.

peppy:
Is there a reason for requiring the `LogicalX` value? Just thinking that we could potentially replace lines like

`            var catchPosition = getReplayOverride() ?? LogicalX;`

with 

`            updatePosition()`

where

```csharp

private void updatePosition()
{
    if (getReplayOverride() is double d)
        X = d;
}
```

and then use the single `X` representation, unless I'm missing something.

(just proof of concept code, would require some further tidying)

ekrctb:
`LogicalX` is not necessary but I prefer making it explicit. See my comment above.

peppy:
Right. i think that's fine but my remaining issue is without understanding the private implementations of the code it's not clear as to *why* this discrimination is required.

Would you be against something along the lines of 

```csharp

        /// <summary>
        /// The immediate horizontal position of the catcher. May return an overridden position based on replay data.
        /// </summary>
        public float LogicalX
        {
            get
            {
                updatePosition(getReplayOverride() ?? LogicalX);

                return logicalX;
            }
            set => updatePosition(value);
        }
```

ie. ensuring that `LogicalX` is always correct, rather than checking on each usage for the override?

ekrctb:
I guess that is fair.

ekrctb:
Superseded by #12349.

PR #11324, opened 2020-12-26
Fix "difficulty adjust" mod overridden settings getting discarded in multiplayer

### LINKED ISSUE #11305: Difficulty adjust not working in multiplayer

**Logs:**

frenzibyte:
Closes #11305 

## What the issue is

The issue was occurring because of the mod getting reinstantiated twice during the room creation process:
https://github.com/ppy/osu/blob/6849eabf2ce56db434d9a68829f63207659b78a1/osu.Game/Screens/OnlinePlay/Multiplayer/MultiplayerMatchSongSelect.cs#L51 https://github.com/ppy/osu/blob/6849eabf2ce56db434d9a68829f63207659b78a1/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs#L443

when it gets reinstantiated, the settings are copied from the previous instance to the new one, and that is happening before `ReadFromDifficulty` is called, which means `userTrackedSettings` wouldn't be tracking the copying, therefore when `ReadFromDifficulty` gets called (by `OsuGame.updateModDefaults()`) the settings would be discarded and the base values would be used instead.

## The solution

I've tried a couple of solutions but none of them worked properly, involving:
 - copying `userTrackedSettings` over, which wouldn't work for `APIMod.ToMod()`.
 - applying difficulty before copying settings, which doesn't really sound sane to do + you would have to track exactly any instantiation of the mod and read difficulty there.
 - forcing difficulty in the `Mod` constructor, which I'm not really sure about, and I don't think will be possible with the current flow.

but I came up with a solution that I am satisfied with:

And what it does is extracting the logic into a `DifficultyTrackingBindable` class, and make it so that it can handle "changing difficulty settings without a base difficulty" cases properly, by tracking any touches to the bindable, until a base difficulty is received.

I've added test coverage for the new mod's settings overriding behaviour, which matches [as it was previously](https://github.com/ppy/osu/blob/3f85fcf0c94da5da35fb3bcad5ef8a4e57493362/osu.Game.Tests/NonVisual/ModDifficultyAdjustOverridingTest.cs#L23-L85) + handles [changing settings before base](https://github.com/ppy/osu/blob/3f85fcf0c94da5da35fb3bcad5ef8a4e57493362/osu.Game.Tests/NonVisual/ModDifficultyAdjustOverridingTest.cs#L87-L118) properly.

### DIFF
--- a/osu.Game.Tests/NonVisual/ModDifficultyAdjustOverridingTest.cs
+++ b/osu.Game.Tests/NonVisual/ModDifficultyAdjustOverridingTest.cs
@@ -1,0 +1,129 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using NUnit.Framework;
+using osu.Game.Beatmaps;
+using osu.Game.Rulesets.Osu.Mods;
+
+namespace osu.Game.Tests.NonVisual
+{
+    [TestFixture]
+    public class ModDifficultyAdjustOverridingTest
+    {
+        private OsuModDifficultyAdjust mod;
+        private int difficultyID;
+
+        [SetUp]
+        public void SetUp()
+        {
+            mod = new OsuModDifficultyAdjust();
+            difficultyID = 0;
+        }
+
+        [Test]
+        public void TestOverridenStopTrackingBase()
+        {
+            var baseDifficulty = createDifficulty(5);
+            mod.ReadFromDifficulty(baseDifficulty);
+
+            mod.ApproachRate.Value = 10;
+            mod.CircleSize.Value = 10;
+
+            Assert.AreEqual(10, mod.ApproachRate.Value);
+            Assert.AreEqual(10, mod.CircleSize.Value);
+            Assert.AreEqual(baseDifficulty.DrainRate, mod.DrainRate.Value);
+            Assert.AreEqual(baseDifficulty.OverallDifficulty, mod.OverallDifficulty.Value);
+
+            baseDifficulty = createDifficulty(3);
+            mod.ReadFromDifficulty(baseDifficulty);
+
+            Assert.AreEqual(10, mod.ApproachRate.Value);
+            Assert.AreEqual(10, mod.CircleSize.Value);
+            Assert.AreEqual(baseDifficulty.DrainRate, mod.DrainRate.Value);
+            Assert.AreEqual(baseDifficulty.OverallDifficulty, mod.OverallDifficulty.Value);
+        }
+
+        [Test]
+        public void TestOverridenAfterBaseTrackOnRevert()
+        {
+            var baseDifficulty = createDifficulty(5);
+            mod.ReadFromDifficulty(baseDifficulty);
+
+            mod.ApproachRate.Value = 10;
+            Assert.AreEqual(10, mod.ApproachRate.Value);
+            Assert.AreEqual(baseDifficulty.ApproachRate, mod.ApproachRate.Default);
+
+            mod.ApproachRate.SetDefault();
+            Assert.AreEqual(baseDifficulty.ApproachRate, mod.ApproachRate.Value);
+
+            baseDifficulty = createDifficulty(3);
+            mod.ReadFromDifficulty(baseDifficulty);
+
+            Assert.AreEqual(baseDifficulty.ApproachRate, mod.ApproachRate.Value);
+        }
+
+        [Test]
+        public void TestOverridenAfterBaseDontTrackWithoutChange()
+        {
+            mod.ReadFromDifficulty(createDifficulty(6));
+
+            mod.ApproachRate.Value = 3;
+            mod.CircleSize.Value = 3;
+
+            Assert.AreEqual(3, mod.ApproachRate.Value);
+            Assert.AreEqual(3, mod.CircleSize.Value);
+
+            mod.ReadFromDifficulty(createDifficulty(3));
+
+            Assert.AreEqual(3, mod.ApproachRate.Value);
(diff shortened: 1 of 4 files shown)


osu.Game.Tests/NonVisual/ModDifficultyAdjustOverridingTest.cs
```diff
@@ -0,0 +1,129 @@
...
+            mod.ReadFromDifficulty(createDifficulty(6));
+
+            Assert.AreEqual(6, mod.ApproachRate.Value);
+            Assert.AreEqual(6, mod.CircleSize.Value);
+
+            mod.ReadFromDifficulty(createDifficulty(3));
+
+            Assert.AreEqual(3, mod.ApproachRate.Value);
+            Assert.AreEqual(3, mod.CircleSize.Value);
+        }
```
bdach, line 118: I don't understand why this should work the way it is written. As far as I'm concerned, the way difficulty adjust works should be dead simple:

* If someone has set a value manually, keep that setting forever, until
* a revert to the default value was requested, in which case the base difficulty should be used.

I'm not going to go into the actual implementation of this as it is way too convoluted and hard to understand for what I think it actually needs to be doing. Probably because these tests are enforcing a very convoluted behaviour. This is all made worse by the fact that all of this data is kept in-band instead of using nullable types.

frenzibyte: This is just handling a case where if a difficulty adjust mod is created with certain values, then `ReadFromDifficulty` is called first time with a difficulty that matches those certain values of the mod, then consider the setting is not overriding anything and should keep tracking base difficulty value. (otherwise, I believe this will make the mod's settings always become "overriden" when jumping to song select with a copied mod or something of the sorts)

Maybe documenting each test and what behaviour is it testing against in a song select/multiplayer context would help in that?

bdach: Relying on order of execution is dodgy any way you slice it. This is a representation problem. If the value hasn't been manually touched by anyone, it should not be deserialised, copied, etc. It should remain the default.

frenzibyte: > If the value hasn't been manually touched by anyone, it should not be deserialised, copied, etc.

I'm not sure I understand what you mean by this, what should be preventing "deserializing, copying", because that doesn't sound like it's possible, can you elaborate more?

bdach: It should be passed around unambiguously as null or similar. None of this "maybe this is the map difficulty, or maybe it's what a user set" stuff.

bdach:
For what it's worth I think the proper way to go forward here is some sort of custom `IBindableNumber` implementation just for this, that essentially does something like the following:

```csharp
public OptionalBindableNumber<T> : IBindableNumber<T>
{
	private bool valueWasChanged;

	public OptionalBindableNumber(T initialValue = default)
	{
		ValueChanged += _ => valueWasChanged = true;
	}

	public T Value
	{
		get => valueWasChanged ? base.Value : base.Default;
		set => base.Value = value;
	}

	public bool IsDefault => !valueWasChanged;

	public void SetDefault => valueWasChanged = false;

	// and probably more things as required
}
```

but this will 100% require larger amendments both game- and framework-side (because `SetDefault` is not virtual, and game is hard-coupled to `BindableNumber` to create settings controls, etc.).

frenzibyte:
Hmm, that might work and would definitely make logic straightforward, good thinking on this! (always have this bad habit of being driven by the current logic of the game and adding to it, instead of rethinking the whole thing in general)

bdach:
An `AggregateBindable` usage (one bindable for map value, second for overriden value, aggregate bindable tracks which one to read from) might also be worth considering, too. Might be even simpler actually.

peppy:
@frenzibyte are you planning to make refactors to this? it would be nice to get a fix in the upcoming hotfix build, but not sure the state on your changes here.

frenzibyte:
Yeah, I'll go through it now.

frenzibyte:
Unfortunately, none of these proposals are possible to implement:
 - `AggregateBindable`: That's not possible given that it exposes an immutable `Result` bindable, and doesn't seem to allow being mutated, if possible to do at all.
 - `OptionalBindableNumber`: With the current bindable flow, members like `IsDefault` and `SetDefault` would not be called from bound bindables, making this impossible to implement properly (no way to revert override back to base).
 
 Seems my current solution is the only way here, unless you have another idea that could work out, I'm free for trying.

bdach:
I still categorically disagree with the current code as written, and doubt it is the "only way".

frenzibyte:
Yeah, that's why I said "unless you have another idea", I've reached a dead end here and I'm very unsure about any other solution.

bdach:
I stared at this problem a bit today, and it was admittedly harder than I expected - but I came up with the following commit: https://github.com/bdach/osu/commit/d1006654e34625ad56b00520b8896183689ac59a.

Mind you, it has rough edges that I'm not entirely sure how to sand, but I think the above is still infinitely more workable than whatever the `DifficultyTrackingBindable` from this PR is trying to be doing. Obviously this does not entirely solve the problem (probably going to need new controls to use that, and adjust serialisation logic), but I think it might clear the major hurdle in all of this.

frenzibyte:
The name might need to be more generic though, something like `OverridableSetting`? and an `Overridable{Slider/Control}` for UI control, unless you think the name you had is fine?

bdach:
Not going to insist on the name given. Maybe it should be `SettingWithBaseValue` or similar, not sure.

frenzibyte:
Closing in favour of #11390 (this just utterly regresses code readability)

PR #11328, opened 2020-12-27
Fix mods using reference equality unless cast to `IMod`

frenzibyte:
As pointed out by @bdach.


osu.Game/Rulesets/Mods/IMod.cs
```diff
@@ -1,12 +1,11 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
-using System;
 using Newtonsoft.Json;
 
 namespace osu.Game.Rulesets.Mods
 {
-    public interface IMod : IEquatable<IMod>
```
peppy, line 9: any reason not to leave this on the interface too? was it never used?

frenzibyte: oh, good point actually

osu.Game/Online/API/APIMod.cs
```diff
@@ -52,7 +52,7 @@ public Mod ToMod(Ruleset ruleset)
             return resultMod;
         }
 
-        public bool Equals(IMod other) => Acronym == other?.Acronym;
+        public bool Equals(APIMod other) => Acronym == other?.Acronym;
```
peppy, line 55: You can just keep the `IMod` comparison and cast in the local one to keep the implementation in one place, I believe.

peppy (dismissed):

osu.Game/Online/API/APIMod.cs
frenzibyte: To be clear, does this mean to explicitly implement `Equals(IMod)` or the opposite?

peppy: Much of a muchness I think. Might as well leave it public?

frenzibyte: Upon thinking, I believe the right way is to make `Equals(IMod)` explicit, I don't think we should be allowing stuff like `Mod.Equals(APIMod)`, right?

bdach: That was how it was before, no? If anything that's a safer way to do it as I don't even know if there wasn't anything that was silently relying on that.

bdach: Also I don't know if making the `IMod` implementation explicit or implicit does anything even. `EqualityComparer`s will still use it anyway via the interface specification.

frenzibyte: This time, if the other `IMod` is not the same type as `this`, then `false` would be returned.
https://github.com/ppy/osu/blob/5efcdbd431153a902fd510d727f9b02400afcb66/osu.Game/Rulesets/Mods/Mod.cs#L152

bdach: `Mod` was actually already using strict type equality before this change.

frenzibyte: Ah, fixed.

bdach:
This PR technically still changes behaviour. Before, `DeserializedMod` and `APIMod` would compare equal with *any* other `IMod` with the same acronym, which is no longer the case. I'm not saying that that's technically good or bad, but I see no discussion of any potential breaking changes (or lack thereof). @frenzibyte did you check usages to make sure this doesn't break anything?

frenzibyte:
Yeah there aren't any at all, no usages of enumerable `IMod`s, and no direct calls to `Equals(IMod)`.

bdach:
I double-checked blame, and found https://github.com/ppy/osu/commit/6bdaca1e3bf0a3ce4411a5d53f9ae64e80c8aec1, which shows that that's not entirely true - what is now `PlaylistsPlayer` was using `Equals(IMod)`, but now binds to `Equals(Mod)`.

So while that statement was not necessarily completely correct, I don't think this should actually cause any problems, knock wood...

osu.Game/Scoring/ScoreInfo.cs
```diff
@@ -252,11 +252,12 @@ public IEnumerable<HitResultDisplayStatistic> GetStatisticsForDisplay()
         }
 
         [Serializable]
-        protected class DeserializedMod : IMod
+        protected class DeserializedMod : IMod, IEquatable<DeserializedMod>
         {
             public string Acronym { get; set; }
 
-            public bool Equals(IMod other) => Acronym == other?.Acronym;
+            bool IEquatable<IMod>.Equals(IMod other) => other is DeserializedMod them && Equals(them);
```
bdach, line 259: before I go pushing in changes, is there a reason this is the only one of three classes that got an explicit `IEquatable<IMod>` implementation instead of a normal public one?

frenzibyte:
> I double-checked blame, and found [6bdaca1](https://github.com/ppy/osu/commit/6bdaca1e3bf0a3ce4411a5d53f9ae64e80c8aec1), which shows that that's not entirely true - what is now `PlaylistsPlayer` was using `Equals(IMod)`, but now binds to `Equals(Mod)`.
> 
> So while that statement was not necessarily completely correct, I don't think this should actually cause any problems, knock wood...

Oh that one, it showed up on me but I was in the wrong branch, and thought that it's not using it, my bad... <_<

osu.Game/Scoring/ScoreInfo.cs
frenzibyte: nope. *I swear to god I had it public.*

bdach (approved):
seems fine... eventually

bdach (changes requested):
Going back to this yet again I realised that this most likely will break server side logic because state models (i.e. `MultiplayerRoomSettings`) use `Equals` implementations, which in turn use `SequenceEquals` on API mod collections, and given that `APIMod.Equals` does not take `Settings` into account, I can imagine that this will break users changing mod settings but not the mod itself.

peppy:
@frenzibyte wanna bring this one up to date or otherwise?

frenzibyte:
Will do in a bit.

peppy:
Going to need to cross-test this with the multiplayer server side to make sure we weren't relying on reference comparisons until now (as mentioned above).

frenzibyte:
> Going to need to cross-test this with the multiplayer server side to make sure we weren't relying on reference comparisons until now (as mentioned above).

https://github.com/ppy/osu-server-spectator/blob/54f03fb9d1dcc64dd542d02de49dd30a66864c3d/osu.Server.Spectator/Hubs/MultiplayerHub.cs#L324-L328

Yeah actually looks like it was relying on reference comparison, which would always return `false` in the above lines, i.e. silently failing.

I've pushed a solution and a test case for the issue pointed above, although I'm not quite sure whether it's good enough but that's as far as I have got.

smoogipoo:
I've tested that this also works with osu-server-spectator. Looks pretty alright.

smoogipoo (approved):

PR #11338, opened 2020-12-28
Allow null users to still be displayed in the participant list

### LINKED ISSUE #11302: ParticipantsPanel null reference
```
2020-12-25 06:31:19 [error]: An unhandled error has occurred.
2020-12-25 06:31:19 [error]: System.NullReferenceException: Object reference not set to an instance of an object.
2020-12-25 06:31:19 [error]: at osu.Game.Screens.Multi.RealtimeMultiplayer.Participants.ParticipantPanel.load()
2020-12-25 06:31:19 [error]: --- End of stack trace from previous location where exception was thrown ---
2020-12-25 06:31:19 [error]: at osu.Framework.Allocation.BackgroundDependencyLoaderAttribute.<>c__DisplayClass6_0.<CreateActivator>b__4(Object target, IReadOnlyDependencyContainer dc)
2020-12-25 06:31:19 [error]: at osu.Framework.Allocation.DependencyActivator.activate(Object obj, IReadOnlyDependencyContainer dependencies)
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Drawable.load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies)
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Drawable.Load(IFrameBasedClock clock, IReadOnlyDependencyContainer dependencies, Boolean isDirectAsyncContext)
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.loadChild(Drawable child)
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.AddInternal(Drawable drawable)
2020-12-25 06:31:19 [error]: at osu.Game.Screens.Multi.RealtimeMultiplayer.Participants.ParticipantsList.OnRoomChanged()
2020-12-25 06:31:19 [error]: at osu.Game.Online.RealtimeMultiplayer.StatefulMultiplayerClient.<>c__DisplayClass43_0.<osu.Game.Online.RealtimeMultiplayer.IMultiplayerClient.UserJoined>b__0()
2020-12-25 06:31:19 [error]: at osu.Framework.Threading.ScheduledDelegate.RunTaskInternal()
2020-12-25 06:31:19 [error]: at osu.Framework.Threading.Scheduler.Update()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Drawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Graphics.Containers.CompositeDrawable.UpdateSubTree()
2020-12-25 06:31:19 [error]: at osu.Framework.Platform.GameHost.UpdateFrame()
2020-12-25 06:31:19 [error]: at osu.Framework.Threading.GameThread.ProcessFrame()
```

Reproduced while sitting in a room as they played.

peppy:
The fix here is correcting the access of `user.Country`. The deicision to have null users display is because this is the best we can do (if osu-web could not resolve the user). We still want the users in the lobby to be aware of this user's presence, rather than hiding them from view.

osu-stable does a similar thing, showing these users as `[Loading]`. I decided to go with blank names instead because having *any* text there causes confusion. We can iterate on this in future design updates.

This also fixes the user lookup cache throwing a null reference if no matches were successful. This came up in the new tests I added.

Closes https://github.com/ppy/osu/issues/11302


bdach:
I imagine similar issues might pop up in other places now if we're accepting that API might return a null user. The leaderboard is most likely one of them, there are some unprotected accesses to a `User` returned from the lookup cache there. Not sure whether in scope of this PR or if should be a separate onw.

osu.Game.Tests/Visual/Multiplayer/TestSceneMultiplayerParticipantsList.cs
```diff
@@ -43,6 +43,16 @@ public void TestAddUser()
...
 
+        [Test]
+        public void TestAddNullUser()
+        {
+            AddAssert("one unique panel", () => this.ChildrenOfType<ParticipantPanel>().Select(p => p.User).Distinct().Count() == 1);
+
+            AddStep("add non-resolvable user", () => Client.AddNullUser(-3));
+
+            AddUntilStep("two unique panels", () => this.ChildrenOfType<ParticipantPanel>().Select(p => p.User).Distinct().Count() == 2);
+        }
```
frenzibyte, line 54: I think it would be better coverage if the null user is added in `TestSceneMultiplayerMatchSubScreen` instead, to ensure there shouldn't be any failures on any components in the whole screen.

frenzibyte (changes requested):

osu.Game.Tests/Visual/Multiplayer/TestSceneMultiplayerParticipantsList.cs
peppy: Can you give an example of what you are hoping to catch? I can't see any such usages.

peppy:
@bdach it's always been part of its API that it may return nulls, but definitely looks like the gameplay leaderboard is not handling this correctly. i'll do that as another PR>

bdach (approved):

osu.Game.Tests/Visual/Multiplayer/TestSceneMultiplayerParticipantsList.cs
frenzibyte: I'm saying that so that this test could track if a new component accesses `User` without null checking, one `Debug.Assert(User.User != null)` line would return this issue again.

peppy: I'm not sure that would help. Components should be tested directly, not at a screen level.

It's enforced now via https://github.com/ppy/osu/pull/11341/files#diff-0db59d73b332bc157b321082be2954f26fe9275f47385a233cd5c5710f69808aR27 so can't happen at compile time going forward.

PR #11339, opened 2020-12-28
Update fastlane and dependencies

peppy:
Trying to get iOS deployment flow online. This doesn't seem to help, but we haven't updated in a while so may as well.


bdach (approved):
no idea what this does but it sure looks like a dependency update alright

PR #11341, opened 2020-12-28
Fix incorrect null handling in GameplayLeaderboard

peppy:
This fixes another oversight of `UserLookupCache` potentially returning null on failure. It also adds annotations and xmldoc to avoid such issues at an earlier stage of development going forward.

We may decide in a future refactor to create an "unknown user" instance which is returned in these scenarios, avoiding the null return situation. This is already done for guest users at the `APIAccess` level.

### DIFF
--- a/osu.Game/Database/UserLookupCache.cs
+++ b/osu.Game/Database/UserLookupCache.cs
@@ -17,6 +17,12 @@ namespace osu.Game.Database
         [Resolved]
         private IAPIProvider api { get; set; }
 
+        /// <summary>
+        /// Perform an API lookup on the specified user, populating a <see cref="User"/> model.
+        /// </summary>
+        /// <param name="userId">The user to lookup.</param>
+        /// <param name="token">An optional cancellation token.</param>
+        /// <returns>The populated user, or null if the user does not exist or the request could not be satisfied.</returns>
         public Task<User> GetUserAsync(int userId, CancellationToken token = default) => GetAsync(userId, token);
 
         protected override async Task<User> ComputeValueAsync(int lookup, CancellationToken token = default)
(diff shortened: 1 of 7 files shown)


osu.Game/Database/UserLookupCache.cs
```diff
@@ -17,6 +18,13 @@ public class UserLookupCache : MemoryCachingComponent<int, User>
         [Resolved]
         private IAPIProvider api { get; set; }
 
+        /// <summary>
+        /// Perform an API lookup on the specified user, populating a <see cref="User"/> model.
+        /// </summary>
+        /// <param name="userId">The user to lookup.</param>
+        /// <param name="token">An optional cancellation token.</param>
+        /// <returns>The populated user, or null if the user does not exist or the request could not be satisfied.</returns>
+        [ItemCanBeNull]
```
bdach, line 27: huh, TIL

osu.Game/Screens/Play/HUD/MultiplayerGameplayLeaderboard.cs
```diff
@@ -65,7 +65,7 @@ private void load(OsuConfigManager config, IAPIProvider api)
                 var trackedUser = new TrackedUserData();
 
                 userScores[userId] = trackedUser;
-                var leaderboardScore = AddPlayer(resolvedUser, resolvedUser.Id == api.LocalUser.Value.Id);
+                var leaderboardScore = AddPlayer(resolvedUser, resolvedUser?.Id == api.LocalUser.Value.Id);
```
bdach, line 68: Mind if I also annotate `AddPlayer`'s first param as `[CanBeNull]` for completeness?

peppy: sure.

bdach (approved):

PR #11342, opened 2020-12-28
Fix user changes in multi song select not reverting on exit without confirmation

### LINKED ISSUE #11297: Current map should reset to multiplayer lobby selection when exiting song select menu
**Describe the bug:** If you change the beatmap, but exit the song select without confirming your choice, it will keep playing instead of the one selected in multiplayer lobby, and will try to load it if you start the match.

**Screenshots or videos showing encountered issue:** https://streamable.com/uvd20f

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5740545/runtime.log)

bdach:
Resolves #11297.

Aside from the beatmap, the mods and the ruleset also have to be reverted. A little awkward to keep three pieces of state inside the screen, but it seems like good practice to have the screen clean up after itself.

Also added tests because why not.


peppy (approved):
works as expected 👍🏻

PR #11344, opened 2020-12-28
Fix file and share associations on Android 

### LINKED ISSUE #11334: osu! is registered to open all file types on Android
**Describe the bug:** 

After installing the latest update, when opening most file types, osu! is presenting itself as an application that can open that file type. Notable examples are PDFs, APKs, etc.

Looking at osu!'s androidmanifest.xml, osu! is registered to open all mime types matching `application/*`

```xml
<intent-filter>
    <action android:name="android.intent.action.VIEW"/>
    <action android:name="android.intent.action.SEND"/>
    <category android:name="android.intent.category.DEFAULT"/>
    <data android:mimeType="application/*"/>
    <data android:pathPattern=".*.osz"/>
    <data android:pathPattern=".*.osk"/>
</intent-filter>
```

[This](https://stackoverflow.com/questions/1733195/android-intent-filter-for-a-particular-file-extension) suggests to use `.*\\.osz` and to specify `android:scheme` and `android:host`

I believe that if you don't use backslashes, a file named `fileosz` would get matched in addition to `file.osz`. Some comments suggest using four backslashes, so test those two filenames with the modified androidmanifest.xml

**Device information:**

Realme 6, Android 10
LG G Pad 8.3, Android 7.1.2

**Logs:**

Not needed as the issue is outside of osu!

Susko3:
Makes osu! registered as a file opener only for `.osz` and `.osk` files instead of for all files.

Narrows the share receiver to only accept mime types `application/octet-stream` and `application/zip` instead of `application/*`

Fixes #11334

### DIFF
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -19,7 +19,7 @@ namespace osu.Android
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osz", DataHost = "*", DataMimeType = "*/*")]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osk", DataHost = "*", DataMimeType = "*/*")]
     [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/octet-stream")]
-    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/zip")]
+    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeTypes = new[] { "application/zip", "application/octet-stream" })]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryBrowsable, Intent.CategoryDefault }, DataSchemes = new[] { "osu", "osump" })]
     public class OsuGameActivity : AndroidGameActivity
     {
--- a/osu.Android/OsuGameActivity.cs
+++ b/osu.Android/OsuGameActivity.cs
@@ -18,7 +18,6 @@ namespace osu.Android
     [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false, LaunchMode = LaunchMode.SingleInstance)]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osz", DataHost = "*", DataMimeType = "*/*")]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osk", DataHost = "*", DataMimeType = "*/*")]
-    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/octet-stream")]
     [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeTypes = new[] { "application/zip", "application/octet-stream" })]
     [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryBrowsable, Intent.CategoryDefault }, DataSchemes = new[] { "osu", "osump" })]
     public class OsuGameActivity : AndroidGameActivity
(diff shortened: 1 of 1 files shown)


bdach:
Please do avoid creating commits via the web interface and PRing from master in the future, as noted [in the contributing guidelines](https://github.com/ppy/osu/blob/master/CONTRIBUTING.md).

osu.Android/OsuGameActivity.cs
```diff
@@ -16,7 +16,10 @@
 namespace osu.Android
 {
     [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false, LaunchMode = LaunchMode.SingleInstance)]
-    [IntentFilter(new[] { Intent.ActionDefault, Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataPathPatterns = new[] { ".*\\.osz", ".*\\.osk" }, DataMimeType = "application/*")]
+    [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osz", DataHost = "*", DataMimeType = "*/*")]
+    [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osk", DataHost = "*", DataMimeType = "*/*")]
+    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/octet-stream")]
+    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/zip")]
```
Game4all, line 22: These two share intent filters can be merged into one without causing any breakage (was tested beforehand):
```suggestion
    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeTypes = new[] { "application/zip", "application/octet-stream" })]
```

osu.Android/OsuGameActivity.cs
```diff
@@ -16,7 +16,10 @@
 namespace osu.Android
 {
     [Activity(Theme = "@android:style/Theme.NoTitleBar", MainLauncher = true, ScreenOrientation = ScreenOrientation.FullUser, SupportsPictureInPicture = false, ConfigurationChanges = ConfigChanges.Orientation | ConfigChanges.ScreenSize, HardwareAccelerated = false, LaunchMode = LaunchMode.SingleInstance)]
-    [IntentFilter(new[] { Intent.ActionDefault, Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataPathPatterns = new[] { ".*\\.osz", ".*\\.osk" }, DataMimeType = "application/*")]
+    [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osz", DataHost = "*", DataMimeType = "*/*")]
+    [IntentFilter(new[] { Intent.ActionView }, Categories = new[] { Intent.CategoryDefault }, DataScheme = "content", DataPathPattern = ".*\\\\.osk", DataHost = "*", DataMimeType = "*/*")]
```
Game4all, line 20: Data MIME type `"*/*"` seems a bit too broad. Even if the file manager / browser fails to guess the correct MIME type of the file, it would set it to `application/octet-stream` (or even `application/*` for some obscure file managers), so `"application/*"` should fit in there for both of them instead of `"*/*"`

Game4all (commented):
Seems to fix cases where the game handles intents with unrelated .zip files / pdf files.

osu.Android/OsuGameActivity.cs
Susko3: Some file managers prompt users to select type of file (Image, Picture, Document, etc.) for unknown file extensions. A user might think: "osu! beatmap, that's like a song, right?"

Song would be MIME type `audio/*`. We want to present the user with the option of importing even if they select alternative options. We should leave `*/*` as is.

bdach:
I've tested this and it seems to behave closer to expectations. It is still possible to open zip files with lazer, but only via the "share" functionality which is probably fine (as it requires explicit user action).

osu.Android/OsuGameActivity.cs
```diff
@@ -16,7 +16,10 @@
...
+    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeType = "application/octet-stream")]
+    [IntentFilter(new[] { Intent.ActionSend }, Categories = new[] { Intent.CategoryDefault }, DataMimeTypes = new[] { "application/zip", "application/octet-stream" })]
```
bdach, line 22: Not sure about why this defines 2 mime types after the preceding line, which seems the same except for only defining ZIP instead of ZIP and octet stream, though.

Game4all: The preceding line can be removed , it was left over by the change I suggested.

Susko3: Removed the line with only ZIP

bdach (approved):
seems sane

bdach:
@Susko3 there was no need to merge `master` into this branch yourself. one of us would have done that eventually.

Susko3:
Okay, will keep that in mind for any future PRs

peppy (approved):

Issue #11345, opened 2020-12-28
In-game leaderboard got grayed out and didn't update its values


Sinamer:
**Describe the bug:** At the start of the match my panel in leaderboard became gray and did not update throughout the beatmap

**Screenshots or videos showing encountered issue:** 
![osu_2020-12-28_17-27-45](https://user-images.githubusercontent.com/53346335/103226390-ac227b00-4934-11eb-9223-07536f3a0b7c.png)

**osu!lazer version:** 2020.1228.0

**Logs:**

[runtime.log](https://github.com/ppy/osu/files/5747897/runtime.log)
[network.log](https://github.com/ppy/osu/files/5747898/network.log)

Gonzalo-Bruna:
shouldn't this issue be closed by #11351 ?

PR #11351, opened 2020-12-28
Change PlayingUsers population logic to match expectations

bdach:
Hopefully better fix for https://github.com/ppy/osu/issues/11345.

I know [we discussed not tracking](https://discord.com/channels/188630481301012481/188630652340404224/793157785478103080), but after looking at what would need to happen for the leaderboard to track quit if `PlayingUsers` got changed to a dumb LINQ/removed, I was no longer sure it was worth it anymore.

* Dumb LINQ means that each consumer of `PlayingUsers` has to do the differential check themselves if they want to get differential updates.
* Removing is even worse as each consumer has to remember and check every possible method of playing users changing (quitting out early, the local user leaving room entirely, the remote user leaving room entirely).

Half the issue was having the self-rolled logic in `MultiplayerMatchSubScreen`, I think. Now that they read from the same source this should not be an issue anymore.

Tested in a MP lobby on dev with @frenzibyte. Seemed to work OK.

### DIFF
--- a/osu.Game.Tests/NonVisual/Multiplayer/StatefulMultiplayerClientTest.cs
+++ b/osu.Game.Tests/NonVisual/Multiplayer/StatefulMultiplayerClientTest.cs
@@ -1,0 +1,57 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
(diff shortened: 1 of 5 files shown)


frenzibyte (dismissed):

peppy:
I've made a couple of clean-up changes. Seems fine otherwise.

peppy (approved):

PR #11353, opened 2020-12-28
Disable multiplayer action buttons after clicks to prevent double operations

### LINKED ISSUE #11317: Creating a room while joining a different one can crash the game
**Describe the bug:** Clicking to join a room, then quickly clicking create room several times will freeze the game and result in a crash shortly after.

https://user-images.githubusercontent.com/37958125/103144098-faacfb00-4723-11eb-9c62-38510e93fcf6.mp4

**Logs:**
[runtime.log](https://github.com/ppy/osu/files/5742576/runtime.log)

bdach:
Resolves https://github.com/ppy/osu/issues/11317 (and an item from https://github.com/ppy/osu/issues/11250).

PRing somewhat for discussion, as I'm not 100% sure of this one. The cached helper concept works well enough and maybe even cleans up some places (loading layer handling, for example), but...

Unfortunately with the way the UI is structured, to make the `CreateMultiplayerMatchButton` button disable work, I had to cache one helper at the very very root abstract `OnlinePlayScreen` level which is not great. Then `MultiplayerMatchSubScreen` caches another one over that one, which is dicey, but I think reusing the root-screen instance is equally so. Also considered naming the cached instances somehow but that was also ugly (to use `nameof()` for tests, I had to make things `internal`, etc. etc.)

I think I prefer having the `OngoingOperationTracker` instances scoped to particular operations, but it might be just me? I dunno. Suggestions how to make this less bad are welcome.


peppy:
For reference, what was the dicey part about only having one instance shared across all usages?

bdach:
My biggest problem with it personally was that right now its usages wouldn't overlap, but at some point they might (it's not unreasonable to expect to have two components do this sort of stuff simultaneously), at which point the pattern breaks down badly as written.

peppy:
So you're thinking there's an operation at an `OnlinePlayScreen` happening at the same time as a `MatchSubScreen` on the same multiplayer client? I'd hope this isn't a thing to keep things simple..

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerReadyButton.cs
```diff
@@ -142,7 +150,10 @@ private void onClick()
             else
             {
                 if (Room?.Host?.Equals(localUser) == true)
+                {
+                    gameplayStartTracker.BeginOperation();
```
peppy, line 154: This is impossible to follow. I can't find the end call.

I was hoping for an interface using `IDisposable` to make following the lifetime of the operations feasible:

```csharp
    /// <summary>
    /// Utility class to track ongoing online operations' progress.
    /// Can be used to disable interactivity while waiting for a response from online sources.
    /// </summary>
    public class OngoingOperationTracker
    {
        /// <summary>
        /// Whether there is an online operation in progress.
        /// </summary>
        public IBindable<bool> InProgress => inProgress;

        private readonly Bindable<bool> inProgress = new BindableBool();

        private LeasedBindable<bool> leasedInProgress;

        /// <summary>
        /// Begins tracking a new online operation.
        /// </summary>
        /// <exception cref="InvalidOperationException">An operation has already been started.</exception>
        /// <returns>An <see cref="IDisposable"/> which should be disposed when the operation and completed.</returns>
        public IDisposable BeginOperation()
        {
            if (leasedInProgress != null)
                throw new InvalidOperationException("Cannot begin operation while another is in progress.");

            leasedInProgress = inProgress.BeginLease(true);
            leasedInProgress.Value = true;

            return new InvokeOnDisposal(endOperation);
        }

        /// <summary>
        /// Ends tracking an online operation.
        /// Does nothing if an operation has not been begun yet.
        /// </summary>
        private void endOperation()
        {
            leasedInProgress?.Return();
            leasedInProgress = null;
        }
    }
```

peppy:
I'd still hope this can all use a single instance of this class. More than one and it falls apart in terms of readability. Still unsure of the case where you would have two ongoing operations.

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerReadyButton.cs
bdach: You've picked the worst usage by far, which is fair, but I'm not sure how to use `IDisposable` there. That disposable, once taken by the ready button, would have to find its way to the parent screen so that it can release the disable, at which point the original problem of having to push it through however many levels of the hierarchy returns (and requires not accidentally dropping the disposable along the way).

peppy:
I'd hope the full lease logic would be at the parent in the first place. Is that not possible?

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerReadyButton.cs
peppy: I'd hope the full lease logic would be at the parent in the first place. Is that not possible?

bdach:
That just flips the problem, doesn't it? To do that, the screen has to know when the actual UI click happens, so then it becomes a problem of jamming a click event callback through multiple levels, but down hierarchy instead of up. It's *possible*, but I hoped it would be avoidable.

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerReadyButton.cs
peppy: Okay, so the ready button logic is *really* not how we want to do things. It shouldn't be interacting with the API directly - this code should be moved to the point of construction and the action should be passed into the button IMO.

The Ready button itself should only be responsible for its display logic (it's fine for it to bind to the client for this purpose).

bdach: Yeah alright, I'll see what I can do. Let me get my bearings.

bdach:
This has sort of spiraled out of control as I was struggling not to upend half of multiplayer code but still ended up moving a decent chunk to make work anyway. Seems to work and I believe the current version should be closer to what you were expecting to see.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
```diff
@@ -178,6 +183,39 @@ public Task ChangeSettings(Optional<string> name = default, Optional<PlaylistIte
...
+                return false;
+
+            switch (localUser.State)
+            {
+                case MultiplayerUserState.Idle:
+                    await ChangeState(MultiplayerUserState.Ready);
+                    return false;
+
+                case MultiplayerUserState.Ready:
+                    if (Room?.Host?.Equals(localUser) == true)
```
peppy, line 205: hmm, not 100% keen on this existing here with the current method name (and the xmldoc not really mentioning *when* a game start is triggered).

osu.Game/Screens/OnlinePlay/Multiplayer/CreateMultiplayerMatchButton.cs
```diff
@@ -10,14 +10,29 @@ namespace osu.Game.Screens.OnlinePlay.Multiplayer
...
         {
             Triangles.TriangleScale = 1.5f;
 
             Text = "Create room";
 
-            ((IBindable<bool>)Enabled).BindTo(multiplayerClient.IsConnected);
+            isConnected = multiplayerClient.IsConnected.GetBoundCopy();
+            isConnected.BindValueChanged(_ => updateState());
+
+            operationInProgress = ongoingOperationTracker.InProgress.GetBoundCopy();
```
peppy, line 32: for safety we usually avoid doing binding logic in async load, can this be moved to `LoadComplete` without issue?

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerMatchFooter.cs
```diff
@@ -19,7 +20,13 @@ public class MultiplayerMatchFooter : CompositeDrawable
 
         public readonly Bindable<PlaylistItem> SelectedItem = new Bindable<PlaylistItem>();
 
+        public Action OnToggleReady
```
peppy, line 23: wow, this is weird syntax. what happens if you call the `get`? usually we fill it in with an exception.

peppy (changes requested):

peppy:
I'm not 100% on this because of the need to both resolve the `OngoingOperationTracker` *and* create a bound bindable. Will have a think about it. Might feel better if it was just part of the `OnlinePlayScreen` class (which is already cached), rather than its own class? Not sure.

osu.Game/Screens/OnlinePlay/Multiplayer/CreateMultiplayerMatchButton.cs
bdach: Yep, this is just an oversight.

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerMatchFooter.cs
bdach: You can't lest you get a compile-time error. I already kinda sorta used it once (see `IBarLine`). Can expose the getter, but not sure there's too much point?

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
bdach: Hm, good point. *Probably* saner to move this to the action itself I suppose (and duplicate in test for testing)...

How does something like this look?

```csharp
                OnReadyClick = async () =>
                {
                    readyClickedOperation = ongoingOperationTracker.BeginOperation();

					if (Client.IsHost && Client.LocalUser?.State == MultiplayerUserState.Ready)
					{
						await Client.StartGameplay();
						return;
					}

                    await Client.ToggleReady() // <- does just the toggle logic
					readyClickedOperation.Dispose();
					readyClickedOperation = null;
                }
```

(that's with surrounding names removed because those didn't make much sense in retrospect, either).

bdach:
So that'd entail resolving the screen instead of the tracker? Would probably still want a bindable to bind to on that.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
peppy: looks better, for sure.

osu.Game/Screens/OnlinePlay/Multiplayer/Match/MultiplayerMatchFooter.cs
peppy: Seems fine as long as it's compile time error.

bdach:
Is there anything else I need to be doing about this PR, or is this backburnered for now?

peppy:
Nothing required, been on my list but haven't had a good chance to review yet. Should get to it today.

osu.Game.Tests/Visual/Multiplayer/TestSceneMultiplayerReadyButton.cs
```diff
@@ -27,6 +31,11 @@ public class TestSceneMultiplayerReadyButton : MultiplayerTestScene
         private BeatmapManager beatmaps;
         private RulesetStore rulesets;
 
+        [Cached]
+        private OngoingOperationTracker ongoingOperationTracker = new OngoingOperationTracker();
```
peppy, line 35: think it's worth moving this to the base `MultiplayerTestScene`?

osu.Game/Screens/OnlinePlay/Lounge/LoungeSubScreen.cs
```diff
@@ -98,7 +106,13 @@ protected override void LoadComplete()
             base.LoadComplete();
 
             initialRoomsReceived.BindTo(RoomManager.InitialRoomsReceived);
-            initialRoomsReceived.BindValueChanged(onInitialRoomsReceivedChanged, true);
+            initialRoomsReceived.BindValueChanged(_ => updateLoadingLayer());
+
+            if (ongoingOperationTracker != null)
+            {
+                joiningRoom.BindTo(ongoingOperationTracker.InProgress);
```
peppy, line 113: not sure this should be called `joiningRoom`; are you okay with renaming to `operationInProgress` in line with other usages? reasoning is that a different operation could be in progress, apart from the local one (if we only cared about the local one than checking if `joiningRoomOperation` is null would be enough).

osu.Game/Screens/OnlinePlay/Multiplayer/MultiplayerMatchSubScreen.cs
```diff
@@ -196,13 +209,55 @@ public override bool OnBackButton()
...
+            client.ToggleReady()
+                  .ContinueWith(t =>
+                  {
+                      t.CatchUnobservedExceptions(true); // will run immediately.
+                      endOperation();
+                  });
+
+            void endOperation()
+            {
+                readyClickOperation?.Dispose();
```
peppy, line 244: I think a non-null assert here could work too, if you want to go crazy with assertions.

osu.Game/Screens/OnlinePlay/OngoingOperationTracker.cs
```diff
@@ -0,0 +1,53 @@
...
+            return new InvokeOnDisposal(endOperation);
+        }
+
+        /// <summary>
+        /// Ends tracking an online operation.
+        /// Does nothing if an operation has not been begun yet.
+        /// </summary>
+        private void endOperation()
+        {
+            leasedInProgress?.Return();
```
peppy, line 49: can this actually ever be null?

peppy (changes requested):

osu.Game/Screens/OnlinePlay/Multiplayer/MultiplayerMatchSubScreen.cs
bdach: Right. This is a vestige of an earlier revision in which I had the tracker as `CanBeNull` (but later realised that it does not need to be).

osu.Game/Screens/OnlinePlay/OngoingOperationTracker.cs
bdach: Theoretically this is permissive against a possible multiple-disposal of the operation, but it might well also be considered programmer error at that point. Not opposed to asserting non-null.

peppy (approved):

Issue #10267, opened 2020-09-27
don't adjust mania scroll speed to bpm changes within a map


LastExceed:
Even when using fixed scroll speed, BPM changes within the map still affect the scroll speed (relative to the map's default BPM). Most of the time mappers counter this using SVs to bring the scroll speed back to its default value, as drastic BPM changes are otherwise unplayable. This in turn makes SV mapping harder as things like teleport's now need to be normalized to a value that isn't x1.0, and throws off beginner mappers who don't know about this. I understand the appeal of BPM scaling from a game design POV, but it's pretty much incompatible with the user's ability to configure scroll speed as most players set their scroll speed as fast as they can read to maximize precision of visual aiming and minimize the number of notes that are on screen at a time (see also: lane covers).

Therefore I suggest making this effect exclusive to the "scale osu!mania scroll speed with BPM" option (if we're even going to bring that back in lazer, see #2870).

This would of course break every map that was re-normalized using SVs (which includes ranked maps), so we'd have to account for that when parsing legacy maps. There's also aspire-like maps that use BPM changes to work around the SV bounds, but that's a different topic. more to that here #10266

peppy:
I don't *think* this is something we want because it basically limits how a mapper can... map.

ReiFan49:
as an unranked mod sure it's a good idea to practice, but still a "constant speed mod" usually breaks what mapper's message throughout the beatmap itself

LastExceed:
> it basically limits how a mapper can... map

@peppy please elaborate. How does this proposal limit anything ?

> a "constant speed mod" usually breaks what mapper's message

@ReiFan49 I am not proposing a constant speed mod (that would be #9802). This proposal only removes BPM based speed changes which can be replaced by SVs if necessary. And regarding "breaking" anything, I explained in the OP that this change needs to be accounted for when parsing legacy maps to avoid any existing maps being changed.

abraker95:
To paraphrase @LastExceed, if you have a beatmap going at 180 BPM and want a short section that is 90 BPM you need:
```
uninherited timing point 90 BPM
inherited timing point 2.0x
uninherited timing point 180 BPM
inherited timing point 0.5x
```
Because uninherited timing points change scroll speed. So inherited timing points need to be there to counteract that. This asks for the uninherited timing point to not affect scroll speed so that those extra inherited timing points would not be needed. So it becomes:
```
uninherited timing point 90 BPM
uninherited timing point 180 BPM
```

[...]

PR #11354, opened 2020-12-28
Add "Constant" osu!mania mod

### LINKED ISSUE #9802: Add "Constant" mod in mania
The mod enforces beatmap play with a constant scrolling speed despite changing with BPM, which is useful to practice SV maps. Also, mappers don't need to create a 'non-sv' difficulty. However, it is not fair to players that played with SV, the score could be nerfed or unranked.

"Const" mod from Malody

chev2:
Resolves #9802 
This adds a new mod to osu!mania that makes all the slider velocities/speed changes the default value (1).
It only affects difficulty points' speed multiplier as of now. BPM timing points are left unchanged, which may cause issues stated in #10267 (speeding up BPM, and slowing down scroll velocity).
The mod is set to unranked.

### DIFF
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
@@ -3,6 +3,7 @@
 
 using System.Linq;
 using osu.Framework.Graphics.Sprites;
+using osu.Framework.Utils;
 using osu.Game.Beatmaps;
 using osu.Game.Beatmaps.ControlPoints;
 using osu.Game.Rulesets.Mods;
@@ -39,9 +40,9 @@ namespace osu.Game.Rulesets.Mania.Mods
 
                     DifficultyControlPoint diffControlPoint = new DifficultyControlPoint();
 
-                    diffControlPoint.SpeedMultiplierBindable.MinValue = double.MinValue; // Uncapped minimum value
+                    diffControlPoint.SpeedMultiplierBindable.MinValue = Precision.DOUBLE_EPSILON; // Uncapped minimum value
                     diffControlPoint.SpeedMultiplierBindable.MaxValue = double.MaxValue; // Uncapped maximum value
-                    diffControlPoint.SpeedMultiplierBindable.Precision = 0.000000001d; // Arbitrary value, may need changing
+                    diffControlPoint.SpeedMultiplierBindable.Precision = Precision.DOUBLE_EPSILON;
                     diffControlPoint.SpeedMultiplier = initialBPM / timingControlPoint.BPM; // Counteract BPM velocity
 
                     group.Add(diffControlPoint);
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Beatmaps;
+using osu.Game.Rulesets.Mods;
+
+namespace osu.Game.Rulesets.Mania.Mods
+{
+    public class ManiaModConstant : Mod, IApplicableToBeatmap
+    {
+        public override string Name => "Constant";
+        public override string Acronym => "CT";
+        public override double ScoreMultiplier => 1;
+        public override string Description => "No more tricky note speed changes!";
+        public override IconUsage? Icon => FontAwesome.Solid.ArrowDown;
+        public override ModType Type => ModType.Conversion;
+        public override bool Ranked => false;
+
+        public void ApplyToBeatmap(IBeatmap beatmap)
+        {
+            // Get all difficulty points
+            var difficultyPoints = beatmap.ControlPointInfo.DifficultyPoints;
+
+            foreach (var difficultyPoint in difficultyPoints)
+            {
+                // Set this difficulty point's speed multiplier to 1
+                difficultyPoint.SpeedMultiplier = 1;
+            }
+        }
+    }
+}
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
@@ -12,8 +12,8 @@ namespace osu.Game.Rulesets.Mania.Mods
 {
     public class ManiaModConstant : Mod, IApplicableToBeatmap
     {
-        public override string Name => "Constant";
-        public override string Acronym => "CT";
+        public override string Name => "Fixed Scroll";
+        public override string Acronym => "FS";
         public override double ScoreMultiplier => 1;
         public override string Description => "No more tricky note speed changes!";
         public override IconUsage? Icon => FontAwesome.Solid.BalanceScale;
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
@@ -33,7 +33,15 @@ namespace osu.Game.Rulesets.Mania.Mods
                 // Get the difficulty control point of this group, if any
                 var difficultyControlPoints = group.ControlPoints.OfType<DifficultyControlPoint>().ToArray();
 
-                // If there is a timing control point
+                // If this group has a difficulty point, remove it
+                if (difficultyControlPoints.Any())
+                {
+                    var diffControlPoint = difficultyControlPoints[0];
+
+                    group.Remove(diffControlPoint);
+                }
+
+                // If this group has a timing point, add counteracting difficulty point
                 if (timingControlPoints.Any())
                 {
                     var timingControlPoint = timingControlPoints[0];
@@ -45,14 +53,8 @@ namespace osu.Game.Rulesets.Mania.Mods
                     diffControlPoint.SpeedMultiplierBindable.Precision = Precision.DOUBLE_EPSILON;
                     diffControlPoint.SpeedMultiplier = initialBPM / timingControlPoint.BPM; // Counteract BPM velocity
 
+                    // Add the new difficulty point to this group
                     group.Add(diffControlPoint);
-                }
-                // Else, there is no timing control point, but a difficulty control point
-                else if (difficultyControlPoints.Any())
-                {
-                    var diffControlPoint = difficultyControlPoints[0];
-
-                    group.Remove(diffControlPoint);
                 }
             }
         }
--- a/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
+++ b/osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+using System.Linq;
 using osu.Framework.Graphics.Sprites;
 using osu.Game.Beatmaps;
 using osu.Game.Beatmaps.ControlPoints;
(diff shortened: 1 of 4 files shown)


osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public void ApplyToBeatmap(IBeatmap beatmap)
+        {
+            // Get all difficulty points
+            var difficultyPoints = beatmap.ControlPointInfo.DifficultyPoints;
+
+            foreach (var difficultyPoint in difficultyPoints)
+            {
+                // Set this difficulty point's speed multiplier to 1
+                difficultyPoint.SpeedMultiplier = 1;
```
bdach, line 28: So what's the plan regarding BPM changes here? I don't think generating a correct set of difficulty control points that would compensate BPM would be overly difficult, so why is this not done? Are you expecting turning off BPM scaling to become a game-wide feature?

chev2: The compensation for BPM didn't cross my mind, I'll look into that. I do not expect BPM scaling to change. The main issue I was facing was that if I were to reset all timing points as well, there would be the chance that an actual change in BPM on the map would be reset and make the rest of the map off-sync.

bdach: I think adding a difficulty control point for each timing control point, with multiplier equal to (initial BPM / current BPM) should do the trick. In fact I think that is already how constant scrolling speed is mapped? And yeah, just nuking timing points isn't going to work.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -0,0 +1,32 @@
...
+using osu.Framework.Graphics.Sprites;
+using osu.Game.Beatmaps;
+using osu.Game.Rulesets.Mods;
+
+namespace osu.Game.Rulesets.Mania.Mods
+{
+    public class ManiaModConstant : Mod, IApplicableToBeatmap
+    {
+        public override string Name => "Constant";
+        public override string Acronym => "CT";
```
abraker95, line 13: Change name to "No SV" and change the acronym to "NSV". NSV is better recognized by mania players as an indication in versions of maps without SV.

bdach: Don't think we do three letter acronyms anywhere. Probably won't fit in button?

Joehuu: https://github.com/ppy/osu/blob/575a00dd7deafdc6a9194dec59d79737db44e0ad/osu.Game.Rulesets.Mania/Mods/ManiaModKey10.cs#L10

bdach: Fair enough on that point - still not sure about using SV as an acronym directly in the name. Sort of inaccessible for newcomers (mania doesn't even really have "sliders").

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -0,0 +1,32 @@
...
+        public override double ScoreMultiplier => 1;
+        public override string Description => "No more tricky note speed changes!";
+        public override IconUsage? Icon => FontAwesome.Solid.ArrowDown;
```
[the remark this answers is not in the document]
chev2: I think `FontAwesome.Solid.BalanceScale` looks the best, I'll change it to that.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
abraker95: I like to thing mania "SV" stands for scroll velocity, but any mania player knows what SV means because it's hard not to come across the term somewhere.

The wiki details it [here](https://osu.ppy.sh/wiki/en/Guides/osu%21mania_Mapping_Guide#slider-velocity-(sv)). [A pinned thread in the osu!mania subforum](https://osu.ppy.sh/community/forums/topics/155780) uses the term for describing how to map SV. [Countless of mappers use "NSV"](https://osu.ppy.sh/beatmapsets?m=3&q=nsv&s=graveyard) in the difficulty name to indicate it's the same map without SV.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -0,0 +1,50 @@
...
+            }
+
+            // Correct changes in speed caused by timing points by adding an associated difficulty point
+            foreach (var timingPoint in timingPoints)
+            {
+                // Create a new difficulty point to counteract timing point
+                DifficultyControlPoint diffControlPoint = new DifficultyControlPoint
+                {
+                    SpeedMultiplier = initialBPM / timingPoint.BPM
+                };
```
chev2, line 43: Regarding this, since difficulty points have a capped speed multiplier value of 0.1-10, there will be cases where the calculated `initialBPM / timingPoint.BPM` exceeds this range. One beatmap that has this is [this one](https://osu.ppy.sh/beatmapsets/586848#mania/1242696). Any thoughts to work around this?

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
bdach: I still think that "SV" as two letters should not be in the name, regardless of if it's established jargon. I'd be fine with "Constant Scroll" or something, as long as there are actual words there and not two-letter acronyms.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -20,30 +21,38 @@ public class ManiaModConstant : Mod, IApplicableToBeatmap
...
+                var difficultyControlPoints = group.ControlPoints.OfType<DifficultyControlPoint>().ToArray();
+
+                // If there is a timing control point
+                if (timingControlPoints.Any())
+                {
+                    var timingControlPoint = timingControlPoints[0];
+
+                    DifficultyControlPoint diffControlPoint = new DifficultyControlPoint();
+
+                    diffControlPoint.SpeedMultiplierBindable.MinValue = double.MinValue; // Uncapped minimum value
```
bdach, line 42: Yeah I'm not sure this should be allowed to be negative (or even zero). I'd agree with

```suggestion
                    diffControlPoint.SpeedMultiplierBindable.MinValue = Precision.DOUBLE_EPSILON; // Uncapped minimum value
```

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -20,30 +21,38 @@ public class ManiaModConstant : Mod, IApplicableToBeatmap
...
+                    diffControlPoint.SpeedMultiplierBindable.MaxValue = double.MaxValue; // Uncapped maximum value
+                    diffControlPoint.SpeedMultiplierBindable.Precision = 0.000000001d; // Arbitrary value, may need changing
```
bdach, line 44: Similarly

```suggestion
                    diffControlPoint.SpeedMultiplierBindable.Precision = Precision.DOUBLE_EPSILON;
```

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
```diff
@@ -20,30 +21,38 @@ public class ManiaModConstant : Mod, IApplicableToBeatmap
...
+                    diffControlPoint.SpeedMultiplier = initialBPM / timingControlPoint.BPM; // Counteract BPM velocity
+
+                    group.Add(diffControlPoint);
+                }
+                // Else, there is no timing control point, but a difficulty control point
+                else if (difficultyControlPoints.Any())
```
bdach, line 50: Not sure this is correct - I don't think there's anything stopping a group from having a timing point _and_ a difficulty control point? Should just probably be

```suggestion
                if (difficultyControlPoints.Any())
```

bdach (dismissed):

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
chev2: Alright, I'll fix this as well as the precision.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
chev2: If it were to check for a difficulty point after it checks for a timing point, there's a possibility that it will remove the difficulty point, which will undo the timing point's counteraction. From my understanding, at least. That's why it's an `else if`. If the group has both a timing and difficulty point in the first place, the difficulty point will be overridden anyway.

bdach: Oh, because there's a difficulty point potentially added before... Yeah I think that's a bit backwards. Should clean up the group from any difficulty points first before proceeding to add own overrides here I reckon.

osu.Game.Rulesets.Mania/Mods/ManiaModConstant.cs
chev2: I'm fine with "Constant Scroll" too, since gives a better idea as to what this mod does compared to just "Constant". "Constant Scrollspeed" might also work, but it might also be a bit long for a name.
Either way, I think the acronym "CT" should stay the same. "CS" is already an acronym for circle size, but that's not necessarily a mania thing, so I don't know.

peppy: how about FS (fixed scroll)? if we're looking to avoid acronym clashes and using confusing terms like SV.

bdach:
On re-testing this, there seem to be blocking issues:

* Hold note duration does not look to change properly on maps with changing BPM. This is most likely due to changing the control points too late in the parsing/conversion process? Not sure right now.
* Worse yet, if the mod is activated and then de-activated, the control points won't get restored to their previous state. This will be due to internal caching of the beatmap and not restoring the changes to control points which were made.

The second point is the bigger issue and I believe this is also an issue that the editor has, and will need addressing properly separately.

bdach:
So the timing point pollution issue is fixed, and I can believe that it was causing the mis-timings I was seeing (because the mod could apply itself to a beatmap multiple times), but there's still one issue left - namely that this mod doesn't work on converts, because of this:

https://github.com/ppy/osu/blob/dd809df07654bb0b8ba104b6f078c874272aa752/osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs#L78-L80

Don't really have any smart ideas how to counteract this, save for just publicising this `isForCurrentRuleset` field...

bdach:
@smoogipoo when you have the time, I'd like to know whether you're fine with the [workaround proposed above](https://github.com/ppy/osu/pull/11354#issuecomment-758225968) which was applied in fe92083 by @chev2, or if you have a better idea on how to make this work.

smoogipoo:
Couldn't this PR be implemented by using the `ConstantScrollAlgorithm`? It would remove the mess of adding "counteracting" control points and the preserve bool.

I'd say find a way to make the scroll algorithm of `DrawableScrollingRuleset` editable.

bdach:
I'd looked into that but ditched the effort due to reasons I don't currently remember. I think I couldn't figure out a way to get the drawable hierarchy to switch from one scrolling mode to another, due to it being cached and all that.

Would have probably warranted a separate pull and I don't even know if we have the bandwidth to look at that right now.

smoogipoo:
I think `IApplicableToDrawableRuleset` should be usable. As long as it's done before the first update frame, SHOC should handle fine.

bdach:
Making the algorithm mutable itself was the fiddly part. I might look into this again but no promises.

peppy:
maybe worth getting this one merged and converting it to use a scroll algorithm as a second step? the implementation is quite isolated so i don't see much harm in it.

smoogipoo:
Do note that if this is merged, your timing point cloning PR needs changes.

peppy:
Even without the mod applied? Maybe I'm misunderstanding the gravity of the ruleset-side change.

smoogipoo:
Yes. Bottom line is this mod modifies the beatmap's `ControlPointInfo`. It needs a mod interface similar to `IApplicableToBeatmapDifficulty` that clones it to be safe.

osu.Game.Rulesets.Mania/UI/DrawableManiaRuleset.cs
```diff
@@ -55,6 +55,11 @@ public class DrawableManiaRuleset : DrawableScrollingRuleset<ManiaHitObject>
         // Stores the current speed adjustment active in gameplay.
         private readonly Track speedAdjustmentTrack = new TrackVirtual(0);
 
+        /// <summary>
+        /// Whether difficulty points should be taken into account, even on convert maps.
+        /// </summary>
+        public bool PreserveDifficultyPoints { get; set; }
```
peppy, line 61: Re-thinking this one, i really don't think this should exist..

bdach: I agree and was reluctant to suggest it in the first place, but in the grand scheme of things I don't deem this pull important enough to spend time on doing things properly by substituting the scroll algorithm.

PR #11358, opened 2020-12-29
Remove selection polling from multiplayer

### LINKED ISSUE #11348: Lobby sorting is not stable
Seems to be easiest to reproduce by joining a game, exiting out, then clicking the same game (or another near it).

peppy:
Looks like this was just copy-paste without any thought into whether it should exist. It really shouldn't exist.

This is a thing for the playlists system because the *whole system* there relies on polling the web API to get updated information. In the case of mutliplayer, we hand off all communications to the realtime server at the point of joining the rooms.

The argument that this was there to do faster polling on the selection isn't valid since the polling times were the same for both cases.

Closes #11348.


bdach:
While I don't have any issues with this on principle, do you know why this doesn't seem to break in playlists? That also has two simultanous polls, but they don't seem to be fighting each other as much as these multiplayer ones...

peppy:
No idea. maybe just good luck.

bdach (approved):

PR #11359, opened 2020-12-29
Silence network errors that don't need to be shown to the user

### LINKED ISSUE #11347: Errors spam when network disconnected
I noticed that  the "connection to multiplayer server was lost" is shown even when not in a multiplayer game (it shouldn't be).

But on top of this, many more error popups constantly appear when no network connection is available:

image

```log
2020-12-28 16:05:26 [error]: Error occurred while handling an API request.
2020-12-28 16:05:26 [error]: System.AggregateException: One or more errors occurred. (No such host is known.) (Value cannot be null. (Parameter 'value'))
2020-12-28 16:05:26 [error]: ---> System.Net.Http.HttpRequestException: No such host is known.
2020-12-28 16:05:26 [error]: ---> System.Net.Sockets.SocketException (11001): No such host is known.
2020-12-28 16:05:26 [error]: at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: --- End of inner exception stack trace ---
2020-12-28 16:05:26 [error]: at System.Net.Http.ConnectHelper.ConnectAsync(String host, Int32 port, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: at System.Net.Http.HttpConnectionPool.ConnectAsync(HttpRequestMessage request, Boolean allowHttp2, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: at System.Net.Http.HttpConnectionPool.CreateHttp11ConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: at System.Net.Http.HttpConnectionPool.GetHttpConnectionAsync(HttpRequestMessage request, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: at System.Net.Http.HttpConnectionPool.SendWithRetryAsync(HttpRequestMessage request, Boolean doRequestAuth, CancellationToken cancellationToken)
2020-12-28 16:05:26 [error]: at System.Net.Http.RedirectHandler.SendAsync(HttpRequestMessage request, CancellationToken 
```

peppy:
- Fix error message being surfaced to user on multiplayer disconnection when not in room
- Handle SocketExceptions and HttpRequestExceptions more silently

These can occur when a network connection is completely unavailable (ie.  host resolution failures are occurring). Currently these would appear as important errors which spammed the notification overlay every retry forever, while no network connection is available.

I also took this opportunity to remove a lot of `bool` passing which was no longer in use (previously the fail count / retry process was different to what we have today).

- [x] Depends on https://github.com/ppy/osu-framework/pull/4115
- Closes #11347

2cb84c5111 (Dean Herbert, 63 minutes ago)

### DIFF
--- a/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
+++ b/osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
@@ -84,7 +84,7 @@ namespace osu.Game.Online.Multiplayer
             IsConnected.BindValueChanged(connected =>
             {
                 // clean up local room state on server disconnect.
-                if (!connected.NewValue)
+                if (!connected.NewValue && Room != null)
                 {
                     Logger.Log("Connection to multiplayer server was lost.", LoggingTarget.Runtime, LogLevel.Important);
                     LeaveRoom().CatchUnobservedExceptions();
--- a/osu.Android.props
+++ b/osu.Android.props
@@ -52,6 +52,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="ppy.osu.Game.Resources" Version="2020.1202.0" />
-    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.1228.0" />
+    <PackageReference Include="ppy.osu.Framework.Android" Version="2020.1229.0" />
   </ItemGroup>
 </Project>
--- a/osu.Game/Online/API/APIAccess.cs
+++ b/osu.Game/Online/API/APIAccess.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Net;
 using System.Net.Http;
+using System.Net.Sockets;
 using System.Threading;
 using System.Threading.Tasks;
 using Newtonsoft.Json.Linq;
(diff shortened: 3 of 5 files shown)


osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
```diff
@@ -84,7 +84,7 @@ protected StatefulMultiplayerClient()
             IsConnected.BindValueChanged(connected =>
             {
                 // clean up local room state on server disconnect.
-                if (!connected.NewValue)
+                if (!connected.NewValue && Room != null)
```
bdach, line 87: Did this actually silence/fix anything? It looks like `LeaveRoom` should be handling the null case correctly already (albeit slightly awkwardly because inheritance) - maybe that logic should be trimmed if we're moving this up here?

bdach (commented):
No issues in principle with the `APIAccess` changes.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
peppy: This should fix the error message showing up at the main menu.

bdach: Ohhhhh right gotcha. Got totally fixated on the `LeaveRoom` operation while simultaneously glossing over the log statement that shows up within.

bdach (dismissed):
looks good - just need the framework bump to go through

bdach (approved):

PR #11361, opened 2020-12-29
Fix being able to enter online screens while API is in a failing mode

peppy:
Without this change, the user could enter the screens from the main menu in APIState.Failing, only to be returned back to the menu in milliseconds. This changes the state check to include all non-online states for the time being.

Also provides better messaging when an online screen needs to be exited due to a connection loss.

### DIFF
--- a/osu.Game/Screens/Menu/ButtonSystem.cs
+++ b/osu.Game/Screens/Menu/ButtonSystem.cs
@@ -156,7 +156,7 @@ namespace osu.Game.Screens.Menu
 
         private void onMultiplayer()
         {
-            if (!api.IsLoggedIn || api.State.Value != APIState.Online)
+            if (api.State.Value != APIState.Online)
             {
                 notifications?.Post(new SimpleNotification
                 {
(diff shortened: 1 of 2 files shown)


osu.Game/Screens/Menu/ButtonSystem.cs
```diff
@@ -156,7 +156,7 @@ private void load(AudioManager audio, IdleTracker idleTracker, GameHost host)
 
         private void onMultiplayer()
         {
-            if (!api.IsLoggedIn)
+            if (!api.IsLoggedIn || api.State.Value != APIState.Online)
```
bdach, line 159: Seems a little weird to check both `IsLoggedIn` and state - at this point just the latter might be enough? Is this future-proofing and at some point we'll want to allow being online (as in there being a physical connection to the internet) but not logged in? Because right now I'm pretty sure you can't have the latter without the former.

peppy: probably can remove the first.

bdach: As a follow-up thought, the notification message should probably be reworded to "You gotta be online..." to match, to avoid user confusion (they're possibly logged in already after all, it's just their connection that's dying). I'll grab both things.

bdach (approved):

PR #11364, opened 2020-12-29
Add SV multipliers to taiko difficulty mods

### LINKED ISSUE #10663: HR/EZ don't adjust taiko SV
**Describe the bug:** In Stable, HR doubles taiko SV and EZ turns it down a little bit (I think 0.75x?). For some people using those mods is the only way to make certain maps readable. For some reason I thought this was known, but I couldn't find any other issues other than the proposal for DifficultyAdjust.

**Screenshots or videos showing encountered issue:** https://www.youtube.com/watch?v=4DMXWGRn7-M

Firmatorenio:
Fixes #10663.

Note that things still don't look quite the same as on Stable due to the hitobject drawables having slightly different scaling.
SV coefficients have been deduced by screenshotting how a stream of notes looks on Stable at 1x SV multiplier with the mods, then adjusting that multiplier in editor until it looks the same (EZ looking best at 0.8x and HR at 1.87x).

Stable nomod at 1x:
image
Lazer nomod at 1x (or with either mod before this PR):
image

Stable at 1x+EZ:
image
Stable at .8x nomod:
image
This PR at 1x+EZ:
image

Stable at 1x+HR:
image
Stable at 1.87x nomod:
image
This PR at 1x+HR:
image


osu.Game.Rulesets.Taiko/Mods/TaikoModHardRock.cs
```diff
@@ -9,5 +10,13 @@ public class TaikoModHardRock : ModHardRock
     {
         public override double ScoreMultiplier => 1.06;
         public override bool Ranked => true;
+
+        private const double slider_multiplier = 1.87;
```
bdach, line 14: As I said on discord, there *is* a 1.4 value in stable, *but* it's fudged further if hidden is *not* active.

It actually comes out that your reverse-engineered factor is ~1.33 times, or 4/3ths, larger than 1.4, which makes sense if you were testing at a 16:9 resolution (as that's the proportion of widths of a 16:9 and a 4:3 screen if they both have the same height). But honestly, between the dependence on both hidden being active and the window resolution, I'm not even sure where to *begin* porting that to lazer.

Firmatorenio: As of right now, the 'field of view' of the bar doesn't really change with the resolution and scales to always have the same amount of notes, amount of playfield, to be visible, and it's roughly the same as it is on stable with 16:9 resoluton - so I want to say it makes sense to have it always as if it's 16:9.

Hidden has yet to be implemented, and I assume the disparity between 16:9 and 4:3 playstyles is exactly the reason - but if the field has already been scaled to be the same regardless of the resolution, Lazer's Hidden implementation will most likely be attached to that scaling and stay fixed too? I think that makes the most sense to make all players have equal conditions in the end.

bdach: Personally I could be fine with this, *provided* that it is documented properly, for example:

```suggestion
		/// <summary>
		/// Multiplier factor applied to slider speed.
		/// </summary>
		/// <remarks>
		/// This factor is made up of two parts: the base part (1.4)
		/// and the aspect ratio adjustment (1.33). Stable applies the latter
		/// by dividing the width of the user's display by the width of a display
		/// with the same height, but a 4:3 ratio.
		/// TODO: This adjustment is only active if Hidden is off.
		/// TODO: Revisit if taiko playfield scaling ever changes away from a hard-coded 16:9
		/// (see https://github.com/ppy/osu/issues/5685).
		/// </remarks>
        private const double slider_multiplier = 1.4 * 1.33;
```

But at that point I'm not sure there's enough of the issue solved here to justify merging.

Firmatorenio: I'm alright with adding the remarks, but I'm dropping the HD-related TODO because, as discussed on Discord, it doesn't actually look like HR is the mod that changes its behavior (from its 'lone' behavior at that aspect ratio) when both are applied, HD is. The actual note velocity with HR seems to stay the same regardless of whether HD is applied or not.

Assuming the hardcoded 16:9 stays, it wouldn't make sense for the new HD to be based around a 4:3 ratio like it is on Stable. And even if hard 16:9 doesn't stay, you can always add another, more global, not mod- but screen-width- dependent multiplication factor to the overall slider velocity that will scale off of 16:9 (the direction of making all the players see the same amount of notes regardless of the aspect ratio makes perfect sense. If only everyone being happy with that was achievable since 4:3 players are clearly used to a different experience from the 16:9 players.)

peppy:
No real test coverage of this but i guess it's fine for now.

PR #11379, opened 2020-12-31
Fix PerformFromMenuRunner failing if CurrentScreen is null

peppy:
Found this while testing. If the window decoration close button is used to exit the game, there's a potential that a `PerformFromMenu` operation is started which can never resolve the menu (due to the screen already being switched to `IntroTriangles` to play the outro). This means the scheduled operations will continue to run while the game shuts down. During this process, the screen stack may be empty for one (or more?) frames, which causes the nullref.

### DIFF
--- a/osu.Game/PerformFromMenuRunner.cs
+++ b/osu.Game/PerformFromMenuRunner.cs
@@ -73,13 +73,16 @@ namespace osu.Game
             // find closest valid target
             IScreen current = getCurrentScreen();
 
+            if (current == null)
+                return;
+
             // a dialog may be blocking the execution for now.
             if (checkForDialog(current)) return;
 
             game?.CloseAllOverlays(false);
 
             // we may already be at the target screen type.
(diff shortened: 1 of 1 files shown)


bdach:
Seems reasonable - but while we're here, the following also caught my attention:

https://github.com/ppy/osu/blob/6ebb267c66d5135527378b163ae72abb2f308ce1/osu.Game/PerformFromMenuRunner.cs#L139-L143

This is also potentially calling `getCurrentScreen()` again and potentially getting a different screen than `checkCanComplete()` - maybe worth passing it down instead of calling locally?

peppy:
Can we just inline it? Looks like it's only used in one place.

bdach:
Yeah that also works.

bdach:
Have pushed that in. I'm hopefully competent enough to be able to inline a method correctly but you can have a second look to be sure.

bdach (approved):

PR #11380, opened 2020-12-31
Catch multiplayer client-related unobserved exceptions better

bdach:
Silencing an exception from a task continuation requires accessing `task.Exception` in any way, which was not done previously if `logOnError` was false.

To resolve without having to worry whether the compiler will optimise away a useless access or not, just always log, but switch the logging level. The unimportant errors will be logged as debug and therefore essentially silenced on release builds (but could still be potentially useful in debugging).

---

This can be semi-tested by gutting chosen `MultiplayerClient` operations (like `ChangeState` or `LeaveRoom`) to look like:

```csharp
public async Task LeaveRoom()
{
	await Task.Delay(100);
	throw new InvalidOperationException();
}
```

The async delay is important (without it the error will just be thrown synchronously, which `CatchUnobservedExceptions()` does nothing for anyway).

### DIFF
--- a/osu.Game/Extensions/TaskExtensions.cs
+++ b/osu.Game/Extensions/TaskExtensions.cs
@@ -1,7 +1,11 @@
 // Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
 // See the LICENCE file in the repository root for full licence text.
 
+#nullable enable
+
+using System;
 using System.Threading.Tasks;
+using osu.Framework.Extensions.ExceptionExtensions;
 using osu.Framework.Logging;
 
 namespace osu.Game.Extensions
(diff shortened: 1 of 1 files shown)


osu.Game/Extensions/TaskExtensions.cs
```diff
@@ -13,13 +17,19 @@ public static class TaskExtensions
...
+        /// </param>
+        public static void CatchUnobservedExceptions(this Task task, bool logAsError = false)
         {
             task.ContinueWith(t =>
             {
-                if (logOnError)
-                    Logger.Log($"Error running task: {t.Exception?.Message ?? "unknown"}", LoggingTarget.Runtime, LogLevel.Important);
+                Exception? exception = t.Exception?.AsSingular();
+                if (logAsError)
+                    Logger.Error(exception, $"Error running task: {exception?.Message ?? "(unknown)"}", LoggingTarget.Runtime, true);
```
peppy, line 30: Logger.Error is a bit of a special one (submits error logs, but also already includes the message in the output, i think?). I think it may be okay to use here, but have you checked the log output doesn't include duplicated message text?

bdach: This is how it looks in the log:

```
[runtime] 2020-12-31 11:17:10 [error]: Error running task: Operation is not valid due to the current state of the object.
[runtime] 2020-12-31 11:17:10 [error]: System.InvalidOperationException: Operation is not valid due to the current state of the object.
[runtime] 2020-12-31 11:17:10 [error]: at osu.Game.Online.Multiplayer.MultiplayerClient.StartMatch() in /home/dachb/Dokumenty/opensource/osu/osu.Game/Online/Multiplayer/MultiplayerClient.cs:line 178
```

and on screen:

osu_2020-12-31_12-17-40

I guess the message is duplicated but I mostly included it, because without it the message looks barren/jank (just an "Error running task"). I guess that can be changed by adjusting the error messages but not sure if worth?

bdach: If I were to suggest an alternative it would be to include a string param to `CatchUnobservedExceptions()` and adjust usages to provide context as to which operation is being bailed out of.

peppy: I think it's fine, as long a the text makes it easy to know the source is a call to this method.

peppy (approved):

PR #11386, opened 2021-01-01
Fix selected mods properties not copied to `ModSelectOverlay`'s mods

frenzibyte:
Spotted while I was testing something else with multiple clients, `ModSelectOverlay` has its own isolated mod instances for each button, when `SelectedMods` bindable is changed externally, `ModSelectOverlay` will select the types of each mod in the new selected list, without copying each's properties into the button's mods.

This will cause a nasty mod desynchronization, which is one of the reasons (if not the single reason) why some players finish early than others, or vice versa.

This fixes the issue by copying the properties from `SelectedMods` to their associated buttons on the selection logic (which has been renamed to `UpdateSelectedMods` with xmldocs).


osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -131,22 +131,31 @@ public virtual string SettingDescription
         /// </summary>
         public virtual Mod CreateCopy()
         {
-            var copy = (Mod)Activator.CreateInstance(GetType());
+            var result = (Mod)Activator.CreateInstance(GetType(), true);
```
bdach, line 134: Is the `true` argument required here? It seemed to work without before - I'm not sure we have mods with non-public parameterless ctors...

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -131,22 +131,31 @@ public virtual string SettingDescription
...
+            result.CopyFrom(this);
+            return result;
+        }
+
+        /// <summary>
+        /// Copies properties of given mod into here, through changing value.
```
bdach, line 140:
```suggestion
        /// Copies mod setting values from <paramref name="them"/> into this instance.
```

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -131,22 +131,31 @@ public virtual string SettingDescription
...
             foreach (var (_, prop) in this.GetSettingsSourceProperties())
             {
-                var origBindable = prop.GetValue(this);
-                var copyBindable = prop.GetValue(copy);
+                var ourBindable = prop.GetValue(this);
+                var theirBindable = prop.GetValue(them);
 
                 // The bindables themselves are readonly, so the value must be transferred through the Bindable<T>.Value property.
-                var valueProperty = origBindable.GetType().GetProperty(nameof(Bindable<object>.Value), BindingFlags.Public | BindingFlags.Instance);
+                var valueProperty = theirBindable.GetType().GetProperty(nameof(Bindable<object>.Value), BindingFlags.Public | BindingFlags.Instance);
```
bdach, line 154: Not sure again why this is a change. It probably doesn't make any functional difference, as any desynchronisation should result in catching fire, but from a readability standpoint alone `ourBindable` seems like a better choice for getting properties to copy into, as it is *this* instance that is being copied into. The type check at the start of the method should ensure there is nothing funny going on wrt types already.

osu.Game/Overlays/Mods/ModSection.cs
```diff
@@ -127,18 +127,28 @@ public void DeselectTypes(IEnumerable<Type> modTypes, bool immediate = false)
         }
 
         /// <summary>
-        /// Select one or more mods in this section and deselects all other ones.
+        /// Updates all buttons with the given list of selected mods.
         /// </summary>
-        /// <param name="modTypes">The types of <see cref="Mod"/>s which should be selected.</param>
-        public void SelectTypes(IEnumerable<Type> modTypes)
+        /// <param name="newSelectedMods">The types of <see cref="Mod"/>s to select.</param>
```
bdach, line 132: No longer just the types.

osu.Game/Overlays/Mods/ModSection.cs
```diff
@@ -127,18 +127,28 @@ public void DeselectTypes(IEnumerable<Type> modTypes, bool immediate = false)
...
+                    if (index < 0)
+                        continue;
+
+                    var buttonMod = button.Mods[index];
+                    buttonMod.CopyFrom(mod);
+                    button.SelectAt(index);
+                    break;
+                }
+
+                if (index < 0)
```
bdach, line 151: Not a fan of how this loop is written. I'd probably prefer something along the lines of

```csharp
        public void UpdateSelectedMods(IReadOnlyList<Mod> newSelectedMods)
        {
            foreach (var button in buttons)
                updateButtonMods(button, newSelectedMods);
        }

        private void updateButtonMods(ModButton button, IReadOnlyList<Mod> newSelectedMods)
        {
            foreach (var mod in newSelectedMods)
            {
                int index = Array.FindIndex(button.Mods, m1 => mod.GetType() == m1.GetType());

                if (index >= 0)
                    continue;

                var buttonMod = button.Mods[index];
                buttonMod.CopyFrom(mod);
                button.SelectAt(index);
                return;
            }

            button.Deselect();
        }
```

as I think it is clearer to read (no more confusion which loop is being continued/broken out of, early-return helps get rid of the weird outer index).

bdach (changes requested):

osu.Game/Overlays/Mods/ModSection.cs
frenzibyte: sure but I believe that `index >= 0` check in the bottom method is flipped, yeah?

osu.Game/Rulesets/Mods/Mod.cs
frenzibyte: just a drive-by one, can revert.

osu.Game/Rulesets/Mods/Mod.cs
frenzibyte: that was also a drive-by change, was making a `MultiMod` private constructor as it had a public *parameterful* one, can revert

osu.Game/Overlays/Mods/ModSection.cs
frenzibyte: sounds good?
```suggestion
        /// <param name="newSelectedMods">The new list of selected mods to set from.</param>
```

osu.Game/Overlays/Mods/ModSection.cs
bdach: that it is, yeah. i messed with that code a few times and missed that.

osu.Game/Overlays/Mods/ModSection.cs
bdach: "set from" sounds weird, would just keep it simple and keep "select".

bdach:
Is this still required/applicable after #11413?

frenzibyte:
Nope, the mods aren't copied to the overlay in master, the linked PR doesn't change anything in here.

bdach:
Please look into resolving the merge conflicts in that case.

frenzibyte:
> Please look into resolving the merge conflicts in that case.

Right, done.

bdach:
hold off with this one for a bit please, want to see if i can figure out and add test coverage for the failure case this is supposed to fix.

bdach:
never mind, there is already a test. weird, could have sworn there wasn't one before.

bdach (approved):
seems fine

peppy (approved):

PR #11387, opened 2021-01-01
Add implementation of `OverridableBindable`

frenzibyte:
>

- [ ] Depends on https://github.com/ppy/osu-framework/pull/4118 (`BindableNumberWithCurrent<T>` part)
- [ ] Depends on framework bump.

Definitely unsure of the name used here (`OverridableBindable`), that's the best I could think of.

### DIFF
--- a/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
+++ b/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
@@ -62,16 +62,17 @@ namespace osu.Game.Tests.NonVisual
 
             Assert.That(setting.CustomValue.Value, Is.EqualTo(8));
 
-            // finally, disable custom value to make sure we revert to base and disable changes.
+            // disable custom value to make sure we revert to base and disable changes.
             setting.HasCustomValue.Value = false;
 
             Assert.That(setting.FinalValue.Value, Is.EqualTo(11));
             Assert.That(setting.FinalValue.Disabled, Is.True);
 
-            // re-enable once more for good measure.
+            // finally, re-enable and ensure custom value inherited base.
             setting.HasCustomValue.Value = true;
 
-            Assert.That(setting.FinalValue.Value, Is.EqualTo(8));
+            Assert.That(setting.FinalValue.Value, Is.EqualTo(setting.CustomValue.Value));
+            Assert.That(setting.FinalValue.Value, Is.EqualTo(11));
             Assert.That(setting.FinalValue.Disabled, Is.False);
         }
 
--- a/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
+++ b/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
@@ -7,12 +7,12 @@ using osu.Game.Rulesets.Mods;
 namespace osu.Game.Tests.NonVisual
 {
     [TestFixture]
-    public class CustomDifficultySettingTest
+    public class OverridableBindableTest
     {
         [Test]
         public void TestBaseValueChanges()
         {
-            var setting = new CustomDifficultySetting<double>(
+            var setting = new OverridableBindable<double>(
                 minValue: 1,
                 maxValue: 11,
                 defaultValue: 5);
@@ -29,7 +29,7 @@ namespace osu.Game.Tests.NonVisual
         [Test]
         public void TestCustomValueSet()
         {
-            var setting = new CustomDifficultySetting<float>(
+            var setting = new OverridableBindable<float>(
                 minValue: 5,
                 maxValue: 15,
                 defaultValue: 6);
--- a/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
+++ b/osu.Game.Tests/NonVisual/OverridableBindableTest.cs
@@ -3,6 +3,7 @@
 
 using NUnit.Framework;
 using osu.Framework.Bindables;
+using osu.Game.IO.Serialization;
 using osu.Game.Rulesets.Mods;
 
 namespace osu.Game.Tests.NonVisual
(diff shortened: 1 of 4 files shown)


osu.Game.Tests/NonVisual/OverridableBindableTest.cs
```diff
@@ -0,0 +1,153 @@
...
+            Assert.That(setting.FinalValue.Value, Is.EqualTo(2));
+
+            parseAndEnsureExpected(7f, 7f);
+            parseAndEnsureExpected(new Bindable<float>(6f), 6f);
+            parseAndEnsureExpected(new OverridableBindable<float>(5f) { HasCustomValue = { Value = true } }, 5f);
+
+            parseAndEnsureExpected(14f, 8f);
+            parseAndEnsureExpected(-5f, 2f);
+
+            void parseAndEnsureExpected(object input, object expected)
```
bdach, line 96:
```suggestion
            void parseAndEnsureExpected(object input, float expected)
```

use `object` as sparsely as humanly possible

osu.Game.Tests/NonVisual/OverridableBindableTest.cs
```diff
@@ -0,0 +1,153 @@
...
+
+            Assert.That(setting.Serialize().Deserialize<float>(), Is.EqualTo(5f));
+
+            var deserializedSetting = setting.Serialize().Deserialize<OverridableBindable<float>>();
+            Assert.That(deserializedSetting.HasCustomValue.Value, Is.True);
+            Assert.That(deserializedSetting.CustomValue.Value, Is.EqualTo(5f));
+
+            setting.HasCustomValue.Value = false;
+
+            Assert.That(setting.Serialize().Deserialize<float?>(), Is.Null);
```
bdach, line 150: should have equivalent testing to the `HasCustomValue = true` case (deserialising to an `OverridableBindable<float>` should yield a non-null one with `HasCustomValue = true`)

osu.Game/Rulesets/Mods/OverridableBindable.cs
```diff
@@ -0,0 +1,165 @@
...
+                    return;
+
+                case OverridableBindable<T> setting:
+                    if (setting.HasCustomValue.Value)
+                    {
+                        CustomValue.Value = setting.CustomValue.Value;
+                        HasCustomValue.Value = true;
+                    }
+                    else
+                        HasCustomValue.Value = false;
```
bdach, line 100: can be simplified to
```suggestion
					HasCustomValue.Value = setting.CustomValue.Value;
                    if (setting.HasCustomValue.Value)
                        CustomValue.Value = setting.CustomValue.Value;
```

bdach (changes requested):

osu.Game/Rulesets/Mods/OverridableBindable.cs
```diff
@@ -0,0 +1,165 @@
...
+        public Bindable<T> FinalValue => finalValue;
+
+        /// <summary>
+        /// Whether the final bindable should be bound to <see cref="CustomValue"/>, otherwise will be immutably bound to <see cref="BaseValue"/>.
+        /// </summary>
+        public Bindable<bool> HasCustomValue { get; } = new BindableBool();
+
+        private readonly BindableNumber<T> immutableBaseValue = new BindableNumber<T>();
+        private readonly BindableNumber<T> customValue = new BindableNumber<T>();
+        private readonly BindableNumber<T> finalValue = new BindableNumber<T>();
```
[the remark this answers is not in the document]
frenzibyte: good catch

osu.Game.Tests/NonVisual/OverridableBindableTest.cs
frenzibyte: that is doing so right above this code

osu.Game/Rulesets/Mods/OverridableBindable.cs
frenzibyte: intentionally did it this way such that `FinalValue` binds to `CustomValue` (`HasCustomValue = true`) after the value is parsed.

bdach: then reverse the operation order? with a comment, obviously. no idea why there was no comment in the first place if this was relying on weird operation ordering.

osu.Game.Tests/NonVisual/OverridableBindableTest.cs
bdach: i'm not seeing a

```csharp
            var deserializedSetting = setting.Serialize().Deserialize<OverridableBindable<float>>();
            Assert.That(deserializedSetting.HasCustomValue.Value, Is.False);
```

anywhere? i was explicitly saying that the deserialise/parse should also cover the no-custom-value case.

osu.Game/Rulesets/Mods/OverridableBindable.cs
frenzibyte: thought it was quite clear, not say "I have custom value" unless I actually do have a valid one, so you mean like this:
```suggestion
                    HasCustomValue.Value = false;

                    if (setting.HasCustomValue.Value)
                    {
                        CustomValue.Value = setting.CustomValue.Value;
                        HasCustomValue.Value = true;
					}
```

bdach: why is the assignment to `HasCustomValue` under a branch at all if you're just copying the value, is what i'm getting at

osu.Game.Tests/NonVisual/OverridableBindableTest.cs
frenzibyte: oh you mean it like that, sorry didn't fully understand your preivous comment

bdach: i did flip true to false i suppose

osu.Game/Rulesets/Mods/OverridableBindable.cs
frenzibyte: Resolved in https://github.com/ppy/osu/pull/11387/commits/831e11dbf2a608bbbc5f061b846cc4a5caedcdf7 (should've been in a separate commit but /shrug)

smoogipoo:

PR #11388, opened 2021-01-01
Add simple UI control for `OverridableBindable`

frenzibyte:
>

- [ ] Depends on #11387 
- [ ] Depends on ppy/osu-framework#4118
- [ ] Depends on framework bump.


frenzibyte:
Failure above is known fwiw, which ppy/osu-framework#4118 is fixing.

peppy:
Is this changing this original UX? When checking the "i want to customise this value" what value is it inheriting?

Would be nice for the tests to cover these scenarios to state intent.

frenzibyte:
Yeah actually now when enabling the checkbox, it'll use whatever `CustomValue` has, instead of making `CustomValue` inherit base value first, will fix in the underlying PR.

smoogipoo:

PR #11389, opened 2021-01-01
Use `IParseable` interface for mod parsing/copying logic

frenzibyte:
Required for the proposed `OverridableBindable` class to be allowed for use as mod property.

Not sure about the `Mod.CreateCopy()` changes part though, going through `IParseable` seems the best way currently, no more reflection hacks with that, but feedbacks would be appreciated.


peppy:
This change has already been applied via recent fixes applied to the same code (albeit in a slightly different place now).

PR #11390, opened 2021-01-01
Replace user-changed value logic with `OverridableBindable` for "difficulty adjust" mod

### LINKED ISSUE #11305: Difficulty adjust not working in multiplayer

frenzibyte:
>

- [ ] Depends on #11388 
- [x] Depends on #11389 
- Closes #11305 
- Alternative to and closes #11324


smoogipoo:

PR #11399, opened 2021-01-01
Update dependencies

bdach:
Bundled PRs from dependabot for easier testing/merging.

- [x] Needs testing on Apple devices (though I rolled back the signalr bumps for iOS).
* Includes #11393, #11394, #11395, #11397.
* Does **NOT** include #11392 - [see reason](https://github.com/ppy/osu/pull/11392#issuecomment-753373677).
* Does **NOT** include #11396 - as I'm pretty sure this was pinned on purpose (comparing with #11038).

Have checked linux, windows and android. All boot and can enter multiplayer.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -16,7 +16,7 @@
     <EmbeddedResource Include="Resources\**\*.*" />
   </ItemGroup>
   <ItemGroup Label="Code Analysis">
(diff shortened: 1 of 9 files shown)


frenzibyte:
macOS seems fine as well, boot and multi.

peppy:
Nothing in here looks like it'll be an issue on iOS.

peppy (approved):

PR #11405, opened 2021-01-02
Fix the beatmap carousel not returning to centre correctly after resizing window

### LINKED ISSUE #11403: Carousel doesn't update its middle position when resizing the window
**Describe the bug:** As title, the carousel doesn't update its middle position when resizing the window.
Can be fixed by re-entering the carousel (as shown in the video).

https://user-images.githubusercontent.com/43073074/103456138-aab9cf80-4cf3-11eb-88f4-5412c0684f71.mp4

**osu!lazer version:** 2020.1229.0

peppy:
Closes #11403.

### DIFF
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -428,7 +428,10 @@ namespace osu.Game.Screens.Select
 
         private void updateSelectedBeatmap(BeatmapInfo beatmap)
         {
-            if (beatmap == null || beatmap.Equals(beatmapNoDebounce))
+            if (beatmap == null && beatmapNoDebounce == null)
+                return;
+
+            if (beatmap?.Equals(beatmapNoDebounce) == true)
                 return;
 
             beatmapNoDebounce = beatmap;
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -428,7 +428,7 @@ namespace osu.Game.Screens.Select
 
         private void updateSelectedBeatmap(BeatmapInfo beatmap)
         {
-            if (beatmap?.Equals(beatmapNoDebounce) == true)
+            if (beatmap == null || beatmap.Equals(beatmapNoDebounce))
                 return;
 
             beatmapNoDebounce = beatmap;
@@ -438,7 +438,7 @@ namespace osu.Game.Screens.Select
 
         private void updateSelectedRuleset(RulesetInfo ruleset)
         {
-            if (ruleset?.Equals(rulesetNoDebounce) == true)
+            if (ruleset == null || ruleset.Equals(rulesetNoDebounce))
                 return;
 
             rulesetNoDebounce = ruleset;
--- a/osu.Game/Screens/Select/SongSelect.cs
+++ b/osu.Game/Screens/Select/SongSelect.cs
@@ -435,13 +435,15 @@ namespace osu.Game.Screens.Select
                 return;
 
             beatmapNoDebounce = beatmap;
-
             performUpdateSelected();
         }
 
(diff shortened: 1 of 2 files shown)


bdach:
As stated in https://github.com/ppy/osu-framework/pull/4124#issuecomment-753473986, this depends on the framework-side bump of the same package now.

bdach:
(Disregard the above mentioned pull, I messed up my tabs.)

Looks like a genuine test failure which is slightly scary...

bdach:
Did a bit of debugging and this fell out... Picture should explain what's wrong here.

2021-01-02-221126_1116x327_scrot

I guess we want something like the following instead?

2021-01-02-221558_1165x447_scrot

That seems to fix the test failure, for what it's worth.

peppy:
I've added the checks you proposed, but also changed the underlying carousel logic to avoid this happening in the first place. The issue is that the carousel would fire selection changed events every `updateYPositions` call when the selection is null.

Before merging this, the performance needs to be checked as well. I did briefly test this myself and didn't notice an issue, but consider that `updateYPositions` is now being called every update loop when a transform is being applied (ie. hiding the toolbar).

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -428,7 +428,7 @@ private void workingBeatmapChanged(ValueChangedEvent<WorkingBeatmap> e)
 
         private void updateSelectedBeatmap(BeatmapInfo beatmap)
         {
-            if (beatmap?.Equals(beatmapNoDebounce) == true)
+            if (beatmap == null || beatmap.Equals(beatmapNoDebounce))
```
bdach, line 431: This is not the correct check for the problem I pointed out - it (correctly) fails other tests. Now if `beatmap` is null but `beatmapNoDebounce` wasn't, this early-returns instead of transferring, therefore breaking deselection. There was a reason I had it written as funny as it was.

bdach (changes requested):

osu.Game/Screens/Select/SongSelect.cs
peppy: Yeah, I was thinking we wanted to keep the debounce from being null but didn't realise it isn't actually a bindable and null is valid in this context.

peppy:
I've fixed the remaining issues with this. I think it's about the best we can do performance-wise without adjusting panel Y positions to *not* include `visibleHalfHeight` (this actually shouldn't be too difficulty, but definitely for another PR and only when necessary). I've tested with my normal beatmap collection and the overhead doesn't look too high (and is limited to toolbar show/hide and window resizes for now).

osu.Game/Screens/Select/SongSelect.cs
```diff
@@ -438,7 +441,7 @@ private void updateSelectedBeatmap(BeatmapInfo beatmap)
 
         private void updateSelectedRuleset(RulesetInfo ruleset)
         {
-            if (ruleset?.Equals(rulesetNoDebounce) == true)
+            if (ruleset == null || ruleset.Equals(rulesetNoDebounce))
```
bdach, line 444: Not sure why this doesn't mirror the `updateSelectedBeatmap` check. Seems like `rulesetNoDebounce` being `null` is also valid (`performUpdateSelected` is at least one place that seems to permit it - it is assigned to a `ruleset` local and then logged).

Also not sure why this doesn't fail any tests; it's probably because `rulesetNoDebounce` seems to have no functional usages besides that one log statement. Still would probably unify treatment with beatmap to reduce confusion/avoid potential future bugs.

peppy: I've made this change 👍🏻

bdach:
Seems to work correctly and I'm also only seeing minimal added overhead in mentioned cases, so looks good to go :+1:

bdach (approved):

PR #11407, opened 2021-01-02
Refactor difficulty calculator framework to be more flexible

Syriiin:
This PR makes 3 changes to the difficulty calculator framework in `osu.Game.Rulesets.Difficulty` to make it more flexible.

# Changes

1. Encapsulate the strain logic into its own abstract `StrainSkill` class
2. Add a list of visual mods to the base `Skill` class
3. Replace `LimitedCapacityStack` with a regular `Stack`

# Reason

### StrainSkill refactor
There is no reason to force rulesets to conform to a strain based difficulty calculator.
This change encapsulates all the strain logic into the `StrainSkill` abstract class so we are free to extend the base `Skill` class for fully custom skill implementations.

### Adding visual mods
Visual mods are not reflected in hitobject or clock rate transformations, and therefore are impossible to account for in difficulty calculation currently.
This change simply adds a list of visual mods to the base skill class so we can choose to have them affect difficulty if we wish.

### Removing LimitedCapacityStack
There is no reason we should be limiting skills to knowing only the previous 2 objects before the current one. This originally existed as an angle implementation detail of the original pp+ codebase which made its way here, but didn't get used in the same way.
If we want to get only the previous 2 objects we can simply use LINQ. ie. `Previous.Take(2)`

# Note

None of these changes _should_ make any difference to current difficulty values, however, for some reason the `StrainSkill` refactor has changed the taiko calculation output very slightly (comparable to a rounding error).
I have simply updated the test to reflect the new value as I haven't seen anything to make me think its a deeper issue.


bdach:
This is a pretty massive diff, all things considered. Not sure these three things should be clumped up as they are. "More flexible" is also a pretty nebulous goal overall - is there an overall larger direction you want to take this in further, or is this just a "code quality" submission?

Syriiin:
> This is a pretty massive diff, all things considered. Not sure these three things should be clumped up as they are. "More flexible" is also a pretty nebulous goal overall - is there an overall larger direction you want to take this in further, or is this just a "code quality" submission?

The diff is pretty deceiving I reckon.
More than half of it is from the second two commits which are extremely simple, but just happen to have a lot of line changes due to their nature.
The only bad one is the 159 additions and 130 deletions from the first commit which are largely just from renaming the file.

Yea "more flexible" maybe isn't the most helpful description. 
I've just noticed a lot of arbitrary limitations when working with the diff calc framework in custom rulesets and I wanted to get rid of them since there is no reason for them to exist.
The specific limitations being:
- visual mods can't affect difficulty
- skills can't look back past 2 previous objects
- difficulty calculators are forced to conform to a strain based approach)

bdach:
> More than half of it is from the second two commits which are extremely simple, but just happen to have a lot of line changes due to their nature.

That may be the case, but now I've got ~500 lines to cross-check, and I'll probably have to read the diff commit by commit too as the full diff is way too noisy. I wouldn't split just yet until we've ascertain whether these changes are actually something we might want (and whether it should be in the current form), which I'm not fully convinced yet.

> visual mods can't affect difficulty

I... don't think that's the case? Difficulty calculators already define `DifficultyAdjustmentMods`, so there seems to already be a mechanism in place for this, which sounds to be even more general than what's proposed here as it has no arbitrary "visual mod" constraint.

> skills can't look back past 2 previous objects

This is fair, but I'm not sure they should be getting the full history, either. There are some ridiculous maps out there with thousands of hitobjects and I'm not sure that should all necessarily reside in memory if it doesn't need to. New custom rulesets might also stretch this further.

> difficulty calculators are forced to conform to a strain based approach

This is probably the one thing I'd be most willing to accept (taiko hacks past this currently).

If this is intended for custom rulesets only, then I'm not sure whether it's not too early for this change. There's no infrastructure to do anything too useful with custom ruleset difficulty calculation just yet. An added factor is that while we're maintaining diffcalc in both stable and lazer, someone will have to go and port these changes so that the two implementations can be easily compared for possible discrepancies.

osu.Game.Rulesets.Mania/Difficulty/Skills/Strain.cs
```diff
@@ -71,8 +75,8 @@ protected override double StrainValueOf(DifficultyHitObject current)
         }
 
         protected override double GetPeakStrain(double offset)
-            => applyDecay(individualStrain, offset - Previous[0].BaseObject.StartTime, individual_decay_base)
-               + applyDecay(overallStrain, offset - Previous[0].BaseObject.StartTime, overall_decay_base);
+            => applyDecay(individualStrain, offset - Previous.First().BaseObject.StartTime, individual_decay_base)
```
peppy, line 78: pretty sure all these `.First()` calls are going to allocate enumerators, which we want to avoid. i think you're looking for `.Peek()`

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
```diff
@@ -1,120 +1,46 @@
...
-
-        /// <summary>
-        /// The weight by which each strain value decays.
-        /// </summary>
-        protected virtual double DecayWeight => 0.9;
-
-        /// <summary>
-        /// <see cref="DifficultyHitObject"/>s that were processed previously. They can affect the strain values of the following objects.
-        /// </summary>
-        protected readonly LimitedCapacityStack<DifficultyHitObject> Previous = new LimitedCapacityStack<DifficultyHitObject>(2); // Contained objects not used yet
```
peppy, line 42: not sure if we want to remove this as opposed to making the limited size configurable. probably need input from @smoogipoo on the reasoning / direction here (it may be done to limit retention of `DifficultyHitObjects`).

peppy (changes requested):
Important to note that we consider difficulty calculation one of the high performance paths in lazer, which needs to run lean for both client and server side processing. The end goal would be to get allocations down to as close to zero as possible, and while we're not there yet, I'd want to avoid regressing further in the other direction.

Syriiin:
> That may be the case, but now I've got ~500 lines to cross-check, and I'll probably have to read the diff commit by commit too as the full diff is way too noisy.

Reading the diff commit by commit is definitely going to be the way to go.
For the `StrainSkill` refactor, you pretty much just need to compare the `Process()` method, since that's where all the logic ties into.

> I... don't think that's the case? Difficulty calculators already define DifficultyAdjustmentMods, so there seems to already be a mechanism in place for this, which sounds to be even more general than what's proposed here as it has no arbitrary "visual mod" constraint.

Afaik, `DifficultyAdjustmentMods` is meant for informing the calculator of which mods will have difficulty implications _based on our implementation_, not which mods _can_ have effects. There is simply no good way (that I have been able to come up with) to access the visual mods from within a `Skill`

> This is fair, but I'm not sure they should be getting the full history, either. There are some ridiculous maps out there with thousands of hitobjects and I'm not sure that should all necessarily reside in memory if it doesn't need to. New custom rulesets might also stretch this further.

I agree to an extend, however it's not something I think we should have a say in. Who are we to tell custom ruleset authors that they can't have the difficulty of an object be affected by all the previous ones? We surely can't come up with a use case right now, but I don't think it's our place to force that limitation without a good reason. The memory usage is a valid point in theory, but I don't believe this would actually be an issue in practice without it causing a myriad of issues beyond just the diff calc.

> This is probably the one thing I'd be most willing to accept (taiko hacks past this currently).

Yea this is the one part that actually can be hacked past (and I have in my use case), but its clearly not ideal to be promoting hacky solutions when they can be supported instead.

osu.Game.Rulesets.Mania/Difficulty/Skills/Strain.cs
Syriiin: Ah good point, will make the change in the morning :+1:

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
Syriiin: I'm fairly certain this only exists in the codebase as a remnant of the original pp+ implementation for angle calculations.
Instead of having a configurable fixed size, what do we think about a method that allows the `Skill` to define before processing each object, which items are no longer needed?
Example use case: a reading skill that wants to have access to as many objects as are on the screen at a given time.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
```diff
@@ -0,0 +1,133 @@
...
+        private readonly List<double> strainPeaks = new List<double>();
+
+        /// <summary>
+        /// Process a <see cref="DifficultyHitObject"/> and update current strain values accordingly.
+        /// </summary>
+        public sealed override void Process(DifficultyHitObject current)
+        {
+            // The first object doesn't generate a strain, so we begin with an incremented section end
+            if (Previous.Count == 0)
+                currentSectionEnd = Math.Ceiling(current.StartTime / SectionLength) * SectionLength;
```
bdach, line 67: I am *relatively* certain that moving this is the cause of the slight taiko diffcalc shift. If you look at all rulesets' implementations of `CreateDifficultyHitObjects()`, they don't generate a difficulty hitobject for the first object of the map, because doing so usually requires at least one preceding object to exist, so they usually start from the second or even third one. That *may* be fine to do, but it is a behavioural change.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
```diff
@@ -0,0 +1,133 @@
...
+
+        /// <summary>
+        /// Returns the calculated difficulty value representing all processed <see cref="DifficultyHitObject"/>s.
+        /// </summary>
+        public sealed override double DifficultyValue()
+        {
+            double difficulty = 0;
+            double weight = 1;
+
+            var finalStrainPeaks = strainPeaks.Select(p => p).ToList(); // Copy list so we don't create side effects
```
bdach, line 118: Can skip the `.Select(p => p)`. It doesn't do anything.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
```diff
@@ -0,0 +1,133 @@
...
+            finalStrainPeaks.Add(currentSectionPeak); // Add final peak since it never gets saved
```
bdach, line 119: This is sort of dodgy. Ideally `Process()` would be doing this instead of having this cleanup in another method that looks to be unrelated.

This is the consequence of the choice of having `Process()` take in individual HOs instead of the entire list. Passing the whole list down would avoid this problem, I think.

osu.Game.Rulesets.Taiko/Difficulty/Skills/Rhythm.cs
```diff
@@ -13,7 +13,7 @@ namespace osu.Game.Rulesets.Taiko.Difficulty.Skills
     /// <summary>
     /// Calculates the rhythm coefficient of taiko difficulty.
     /// </summary>
-    public class Rhythm : Skill
+    public class Rhythm : StrainSkill
```
bdach, line 18: This skill is special (it specifies `StrainDecayBase => 0`, as the goal is to have this one be based on object index, not time between objects). I'd ideally want that zero specification gone from this to avoid having to explain it, so in that regard I don't think this change goes far enough.

osu.Game/Rulesets/Difficulty/DifficultyCalculator.cs
```diff
@@ -179,7 +180,8 @@ static IEnumerable<Mod> createDifficultyAdjustmentModCombinations(ReadOnlyMemory
         /// Creates the <see cref="Skill"/>s to calculate the difficulty of an <see cref="IBeatmap"/>.
         /// </summary>
         /// <param name="beatmap">The <see cref="IBeatmap"/> whose difficulty will be calculated.</param>
+        /// <param name="visualMods">Visual mods to calculate difficulty with.</param>
         /// <returns>The <see cref="Skill"/>s.</returns>
-        protected abstract Skill[] CreateSkills(IBeatmap beatmap);
+        protected abstract Skill[] CreateSkills(IBeatmap beatmap, List<ModWithVisibilityAdjustment> visualMods);
```
bdach, line 185: As I said already, I find constraining this to just "visual" mods sort of arbitrary. I think if we want skills to be aware of active mods, they should just receive *all* active mods - that will surely make even more things possible than passing just "visual" mods?

bdach (changes requested):

osu.Game/Rulesets/Difficulty/DifficultyCalculator.cs
peppy: Can agree with this. If we're going to make things more flexible there's no reason to have an arbitrary limit like this.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
Syriiin: Ah that could definitely be it! I don't think that is enough of a reason to require strain calculations to happen at a `DifficultyCalculator` level though. If we want to make certain that this tiny output change doesn't occur we can find another way around it.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
Syriiin: This exists to ensure that `DifficultyValue()` is kept as a pure function. It merely interprets the current state of the `Skill` and gives you a result. If we were to pass the entire list of hitobjects to the `Skill` in a single process call, then in that case why even let it have state, static would work too? why have a `DifficultyValue()` method? we could just return the difficulty value as the result of the `ProcessAll()` method.
This is certainly a valid approach, but I believe it's not the correct one. Correct me if I'm wrong but, I believe we made the conscious decision to have `Skill`s follow a stateful approach rather than a static one for this very reason.

Consider if we wanted to get the difficulty of a `Skill` after processing each object for some reason (perhaps constructing a difficulty graph).
If we were to add this incomplete strain to `strainPeaks`, the following `Process()` calls would result in incorrect values.
I think we should be able to call `Process()` `Process()` `DifficultyValue()` and end up with the same result as `Process()` `DifficultyValue()` `Process()` `DifficultyValue()`.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
Syriiin: As mentioned in the code comment, this is to ensure that `DifficultyValue()` doesn't produce side effects.
More: https://github.com/ppy/osu/pull/11407#discussion_r550949827

osu.Game/Rulesets/Difficulty/DifficultyCalculator.cs
Syriiin: I have no issue with simply giving skills access to all mods, however I don't see any need to and think it might encourage misuse of them. There is a fundamental difference with visual mods in that they are not reflected in the hitobject's values. Regular mods make physical transformations to the hitobjects which are used for difficulty calculation.

osu.Game.Rulesets.Mania/Difficulty/Skills/Strain.cs
Syriiin: Fixed in 7e0cdb250c9008140ba363760ac6e031718e8d74

Syriiin: Fixed in 7e0cdb250c9008140ba363760ac6e031718e8d74

osu.Game/Rulesets/Difficulty/DifficultyCalculator.cs
peppy: If we're going to go in that direction, I'd have to add the argument that visual mods don't actually affect difficulty and probably shouldn't be considered, but that's a whole different discussion to be had.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
Syriiin: If performance is our chief concern here, we can perhaps achieve the best of both worlds by having the `DifficultyCalculator` track the previous hitobject stack and simply pass the reference to it in the Skill ctors?

osu.Game/Rulesets/Difficulty/DifficultyCalculator.cs
Syriiin: That is true in what we have always considered for the official gamemodes, but is it right for us to impose that limitation on custom ruleset developers?

peppy: If we're looking for any kind of consistency.

At the point mods are contributing to difficulty a lot of considerations need to change. I'm not in the position to make a call like this, but I'd appreciate it not being hard-coded into the difficulty algorithm in the way you've done. It's either all or no mods.

Syriiin: Changed in 95dcc73c742d4b02db88684656c57bf3916bd1f3.
Now passing all mods :+1:

osu.Game.Rulesets.Taiko/Difficulty/Skills/Rhythm.cs
Syriiin: I agree, we should ideally not have that `StrainDecayBase => 0` in the skill (and in fact the whole skill is pretty hacky to just fit into the old `Skill` requirements), however I don't think this is the place to make that change.
I think the most elegant solution would be to split `StrainSkill` again perhaps into `StrainSkill` and `TimeStrainSkill` or something since the current strain implementation is centered around consistent time-based decay, where as the taiko `Rhythm` skill is fundamentally different in that it has different decay requirements and forcefully resets the strain to 0 in places.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
Syriiin: Ah nevermind! I misunderstood what you mean. 😅 
Fixed in 645b0c84d8d1e07b0f9b98c74086ae7144ad9f8e

osu.Game.Rulesets.Taiko/Difficulty/Skills/Rhythm.cs
bdach: If the strain skill split was its own separate pull then it would be the best place to make this change, but clumped up like this it's definitely not a good idea to pile more on.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
bdach: I misunderstood what this was doing, partially because of the xmldoc that states the value represents "all processed difficulty hitobjects". I thought it was a hack that works around the fact that `Process()` doesn't add the final section strain (because it doesn't actually know when it receives the final difficulty hit object). Maybe the xmldoc of this method should state "...up to now" or similar to clarify.

Syriiin: Yea that's understandable. I'll make it a bit more clear in the xmldoc.
FIxed in 79a6222a23e828d59a1cfeb34cdb9695d40758dc

osu.Game.Rulesets.Taiko/Difficulty/Skills/Rhythm.cs
Syriiin: Yea I agree. I'm happy to start work on that refactor as a new PR following this one.

osu.Game/Rulesets/Difficulty/Skills/StrainSkill.cs
Syriiin: In fact thinking about this again, doesn't this logically make more sense than before?
If a calculator doesn't create difficulty hitobjects for the first one or two hitobjects, then that carries the implication that those first objects "don't matter for difficulty".
With that implication, then it would make sense for the strain decay to begin at the first object that does contribute to difficulty, rather than letting it decay from the starting value 1 to some value slightly lower before real difficulty calculation begins (which is what happened before, with it starting at the first REAL hitobjects time).

My preference would be to actually enforce 1:1 hitobject to difficulty hitobject (excluding spinners for example) transformation and just handle the cases for missing previous objects within the difficulty code... but that's a discussion for another time.

bdach: Sure, I'm not expecting my original remark to be "resolved" in any capacity, per se. Was more of an observation if anything.

bdach:
The single point of contention that I have with this diff in its current state is still the removal of the limited stack - still think that a configurable but limited stack size would be preferable. Everything else I'm pretty okay with.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
Syriiin: I've addressed this in 13c12c246e6c7463626aec3d957349e32cdbe77b and 3a6aaf49f814afe1dd0f4703d662f99dcc849a3e.
13c12c246e6c7463626aec3d957349e32cdbe77b replaces LimitedCapacityStack<T> with DropoutStack<T>, which allows 3a6aaf49f814afe1dd0f4703d662f99dcc849a3e to implement a programmable history retention, rather than simply static size, or everything.
With these changes, we can have the stack retain only what is absolutely necessary as determined by the Skill, be that only the last n objects, objects within the last n milliseconds, or any other implementation skills could require.

Syriiin:
> The single point of contention that I have with this diff in its current state is still the removal of the limited stack - still think that a configurable but limited stack size would be preferable (could make it optional if somebody _really_ wants all objects). Everything else I'm pretty okay with.

@bdach see https://github.com/ppy/osu/pull/11407#discussion_r559937932 my update on this. What do you think?

bdach:
I don't know where the insistence on having the stack be "programmable" wrt object retention is coming from - seems like an unnecessary overcomplication to solve a problem that I'm not convinced people actually need a solution for, but if it works I guess it's tolerable. Need to read the code properly which I can't do right now.

Syriiin:
> I don't know where the insistence on having the stack be "programmable" wrt object retention is coming from - seems like an unnecessary overcomplication to solve a problem that I'm not convinced people actually need a solution for, but if it works I guess it's tolerable. Need to read the code properly which I can't do right now.

The reason I'm pushing for it is because I have a calculator implementation that requires a solution to it 😅 (pp+ lazer port)

bdach:
Sounds like something that should have been mentioned when you suggested that, to give some justification for the introduced complexity.

Syriiin:
> Sounds like something that should have been mentioned when you suggested that, to give some justification for the introduced complexity.

Yea probably. I was trying to frame the discussion in a sense of it being beneficial to _anyone_ implementing custom calculators, but I guess it would have been good to give a concrete example.

osu.Game/Rulesets/Difficulty/Utils/DropoutStack.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace osu.Game.Rulesets.Difficulty.Utils
+{
+    /// <summary>
+    /// An indexed stack where items are dropped from the bottom rather than popped from the top.
+    /// Indexing starts at the top of the <see cref="DropoutStack{T}"/>.
+    /// </summary>
```
bdach, line 12: Not overly sure about the chosen semantics. What this is is essentially a queue, but with the "indexing reversed" (tail of queue has smallest index? I think?). Might as well turn it into a deque at this point if we're going for flexibility? Or just plain list? Not sure.

Definitely can say that performance-wise it is definitely worse, but not sure by how much (see the `O(n)` implementation of `Drop()` - popping off the `LimitedStack` was `O(1)`).

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
```diff
@@ -1,120 +1,70 @@
...
+        /// Remove objects from <see cref="Previous"/> that are no longer needed for calculations from the current object onwards.
         /// </summary>
-        public void SaveCurrentPeak()
+        /// <param name="current">The <see cref="DifficultyHitObject"/> to be processed.</param>
+        protected virtual void RemoveExtraneousHistory(DifficultyHitObject current)
         {
-            if (Previous.Count > 0)
-                strainPeaks.Add(currentSectionPeak);
+            // Default implementation to not retain objects
+            Previous.Clear();
```
bdach, line 56: Would probably keep the previous implementation rather than just clearing like this. It makes `Previous` essentially useless unless overridden.

osu.Game/Rulesets/Difficulty/Utils/DropoutStack.cs
Syriiin: Good points.

I've addressed this with 4eb125374f7c0589cf5a6312e02e68785a265ab9 by replacing the `DropoutStack` with a `ReverseQueue`.
Is has essentially the same functionality, but with a more semantically appropriate name and lower level implementation allowing for O(1) enqueue*, dequeue, and access operations.
*The only exception is enqueue being O(n) when it hits its internal capacity and resizes to double the buffer size.
This O(n) case will only be a possibility for skills with truly unknown size requirements however because of the configurable initial capacity added in 3cf8635c61f3a84056e833975e34f6f277671f5e.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
Syriiin: That's the point though?
`Previous` _should be_ useless (in the sense that it doesn't store anything) if there is no requirement for it by the `Skill`.

Either way, with these changes the `Previous` collection is actually more efficient than the existing `LimitedCapacityStack` even for the same `Skill` implementations, since now we are only storing at most two objects in the collection at any given time, as opposed to before where it was a constant three, right?
Although that wasn't a flaw of the `LimitedCapacityStack` but instead that for whatever reason we left the capacity as three even though it was never used.

Syriiin: I get what you mean though. For the base skill we have this flow of adding an object to the collection, just for it to be inevitably cleared out when processing the next object.
But this is a case where I think the performance benefit wouldn't be worth the slightly unintuitive code to handle this case.
My initial thought would be to have an empty virtual method for `RemoveExtraneousHistory` and a conditional in `Process` and the ctor to check for the special case where `PreviousCollectionSoftCapacity` is 0 and simply skip initialising or enqueuing to `Previous`.

osu.Game/Rulesets/Difficulty/Utils/ReverseQueue.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Retrieves the item at an index in the <see cref="ReverseQueue{T}"/>.
+        /// </summary>
+        /// <param name="index">The index of the item to retrieve. The most recently enqueued item is at index 0.</param>
+        public T this[int index]
+        {
+            get
+            {
+                if (index > Count - 1)
+                    throw new ArgumentOutOfRangeException(nameof(index));
```
bdach, line 44: Should also throw when `< 0` if we're being accurate about it.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
bdach: Empty override is fine, but the size checks are weird. If anything I'd also leave the addition to the consumer, the same way removal is currently. Maybe make it one method, even, after `Calculate()`. `UpdateHistory(DifficultyHitObject newObject)` or something?

osu.Game/Rulesets/Difficulty/Utils/ReverseQueue.cs
```diff
@@ -0,0 +1,110 @@
...
+        /// <summary>
+        /// Enqueues an item to this <see cref="ReverseQueue{T}"/>.
+        /// </summary>
+        /// <param name="item">The item to enqueue.</param>
+        public void Enqueue(T item)
+        {
+            if (Count == capacity)
+            {
+                // Double the buffer size
+                var buffer = new T[capacity * 2];
```
bdach, line 60: This fails if `capacity == 0` (which it can be, in particular initially).

bdach (changes requested):

osu.Game/Rulesets/Difficulty/Utils/ReverseQueue.cs
Syriiin: Reason I didn't do that initially was because indexing the array would throw an error for that anyway, but you're right, it would probably make more sense to handle all the range exceptions with a single error here. Will fix later today.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
Syriiin: I'm not a super big fan of this approach, since the `Previous` stack is an idea handled by the `Skill` class, it makes sense for the `Skill` class to handle as much of the logic as it reasonably can.
I'm happy to make the change if you insist since it doesn't affect functionality, but I don't like that we are going to now have copy pasted code in all `Skill` implementations simply to accommodate for this edge case which is almost never relevant (since `Skill`s almost always need access to atleast the single prior object).

Syriiin: > Maybe make it one method, even, after `Calculate()`. `UpdateHistory(DifficultyHitObject newObject)` or something?

This also can't be a single method, since removal of irrelevant objects needs to be done as a preprocessing step for a given object.
After adding the current object, we can't know for sure what the next objects requirements will be.

Syriiin: The only way we can make it a single method is by passing the _previous_ object along to `Skill.Process` along with the current one, so that we can add the previous and remove the unnecessary all in one step.
But this now creates inconsistent `Skill` state between `Process` calls again which is something I want to avoid if at all possible.

osu.Game/Rulesets/Difficulty/Utils/ReverseQueue.cs
Syriiin: Fixed in 214e0096fe954ea36918a7eaffc93eb564537210

osu.Game/Rulesets/Difficulty/Utils/ReverseQueue.cs
Syriiin: Fixed in 214e0096fe954ea36918a7eaffc93eb564537210

bdach:
I'm not sure if you're following [discussion on discord](https://discord.com/channels/188630481301012481/188630652340404224/801747101212999720), but to reiterate it here for transparency just in case: the scope of changes in this pull is so large that porting this to stable would either take a considerable amount of time or not even be done at all; either way the timeline of this getting merged in its current clumped-up form might be on a timescale of months. For whatever that's worth, I pointed it out once already, back when this was opened.

osu.Game/Rulesets/Difficulty/Skills/Skill.cs
bdach: > since the `Previous` collection is an idea handled by the `Skill` class, it makes sense for the `Skill` class to handle as much of the logic as it reasonably can

It's not unreasonable to have someone i.e. want to ignore spinners entirely in a skill, or something along these lines. You could have a sane base implementation that adds the object to the queue and doesn't need to be touched in the 99% case.

Syriiin: Ah yea that's true. I'm happy to make it the same as the `RemoveExtraneousHistory` then. Both can be empty in the base `Skill`, with proper logic in the `StrainSkill override`.
Changed in e61ae58ff1e771f50cfe6e7e2b5de85fba607b2e

Syriiin:
> the scope of changes in this pull is so large that porting this to stable would either take a considerable amount of time or not even be done at all; either way the timeline of this getting merged in its current clumped-up form might be on a timescale of months.

Yea that's understandable, if a little annoying. I'm happy to split this up into its individual parts once we are happy with the changes.
I'm worried the same problem would still be the case though, since really the only part of this PR that I can see being quite troublesome to backport would be the `StrainSkill` refactor, and that can't really be broken down any further.
I don't imagine adding a mods param to the ctors or replacing the `LimitedHistoryStack` with the `ReverseQueue` would be incredibly difficult to backport due to their clean cut nature.

Either way, I'm happy to split the PR up once we are happy with the changes. 👌

bdach:
> Yea that's understandable, if a little annoying.

I mean I did signpost it as a potential issue in the very first comment here... Don't know how that could have been conveyed any sooner.

And I don't necessarily think the strain refactor is the only or even biggest problem. The `Mod` class doesn't actually exist in stable, for instance.

Syriiin:
> I mean I did signpost it as a potential issue in the very first comment here... Don't know how that could have been conveyed any sooner.

Oh I wasn't insinuating its someone's fault or anything. Just expressing that its an unfortunate reality that we need to deal with. 😅 

> And I don't necessarily think the strain refactor is the only or even biggest problem. The `Mod` class doesn't actually exist in stable, for instance.

Would that really be difficult to backport though? It's not like we are using special characteristics of the `Mod` class itself. Couldn't it reasonable just be implemented in practically the same way, but as the legacy bitmask/enum (or however stable represents mods)? Obviously I don't know the code of the stable implementation, but I can't imagine its THAT unwieldy.

Syriiin:
I have split this PR into 4 smaller PRs for the sake of backporting to osu-stable in smaller chunks.
- Add a list of mods to `Skill` class (#11687)
- Implement dynamic previous hitobject retention for `Skill` class (#11688)
- Refactor to encapsulate strain logic into `Skill` class (#11689)
    - Refactor to abstract out strain logic into `StrainSkill` class (#11690, branched from #11689)

I'll close this PR here so we can work on merging the parts individually.

PR #11410, opened 2021-01-03
Add change state methods for multiplayer user beatmap availability

frenzibyte:
Required to allow supporting this server-side. Few things to note:
- `BeatmapAvailability` can be promoted into a `ModelAvailability` and get moved into `osu.Game.Online`, but since the usage is only here in multiplayer, I've decided to scope it there.
- Not sure about the pattern I used for the constructors in `BeatmapAvailability`, but I think it should be fine.


osu.Game/Online/Rooms/BeatmapAvailability.cs
```diff
@@ -0,0 +1,56 @@
...
+        /// <param name="state">The beatmap availability state.</param>
+        /// <exception cref="ArgumentException">Throws if <see cref="DownloadState.Downloading"/> was specified in this constructor, as it has its own constructor (see <see cref="BeatmapAvailability(DownloadState, double)"/>.</exception>
+        public BeatmapAvailability(DownloadState state)
+        {
+            if (state == DownloadState.Downloading)
+                throw new ArgumentException($"{nameof(DownloadState.Downloading)} state has its own constructor, use it instead.");
+
+            State = state;
+            DownloadProgress = null;
+        }
```
bdach, line 35: This is all sorts of wrong from an API design standpoint. If you want to do this, do it like this or similarly:

```csharp
private BeatmapAvailability()
{
}

public static BeatmapAvailability NotDownloaded() => new BeatmapAvailability
{
	State = NotDownloaded
};

public static BeatmapAvailability DownloadProgress(double progress) => new BeatmapAvailability
{
	State = Downloading,
	DownloadProgress = progress
};
```

Notice how you *can't* misuse the above without messing with this class.

osu.Game/Online/Rooms/BeatmapAvailability.cs
```diff
@@ -0,0 +1,56 @@
...
+        public BeatmapAvailability(DownloadState state, double downloadProgress)
+        {
+            if (state != DownloadState.Downloading)
+                throw new ArgumentException($"This is a constructor specific for {DownloadState.Downloading} state, use the regular one instead.");
+
+            State = DownloadState.Downloading;
+            DownloadProgress = downloadProgress;
+        }
+
+        public bool Equals(BeatmapAvailability other) => State == other.State && DownloadProgress == other.DownloadProgress;
```
bdach, line 52: Implementation is not safeguarding against `other == null` (not sure how you're not seeing this as an inspection in rider, or how it's not failing CI).

osu.Game/Online/Multiplayer/IMultiplayerRoomServer.cs
```diff
@@ -40,6 +41,13 @@ public interface IMultiplayerRoomServer
         /// <exception cref="NotJoinedRoomException">If the user is not in a room.</exception>
         Task ChangeState(MultiplayerUserState newState);
 
+        /// <summary>
+        /// Change the user's local availability state of the beatmap set in joined room.
+        /// This will also force user state back to <see cref="MultiplayerUserState.Idle"/>.
+        /// </summary>
+        /// <param name="newBeatmapAvailability">The proposed new beatmap availability state.</param>
+        Task ChangeBeatmapAvailability(BeatmapAvailability newBeatmapAvailability);
```
bdach, line 49: how will the user signal that they're done downloading? will they send a `Downloading` message with a `100%` progress? not sure about that design

osu.Game/Online/Multiplayer/MultiplayerRoomUser.cs
```diff
@@ -16,6 +17,11 @@ public class MultiplayerRoomUser : IEquatable<MultiplayerRoomUser>
 
         public MultiplayerUserState State { get; set; } = MultiplayerUserState.Idle;
 
+        /// <summary>
+        /// The availability state of the beatmap, set to <see cref="DownloadState.LocallyAvailable"/> by default.
+        /// </summary>
+        public BeatmapAvailability BeatmapAvailability { get; set; } = new BeatmapAvailability(DownloadState.LocallyAvailable);
```
bdach, line 23: will `Downloaded` be shown as a status to other users? not sure why `DownloadState` was reused here.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
```diff
@@ -311,6 +313,24 @@ Task IMultiplayerClient.UserStateChanged(int userId, MultiplayerUserState state)
...
+        {
+            if (Room == null)
+                return Task.CompletedTask;
+
+            Scheduler.Add(() =>
+            {
+                if (Room == null)
+                    return;
+
+                Room.Users.Single(u => u.UserID == userId).BeatmapAvailability = beatmapAvailability;
```
bdach, line 326: should be tolerant against user potentially not being in the room. we don't really ever care if this fails.

osu.Game/Tests/Visual/Multiplayer/TestMultiplayerClient.cs
```diff
@@ -77,6 +78,14 @@ public void ChangeUserState(int userId, MultiplayerUserState newState)
             });
         }
 
+        public void ChangeUserBeatmapAvailability(int userId, BeatmapAvailability newBeatmapAvailability)
+        {
+            Debug.Assert(Room != null);
+
+            ((IMultiplayerClient)this).UserBeatmapAvailabilityChanged(userId, newBeatmapAvailability);
+            ChangeUserState(userId, MultiplayerUserState.Idle);
```
bdach, line 86: this logic should probably be in `StatefulMultiplayerClient`, as a compound operation. the normal multi client should really be doing this the same way. there shouldn't be any half-and-half; either it's one server call (just `UserBeatmapAvailabilityChanged`, which automagically resets user state back to idle server-side) both in tests and in real usage, or two server calls (`UserBeatmapAvailabilityChanged` + `UserStateChanged`) in both.

bdach (changes requested):

osu.Game/Online/Rooms/BeatmapAvailability.cs
frenzibyte: Just wasn't more creative you could say /shrug

osu.Game/Online/Rooms/BeatmapAvailability.cs
frenzibyte: because I'm guessing you've already changed it to a class on your end :p, it's a `struct` currently

osu.Game/Online/Multiplayer/IMultiplayerRoomServer.cs
frenzibyte: I didn't go through the downloading logic we have to check for that because it's pretty much unnecessary, `100%` `Downloading` means the user downloading progress has reached 100%, `Downloaded` will mean it finished downloading and is in the importing process.

osu.Game/Online/Multiplayer/MultiplayerRoomUser.cs
frenzibyte: `Downloaded` means `Importing`, really thinking about renaming that already.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
frenzibyte: I'm not sure about that, `UserLeft` will remove the user immediately (technically in the main thread but whatever), there shouldn't be any way at all that availability is changed in a user that is not in the room anymore, and sounds like a nasty bug that needs fixing.

osu.Game/Tests/Visual/Multiplayer/TestMultiplayerClient.cs
frenzibyte: The way this is doing in the enclosing method is exactly like the server, propagate availability change and change user state afterward, I could make it such that the "change user state" to idle should be part of the client's job, but I'm not quite sure how sensible that would be (and also actually won't be possible because `IMultiplayerServer.ChangeState` is meant to guard against changing state while beatmap availability not set to `LocallyAvailable)

osu.Game/Online/Rooms/BeatmapAvailability.cs
```diff
@@ -0,0 +1,56 @@
...
+// See the LICENCE file in the repository root for full licence text.
+
+using System;
+
+namespace osu.Game.Online.Rooms
+{
+    /// <summary>
+    /// The local availability information about a certain beatmap for the client.
+    /// </summary>
+    public readonly struct BeatmapAvailability : IEquatable<BeatmapAvailability>
```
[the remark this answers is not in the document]
frenzibyte: I have gone with a `struct` given that this really is just encapsulating two properties, but yeah can convert to class, seems fine.

osu.Game/Tests/Visual/Multiplayer/TestMultiplayerClient.cs
bdach: if receiving an `UserBeatmapAvailabilityChanged` message from the server implicitly means that the user has reverted to idle, then `StatefulMultiplayerClient.UserBeatmapAvailabilityChanged` should be updating the user's state to idle, no? which is another shortcoming of merging the two messages.

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
bdach: this is for display only. if the user has somehow left the room already it's way worse to have this throw an error than not.

osu.Game/Tests/Visual/Multiplayer/TestMultiplayerClient.cs
frenzibyte: no `UserStateChanged` would also be invoked along the state set in `changeAndBroadcastUserState`, before `UserBeatmapAvailabilityChanged` is invoked.

bdach: never mind, got confused by method naming in the test client

osu.Game/Online/Multiplayer/StatefulMultiplayerClient.cs
frenzibyte: hmm, you have a point, I'm fine with your suggestion, but at the very least logging the error such that it goes to sentry would be a better idea, because this sounds like it should never be the case. sounds good?

peppy: if you're going to go that far, you should be arguing that the `room == null` case should also be reported. i don't think either need to be. reporting to sentry should be cases where the game is in a state it cannot continue (ie. something we need to fix). logging things like this just adds noise.

frenzibyte: oh, my bad. I see.

osu.Game/Online/Rooms/BeatmapAvailability.cs
```diff
@@ -20,36 +20,18 @@ namespace osu.Game.Online.Rooms
...
-                throw new ArgumentException($"This is a constructor specific for {DownloadState.Downloading} state, use the regular one instead.");
-
-            State = DownloadState.Downloading;
             DownloadProgress = downloadProgress;
         }
 
-        public bool Equals(BeatmapAvailability other) => State == other.State && DownloadProgress == other.DownloadProgress;
+        public static BeatmapAvailability NotDownload() => new BeatmapAvailability(DownloadState.NotDownloaded);
+        public static BeatmapAvailability Downloading(double progress) => new BeatmapAvailability(DownloadState.Downloading, progress);
+        public static BeatmapAvailability Downloaded() => new BeatmapAvailability(DownloadState.Downloaded);
```
peppy, line 33: what does this state mean? should it maybe be renamed to `Importing` if that's the intended purpose? `Downloaded` sounds like a final state.

peppy (changes requested):

bdach:
I believe [this server-side review comment](https://github.com/ppy/osu-server-spectator/pull/29#discussion_r551006982) is also relevant here and is not yet addressed on either side.

frenzibyte:
> I believe [this server-side review comment](https://github.com/ppy/osu-server-spectator/pull/29#discussion_r551006982) is also relevant here and is not yet addressed on either side.

Addressed changes here, mentioned in that PR that I won't address the reviews there until this is merged and game is bumped, see https://github.com/ppy/osu-server-spectator/pull/29#discussion_r554735285

bdach (dismissed):
seems reasonable

osu.Game/Online/Rooms/BeatmapAvailability.cs
```diff
@@ -0,0 +1,40 @@
...
+        [JsonConstructor]
+        private BeatmapAvailability(DownloadState state, double? downloadProgress = null)
+        {
+            State = state;
+            DownloadProgress = downloadProgress;
+        }
+
+        public static BeatmapAvailability NotDownloaded() => new BeatmapAvailability(DownloadState.NotDownloaded);
+        public static BeatmapAvailability Downloading(double progress) => new BeatmapAvailability(DownloadState.Downloading, progress);
+        public static BeatmapAvailability Importing() => new BeatmapAvailability(DownloadState.Downloaded);
```
peppy, line 33: any reason to not update the underlying enum to match?

frenzibyte: 9 files of change caused by this, is it fine to push now or should this go in another PR?

bdach (approved):

peppy (approved):

PR #11411, opened 2021-01-03
Add ability to toggle discord rich presence

### LINKED ISSUE #5090: Add a toggle for disabling Discord Rich Presence

cillynder:
There are 3 modes: enabled, limited, and disabled.

The limited mode hides identifiable information such as username, rank, and (if participating in one) multiplayer lobby name.

Closes #5090
Supersedes / closes #8258

### DIFF
--- a/osu.Game/Configuration/DiscordRichPresenceMode.cs
+++ b/osu.Game/Configuration/DiscordRichPresenceMode.cs
@@ -7,11 +7,11 @@ namespace osu.Game.Configuration
 {
     public enum DiscordRichPresenceMode
     {
-        Disabled,
+        Off,
 
         [Description("Hide identifiable information")]
         Limited,
 
-        Enabled
+        Full
     }
 }
--- a/osu.Game/Configuration/DiscordRichPresenceMode.cs
+++ b/osu.Game/Configuration/DiscordRichPresenceMode.cs
@@ -1,0 +1,17 @@
+// Copyright (c) ppy Pty Ltd <contact@ppy.sh>. Licensed under the MIT Licence.
+// See the LICENCE file in the repository root for full licence text.
+
+using System.ComponentModel;
+
+namespace osu.Game.Configuration
+{
+    public enum DiscordRichPresenceMode
+    {
+        Disabled,
+
+        [Description("Hide identifiable information")]
+        Limited,
+
+        Enabled
+    }
+}
--- a/osu.Desktop/DiscordRichPresence.cs
+++ b/osu.Desktop/DiscordRichPresence.cs
@@ -9,6 +9,7 @@ using osu.Framework.Allocation;
 using osu.Framework.Bindables;
 using osu.Framework.Graphics;
 using osu.Framework.Logging;
+using osu.Game.Configuration;
 using osu.Game.Online.API;
 using osu.Game.Rulesets;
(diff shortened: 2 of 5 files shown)


osu.Game/Configuration/DiscordRichPresenceMode.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace osu.Game.Configuration
+{
+    public enum DiscordRichPresenceMode
+    {
+        Disabled,
+
+        [Description("Hide identifiable information")]
+        Limited,
+
+        Enabled
```
peppy, line 15: Maybe `Off` and `Full` would be better than `Disabled`/`Enabled`? To fit with the ternary middle value better.

peppy (changes requested):

osu.Game/Configuration/DiscordRichPresenceMode.cs
LavaDesu: On second thought, I think this would look a bit weird to users. Should descriptions be added? Maybe something like this:
- Disabled
- Show only non-identifiable information
- Show all information

bdach: `Off`/`Only non-identifiable information`/`Full` could maybe work? Although "non-identifiable information" is also sort of opaque...

Sometimes services will refer to the "non-identifiable" mode as to "streamer mode" but again that's a little weird to use in this context.

Generally not sure of the usefulness of the "limited" mode. Might as well turn off entirely at that point?

LavaDesu: > Generally not sure of the usefulness of the "limited" mode. Might as well turn off entirely at that point?

I'm mostly implementing up to the linked issue. We could indeed implement only a simple boolean toggle instead of multiple modes, though I personally think that players should be given the option to show off what they're playing without being potentially followed around by random people.

peppy: I think there's definitely an audience for the limited mode.

bdach: Hm, fair enough. But yeah, should keep the names consistent whichever way it goes. It should either be `Hide`/`Show only non-identifiable information`/`Show all information`, or the version I proposed above, or something. Having `Disabled` in there alongside two `Show`s seems super off.

peppy: For sure. Has already been updated to something I can call acceptable, check latest commit.

bdach: Ah yeah I actually missed the fact that that got applied entirely.

peppy (dismissed):

peppy (approved):

bdach (approved):
seems to work as expected

PR #11413, opened 2021-01-03
Fix difficulty adjust settings not being transferred correctly in multiplayer/playlists

### LINKED ISSUE #11305: Difficulty adjust not working in multiplayer

peppy:
Closes https://github.com/ppy/osu/issues/11305

This fixes a number of issues with serialisation and deserialisation of `Mod`/`APIMod` when `ModDifficultyAdjust` is involved. I've aimed to make the minimum number of changes possible to fix mutliplayer/playlists usage. The main issues were:

- Depends on application order of difficulty settings (from the beatmap) and user difficulty settings (from the serialisation), user adjustments could be overwritten by beatmap values as they are not considered tracked. This provides a pathway to forcefully apply and track user adjustments when we are sure they are non-defaults.
- To ensure they are non-defaults, only adjusted values are serialised. Previously the client was sending all values to the server, including map defaults (which was very wrong). Now, only adjusted values are sent.
- Tidied up usage of `GetOrderedSettingsSourceProperties` where it was not assumed that the retrieved properties were `IBindable` (the whole underlying system relies on them being such).

Alternative to the other pathway being explored by @frenzibyte. We may still want to go down that road to better solidify the API (and provide a questionably better user experience via the checkbox model?), but I'm hoping to get this in as a hotfix solution in the mean time.

### DIFF
--- a/osu.Game/Online/API/APIMod.cs
+++ b/osu.Game/Online/API/APIMod.cs
@@ -31,7 +31,12 @@ namespace osu.Game.Online.API
             Acronym = mod.Acronym;
 
             foreach (var (_, property) in mod.GetSettingsSourceProperties())
-                Settings.Add(property.Name.Underscore(), property.GetValue(mod));
+            {
+                var bindable = (IBindable)property.GetValue(mod);
+
+                if (!bindable.IsDefault)
+                    Settings.Add(property.Name.Underscore(), property.GetValue(mod));
+            }
         }
 
         public Mod ToMod(Ruleset ruleset)
--- a/osu.Game/Online/API/APIMod.cs
+++ b/osu.Game/Online/API/APIMod.cs
@@ -35,7 +35,7 @@ namespace osu.Game.Online.API
                 var bindable = (IBindable)property.GetValue(mod);
 
                 if (!bindable.IsDefault)
-                    Settings.Add(property.Name.Underscore(), property.GetValue(mod));
+                    Settings.Add(property.Name.Underscore(), bindable);
             }
         }
 
--- a/osu.Game/Online/API/APIMod.cs
+++ b/osu.Game/Online/API/APIMod.cs
@@ -51,7 +51,7 @@ namespace osu.Game.Online.API
                 if (!Settings.TryGetValue(property.Name.Underscore(), out object settingValue))
                     continue;
 
-                ((IBindable)property.GetValue(resultMod)).Parse(settingValue);
+                resultMod.CopyAdjustedSetting((IBindable)property.GetValue(resultMod), settingValue);
             }
 
             return resultMod;
--- a/osu.Game/Online/API/APIMod.cs
+++ b/osu.Game/Online/API/APIMod.cs
@@ -36,7 +36,7 @@ namespace osu.Game.Online.API
 
         public Mod ToMod(Ruleset ruleset)
         {
-            Mod resultMod = ruleset.GetAllMods().FirstOrDefault(m => m.Acronym == Acronym);
+            Mod resultMod = ruleset.GetAllMods().SingleOrDefault(m => m.Acronym == Acronym);
 
             if (resultMod == null)
                 throw new InvalidOperationException($"There is no mod in the ruleset ({ruleset.ShortName}) matching the acronym {Acronym}.");
--- a/osu.Game/Online/API/APIMod.cs
+++ b/osu.Game/Online/API/APIMod.cs
@@ -41,7 +41,7 @@ namespace osu.Game.Online.API
 
         public Mod ToMod(Ruleset ruleset)
         {
-            Mod resultMod = ruleset.GetAllMods().SingleOrDefault(m => m.Acronym == Acronym);
+            Mod resultMod = ruleset.GetAllMods().FirstOrDefault(m => m.Acronym == Acronym);
 
             if (resultMod == null)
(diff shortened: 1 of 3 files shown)


osu.Game/Online/API/APIMod.cs
```diff
@@ -31,12 +31,17 @@ public APIMod(Mod mod)
             Acronym = mod.Acronym;
 
             foreach (var (_, property) in mod.GetSettingsSourceProperties())
-                Settings.Add(property.Name.Underscore(), property.GetValue(mod));
+            {
+                var bindable = (IBindable)property.GetValue(mod);
+
+                if (!bindable.IsDefault)
+                    Settings.Add(property.Name.Underscore(), property.GetValue(mod));
```
frenzibyte, line 38:
```suggestion
                    Settings.Add(property.Name.Underscore(), bindable);
```

frenzibyte (changes requested):
This seems to fail when:
 - selecting difficulty adjust mod and setting `5` as a custom value in, circle size, for example.
 - creating room
 - going back to song select and come back again.
 - check mod tooltip, it would show `Difficulty Adjust`, instead of `Difficulty Adjust (CS 5.0)`.

Upon debugging, it appears to be that `Default` value is not copied through mods, causing `APIMod` to disregard the custom circle size (example property) setting, because it is set to `5.0`, and while the initial default of every property is `5.0`, `IsDefault` will return true.

Using the `userChangedSettings` dictionary through a method or something (`IsUserAdjusted`) instead of `IsDefault` should solve it I believe.

peppy:
sounds like that may potentially also result in incorrect default display on the controls - is that not the case?

frenzibyte:
It would appear correct in the controls because the mod would be passed to the global `SelectedMods` bindable at that point, which would set `Default` to their proper value, but the `APIMod` is constructed before that happens.

bdach:
That `IsDefault` thing is also why I suggested going down the seemingly insane rabbit hole of changes. As I tried to convey in that other PR thread, with the way difficulty adjust is currently set up to work and display on the UI (one bindable that's supposed to both show base values from a map, but also adapt to user customisation), information on whether or not the user has customised a setting can't be reliably kept "in band" with the actual setting value by using bindable defaults, because they're trying to do multiple things at a time. It needs to be separated somehow.

You could *maaaybe* get close enough, by making it work such that every beatmap change moves the default value of each difficulty setting to be the base difficulty of the map, but IMO that should still be considered temporary, and will probably also be broken in some other way I can't think of right now.

peppy:
The only flaw of the method presented here is that `bindable.Parse` doesn't copy over anything but value. I've changed it to use `Bind`/`Unbind` for now.

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -136,19 +133,35 @@ public virtual Mod CreateCopy()
...
+        /// </summary>
+        /// <param name="bindable">The target bindable to apply the adjustment.</param>
+        /// <param name="value">The adjustment to apply.</param>
+        internal virtual void CopyAdjustedSetting(IBindable bindable, object value)
+        {
+            if (value is IBindable incoming)
+            {
+                //copy including transfer of default values.
+                bindable.BindTo(incoming);
+                bindable.UnbindFrom(incoming);
```
bdach, line 159: missing the fact that the recent framework PR that got merged did not transfer default/disabled was a huge oversight on my part, but i think i'm just going to look away for now to not complicate matters even more than i already have. hopefully most/all of this will get torn out at some point anyway.

osu.Game/Rulesets/Mods/Mod.cs
```diff
@@ -136,19 +133,35 @@ public virtual Mod CreateCopy()
...
             return copy;
         }
 
+        /// <summary>
+        /// When creating copies or clones of a Mod, this method will be called to copy explicitly adjusted user settings.
+        /// The base implementation will transfer the value via <see cref="Bindable{T}.Parse"/> and should be called unless replaced with custom logic.
+        /// </summary>
+        /// <param name="bindable">The target bindable to apply the adjustment.</param>
+        /// <param name="value">The adjustment to apply.</param>
+        internal virtual void CopyAdjustedSetting(IBindable bindable, object value)
```
bdach, line 153: mind if i rename to `target` and `source`? kinda confusing to read otherwise. with the rename at least the rider param name hints help some.

peppy: sounds better, yes.

osu.Game/Rulesets/Mods/Mod.cs
peppy: i'm not 100% sure a call to `Parse` *should* transfer more than value though. that's up for further discussion.

bdach:
made my (tiny) changes, double-check to make sure you're OK with them. let's get this over with for now.

bdach (approved):

osu.Game/Rulesets/Mods/Mod.cs
frenzibyte: yeah, I've intentionally made `Parse` not copy default and disabled because of the above reason, it shouldn't be doing so.

frenzibyte:
Late here but seems to fix it well! 👍🏼

PR #11415, opened 2021-01-04
Fix multiplayer ready button crashing when deleting selected beatmap set

frenzibyte:
Fixes "Ready button crashes when deleting beatmap" card in #11250 

The issue was occurring because `ArchiveModelManager.Delete()` was querying the model from database with only including the file info, and the queried model has been passed directly to `ItemRemoved`, which caused `ReadyButton` to receive a null `Beatmaps`.

~~This fixes it by actually using the right way to query models, which is through `ModelStore.ConsumableItems`.~~

Database-side changes are avoided, as per https://github.com/ppy/osu/pull/11415#issuecomment-753767936, this is fixing the issue by performing a full query on the beatmaps list instead.

### DIFF
--- a/osu.Game/Screens/OnlinePlay/Components/ReadyButton.cs
+++ b/osu.Game/Screens/OnlinePlay/Components/ReadyButton.cs
@@ -39,9 +39,9 @@ namespace osu.Game.Screens.OnlinePlay.Components
             SelectedItem.BindValueChanged(item => updateSelectedItem(item.NewValue), true);
         }
 
-        private void updateSelectedItem(PlaylistItem _) => updateBeatmapState();
-        private void beatmapUpdated(ValueChangedEvent<WeakReference<BeatmapSetInfo>> _) => updateBeatmapState();
-        private void beatmapRemoved(ValueChangedEvent<WeakReference<BeatmapSetInfo>> _) => updateBeatmapState();
+        private void updateSelectedItem(PlaylistItem _) => Scheduler.AddOnce(updateBeatmapState);
+        private void beatmapUpdated(ValueChangedEvent<WeakReference<BeatmapSetInfo>> _) => Scheduler.AddOnce(updateBeatmapState);
+        private void beatmapRemoved(ValueChangedEvent<WeakReference<BeatmapSetInfo>> _) => Scheduler.AddOnce(updateBeatmapState);
 
         private void updateBeatmapState()
         {
--- a/osu.Game/Screens/OnlinePlay/Components/ReadyButton.cs
+++ b/osu.Game/Screens/OnlinePlay/Components/ReadyButton.cs
@@ -53,14 +53,7 @@ namespace osu.Game.Screens.OnlinePlay.Components
 
             var databasedBeatmap = beatmaps.QueryBeatmap(b => b.OnlineBeatmapID == beatmapId && b.MD5Hash == checksum);
 
-            if (databasedBeatmap == null)
-                hasBeatmap = false;
-            else
-            {
-                // DeletePending isn't updated in the beatmap info query above, need to directly query the beatmap set from database as well.
-                var databasedSet = beatmaps.QueryBeatmapSet(s => s.Equals(databasedBeatmap.BeatmapSet));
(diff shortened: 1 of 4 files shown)


peppy:
This fix is too hard for me to parse. It's affecting the database backend (which we're trying to avoid changing) and adjusting file stores which have been working until now.

I propose just doing a full query each time in the ready button to keep things simple (rather than attempting to parse then delete/restore events).

osu.Game/Screens/OnlinePlay/Components/ReadyButton.cs
```diff
@@ -41,38 +39,28 @@ private void load(OsuColour colours)
...
+            var databasedBeatmap = beatmaps.QueryBeatmap(b => b.OnlineBeatmapID == beatmapId && b.MD5Hash == checksum);
 
-            return expression(b => b.OnlineBeatmapID == beatmapId && b.MD5Hash == checksum) != null;
+            if (databasedBeatmap == null)
+                hasBeatmap = false;
+            else
+            {
+                // DeletePending isn't updated in the beatmap info query above, need to directly query the beatmap set from database as well.
+                var databasedSet = beatmaps.QueryBeatmapSet(s => s.Equals(databasedBeatmap.BeatmapSet));
+                hasBeatmap = databasedSet?.DeletePending == false;
```
frenzibyte, line 62: @peppy so that's what I was talking about, for some reason, when the beatmap is restored from deletion, the `QueryBeatmap` call above gives an outdated value of `DeletePending` (true), while `QueryBeatmapSet` here gives the correct value of `DeletePending` (false).

that's why I currently have a query performed on the beatmap info table and a query performed on the beatmap set table, not sure how to fix.

peppy: No test for this PR? Looks like it should be pretty low effort (`ReadyButton` already has good coverage otherwise).

peppy: Testing the bad way (without a test) looks correct at my end?

image

How are you testing / seeing this?

frenzibyte: I'm talking about the "restore from deletion" case, deleting it updates the value yes, but restoring from deletion doesn't.

peppy: Have you investigated why?

frenzibyte: I haven't yet, but I don't think I'll be able to achieve any results, not familiar with this stuff at all currently.

peppy: Either way, adding test coverage would make testing this a lot easier.

peppy (changes requested):
Waiting on test coverage to check why database issues are arising.

frenzibyte:
so I've made a test case for this, but it seems to succeed no matter how many times I run it:
```csharp
            OsuButton readyButton = null;

            AddAssert("ensure ready button enabled", () =>
            {
                readyButton = button.ChildrenOfType<OsuButton>().Single();
                return readyButton.Enabled.Value;
            });

            AddStep("delete beatmap", () => beatmaps.Delete(importedSet));
            AddAssert("ready button disabled", () => !readyButton.Enabled.Value);
            AddStep("undelete beatmap", () => beatmaps.Undelete(importedSet));
            AddAssert("ready button enabled back", () => readyButton.Enabled.Value);
```

after playing with the issue inside lazer, it looks like that issue only occurs when multi-threaded, so it appears to be threading-related, I'll try investigating more.

peppy:
This is due to ef state tracking and the fact we are actually not doing it correctly. I'll take a look and see if it's a relatively simple fix.

peppy:
I've removed the weird workaround you added. It's fine to let it be incorrect in the said scenario (if anything, it should be tracked as a separate issue). You can find that song select struggles in similar ways due to the same underlying issue.

peppy (approved):
Seems fine for now.

PR #11419, opened 2021-01-04
Fix RemoveBlockingOverlay causing transform mutation from disposal threads

peppy:
This is just one of multiple cases of transform mutation from incorrect threads that I unearthed (causing enumeration-on-modified-collection exceptions).

### DIFF
--- a/osu.Game/OsuGame.cs
+++ b/osu.Game/OsuGame.cs
@@ -151,11 +151,11 @@ namespace osu.Game
             updateBlockingOverlayFade();
         }
 
-        public void RemoveBlockingOverlay(OverlayContainer overlay)
+        public void RemoveBlockingOverlay(OverlayContainer overlay) => Schedule(() =>
         {
             visibleBlockingOverlays.Remove(overlay);
             updateBlockingOverlayFade();
-        }
+        });
 
         /// <summary>
         /// Close all game-wide overlays.


osu.Game/OsuGame.cs
```diff
@@ -151,11 +151,11 @@ public void AddBlockingOverlay(OverlayContainer overlay)
             updateBlockingOverlayFade();
         }
 
-        public void RemoveBlockingOverlay(OverlayContainer overlay)
+        public void RemoveBlockingOverlay(OverlayContainer overlay) => Schedule(() =>
```
bdach, line 154: Not sure about doing this here instead of locally. It raises questions about why `AddBlockingOverlay()` is not scheduled but this is, which I guess can be answered by the commit message, but it still sorta feels like this would be better placed in `OsuFocusedOverlayContainer.Dispose()` - as I'm guessing that is the only infringing usage? `UpdateState` should be safe because `UnbindAllBindables()` runs first, right?

peppy: Not sure. That method exists only for that one class so it's a bit arbitrary where it's placed.

bdach: Hm, fair enough. Slipped by past me that the schedule uses the implicit `this` which is an `OsuGame`.

bdach (approved):

Issue #10049, opened 2020-09-03
Dashboard description goes off-screen


Gabixel:
It should probably "expand" the description to the left rather than the right, like the "now playing" one.

**Screenshots or videos showing encountered issue:** 
![dashboard](https://user-images.githubusercontent.com/43073074/92128480-bb6c3e80-ee02-11ea-8e7c-17b8bc315a43.png)

**osu!lazer version:** 2020.903.0

peppy:
I saw this while implementing, but disagree that it should just go to the left, unless we apply that to *all* of the icons on the right hand side of the toolbar.

peppy:
Closed via #11420.

Joehuu:
Fixed in https://github.com/ppy/osu/pull/11420.

PR #11420, opened 2021-01-04
Fix right aligned toolbar button tooltips leaving the screen

RizonFTW:
This will change it so it doesn't go off screen. If you want me to try something else to solve this issue let me know. 

Issue: #10049


bdach:
I don't know whether the order of the toolbar buttons was explicitly intended, but changing their order to fix a text overflow issue sounds like a pretty shoddy fix. You should probably be looking to align all tooltip texts to the right, as suggested in the issue thread.

peppy (approved):

